PR #63, opened 2011-12-08
Enhance `stream-map' to work with more than one list

danking:
Stream map should work on multiple lists. From the tests:

```
(define fibonacci
  (stream-cons 0 (stream-cons 1 (stream-map +
                                            fibonacci
                                            (stream-rest fibonacci)))))
```


samth:
@danking I'm going to close this; if you want to make this work again, feel free to re-open.

PR #552, opened 2014-01-17
Fix misspelling of scheme_detach_multple_array

takikawa:
Fixes a misspelling in the C API. The reason I'm submitting a PR is in case anyone thinks we should leave the old name in for backwards compatibility (this function has existed for about six years). I'm leaning against since it's been documented with the correct name for all that time.

Relevant people: @mflatt


pmatos:
@takikawa ping

samth:
Now that Racket CS is likely to become the default soon, this seems less valuable, so I'm closing.

PR #617, opened 2014-04-14
Add sequence support for id-tables

deeglaze:
An obvious deficiency now patched for id-tables.


pmatos:
@rmculpepper re-ping?

soegaard:
@samth @rmculpepper 

This one seems to have fallen into the cracks.

PR #636, opened 2014-05-02
New Indentation Rules Implemented

LeiWangHoward:
1) Indentation rules applied to count-paren function:
Count number of parenthesis before given position till the outmost '@'
annotation, if the there is '[', we check if it has '@' right after it within the
same line, if so, we add the number of characters between this '[' and the
beginning of the line it appears to the total parenthesis counting
2) Line push back rules:
We do not push back lines begin with '@ that
a) Has keyword 'codeblock' or 'verbatim' after
b) Contains '[' and multiple lines between it and the closing ']'


rfindler:
Sorry

PR #639, opened 2014-05-07
Allow caching to be disabled for individual metafunctions.

jrslepak:
A metafunction with the #:cache-poison keyword will not have its results
cached. Other metafunction results which depend on an uncached
metafunction are also not cached. Some of the parsing logic in
split-out-contract is changed to allow the keyword to be checked more
easily. The metafunction structure has an additional boolean field
indicating whether its results should be cached.

### DIFF
--- a/pkgs/typed-racket-pkgs/typed-racket-lib/typed-racket/base-env/prims.rkt
+++ b/pkgs/typed-racket-pkgs/typed-racket-lib/typed-racket/base-env/prims.rkt
@@ -139,7 +139,7 @@ This file defines two sorts of primitives. All of them are provided into any mod
                 [syntax/define (normalize-definition)]
                 [typed-racket/private/parse-type (parse-type)]
                 [typed-racket/private/type-contract (type->contract type->contract-fail)]
-                [typed-racket/env/type-name-env (register-type-name)]))
+                [typed-racket/env/type-alias-env (register-resolved-type-alias)]))
 
 (define-for-syntax (with-type* expr ty)
   (with-type #`(ann #,expr #,ty)))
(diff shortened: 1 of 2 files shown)


samth:
@jrslepak @rfindler has the status of this changed since May?

rfindler:
I don't think so. Sorry I have not had time to do this myself.

Robby

samth:
@rfindler I've assigned this to you (just to keep track of who should review it when it's ready).

howell:
I found myself wanting this today, is there any hope for reviving this?

rfindler:
@howell: can you say more about the model where you found you needed it and why you needed it?

I am getting even more hesitant to add it because it is a performance maintence burden in the future and it doesn't seem like it translates well into something you'd typeset in a paper. So I'd like to hear more about why you wanted to use it to better understand why it seems useful to you.

rfindler:
Oh: and I just now noticed that this isn't on the redex repo. @samth: do you know if that's fixable?

And one other request for @howell: would it work for your usecase if the poisoning was syntactic instead of dynamic? So if one metafunction calls a poisoned one, then it wouldn't be cached, but otherwise it would (use the current-cached-metafunctions parameter)?

samth:
https://github-issue-mover.appspot.com/ is the best thing you can do for
moving issues.

rfindler:
Looks like it requires authorization. Is that something we can revoke after doing the move?

samth:
Yes, you can.

rfindler:
Okay, that didn't work. How do I deauthorize it?

samth:
In your github settings.

howell:
My model is pretty similar to STLC. I wanted something like this when I implemented type inference to generate fresh type variables. I don't think `variable-not-in` is suitable because I want the variables to be unique even across parts of the term I don't currently have a handle on. My next thought was to have a `fresh` metafunction that uses racket `gensym`, which is when I posted here. Ultimately I wound up calling out to gensym without going through a metafunction.

And to answer the other question, I think a static/syntactic approach would suffice.

rfindler:
Ah, yes. That can be problematic. In this particular case, however, you do not need the "poisoning" aspect, right? It would be enough to simply disable caching for the one metafunction that creates the fresh variables?

Also: did you consider using `#:binding-forms` ?

howell:
> Ah, yes. That can be problematic. In this particular case, however, you do not need the "poisoning" aspect, right? It would be enough to simply disable caching for the one metafunction that creates the fresh variables?

In my case the metafunction would only be called from the type judgment, so I believe this is correct.

> Also: did you consider using #:binding-forms ?

I belatedly wondered if using this might have made things work out right auto-magically. I'm not very familiar with it, though (I haven't used it in a model yet). I'll take a closer look and see.

howell:
After perusing the docs, I'm not sure `#:binding-forms` will help me because there is no binding form for type variables. (This is STLC).

rfindler:
Why do you need variables to type check STLC?

howell:
for type inference.

rfindler:
We're a bit off topic here, but STLC doesn't require inference. Anyway.

PR #650, opened 2014-05-21
Cleanup tc/funapp and make it use infer like tc/apply

endobson:


samth:
@shekari can you open this as a PR on the racket/typed-racket repo?

sorawee:

PR #671, opened 2014-06-10
Make generated struct contracts lazy.

endobson:
This makes it so generated struct contracts are lazy instead of strict. This makes contract checking O(1) instead of having to check the entire data structure on every pass through the contract boundary. The slow down is a factor of 60 for simple checks since chaperone structs are slower than flat structs on the constant factor.


sorawee:

PR #682, opened 2014-06-21
Split HashTable out into mutable and immutable kinds.

deeglaze:
There are some inference problems making the marketplace tests fail. I'll need same help fixing that up.


bennn:
Closed by https://github.com/racket/typed-racket/pull/559

PR #691, opened 2014-06-28
Many changes to avoid generic equals.

endobson:
This is not ready for merge, but I wanted to put it out for comment. I found may places where we were using generic equals and replaced them with type specific ones. I found a small (2-3%) improvement in timing of for-vector. I think that it is worth keeping the changes that do not make the code harder to read, but not on the others. Thoughts?


sorawee:

PR #701, opened 2014-07-01
Make subst-gen not require having all constraints bound.

endobson:
Removing this requirement means that we can produce more minimal csets when adding expected csets in later changes.


sorawee:

PR #702, opened 2014-07-01
Push expected constraints all the way down

endobson:


sorawee:

PR #705, opened 2014-07-02
Add test for mu and invariant recursive calls.

endobson:
This case used to generate a substitution and now doesn't. I believe this is the correct behavior because the substitution generated before (b2 = Symbol) does not correctly unify with the first item because of the unfolding inside the vector.

Can someone doublecheck my reasoning?


sorawee:

PR #706, opened 2014-07-02
Add aliases to lexical environment

shekari:
This adds support for aliases to the lexical environment. Currently only support for entire aliases is implemented but the scafolding for support of subparts is there. Subparts would be support for understanding that
in
```
(define x (car y))
(define z (car y))
```
`x` and `z` are aliases of each other, since they both refer to `(make-Path (list (make-CarPE)) #'y)`.


samth:
This ended up being done differently by @andmkent in racket/typed-racket#2.

PR #709, opened 2014-07-05
Make struct overlap work correctly.

endobson:
This does 3 things.
1. Make remove intersect handle overlapping structs correctly.
2. Fix resulting breakage in restrict, by making restrict understand how
to instantiate type variables when restricting to a substruct.
3. Prevent infinite loops that occur when trying to acheive #2.


sorawee:

PR #711, opened 2014-07-06
Improve types of rest and cons.

endobson:
Add accessor to second case in rest type.
Make cons not widen list types.
Introduces a slight backwards incombatibility in how cons can be
instantiated.

Closes PR 14457.

The backwards incombatibility of cons seems like it is unavoidable because the previous type of cons just ignored the second type argument and dropped all structure from list arguments. I think all necessary changes will be like for little schemer, where the instantiation is no longer needed.


sorawee:

PR #712, opened 2014-07-06
Fix or/sc optimization with non flat contracts.

endobson:
or/sc cannot optimize away all contracts if one is any/sc, because some
of the values passing through the contract barrier may still need the
protection of the other contracts.

Closes PR 14581.


bennn:
https://github.com/racket/typed-racket/commit/9df037b0f6e6b2df3cc0993631bb2870367e291a for a conservative fix to the optimization

PR #729, opened 2014-07-12
Add `prop:object-name`.

samth:
This adds `prop:object-name`, as discussed [on the users list](http://lists.racket-lang.org/users/archive/2014-April/062300.html). 

This removes the invariant that object names are always symbols.


pkgs/racket-doc/scribblings/reference/struct-inspectors.scrbl
```diff
@@ -169,3 +172,22 @@ example).
...
+ procedure @racket[_proc] receives the structure as an argument, and the result
+ of @racket[_proc] is the @racket[object-name] of the structure.}
+
+ @item{An exact, non-negative integer between @racket[0] (inclusive) and the
+ number of non-automatic fields in the structure type (exclusive, not counting
+ supertype fields): The integer identifies a field in the structure, and the
+ field must be designated as immutable. The value of the field is used as the
+ @racket[object-name] of the structure.}
+]
+@history[#:added "6.1.1.5"]
```
AlexKnauth: should this be 6.1.1.9?

samth:
@mflatt Does this look reasonable now?

mflatt:
@samth: looks good

PR #732, opened 2014-07-15
Struct-info extended with field name information, match extended with named field patterns.

deeglaze:
Split into two commits to separate the concerns.


samth:
@sorawee Is there still something here that is needed, given your recent changes?

sorawee:
I still have not applied the `struct*` change, so no, let's keep it open for now.

PR #756, opened 2014-08-20
Add missing compare of last word in `equal-proc`.

greghendershott:
See http://lists.racket-lang.org/users/archive/2014-August/063941.html


AlexKnauth:
should this pull request be closed?

PR #758, opened 2014-08-20
add pattern-expanders to syntax/parse

AlexKnauth:


stchang:
Just used this. Works great, thanks! I did spend a lengthy debugging session to learn that pattern expander names *must* have a `~` prefix. I guess it's a limitation of `syntax-parse` so I don't know if anything can be done about it. For now, I added a note to the docs.

PR #815, opened 2014-11-05
add identifier sets (in syntax/id-set)

stchang:
Second attempt at adding identifier sets (some previous discussion here: https://github.com/plt/racket/pull/807).

If there are no complaints, I would like to merge. @rmculpepper and @carl-eastlund, would you mind taking a look if you have time?


stchang:
Merged in with 7d434d266e10744d27123ad91c75502e8951d488

PR #844, opened 2014-12-08
add define-syntax-parser

AlexKnauth:
idea from
https://github.com/jackfirth/generic-bind/commit/863d39229fb42395face91de733870c0d02f0922#diff-3252674930bbd0c4e113856a2a3a5747R118
```racket
(define-syntax-parser my-or
  [(or) #'#f]
  [(or e1:expr e:expr ...)
   #'(let ([v e1])
       (if v v (or e ...)))])
```


samth:
@rmculpepper Thoughts?

rmculpepper:
Merged, thanks!

PR #855, opened 2015-01-03
Mark more errors during optimization

gus-massa:
These small commits are independent but related, so they can be squashed before merging


mflatt:
Merged - thanks!

PR #857, opened 2015-01-07
index-scope.rkt: convert find-user-pkgs-dir to a simple-form-path

bdeket:
find-relative-path expects a simple-form-path, but according to the documentation PLTUSERHOME as propagated by find-user-pkgs-dir must only be a complete-path?
Without this building of the documentation fails if PLTUSERHOME contains ".."


mflatt:
I've merged your repair - thanks!

PR #858, opened 2015-01-07
Fix typos in filesystem.scrbl

gus-massa:
Fix typos in the documentation of `call-with-atomic-output-file` and `find-executable-path`.

I don't like the names of the arguments of `find-executable-path`. The last commit change them.


mflatt:
Merged - thanks!

PR #860, opened 2015-01-16
Add contracts (and chaperones/impersonators) for asynchronous channels

lexi-lambda:
This adds an `async-channel/c` contract, as well as `chaperone-async-channel` and `impersonate-async-channel` procedures. The `async-channel/c` contract is a contract combinator that takes a contract as a parameter and performs contract checking on the values placed into or received from the channel.

The motivation behind this change—in addition to it simply being a useful tool to have—is to allow Typed Racket to generate contracts for `Async-Channelof` types to facilitate typed-untyped interaction with code that makes use of asynchronous channels.

In order to support these modifications, the `async-channel` struct has been restructured to use generics on its various procedures, allowing these to be impersonated using the `impersonate-generics` function. I also took the liberty to make use of the new `struct` form to simplify the struct definition instead of using `make-struct-type`.

This implements all changes in the `racket/async-channel` module, which is where the accompanying documentation is located. It also adds some unit tests to check the validity of the contracts.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/async-channels.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/async-channels.scrbl
@@ -11,6 +11,8 @@
 
 @note-lib-only[racket/async-channel]
 
+@section{Creating and Using Asynchronous Channels}
+
 @margin-note/ref{See also @secref["threadmbox"].}
 
 An @deftech{asynchronous channel} is like a @tech{channel}, but it buffers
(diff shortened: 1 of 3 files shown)


pkgs/racket-doc/scribblings/reference/async-channels.scrbl
```diff
@@ -105,5 +107,58 @@ limit); @resultItself{asychronous channel-put event}.}
...
+The @racket[put-proc] must accept the value passed to @racket[async-channel-put]
+called on @racket[channel]; it must produce a replacement value, which is the
+value passed to the put procedure called on the original channel.
+
+The @racket[get-proc] and @racket[put-proc] procedures are called for all
+operations that get or put values from the channel, not just
+@racket[async-channel-get] and @racket[async-channel-put].
+
+Pairs of @racket[prop] and @racket[prop-val] (the number of arguments to
+@racket[impersonate-async-channel] must be odd) add impersonator properties or
```
samth: Can you use a `@tech` reference for impersonator properties? I'm not sure what the right name is, though.

racket/collects/racket/async-channel.rkt
```diff
@@ -147,9 +136,143 @@
...
+    (let ([pos-elem-proj ((contract-projection elem-ctc) blame)]
+          [neg-elem-proj ((contract-projection elem-ctc) (blame-swap blame))])
+      (check-async-channel/c ctc val blame)
+      (impersonate/chaperone-async-channel val
+                                           (λ (v) (pos-elem-proj v))
+                                           (λ (v) (neg-elem-proj v))
+                                           impersonator-prop:contracted ctc
+                                           impersonator-prop:blame blame))))
+
+(define-struct base-async-channel/c (content))
```
samth: Why use `define-struct` and not `struct`?

lexi-lambda: @samth ...because the code I copied this from used it and I forgot to update it. I'll change it, thanks.

lexi-lambda:
@samth Implemented those changes.

racket/collects/racket/async-channel.rkt
```diff
@@ -147,9 +136,147 @@
...
+
+(define ((ho-projection impersonate/chaperone-async-channel) ctc)
+  (let ([elem-ctc (base-async-channel/c-content ctc)])
+    (λ (blame)
+      (let ([pos-elem-proj ((contract-projection elem-ctc) blame)]
+            [neg-elem-proj ((contract-projection elem-ctc) (blame-swap blame))])
+        (λ (val)
+          (check-async-channel/c ctc val blame)
+          (impersonate/chaperone-async-channel val
+                                               (λ (v) (pos-elem-proj v))
```
samth: Why the extra `lambda` here?

lexi-lambda:
@samth You were right, that was unnecessary. I've removed the extra lambdas.

PR #862, opened 2015-01-21
Correct 'german time format

mbertheau:
We separate hour and minute with a colon, not with a period.


rfindler:
Thanks, I've pushed the commit to the racket repo.

mbertheau:
@rfindler great, thanks :)

PR #864, opened 2015-01-24
Add docs of unsafe-fxvector

gus-massa:
The first commit adds the docs of `unsafe-fxvector`.

The second changes the type of the result of `unsafe-fxvector-ref` from `any/c` to `fixnum?`. It also does the equivalent changes in `unsafe-flvector-ref` and `unsafe-extflvector-ref`. I'm not sure because it's easy (but risky and not very wise) to use the unsafe operations to store anything in a `fxvector` and retrieve it later (if you are lucky and there has been no garbage collection).

The third adds the `SCHEME_PRIM_PRODUCES_FIXNUM` flag to `unsafe-fxvector-ref`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/unsafe.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/unsafe.scrbl
@@ -282,6 +282,18 @@ fixnum).}
 
 
 @deftogether[(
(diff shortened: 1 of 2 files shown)


mflatt:
Merged, and much appreciated!

PR #867, opened 2015-01-27
Add writeln and println functions to complement displayln

lexi-lambda:
This is a fairly simple addition. It just adds `writeln` and `println` functions which simply do `write` or `print` followed by `newline`, respectively, similar to `displayln`'s relationship to `display`.

Considering `displayln` already exists, and I frequently find myself writing `(printf "~v\n" foo)` for debugging purposes, I think this would be a convenient addition to the language.


soegaard:
I think write feels left out :-)

lexi-lambda:
@soegaard True. Perhaps I should add `writeln` while I'm at it? :)

soegaard:
I'd like that.

lexi-lambda:
@soegaard Okay, I added `writeln` as well. I also tweaked the `println` documentation a little to indicate that users looking for an equivalent to Java's `println` should use `displayln` instead, to avoid confusion.

samth:
I think this needs some tests.

lexi-lambda:
I'm going to close this, not because I don't want to add test, but because I don't think this is actually all that important. For debugging, using `pretty-print` makes much more sense anyway, making this less relevant.

samth:
I actually think we should do this -- I've missed these functions before.

samth:
I've merged this.

PR #868, opened 2015-01-27
More tests for extflonum

gus-massa:
This is something between a bug report and a pull request.

Racket has a minimal support to read and write `extflonum`s when the `exflonums`
are not available. In this configuration they use a different path code, so it's
necessary to test this version of the code too.

After these commits I get: 

In Linux (Travis) I get these errors:

    ((extflonum) ("007.000t000" "7.0t0" (format "~a" 007.000t000)))
    ((extflonum) ("-007.000t000" "-7.0t0" (format "~a" -007.000t000)))

I'm not sure if they are error or strange behavior, but the result is different in Windows.

In Windows 7 (32 bits) with Visual Studio 2008 Express, if I cut the build just after the RacketCGC.exe and Racket.exe appear I get: 

    ((extflonum) ("0.0t0" "+inf.t" (#<procedure:format> "~a" 0.0t0)))
    ((extflonum) ("0.0t0" "-inf.t" (#<procedure:format> "~a" 0.0t0)))
    ((extflonum) ("0.0t0" "+nan.t" (#<procedure:format> "~a" 0.0t0)))

[It should be

    ((extflonum) ("0.0t0" "+inf.t" (#<procedure:format> "~a" +inf.t)))
    ((extflonum) ("0.0t0" "-inf.t" (#<procedure:format> "~a" -inf.t)))
    ((extflonum) ("0.0t0" "+nan.t" (#<procedure:format> "~a" +nan.t)))

But the problem is precisely that the print procedure has a problem.]

The problem disapears after the build is completed. I tried to reproduce the error in AppVeyor but I was unsuccessful. I thought it was a problem with `--no-foreign-libs` but now I now I think it's some kind of configuration after the creation of Racket.exe. (`--only-foreign-libs`?)

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -9,6 +9,7 @@ build_script:
 - '"c:\Program Files (x86)\Microsoft Visual Studio 10.0\vc\vcvarsall.bat" x86'
 - echo %cd%
 - nmake win32-in-place PKGS="racket-test-core"
+- nmake win32-base
 
 test_script:
 - echo %cd%
--- a/pkgs/racket-test-core/tests/racket/all.rktl
+++ b/pkgs/racket-test-core/tests/racket/all.rktl
@@ -12,6 +12,7 @@
 (load-in-sandbox "dict.rktl")
 (load-in-sandbox "fixnum.rktl")
 (load-in-sandbox "flonum.rktl")
+(load-in-sandbox "extflonum.rktl")
 (load-in-sandbox "string.rktl")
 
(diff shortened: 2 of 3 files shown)


gus-massa:
I successfully misconfigured the Windows build to get a failure in AppVeyor :).

This builds first the `win32-at-place` version to install the packages, but it doesn't run the tests. Then it builds the `win32-base` version (that deletes a few dll?) and runs the tests.

mflatt:
merged the test changes - thanks!

PR #872, opened 2015-02-08
Add stream/c contract to racket/stream

lexi-lambda:
This adds a `stream/c` contract to the `racket/stream` module. It is a contract combinator which lazily checks the contents of a stream based on the parameter provided to `stream/c`. It works with both structures that implement `gen:stream` and ordinary lists, which satisfy `stream?`. Ordinary lists are checked strictly, not lazily, as if they had been checked with the `listof` contract.

As with my addition of `async-channel/c`, this addition should allow `racket/stream` to be effectively ported to Typed Racket, fixing [PR 13523](http://bugs.racket-lang.org/query/?debug=&database=default&cmd=view+audit-trail&cmd=view&pr=13523).

Alongside adding `stream/c`, this change also adds the relevant documentation, and it includes unit tests for ensuring the correctness of the contracts.


racket/collects/racket/stream.rkt
```diff
@@ -179,3 +184,129 @@
...
+                           val '(expected "a stream" given: "~e") val))))
+
+(define (stream/c-stronger? a b)
+  (contract-stronger? (base-stream/c-content a) (base-stream/c-content b)))
+
+; streams are lazy, so we need to contract the rest of the stream lazily (which can be a list)
+(define (contract-stream-rest v ctc blame)
+  (define elem-ctc (base-stream/c-content ctc))
+  (define new-ctc (if (list? v) (listof elem-ctc) ctc))
+  (contract new-ctc v
```
rfindler: I don't think you should call 'contract' here. Just call directly into the projection (or a helper function the projection also uses)

racket/collects/racket/stream.rkt
```diff
@@ -179,3 +184,129 @@
...
+                 (λ (v) (contract-stream-rest v ctc blame))
+                 impersonator-prop:contracted ctc
+                 impersonator-prop:blame (blame-add-missing-party blame neg-party))))))))
+
+(define ((ho-projection impersonate/chaperone-stream) ctc)
+  (let ([elem-ctc (base-stream/c-content ctc)])
+    (λ (blame)
+      (let ([pos-elem-proj ((contract-projection elem-ctc) blame)]
+            [neg-elem-proj ((contract-projection elem-ctc) (blame-swap blame))])
+        (λ (val)
```
rfindler: Use blame-add-context (or one of its variants) here. 

To get a sense of what to put, do something like (list/c (stream/c (box/c none/c))) and look at the error message. You should seem some context messages there. (Let me know if that's not enough of a hint.)

rfindler: Oh I see you're already using it above.

rfindler:
This looks good to me overall. Small note inside.

racket/collects/racket/stream.rkt
rfindler: (the reason not to call contract is you want to use the same blame record, not make a new one)

lexi-lambda: @rfindler Unless I'm missing something about how contracts work, I think that attaching a new contract is necessary, since streams are lazy, and different elements of the stream are fundamentally different objects.

Because of that, contracts need to be assigned both *lazily* and deeply. Therefore, `(stream-rest s)` needs to have a new contract attached to it so that further calls to `stream-rest` on the same stream still include the contract.

rfindler: Yes that is true. But you don't call "contract". You have some contract and
you get its projection and call that.

lexi-lambda:
@rfindler Okay, after thinking about it a little more, I realized what you were saying. I've fixed that. I've also removed the `val-first-projection` implementation because I don't think it matters here, so it just seemed to make things more complicated.

rfindler:
Sounds good.

lexi-lambda:
@rfindler Can I get this merged?

lexi-lambda:
@rfindler Removed the unnecessary `neg-elem-proj`.

rfindler:
Thanks, I've pushed the commit.

samth:
@lexi-lambda turns out there's a failure here: https://travis-ci.org/plt/racket/jobs/51157234#L1521

I think the problem is that PR 14970 has partly fixed, in 2ada651dd3, and thus the use of `struct-info`, where you note the issue w/ that PR, no longer works, but I'm not 100% sure.

@mflatt @rfindler more thoughts?

rfindler:
I'm not really sure, but if someone else doesn't get time to look into it, I can try

mflatt:
I think the solution might be to extend `chaperone-generics` and `impersonate-generics` to accept properties, since they are in control of the relevant representation. Then, `impersonate-stream` and `chaperone-stream` can just pass along the properties, instead of using `impersonate-struct` and `chaperone-struct` directly.

lexi-lambda:
Yes, the entire reason I used `impersonate-struct` in the first place was because `impersonate-generics` doesn't support impersonator properties. I think changing that would be the most logical fix.

mflatt:
@lexi-lambda I started changing `impersonate-generics`, but are you already working on it?

mflatt:
Done and merged.

PR #874, opened 2015-02-10
Link to the guide and reference source locations

hoelzro:
I had a lot of trouble figuring out how to modify the guide to fix a typo, so I figured a link in the guide itself would help others.  I added one to the reference too!

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/guide.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/guide.scrbl
@@ -15,6 +15,9 @@ Racket. From @seclink["datatypes"]{Chapter 3} on, this guide dives
 into details---covering much of the Racket toolbox, but leaving
 precise details to @|Racket| and other reference manuals.
 
+If you would like to contribute to the maintenance of this guide,
+you can find it the source on @hyperlink[https://github.com/plt/racket/tree/master/pkgs/racket-doc/scribblings/guide]{GitHub}.
+
 @table-of-contents[]
 
 @include-section["welcome.scrbl"]
--- a/pkgs/racket-doc/scribblings/reference/reference.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/reference.scrbl
@@ -49,6 +49,9 @@ Unless otherwise noted, the bindings defined in this manual are
 exported by the @racketmodname[racket/base] and @racketmodname[racket]
 languages.}
 
+If you would like to contribute to the maintenance of this reference,
+you can find it the source on @hyperlink[https://github.com/plt/racket/tree/master/pkgs/racket-doc/scribblings/reference]{GitHub}.
+
 @margin-note{The @racketmodname[racket/base] library is much smaller than
 the @racketmodname[racket] library and will typically load faster.
 


samth:
@mflatt I saw your comment after merging this. I've answered this question enough times in the IRC channel that I'm inclined to deal with fixing the links when sources change, but if you have another suggestion I'm happy to change it.

mflatt:
In the case of the self link in the Scribble docs, it has been a problem that the release gets out of sync with the development version; changing the source doesn't affect the docs that lots of people are reading.

I think that a "contribute via ..." link makes sense imposed on installation, similar to the way the Racket banner is added at http://docs.racket-lang.org, instead of having the path built into the source and rendered docs. I imagine that a JavaScript fragment could add a suitable path onto the page; the JavaScript to add the path might be updated daily by consulting pkgs.racket-lang.org. The rendered docs need to record the source of the document for use by the script; fortunately, they already do that (to support the information you see when you click on a section title).

samth:
Are you imagining that the JS would be served from some location that gets modified daily? I bet it would be easier to just consult pkgs.racket-lang.org on every load of the page, since that information is hosted statically anyway on the server.

mflatt:
I had in mind that pkgs.racket-lang.org doesn't currently provide a compact mapping from module names to package sources. But adding and updating that information at pkgs.racket-lang.org (and referencing it directly in the script) sounds right.

samth:
Ah, right, because the current static table has the full data. @jeapostrophe or @tonyg, how hard would it be to serve http://pkgs.racket-lang.org/racket-doc as a JSON document with the relevant data for that pkg?

jeapostrophe:
It already is

http://pkgs.racket-lang.org/pkg/racket-doc.json

On Tuesday, February 10, 2015, Sam Tobin-Hochstadt <notifications@github.com>
wrote:

> Ah, right, because the current static table has the full data.
> @jeapostrophe <https://github.com/jeapostrophe> or @tonyg
> <https://github.com/tonyg>, how hard would it be to serve
> http://pkgs.racket-lang.org/racket-doc as a JSON document with the
> relevant data for that pkg?
>
> —
> <https://github.com/plt/racket/pull/874#issuecomment-73715883>.
>


-- 
Jay McCarthy
http://jeapostrophe.github.io

                          - D&C 64:33

samth:
Even better!

mflatt:
@jeapostrophe: I don't think that's the format that we need. The example path you provided embeds the package name, but the initial problem is mapping a document-source module path to a package name. The catalog overall contains that information, but not in a compact and easily searchable (in JS) format.

samth:
@jeapostrophe you need to allow cross-origin requests in order to make this work -- just adding the header `Access-Control-Allow-Origin: *` should fix it.

samth:
@mflatt I think we can embed the pkg name in the source, unless you're worried about the pkg that provides the reference changing names?

mflatt:
@samth: Yes, I did have in mind trying to support documents that move to different packages. Maybe it's good enough to embed the package name in rendered document, though.

mflatt:
I've added `x-source-pkg` to Scribble's HTML rendering

bennn:
https://github.com/racket/scribble/pull/223 is trying to solve the same problem, but embeds more than the package name. (I need to read the comments PR more carefully.)

PR #875, opened 2015-02-10
Fix a typo in define-cstruct

BartAdv:
This prevented correct message to be outputted in the case of multiple `#:malloc-mode` specifications

### DIFF
--- a/racket/collects/ffi/unsafe.rkt
+++ b/racket/collects/ffi/unsafe.rkt
@@ -1627,7 +1627,7 @@
                        [(#:malloc-mode m . rest) 
                         (not malloc-mode)
                         (loop #'rest alignment #'m properties property-bindings no-equal?)]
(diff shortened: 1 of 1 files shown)


BartAdv:
That is rather small issue, but I've figured out I could add test case here anyway, as I noticed there are test cases for those options. However, I noticed those test cases could yield false positives, due to `syntax-test`:

https://github.com/plt/racket/blob/master/pkgs/racket-test-core/tests/racket/cstruct.rktl#L75-L76

Both those linex yield success, however, both tested cases are in fact reporting an error message not related to what they are supposed to test (if I'm reading this correctly, that is). Because of this, when I add another test case for that small fix I've made (before applying it, of course), tests report nothing.

mflatt:
Merged (with test corrections) - thanks!

PR #877, opened 2015-02-11
Clarify dynamic-require docs and add examples

takikawa:
The current docs for `dynamic-require` imply that the function may return multiple values if the binding is defined as syntax and if the syntax expands to multiple values.

It turns out that the implementation restricts the evaluation to a single value. I've adjusted the docs to specify that explicitly, which I think makes more sense than changing the behavior of the function.


mflatt:
Looks good to me, so I merged - thanks!

PR #878, opened 2015-02-12
Re-do projections for struct fields in chaperone handlers.

samth:
If the selector is itself a chaperone, then doing the access once will mean
that the saved result is not `chaperone-of?` the result of doing the access a
second time, at least in some cases (such as when the accessor uses vector
contracts).

- [x] run the tests
- [x] add more tests

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/struct-dc.rkt
+++ b/pkgs/racket-test/tests/racket/contract/struct-dc.rkt
@@ -178,6 +178,38 @@
      (and (exn:fail:contract:blame? x)
           (regexp-match #rx"promised: foo?" (exn-message x)))))
   
+  (test/spec-passed/result
+   'struct/c-contract-accessor
+   '(begin
+      (require (for-syntax racket/base))
+      (let ()
+        (struct d (vec))
+
+        (define dx-vec (contract (-> (struct/c d (vectorof any/c)) any) d-vec 'pos 'neg))
+        (define-syntax dx (list #'struct:d #'d #'d? (list #'dx-vec) (list #f) #f))
+
+        (struct gds dx ())
+        (define gd (contract (struct/c gds (vectorof any/c)) (gds (vector 1)) 'pos 'neg))
+        (vector-ref (d-vec gd) 0)))
+   1)
+  
+  (test/spec-passed
+   'struct/c-simple-contract-accessor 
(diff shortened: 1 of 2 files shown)


rfindler:
I'm not sure how a pull request is useful here.

But if you want comments then let me suggest you add a test. Also, this
description isn't accurate.

Vectors are relevant only because they create chaperones.

Robby

samth:
I created a PR both because it gives us a concrete diff to talk about, and because it stores my progress, and it automatically runs all the tests.

rfindler:
Thanks for doing this, Sam. The fix seems right to me, but I would do the tests differently. Specifically, I think that test case must cover only one of the changes, and also I think a test case that involves only struct/c (or struct/dc) with a chaperone'd input is more directly focused on the issue at hand (that is, it isn't really about vectors). Including all three tests seems fine too, but if we ever have to come back here, it would be nice to have a test case that gets directly at the issue. 

I'm thinking of something very much like your test, but using something like this revision of some code you sent me earlier:

     (define val1
       (x (vector 1)))
     (define val2
       (chaperone-struct 
        val1 
        x-v 
        (λ (f v) 
          (chaperone-vector v
                            (λ (a b c) c)
                            (λ (a b c) c)))))


But also, feel free to push and I'll adjust the tests later. (Just in case, as I'm not clear how things work: thi is not going to be a merge commit, right?)

samth:
@rfindler I wrote the test you suggested, but it found a bug, I think in `opt/c`. Here's the error I see:

```
FINISHED struct-dc.rkt
FAILED struct/c-simple-contract-accessor rewrite-to-add-opt/c
       got "struct-ref: non-chaperone result;\n received a result that is not a chaperone of the original result\n  original: #<procedure>\n  received: #<procedure>"
  expected no-exn-raised
```

rfindler:
Probably this code needs to be adjusted in the same way:

                           #`(let ([answer (let ([#,sub-val 
                                                  (#,sel-id
                                                   #,(opt/info-val opt/info))])
                                             #,this-body-code)])
                               (let ([proc-name (λ (strct fld) answer)])
                                 proc-name))))))

Around 1355 in the same file.

rfindler:
I can do it later on, if you want. If you get to it first, feel free to push.

Thanks again.

PR #879, opened 2015-02-12
Add a contract combinator for prefab structs

takikawa:
The `prefab/c` in this PR matches up with the proposed `Prefab` type in TR and specifies a prefab key and a contract for each of the fields.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -591,6 +591,49 @@ inspect the entire tree.
 @history[#:changed "6.0.1.6" @elem{Added @racket[#:inv].}]
 }
 
+@defproc[(prefab/c [key prefab-key?] [field contract?] ...)
+         contract?]{
+  Produces a contract that recognizes instances of the @tech{prefab}
+  structure type corresponding to @racket[key]. See @racket[make-prefab-struct]
+  for a description of valid key shapes. The number of @racket[field]
+  contracts provided must be consistent with the provided @racket[key].
+
+  @examples[#:eval (contract-eval)
+    (prefab/c 'foo integer?)
+    (prefab/c '(foo bar 1) integer?)
+    (prefab/c '(foo 3) integer? string?)
+  ]
+
+  Each @racket[field] contract is applied to an instance's fields in
+  the same order as the order in the printed form of a prefab struct.
+  That is, the parent fields come before the child fields and auto-value
+  fields come after the ordinary fields.
+
+  The contract will detect if @racket[key] corresponds to a struct
+  type with mutable fields and check field contracts on mutation appropriately.
+  Impersonator contracts are only allowed on fields that are mutable.
+
+  @examples[#:eval (contract-eval)
+    (struct band (members) #:prefab)
+    (define/contract some-band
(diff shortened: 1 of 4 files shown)


rfindler:
This looks good to me. Feel free to push it.

The only things I'd add would be a test in context.rkt and support for random generation (the random generation should probably look like cons/c's random generation (the function cons/c-generate in misc.rkt)).

takikawa:
While trying to integrate this into TR, I hit a design issue that hadn't come up before.

Here's an example (that doesn't actually run right now) to illustrate the issue:

````
#lang racket

(module t typed/racket
  (struct parent ([x : String]) #:prefab)
  (struct child parent ([y : Integer]) #:mutable #:prefab)
  (: a-child Parent) ; subtyping
  (define a-parent (child "foo" 0))
  ;; exported with a contract for `parent` struct
  (provide a-parent))

(require (submod "." t))

;; the contract doesn't prevent this
(struct parent (x) #:prefab)
(struct child parent (y) #:mutable #:prefab)
(set-child-y! a-parent "not an integer")
````

This example illustrates how a `prefab/c` isn't sufficient protection when exporting a prefab with a supertype of its actual type, because that doesn't protect the subtype fields.

One solution might be to add `#:substruct-accessor` and `#:substruct-mutator` options to `prefab/c` that allow a contract to be specified for the extra accessors/mutators that might exist (with the default being to do nothing). This is a little weird in the case that a non-prefab struct extends a prefab struct (surprisingly this is allowed in Racket), but in that case it can still find all the prefab parts of the struct hierarchy via `struct-info`.

Another would be to define a custom contract just for TR that basically does what I described above but only for TR. It could re-use `prefab/c`'s projection but also apply another layer of chaperoning.

Do either of these approaches make sense?

rfindler:
I think it makes sense to add that as an option to prefab/c.

Robby

takikawa:
Sadly this seems difficult to implement because you can't figure out which parts of the actual struct hierarchy you get from the struct instance are substructs of the type corresponding to the prefab key you're given.

That's because a struct type obtained from the key won't necessarily be equal to the struct type you get from the actual struct instance. The name in the struct type info doesn't help because you can easily get different structs with the same name (you can even have them in the same hierarchy!).

It would help if struct type information could tell you if the type is a prefab type and, if so, its prefab key. I don't think there's any way to do that currently though.

(also there's a question of what to do with transparent non-prefab substructs of a prefab struct... should fields in those be checked or not?)

rfindler:
On Wed, Mar 11, 2015 at 3:13 PM, Asumu Takikawa <notifications@github.com>
wrote:

> Sadly this seems difficult to implement because you can't figure out which
> parts of the actual struct hierarchy you get from the struct instance are
> substructs of the type corresponding to the prefab key you're given.
>
> That's because a struct type obtained from the key won't necessarily be
> equal to the struct type you get from the actual struct instance. The name
> in the struct type info doesn't help because you can easily get different
> structs with the same name (you can even have them in the same hierarchy!).
>
> It would help if struct type information could tell you if the type is a
> prefab type and, if so, its prefab key. I don't think there's any way to do
> that currently though.
>

That sounds like a completely reasonable addition.


> (also there's a question of what to do with transparent non-prefab
> substructs of a prefab struct... should fields in those be checked or not?)
>
>
yeah. Not clear.

Robby

PR #881, opened 2015-02-13
Allow chaperones for methods in class expressions

takikawa:
I'd like to propose allowing `chaperone-procedure` around method definitions inside of a `class` expression in addition to the currently allowed method shapes. Since a chaperone only introduces extra errors (modulo side effects in the wrapper function and properties), I think this is a fairly conservative change.

My motivation is to allow TR to add chaperone properties to methods, so that contract combinators in TR can identify whether a given method is defined in typed code or untyped code.

The other commit introduces an alternative `object/c` constructor. I think I've submitted this for review before, so I'm just including it here because I intended to push both commits at once.

@mflatt @rfindler Any thoughts?


rfindler:
These both seem good to me.

mflatt:
Looks ok to me. As a minor suggestion, I think the docs should say `other-arg-expr ...` instead of `. other args`.

takikawa:
Thanks for the feedback. Merged.

PR #882, opened 2015-02-14
Fix Ignore fuel in optimize_for_inline

gus-massa:
In 7f61a685523b0481bd2fe161d72e3f804f75a18b I added a check to ignore the fuel in `optimize_for_inline` when it's used to get a known procedure instead of actually inlining. There are a lot of double and triple negations in the code and I messed up the Boolean algebra. 

I think this fix the problem, but I couldn't make test example.


mflatt:
Merged - thanks!

PR #889, opened 2015-02-22
Add a true? predicate to racket/bool

lexi-lambda:
So, `racket/bool` already has `false?`. This is an utterly useless function for everything besides clarity—it is identical in every way to `not`. Yet for whatever reason, it does not include `true?`, which would actually be *useful*.

This has been disputed a few times on IRC, so let me explain. Racket, like Scheme, uses the concept of truthiness. Most of the time, there's no need to cast a value to a boolean explicitly; instead of `boolean?`, the Racketeer uses `any/c`. Constructs like `if` handle these values appropriately.

I know this, but a `true?` function is still sometimes useful. Why? Well, sometimes you want to return a value from a function, but you don't want to return some truthy value that exposes some internal implementation details. There are various ways around this, the most common being `(not (not ...))` or `(and ... #t)`.

Don't believe me? Run `git grep -E "\\(and \\S+ #t\\)"`. This is a real pattern. It's also a bad one, and considering we have `false?`, there's no reason to not also include `true?`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/booleans.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/booleans.scrbl
@@ -265,6 +265,9 @@ Returns @racket[(equal? a b)] (if @racket[a] and @racket[b] are symbols).}
 
 Returns @racket[(equal? a b)] (if @racket[a] and @racket[b] are booleans).}
 
+@defproc[(true? [v any/c]) boolean?]{
+Returns @racket[#f] if @racket[v] is @racket[#f], @racket[#t] otherwise.}
+
 @defproc[(false? [v any/c]) boolean?]{
 
 Returns @racket[(not v)].}
--- a/pkgs/racket-test/tests/racket/bool.rkt
+++ b/pkgs/racket-test/tests/racket/bool.rkt
@@ -18,6 +18,10 @@
 (check-false (false? #t))
 (check-false (false? "11"))
 
+(check-false (true? #f))
+(check-true (true? #t))
+(check-true (true? "11"))
+
 (for ([x (in-list '(#f #t))])
   (for ([y (in-list '(#f #t))])
     (check-equal? (implies x y)
--- a/racket/collects/racket/bool.rkt
+++ b/racket/collects/racket/bool.rkt
@@ -1,6 +1,6 @@
 #lang racket/base
 
-(provide true false false?
+(provide true false true? false?
          boolean=?
          symbol=?
          implies nand nor xor)
@@ -9,6 +9,7 @@
 (define true #t)
 (define false #f)
 
+(define (true? v) (if v #t #f))
 (define (false? v) (eq? v #f))
 
 (define (boolean=? x y)


mflatt:
The story behind `false?` is that it's useful for following the HtDP design recipe in a literal way. That's why `false?` exists in the student languages, and we have included bindings from the student languages in `racket` so that the student programs are easy to port to Racket.

To explain a little more, the `false?` predicate is useful for following the design recipe when you have a data definition such as

```
 ; A grade is either
 ;  - #f
 ;  - a number
```

The template for a function that consumes a grade `g` has a conditional with `(false? g)` and `(number? g)` cases. It's true that `not` works just as well as `false?`, but for teaching purposes, it's more consistent to have a predicate that ends in `?`.

A `true?` function that recognizes anything other than `#f` isn't useful in the same way. It would even be confusing from that perspective.

Overall, I see your point, but I think I prefer the `(and ... #t)` idiom for coercing a value to a boolean, because it makes the coercion more explicit to me. Meanwhile, I don't think I ever use `false?`, but I like having it as a way to smooth the path from one direction into Racket.

lexi-lambda:
@mflatt That makes sense, but I still think `(and ... #t)` is quite unclear for people not familiar with the idiom. Would you prefer something like `->boolean`? That would make the cast more explicit while still providing the functionality.

lexi-lambda:
Another valid use case stemming from a discussion on #racket:

```racket
(andmap true? lst)
```

I'm actually *perfectly* find with `(andmap identity lst)`, but a few people on IRC disagreed with me, and either way, this would make things more explicit.

rfindler:
I'm pretty sure I've written (andmap values ...) before and I know I write
(filter values ...) with perhaps more regularity than one might wish for
(because the ellipses often begin (for/list....)").

LeifAndersen:
For whatever it's worth, I agree `true?` would be useful. For two (possibly negligible) reasons:

1. I mean, if we have `false?` one would expect `true?` to be there too. (It's news to me that it's not.)

2. In languages like nanopass (or at least the version of nanopass Andy designed), it expects a `?` to be on the end of a predicate. So if you want a terminal in the language that is only `#t`, then you (currently) need to make it yourself.

lexi-lambda:
Note that your point highlights a potential confusion with `true?`: you seem to expect that it would only return `#t` when given the value `#t`, but this implementation returns `#t` whenever the value is non-`#f`.

LeifAndersen:
Oh, so it's the same as `(not (not ...))`? Okay, so that may or may not be useful. I very much had in mind though something like `(equal? #t ...)`, which is at least what I think of as `true?`. But I could totally be wrong about that. :)

lexi-lambda:
I’m going to close this because:

  1. The desire for such a feature does not seem to be very high.
  2. Whether it should be `(eq? #t _)` or `(not (not _))` is not well-defined or agreed upon.
  3. Either behavior could be confusing.

I think a `->boolean` function might be useful, but Racket currently seems to avoid general coercion functions (all the existing builtins that start with `->` are contract combinators, with the exception of `->fl` and `->extfl`).

PR #891, opened 2015-02-27
Fix call to optimize_for_inline in procedure? reduction

gus-massa:
If I understand correctly, `optimize_for_inline` assumes that the local reference are in the old coordinates but in the `procedure?` reduction they have the new coordinates, so it's necessary to call `optimize_reverse`.

The first commit includes only the tests, because I'm sure they are correct and they identify the error. (One of them produces an error with the old code, and another produces an error with one of my draft implementations that was "almost" correct.)

The second commit is an attempt to fix this, but I'm not so sure about the details here. Also, this commit removes the additional argument of `optimize_for_inline` that I added in 0c5944d64a07e2cc2ebd8af36d0bd2057541b24c because it was used only here and it was a special case that didn't mix with the other code. Now the shifting is done with the reversing in `finish_optimize_application2`.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -2145,6 +2145,53 @@
               (define foo integer?)
               (display #t)))
 
+(test-comp '(lambda ()
+              (let ([is3 (lambda () 3)])
+                (letrec ([g (lambda () 3)]
+                         [is0 (lambda () 0)])
+                  g)))
+           '(lambda ()
+              (let ([is3 (lambda () 3)])
+                (letrec ([g (lambda () (is3))]
+                         [is0 (lambda () 0)])
+                  g))))
+
+(test-comp '(lambda (f)
+              (let ([x (f)])
+                (letrec ([g (lambda () (list x #t))]
+                         [dummy (lambda () 0)])
+                  g)))
+           '(lambda (f)
+              (let ([x (f)])
+                (letrec ([g (lambda () (list x (procedure? x)))]
+                         [dummy (lambda () 0)])
+                  g)))
+           #f)
+
+(test-comp '(lambda (f)
+              (let ([x (f)])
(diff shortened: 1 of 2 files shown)


mflatt:
I think the solution is to keep the call to optimize_for_inline() as before, but pass true for the `optimized_rator` argument, and also fix optimize_for_inline() to treat a local variable correctly --- that is, use optimize_reverse() when `optimized_rator` is true.

Currently, I don't see a way for the optimize_reverse() code to fire and still lead to an optimization. Simply passing true for `optimized_rator` avoids the bug, since that disables the use of a local variable. I'll likely leave the generalization in place, anyway, since it might make things a little clearer.

gus-massa:
I agree. I'm my test, I couldn't couldn't find an example that uses
the `optimize_reverse` path, but I was not sure.

I don't like the idea of using `optimized_rator = 1` because in that
case `optimize_for_inline` might go inside `begin`s and `let`s with
side effects, and the current reduction implementation is very simple
and just drops the whole expression. Nevertheless,
`finish_optimize_application2` already goes inside the `begin`s and
`let`s, so probably there is no counterexample.

I tried a few variations of

  (let ([f (lambda (x y z)
             (let ([v (display x)]) (lambda () v)))])
    (procedure? (f (random) 0 0)))

but the reduction was correct in all the cases.

(As I said a few days ago, I'm trying  to make a few changes in
`try_reduce_predicate` and nearby code, that I think will make this
more robust. While writing that I found this error, so that code is
not complete yet. I think that I don't have to add more shifting and
reversing in the new code, so I hope it will be correct in the first
try.)

mflatt:
Good point about the possibility of dropping side effects, but maybe that's better handled by make_discarding_sequence() instead of trying to limit the use of optimize_for_inline()

gus-massa:
This problem was fixed and merged by @mflatt in 04ce921e8fea57dee9df05ffdcb0ce6b08fb8ee6

PR #892, opened 2015-03-02
Reduce repeated object/c wrapping & modify opaque class/c

takikawa:
This PR combines two object/class contract changes. Both support upcoming Typed Racket changes.

One is adapting `object/c` to use the projection of `instanceof/c` in order to get the space-efficiency benefits of the latter (avoiding exponential wrapping). The motivation is to allow Typed Racket to switch to using `object/c` over `instanceof/c` in more cases.

The second changes opaque `class/c` contracts (i.e., ones where `#:opaque` is specified) to ignore local member names. I think the initial design choice may have been a mistake, since if the module specifying the opaque `class/c` doesn't have access to those local member names, it's impossible to specify a passing contract. This is especially a problem for Typed Racket---it doesn't really make sense to have to specify types for, say, internal GUI class member names which are never meant to be accessed.

@rfindler does this look reasonable?

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/class.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/class.scrbl
@@ -1785,7 +1785,8 @@ A class contract can be specified to be @emph{opaque} with the @racket[#:opaque]
 keyword. An opaque class contract will only accept a class that defines
 exactly the external methods and fields specified by the contract. A contract error
 is raised if the contracted class contains any methods or fields that are
-not specified.
+not specified. Methods or fields with local member names (i.e., defined with
+@racket[define-local-member-name]) are ignored for this check.
 
 The external contracts are as follows:
 
@@ -2009,7 +2010,10 @@ As with the external contracts, when a method or field name is specified
    such subclasses.}
 
 @history[#:changed "6.1.1.8"
-         "opaque @racket[class/c] now ignores local member names"]
+         @string-append{Opaque @racket[class/c] now ignores local member names.
+                        This is backwards incompatible, but only affects uses of
+                        opaque @racket[class/c] contracts applied to classes
+                        with methods or fields with local names.}]
 ]}
 
 @defform[(absent absent-spec ...)]{
--- a/pkgs/racket-doc/scribblings/reference/class.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/class.scrbl
@@ -2008,6 +2008,8 @@ As with the external contracts, when a method or field name is specified
    checked on any access and/or mutation of the field that occurs in
    such subclasses.}
 
+@history[#:changed "6.1.1.8"
+         "opaque @racket[class/c] now ignores local member names"]
 ]}
 
 @defform[(absent absent-spec ...)]{
--- a/pkgs/racket-doc/scribblings/reference/class.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/class.scrbl
@@ -1737,7 +1737,8 @@ resulting trait are the same as for @racket[trait-sum], otherwise the
 
 ([maybe-opaque
   (code:line)
-  (code:line #:opaque)]
+  (code:line #:opaque)
+  (code:line #:opaque #:ignore-local-member-names)]
 
  [member-spec
   method-spec
@@ -1786,7 +1787,8 @@ keyword. An opaque class contract will only accept a class that defines
 exactly the external methods and fields specified by the contract. A contract error
 is raised if the contracted class contains any methods or fields that are
 not specified. Methods or fields with local member names (i.e., defined with
-@racket[define-local-member-name]) are ignored for this check.
+@racket[define-local-member-name]) are ignored for this check if
+@racket[#:ignore-local-member-names] is provided.
 
 The external contracts are as follows:
 
@@ -2010,10 +2012,8 @@ As with the external contracts, when a method or field name is specified
    such subclasses.}
 
 @history[#:changed "6.1.1.8"
-         @string-append{Opaque @racket[class/c] now ignores local member names.
-                        This is backwards incompatible, but only affects uses of
-                        opaque @racket[class/c] contracts applied to classes
-                        with methods or fields with local names.}]
+         @string-append{Opaque class/c now optionally ignores local
+                        member names if an additional keyword is supplied.}]
 ]}
 
 @defform[(absent absent-spec ...)]{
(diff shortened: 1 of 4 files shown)


rfindler:
Can you say more what you mean by the second point? Maybe give a small
example so I can understand what's going on?

Also, this seems a little fishy:

https://github.com/plt/racket/pull/892/files#diff-63e7e495c3d1d91da7fe8b2e40e5eebbR1473

Why do we have two different contract combinators with the same exact
projection? Do they really check exactly the same things?

Robby

takikawa:
Sure, here's an example (from the tests) of the second point:

````
(let ()
  (define-local-member-name n)
  (contract (class/c #:opaque [m (-> any/c number? number?)])
            (class object% (super-new) (define/public (m x) 3) (define/public (n) 4))
            'pos
            'neg))
````

This test currently raises an error, because the `n` method is not given a contract. I'm proposing to make this succeed, because in general the local member name `n` could have been defined (and un-exported) in another module. In that case, it's not possible to write an opaque contract on that class outside of that module that does not fail.

---

On the exact same projection, I've modified the `instanceof/c` projection to check if the given contract is an `instanceof/c` or an `object/c` and have slightly different behavior. But they do have pretty similar behavior to begin with, since they both ultimately wrap the class associated with the object. The difference is that `instanceof/c` applies the class contract projection, while `object/c` would construct a more minimal wrapper. Also, the `object/c` case needs a bit more checking to ensure the fields/methods are there (which the class projection normally does).

rfindler:
Well, on the principle that #:opaque was added for TR and it isn't the
right thing to support TR then changing it seems fine.

But on the principle that #:opaque is out as it's own thing there and
you're changing its behavior, it doesn't seem so fine.

Are there any uses of #:opaque in any pkgs on pkgs.racket-lang.org?

Also, I think that if you do change it, a little more discussion should go
into the notes in the history.

-----------------

re same projection: I feel like that's a less than ideal code organization
but it is also something that can be changed another day, so I'm fine with
pushing it.

takikawa:
@rfindler I agree with the potential concern about backwards compatibility, but I think this is a very niche feature that hasn't seen much use so far. 

My main concern is actually that it would prevent Typed Racket from supporting local member names. I talked with Matthias about this, and we concluded that there isn't a good solution that keeps that door open. Any alternative requires Typed Racket to violate abstraction boundaries by listing the types of internal and potentially unexported methods or fields.

I looked on the package server at a few packages that seemed likely to use classes/class contracts. I did find some  uses of `class/c`, but none were opaque.

I also just pushed more documentation discussion in a separate commit.

rfindler:
On Tue, Mar 3, 2015 at 1:40 PM, Asumu Takikawa <notifications@github.com>
wrote:

> @rfindler <https://github.com/rfindler> I agree with the potential
> concern about backwards compatibility, but I think this is a very niche
> feature that hasn't seen much use so far.
>
I don't disagree, but I think that since it has been available for years,
we must check.

>  My main concern is actually that it would prevent Typed Racket from
> supporting local member names. I talked with Matthias about this, and we
> concluded that there isn't a good solution that keeps that door open. Any
> alternative requires Typed Racket to violate abstraction boundaries by
> listing the types of internal and potentially unexported methods or fields.
>

What is the "it" referring to here? Just to be clear, I have no problem
with a new, different keyword and TR using that one. Is that somehow
problematic?


> I looked on the package server at a few packages that seemed likely to use
> classes/class contracts. I did find some uses of class/c, but none were
> opaque.
>

Thanks. Do you know if there's a way to be more systematic about this? I'll
ask Jay if he has any ideas.


> I also just pushed more documentation discussion in a separate commit.
>
>
Thanks.

Robby

rfindler:
As Matthew points out, you can just `raco pkg catalog-archive` and
then unzip and grep. Does that turn anything up?

takikawa:
> What is the "it" referring to here? Just to be clear, I have no problem with a new, different keyword and TR using that one. Is that somehow problematic?

What I meant was that my proposed behavior for `#:opaque` prevents TR from being able to support local member names if that ever happens (I've tried and I'm not sure it's possible).

Also, using a new keyword is fine as far as Typed Racket is concerned. The only issue is that that creates two features that only differ in a very small way (and probably nobody but TR uses anyway).

> As Matthew points out, you can just `raco pkg catalog-archive` and then unzip and grep. Does that turn anything up?

Thanks, I'll try that.

rfindler:
I think if that doesn't turn up any uses of the keyword, then it is okay to
change. I'd be slightly in favor of changing the name of the keyword at the
same time as the behavior changes, if that's the way things end up going,
but I don't have a strong opinion here.

rfindler:
I feel like there should be a way to make local-member-names work with TR. Perhaps by putting type information into the define-local-member-name declaration instead of elsewhere?

takikawa:
Grepping through all the packages revealed that only the tests use the `#:opaque` keyword. I also looked at all the packages that use `class/c` manually just in case and that confirmed it. Ok to merge?

Re: local-member-name, I think it's tricky because the goal for TR is to be able to detect when a `define/public` would fail (due to a duplicate name in the superclass) at typechecking time. With local member names, though, whether it would fail depends on what names are in scope. (and similarly contract checking depends on what names are in scope)

rfindler:
On Tuesday, March 3, 2015, Asumu Takikawa <notifications@github.com> wrote:

> Grepping through all the packages revealed that only the tests use the
> #:opaque keyword. I also looked at all the packages that use class/c
> manually just in case and that confirmed it. Ok to merge?
>

What do you think about changing the name while we are at it? That way
people will learn quickly instead of slowly if something could go wrong? Or
if that is too drastic is there something else? Or do you think I worry too
much? :)



> Re: local-member-name, I think it's tricky because the goal for TR is to
> be able to detect when a define/public would fail (due to a duplicate
> name in the superclass) at typechecking time. With local member names,
> though, whether it would fail depends on what names are in scope. (and
> similarly contract checking depends on what names are in scope)
>

Dynamic check there seems fine!

Robby

takikawa:
> What do you think about changing the name while we are at it? That way people will learn quickly instead of slowly if something could go wrong? Or if that is too drastic is there something else? Or do you think I worry too much? :)

Maybe we could make the `#:opaque` keyword take an argument? Like `#:opaque 'ignore-local-member-name`? Or is it better to use multiple keywords? `#:opaque #:ignore-local-member-name`?

I think maybe you are right that we should keep compatibility, since maybe TR will want to revert to the old behavior if it gets support for local member names.

rfindler:
Something like that seems better I think.


Robby

takikawa:
Okay I pushed a commit that adds an extra keyword to customize `#:opaque` and documents it. Also have tests for both old and new behaviors.

rfindler:
Thanks!

takikawa:
Merged.

PR #893, opened 2015-03-07
More redutions of predicates

gus-massa:
The optimizer had some reductions of predicates applications, like `(pair? X)`, only when `X` was very simple and the type of `X` was obvious. Use `expr_implies_predicate` and `make_discarding_sequence` to allow the reduction of more complex expressions.

Also, the reduction of `procedure?` and `fixnum?` were special cases in `finish_optimize_application2`. Move the checks to `expr_implies_predicate`  to take advantage of the reductions in more general cases.

The seccond commit is adds a special case to reduce `list?` The predicate reductions use simple disjoint types, so `list?` needs to be handle specially since is a mix of `null?` and only some instances of `pair?`.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1237,6 +1237,14 @@
            '(lambda (w z) #t)
            #f)
 
+(test-comp '(lambda (w z) (list? (begin (random) null)))
+           '(lambda (w z) (random) #t))
+(test-comp '(lambda (w z) (list? (begin (random) (void))))
+           '(lambda (w z) (random) #f))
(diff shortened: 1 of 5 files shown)


mflatt:
I hope to look closely in a day or two. Meanwhile, beware that you need to register new global variables with the GC before initializing them. For example, register `scheme_fixnum_p_proc` with `REGISTER_SO(scheme_fixnum_p_proc)` .

gus-massa:
I added the `REGISTER_SO` and changed a `0` to `NULL`.

And, I just added a new commit/question. Most of the uses of `extflonum` are guarded by an `#ifdef`, but I think the variables are defined in all the cases. Is this necessary?

mflatt:
Merged, including a commit to remove the unneeded `MZ_LONG_DOUBLE`s.

PR #894, opened 2015-03-08
Add break-parameterization?.

shekari:
Closes PR 11966.


mflatt:
Merged (after adding a `history` annotation to the docs) --- thanks!

PR #895, opened 2015-03-08
Make define-cpointer-type declare its function with an inferred-name.

shekari:
Closes PR 12296.


mflatt:
Merged (after moving the test case to the less friendly ".rktl" format) --- thanks!

PR #896, opened 2015-03-09
Initial version of adding disappeared uses for provide forms.

shekari:
This is meant to fix one part of pr/13186, that the bindings used for the provide transformer are not caught by check-syntax.

I currently had to reimplement `with-disappeared-uses` because of circularity reasons. Should I split that out to its own file and reuse that machinery, or just re implement a simpler version? I'm thinking the former would be better.

### DIFF
--- a/racket/collects/racket/private/reqprov.rkt
+++ b/racket/collects/racket/private/reqprov.rkt
@@ -683,15 +683,16 @@
       [(module)
        (syntax-case stx ()
          [(_ out ...)
-          (with-syntax ([(out ...)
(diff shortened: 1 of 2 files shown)


mflatt:
I don't think `with-disappeared-uses` is the right idea here. The `pre-expand-export` function should attach `'disappeared-use` properties, `provide-trampoline` should propagate them and add more, and `provide` should propagate properties from expansions of `(expand ....)` clauses. I have that working, I think.

samth:
This can now be closed, right?

PR #897, opened 2015-03-09
Make for/fold/derived check the values in the body in the last clause.

shekari:
Closes PR 13285.


mflatt:
Merged (after moving the test case to the less friendly ".rktl" format) --- thanks!

PR #898, opened 2015-03-14
Improve reference to code value

projanen:
The documentation prose uses "???" to refer to specific Racket output displayed as "temp7".  Did "temp7" originally print out as "???"?

I would have understood this section more quickly if the reference to "temp7" had been made clearer, which is my attempt in this change.


stamourv:
Merged, thanks!

PR #904, opened 2015-03-18
Add docs for fsemaphores in the Synchronization section.

shekari:
Closes PR 13380.

It feels somewhat wierd to add this here and not have the docs under synchronization, but I originally filed the bug because I was looking up synchronization primitives and didn't find fsemaphores until much later, so at least wanted a link from this page.


mflatt:
Merged, but in case you wonder why I changed "works well", it's because I'm unconvinced that fsemaphores are a good idea overall.
Thanks!

PR #905, opened 2015-03-22
fix integer-length overflow (PR14986)

jkominek:
This fixes the overflow in integer-length when the argument is a bignum with ~2^31 bits in it. It does so by checking to see if the number is large enough it might cause an overflow, and if so, falls back to using bignum arithmetic. (It also uses a datatype that will be larger on 64 bit platforms, so they're completely in the clear for at least a few years.)

This also improves the performance on negative bignums, by avoiding the call to scheme_bignum_not and handling negatives directly.

It includes some test cases. In addition to the travis ci & appveyor tests, I also tested this on 32 bit linux, with and without USE_LONG_LONG_FOR_BIGDIG.

I'd be happy to make changes/fixes/adjustments to anything.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/number.rktl
+++ b/pkgs/racket-test-core/tests/racket/number.rktl
@@ -2633,6 +2633,39 @@
 (test #x100000000 arithmetic-shift #x200000000 -1)
 (test #x200000000 arithmetic-shift #x100000000 1)
 
+(test 0 integer-length -1)
+(test 0 integer-length 0)
+(test 1 integer-length 1)
+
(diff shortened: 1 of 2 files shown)


mflatt:
Merged, thanks!
I changed a `1L` to `(intptr_t)1` because I don't trust "L" to be long enough. For 64-bit Windows, `long` is the same as `int`, and I think `LL` might be needed to get to `intptr_t`; I'm not sure that's right, especially if the appveyor tests passed, but it seems safer to be explicit with `(intptr_t)`.

PR #906, opened 2015-03-23
add max year to sql-interval

jbclements:
I got an unpleasantly low-level error yesterday when I tried to issue a db query and had mistaken milliseconds for seconds in my conversion. This pull request signals an error on creation of a sql-interval whose year can't be serialized in 2 bytes. I'm about to make another pull request against db, for a tightened contract on creation of sql-date and sql-timestamp. I wrote a simple test case, but I wasn't quite sure where to put it, and I figured I'd just ask.


jbclements:
@rmculpepper have a moment to review this?

jbclements:
So as not to make you read the code again: I put - #x8000 up to #x7fff, a two-byte signed range.

PR #907, opened 2015-03-24
add docs on not running tests

jbclements:
It took me quite a while to figure out how to use submodules to prevent a file from being tested (the secret is module*). This PR adds docs that show how to do it. 

Please note that the 13 seems to be necessary to get the renderer not to put the close-paren at the end of the comment.


jbclements:
@mflatt  , have time to review this? It seems like a relatively uncontroversial change, I'll push it myself in 24 hours if I don't hear back.  Thanks!

mflatt:
I'd recommend `(module test racket/base)`, but ok otherwise.

jbclements:
@mflatt squash fail, wound up making two separate pushes, alas.

PR #909, opened 2015-03-27
Add first order check to parametric->/c

shekari:
I'm sure there is a better way to write the tests, but I couldn't figure it out.


rfindler:
There is a pile of first-order check tests in tests/racket/contract/first-order.rkt.

Also, what do you think about making up some `barrier/c`s and passing them in to get a more accurate first-order check?

shekari:
Add the `barrier/c`s and moved the tests. I decided to delay the `barrier/c` creation as much as possible, not sure if that is better or not.

rfindler:
I am wondering if perhaps there is a better way to do this. What if polymorphic->/c insisted that there was an arrow contract in its body, syntactically? So you could write

(let ([f (lambda (x) (-> x x))]) (polymorphic->/c (x) (f x)))

anymore, but you had to actually have the -> syntactically inside the polymorphic->/c. 

Then we could read the arity checks off of the syntax and not have to call that function and not have to deal with things like this:

(polymorphic->/c (x) (if (zero? (random 2)) (-> x x x) (-> x x)))

shekari:
The issue that I think you are trying to solve is that `polymorphic->/c` can possibly have two different first order checks on different applications. But that problem is present in `if/c` as well. Also the syntactic check seems wrong as it prevents using `->*` or `case->`, or it allows some well defined list of accepted forms but then people cannot expand those. In addition it breaks backwards compatability in ways that may break TR program's contract generation.

rfindler:
Yes, the problem is present in `if/c`. But I'm not sure that that should stop us from fixing this problem.

And the syntactic check I had in mind would include checking for all arrow combinators.

But is there another reason you think this is a bad idea? Too much work? (That seems reasonable, actually.)

shekari:
The issue is that it will break TR (and possibly other usecases of polymorphic->/c).

```
#lang typed/racket/base

(provide f)

(: f (-> (All (A) (Rec X (-> A X))) #f))
(define (f x) #f)
```

The body of the inner function turns into something like
```
(polymorphic->/c (a) (letrec ([x-typed ...] [x-untyped ...] [x-both ...]) x-untyped))
```

Which would not pass your proposed check.

rfindler:
Yes, it would have to have a different name than `polymorphic->/c` -- sorry I thought that, but didn't say it.

And I see that TR would have to change, but presumably that's in the "do more work" category.

rfindler:
Looking at the commit again, I think it should just be pushed. Do you want to do it, or shall I?

shekari:
I can push it.

PR #910, opened 2015-03-27
further notes on Git workflow

mbutterick:
Inspired by commit 7408ee4709c6bd861a42117cb93be39e8156c9f1, a few more clarifications about Git workflow with `raco pkg update --clone`, including an explanation of the special `extra-pkgs` directory.


mflatt:
Just moving the note about "extra-pkgs" to a margin note makes it all seem right to me – visible enough to people who can use the information, but enough out of the way for others. Assuming that's ok, I'll push that adjustment (along with some other adjustments for style and consistency; please check that I didn't mangle anything).
Overall, this section has become distressingly long, but maybe there's a lower bound to how simple it can be to mix two complex systems.

mflatt:
Merged (a while ago)

PR #912, opened 2015-04-02
Implement class sealing/unsealing

takikawa:
This implements two class operations `class-seal` and `class-unseal` that are not tied to contracts, but are primarily intended to be used to implement polymorphic contracts on classes.

@rfindler I showed you a previous incarnation of this code, but I've basically re-written it from scratch since then. More specifically, I took your advice and separated the `make-seal/c` operation into two more primitive operations that aren't tied to contracts.

I also use the new feature of impersonator properties that lets a property be overriden by a chaperone/impersonator that does no redirection and just supplies a witness instead. This lets sealing/unsealing use memory proportional to the number of seals and the number of classes in the hierarchy, rather than proportional to the number of times something gets sealed and/or unsealed (even with the same seal). It also avoids allocating a whole class structure for each seal/unseal operation.

The client of this PR is this Typed Racket PR: https://github.com/racket/typed-racket/pull/80

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/class.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/class.scrbl
@@ -2701,6 +2701,47 @@ a method that is not supplied by an object.
 
 }
 
+@defproc[(class-seal [class class?]
+                     [key symbol?]
+                     [unsealed-inits (listof symbol?)]
+                     [unsealed-fields (listof symbol?)]
+                     [unsealed-methods (listof symbol?)]
+                     [inst-proc (-> class? any)]
+                     [member-proc (-> class? (listof symbol?) any)])
+         class?]{
+
+Adds a seal to a given class keyed with the symbol @racket[key]. The
+given @racket[unsealed-inits], @racket[unsealed-fields], and
+@racket[unsealed-methods] list corresponding class members that are
+unaffected by sealing.
+
+When a class has any seals, the @racket[inst-proc] procedure is called
+on instantiation (normally, this is used to raise an error on
+instantiation) and the @racket[member-proc] function is called
+(again, this is normally used to raise an error) when a subclass
+attempts to add class members that are not listed in the unsealed lists.
+
+The @racket[inst-proc] is called with the class value on which an
+instantiation was attempted. The @racket[member-proc] is called with
+the class value and the list of initialization argument, field, or
+method names.
+}
+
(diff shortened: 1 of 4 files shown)


mflatt:
@takikawa Ok by me. Nit: "which" should be "that" in the first paragraph of the docs.

rfindler:
Sorry for the long delay in getting back to this. 

The code looks okay to me.

Can you give a little more hint on how you expect TR to use this? In particular, I guess it will seal things when they cross the boundary into the untyped code. How does the unsealing work? That is, under which conditions will a class be unsealed (at a high-level)?

takikawa:
@rfindler Yes it'll get sealed going into untyped code. Unsealing happens whenever it flows back into the typed side, such as being returned from a mixin. So it'll be used like the barriers in `parametric->/c`.

rfindler:
Okay, but how do you know it is safe to unseal?

takikawa:
By safe you mean the thing is actually sealed? So you don't know this in general, which is why the unsealing errors (with blame if it's used for a contract) if the thing isn't sealed. Detecting this is actually important since (just like with ordinary polymorphism for functions) you don't want to let a function return an result that is "unrelated" to the input (i.e., not sealed) if it's supposed to return a parametric thing.

rfindler:
I somehow thought that subclassing was allowed on the untyped side, so
there must be something more going on there? I guess you put contracts on
the new methods when the class comes back across the boundary?

Does this mean there is now a linked representation for methods somehow in
a class?

takikawa:
Yeah subclassing is fine because seals are inherited. If it subclasses with methods that are not supposed to be there it will error though. The method representation doesn't change at all. It's just that there's a new chaperone property added onto the class (that is copied to the child in inheritance).

rfindler:
Okay, well I'm missing something, but I am pretty sure its just me being
dense.

Meanwhile, small nit, don't do this:

   (for ([f all-init-checkers]) ...)

put an `in-something` on there.

takikawa:
@rfindler I may also be explaining this poorly. It really does work mostly like `parametric->/c` does, though, but just where the positive and negative occurrences are replacing by either sealing or unsealing on classes.

In particular here is where TR defines the sealing function contract: https://github.com/takikawa/typed-racket/blob/sealing-ctc/typed-racket-lib/typed-racket/utils/sealing-contract.rkt

PR #914, opened 2015-04-05
Clarify the minimum CPU requirements in README

juanfra684:


mflatt:
It sounds like "scheme.h" should look for `__SSE2_MATH__`, instead of `__SSE_MATH__`, as an indication that SSE2 is available. Right?

juanfra684:
Yes, thanks for the update. I think that you should also change "-mfpmath=sse" to "-msse2 -mfpmath=sse" in the README.

PR #915, opened 2015-04-10
Make docs for sgn correct.

shekari:
The docs for `sgn` are currently incorrect as they leave out negative zero, nan, and single flonums. This adds all of those to the doc. But I think it might be better to find a way to describe all the values other than by listing them all out.


mflatt:
merged, but using `(or/c (=/c -1) ...)`

PR #916, opened 2015-04-11
Mark the result of more procedures as vector?

gus-massa:
The result of some procedures is a `vector`, but they are not omittable because they may rise an error. With the recent changes of the predicate reduction these cases are correctly handled.

This has a few commits and many questions

The first commit marks the result of `make-vector` and `list->vector` as a `vector?`. A added  the global variables `scheme_make_vector_proc` and `scheme_list_to_vector_proc`. Is it a problem to add too many globals?

The second commit adds `struct->vector`. The result of this function depends on the `current-inspector` and I think that that is the cause of the unnecessary `values` that the optimizer adds arround `struct->vector`. I tried to fix it, but I left it there because I didn't want to break something, nor add another special case.

The third commit remove some of the arity checks in  `rator_implies_predicate` because they are unnecessary and I'm too lazy to add the arity check to the new procedures :).

The fourth commit makes `(make-vector <number>)` omittable. In many use cases the length of the vector is fixed at compile time, for example (make-vector 3) is equivalent to (vector 0 0 0). So we are sure that make-vector will not raise an error and we can recognize these expressions as omittable and drop them when the result is ignored. The problem is that `scheme_omittable_expr` is used for `module_optimize` and when `resolved=0` this cause a error in the validation step. So I added a check for `resolved!=0`, but I'm not sure this is correct.

Also, `scheme_omittable_expr` is very similar to an ideal extended/enhanced version of `optimize_ignored`. I think they should be merged, but they have too many options, so it looks like a difficult project and I will not try it for now.


mflatt:
The issue with `struct->vector` is that it's not marked as "noncm". I think it would be safe to use `scheme_make_noncm_prim` for that one, instead of `scheme_make_prim_w_arity`, and that would fix `values`.

I worry about dropping the arity checks in `rator_implies_predicate`. Does that allow `finish_optimize_any_application` to discard a call that would raise an exception?

So far, I don't see why `resolved` would be needed for `scheme_omittable_expr`. I'll have to try it out.

mflatt:
I see that `finish_optimize_any_application` is ok because it defers to 

Dropping the `resolved` requirement for `make-vector` as omitable breaks validation because the validator can't reconsrtuct the optimizer's reasoning that `(make-vector <num> <const>)` will never fail. The `scheme_is_functional_nonfailing_primitive` normally reconstructs the optimizer's conclusion, but that function cannot handle the argument-sensitivity of the `make-vector` case. I think the solution is to extend `validate_expr` in the three places that it calls `is_functional_nonfailing_rator`.

gus-massa:
I read `finish_optimize_any_application` more carefully and I added a check, so in a Boolean context an `<error>` is not wrapped as `(begin <error> #t)`.

There are still a few corner cases like the `make-vector` in

    (if (if z 
            (make-vector (f z))
            (void))
        'a
        'b)

but they are more difficult to detect and I hope they are very unusual.

gus-massa:
I change the definition of `struct->vector` to use `scheme_make_noncm_prim`. (Do you have an easy example of a primitive that really needs `scheme_make_prim_w_arity`?)

I also changed `validate_expr` to handle `(make-vector <number> x)`.

mflatt:
These changes are fine as far as I can tell.

The `hash-ref` function is a an example of one that is not non-cm, because it accepts a function argument that it calls in tail position. The `values` or `struct-info` functions are not non-cm, because they can multiple values.

PR #917, opened 2015-04-15
Check the type of the arguments of more procedures

gus-massa:
The optimizer checks the type of the argument of some unary procedures and uses the gathered information to replace them by the unsafe version, reduce predicates and detect type errors. 

This extends the checks to more procedures that have no unsafe version and procedures that have more than one argument.

--

I also thought about checking the primitives compearing the pointers instead of the names, for example change

    check_known(info, app, rator, rand, offset, "car", s__pair_p_proc, s__unsafe_car_proc);

to

    check_known(info, app, rator, rand, offset, s__car_proc, s__pair_p_proc, s__unsafe_car_proc);

But I didn't implement it because I was not sure if this is useful and most of the primitives don't have a global variable.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -20,6 +20,7 @@
   (namespace-require 'racket/extflonum)
   (namespace-require 'racket/fixnum)
   (namespace-require 'racket/unsafe/undefined)
+  #;(namespace-require '(rename '#%kernel k:map map))
   (eval '(define-values (prop:thing thing? thing-ref) 
            (make-struct-type-property 'thing)))
   (eval '(struct rock (x) #:property prop:thing 'yes))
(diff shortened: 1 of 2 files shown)


mflatt:
Nice!

I agree that it's fine to leave the checks as strings in this case, because it gets tedious to make a global variable for every function. Maybe there's a better way overall.

I worry a little about introducing `app_o` as an alias of `app` to avoid a cast. I've gotten into trouble with that pattern by later adding an assignment of the original (i.e., `app`) without noticing that I need to also change the alias (i.e., `app_o`).

gus-massa:
What about `rator`? Should I change it back to `app->rator`?

gus-massa:
I just made `app_o` local to reduce the problems of overwriting `app`. But I didn't change `rator=app->rator`.

There are a few commented primitives checks: `vector-copy!`, `vector->immutable-vector` and `vector-fill!`. It's possible to check them, but they are used very seldom. Should I remove them or uncomment them?

mflatt:
Ok. I wouldn't bother with `vector-copy!`, etc., but I have no strong opinion.

PR #918, opened 2015-04-16
Tests for 16ce8fd90d

gus-massa:
Due to a bad coordinates use, the optimizer confused the variables `f-two` with `unuser-pair` of the test, and in a boolean context reduces it to `#t`. See 16ce8fd90d and PR15038.

The right number of let's is dificult to calculate, so we generate many variations. Currently, to get the error `n` should be `16`.


mflatt:
Just to confirm, this is a great addition.

PR #921, opened 2015-04-23
at-exp: make language-info compose with other languages

AlexKnauth:
If I use for instance #lang at-exp rackjure, then rackjure #λ literals will still work in the repl, though before this they didn't.  Also if the base language provides other language-info, it will still use that, instead of throwing it away and replacing it completely.


mflatt:
Merged - thanks!

PR #924, opened 2015-04-28
Documentation: Note that exn:break-continuation doesn't work with with-handlers

greghendershott:


greghendershott:
I changed this to caution only about `with-handlers`, in response to @mflatt 's comment.

I amended the commit and force-pushed, because I figured that would effectively be the end result of rebasing and squashing. But that "sliced off" the commentary attached to the original commit, from this pull request. For reference, it's here: https://github.com/greghendershott/racket/commit/5c74d581f8952fa80c9fddabb11422b86951af6e#commitcomment-10952304

mflatt:
Merged - thanks!

Issue #919, opened 2015-04-19
Inheritance of parent structs is not clear on fields in documentation


bluerider:
It is not noted in the documentation that inheriting fields from a parent struct requires calling them by the parent struct's name. Example:
(struct parent (test-field))
(struct child parent (dummy-field))

(child-test-field child) is unknown
(parent-test-field child) is known

zenspider:
http://docs.racket-lang.org/guide/define-struct.html#%28part._struct-subtypes%29

makes it pretty clear, but that example, or equivalent, should probably be added to the reference for completeness.

bluerider:
I agree now that I look at it more closely. However, I think it should be mentioned explicitly that class methods are inherited verbatim.

zenspider:
> On Apr 21, 2015, at 23:57, bluerider <notifications@github.com> wrote:
> 
> I agree now that I look at it more closely. However, I think it should be mentioned explicitly that class methods are inherited verbatim.

agreed

bluerider:
Submitted Pull Request #925

bluerider:
Fixed in c44cffe5a88c1221697cffe23dacaec42e976c4d

PR #925, opened 2015-04-29
Fix up Issue #919

bluerider:
Updated guide to clarify subtype struct inheritance of fields. A subtype inherits supertype's prefixes for fields.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/define-struct.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/define-struct.scrbl
@@ -136,6 +136,10 @@ supertype.
 (define p (3d-posn 1 2 3))
 p
 (posn? p)
+(3d-posn-z p)
+(code:comment "a 3d-posn has an x field, but there is no 3d-posn-x-selector:")
+(3d-posn-x p)
+(code:comment "use the supertype's posn-x selector to access the x-field")
 (posn-x p)
 (3d-posn-z p)
 ]


samth:
This seems to have an extra merge commit, and also to add a bunch of .css files and others.

bluerider:
samth: Not sure were those .css files are. Let me check. The extra merge commit is when I updated my fork against PLT/Racket master.

bluerider:
Must've occured when I ran raco scribble to check if the HTML doc was fine. I'll remove them

bluerider:
Not sure how to solve the merge commit though. Any help there?

samth:
You should rebase your branch that you're working on, and then do a force push to `bluerider:master`.

bluerider:
Okay, I rebased. Is it fine now?

samth:
Yes, the commits are fixed, thanks! That makes it much easier to review just the change you've made.

mflatt:
I think the idea here is good, but it doesn't currently render well. The long comment should be broken into two lines. I'd drop the last line (intended to be a comment?) 

Nit: I think a comma should appear before "since" in the first comment

bluerider:
All right. I've rebased and reworded the comment.

mflatt:
Thanks for fixing the rendering problems! The second line is still long on my screen, but after further thought, I'd like to suggest text changes, anyway.

How about

   a 3d-posn has an x field, but there is no 3d-posn-x selector:

and

 use the supertype's posn-x selector to access the x field:

?

(I'm happy to adjust the patch after we arrive at text that we all like.)

bluerider:
All right; I've updated the patch to use your suggestions verbatim.

mflatt:
Merged - many thanks!

Issue #866, opened 2015-01-26
The documentation for `count` from `racket/list` is inaccurate


lexi-lambda:
The documentation for [`count`](http://docs.racket-lang.org/reference/pairs.html#%28def._%28%28lib._racket%2Flist..rkt%29._count%29%29) states the following:

> Returns `(length (filter proc lst ...))`, but without building the intermediate list.

This isn't quite accurate, because unlike `map`, `filter` does not accept multiple list arguments. Instead, `count` is much closer to something like this:

```
(foldl (lambda (lst ... result) (if (proc lst ...) (add1 result) result))
       0 lst ...)
```

bluerider:
Does count accept multiple lists?

bluerider:
Well, I've just been shown on the IRC that it can accept multiple lists. filter should be replaced with filter-map then.

bluerider:
Fixed up to use filter-map instead of filter:
https://github.com/plt/racket/pull/927

samth:
Fixed.

PR #927, opened 2015-05-03
Fix up Issue #866

bluerider:
Fix up documentation issue #866. Procdure of count is more similar to filter-map than filter.


samth:
Merged.

PR #928, opened 2015-05-03
Add expansions for and/or to permit higher-order usage

lexi-lambda:
This is a work in progress experiment I've implemented based on a discussion I started on the IRC channel. Effectively, this modifies how `and` and `or` expand to permit them to be used as functions.

```racket
(apply and 1 #f 3) ; => #f
```

Obviously, using `and` or `or` as functions disables their short-circuting functionality, but using them as before retains it. Importantly, all usages of these forms in a higher-order manner would have previously been a syntax error, so no working code can possibly be affected by this change.

I *personally* believe this is a worthwhile change, but obviously these are two fairly important forms, so I think a discussion of the pros and cons is needed. I think think the benefits are straightforward:

- Using these operations in a higher-order manner can frequently be useful, and it is not uncommon to have new users unfamiliar with how the macros operate ask questions on IRC after running into a syntax error from a bare use of `and` or `or`.
- Implementing these as alternative expansions of the existing macros makes using them seamless, and follows the principle of least astonishment.
- Any higher-order use of these logical functions wouldn't be able to take advantage of short-circuiting evaluation, anyway (assuming a strict evaluation model), so it is unlikely to cause any surprise in that regard.

The possible downsides I've considered are as follows:

- The "magical" behavior could be even more confusing to new users if they were to accidentally stumble into it. Trying to use `and` as a higher-order function currently just throws a syntax error, which is annoying, but at least the user is prompted to figure out why. On the other hand, if the user expects short-circuit evaluation with this implementation, it will silently "misbehave".
- Introducing this kind of change in `racket/base` could likely have effects on other `#lang`s that make use of the base language's `and` and `or` forms. For example, `r5rs` just uses `and` and `or` from `scheme/base`, so it would need to be modified to use the old behavior.

There are also topics of consideration that aren't necessarily benefits *or* downsides.

- Should these be a part of `racket/base`, `racket/bool`, or some other package? Having them in the standard library would be very nice, but maybe this sort of change isn't suitable for the core language.
- If these did end up in `racket/base`, would there be any easy way to access the legacy behavior? Since this is a superset of the old behavior, it is unlikely that it would really be necessary, but for languages like `r5rs`, it might be useful.

Finally, if moving forward is determined to be a good idea, there are things I'd need to do to get this to an acceptable state.

- [ ] Update reference and guide to adequately explain and document behavior
- [ ] Add tests!
- [ ] Fix the rnrs langs, maybe by changing their behavior in `scheme/base`?

### DIFF
--- a/racket/collects/racket/private/applicable-and-or.rkt
+++ b/racket/collects/racket/private/applicable-and-or.rkt
@@ -1,0 +1,37 @@
+;; -----------------------------------------------------------------------------
+;; applicable expansions for and, or
+
+(module applicable-and-or '#%kernel
+  (#%require (prefix s: "small-scheme.rkt")
+             "define.rkt" "misc.rkt" "for.rkt"
+             (for-syntax "stxcase-scheme.rkt"))
+  
+  (#%provide and or)
+  
+  (define-syntax-rule (define/renamed rename (name . args) body ...)
+    (define name
+      (procedure-rename
+       (λ args body ...)
+       'rename)))
+
+  (define/renamed and (applicable-and . args)
+    (for/fold ([v #t])
+              ([arg (in-list args)])
+      #:break (not v)
+      (s:and v arg)))
+  
+  (define/renamed or (applicable-or . args)
+    (for/fold ([v #f])
+              ([arg (in-list args)])
+      #:break v
+      (s:or v arg)))
+
+  (define-syntax and
+    (syntax-id-rules ()
+      [(_ . args) (s:and . args)]
+      [_ applicable-and]))
+  
+  (define-syntax or
+    (syntax-id-rules ()
+      [(_ . args) (s:or . args)]
+      [_ applicable-or])))
--- a/racket/collects/racket/private/base.rkt
+++ b/racket/collects/racket/private/base.rkt
@@ -12,6 +12,7 @@
              "cert.rkt"
              "submodule.rkt"
              "generic-interfaces.rkt"
+             "applicable-and-or.rkt" ; shadows `and', `or'
              (for-syntax "stxcase-scheme.rkt"))
 
   (#%provide (all-from-except "pre-base.rkt"
@@ -38,6 +39,7 @@
              (all-from "cert.rkt")
              (all-from "submodule.rkt")
              (all-from "generic-interfaces.rkt")
+             (all-from "applicable-and-or.rkt")
              (for-syntax syntax-rules syntax-id-rules ... _)
              (rename -open-input-file open-input-file)
              (rename -open-output-file open-output-file)


mflatt:
I mostly worry about the "magical" downside. A related downside is that `(apply and (list e1 e2))` would not be the same as `(and e1 e2)` for all `e2`, which seems awkward.

As extra fodder for discussion: Would it make sense to have non-short-circuiting variants of `and` and `or` using different names? For the sake of discussion, we might call them `and*` and `or*`. That way, there's no issue with different behavior in first-order and higher-order applications. We might even adjust the `and` and `or` macros to suggest `and*` or `or*` when used in a non-application position (and when those identifiers are bound at the use site?).

rfindler:
Oh, I like the "different names with a note in the error message approach"! That's a great idea.

samth:
If it's not bound at the use site, can we make the error message describe how to get it?

lexi-lambda:
I personally like the magic, since it doesn't have any downsides for those aware of it, but I also recognize the potential confusion, so `and*` and `or*` would still be nice to have.

rfindler:
Well, we strive to make Racket work well for people who don't already know
the magic (as much as we can).

lexi-lambda:
From a certain perspective, I agree. It would be absolutely possible to add magic in places where magic does not belong, to the point of making the language incomprehensible to newcomers. That said, it is my opinion (and nothing more than opinion) that the advantage of the magic in this case is a big enough win to warrant it, especially since I'd find it unlikely that a user would stumble upon it accidentally in 99% of use-cases.

I suppose one could make an argument for a "limited" version of the language with less magic that would be more accessible to beginners. Of course, we actually already have that: `htdp/*sl`. BSL, for example, turns off the `cons` → `list` magic. Racket, on the other hand, provides a number of conveniences for the sake of the experienced programmer.

But again, that's my opinion, and I completely understand the argument that there's a rather big difference between `cons` cells printing differently and `and`/`or` behaving in different ways in different usages. The general consensus seems to be that adding separate strict versions would be preferable, and I'd be more than happy to go that route instead.

rfindler:
Thanks, Alexis, for your thoughts and your flexibility.

samth:
My feeling is that the magic isn't so bad here, and in keeping with other areas of Racket where similar magic happens: `in-range` in `for` loops, keyword argument function warnings.

I think the change @lexi-lambda proposes would continue to work well (better, really) for people who don't know the magic.

stamourv:
(add1 @samth)

rmculpepper:
On 06/22/2015 12:25 PM, Sam Tobin-Hochstadt wrote:
> My feeling is that the magic isn't so bad here, and in keeping with
> other areas of Racket where similar magic happens: |in-range| in |for|
> loops, keyword argument function warnings.

I disagree. It's a bad idea to have an expression that changes its 
meaning depending on whether it's used in operator position: eg,

   (or 1 (/ 0)) => 1
   ((values or) 1 (/ 0)) => error
   (let ([f or]) (f 1 (/ 0))) => error

That's not the same as in-range; for detects direct uses of in-range 
(and others) to expand into a different (more efficient) implementation 
of the same behavior.

It's also different from constructs like method names within a class 
body. Method names are not expressions; they can only be used in 
operator position. (But allowing method names used in non-operator 
position to expand into the appropriate lambda expression would be okay, 
because it would be consistent with their behavior in operator position.)

Ryan

rfindler:
I'm with Ryan on this point overall, but I strongly agree that the
situations Sam suggests are not well characterized by the phase "similar
magic".

Robby

greghendershott:
I like the motivation behind this PR!  Probably everyone agrees that the status quo -- a terse "bad syntax" error message -- is not helpful.

More-helpful alternatives seem to include:

1. The "magic" behavior in the PR.

2. A helpful error message suggesting using new `and*` `or*` functions, as @mflatt suggested.

3. A helpful error message suggesting using what's already available. For example in `qq-and-or.rkt` change

        (raise-syntax-error #f "bad syntax" x)

  to something like

        (raise-syntax-error #f "is syntax not a function; try `ormap' or `for/or`" x)

I wonder if 2 and/or 3 might be the most helpful, because it's teaching (or reminding) people something they do need to understand sooner rather than later?

[What to understand: It's not apparent whether `(f x)` is a function application or a macro invocation. Macros can do things like short-circuit. And so on. I think it's pretty common to "know" this from reading, but still need a few reminders in practice before it sinks in. So instead of eliminating this learning opportunity, maybe provide a clearer reminder and more enjoyable learning?]

p.s. I guess 2 and 3 are compatible; the error message could mention all of `ormap`, `for/or`, and a new `or*`.

samth:
Just a note that a new Racketeer was confused by this issue today.

PR #929, opened 2015-05-06
use better url-regexp in string->url contract

stchang:
Any reason not to use this better url regexp?

### DIFF
--- a/racket/collects/net/url.rkt
+++ b/racket/collects/net/url.rkt
@@ -704,10 +704,7 @@
 (provide (struct-out url) (struct-out path/param))
 
 (provide/contract
- (string->url (-> (and/c string?
-                         (or/c #rx"^[a-zA-Z][a-zA-Z0-9+.-]*:"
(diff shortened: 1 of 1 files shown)


jeapostrophe:
Looks good to me!

lexi-lambda:
The documentation for `string->url` should probably be updated as well to include the better regex in the contract.

stchang:
Can I provide `url-regexp`? (I won't merge until #948 gets merged.)

jeapostrophe:
I think it would be a good idea

stamourv:
#948 is now merged.

mflatt:
For the record, one reason this probably wasn't a good idea is that the regexp matches all strings.

At, least, I think it does: it can match just a path as any character except `?` and `#`, but that can be followed by a query as `?` and then any sequence of characters except `#`, and that can be followed by a fragment as `#` and any sequence of characters (with no exceptions).

PR #930, opened 2015-05-07
Add shorthand syntax-parameterize-rename form

jackfirth:
Would find this very useful as it's what I almost always use syntax parameters for. I can't figure out what's going on in the file that tests syntax parameters, so I'm unsure how to add any tests to this.


mflatt:
I can see how this would be useful. The shorthand doesn't quite feel worthwhile to me, but I have no technical objection, and I'm willing to be overridden or outvoted. Do others find this useful enough to merge?

jeapostrophe:
I think it's too specific to be included, but I'm a bit of an RTFM
curmudgeon.

Jay

On Sat, Jun 20, 2015 at 9:57 AM, Matthew Flatt <notifications@github.com>
wrote:

> I can see how this would be useful. The shorthand doesn't quite feel
> worthwhile to me, but I have no technical objection, and I'm willing to be
> overridden or outvoted. Do others find this useful enough to merge?
>
> —
> <https://github.com/plt/racket/pull/930#issuecomment-113770627>.
>



-- 
Jay McCarthy
http://jeapostrophe.github.io

                          - D&C 64:33

rfindler:
Would it help the guide exposition to use something like this?

Robby

samth:
Is there a reason not to make this a new package?

jackfirth:
@rfindler Greatly, I think. Making the guide show how to use syntax parameters for the common use case and then showing the more general syntax-parameterize equivalent form would help readability

@samth No technical reason. I've personally found this to be the main use case for syntax parameters, and making the syntax system simple to use in the common case would be valuable for those learning it. But that is a personal position, and I am fine putting this in a package.

rfindler:
On Sat, Jun 20, 2015 at 6:03 PM, Jack Firth <notifications@github.com>
wrote:

> @rfindler <https://github.com/rfindler> Greatly, I think. Making the
> guide show how to use syntax parameters for the common use case and then
> showing the more general syntax-parameterize equivalent form would help
> readability
>
> <https://github.com/samth>
>

Dare I challenge you to put your money where your mouth is with an expanded
pull request to show us just how awesome it would be?

:)

jackfirth:
You're on :)

PR #931, opened 2015-05-08
Add a char-in flat contract

lexi-lambda:
This adds a `char-in` flat contract to `racket/contract`, which is something I've found myself wanting before. I think it's a fairly straightforward change. Includes tests and documentation.


rfindler:
Thanks! I pushed this commit.

lexi-lambda:
Great, thanks for fleshing it out more, too.

PR #932, opened 2015-05-10
json: Remove regexp-try-match used to parse strings, replace with less-memory intensive functions.

e45lee:
I've encountered a few issues where the JSON parser uses tons of memory to parse simple strings.  (when trying to parse a 38MB string, the JSON parser uses at least 1GB memory, according to DrRacket).

I've tracked it down to the usage of regexp-try-match to match the string to the next escape or quote in read-string.

This PR replaces regexp-try-match with a simple loop over the string instead.  It passes all the tests in racket-test/json, and it doesn't crash on large strings either.

### DIFF
--- a/pkgs/racket-test/tests/json/json.rkt
+++ b/pkgs/racket-test/tests/json/json.rkt
@@ -119,6 +119,9 @@
         (string->jsexpr "") => eof
         (string->jsexpr " \t\r\n") => eof
 
+        ;; Invalid UTF-8 input.
+        (bytes->jsexpr #"\"\377\377\377\"") =error> exn:fail:contract?
+
         ;; More string escapes:
         (string->jsexpr @T{ "hel\"lo" }) => "hel\"lo"
         (string->jsexpr @T{ "\\//\\\\//" }) => "\\//\\\\//"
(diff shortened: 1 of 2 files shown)


mflatt:
This looks like a good improvement, but I wonder about encoding issues. The comment on the old version suggests that a string regexp was used so that it would only match a valid UTF-8 encoding. Does the new version need to defend against a bad encoding? Or is the implicit fixup by `get-output-string` (replacing bad encodings with "?") appropriate?

e45lee:
Currently, what happens in the old read-string is that it throws "unterminated string" when given non-UTF-8 text.  (The new read-string just replaces with "?")

<pre>
-> (require json)
-> (bytes 255 255 255)
#"\377\377\377"
-> (bytes->jsexpr #"\"\377\377\377\"")
; string::2: bytes->jsexpr: unterminated string [,bt for context]
</pre>

I'm not sure if either behaviour is ideal.  Alternatively, I could use `(compose bytes->string/utf-8 get-output-bytes)`, and have it toss a more descriptive exception when given invalid UTF-8, if that's better.

plt:
What do other JSON parsers do on this input?

e45lee:
Python's JSON parser tosses an UnicodeDecodeError.  I don't have anything else handy to test with right now.  I'll update the PR to do the equivalent thing -- use `(compose bytes->string/utf-8 get-output-bytes)`, and have it toss a more descriptive exception when given invalid UTF-8.

mflatt:
Merged - thanks!

PR #934, opened 2015-05-16
contract docs: explain positive and negative as server and client

AlexKnauth:


rfindler:
Thanks! I've added this commit.

PR #935, opened 2015-05-17
More types propagation

gus-massa:
The fist commit enables the propagation of types from the body of a `let` form to the surrounding code, for example in

    (lambda (z) (let ([o (random)]) (car z)) (pair? z))
    (lambda (z) (let ([o z]) (car o)) (pair? z))

reduce `(pair? z) ==> #t`.

The second commit extends this propagation to inlined functions.

I'm still worried because I'm using a very simple method to shift the coordinates in both cases.

The third commit calculates the intersection of the types of the two `if` branches. I tried to minimize the creation of new `trees` to hold the type information. I assume that most of the time one of the branches doesn't add new type information.

The last commit just add a test for an inference with `map`, that is possible using all the modifications of the previous commits.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1261,12 +1261,41 @@
 (test-comp '(lambda (f l) (f l) #t)
            '(lambda (f l) (f l) (procedure? f)))
 
+(test-comp '(lambda (z) (let ([o #f]) (car z)) #t)
+           '(lambda (z) (let ([o #f]) (car z)) (pair? z)))
+(test-comp '(lambda (z) (let ([o (random)]) (car z)) #t)
+           '(lambda (z) (let ([o (random)]) (car z)) (pair? z)))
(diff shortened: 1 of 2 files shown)


mflatt:
This looks good to me. In particular, I detect no problems with the shifts.

In the implementation of intersection, I think it would be a little clearer to set `base_pred` to NULL when `base_types` is NULL – or somehow avoid the relatively subtle invariant that `base_pred` is still initialized to NULL when `base_types` is NULL.

gus-massa:
Do the `Scheme_Hash_Tree` have a method to count the number of items in them? Is this method O(n) or O(1)? 

I guess that the `else` brach usually adds less information than the `then` branch, so I use the `f_tree` in the outer loop and I use `hash_tree_get` to see if the item is in the `t_tree`.

gus-massa:
I just changed the intersection method to make the use of `NULL` clearer and I use `tree->count` to select the smaller tree. The other commits are unchanged.

mflatt:
Looks good to me.

PR #938, opened 2015-05-20
Initial fix for handling truncated zip files.

samth:


samth:
This is the code @rfindler and I developed in London to attempt to fix the test failure seen [here](http://pkg-build.racket-lang.org/server/built/test-fail/io.txt) in the "io" package. It's far from clear that this is the right fix, but it fixes the bug and doesn't break any other tests.

PR #939, opened 2015-05-21
Fix the first order check on async-channel/c

lexi-lambda:
This is my fault. When I wrote this initially I had a shaky understanding of how the contract system worked internally, and the first order check was wrong. Without this fix, simply calling `contract-first-order` on an async channel contract would throw an exception because the first order check isn't properly curried.

Anyway, it's fixed now, and I've added a test covering it.


samth:
@rfindler or @takikawa, are you the right people to look at this?

rfindler:
The commit looks good to me.

lexi-lambda:
Just a small bump and a request for this to be merged, so long as everything looks fine (which so far appears to be the case).

rfindler:
I've pushed it. Sorry for the delay.

lexi-lambda:
No problem; thanks for merging it.

PR #940, opened 2015-05-21
fixing path->string to work with explode-path

florence:
Currently `(map path->string (explode-path "../something"))` errors, because `explode-path` produces an `'up`. As noted in https://github.com/racket/compiler/pull/1#discussion_r29418697 it seems like `path->string` should handle this.

Does this seem reasonable? Should `some-system-path->string` do the same?


samth:
If we want to resurrect this, it needs a CS implementation (in racket/src/io).

mflatt:
I'm skeptical that this is a good idea. (I also vaguely remember a related discussion somewhere else, but I can't find it, so maybe I misremember.)

A problem with the motivation is that `path->string` is not a good choice for the path elements produced by `explode-path`. If there's any right operation, it would be `path-element->string`. Even then, I can't quite imagine where that would be a good idea. So, maybe this is one of those cases where the mismatch is a symptom of going the wrong direction, or maybe there's a good use case where more explanation would help.

samth:
Looking at the (not-actually-used) code in racket/compiler#1, it doesn't seem like the place it was invented for was a good idea, so I'm just going to close this.

PR #941, opened 2015-05-23
Add stream* to complement stream

lexi-lambda:
Adds `stream*` to `racket/stream`, which functions similarly to `list*`, but produces a stream.


lexi-lambda:
Can I get some feedback on this and/or get it merged?

mflatt:
I'll merge this, but with two extra tests to provide multiple elements to `stream*`.

PR #943, opened 2015-05-25
and/c: use projections more on flat-contract arguments

AlexKnauth:
Should fix the problem outlined here:
http://www.mail-archive.com/racket-users@googlegroups.com/msg27144.html
http://www.mail-archive.com/racket-users@googlegroups.com/msg27163.html

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/context.rkt
+++ b/pkgs/racket-test/tests/racket/contract/context.rkt
@@ -491,6 +491,11 @@
                                  'pos 'neg)
                        save-file))
   
+  (context-test '("an and/c case of")
+                '(contract (and/c integer? positive?)
+                           5.9
+                           'pos 'neg))
+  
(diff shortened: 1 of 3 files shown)


rfindler:
Can you add a test case somewhere to demonstrate the difference with this change? Maybe in `tests/racket/contract/context` ?

AlexKnauth:
I added a test in `tests/racket/contract/context.rkt` and fixed another test in `tests/racket/contract/or-and.rkt`.

lexi-lambda:
I haven't tested this or looked it over in detail, but the premise of the change looks good to me.

rfindler:
I've pushed this. Does it help, Alexis?

lexi-lambda:
Yep, this works well! Thanks to both of you. I'll tweak my own error message slightly to fit this better, but this is definitely what I was looking for.

PR #945, opened 2015-05-26
Preprocessed C code requires the use of CFLAGS

juanfra684:
Racket doesn't use the CFLAGS for the preprocessed C code (gcc -E) which
is not consistent with the later use of the C compiler to compile the
code.

Silences a warning when -D_FORTIFY_SOURCE=2 is used on Linux:

In file included from /usr/include/inttypes.h:25:0,
                 from ./../include/../mzconfig.h:36,
                 from ./../include/../sconfig.h:20,
                 from ./../include/scheme.h:33,
                 from ./../src/schpriv.h:23,
                 from ./precomp.c:2:
/usr/include/features.h:328:4: warning: #warning _FORTIFY_SOURCE requires compiling with optimization (-O) [-Wcpp]
 #  warning _FORTIFY_SOURCE requires compiling with optimization (-O)

**Only tested on OpenBSD. I don't know if the change breaks something on Windows with VS toolkit.**


mflatt:
Shouldn't flags needed for preprocessing should be in `CPPFLAGS` instead of `CFLAGS`?

juanfra684:
No, because the `-O` flags are part of the compiler, a real C preprocessor wouldn't support this type of flags.

mflatt:
I think that passing `CFLAGS` to the proprocessor causes problems on some platforms and configurations, but I don't remember how it goes, exactly. (Maybe it has to do with platforms with a "real" preprocessor.) Given how many times I've been around that circle – where each time I end up concluding that `CFLAGS` and `CPPFLAGS` really are different and should not be confused – I'm reluctant to guess again.
Meanwhile, if `-D_FORTIFY_SOURCE=2` requires that `-O` is specified at preprocessing time, then `-O` must be a preprocessor option, so I think it's fair to require `-O` added to `CPPFLAGS` in that case. Note that all `CPPFLAGS` are passed to the C compiler.

juanfra684:
I see your point. How about this?. The code would be in the section of Linux (FORTIFY_SOURCE is only available on glibc) within configure.ac and configure:

```shell
if test "${CPP}" = "gcc -E" || test "${CPP}" = "clang -E" || test "${CPP}" = "cc -E" ; then
      OPTGREP=`echo "${CFLAGS}" | egrep -o '\-O[0-9]'`
      CPPFLAGS="${OPTGREP} ${CPPFLAGS}"
```

mflatt:
After thinking about it a lot more, I'm inclined to blame autoconf's selection of `CC`, `CPP`, `CFLAGS`, and `CPPFLAGS` – and try to fix things up at that level:

https://github.com/mflatt/racket/commit/7dee02f3735ce2d91a569d4fe9ce6ef8df34c245

In other words, if a programmer specifies any of the first three, the programer takes on the burden of keeping the C compiler consistent with the preprocessor. Otherwise, we make automatic selection set the flags in a way that is consistent with how the tools work.

How does that look?

juanfra684:
Yes, that looks better. Feel free to commit the change.

PR #948, opened 2015-05-29
url-strings: Separate URL parsing from URL actions

tim-brown:
`net/url` provides functions for both converting strings
and paths to and from URLs.

`net/url` also includes functions for creating (pure and import)
network ports. This functionality `require` the HTTP client stack
which is unnecessary when URLs simple need parsing for their
"bits".

New library: `net/url-strings` handles `url->string` and `string->url`
(and also the related `path->url` and `url->path` functions). This is
required by net/url for compatability.

`net/url-exception.rkt` is factored out for use by both libraries.

- See also racket/net changes for T&D

### DIFF
--- a/racket/collects/net/url-exception.rkt
+++ b/racket/collects/net/url-exception.rkt
@@ -1,0 +1,19 @@
+#lang racket/base
+(require racket/string
+         racket/contract/base
+         racket/list)
+
+(define-struct (url-exception exn:fail) ())
+(define (-url-exception? x)
+  (or (url-exception? x)
+
+      ;; two of the errors that string->url can raise are
+      ;; now contract violations instead of url-expcetion
+      ;; structs. since only the url-exception? predicate
+      ;; was exported, we just add this in to the predicate
+      ;; to preserve backwards compatibility
+      (and (exn:fail:contract? x)
+           (regexp-match? #rx"^string->url:" (exn-message x)))))
+
+(provide (struct-out url-exception))
+(provide/contract (-url-exception? (any/c . -> . boolean?)))
--- a/racket/collects/net/url-string.rkt
+++ b/racket/collects/net/url-string.rkt
@@ -1,0 +1,400 @@
+#lang racket/base
+(require racket/string
+         racket/contract/base
+         racket/list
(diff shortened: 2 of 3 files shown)


mflatt:
This looks great to me. Two requests:

 * I think the library name should be `net/url-string` (i.e., single case, consistent with libraries like
   `racket/string` – granting that, unfortunately, `net/url-structs` didn't follow that convention).

 * The `net/url` library should `(provide (all-from-out "url-string.rkt"))` instead of re-listing functions and
   adding an extra layer of contracts.

tim-brown:
On 29/05/2015 21:53, Matthew Flatt wrote:
> This looks great to me. Two requests:
>
>   *
>
>     I think the library name should be |net/url-string| (i.e., single
>     case, consistent with libraries like
>     |racket/string| – granting that, unfortunately, |net/url-structs|
>     didn't follow that convention).
>
>   *
>
>     The |net/url| library should |(provide (all-from-out
>     "url-string.rkt"))| instead of re-listing functions and
>     adding an extra layer of contracts.

Does (provide (all-from-out ...)) preserves the contract provisions?
That's the reason I chose to keep the provide/contracts as they were
passed up to net/url.

Tim

---
This email has been checked for viruses by Avast antivirus software.
http://www.avast.com

mflatt:
Yes, `(all-from-out)` keeps contracts.

tim-brown:
Is this pull request waiting on me for anything?

mflatt:
@tim-brown: Assuming we agree on the changes, did you already make a patch with the library name changed to `net/url-string` and re-exporting from `net/url` without an extra contract layer? I was hoping that you'd update the pull request with those changes (but haven't pressed or tried to do it myself, because I've been occupied over the past couple of weeks). Thanks!

tim-brown:
I think this is ready to merge... AppVeyor is complaining about something in places -- I don’t think it’s my fault.

samth:
Yes, that error from AppVeyor is unrelated.

tim-brown:
again: is there anything holding this up from my end?

stamourv:
Merged.

PR #949, opened 2015-05-29
Have (ctype-sizeof _bool) report SIZEOF_BOOL

lehitoskin:
_bool should not be sizeof(int) since the compiled sizeof(bool) is 1 byte and not 4 bytes. This is a real problem when using cstructs that contain bools: doing operations on a library-generated struct that contains bool variables will segfault.


mflatt:
Please use `_stdbool` instead of `_bool` to get a C99 boolean.

For more information, see http://lists.racket-lang.org/dev/archive/2014-April/014028.html

lehitoskin:
This is exactly what I need. No idea how I could have missed _stdbool. Thanks for clearing this up.

PR #950, opened 2015-05-29
Add racket/os library.

takikawa:
Adds a `racket/os` library that contains just two functions from `mzlib/os` so far. It would be nice to extend it in the future with more functions that query OS information. I'll also submit a PR to change `mzlib/os` to just re-provide these functions.


mflatt:
This looks good to me.

mflatt:
Looks good to me.

mflatt:
Merged (with minor changes, mainly for the doc location)

PR #954, opened 2015-06-03
Make pair accessor shorthand examples properly link to relevant identifier documentation

lexi-lambda:
The macro that generates the documentation for the `c*r` pair accessor shorthand does not properly include lexical context, so the resulting examples are not hyperlinked to the proper docs. This fixes that.


mflatt:
Merged - thanks!

PR #955, opened 2015-06-04
Fix contract error in in-vector documentation example

lexi-lambda:
The example given for `in-vector` uses an immutable hash where it should use a mutable one, which raises a contract error that should definitely not be a part of the example. This changes it to use a mutable hash.


rfindler:
Thanks! I've pushed the commit.

PR #956, opened 2015-06-05
Generalize inferred names

gus-massa:
After some expansions, a expression with the syntax property `'inferred-name` of
`'x` is converted to one with ` ('x . 'x) `, so it's not useful to get the name of a
procedure. 

So we simplify the syntax property `'inferred-name` to handle
these cases, to trees deep <=5, where all the leaves are the same symbol. (In the wild, I only saw the 1-level case ` ('x . 'x) `).

The same simplification is useful for properties like ` (#<void> . #<void>)` that are simplified to `#<void>` and replace the local inferred name with the location.

This commit also changes `syntax-local-infer-name` in the case that the property is not a symbol. This is a small breaking change, but the new behavior is more consistent with the names of the procedures.

Some examples:

    #lang racket

    (define-syntax (thunk1 stx)
      (syntax-case stx ()
        [(_ v)
         (begin
           #;(displayln (eval #'v))
           (syntax-property
            #'(lambda () 1)
            'inferred-name
            (eval #'v)))]))

     (let ([tmp (thunk1 (cons 'one 'one))]) tmp)     ;==> #<procedure:tmp> 
                                                          ;changes to #<procedure:one>
     (let ([tmp (thunk1 (cons 1 1))]) tmp)           ;==> #<procedure:tmp>
     (let ([tmp (thunk1 (cons (void) (void)))]) tmp) ;==> #<procedure:tmp> 
                                                          ;changes to #<procedure:test.rkt:77:77>
     (let ([tmp (thunk1 (cons #f #f))]) tmp)         ;==> #<procedure:tmp>
     (let ([tmp (thunk1 (cons "one" "one"))]) tmp)   ;==> #<procedure:tmp>

TODO:
* Fix the documentation of `'inferred-name` and `syntax-local-infer-name`
* Add tests. (I'm not sure where I should to put them)

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/syntax-model.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax-model.scrbl
@@ -984,6 +984,12 @@ property value of @|void-const| hides a name that would otherwise be
 inferred from context (perhaps because a binding identifier's was 
 automatically generated and should not be exposed).
 
+To support the propagation of properties during expansions, instead
+of a symbol it is possible to use a tree made of @racket[cons] where
+all the leaves are that symbol, for example @racket[('name . 'name)]
+is equivalent to @racket['name]. Similarly, @racket[(<void> . <void>)]
+is equivalent to @|void-const|.
+
 When an inferred name is not available, but a source location is
 available, a name is constructed using the source location
 information. Inferred and property-assigned names are also available
--- a/pkgs/racket-doc/syntax/scribblings/define.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/define.scrbl
@@ -27,7 +27,7 @@ If the definition is ill-formed, a syntax error is raised. If
 an expression context. The default value of @racket[check-context?] is
 @racket[#t].
 
-If @racket[opt-kws?] is @racket[#t], then arguments of the form
+If @racket[opt+kws?] is @racket[#t], then arguments of the form
 @racket[[id expr]], @racket[keyword id], and @racket[keyword [id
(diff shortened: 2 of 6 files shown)


mflatt:
Good – thanks! I think tests go in "name.rktl".

gus-massa:
I added some tests and a draft of the documentation changes.

I also changed `scheme/unit` to `racket/unit`. Is this ok?

I still don't like the new redaction of the docs, and I couldn't typeset correctly `(#<void> . #<void>)`. I propose to squash and merge all but the docs changes, and perhaps you can later fix the docs.

gus-massa:
I'm stuck here. Is it ok to merge this without updating the docs?

mflatt:
I'd lost track of this one. The change looks generally ok, although I don't think a fuel limit is needed here.

For the docs, I think the right solution is to use expressions: `(cons 'x 'x)` or `(cons (void) (void))`. (Using `('x . 'x)` seems like a hybrid of `write` and `print`.)

Since I've let this sit so long, I'll adjust and merge.

mflatt:
merged

gus-massa:
I've just read the email with the notification, thanks for fixing the details.

I added the fuel to avoid problems with circular lists and cyclic trees. It's not an expected use case. It was esier to choose an arbitrary bound than detect all the posible cicles in the `car` and `cdr` parts.

IIRC `#<void>` is difficult to typeset in scribble. The official way to typeset `#<void>` uses a custom function that creates the correct link. I think that `(void)` alone doesn't create a link.

PR #961, opened 2015-06-12
Add subprocess-signal.

Hardmath123:
This PR creates a new primitive, `subprocess-signal`, which allows users to send arbitrary POSIX signals to a subprocess. There are many, many use-cases for this, such as communicating with obscure processes that accept `SIGUSR*`, suspending and resuming process with `SIGSTOP` and `SIGCONT`, and even implementing `subprocess-kill` within Racket rather than as a separate C primitive.

A simple test case and example is:

```scheme
#lang racket
(define-values (p stdout stdin stderr) (subprocess #f #f #f "/usr/bin/python" "signal.py"))
(display (read-line stdout))
(newline)
(subprocess-signal p 'SIGTERM) ; or any other signal name like SIGUSR1
(display (read-line stdout))
(newline)
```

In file `signal.py`, we have:

```python
import signal
import sys
import time

print "Bananas!"
sys.stdout.flush()

def beep(num, stack):
    print "Got signal:", num
    sys.stdout.flush()

signal.signal(signal.SIGHUP, beep)
signal.signal(signal.SIGINT, beep)
signal.signal(signal.SIGQUIT, beep)
signal.signal(signal.SIGILL, beep)
signal.signal(signal.SIGTRAP, beep)
signal.signal(signal.SIGABRT, beep)
signal.signal(signal.SIGEMT, beep)
signal.signal(signal.SIGFPE, beep)
#signal.signal(signal.SIGKILL, beep)
signal.signal(signal.SIGBUS, beep)
signal.signal(signal.SIGSEGV, beep)
signal.signal(signal.SIGSYS, beep)
signal.signal(signal.SIGPIPE, beep)
signal.signal(signal.SIGALRM, beep)
signal.signal(signal.SIGTERM, beep)
signal.signal(signal.SIGURG, beep)
#signal.signal(signal.SIGSTOP, beep)
signal.signal(signal.SIGTSTP, beep)
#signal.signal(signal.SIGCONT, beep)
signal.signal(signal.SIGCHLD, beep)
signal.signal(signal.SIGTTIN, beep)
signal.signal(signal.SIGTTOU, beep)
signal.signal(signal.SIGIO, beep)
signal.signal(signal.SIGXCPU, beep)
signal.signal(signal.SIGXFSZ, beep)
signal.signal(signal.SIGVTALRM, beep)
signal.signal(signal.SIGPROF, beep)
signal.signal(signal.SIGWINCH, beep)
signal.signal(signal.SIGINFO, beep)
signal.signal(signal.SIGUSR1, beep)
signal.signal(signal.SIGUSR2, beep)

while True:
    time.sleep(1)
```

This feature is UNIX-only, though a small subset of the signals have Windows equivalents which I can try to implement later.

### DIFF
--- a/racket/src/racket/src/port.c
+++ b/racket/src/racket/src/port.c
@@ -430,6 +430,7 @@ static void flush_if_output_fds(Scheme_Object *o, Scheme_Close_Custodian_Client 
 static Scheme_Object *subprocess(int c, Scheme_Object *args[]);
 static Scheme_Object *subprocess_status(int c, Scheme_Object *args[]);
 static Scheme_Object *subprocess_kill(int c, Scheme_Object *args[]);
+static Scheme_Object *subprocess_signal(int c, Scheme_Object *args[]);
 static Scheme_Object *subprocess_pid(int c, Scheme_Object *args[]);
 static Scheme_Object *subprocess_p(int c, Scheme_Object *args[]);
 static Scheme_Object *subprocess_wait(int c, Scheme_Object *args[]);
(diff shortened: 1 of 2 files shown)


Hardmath123:
Note: The reason it fails on Travis/Linux is because Linux defines a different set of signal constants than OSX does. We should use a smaller set of signals which is guaranteed to be defined everywhere.

mflatt:
Thanks for the patch! I like the idea of `subprocess-signal`, but I'd like to look more into how the set of supported signals can be generally portable and complete (where the latter might mean accepting numerical values as well as symbols). I can't get to it just now, but I'll hopefully be able to follow up soon.

sorawee:
Let me assign this PR to @mflatt then.

PR #963, opened 2015-06-15
Add pointer accessor for cstruct fields.

endobson:
The use case for this feature is allocating a struct that will hold many different values. Then passing the pointers to those values on to the foreign interface. Having them all in the same struct means that lifetime management is easier.


endobson:
ptr-add seems to leave the pointers tag the same. I don't see an easy way to allocate a new pointer. Should it just be `(cast val _pointer _pointer)`?

mflatt:
Yes, casting seems like the right way to drop tags.

Is there much of a compatibility issue here with defining new names? Offhand, I think that conflicts are unlikely, but I thought I'd mention it, just in case.

PR #966, opened 2015-06-29
Reoptimize propagated constants

gus-massa:
Reduces them to #t or #f when they are copied to a Boolean context.

For example in

    (lambda (x) (let ([z 2]) (not (if x #f z)))

reduce the bound `z` to `#t`.


mflatt:
Looks ok to me.

PR #967, opened 2015-07-01
disabled HAVE_POLL_SYSCALL in solaris

wargrey:
poll(2) is the CPU eater, and it affects all TCP applications including the plt-web-server.

I am not sure if this is a bug of poll(2), or just a poor implementation in RVM.
It's reasonable to suggest that it could be improved with another Event Completion Framework in the future.


mflatt:
Thanks! I'll copy the change to the source "configure.ac" and merge the change.

mflatt:
merged

PR #968, opened 2015-07-04
A bugfix and more types features

gus-massa:
The first commit fix bug, the others are a few new related features.

**1)** When a **mutable** variable with a known type is in a Boolean context, it's incorrectly reduced to `#t`.
This fixes the bug twice (each change is enough to fix it). 
* Don't reduce mutable variables with a type to `#t` in a Boolean context.
* Don't record the type of mutable variables when a predicate is checked in a test condition. (This is more consistent with other parts that record types, like `check_known`.)

**2)** Reduce variables with type `null?` to `null`. The optimizer reduces the variables with a known type to `#t` in a Boolean context. But some predicates imply that the variable has a definite value, so they can be reduced in a non-Boolean context too. For example, in `(lambda (x) (if (null? x) x 0)))` reduce the last `x` to `null`.
This also adds `void? ` and `eof-object? ` to the relevant predicates list and when a variable has these types reduce it to `#<void>` or `eof` respectively.

**3)** Infer type from `eq?` comparisons in test positions. For emple, in `(if (eq? x <pred?-expr>) <tbranch> <fbranch>)` infer that the type of `x` is `pred?` in the `<tbranch>`. This extends the previous type inference that recorded types from constructions like `(if (pred? x) <tbranch> <fbranch>)`.
This also, reduce `(eq? x y)` to `#f` when the types of `x` and `y` are different, to reduce the unnecessary types recording.

 *[Note: This idea can be extended to handle `equal?` and `eqv?`, but the relevant predicates list is not guaranty to be `equal?-disjoint` in the future, so I used only `eq?`.]*

**4)** Add `not` to the relevant predicates list. Previously all the predicates recognized only `non-#f` things, so `not` can be safely added to the list of disjoint predicates. But many of the parts of the optimizer relied on the `non-#f` property and had to be modified.
This also records that in `(if x <tbranch> <fbranch>)` the type of `x` is `not` in the `<fbranch>` so it can be reduced to `#f`.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1763,7 +1763,8 @@
                     `(lambda (x) (if (,pred? x) x 0))))])
   (test-pred-implies-val 'null? 'null)
   (test-pred-implies-val 'void? '(void))
-  (test-pred-implies-val 'eof-object? 'eof))
+  (test-pred-implies-val 'eof-object? 'eof)
+  (test-pred-implies-val 'not '#f))
 (test-comp '(lambda (x) (if (null? x) 1 0) null)
            '(lambda (x) (if (null? x) 1 0) x)
            #f)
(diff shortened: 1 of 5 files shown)


mflatt:
These changes look good to me.

PR #970, opened 2015-07-11
Remove innecesary wraps in values

gus-massa:
While reducing some ignored expressions, the optimizer may wrap the arguments `<expr>` in `(values <expr>)` when it isn't sure that they are single-value non-cm expression. 

1) Many of the portfun.c primitives are defined with `GLOBAL_PRIM_W_ARITY` but I think they can be defined with  `GLOBAL_NONCM_PRIM`. The only exception is `filesystem-change-evt` because it may tailcall a thunk in case of an error. The most common case is `(values (get-output-string x))`.

2) The expression `(values ...)` is non-cm, but the number of return values is clearly variable. But in case it has only one argument, it can be marked as singled valued too. This eliminate the construction of same expressions like `(values (values (f))`.

I didn't see too many cases where 3) and 4) are relevant, but they are quite straightforward.

3) To check that `(if <t> <tb> <fb>)` is a s-v-noncm it's enough to check `<tb>` and `<fb>` are. If `<t>` is not singled-valued, then it will generate an error that will get caught elsewhere.

4) Add the cases for `begin` and `begin0` to the s-v-noncm test, and makes the code of the `let` more general. With this change, it jumps directly to the tail of a `begin` or `let` form, so the same level of fuel allows more nesting.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1563,6 +1563,17 @@
 (test-comp '(lambda (w) (if (void (list w)) 1 2))
            '(lambda (w) 1))
 
+; Diferent number of argumets use different codepaths
+(test-comp '(lambda (f x) (void))
+           '(lambda (f x) (void (list))))
+(test-comp '(lambda (f x) (begin (values (f x)) (void)))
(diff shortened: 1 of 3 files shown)


mflatt:
LGTM

PR #972, opened 2015-07-16
Start moving contents of unstable-list-lib to a more stable place

stamourv:
This moves the contents of `unstable/function` to `racket/function`, and most of the contents of `unstable/list` to `racket/list`. Some of those functions are useful (and used), so I think we should include them in the standard library to encourage their use.

In the process, I changed the interface to some of those functions (list prefix functions and `check-duplicate`) to align it more closely with the conventions of `racket/list`. None of those changes break uses of those functions, either in the main distribution or in packages.

Once this gets in, I'll push commits to the `unstable` repo to have it just re-export the relevant functions from `racket/*` (for backwards compatibility). I'm also working on PRs to replace uses of `unstable/*` with uses of `racket/*`. After that, I plan to tackle the other subcollections provided by `unstable-list-lib`.

Some questions that remain open:
* Should we change the name of `check-duplicate` to `check-duplicates`, to mirror `remove-duplicates`?
* What to do about the remaining functions in `unstable/list`. Each of the functions I did not move are used in a single package (usually TR) and could be moved there as internal helper functions. If we want to preserve backwards compatibility for users of `unstable`, that would mean duplicating the implementation.
* What do to with the leftover `unstable-list-lib` package, and the collections it provides (once I'm done with the other subcollections in there)? I'd be in favor of removing the package from the main distribution, but keeping it available as a separate package, for backwards compatiblity.

Feedback welcome.


samth:
This is great! Thanks for getting this started.

On the three questions:
1. I'm in favor of changing the name, as long as we also change all clients.
2. I think we should make a new package called "list-extra" or something like that, and put the remaining functions there. Then TR etc can depend on that package. But maybe looking at the functions individually makes more sense.
3. I agree with your plan for keeping the pkg.

stamourv:
1. The clients I found are TR and `refined-acl2`. We can change the former, and I can submit a PR to the latter.
2. The remaining functions are:
 * `filter-multiple`: one use, in TR
 * `extend`: a few uses, in TR. IMO, the name is too general (a lot of code uses it to mean other things, including other parts of TR), and the function is of limited usefulness, so I'd be inclined to just move it to TR, and rename it.
 * `map/values`: one use, in TR. Also, `for/lists` covers most of its use cases.
 * `map2`: one use, in `minipascal`. Special case of the above.
Given that this is all that's left, and that those functions are not (IMO) very compelling, I'm not sure we'd want a new package for them.
3. Good. How do I remove it from the main-distribution roster?

Oh, and 4., which I forgot to bring up:
* Some of the functions I moved to `racket/list` may make more sense in `racket/base`, to be with similar functions. `list-update` and `list-set` feel like they belong with `list-ref`, and `remf` and `remf*` feel like they belong with `remove` and `memf`. Thoughts?

stamourv:
Update:
* Moved `unstable/hash`.
* Decided against moving `unstable/bytes` (only two packages use it, and there was disagreement on its usefulness in the comments). I have updated its uses to remove the dependency, and will submit PRs once this one gets in.
* Waiting on @rmculpepper's input re what to do with `unstable/class-iop` (he's its author and sole user). That's the last collection in `unstable-list-lib`.
* Changed `check-duplicate` to `check-duplicates`. Undecided on the name, as there is both a `remove-duplicates` and `check-duplicate-identifier`, so it's not clear which one to align to.
* Fixed all packages that use `unstable-list-lib` to remove the dependency (with the exception of those that use `unstable/class-iop`). I will send pushes and PRs once this one gets in.

samth:
* If we're not moving `unstable/bytes` (which seems fine to me) we should still get rid of it in this package.
* I would create the PRs now for other repos so that we can see what they'll look like.
* I think similarity with `remove-duplicates` is more compelling than `check-duplicate-identifier`, which is a lot more specialized.
* I think adding more things to `racket/base` should be avoided.
* For the various list functions:
  - `extend` is a terrible name, and not a very useful function. Haskell's list libraries may have a better name, but either way we should just confine this to TR.
  - I like `filter-multiple`, and would keep it.
  - I agree about `map/values`.
  - `map2` seems more likely to be useful. @soegaard, what do you think, since it's used in minipascal?

To remove things from `main-distribution`, just take them out of the relevant `info.rkt` file.

stamourv:
* Re `unstable/bytes`: I'm not sure I understand what you mean. My plan is to remove all uses of it, but leave it in the `unstable-list-lib` package (which still provides everything it always has, though now mostly with re-exports), which would be out of the main distribution, and so not generally available. That course of action has the advantage of preserving backwards compatibility. Are you suggesting that we remove `unstable/bytes` altogether?

* Re PRs: They're all ready. IIRC (wrote them on Friday), they're mostly removing/changing `require`s. I'll send them out today.

* Re `filter-multiple`: It took me a few reads of the documentation to figure out what it did, so I don't think the interface is great. I'm also not fond of the name, since it feels closer to `partition` than to `filter` to me. Let's leave it out for now, and feel free to submit another PR to include it.

* Re `map/values` and `map2`: Both can be replaced by `for/lists`, so I don't think we're losing much.

samth:
* `unstable/bytes`: I'm agreeing with you.
* `filter-multiple`: it's not like `partition`, because input elements can end up in multiple output lists. A better name is welcome. But I'm happy to do whatever you prefer here.

soegaard:
2015-07-20 15:28 GMT+02:00 Sam Tobin-Hochstadt <notifications@github.com>:
>
>
>    -
>       - map2 seems more likely to be useful. @soegaard
>       <https://github.com/soegaard>, what do you think, since it's used
>       in minipascal?
>
> I don't know. I had to look up what it did. If I recall correctly, I wrote
the equivalent of map2 for MiniPascal. Then stumbled upon map2 in unstable
the day after - and used that instead.

/Jens Axel

stamourv:
@samth: Let's think of a better name for `filter-multiple` (and maybe a better interface) and revisit the question.

@soegaard: FWIW, I have a PR forthcoming that replaces it with `for/lists`.

stamourv:
Merged.

PR #977, opened 2015-07-22
Provide `promise/name?`.

samth:
This is needed for Typed Racket, since by-name promises are not
idempotent and thus can't be typed like existing promises.
Therefore, the contract for `(Promiseof T)` needs to reject
them.

In particular, see racket/typed-racket#159.


mflatt:
Looks fine to me

PR #981, opened 2015-07-23
raco pkg new template updates for travis.yml

apg:
- Utilize RACKET_DIR for prefix to raco commands
- Add 6.2 as a build matrix option.

### DIFF
--- a/racket/collects/pkg/private/new.rkt
+++ b/racket/collects/pkg/private/new.rkt
@@ -99,11 +99,7 @@ language: c
 
 # Based from: https://github.com/greghendershott/travis-racket
 
-# Optional: To use Travis CI's newer container infrastucture,
-# un-comment the following line. (Also be sure RACKET_DIR is set to
-# somewhere like ~/racket that doesn't require sudo.)
-#
-# sudo: false
+sudo: false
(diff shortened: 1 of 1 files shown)


apg:
I'm not sure why this failed. Will pull in master and try again.

samth:
None of those failures are your fault -- the Travis ones are intermittent errors that I've seen before, and the Appveyor one is weird but not your fault.

samth:
Also, we should probably switch to `sudo: false`, since Travis is now explicitly marking other builds as "legacy".

apg:
Ok. I can add a sudo: false. Will add another commit.

On Thu, Jul 23, 2015 at 1:58 PM, Sam Tobin-Hochstadt <
notifications@github.com> wrote:

> Also, we should probably switch to sudo: false, since Travis is now
> explicitly marking other builds as "legacy".
>
> —
> <https://github.com/plt/racket/pull/981#issuecomment-124238525>.
>



-- 
http://www.apgwoz.com

sorawee:
Thanks; the current `raco pkg new` template fixes all of these issues already, so I think we should close this PR.

PR #984, opened 2015-07-25
Jitinline string-length and bytes-length

gus-massa:
Previously only the `unsafe-X-length` versions were inlined in the jit compiler.

(I didn't write jit code before, I hope I got all the details correctly.)

One small detail is the name conventions of the the intermediate constructions. At the C level the names are `char_string` and `byte_string`, but they are renamed at the Racket level as `string` and `bytes`. So I'm not sure if `bad_string_length_code` and `bad_bytes_length_code` are the best names election.

### DIFF
--- a/racket/src/racket/src/jit.h
+++ b/racket/src/racket/src/jit.h
@@ -313,6 +313,8 @@ struct scheme_jit_common_record {
   void *bad_vector_length_code;
   void *bad_flvector_length_code;
   void *bad_fxvector_length_code;
+  void *bad_string_length_code;
+  void *bad_bytes_length_code;
   void *vector_ref_code, *vector_ref_check_index_code, *vector_set_code, *vector_set_check_index_code;
   void *chap_vector_ref_code, *chap_vector_ref_check_index_code, *chap_vector_set_code, *chap_vector_set_check_index_code;
   void *string_ref_code, *string_ref_check_index_code, *string_set_code, *string_set_check_index_code;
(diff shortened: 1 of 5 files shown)


mflatt:
I think you need a JIT_UPDATE_THREAD_RSPTR() in each of the common functions, which ensures that the global runstack pointer used by C code (including the GC) is synced with the register that JITted code uses.

I don't see any other problems, but it's difficult to be sure. The names of the common-function pointers are fine with me.

gus-massa:
I just added them to the github branch.

 I copied the code from the `bad_XXvector_length_code`. They don't have any `JIT_UPDATE_THREAD_RSPTR()`. Is that ok?

--
Travis just finished with an error. It's the same error in the pull request in plt/racket/ travis test and in my gus-massa/racket travis test.

    $ export RACKET_CONFIGURE_ARGS="--disable-jit"
    [...]    
    Section(namespaces)
    ERROR: id: unbound identifier;
     also, no #%top syntax transformer is bound
      in: id
    The command "raco test -l tests/racket/test" exited with 2.

mflatt:
Missing `JIT_UPDATE_THREAD_RSPTR()` in `bad_XXvector_length_code` seems likely to be a bug.

I just saw the `Section(namespaces)` failure in my build without this change, too, so it must be unrelated.

gus-massa:
So I just rebase and push this?

mflatt:
Yes

mflatt:
Just for the record, that really needs to be `JIT_UPDATE_THREAD_RSPTR()`, not `JIT_UPDATE_THREAD_RSPTR_IF_NEEDED()`. The "common" stubs don't have the context to say whether an update is needed, so they should always update.

I see that one other place is broken along those lines, and I'll also add the missing calls in the vector functions.

gus-massa:
Ups! I copied it from `box_cas_fail_code` in line 545. You should check that.

PR #985, opened 2015-07-27
Fix bugs/typos in docs for contracts and quote-syntax

btlachance:
1st is a small grammatical mistake
2nd is in a section about ->* yet mistakenly -> is referred to
3rd is about recontract-out yet contract-out is mentioned instead
4th clarifies return value for value-contract
5th replaces free-identifier? with free-identifier=?


rfindler:
The contract changes look good to me. Thanks!

Robby

mflatt:
merged - thanks!

btlachance:
D"oh, thanks for the formatting fixes

PR #986, opened 2015-07-28
Update INSTALL.txt

enedil:
I removed reference to OS X in Unix section since OS X is a certified UNIX, in contrary to for example Linux.


mflatt:
Thanks for the patch! I ended up revising "INSTALL.txt" in other ways, based on your suggestion (see a3e359d9e167365887b0a9b4948c7fe859869021)

PR #987, opened 2015-07-28
Add string-startswith? and string-endswith?

bennn:
Two little functions that I think should be upgraded from [srfi-13](http://docs.racket-lang.org/srfi/srfi-std/srfi-13.html#PrefixesSuffixes).

People coming from [Java](http://docs.oracle.com/javase/7/docs/api/java/lang/String.html#startsWith(java.lang.String)), [Python](https://docs.python.org/3/library/stdtypes.html#str.startswith), [C#](https://msdn.microsoft.com/en-us/library/baketfxw(v=vs.110).aspx), or [JavaScript](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/startsWith) might like them too.


stamourv:
I'd prefer calling them `string-prefix?` and `string-suffix?`, for consistency with the newly added `list-prefix?` family of functions.

Also, I'm not sure these particular shortcuts are all that useful given the existence of `regexp-match` and co. ISTR us deciding against adding similar shortcuts to `racket/string` for that reason in the past.

bennn:
I figured there was a reason there weren't added already, and did find a brief [mailing list thread](http://lists.racket-lang.org/users/archive/2012-April/051406.html) (their conclusion: use `regexp` or `srfi-13`).

But IMO these are convenient, self-documenting functions. Just like `list-prefix?` and `sixth` are nice to have.

samth:
I think we should add useful string functions, instead of expecting everyone to use regexps. Learning regexp syntax is much more complicated than a simple function call.

jackfirth:
Strongly agree with @samth here. Simple string operations for common cases are very helpful, even for regexp masters.

gus-massa:
The implementation of `string-replace` uses regexps. https://github.com/plt/racket/blob/master/racket/collects/racket/string.rkt#L103-L117

It's tricky, so it's not straightforward to use the regexp because you need to correctly escape the string, and cache the escaped version to improve the performance. So I agree that it's good to include specialized functions.

I didn't test the time, but I think that an implementation of `string-prefix/sufix` based in regexps will be faster than the proposed implementation that use `for`.

bennn:
@gus-massa good point, I will run some benchmarks & share the results here

bennn:
Unless I've done something wrong, the non-regexp implementation is winning.
```
Running `string-prefix?` on all pairs of words in 'common-words.rktd'
Average of 30 runs: 416.43 ms
Running `string-suffix?` on all pairs of words in 'common-words.rktd'
Average of 30 runs: 685.8 ms
Running `string-prefix2` on all pairs of words in 'common-words.rktd'
Average of 30 runs: 14487.2 ms
Running `string-suffix2` on all pairs of words in 'common-words.rktd'
Average of 30 runs: 21745.23 ms
```

Code & test data is here:
https://gist.github.com/bennn/7efd60b4809a475af297

but for reference, here are the implementations:
```
#lang racket/base

(require
  (only-in racket/file file->value)
  (only-in racket/format ~r)
)

;; =============================================================================
;; Simple implementation

(define (string-prefix? str prefix)
  (and
    (<= (string-length prefix) (string-length str))
    (for/and ([c1 (in-string str)]
              [c2 (in-string prefix)])
      (char=? c1 c2))))

(define (string-suffix? str suffix)
  (define offset (- (string-length str) (string-length suffix)))
  (and
    (not (negative? offset))
    (for/and ([c1 (in-string str offset)]
              [c2 (in-string suffix)])
      (char=? c1 c2))))

;; =============================================================================
;; Regexp implementation

;; The real thing would probably share 1 cache with replace
(define prefix-cache (make-weak-hasheq))
(define suffix-cache (make-weak-hasheq))

(define (string-prefix2 s1 s2)
  (define s2* (if (regexp? s2)
                  s2
                  (hash-ref! prefix-cache s2 (lambda () (regexp-quote s2)))))
  (define s2** (regexp (string-append "^" (regexp-quote s2*))))
  (if (regexp-match s2** s1) #t #f))

(define (string-suffix2 s1 s2)
  (define s2* (if (regexp? s2)
                  s2
                  (hash-ref! suffix-cache s2 (lambda () (regexp-quote s2)))))
  (define s2** (regexp (string-append (regexp-quote s2*) "$")))
  (if (regexp-match s2* s1) #t #f))
```

gus-massa:
The regexp implementations have a few problems, for example they call `regexp-quote` twice on each string. Also, it's probably better to cache the final regexp than the intermediate escaped string. Anyway, after fixing this I get the same results than you, i.e. the regexp versions are slower.

    Running `string-prefix1?` on all pairs of words in 'common-words.rktd'
    Average of 1 runs: 1357 ms
    Running `string-suffix1?` on all pairs of words in 'common-words.rktd'
    Average of 1 runs: 1935 ms
    Running `string-prefix2?` on all pairs of words in 'common-words.rktd'
    Average of 1 runs: 49296 ms
    Running `string-suffix2?` on all pairs of words in 'common-words.rktd'
    Average of 1 runs: 52697 ms
    Running `string-prefix3?` on all pairs of words in 'common-words.rktd'
    Average of 1 runs: 7956 ms
    Running `string-suffix3?` on all pairs of words in 'common-words.rktd'
    Average of 1 runs: 8580 ms

PS: Looking at this, I found a bug in `string-replace` when the string is mutable. I've already submitted the PR.

The code, in case you want to rerun the tests:

     (define prefix-cache3 (make-weak-hasheq))
     (define suffix-cache3 (make-weak-hasheq))

     (define (string-prefix3? s1 s2)
      (define s2* (if (regexp? s2)
                      s2
                      (hash-ref! prefix-cache3 s2
                                 (lambda () (regexp (string-append "^" (regexp-quote s2)))))))
      (if (regexp-match s2* s1) #t #f))

     (define (string-suffix3? s1 s2)
      (define s2* (if (regexp? s2)
                      s2
                      (hash-ref! suffix-cache3 s2 
                                 (lambda () (regexp (string-append (regexp-quote s2) "$"))))))
      (if (regexp-match s2* s1) #t #f))

bennn:
Thanks for fixing the sloppy regexp code.

I just tried another implementation with loops & it's faster, so I've pushed it. (Not sure what this `AppVeyor` failure is about.)

Results:
```
Running `string-prefix-loop` on all pairs of words in 'common-words.rktd'
Average of 30 runs: 273.2 ms
Running `string-suffix-loop` on all pairs of words in 'common-words.rktd'
Average of 30 runs: 302.13 ms
```

Implementation:
```
(define (string-prefix-loop str prefix)
  (define l1 (string-length str))
  (define l2 (string-length prefix))
  (let loop ([i 0])
    (cond
     [(= l2 i)  #t] ;; Finished reading all chars in prefix
     [(= l1 i)  #f] ;; Prefix is longer than string
     [else      (and (char=? (string-ref str i) (string-ref prefix i))
                     (loop (add1 i)))])))

(define (string-suffix-loop str suffix)
  (define l2 (string-length suffix))
  (define offset (- (string-length str) l2))
  (and (not (negative? offset)) ;; Suffix isn't longer than string
       (let loop ([i+o offset] [i 0])
         (or (= i l2)
             (and (char=? (string-ref str i+o) (string-ref suffix i))
                  (loop (add1 i+o) (add1 i)))))))
```

stamourv:
Merged, thanks!
@bennn, could you close this? I can't.

bennn:
Yes!

PR #989, opened 2015-07-28
Fix check-syntax arrows within units

AlexKnauth:
Programs like this now show check-syntax arrows in DrRacket from the `x` definition outside the unit to the `x` in the unit body, and from the `y` definition inside the unit to the `y` in the `(define z y)`, where before both were broken.
```racket
#lang racket
(define x 3)
x
(define-signature z^ (z))
(define-unit z@
  (import)
  (export z^)
  (define y x)
  (define z y))
```

### DIFF
--- a/racket/collects/racket/unit.rkt
+++ b/racket/collects/racket/unit.rkt
@@ -942,7 +942,7 @@
        ;; the introduced parts, which we implement by adding the
        ;; mark to the introduced parts and then flipping it
        ;; evenrywehere.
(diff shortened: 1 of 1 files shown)


AlexKnauth:
This also fixes many check-syntax arrows in programs that use `#lang racket/unit`, such as many of the modules in the type-checker for `typed/racket`, where their non-existance has annoyed me for a long time.

mflatt:
merged - thanks!

PR #990, opened 2015-07-28
Fix match with (list-no-order p ..k) patterns

khinsen:
`p ..k` matches exactly k repetitions of p, but its documented behavior is to match "k or more" repetitions. This fix implements the documented behavior, and adds tests for this.

Fixes [bug number 15122](http://bugs.racket-lang.org/query/?debug=&database=default&cmd=view+audit-trail&cmd=view&pr=15122).

### DIFF
--- a/pkgs/racket-test/tests/match/examples.rkt
+++ b/pkgs/racket-test/tests/match/examples.rkt
@@ -202,6 +202,26 @@
            [(list-no-order 1 2 3 rest ...) rest]
            [_ 'no]))
 
+   (comp '(x y z)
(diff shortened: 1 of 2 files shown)


samth:
Are there already tests for `..k` patterns outside of `list-no-order`?

samth:
Thanks for the patch, btw!

khinsen:
There are `..k` patterns in `other-plt-tests.rkt` and in `other-tests.rkt`, but none of those places looks like `..k` patterns are the main topic there.

samth:
Merged, finally.

PR #991, opened 2015-07-28
typo: `prefix` struct field name

bennn:
Change `inspector-desc` to `src-inspector-desc`, to match `zo-structs.rkt`


mflatt:
merged - thanks!

PR #994, opened 2015-07-29
JIT: Inline keyword?

gus-massa:


mflatt:
The "keyword?" function in "symbol.c" needs to be marked as inlinable, right?

Tests can be added to "pkgs/racket-test-core/tests/racket/optimize.rktl" in the same way as for `symbol?`.

gus-massa:
Fixed and rebased. Is it ok now?

mflatt:
Yes, that looks right to me.

PR #995, opened 2015-07-31
Move `unstable/wrapc` and `unstable/macro-testing` to `syntax` collection

stamourv:
After discussion with @rmculpepper.


stamourv:
Update: Moved the used parts of `unstable/syntax` to `syntax/X`. Also after discussion with @rmculpepper.

stamourv:
Merged already.

PR #997, opened 2015-08-01
update packges sources

bianyali:
Add Racket package manager to package sources


mflatt:
Adjusted (the identifier `pkg` isn't bound, the PLaneT URL wasn't right!, and noted that PLaneT is the legacy site) and merged – thanks!

PR #1001, opened 2015-08-07
add define-pattern-expander

stchang:
There should be a special form to define `pattern-expander`s, to check that the name is valid.


stchang:
Not really needed anymore, due to https://github.com/racket/racket/pull/1073

PR #1003, opened 2015-08-09
JIT: Early attempt to inline string=?

gus-massa:
This works, but it is still unfinished and has a few unnecessary comments. This commit inlines `string=?`, my plan is to inline `bytes=?` too in the final version. I have a lot of doubts / comments:

* When is it necessary to use `CHECK_LIMIT()`?

* I tried to check the first character outside the loop to avoid the use of `LOCAL3`, but I didn't notice any difference in time, so I removed the check.

* This version puts `scheme_true` or `scheme_false` in `dest` and then checks if it is a `for_branch` use or a normal one. I tried to make a `for_branch` version, but I didn't notice any difference in time and the code was more confusing, so I keep the simple version.

* I´ll move the call to `mz_finish_prim_lwe(ts_scheme_string_eq_2, refr); ` to jitcommon.c

* Is it a good idea to move the main loop to jitcommon.c, and left in jitinline.c only the initial type and length checks?

* I just notice `result_ignored`. Is it ok to just check the types and return `scheme_false` in that case.


gus-massa:
I just made a small fix in line 2579: `mz_patch_UCbranch` -> `mz_patch_ucbranch`

mflatt:
A `CHECK_LIMIT()` call is needed often enough to ensure that the buffer used to general code is not over-filled. It looks like you've included enough to me.

Yes, the comparison code is long enough that I would have moved the loop to common code.

I agree about moving the call to `scheme_string_eq_2()` to common code. The call to `JIT_UPDATE_THREAD_RSPTR_IF_NEEDED()` can then be removed and changed to `JIT_UPDATE_THREAD_RSPTR()` in the common code, but since `string=?` is almost always used in a branch where and update will be inserted, anyway, I guess that won't make a difference.

If you don't want to specialize `for_branch` mode, you can just not inline in that mode, and inlining will be tried again in regular mode. I recommend specializing `for_branch` mode, though. I'm not too surprised that you couldn't measure a difference, but think the effect of `for_branch` mode is that it keeps the generated code more compact overall.

Yes, if `result_ignored` is set, then the generated code could make sure that the arguments are strings and do nothing else (i.e, not even return false).

Don't call `mz_pop_locals()`. That's for the entire enclosing function. There's no need to balance `mz_set_local_p()`s with mz_get_local_p()`s.

I think the main loop might generate too much code for a tiny jump (in which case short jumps are appropriate). Have you tried on 32-bit x86?

Does it work to avoid LOCAL3 in the main loop by checking from the end to the start with: R2 = number of bytes left to check (start at end of strings, decrement by `int` size on each iteration, stop when it hits zero), R0 = start of first string (add R2 to R0 to get each character), and R1 = start of second string (add R2 to R1 to get each character)?

mflatt:
On second thought, I'm not sure about moving the loop to common code. Maybe it's fine inline, especially if it can be made more compact.

gus-massa:
I compiled in 32 bits and there are no problems. Nevertheless, sometimes when I added more code the program breaks, so I guess it's close to the maximal size for tiny jumps.

I think it's necessary to use 5 registers: One for each character, one for a pointer to any place in the string, and one counter. I thought about keeping the pointers constant and counting backwards, but that doesn't change the number of registers. 

But this can be useful to unroll the loop for small strings, I didn't see a big difference in the time, but it really looks like a good idea so I'll try again.

gus-massa:
I did most of the changes, but I didn't move the main loop to jitcommon.c.

Some `jit_???` commands have a `(void)` cast. Is that necessary? Do all of them need one? Can I remove all of them?

I'm planning to add a special case for strings of small length that are known at jit time, because I think that `(string=? x "XYZ")` is a common use case.

mflatt:
Don't forget to get rid of the `mz_pop_locals()` calls.

Put a `(void)` cast only if the `jit_....` function would normally return an address to patch but you won't need to patch it. That `(void)` silences warnings on some compilers. I don't think you need `(void)` anywhere in this patch.

gus-massa:
To remove `mz_pop_locals()` I have to remove `mz_push_locals()` too, because otherwise I get an error.

mflatt:
I missed the `mz_push_locals()` – that shouldn't be there, either

gus-massa:
Another question. Is this

       __START_TINY_JUMPS__(1);
      jit_ldxi_l(JIT_R2, JIT_R0, &SCHEME_CHAR_STRLEN_VAL(0x0));
      jit_ldxi_l(JIT_V1, JIT_R1, &SCHEME_CHAR_STRLEN_VAL(0x0));
      __END_TINY_JUMPS__(1);
      __START_TINY_OR_SHORT_JUMPS__(!for_branch, branch_short);
      ref_false1 = jit_bner_i(jit_forward(), JIT_R2, JIT_V1);
      __END_TINY_OR_SHORT_JUMPS__(!for_branch, branch_short);

equivalent to this

      __START_TINY_OR_SHORT_JUMPS__(!for_branch, branch_short);
      jit_ldxi_l(JIT_R2, JIT_R0, &SCHEME_CHAR_STRLEN_VAL(0x0));
      jit_ldxi_l(JIT_V1, JIT_R1, &SCHEME_CHAR_STRLEN_VAL(0x0));
      ref_false1 = jit_bner_i(jit_forward(), JIT_R2, JIT_V1);
      __END_TINY_OR_SHORT_JUMPS__(!for_branch, branch_short);

?

I.E. Do `jit_ldxi_l` have some kind of internal hidden jumps? (This would be usefull to simplify some parts of my code.)

mflatt:
Those are equivalennt. Short- and tiny-jump modes affect only `jit_br....()` and `jit_jmp()`.

gus-massa:
I added in [another branch] ( https://github.com/gus-massa/racket/commits/15-7-X-JIT-Inline-String-EQ-Z ) an especial case for explicit strings of length 0, 1 and 2, but the difference in speed is very small and the run time difference is at the noise level. I think that it's better to keep only the special case for `""`, because it reduce the size of the jit code without increasing too much the complexity of jitinline.c

Also, to move the main loop to jitcommon.c it's necesary to have another register for the jump address (usually `JIT_V1`), so it looks difficult.

gus-massa:
I added `bytes=?` and changed a few unrelated `SCHEME_CHAR_STR_VAL` to `SCHEME_BYTE_STR_VAL`.

I hope this is the final version. (Please review extra carefully.)

Anyway, another question. For the bytes version, I didn't find:

    ref_false2 = jit_bner_c(jit_forward(), JIT_R0, JIT_V1);

so I replace it with:

    jit_extr_c_i(JIT_R0, JIT_R0);
    jit_extr_c_i(JIT_V1, JIT_V1);
    ref_false2 = jit_bner_i(jit_forward(), JIT_R0, JIT_V1);

mflatt:
It turns out that `jit_ldr_c()` sign-extends to an integer, so the `jit_extr_c_i()`s shouldn't be needed. Otherwise, this looks good to me.

Since `string=?` and `bytes=?` have so much in common, I would have made them the same, parameterizing the code generation over `scheme_char_string_type` and `&SCHEME_CHAR_STRLEN_VAL(0x0)` and using conditionals as necessary otherwise. I'm happy to defer to your choice of separate blocks if you prefer that approach, though.

gus-massa:
I tried to parameterize it, but half of the code is common and half of the code needs parameterizations, so it needs an `if (string) {` every other line. 

I'll try with something like 

    intptr_t strlen_val;
    [...]
    if (string) 
      strlen_val = &SCHEME_CHAR_STRLEN_VAL(0x0);
    else
      strlen_val = &SCHEME_BYTE_STRLEN_VAL(0x0);

gus-massa:
Sorry for the delay. While collapsing `string=?` and `bytes=?` to a common implementation I found a small bug. (It said something like `app->rand1` instead of `app->rand2`.) The problem is that fixing the bug caused an unexpected error in code like `(string=? x "")`, but it's not easy to trigger that because of the inlining and jit-inlining heuristics. (see 2nd patch).

It's easily solved removing the special case for explicit strings (see 3rd patch). It would be not bad to remove both special cases. I think they don't improve the speed or code size too much. The problem is that I really don't understand where the bug is, so I suspect the real bug is elsewhere and the deletion only makes the bug more difficult to find.

mflatt:
I wonder if the problem is that you're passing 0 (i.e., false) for the `order_matters` argument to `scheme_generate_two_args()`, but then assuming a particular assignment of arguments to R0 and R1. Does passing true for that argument help?

gus-massa:
Obviously! :) That fixed the problem. (5 seconds to write, 1 hour to rebase, retest, ...)

mflatt:
Looks good to me!

gus-massa:
Thank for all the reviews! This is finally merged.

PR #1008, opened 2015-08-12
Fix an error in racket-test-core/number.rktl on OpenBSD-current/sparc64

juanfra684:
"Errors were:
((numbers) (0.0 -0.0 (#<procedure:round> -0.0)))
((numbers) (125.0+0.0i 125.0-0.0i (#<procedure:z-round> 125.00000000000023-0.0i)))
((numbers) (100.0+0.0i 100.0-0.0i (#<procedure:z-round> 99.99999999999999-0.0i)))"

I tried every applicable option to OpenBSD/sparc64 in "Inexact Arithmetic", and this is the only one which fixed the error.


mflatt:
merged - thanks!

PR #1010, opened 2015-08-13
Use scheme_getenv in eval.c

gus-massa:
Replace `getenv` with `scheme_getenv` to read some environment variables. IIRC this is different only when the variable has an non ascii character like `é` or `ñ`, that is not an expected usecase of these variables. Anyway I think that it's better to make this change for consistency.


mflatt:
I agree

PR #1012, opened 2015-08-14
Fix string-replace when the string is mutable

gus-massa:
The `from` string argument is converted to a regexp and cached. When `from` is
a mutable string this can cause wrong results in the following calls
to `string-replace`. So the string is first converted to an immutable string to
be used as the key for the cache.

The cache has two steps:
A mutable string weakly holds a immutable copy until it is collected
or modified (and used as a argument of `string-replace`).
The immutable copy weakly holds the regexp that is used in string-replace.
Using `string->immutable-string` directly in `string-replace` is not a useful
because the immutable copy could be immediately collected.


mflatt:
LGTM

PR #1014, opened 2015-08-15
write-xexpr: consistently use argument port.

geoffhill:
Fix the behavior of the `#:insert-newlines?` keyword argument, to write newlines to the argument port, rather than `(current-output-port)`.


jeapostrophe:
Great! Thanks!

PR #1019, opened 2015-08-19
Fixes to prop:rename-transformer

takikawa:
Fixes an error message and removes a line that accommodates a box argument in the property. As far as I can tell, it was never possible to put boxes in the `prop:rename-transformer` property so I think this line can go.


mflatt:
merged

PR #1020, opened 2015-08-20
Add property to cooperate with syntax-local-value

takikawa:
This PR adds a struct property that cooperates with `syntax-local-value`.

The motivation for this change is to allow TR to close some bugs related to its uses of rename transformers: https://github.com/takikawa/typed-racket/tree/gh-issue-163

Some design questions:
  * Should `syntax-local-value/immediate` on a struct with this property return something other than the final redirected value? Currently I chose no, since existing clients of this function don't expect any special behavior from anything but rename transformers. Changing that behavior could break code.
  * What should be allowed in the property? I chose to allow either an integer index or a procedure with one argument. The index case is similar to rename transformers but the field must be an identifier. This seems reasonable to me, but there are other choices (like allow an identifier in the property too).
  * Should rename transformers or this take precedence when both properties are provided? I chose to let rename transformers take precedence, which I think is reasonable.

I did run into a situation in TR where I wanted `s-l-v/immediate` to give me the struct with this new property back, but I was able to work around it with another layer of indirection using a rename transformer.

Also the name of the property is tentative. If anyone has a better name that'd be great. Maybe `prop:redirect-syntax-local-value`?


mflatt:
@asumu: The issue with  `syntax-local-value/immediate` makes me reconsider the value of separating rename transformers from `prop:syntax-local-value`. If `prop:rename-transformer` allowed a single-argument function as its value, then the meaning of "rename transformer" would change so that it's target identifier is not a constant, but existing code is likely to work fine. With the separation, I can't help thinking that we'll need a variant of `syntax-local-value/immediate` eventually. And, looking at your message again, I see that you already wanted it once.

If you think the separation is better, then this commit is fine with me. If you'd like to try changing `prop:rename-transformer` after all, though, I think that would be a worthwhile experiment (despite my earlier advice).

takikawa:
@mflatt I've implemented your new suggestion and pushed it to this PR. I like this approach better because it adds fewer primitives. It works fine for all of my Racket test cases, but it doesn't quite solve all of my TR issues. Here's a unit test that I added for TR:

````racket
#lang racket

(module m typed/racket/base
  (define x 1)
  (provide x))

(module n typed/racket/base
  (require (submod ".." m))
  (module* a #f x))
````

The use of `x` in the `module*` at the bottom hits the "single form in the module body" special case and is therefore expanded in the `'module-begin` context. This is before TR's `#%module-begin` has installed its typed context flag, so it expands to the wrong thing.

In my previous `prop:syntax-local-value` + `prop:set!-transformer` solution, I had the set! transformer detect when it's in a module-begin context and apply a `#%expression` to delay expansion. You can't do that with rename transformers though, so it seems I'm stuck even with the delayed target identifier selection. Any idea how to work around this?

mflatt:
Maybe the `module` expander shouldn't try to expand a single identifier into `#%module-begin` if it's bound to a rename transformer, since the result can't possibly be a `#%module-begin` form.

samth:
Won't that have the same problem if we replace x with (x)?

mflatt:
Good point. Another possibility is that a rename transformer could have some sort of property to indicate the kinds of contexts where it can expand (to play a similar role as a `(syntax-local-context)` test in a recgular macro).

takikawa:
How about instead of putting the expanded syntax under a `#%module-begin`, re-expanding it by putting the unexpanded original form into the `#%module-begin`? (assuming it didn't expand to a `(#%module-begin ...)` form)

I suppose this has potential backwards incompatibilities for macros that have side effects.

mflatt:
Yes, as a general rule, it's bad to re-expand forms, so reverting to the unexpanded form to put in `#%module-begin` doesn't sound like a good idea to me.

samth:
I still like the opportunity to opt-out of module-begin expansion, which
avoids a lot of semantic complexity.

rmculpepper:
Perhaps there should be a separate type (and struct type prop) for module-begin transformers, similar to `set!`-transformers. I don't know if it's possible to do that in a backwards-compatible way, though. If the macro expander is already changing in 7.0 anyway, though...?

mflatt:
A general property to opt out of `#%module-begin` sounds like a better approach than what I had in mind, which was something specific to rename transformers.
Having to opt-in for `#%module-begin` expansion seems like an especially breaking change. (All my tutorials on Racket languages would break, I think.)

takikawa:
So something like `prop:not-#%module-begin`? And anything that's a `prop:procedure`, `prop:set!-transformer`, or a `prop:rename-transformer` with this property set to `#t` would not get expanded when directly in a `'module-begin` context? (presumably if the macro that's in module-begin position does a local expansion or something it will still expand normally?)

mflatt:
How about `prop:expand-contexts` to hold a list of symbols like the second argument to `local-expand` (but with a generic `'definition-context` symbol)? I could add that.

mflatt:
Added `prop:expansion-contexts` as 929db29b67e80ac155868fca5e8823235a745b4e.

takikawa:
Combining this with `prop:expansion-contexts` solved TR's issues nicely, thanks. So I'll go ahead and merge a squashed version of this unless anyone has any objections.

takikawa:
Closed by 33bb5e906086e4e0c0951d840916f8fa4bef602c.

PR #1030, opened 2015-09-02
http-client: don't try to gunzip an empty stream

jbclements:
Per (brief) discussion on mailing list, this change adjusts http-recv! to not try to gunzip a stream if it's empty; just return it as the empty stream.

### DIFF
--- a/racket/collects/net/http-client.rkt
+++ b/racket/collects/net/http-client.rkt
@@ -261,7 +261,9 @@
   (define decoded-response-port
     (cond
       [head? raw-response-port]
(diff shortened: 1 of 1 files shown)


jbclements:
@jeapostrophe , can you take a look at this PR?

jeapostrophe:
My pedantic side says we shouldn't do this, because it is not really gzip'd (which requires a header & footer), so there is an error, but my pragmatic side can deal with it. Regarding Greg's comment, I'm more comfortable taking the version that checks the code, but perhaps that's just best left as a comment with a link to the email thread.

PR #1031, opened 2015-09-03
net: make #<url> structure transparent

jbclements:
As I mentioned in an email thread on racket-users, it was a great help to me to make this structure transparent; it seems quite "Racket-y" to do so, and I don't see any drawbacks (since the constructor is already exposed, and url->string and string->url both exist).

@jeapostrophe , I think this is again your part of the world...

### DIFF
--- a/racket/collects/net/http-client.rkt
+++ b/racket/collects/net/http-client.rkt
@@ -261,7 +261,9 @@
   (define decoded-response-port
     (cond
       [head? raw-response-port]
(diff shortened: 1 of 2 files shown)


jbclements:
Actually, one request I might reasonably make of myself is to go ahead and make path/param transparent as well, and I'd be happy to fulfill this request if asked.

jeapostrophe:
Yes, you should do that too!

PR #1034, opened 2015-09-04
Remove ignored call/cc

gus-massa:
Reduce `(call/cc (lambda (<ignored>) body ...))` to `(begin body ...)`.

I'm not sure about `call/ec`. I'm woried that `call/ec` may add some kind of prompt that may be visible.

This can be usefull for code like

    (let/ec exit
      (for ([i (in-range 1000)])
        (do-something-without-exit i)))

I have to increment `vclock` because it was necesary to avoid the move of some expressions like

    (let ([z (random)])
      (call/cc (lambda (ignored) z)))

to

    (let ([unused-z (random)])
      (random))

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -3540,6 +3540,16 @@
            #f)
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Check that the unused continuations are removed
+
+(test-comp '(call-with-current-continuation (lambda (ignored) 5))
+           5)
+(test-comp '(call-with-composable-continuation (lambda (ignored) 5))
+           5)
+(test-comp '(call-with-escape-continuation (lambda (ignored) 5))
+           5)
+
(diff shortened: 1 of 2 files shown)


samth:
If you're thinking about optimizing `call/cc`, @gus-massa, I can mention my favorite desired optimization: transforming

```
(call/cc (lambda (k) (if e1 e2 (k e3))))
```

into

```
(if e1 e2 e3)
```

And more generally when `k` is only used as a operator in tail position wrt the `lambda`.

gus-massa:
The "ignored" case is much easier, because all the information is already available. Each argument has a "use count", but it's not very reliable during the optimization pass, because it's not updated when the code is eliminated or duplicated.

0 means unused. 1 means probably at most one, and this was the source of a recent bug. More than 1 actually means undefined but probably many times. (I think that the number is correct initially, but some parts are optimized multiple times.)

With a reliable count it would be easier, because we can count all the tail calls and compare the number with the total use count. The alternative is to inspect all the expression if it's small enought, using a fuel to bound the exploration deep. For example, currently there is a similar function that checks that the expression uses no mutable bindings, to detect that it's safe to move.

mflatt:
This seems right as far as I can tell (but with a `printf` that I expect you already plan to remove).

gus-massa:
I'll merge the simple version now. I'll try to write the more complete version later.

PR #1035, opened 2015-09-04
Remove nested begin0's added by sfs pass

gus-massa:
An example is

    (let ([z (random)])
      (begin0 (lambda () z) (display 0)))

but I don't know how to test the lack of repetition of the `begin0`.


mflatt:
Looks good to me. I don't have a good idea for the test, either, but probably the example should go in "optimize.rktl" to ensure that the flattening code is run.

gus-massa:
I think that the only way is to inspect the bytecode. Is this test ok?

mflatt:
Yes, that looks like a good test.

PR #1037, opened 2015-09-04
added args checking to prefix functions

florence:
Adds argument checking for `take-common-prefix` and friends.


samth:
Looks good.

stamourv:
Merged, thanks!

PR #1039, opened 2015-09-07
Fix make-contract example to show the correct contract violation

btlachance:
There was a duplicate quote around `given:' which caused a violation in
make-contract's internals.


rfindler:
Thanks! I've pushed it.

PR #1040, opened 2015-09-07
Allow user to set there personal github api token

LeifAndersen:
Using the PLT_GITHUB_API_KEY environment variable.

### DIFF
--- a/racket/collects/pkg/private/stage.rkt
+++ b/racket/collects/pkg/private/stage.rkt
@@ -722,7 +722,7 @@
      (or
       (for/or ([kind '("branches" "tags")])
         (define api-u
-          (url "https" #f "api.github.com" #f #t
+          (url "https" (getenv "PLT_GITHUB_API_KEY") "api.github.com" #f #t
                (map (λ (x) (path/param x empty))
                     (list "repos" user repo kind))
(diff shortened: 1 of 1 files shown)


samth:
_their_!

mflatt:
@LeifAndersen Are you running into GitHub query limits? I think I ran into that problem a week or two ago, and I have been thinking that we should switch `raco pkg` to always use the Git protocol, instead of using GitHub API queries.

samth:
I've run into the API limits, such as with `raco pkg update -a` run repeatedly.

LeifAndersen:
@samth Doah.

@mflatt Yes, I run into them quite regularly when updating.

stamourv:
Ditto. The limits make `raco pkg catalog-archive` unreliable, too.

mflatt:
@LeifAndersen I changed `raco pkg` to use the Git protocol for GitHub. Are you still interested in adding `PLT_GITHUB_API_KEY`?

LeifAndersen:
Cool, thanks.

Nah, I'm not really interested then. I only cared because we were hitting the API rate limit, and if that's not a problem any more then I don't care.

Thanks for pushing this.

PR #1044, opened 2015-09-10
Propagate object inspectors in class/c wrappers

takikawa:
This allows reflection and equality to work on objects produced from a contracted class if the original class had an appropriate inspector or was transparent.

This required adding a new field to the class struct, which I'm not too happy about but I don't see a way to avoid it. Also the one problem I've found so far is that you can still observe whether an object came from a contracted class or not via reflection using `object->vector` because it will have more `...`s in the output. The other reflective functions work as expected though.

See this mailing list thread for reference: https://groups.google.com/forum/#!topic/racket-users/8bUoIhd07f8

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/class.rkt
+++ b/pkgs/racket-test/tests/racket/contract/class.rkt
@@ -2566,4 +2566,53 @@
        (class object%
          (super-new)
          (define/public (callback f) (f 1))))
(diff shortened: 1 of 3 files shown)


mflatt:
LGTM

takikawa:
Merged as 52aa11c4074041b2cfb4628b7c960c79419e3104

PR #1045, opened 2015-09-10
Format string and arguments match.

acarrico:


acarrico:
Just looking at Greg Hendershott's guide fo infrequent contributers to Racket: sorry this isn't in a branch.

samth:
@acarrico This is totally fine -- how you organize the `acarrico/racket` fork doesn't affect how it looks when reviewing.

samth:
Is there a test case for this, though? It looks like one ought to be pretty easy to construct.

acarrico:
Sorry: test case: (sort '(1 2 3))

Let me know if you add a test case to build, since I don't really know where that would go.

acarrico:
Actually, to test it you need a special environment. I got from in racket-mode.

samth:
Test cases would go here: https://github.com/plt/racket/blob/master/pkgs/racket-test-core/tests/racket/function.rktl

But running that expression gives me what I expect:

```
-> (sort '(1 2 3))
; sort: arity mismatch;
;  the expected number of arguments does not match the given number
;   expected: 2 plus optional arguments with keywords #:cache-keys? and #:key
;   given: 1
```

Is there something else I need to do to see the error?

acarrico:
Try in the racket-mode *racket REPL*. I guess there needs to be a missing (syntax-line stx), so it is context sensitive.

acarrico:
t﻿> (sort '(1 2 3))
; format: format string requires 2 arguments, given 3; arguments were: "~a:::~a: " 'stdin #f 6965

So a proper test case would presumably need to synthesize syntax without line numbers for evaluation.

samth:
Here's a test program that errors currently.

```
(let ()
 (define l (make-log-receiver (current-logger) 'warning))
 (eval (datum->syntax #'here '(lambda () (sort 1)) (list 'a #f #f #f #f))))
```

You can make that into a test case for that file as follows:

```
(let ()
  (define l (make-log-receiver (current-logger) 'warning))
  (test procedure? (eval (datum->syntax #'here '(lambda () (sort 1)) (list 'a #f #f #f #f))) #t))
```

acarrico:
You forgot the "expect" argument (below), but otherwise that works (fails before, doesn't fail after the change). I don't know the racket dev test culture, but wouldn't commiting this in "function.rktl" be a confusing legacy? It is a contrived test for a random bug, whereas all the other tests are readable tests of function operators. 

(let ()
  ;; ISSUE: format: format string requires 2 arguments, given 3; arguments were: "~a:::~a: " 'a #f #f
  (define l (make-log-receiver (current-logger) 'warning))
  (test #t
        procedure?
        (eval (datum->syntax #'here '(lambda () (sort '(1))) (list 'a #f #f #f #f)))))

rfindler:
On Thu, Sep 10, 2015 at 9:32 PM, Anthony Carrico <notifications@github.com>
wrote:

> You forgot the "expect" argument (below), but otherwise that works (fails
> before, doesn't fail after the change). I don't know the racket dev test
> culture, but wouldn't commiting this in "function.rktl" be a confusing
> legacy? It is a contrived test for a random bug, whereas all the other
> tests are readable tests of function operators.
>
>
Many of the tests I use are there to check correct behavior of functions
(like, iiuc, this one) that were put in place because of some strange bug
somewhere. My experience is that adding those tests is a win.

acarrico:
Hi Robbie, the issue here is that you get a dynamic error (a format someplace in the implementation has wrong args). Sam's test works for this case, but doesn't really test anything about functions per se. I guess my question is, why test this format string, but none of the others in the code base, and why test it here in this very indirect way (my EE training "test at the pin")? Nobody could look at the test case and say, "oh, that is testing for a bad format string". We now presume this one is correct, and unlikely to regress. I suppose, the proper solution would be for "somebody :)" to switch the code base to a format macro that checks args statically (for non higher order use of format), and then to use test cases to ensure the static checker really catches bad format strings. That would solve the problem across the code base.

rfindler:
That all makes sense and I understand. If there is ​a better place to put a
test case, then by all means, please put it there! If there are other test
cases that come to mind as worth testing (or perhaps an entire class of
problems to test for that you could automate by using random testing) then
make more tests.

But putting this one test in is a good idea because it makes sure in the
most concrete way that this one bug is no longer in the codebase. It is a
minimum bar to pushing a bug fix. One changed piece of behavior that makes
sure that this one specific bug is for sure gone.

Does this make more sense?

mflatt:
I agree about adding a test case. I'd put it in "syntax.rktl", since the intent is to test error formatting from the keyword-argument `lambda` layer.

acarrico:
Seems like a better place. I've put it there and committed.

samth:
Merged now, thanks!

PR #1050, opened 2015-09-11
syntax/parse: fix disappeared-use property in attribute macro

AlexKnauth:
This fixes the DrRacket check syntax arrows in uses of `attribute` like this:
```racket
#lang racket/base
(require syntax/parse)
(syntax-parse #'a
  [a
   (attribute a)])
```


AlexKnauth:
This was merged as https://github.com/plt/racket/commit/2923a3adcfed4a10fc756ee8b8e208d266ae191b

PR #1055, opened 2015-09-13
Binding for OpenSSL SHA-256

eu90h:
Figured it would be nice if people could easily use SHA-256, so I wrapped openssl's implementation.
The first commit adds the wrapper itself, while the second consists of the documentation.

### DIFF
--- a/pkgs/racket-doc/openssl/openssl.scrbl
+++ b/pkgs/racket-doc/openssl/openssl.scrbl
@@ -4,7 +4,8 @@
           (for-label openssl
                      racket
                      openssl/sha1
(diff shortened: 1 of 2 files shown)


samth:
Any particular reason you closed this, @eu90h?

eu90h:
I fudged my racket fork by accidentally pushing to the master branch, so as
I pushed to OS rng stuff it was "on top" of the sha-256 binding commits.
I'll reopen later tonight/tomorrow.

eu90h:
additionally, it probably wouldn't hurt to wrap the whole sha-2 family
while I'm in there

PR #1057, opened 2015-09-13
typo: construct -> contract

bennn:


rfindler:
Thanks! Pushed: http://git.racket-lang.org/plt/aa0823daf6..dedde2cd60

PR #1060, opened 2015-09-14
make sure git checkouts have correct file perms

florence:
This fixes and issue where package installs sometimes have the wrong permissions a file.

I don't quite know how to tests this. I did check locally on the package that was having issues (https://github.com/rpless/cover-coveralls)

### DIFF
--- a/racket/collects/net/git-checkout.rkt
+++ b/racket/collects/net/git-checkout.rkt
@@ -676,6 +676,7 @@
             #"100755" #"755")
            (copy-object tmp
                         (this-object-location)
+                        mode
                         (build-path dest-dir fn))]
           [(#"40000" #"040000")
            (extract-tree id obj-ids tmp (build-path dest-dir fn))]
@@ -938,8 +939,8 @@
    [else
     (call-with-input-file* (build-path (tmp-info-dir tmp) location) proc)]))
 
-;; copy-object : tmp-info location path -> void
-(define (copy-object tmp location dest-file)
+;; copy-object : tmp-info location bytes path -> void
+(define (copy-object tmp location mode dest-file)
   (cond
    [(pair? location)
     (define bstr (object->bytes tmp location))
@@ -950,7 +951,12 @@
    [else
     (copy-file (build-path (tmp-info-dir tmp) location)
                dest-file
-               #t)]))
+               #t)])
+  (define perms
+    (case mode
+      [(#"755" #"100755") #o755]
+      [(#"644" #"100644") #o644]))
+  (file-or-directory-permissions dest-file perms))
 
 ;; object->bytes : tmp-info location -> bytes
 (define (object->bytes tmp location)
--- a/racket/collects/net/git-checkout.rkt
+++ b/racket/collects/net/git-checkout.rkt
@@ -671,13 +671,16 @@
        (when id
          (define (this-object-location)
            (object-location (hash-ref obj-ids id)))
-         (case (datum-intern-literal mode)
-          [(#"100644" #"644"
-            #"100755" #"755")
+         (define (copy-this-object perms)
            (copy-object tmp
                         (this-object-location)
-                        mode
-                        (build-path dest-dir fn))]
+                        perms
+                        (build-path dest-dir fn)))
+         (case (datum-intern-literal mode)
+          [(#"100755") #"755"
+           (copy-this-object #o755)]
+          [(#"100644" #"644")
+           (copy-this-object #o644)]
           [(#"40000" #"040000")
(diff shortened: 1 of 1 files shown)


rfindler:
Would you test this by having a canned git repo somewhere and then asking
this to do a checkout?

Is there already a test suite for this code?

florence:
A grep of the main repo shows no usages of `git-checkout` in a test directory (or in fact out side of the pkg collection), so I don't think there is a test suite yet.

mflatt:
My "test" originally was to clone the Racket repo and diff the result. I think that created the illusion of a test, so that I forgot to automate it.

florence:
I've just done the same "test" as Matthew, and everything checks out. I don't think I'll have time to actually automate that test before racket con, and `cover-coveralls` is not working without this. Is it okay to merge it anyways?

mflatt:
That's fine with me.

rfindler:
Looks like something goes wrong on windows. From the AppVeyor build (above)

```
Downloading repository git://github.com/racket/data/?path=data-lib
file-or-directory-permissions: update failed due to unsupported bit combination
  path: C:\Users\appveyor\AppData\Local\Temp\1\14422706421442270642484-racket_data_master\.gitignore
  permission value: 420 
  context...:
   C:\projects\racket\racket\collects\net\git-checkout.rkt:669:5: loop 
   C:\projects\racket\racket\collects\net\git-checkout.rkt:141:11 
   C:\projects\racket\racket\collects\net\git-checkout.rkt:38:2: retry-loop 
   C:\projects\racket\racket\collects\pkg\private\download.rkt:99:2: download! 
   C:\projects\racket\racket\collects\file\cache.rkt:63:2: fetch-and-continue 
   C:\projects\racket\racket\collects\racket\contract\private\arrow-val-first.rkt:324:3 
   C:\projects\racket\racket\collects\pkg\private\download.rkt:91:0: download-repo!24 
   C:\projects\racket\racket\collects\pkg\private\stage.rkt:299:11 
   C:\projects\racket\racket\collects\pkg\private\stage.rkt:106:0: stage-package/info46 
   C:\projects\racket\racket\collects\pkg\private\stage.rkt:106:0: stage-package/info46 
   C:\projects\racket\racket\collects\pkg\private\install.rkt:641:4: for-loop 
   C:\projects\racket\racket\collects\pkg\private\install.rkt:141:0: install-packages74 
   C:\projects\racket\racket\collects\pkg\private\..:261:28 
   C:\projects\racket\racket\collects\racket\contract\private\arrow-val-first.rkt:324:3 
   C:\projects\racket\racket\collects\racket\file.rkt:374:8 
   C:\projects\racket\racket\collects\racket\file.rkt:363:0: call-with-file-lock42 
   ... 
NMAKE : fatal error U1077: 'racket\racket.EXE' : return code '0x1'
Stop.
```

rfindler:
I pushed the commits, but then reverted them when I realized the issue above. Sorry about the churn.

florence:
As it turns out, `644` is not a valid set of permissions on windows. My google fu tells me that `git` handles this by converting every file to `755` on windows, so I changed things to do that. I have no local windows installs to I'm letting the automatic appveyor build do that.

florence:
Hrm, doesn't look like the appveyor build here does any kind of `raco pkg install`. Does anyone have a windows machine I can test this on?

samth:
It installs a number of packages for testing -- that's why it failed.

florence:
Oh I see, around line 2201 of the output. Missed that. I guess that means its fixed, now that the appveyor build is passing?

samth:
Yeah, looks good.

rfindler:
Okay if I squash the commits?

rfindler:
I've pushed https://github.com/racket/racket/commit/35a0f0c71c6a53a0cb09520ef9de335cb06e66d4

florence:
This seems to have caused a failure on a different appveyor build: https://ci.appveyor.com/project/plt/racket/build/1.0.785

@mflatt Am I missunderstanding the code in file_or_dir_permissions? I thought `755` was valid because the read and write were set for every level of permissions?

mflatt:
@florence write needs to be the same across all levels, and read and exec always need to be set; so, I think that means the valid permissions are 777 and 555.

mflatt:
Maybe the right strategy is just to not try to set permissions on Windows

samth:
Do we know what git itself does?

rfindler:
Spencer suggests reverting this commit, but I think a better intermediate
state is one that just doesn't make these changes on windows. Is that okay
with everyone?

samth:
Seems sensible to me.

PR #1061, opened 2015-09-15
Minor fixes found by cppcheck

juanfra684:


vendethiel:
minor?

samth:
I think both of those are in never-on config options.

juanfra684:
Yeah, probably that is the reason why nobody complained before :) .

@samth, if some of the devs have some free time, try to check the code with cppcheck. It shows a lot of warnings but I don't understand the GC code.

mflatt:
merged - thanks!

Issue #1062, opened 2015-09-15
Standard access to kernel-level PRNGs


eu90h:
Many languages have a standard way to access random numbers generated from the operating system's  entropy pool. 

They seem to be invariably implemented in the same way: reading bytes from `/dev/urandom` on *nix based systems or calling Advapi32.dll's CryptGenRandom/RtlGenRandom on Windows systems. 
See for instance [Python](https://docs.python.org/2/library/os.html#os.urandom) or [Go](https://golang.org/src/crypto/rand/rand.go). 

The Racket [system](http://docs.racket-lang.org/reference/os.html?q=system) modules lack any such procedure for OS PRNG access. 

I have a patch which adds a procedure `(secure-random n)` which reads `n` random bytes from the OS PRNG. I would like to test it a little bit more (particularly on Windows), but it should be ready to go soon.

The proposed `secure-random` function has the added bonus of being cryptographically secure, unlike the already present `random` function. Interestingly, using the MRG32k3a PRNG (which is what Racket's `random` is) was the cause of Hacker News being [compromised](https://news.ycombinator.com/item?id=639976).

If you folks like the proposal, I'll post the patch here soon. 

Thanks to @lexi-lambda for suggesting the name `secure-random`

rfindler:
This sounds great!

Does it make sense to use the word `system` in the name of the function?
Like system-random or something like that to indicate that it's coming from
the underlying OS?

eu90h:
I believe that `secure-random` is best, as it makes the procedure more "self-documenting". 

If we went with `system-random` for instance, the programmer wonders "what's the difference between this and the normal `random`?". By naming it `secure-random`, the programmer need not wonder the difference. This point was convincingly argued by lexi-lambda in an IRC discussion.

I believe the documentation would be the best place to explain the source of randomness.

rfindler:
Okay.

jeapostrophe:
I feel pretty strongly that "secure-random" is a misleading name, because
it is very opinionated and "blesses" the function too much. There's no
reason that the OS is particularly secure and there are many instances of
mistakes in this part of OSes. I think a name like "system-random" is
perfectly apt: it is descriptive and you can use the documentation to
explain why you'd like to use it over others.

Jay

On Tue, Sep 15, 2015 at 9:12 PM, Robby Findler <notifications@github.com>
wrote:

> Okay.
>
> —
> Reply to this email directly or view it on GitHub
[...]

PR #1064, opened 2015-09-16
added access to os' rng

eu90h:
patch fulfilling issue #1062

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/numbers.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/numbers.scrbl
@@ -1,10 +1,11 @@
 #lang scribble/doc
-@(require "mz.rkt" racket/math scribble/extract
+@(require "mz.rkt" racket/math scribble/extract 
           (for-label racket/math
                      racket/flonum
                      racket/fixnum
                      racket/unsafe/ops
-                     racket/require))
+                     racket/require
+                     racket/random))
 
 @(define math-eval (make-base-eval))
 @(interaction-eval #:eval math-eval (require racket/math))
@@ -835,6 +836,8 @@ both in binary and as integers.
 @; ------------------------------------------------------------------------
 @subsection{Random Numbers}
 
+@margin-note{If security is a concern, use @racket[crypto-random-bytes]}
+
 @defproc*[([(random [k (integer-in 1 4294967087)]
                     [rand-gen pseudo-random-generator?
                                (current-pseudo-random-generator)])
@@ -854,7 +857,6 @@ generator (which defaults to the current one, as produced by
 internal state for generating numbers. The random number generator
 uses a 54-bit version of L'Ecuyer's MRG32k3a algorithm
 @cite["L'Ecuyer02"].}
-
 
 @defproc[(random-seed [k (integer-in 1 (sub1 (expt 2 31)))])
           void?]{
@@ -922,6 +924,18 @@ where the first three integers are in the range @racket[0] to
 range @racket[0] to @racket[4294944442], inclusive; at least one of
 the first three integers is non-zero; and at least one of the last
 three integers is non-zero. Otherwise, the result is @racket[#f].}
+
+@subsection{System-provided Randomness}
+
+The @racketmodname[racket/random] module provides an interface to randomness from the underlying operating system.
+Use this over @racket[random] wherever security is a concern.
+
+@defproc[(crypto-random-bytes [n exact-positive-integer?])
+         bytes?]{
+Returns n bytes. On Unix systems, the bytes are obtained from @filepath{/dev/urandom}, while Windows uses RtlGenRand.}
+
+@examples[#:eval(make-base-eval #:lang 'racket '(define (crypto-random-bytes n) #"\0\1\1\2\3\5\b\r\25\"7Y\220\351"))
+ (crypto-random-bytes 14)]
 
 @; ------------------------------------------------------------------------
 @subsection{Number--String Conversions}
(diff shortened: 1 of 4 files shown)


eu90h:
Ugh.. not sure why it's failing - give me a while.

mflatt:
As others have said in #1062, this good addition is a good idea.

Others suggested `racket/system`, but I don't think that's the right place, because "system" in `racket/system` really refers to the `system` function. I suggest making it `racket/random`.

Various tweaks would make it easier to merge:

 * The internal modules "unix_rand.rkt" and "windows_rand.rkt" should be moved into a "private" subdirectory. Also, using a hyphen instead of an underscore in the name would be more consistent with our naming conventions.

 * My sense from the earlier discussion is that "Provides strong randomness from the underlying OS" in the docs is too strong, and it should just be "Provides randomness from the operating system".

 * Use "Unix" instead of "*NIX". I know that's less technically correct, but it's our convention for now.

 * Use `@filepath{...}` for path names.

 * Use `@racket[crypto-random-bytes]` in the note to make sure it typesets and hyperlinks right. But I think the note really belongs next to `random`, not next to `crypto-random-bytes` itself. Less importantly, please reword to avoid starting the sentence with a lowercase name.

Thanks

jackfirth:
Rather than "strong randomness", would "cryptographically secure randomness" be preferable? The function is meant to guarantee that the numbers it produces are from a cryptographically secure pseudorandom number generator provided by the OS. That's more technically accurate and conveys why you'd want to use this over `random`.

eu90h:
Thanks a lot Matthew, `racket/random` sounds good. I'll rewrite the docs & fix everything to align with convention (sorry, still learning the ropes!)

Additionally, I'd like to thank @jackfirth a bunch for helping me review this.

samth:
I agree with @jackfirth -- "strong" is unclear, but in cases where the OS doesn't provide genuine randomness or a well-seeded CSPRNG, we should error rather than returning something weaker.

eu90h:
Would a statement like, "provides randomness from the underlying OS intended for use in cryptographic applications" be ok?

This statement conveys both the source and the intent.

rfindler:
That sounds right to me.

eu90h:
OK. Moved the procedure to a new file `racket/random`. Added new documentation for it & margin note on `random` procedure.

rfindler:
Looks like `math`'s `random-bits`, `random-natural`, and `random-integer` should change to either call this function or to get a similar pointer/disclaimer. @ntoronto ?

Also, I noticed this slightly sub-optimal error message:

```
> (random-integer 1 1)
random-natural: contract violation
  expected: Positive-Integer
  given: 0
> 
```

soegaard:
2015-09-20 15:03 GMT+02:00 Robby Findler <notifications@github.com>:

> Looks like math's random-bits, random-natural, and random-integer should
> change to either call this function or to get a similar pointer/disclaimer.
> @ntoronto <https://github.com/ntoronto> ?
>

It would be a bad idea to change them - so a pointer is best.

The standard random functions generate pseudo-random numbers.
This implies that we can repeat experiments using random-seed.

    #lang racket
    (require math/base)
    (random-seed 42)
    (random-bits 10)
    (random-seed 42)
    (random-bits 10)

Output:  75 75

Note: The new random generator is slower than the standard random number
generator.
          (Not a criticism - cryptographically randomness cost more)

/Jens Axel




> Also, I noticed this slightly sub-optimal error message:
>
> > (random-integer 1 1)
> random-natural: contract violation
>   expected: Positive-Integer
>   given: 0
> >
>
> —
> <https://github.com/racket/racket/pull/1064#issuecomment-141784144>.
>



-- 
-- 
Jens Axel Søgaard

rfindler:
That makes a lot of sense.

Thanks,
Robby

eu90h:
I'd _totally_ forgotten the random functions exported by `math`, apologies. 
I'll add a margin note on those as well.

rfindler:
Some other comments:

 - please make a single commit

 - please remove changes that are only whitespace (don't change pkgs/racket-doc/scribblings/reference/subprocess.scrbl or racket/collects/racket/system.rkt at all)

- This is a better definition of `crypto-random-unix-bytes`:

```
(define (crypto-random-unix-bytes n)
  (check-urandom-exists)
  (call-with-input-file* "/dev/urandom"
    (λ (port)
      (read-bytes n port))))
```

(It won't leak ports.)

- The file `random.scrbl` isn't linked in anywhere. I guess the `racket/string` organization is the best one to follow. So you would create a new subsection of the "Random Numbers" section in `numbers.scrbl` and put the content of `random.scrbl` there. 

- Please include some examples in the docs. Use fake output, however; don't actually call the real functions during documentation building. (Maybe the resulting bytes could be `#"\0\1\1\2\3\5\b\r\25\"7Y\220\351"` ?)

rfindler:
Also: should racket/random be provided by racket? I'd say yes, but I'm not positive that's a good idea.

eu90h:
Sorry about all that :|

I'll fix it up, thank you.

rfindler:
On Sunday, September 20, 2015, Stefan <notifications@github.com> wrote:

> Sorry about all that :|
>
> I'll fix it up, thank you.
>
>
>

No worries! I am thankful for your contribution. Just want to make it as
good as it can be. :)

eu90h:
Sorry for the delay, been feeling a bit ill the past week.

Thanks @rfindler for fixing the port leaking and general tips. (Learned to squash commits - glad the git voodoo actually worked!)

I moved the documentation to a subsection below the random section, and included an example (had to do some whacky stuff to not actually eval the procedure, hopefully this was done ok.)

mflatt:
Merged (with some small changes)

PR #1065, opened 2015-09-16
added bytes->number

eu90h:
A procedure that turns byte-strings into numbers would be useful, especially when dealing with the proposed `crypto-random-bytes` function. This patch adds such a procedure.

### DIFF
--- a/racket/collects/racket/bytes.rkt
+++ b/racket/collects/racket/bytes.rkt
@@ -1,12 +1,13 @@
 #lang racket/base
 
-(provide bytes-append* bytes-join)
+(provide bytes-append* bytes-join bytes->number)
 
 (define bytes-append*
   (case-lambda [(strs) (apply bytes-append strs)] ; optimize common case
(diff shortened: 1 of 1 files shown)


mflatt:
I'd expect `bytes->number` to be like `string->number`, but I don't have a good name for this function. I also worry that it's not needed enough to include in `racket/bytes`.

jackfirth:
It might make more sense to do it in terms of `bytes->list` rather than `bytes->hex-string`:

```racket
(define (bytes->number bytes)
  (digits->number (bytes->list bytes) 256))

(define (digits->number digits base)
  (let loop ([n 0] [digits digits])
    (if (empty? digits)
        n
        (loop (+ (* n base)
                 (car digits))
              (cdr digits)))))
```

The rationale for this being different from `string->number` is that a byte string is already canonically one number - the result of treating it as a list of digits in base 256.

jeapostrophe:
FWIW, I have written this function a few times using Jack's version. I've also written it before using integer-bytes->integer and using base 2^32

PR #1066, opened 2015-09-17
wrapped openssl's sha-256

eu90h:
@samth here you go :)
 sorry about closing #1055, just couldn't think of another way.

### DIFF
--- a/pkgs/racket-doc/openssl/openssl.scrbl
+++ b/pkgs/racket-doc/openssl/openssl.scrbl
@@ -4,7 +4,8 @@
           (for-label openssl
                      racket
                      openssl/sha1
-                     openssl/md5))
+                     openssl/md5
+                     openssl/sha256))
 
 @title{OpenSSL: Secure Communication}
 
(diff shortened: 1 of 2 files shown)


eu90h:
AppVeyor failure: `Build execution time has reached the maximum allowed time for your plan (40 minutes).`

mflatt:
I'm not opposed to this addition, but I wonder whether it would be better to add it in a separate package (while `openssl/libcrypto` should be documented so that packages can reasonably depend on it)

eu90h:
Improved the wrapper slightly, making better use of the FFI libs (thanks @samth). 

Where would a better place for this be @mflatt? Or do you mean I should just make it a package and submit it to the index?

mflatt:
@eu90h Yes, I think this library makes sense as a new package. I've finally added documentation for `openssl/libcrypto`.

eu90h:
Understood. Thanks for for documenting libcrypto. I'll go ahead and close this.

PR #1067, opened 2015-09-17
Refactor raco's tool listing code

jackfirth:
This PR makes some stylistic and organizational changes to the implementation of `raco`'s listing of installed tools, with the aim of making the implementation easier to read.

I'd like to add tests, but I don't see any here and I'm not sure where to add them.


mflatt:
merged - thanks!

PR #1074, opened 2015-09-23
fixing git-checkout permissions for windows

florence:
Follow up to #1060, in which this finally works for Windows.


mflatt:
I don't know why this would be needed. Can you explain more @florence?

PR #1076, opened 2015-09-25
Try to collapse references in a branch using the type information of the other branch

gus-massa:
First commit: Create a new sub_info for each branch to hold the type information of the local variables, instead of handling the types manually. It's makes easier to write the second commit. 

Seccond commit: A reference to a local may be reduced in a branch to a constant, while it's unchanged in the
other because the optimizer has different type information for each branch. Try to use the
type information of the other branch to see if both branches are actually equivalent.

For example, `(if (null? x) x x)` is first reduced to `(if (null? x) null x)` using the type
information of the `#t` branch. But both branches are equivalent so they can be
reduced to `(begin (null? x) x)` and then to just `x`.

Third commit: In some cases, for example while using `no_types`, the optimizer can try to
add again the type information of a local variable. This creates unnecessary
internal storage to save the repeated information.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1762,6 +1762,12 @@
 (test-comp '(lambda (x) (if (values 1 2) 78 78))
            '(lambda (x) (values 1 2) 78)
            #f)
+(test-comp '(lambda (x) (if (null? x) x x))
+           '(lambda (x) x))
+(test-comp '(lambda (x) (if (null? x) null x))
+           '(lambda (x) x))
+(test-comp '(lambda (x) (not (if (null? x) #t x)))
+           '(lambda (x) (not x)))
 
 (test-comp '(lambda (x) (if (let ([r (something)])
                               (if r r (something-else)))
--- a/racket/src/racket/src/optimize.c
+++ b/racket/src/racket/src/optimize.c
@@ -23,7 +23,7 @@
   All rights reserved.
 */
 
-/* This file implements bytecode optimzation.
+/* This file implements bytecode optimization.
 
    See "eval.c" for an overview of compilation passes. */
 
@@ -79,8 +79,8 @@ struct Optimize_Info
 
   /* Set by expression optimization: */
   int single_result, preserves_marks; /* negative means "tentative", due to fixpoint in progress */
-  int escapes; /* flag to signal that the expression allways escapes. When escapes is 1, it's assumed
-                  that single_result and preserves_marks are also 1, and that it's not necesary to
+  int escapes; /* flag to signal that the expression always escapes. When escapes is 1, it's assumed
+                  that single_result and preserves_marks are also 1, and that it's not necessary to
                   use optimize_ignored before including the expression. */
 
   char **stat_dists; /* (pos, depth) => used? */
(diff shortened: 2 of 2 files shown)


mflatt:
This looks plausible, and I hope to take a closer look soon. Anote on timing the merge: I think we'll want to wait until after the branch is made for the next release (which is about a week from now).

gus-massa:
I agree that a shift error is very difficult to detect and it's better to wait until the branch for next version.

I think that https://github.com/racket/racket/pull/956 is safe to include in this version.

mflatt:
I finally looked at this a little more, and based on just skimming the code, it seems ok.

gus-massa:
I rebased this and the test passed. I think this still needs a thoughtfully review, because I'm worry that there are any hidden problems with the coordinate shifting of the locals.

mflatt:
I tried applying the patch in a fresh build (just `make` at the repo top level), and I got a seg fault. Here's the most interesting fragment of the crash report, and I can send the whole report if it seems helpful – but the crash happened both times I tried, so I expect it's repeatable:

```
Thread 9 Crashed:
0   Racket                        	0x0000000100c38936 add_type + 166 (optimize.c:8863)
1   Racket                        	0x0000000100c2918b merge_types + 251 (optimize.c:4224)
2   Racket                        	0x0000000100c26ec4 optimize_for_inline + 5716 (optimize.c:1946)
3   Racket                        	0x0000000100c2a20e optimize_application2 + 1054 (optimize.c:3213)
4   Racket                        	0x0000000100c2f9ef scheme_optimize_expr + 2287 (optimize.c:7972)
5   Racket                        	0x0000000100c35a97 scheme_optimize_expr + 27031 (optimize.c:4491)
6   Racket                        	0x0000000100c2e5fd scheme_optimize_lets + 11677 (optimize.c:6690)
7   Racket                        	0x0000000100c311a3 scheme_optimize_expr + 8355 (optimize.c:7988)
8   Racket                        	0x0000000100c35b9d scheme_optimize_expr + 27293 (optimize.c:4505)
```

gus-massa:
I was testing with
 
    nmake win32-in-place PKGS="racket-test-core"

and a few additional packages like `compiler-lib` to minimize the compilation time. With that I got no error. Neither in Travis or AppVeyor.

I tried again with all the packages of  `main-distribution` and I got an error :). In all runs, the last packages before the error was: 

    raco setup: 1 making: <pkgs>/compatibility-doc/mzlib/scribblings

So I'll try to minimize it to isolate the error.

gus-massa:
I still didn't fix this, but I got a minimized test case to get the error:

    #lang racket/base
    
    (define {{zz x} y} 0)
    
    (define {ff x}
     {{zz (letrec ([loop (lambda () loop)]) loop)} (car x)})

I tried to inline `zz` manualy but I coudn't.

gus-massa:
I got an advance, but this is still not the final version.

I think I found the error. One of the `merge_types` was shifting in the wrong direction. I want to take one day to review carefully all the other `merge_types` to check if they have the right signs.

Also, after reading many times the commit "Don't add type information twice" I think it must be rewriten, so I'll change it.

gus-massa:
I submitted the fix as another PR in https://github.com/racket/racket/pull/1197

I think it would be better to merge now only the fix, and wait until the release branch of 6.3.1 is created to analize these new features.

gus-massa:
Let's try again ...

I rebased the old version. The first two commits are unchanged, I'm happy with them.

I rebrote the `add_type` fucntion of the third commit. I don't like the old version. One of the problems is that this commit doesn't have any way to test it, so it's difficult to compare with the old version.

The current version says in [optimie.c#L4225](https://github.com/gus-massa/racket/commit/565556403f472e09d9996b87f66cf221fdf87b08#diff-5e09d7df9102fa81a4736e357a0c3e1aR4225)

    /* Don't add the type if something is already there, this may happen when no_types. */
    if (do_optimize_get_predicate(pos, info, 1)
        || optimize_is_local_type_valued(info, pos)) {
      return;
    } 

But I'm not sure if it's a good idea to check the local type value, perhaps it's better to just use

    if (do_optimize_get_predicate(pos, info, 1)) {
      return;
    }

Finally, I just added the last commit. It adds no new functionality, but it was annoying that the arguments of `optimize_get_predicate` are in the inverse order of all the other similar functions.

mflatt:
This looks ready to merge.

gus-massa:
I rebased this, but I got errors in the CI.

The error in Travis is a common error in previous commits, sometimes fail gcc and sometimes clang. It's unrelated so I'm ignoring it.

The error in AppVeyor is new. It fails randomly approximately one half of the times. The error is in random places and is not repeatable. The error is so weird that I guess it's a bad interaction with the garbage collector????? I'll try to bisect and repeat the test, to try to find the problem. Test history: https://ci.appveyor.com/project/gus-massa/racket/history

gus-massa:
Almost all the error were related to alams ans similar code. I guess the fix goes elsewhere.

The only case where I got another error was:

    FAILED ->i-values17 rewrite-to-add-opt/c
      got "y-dom-blame: undefined;\n cannot reference undefined identifier"
      expected no-exn-raised
    FINISHED arrow-i.rkt

I tried to reproduce the error in my computer and AppVeyor, but I was unsuccesaful.

If you agree, I'll rebase and merge this as is.

mflatt:
Yes, I agree with merging.

samth:
That error is most likely an intermittent macro expander bug that we've seen before but not been able to track down.

PR #1078, opened 2015-10-03
Correct shown portion of error message

Blaisorblade:
Currently this documentation page [1] ends with the unexpected

> and the source location where the contract appears.
>> (assuming the contract is correct)

Please see the original at http://docs.racket-lang.org/guide/contract-func.html to see the above in context.

With this change, it ends with the more sensible:
> and the source location where the contract appears.
>> at: eval:5.0

I suspect that this was the original output, but that the message changed and the docs weren't updated.

[1] http://docs.racket-lang.org/guide/contract-func.html


rfindler:
Thanks! I've pushed a fix (only slightly different from your suggestion).

https://github.com/racket/racket/commit/2516374744153cd571e18e773df04636fad930ce

Blaisorblade:
Thank you for the prompt reaction!

PR #1086, opened 2015-10-08
Don't treat file://a/b as being a vaild URL on Windows

bmastenbrook:
`string->url` was treating `file://c:/fish.plt` as if it was equivalent to the path `c:\fish.plt`; however, this is not a valid `file://` URL on Windows (see http://blogs.msdn.com/b/ie/archive/2006/12/06/file-uris-in-windows.aspx). Additionally, it doesn't round trip this way: `(url->string (string->url "file://fish.plt"))` returned `"file:///fish.plt"`. Removing this behavior should fix a pkg test on appveyor.

### DIFF
--- a/racket/collects/net/url-string.rkt
+++ b/racket/collects/net/url-string.rkt
@@ -198,29 +198,16 @@
      (when (and scheme (not (regexp-match? #rx"^[a-zA-Z][a-zA-Z0-9+.-]*$"
                                            scheme)))
        (url-error "Invalid URL string; bad scheme ~e: ~e" scheme str))
-     ;; Windows => "file://xxx:/...." specifies a "xxx:/..." path
-     (let ([win-file? (and (or (equal? "" port) (not port))
-                           (equal? "file" scheme)
-                           (eq? 'windows (file-url-path-convention-type))
-                           (not (equal? host "")))])
-       (when win-file?
-         (set! path (cond [(equal? "" port) (string-append host ":" path)]
-                          [(and path host) (string-append host "/" path)]
-                          [else (or path host)]))
-         (set! port #f)
-         (set! host ""))
-       (let* ([scheme   (and scheme (string-downcase scheme))]
-              [host     (and host (string-downcase host))]
-              [user     (uri-decode/maybe user)]
-              [port     (and port (string->number port))]
-              [abs?     (or (equal? "file" scheme)
-                            (regexp-match? #rx"^/" path))]
-              [path     (if win-file?
-                          (separate-windows-path-strings path)
-                          (separate-path-strings path))]
-              [query    (if query (form-urlencoded->alist query) '())]
-              [fragment (uri-decode/maybe fragment)])
-         (make-url scheme user host port abs? path query fragment))))
+     (let* ([scheme   (and scheme (string-downcase scheme))]
+            [host     (and host (string-downcase host))]
+            [user     (uri-decode/maybe user)]
+            [port     (and port (string->number port))]
+            [abs?     (or (equal? "file" scheme)
+                          (regexp-match? #rx"^/" path))]
+            [path     (separate-path-strings path)]
+            [query    (if query (form-urlencoded->alist query) '())]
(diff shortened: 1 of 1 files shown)


bmastenbrook:
Note that at the very least I need to fix the docs and the tests in the racket/net repository before this can be merged.

mflatt:
While I agree that `string->url` accepts some ill-formed "file://" URLs, but I worry that changing the function to be more strict at this point is going to break something.

Support for broken "file://" URLs on Windows was added in 0153e122b7423d578acc480904f7a0fcfd52f46d. Unfortunately, I hadn't yet learned to write down motivations in comments or commit messages, but by correlating the commit date and my e-mail log, I've tracked down the motivation to this (long) discussion:

http://bugs.racket-lang.org/query/?debug=&database=default&cmd=view+audit-trail&cmd=view&pr=8060

I didn't actualy read that whole discussion again, but I think the gist is that we do something with broken URLs for the same reason that IE 7 does (as noted in the referenced blog post).

As for the original problem, I think the solution is probably to adjust the test suite to take into account the conventions of `url->string` on Windows.

Meanwhile, I think the docs for `url->string` are not quite right, since "file:///..." doesn't produce a path that starts with "/", and "%" escapes are handled as you'd expect.

samth:
The original problem came up because `(string->url "file://pkg.plt")` produces substantially different results on Windows than on other platforms. I agree that making it reject ill-formed paths is probably a bad idea, but could `string->url` produce consistent results for this URL on Windows?

samth:
Also, what's the correct URL to use for all the places that `"file://something"` appears in the pkg tests?

mflatt:
@samth: I don't understand the suggestion. The consistent (with respect to standards and other platforms) treatment of the URL is to make it not refer to a file, while the current Windows-specific treatment of the path is intended to make it refer to a file. So, I think it can't produce consistent results without rejecting (in the sense I meant) the URL as a reference to a file — but maybe you have some other point in the space in mind.

If we go with changing the package tests, I think the URLs doesn't change, but some tests' expected results are dependent on the platform; "file://something" refers to a file on Windows, while it doesn't refer to a file on other platforms.

samth:
Ah, ok, I understand what you mean now. One interesting question is if any currently-supported version of IE accepts these URLs.

As for the tests, I'm thinking of tests like this one: https://github.com/racket/racket/blob/master/pkgs/racket-test/tests/pkg/tests-name.rkt#L152 It seems weird to me that the behavior of parsing a package source would differ depending on platform. But maybe the right thing is to be more strict in the pkg name parsing?

mflatt:
The latest I have at hand at the moment is IE 9, which does support some bad URLs, as does Chrome. It's not clear which bad ones are supported, though.

A strict mode that is enabled for package URL parsing sounds like a good option.

samth:
One possibility would be to just set `file-url-path-convention-type` to `'unix` when calling `string->url` in the pkg library. Is that likely to be sufficient here?

bmastenbrook:
It would certainly fix those tests. After looking at it further, I don't think those tests are doing anything useful though - they're basically asserting that `(string->url "file://fish.plt")` has no path. Maybe the best thing to do is to parameterize those tests and leave this behavior alone, if those tests even need to exist at all.

(I'm still not convinced this `string->url` behavior is correct though, and other things seem to stick out to me - I would expect turning a UNC path on a server into a URL would produce something with a hostname, and in practice it doesn't round trip at all.)

mflatt:
@bmastenbrook You're right that `string->url` isn't compatible with the IE encoding of UNC paths. (Probably we failed to find any spec back then. As far as I can tell, blog posts are still the closet thing that exists to a spec.)

I guess I'll take it from the top and try to figure out the conventions again. :(

samth:
@jeapostrophe can you say more about the goal of the tests in question?

mflatt:
@samth I think I wrote that test; it was to catch problems parsing URLs, so it did it's job in provoking this thread

Issue #1083, opened 2015-10-06
string-prefix? change argument order


bennn:
@stchang just pointed out that  `string-prefix?` in `racket/base` has the signature:
```
(string-prefix? str prefix)
```
but [srfi/13](http://docs.racket-lang.org/srfi/srfi-std/srfi-13.html#string-prefix-p) implements
```
(string-prefix? prefix str)
```

Same goes for `string-suffix?`. I personally like giving the prefix 2nd, but we should probably change this to match the srfi. Also `list-prefix?` in `racket/base` takes its prefix first.

samth:
I don't think there's any need to change to match the SRFI. Consistency with `list-prefix?` is more important, but not necessarily definitive.

PR #1087, opened 2015-10-10
string-prefix/suffix, change arg order

bennn:
For #1083, in case we want to make the change.


bennn:
Nevermind then. I'm personally happier with this order :smile:

PR #1090, opened 2015-10-13
Add a check in the compilation-top reader.

samth:
This bug was found by fuzz testing.


samth:
@mflatt I'm mostly wondering if there is any other constraint that should be checked here.

PR #1091, opened 2015-10-15
fix ranges of `set-union!`, `set-intersect!`, `dict-clear!` to be `vo…

philnguyen:
Fix typos in documentation for functions that do imperative updates.


mflatt:
merged - thanks!

PR #1093, opened 2015-10-16
Add a missing property to an expression lifted by the contract system.

stamourv:
Caused TR to miss it.

Submitted as a PR pending changes to TR.

### DIFF
--- a/racket/collects/racket/contract/base.rkt
+++ b/racket/collects/racket/contract/base.rkt
@@ -43,7 +43,8 @@
                "private/struct-dc.rkt"
                "private/struct-prop.rkt")
  (except-out (all-from-out "private/base.rkt")
-             current-contract-region)
+             current-contract-region
+             (for-syntax lifted-key add-lifted-property))
  (except-out (all-from-out "private/misc.rkt")
              check-between/c
              check-unary-between/c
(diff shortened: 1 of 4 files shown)


rfindler:
Is TR generating `opt/c` contracts?

samth:
This isn't for contracts TR generates, but for contracts on functions used in typed modules.

rfindler:
Are you saying that someone is writing `opt/c` in TR code and then TR uses
that as a contract?

samth:
No. @stamourv can give you the actual code that ran into this issue, but what's happening here is that some code in the Racket standard library, unrelated to TR, uses these contract features. Then the function with that contract is used in a typed module, which causes the expander to lift some things. TR attempts to ignore that code while type checking, but the relevant annotation was missing in this case, so TR did the wrong thing.

rfindler:
Ah, got it.

This seems fine to push (please don't use a merge commit -- I can do it if
you want)

stamourv:
FWIW, one of the failing tests was using `~r` and other functions from `racket/format` in TR code. Those functions have contracts, and so their call sites in the typed module triggered some lifts.

stamourv:
Merged.

PR #1100, opened 2015-10-23
Allow separate read and write contracts for box/c

AlexKnauth:
This allows `(box/c write-ctc read-ctc)` to specify that a write to a box goes through `write-ctc` and a read from the box goes through `read-ctc`. This allows box contracts that restrict mutation of the box. 

Also, the `write-ctc` is contra-variant and the `read-ctc` is co-variant, so this means that a contract for the union of two mutable box types can be written using a more restrictive `write-ctc`, which is `and/c` of the two write-ctc contracts, and a less restrictive `read-ctc`, which is the `or/c` of the two read-ctc contracts.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -363,14 +363,16 @@ Returns the same contract as @racket[(vector/c c ... #:immutable #t)]. This form
 reasons of backwards compatibility.}
 
 
-@defproc[(box/c [c contract?]
+@defproc[(box/c [in-c contract?]
+                [c contract? in-c]
                 [#:immutable immutable (or/c #t #f 'dont-care) 'dont-care]
                 [#:flat? flat? boolean? #f])
          contract?]{
-Returns a contract that recognizes boxes. The content of the box must match @racket[c].
+Returns a contract that recognizes boxes. The content of the box must match @racket[c],
+and mutations on mutable boxes must match @racket[in-c].
 
 If the @racket[flat?] argument is @racket[#t], then the resulting contract is
-a flat contract, and the @racket[c] argument must also be a flat contract.  Such
+a flat contract, and the @racket[out] argument must also be a flat contract.  Such
 flat contracts will be unsound if applied to mutable boxes, as they will not check
 future operations on the box.
 
--- a/pkgs/racket-test/tests/racket/contract/box.rkt
+++ b/pkgs/racket-test/tests/racket/contract/box.rkt
@@ -6,6 +6,18 @@
                             (λ (box v) v)
                             (λ (box v) v))])
       (contract (box/c any/c) v 'pos 'neg)))
+
+  (test/no-error
+   '(let ([v (chaperone-box (box-immutable 1)
+                            (λ (box v) v)
+                            (λ (box v) v))])
+      (contract (box/c none/c any/c) v 'pos 'neg)))
+
+  (test/no-error
+   '(let ([v (chaperone-box (box 1)
+                            (λ (box v) v)
+                            (λ (box v) v))])
+      (set-box! (contract (box/c boolean? none/c) v 'pos 'neg) #t)))
   
   (test/pos-blame
    'box/c1
@@ -30,16 +42,33 @@
   (test/neg-blame
    'box/c6
    '(set-box! (contract (box/c boolean?) (box #f) 'pos 'neg) 11))
+
+  (test/neg-blame
+   'box/c6
+   '(set-box! (contract (box/c boolean? any/c) (box #f) 'pos 'neg) 11))
   
   (test/neg-blame
    'box/c7
    '(set-box! (contract (box/c boolean?) (box 12) 'pos 'neg) 11))
+
+  (test/neg-blame
+   'box/c7
+   '(set-box! (contract (box/c boolean? any/c) (box 12) 'pos 'neg) 11))
   
   
   (test/neg-blame
    'box/c-with-cons/c-inside
    '(let ([f
            (contract (box/c (cons/c (-> boolean? boolean?) '()))
+                     (box (list values))
+                     'pos 
+                     'neg)])
+      ((car (unbox f)) 3)))
+
+  (test/neg-blame
+   'box/c-with-cons/c-inside
+   '(let ([f
+           (contract (box/c (cons/c (-> boolean? boolean?) '()) any/c)
                      (box (list values))
                      'pos 
                      'neg)])
--- a/racket/collects/racket/contract/private/box.rkt
+++ b/racket/collects/racket/contract/private/box.rkt
@@ -12,10 +12,9 @@
 (define/subexpression-pos-prop (box-immutable/c elem)
   (box/c elem #:immutable #t))
 
-(define-struct base-box/c (content immutable))
+(define-struct base-box/c (content-w content-r immutable))
 
 (define (check-box/c ctc val blame)
-  (define elem-ctc (base-box/c-content ctc))
   (define immutable (base-box/c-immutable ctc))
   (unless (box? val)
     (raise-blame-error blame val '(expected "a box" given: "~e") val))
@@ -29,7 +28,6 @@
     [(dont-care) (void)]))
 
 (define (check-box/c-np ctc val blame)
-  (define elem-ctc (base-box/c-content ctc))
   (define immutable (base-box/c-immutable ctc))
   (cond
     [(box? val) 
@@ -55,7 +53,7 @@
                           val '(expected "a box" given: "~e") val))]))
 
 (define (box/c-first-order ctc)
-  (define elem-ctc (base-box/c-content ctc))
+  (define elem-r-ctc (base-box/c-content-r ctc))
   (define immutable (base-box/c-immutable ctc))
   (λ (val)
     (and (box? val)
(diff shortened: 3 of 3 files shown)


rfindler:
Thanks! This looks like a nice improvement.

Please don't delete those tests (just add new ones). That is, make it so that a one-argument box/c still prints the way it currently does, perhaps by having a different internal struct for one argument contracts or perhaps by putting a #f in a field somewhere.

Also, what should happen with this kind of interaction:

```
(let ([b (contract (box/c integer? string?)
                   (box 0)
                   'pos 'neg)])
  (set-box! b "x")
  (unbox b))
```

I don't think that should be a positive (or a negative) violation. Instead it seems to be some other kind of violation somehow. Well, I'm not certain, but it seems wrong.

One idea would be to insist that the mutation side contract must be `contract-stronger?` than the lookup side.

@mfelleisen @chrdimo any opinions?

AlexKnauth:
Re: Please don't delete those tests:
Where did I delete tests?

Re: write contract stronger than read contract:
Yes, but would that disallow contracts that are actually stronger but that contract-stronger? doesn't know about?

AlexKnauth:
Oh, you mean don't change printing when the two contracts are the same. Ok I'll do that.

rfindler:
Re: deleting. I would say "no minus signs in the diff for the testing
files".

Re: stronger: yes. But we should try to make stronger work better. So if
there are missing cases we should fix them. Except we cannot do that for
predicates (or only to a
Limited extent). Is that the worry?

But if you don't like that plan then can you say what should happen when
the two contracts are legitimately incomparable? Like in the example? Is
that behavior actually right.

AlexKnauth:
Does a write-only reference to a box make sense? I'm not sure. But either way it doesn't seem like box/c can enforce it without being way too restrictive. Originally I thought people would be get around this problem by saying `(box/c (and/c write-ctc read-ctc) read-ctc)`, but `(contract-stronger? (and/c pair? pair?) pair?)` doesn't even work.

rfindler:
I do see that and/c has a fairly naive contract-stronger implementation and
I seem to recall I did that because the obvious improvement is an n^2
search. But maybe it is worth trying that out now.

I think a write-only reference to a box makes sense. Your commit would
allow such a thing, as do chaperones currently.

My concern is with blame assignment. If I say that you may write this box
only with numbers and read it only with strings and then you write number
and I read that number, someone must be blamed. It seems unfortunate to
blame me for that problem with the read and you certainly should have been
allowed -- that's what the contract explicitly says. So what does that
leave? My only conclusion is that it is a bad contract. So we could use
contract-stronger to rule out contracts where this is a problem. We could
also redirect the blame to the contract itself the same way that ->i does.
That would eliminate the need to improve contract-stronger.

Maybe there's another thing we could do, I'm not sure.

AlexKnauth:
Redirecting the blame work, how would that work?

rfindler:
When you write an `->i` contract, the contract itself can be blamed for a self-violation. Here's an example.

I see the first line of this contract error message is wrong and needs to be fixed; look at the blaming line.

```
#lang racket

(module contract-lib racket
  (provide a-contract)
  (define a-contract
    (->i ([f (-> integer? integer?)]
          [x (f) (<= (f #f) 3)])
         any)))

(module server racket
  (require (submod ".." contract-lib))
  (provide
   (contract-out
    [app a-contract]))
  (define (app g x) 1))

(module client racket
  (require (submod ".." server))
  (app (λ (x) 0) 0))

(require (submod "." client))
```


What I mean, however, was that the `box/c` contract could pick up the name of the containing module and then assign blame to that when it gets a violation of this kind (where the contract on the "mutate" side of the box passes but the one on the "lookup" side fails).

But now that I say this, I realize this situation isn't quite the same, as we aren't in a situation where the contract is somehow blaming itself.

I'm much less convinced of this "shift the blame" approach now.

The interesting thing is that if we think of a box as a pair of functions (a getter and a setter) then this kind of blame seems totally fine. The issue is somehow solely because the box comes with this promise that there is exactly one place that's being updated.

chrdimo:
I think the read/write contracts distinction makes sense.

Also I don’t see any problem with blame.

For instance in Robby’s example, the module (code region) that provides the box with the contract (box/c integer? string?)
 basically says to its clients: “It is your responsibility to write to the box values that satisfy integer? and I promise that when you read from the box you will get values that satisfy string?.” Of course this is hopeless but the provider should know better before imposing on the box a contract that it is impossible for the provider to meet.

This is not indy blame. Indy blame has to do with the contract itself creating a value that violates a contract. Here we do not have such a case.

Another way to see this is to view the box via its second-order interpretation as the values function with contract (->  (-> write/c) (-> read/c)).


On Oct 24, 2015, at 9:52 AM, Robby Findler <notifications@github.com<mailto:notifications@github.com>> wrote:


When you write an ->i contract, the contract itself can be blamed for a self-violation. Here's an example.

I see the first line of this contract error message is wrong and needs to be fixed; look at the blaming line.

#lang racket

(module contract-lib racket
  (provide a-contract)
  (define a-contract
    (->i ([f (-> integer? integer?)]
          [x (f) (<= (f #f) 3)])
         any)))

(module server racket
  (require (submod ".." contract-lib))
  (provide
   (contract-out
    [app a-contract]))
  (define (app g x) 1))

(module client racket
  (require (submod ".." server))
  (app (λ (x) 0) 0))

(require (submod "." client))


What I mean, however, was that the box/c contract could pick up the name of the containing module and then assign blame to that when it gets a violation of this kind (where the contract on the "mutate" side of the box passes but the one on the "lookup" side fails).

But now that I say this, I realize this situation isn't quite the same, as we aren't in a situation where the contract is somehow blaming itself.

I'm much less convinced of this "shift the blame" approach now.

The interesting thing is that if we think of a box as a pair of functions (a getter and a setter) then this kind of blame seems totally fine. The issue is somehow solely because the box comes with this promise that there is exactly one place that's being updated.

rfindler:
@chrdimo : I agree with you that if this were two functions then that's fine. The point is that these are NOT functions. Accessor and getter functions do not provide the same guarantees to the programmer as boxes do. In particular, when you have a box and you know you have the only reference to the box, then you know that what you put in is what you get out.

I think we should be taking into account the semantics of the underlying contract.

It is this "oh this could be modeled by functions"-style of thinking that led to us not realizing we needed chaperones for so long. And this was an error.

samth:
But note that it's easy to satisfy this contract by implementing the box with an impersonator.

rfindler:
@samth yes, good point. And so I suppose I should be fine with this contract as an impersonator contract.

samth:
I think you should be fine with it even as a chaperone contract. The impersonator is only needed to implement something that satisfies the contract -- the contract itself doesn't violate the chaperone invariant.

chrdimo:
On Oct 24, 2015, at 11:43 AM, Robby Findler <notifications@github.com<mailto:notifications@github.com>> wrote:


@chrdimo<https://urldefense.proofpoint.com/v2/url?u=https-3A__github.com_chrdimo&d=CwMCaQ&c=WO-RGvefibhHBZq3fL85hQ&r=M8TmlY7TpuYqG_DGXpL4c3JwGKDvIYRdn-6BrSn9UR4&m=NUaK0x2WT9HuSiA8SboxF5PBp0YGuUmr2appErRQdN8&s=shJowyr7fjJslUF7hte8JKH_ApjmEiSBMkkEOdDWZc4&e=> : I agree with you that if this were two functions then that's fine. The point is that these are NOT functions. Accessor and getter functions do not provide the same guarantees to the programmer as boxes do. In particular, when you have a box and you know you have the only reference to the box, then you know that what you put in is what you get out.

I think we should be taking into account the semantics of the underlying contract.

It is this "oh this could be modeled by functions"-style of thinking that led to us not realizing we needed chaperones for so long. And this was an error.


I disagree. Chaperones and impersonators are necessary to obtain the same interposition points that one would have if boxes were second-order functions. The problem with the previous situation was that boxes were treated as first-order functions.

As for the ownership of the box, I don’t see how the contract reflects it or how the semantics of box/c implies it.

The way I understand box/c with write/c and read/c is that the client of the box/c is obliged to write write/c values and is promised to read read/c values. If read/c and write/c are the same contract content/c then content/c becomes an invariant of what is in the box (the same invariant that box/c enforces currently). A similar invariant exists if write/c implies read/c and so on. But the invariant does not hold in the general case and this is something that the writer of the module that imposes the box/c contract should easily figure out.

rfindler:
I think what you're saying, @chrdimo, is that it is possible to have a contract with incompatible read and write aspects, even with a chaperone contract where a program still can (think it can) succeed. That is, the server could retain the uncontracted version of the box and then when a string/integer is put in from the outside, it can notice that and put a integer/string back into the box before the next time someone looks inside the box. And if it was wrong about it's ability to do this dance then it gets blame. Appropriately.

And, of course, I cannot disagree with this. 

The problem is that there is no useful way to have such a contract, as we cannot give out the "mutate" side of a box without simultaneously giving out the "lookup" side. So such contracts are just stupid.

chrdimo:
@rfindler: I agree. But we cannot prevent stupid contracts. Think about the classic example from the Blume-MacAllester paper:

(contract-out [values (-> (-> integer? integer?) any)]

The best we can should do is that contracts have clear meaning in terms of what responsibilities they imply. And I think this new box/c has that.

rfindler:
@chrdimo but that contract is only stupid with a specific value. This variation on `box/c` admits contracts that are stupid for ANY box.

(I suppose at this point you could ask why I put `none/c` into the contract library. Good question!)

chrdimo:
@rfindler I think this is a question about pragmatics and not about semantics. Technically, the new box/c does not challenge any of the contracts meta-theory. Now the point you make is whether the extra expressiveness is justified. I would say maybe it is. Yes, you can write many stupid contracts for boxes that you were not able to write before and that may require some extra thought about what you have to do to make sure you live up to the contract (it is intractable to exclude or even detect after failure all such contracts without hampering expressiveness). But at the same time you can write many useful contracts such as the one that allows you to share a mutable box with another module but prevent the other module from writing into the box. Maybe there is a sweet-pot between expressiveness and avoiding stupid contracts but this only practice will reveal.

One question I have is what triggered this modification of box/c?

rfindler:
@chrdimo I think one could ask the semantic question "Are there any contracts for which there is no value that escapes blame?". I think any good semantic question is motivated by pragmatics.

@AlexKnauth submitted this PR; I'm not sure of his motivation.

AlexKnauth:
Motivation: Having coffee with Northeastern PL people, and Brian said something about box/c contracts being invariant, Ben Lerner said something about boxes where reading is co-variant and writing is contra-variant, and I started thinking about that, and I remembered [this](http://lists.racket-lang.org/users/archive/2014-July/063551.html) discussion on the mailing list about unions of vector types in typed racket. 

This PR fixes that for box contracts. Unions of mutable box types will make sense, and this could also be done for vectors and other mutable values like vectors in the future. I was also thinking that this would allow better typed racket support for read-only boxes, vectors, etc. without making a completely separate type.

rfindler:
Alright, your motivation to do it is far stronger than my wanking around to try to find a variation on it that avoids this problem I was talking about.

Please push away. (Or I can, if you don't have access.)

AlexKnauth:
I don't have access, so you should push.
But I do think it's a bad thing for a box contract to accept completely unrelated contracts for these; I just don't think it's something we can enforce without being too restrictive.

rfindler:
Okay, I've pushed it: https://github.com/racket/racket/commit/67e3899272792c2e510abdf39865b3a40dfd8a8d

PR #1102, opened 2015-10-24
Improve how multi-in assigns source location info for its expansion

lexi-lambda:
This changes how `multi-in` is implemented so that the location for each expanded element in the final `require` spec is tied to the last relevant module path element. This allows DrRacket to intelligently show arrows linking each imported binding with a relevant piece of the multi-in import spec.

To demonstrate how this looks in DrRacket, I've taken two screenshots:

screen shot 2015-10-24 at 1 27 10 pm
screen shot 2015-10-24 at 1 27 30 pm

In contrast, without this change, no syntax arrows whatsoever are displayed for any identifiers imported with `multi-in`.

### DIFF
--- a/racket/collects/racket/require.rkt
+++ b/racket/collects/racket/require.rkt
@@ -1,6 +1,6 @@
 #lang racket/base
 
-(require (for-syntax racket/base racket/require-transform racket/list
+(require (for-syntax racket/base racket/require-transform racket/list syntax/stx
                      (only-in racket/syntax syntax-local-eval))
          "require-syntax.rkt")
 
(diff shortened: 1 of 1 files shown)


rfindler:
This looks great to me! Not that it is really required, but would it be okay to put those tests into a `(module+ test ...` ?

lexi-lambda:
I actually tried to lift the tests into a test submodule, but it didn't work as expected: because the tests are run in phase 1, they were getting run when required by another module, even though they were in a test submodule. I'm not really sure if there's a good way to handle that, but I just pulled the test back out of the submodule for now and kept the comment.

racket/collects/racket/require.rkt
```diff
@@ -128,25 +130,25 @@
...
-;;       (multi '(("a" "b" "c" "d")))   => '("a" "b" "c" "d")
-;;       (multi '(("a" "b" ("c" "d")))) =error> ""
-;;       (multi '(a b c))       => '(a/b/c)
-;;       (multi '(a (b c) d))   => '(a/b/d a/c/d)
-;;       (multi '(a b (c d)))   => '(a/b/c a/b/d)
-;;       (multi '((a b) c d))   => '(a/c/d b/c/d)
-;;       (multi '((a b) (c d))) => '(a/c a/d b/c b/d)
-;;       (multi '((a b c d)))   => '(a b c d)
-;;       (multi '((a b (c d)))) =error> "")
+#;(begin-for-syntax
```
samth: Couldn't this live in `module+ test` instead of being commented?

lexi-lambda: @samth See the comment I left in the main PR discussion.

rfindler:
I'm okay with merging this and I see that Eli's the primary maintainer so shall I do it?

mflatt:
A related issue is that tests in "collects" cannot depend on packages like `eli-tester`. Tests for things in "collects" generally should be in a separate package, like "racket-test-core".

Out of curiosity, @lexi-lambda, did you try `(module+ test (begin-for-syntax ...))` or `(begin-for-syntax (module+ test ...))`? The former would make tests run when compiling the module (which is bad), while the latter should work right.

lexi-lambda:
> I'm okay with merging this and I see that Eli's the primary maintainer so shall I do it?

FWIW, it looks like @stamourv is the author of `multi-in`, if you want some more eyeballs. That said, I'm pretty confident this is good to be merged. The only potential problem is if people were writing really crazy things that expand to `multi-in`... the source location changes could *potentially* cause behavioral differences. Doing something like that seems incredibly unlikely, though.

> Out of curiosity, @lexi-lambda, did you try `(module+ test (begin-for-syntax ...))` or `(begin-for-syntax (module+ test ...))`? The former would make tests run when compiling the module (which is bad), while the latter should work right.

Aha, you're right, I did the former. The difference doesn't seem super intuitive, but upon further thought, that makes sense. The "collects" point is reasonable, though.

rfindler:
I'll let Vincent do the needful, then. I agree that this commit is pretty
safe and a clear improvement.

stamourv:
Implementation looks good. I'll merge and move the tests.

stamourv:
Merged, thanks!

PR #1104, opened 2015-10-27
typo in doc for make-syntax-delta-introducer

stchang:
I'm not sure about the _m-id_ part but I think this is what the docs are trying to say.


mflatt:
merged - thanks!

PR #1116, opened 2015-11-01
Add equivalents of hash-keys, hash-values, in-hash for id-tables

takikawa:
Added some missing id-table functions. Bumped Racket version because of API additions.


stamourv:
LGTM

PR #1119, opened 2015-11-04
provide example for in-directory

opyate:
Given  [in-directory](http://docs.racket-lang.org/reference/sequences.html?q=filesystem#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._in-directory%29%29) is a "start quickly" example on the [main site](http://racket-lang.org/), I thought might benefit from an example, especially considering the `use-dir?` parameter, which had me slightly confused (late) last night (I expected to see only `/example/b` and `/example/b/beta` in the output, but re-read the doc and saw **except for the CONTENTS of any directory for which**).


mflatt:
Merged - thanks!

PR #1124, opened 2015-11-06
fix documented contract ranges of `set-subtract!` and `set-symmetric-difference!`

philnguyen:
This pull request fixes the ranges of the remaining imperative set methods. Somehow I missed these last time...


mflatt:
merged - thanks!

PR #1127, opened 2015-11-09
net/head: better error message for bytes/string mismatch

jbclements:
Currently, 

```racket
 (extract-field "def" #"abc: 3\r\ndef: 7\r\n\r\n")
```
produces a really nasty error message that exposes regexp internals. This commit checks and gives a better error message.

### DIFF
--- a/racket/collects/racket/contract/private/misc.rkt
+++ b/racket/collects/racket/contract/private/misc.rkt
@@ -630,7 +630,7 @@
   (raise-blame-error blame #:missing-party neg-party val
                      '(expected: "~s" given: "~e")
                      (if empty-ok?
-                         "list?"
+                         'list?
                          (format "~s" `(and/c list? pair?)))
                      val))
 


stamourv:
LGTM

jbclements:
@stamourv  Should I be concerned about the continuous integration failures? They both appear related to xml code, and I don't see an obvious direct connection to this pr, but I could test further if they don't look familiar to you....

samth:
The XML test failure has been fixed.

stamourv:
I don't see a connection either. If those tests pass on your end with your patches, then I wouldn't worry.

samth:
If you just rebase and force-push this PR, it will rerun the CI and pass.

mflatt:
@jbclements This is a slight abuse of `raise-argument-error`; the second argument is intended to be a contract (but in string form), as opposed to prose

PR #1128, opened 2015-11-09
Add more hash-like operations to id-tables

takikawa:
The operations are ref!, set*, set*!, update, and update!. Also bumps version number.


stamourv:
LGTM

rmculpepper:
What's wrong with the generic versions of these functions from `racket/dict`?

takikawa:
The `racket/dict` library adds a lot of overhead due to contracts. I'm trying to remove uses of dict functions from TR in favor of specialized operations to speed up typechecking.

mflatt:
This is ok with me.

PR #1131, opened 2015-11-10
Fix pattern-expander example

AlexKnauth:
Pointed out by https://groups.google.com/d/msg/racket-users/YIYOVPo7W9Y/CEQCejGVBAAJ


AlexKnauth:
Replaced by https://github.com/racket/racket/pull/1133, which actually fixes the problem.

PR #1133, opened 2015-11-11
syntax/parse: disarm the result of expanding patterns

AlexKnauth:
Fixes the bug pointed out in https://groups.google.com/d/msg/racket-users/YIYOVPo7W9Y/CEQCejGVBAAJ

### DIFF
--- a/pkgs/racket-test/tests/stxparse/test.rkt
+++ b/pkgs/racket-test/tests/stxparse/test.rkt
@@ -618,6 +618,17 @@
                                       (+ x y z))))
                   (syntax->datum #'(let ([x 1] [y 2] [z 3])
                                      (+ x y z))))
+
+    (define-syntax ~maybe
+      (pattern-expander
+       (syntax-rules ()
+        [(~maybe pat ...)
+         (~optional (~seq pat ...))])))
+
+    (check-equal? (syntax-parse #'(1 2 3 4 5 6)
+                    [((~maybe a b) ...)
+                     (syntax->datum #'(a ...))])
+                  '(1 3 5))
     ))
 
 (test-case "this-syntax"
--- a/racket/collects/syntax/parse/private/rep.rkt
+++ b/racket/collects/syntax/parse/private/rep.rkt
@@ -475,7 +475,7 @@
             [mstx (introducer (syntax-local-introduce stx))]
             [mresult (parameterize ([current-syntax-parse-pattern-introducer introducer])
                        (proc mstx))]
-            [result (syntax-local-introduce (introducer mresult))])
+            [result (syntax-local-introduce (introducer (syntax-disarm mresult #f)))])
        (disappeared! #'id)
        (recur result))]
     [(id . rst)
@@ -487,7 +487,7 @@
             [mstx (introducer (syntax-local-introduce stx))]
             [mresult (parameterize ([current-syntax-parse-pattern-introducer introducer])
                        (proc mstx))]
-            [result (syntax-local-introduce (introducer mresult))])
+            [result (syntax-local-introduce (introducer (syntax-disarm mresult #f)))])
        (disappeared! #'id)
        (recur result))]
     [wildcard


sorawee:
@AlexKnauth: what's the status of this PR? I just got hit by a bug that this PR would fix. But then I saw https://github.com/racket/racket/commit/759c71e15824b5e0ba63fecbe76f80b1c4565fa2 which seems to indicate that the current behavior is expected, so I'm now not sure if this is a bug or not.

In any case, I'll switch to `syntax-case` for now.

AlexKnauth:
I'll try rebasing and resolving the conflicts.

AlexKnauth:
I have rebased, and am testing now. I will follow-up here if the tests succeed.

rmculpepper:
I think this is the wrong solution, and the right solution is probably just "don't use `syntax-rules` to write anything other than expression or definition macros".

Syntax protection is a security mechanism that macros use to prevent adversarial contexts from reinterpreting the syntax they produce. An expression transformer that tries to protect itself (using `syntax-rules`) can get put inside a pattern expander, and if `syntax-parse` disarms it and (re)interprets the syntax as a pattern, then that defeats the security mechanism.

Perhaps syntax protection should be tied to a specific nonterminal (eg, "Racket form", "syntax-parse pattern", etc). But that wouldn't change the fact that you can't use the default `syntax-rules` (for Racket forms) to implement a pattern expander.

AlexKnauth:
I think `syntax-parse`, when expanding syntax parse patterns, should have the same powers of traversing protected-but-not-tainted syntax objects as the Racket expander does. Perhaps it should disarm, for traversing, but rearm after traversing.

rmculpepper:
No. That breaks the security of the whole syntax protection system.

Here's a contrived example.

    #lang racket/load

    (module A racket/base
      (require syntax/parse (for-syntax racket/base))

      (define secret 42)

      (define-syntax (m stx)
        (syntax-case stx ()
          [(_ x)
           (syntax-protect
            #'(quote (~parse x secret)))]))

      (provide m))

    (require 'A)

    (require (for-syntax racket/base))
    (require syntax/parse syntax/parse/private/parse) ;; work around lazy-require problem...

    (current-code-inspector (make-inspector (current-code-inspector)))

    (define-syntax ~evil
      (pattern-expander
       (syntax-local-value #'m)))

    (printf "extracted ~s\n"
            (syntax-parse #'(blah)
              [(~evil x) #'x]))

The macro `m` produces a `quote` expression, and it uses `syntax-protect` to enforce that the syntax is not taken apart or reinterpreted. And since it's a quote expression, even though it mentions dangerous words like `secret`, it *knows* that that is interpreted as a datum and not as a variable reference.

But if you remove the `syntax-protect` (which is equivalent to changing `syntax-parse` to remove it automatically, then the syntax can be reinterpreted, and an evil context can extract the value of the secret.

AlexKnauth:
Okay. So disarming without rearming is bad, and so this PR as-is is wrong. However, if I modify this PR to disarm and re-arm syntax objects in certain cases, would that be okay?

On patterns that combine sub-patterns, such as list patterns, seqs, optionals-without-defaults, ands, and ors, I would disarm before destructuring, and rearm after destructuring. On patterns that have both sub-patterns and expressions, such as optionals-with-defaults and parse patterns, I would disarm-and-rearm for destructuring the sub-patterns, but not disarm for destructuring the sub-expressions. Would that fix the security issues you're talking about?

rmculpepper:
No.

rmculpepper:
To elaborate, "This syntax must be interpreted whole as an expression (or definition)." is not compatible with interpreting that syntax as a pattern, or interpreting that syntax as a pattern but doing lots of other complicated things, or taking it apart and interpreting the pieces in various ways.

AlexKnauth:
@rmculpepper 
So @sorawee and I were talking about updating the documentation, and we discovered that `match` expanders already do the sort of disarming and rearming that I suggested here and you warned against.

So, I adapted your example `~evil` exploit to use `match` instead of `syntax-parse`, so I guess this appears to be a security problem in `match`.
```
#lang racket

(module A racket
  (define secret 42)

  (define-syntax-rule (list-rest datum) (quote datum))

  (define-syntax (m stx)
    (syntax-case stx ()
      [(_ x)
       (syntax-protect
        #'(list-rest (app (λ (_) secret) x)))]))

  (provide m))

(require 'A)

(define-match-expander ~evil
  (syntax-local-value #'m))

(printf "extracted ~s\n"
        (match '(blah)
          [(~evil x) x]))
```

PR #1134, opened 2015-11-12
Flatten nested begin and begin0 forms after optimization

gus-massa:
The nested `begin`/`begin0` are flattened at read time, but some
optimizations may create new instances. So try again after optimizations. 

The main function is `flatten_sequence`. It has the same name than a function is `sfs.c` that is similar but not equivalent. This is not a problem for the c compiler, but I'm slightly worried about copy&paste errors in the future.

This commit has no tests, because `test-comp` uses the reader to see inside the compiled expressions, so (most?) of these `begins` are cleared. One possibility is to use `match` to inspect the compiled version. Another is to just add some test to ensure that the order is not messed. For example, reduce

    (begin (random 1) 0 (car (list (random 3) (random 3))) 0 (random 4))

to

    (begin (random 1) (random 2) (random 3) (random 4))

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1528,6 +1528,32 @@
 (test-comp '(lambda () (begin (random 1) (random 2)))
            '(lambda () (cdr (cons (random 1) (random 2)))))
 
+(test-comp '(lambda () (begin (random 1) (random 2) (random 3) (random 4)))
+           '(lambda () (begin (car (cons (random 1) (random 2))) (random 3) (random 4)))) ;
+(test-comp '(lambda () (begin (random 1) (random 2) (random 3) (random 4)))
+           '(lambda () (begin (cdr (cons (random 1) (random 2))) (random 3) (random 4))))
+(test-comp '(lambda () (begin (random 1) (random 2) (random 3) (random 4)))
+           '(lambda () (begin (random 1) (car (cons (random 2) (random 3))) (random 4)))) ;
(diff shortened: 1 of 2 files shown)


mflatt:
This seems like a good idea.
Is there any danger of quadratic-time behavior due to nested `begin`s? I think probably not, on the grounds that this flattening only cleans up `begin`s introduced by optimizations, but I'm not sure.
For testing. maybe inspect the result for this particular case via `zo-parse` instead of `read`.

gus-massa:
I rebased, fix a few parts, and added some tests. 

In a separate commit, I added a new fuel to avoid quadratic behavior. It's almost a copy of `shift_fuel`.


An example of code that will trigger quadratic behavior is

    (lambda (f)  
      (cdr (cons (f)
                 (cdr (cons (f)
                            (cdr (cons (f) (f))))))))

I don't know how usual something like this is in real code.

mflatt:
Ok! (I can easily imagine code like that generated from a macro.)

gus-massa:
Then I can squash and push them to the main repository, or you want to more time to do a review?

gus-massa:
I rebased and squashed them. The test are ok. Do I push this?

mflatt:
Looks ok to me.

gus-massa:
I tried but I got an error. Do my account have the permissions in this repository?

    git.exe push --progress  "origin-push" 15-11-Nested-Begin:master

    ERROR: Permission to racket/racket.git denied to gus-massa.
    fatal: The remote end hung up unexpectedly

gus-massa:
Fixed!

PR #1136, opened 2015-11-13
reformat racket/pretty code

AlexKnauth:
Mostly fixes indentation and square brackets.


stamourv:
While I agree that the result is nice, that kind of change kills, e.g., `git blame`, which makes it much harder to figure out the history of a given bit of code.
Because of this, I'll close this PR.
Thanks, though, and I agree with the sentiment! :)

PR #1137, opened 2015-11-13
add #:newline? argument to pretty-printing functions

AlexKnauth:
Versions of pretty-print, pretty-write, and pretty-display that don’t print extra newlines at the end.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/pretty-print.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/pretty-print.scrbl
@@ -6,14 +6,16 @@
 @note-lib[racket/pretty]
 
 @defproc[(pretty-print [v any/c] [port output-port? (current-output-port)]
-                       [quote-depth (or/c 0 1) 0])
+                       [quote-depth (or/c 0 1) 0]
+                       [#:newline? newline? boolean? #t])
          void?]{
 
 Pretty-prints the value @racket[v] using the same printed form as the
 default @racket[print] mode, but with newlines and whitespace inserted
 to avoid lines longer than @racket[(pretty-print-columns)], as
 controlled by @racket[(pretty-print-current-style-table)]. The printed
-form ends in a newline, unless the @racket[pretty-print-columns]
+form ends in a newline by default, unless the @racket[newline?]
+argument is supplied with false or the @racket[pretty-print-columns]
 parameter is set to @racket['infinity]. When @racket[port] has line
 counting enabled (see @secref["linecol"]), then printing is sensitive
 to the column when printing starts---both for determining an initial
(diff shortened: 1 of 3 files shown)


samth:
I think a keyword argument might be nicer here.

mflatt:
I was about to write the same as @samth...

AlexKnauth:
Ok done.

mflatt:
@AlexKnauth You seem to have added a keyword argument internally, but what I meant (and what I thought Sam meant) was to add an optional `#:newline?` argument to `pretty-print` instead of introducing the name `pretty-print/no-inline`, etc.

AlexKnauth:
Oh, okay, I'll do that then.

AlexKnauth:
I've updated this to add optional `#:newline?` keyword arguments to `pretty-print`, `pretty-write`, and `pretty-display`. I'll add documentation once the `make` is done.

rfindler:
Maybe in the docs point out that passing `#t` means that `printing-port-print-line` isn't called with `#f` any more?

AlexKnauth:
What do you mean? It isn't being called at all, in my current code...

AlexKnauth:
I see what you mean, okay.

rfindler:
Yeah: I just mean that in addition to the high-elvel description of the new argument, also include the technical note as to how it changes which callbacks are called back.

Thanks for this, btw.

AlexKnauth:
Does that documentation look good?

It's a bit weird because the `#:newline?` argument determines whether `pretty-print-print-line` is called, but `pretty-print-print-line` doesn't always print a newline. I'm not sure what to do about that though.

rfindler:
Generally these functions don't accept booleans, but accept `any/c` and treat non-`#f` values as true (eg `pretty-print-show-inexactness`).

I think the convention for documenting such things is to write "a true value" and "@racket[#f]", but I see a few other places in those docs where that convention is violated.

The sentence "If it is supplied with false, the @racket[pretty-print-print-line] callback is not called." doesn't seem right. I think it is just not called with `#f` anymore, but it is still called on all the other line breaks.

Also the second paragraph of `pretty-print-print-line`'s docs needs adjusting.

AlexKnauth:
I meant that it's not called for the last newline, which the previous sentence mentions, but you're right, it's not clear.

AlexKnauth:
I've modified it to clarify that if `newline?` is supplied with false, the `pretty-print-print-line` callback is not called after the printed value. I don't want to confuse people by saying "not called with false".

rfindler:
That is better, I agree.

AlexKnauth:
Logically, it's good to think of it as a boolean, and if we ever want to put a third setting in there, like `'just-internal` or `'just-end`, it might be better to document it as a boolean for now.

rfindler:
In that case,  you'll want to add a check?

AlexKnauth:
Okay, I've added the check

AlexKnauth:
Does this look good now?

rfindler:
Commits 1bfa0c5292679ec66271839417eb5afb8b1c7098 and
203dbf78d9deb3999a8eea236410b55a6d8f4456 look good to me, except that I
don't see any test suites. I didn't look over the others.

Robby

AlexKnauth:
Oh, right. Where are the pretty-printing tests?

rfindler:
Grep suggests pkgs/racket-test-core/tests/racket/pretty.rktl.

rfindler:
Tests look good to me.

AlexKnauth:
For some reason they error on my machine
```
Section(namespaces)
Section(collects)
Section(modprot)
ERROR: UNKNOWN::79: read (compiled): ill-formed code [../../../racket/gc2/../src/read.c:5585]
```
But they're passing on Appveyor and Travis

AlexKnauth:
It's possible that something weird is happening with my compiled files, since I've been working with both a snapshot version a few weeks old and the git HEAD version. So should this be merged?

rfindler:
It is okay with me to push those three commits (although do you see a
reason not to squash them?). Do you want me to do it?

AlexKnauth:
I don't have push access. So this shouldn't include the re-formatting stuff?

rfindler:
I didn't check over the formatting stuff. I can see how, tho, in this case my usual objections to formatting changes don't apply (this is old code that is almost never touched so the lost 'git blame' information is minor compared to the readability).

rfindler:
Thanks, I've pushed the commits (except that I combined the three commits that were all pieces of the same change)

PR #1139, opened 2015-11-16
Don't burn fuel when a single use function is inlined (with error, don't merge)

gus-massa:
Since this operation only moves the code and doesn't make the final
bytecode bigger, it's not necessary to decrease the fuel and then it
is available for further inlining.

With this change, the Racket build and tests complete without errors. But I tried a few additional changes and discovered a problem. [In particular, I was trying to burn less fuel when a function is small, there is a similar problem when increasing the initial fuel.]

The problem is that this program doesn't finish compiling:

    #lang racket/base
    (define ((proc))
      ((proc)))

And this neither:

    #lang racket/base
    (let ()
      (define ((proc))
        ((proc)))
      (void proc proc))

I think that the problem is the internal `lambda`s in `(lambda () (lambda () ...))`.

### DIFF
--- a/racket/src/racket/src/optimize.c
+++ b/racket/src/racket/src/optimize.c
@@ -1631,7 +1631,8 @@ static Scheme_Object *no_potential_size(Scheme_Object *v)
 static Scheme_Object *apply_inlined(Scheme_Object *p, Scheme_Closure_Data *data, Optimize_Info *info,
 				    int argc, Scheme_App_Rec *app, Scheme_App2_Rec *app2, Scheme_App3_Rec *app3,
                                     int context,
-                                    int nested_count, Scheme_Object *orig, Scheme_Object *le_prev)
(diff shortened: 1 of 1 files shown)


mflatt:
I'm not sure exactly what's going on, either. I agree that it has something to do with the `lambda` produced by inlining `proc` in `(proc)`, but it's not clear to me why inlining `proc` fails to decrement the fuel enough.

mflatt:
I merged this change, but with an extra flag on duplicated `lambda`s to prevent the infinite-loop problem.

The problem happens because inline-fuel decrements affect only more nested expressions, and the outer call for `((proc))` is not nested in the inner one. The new flag on `lambda`s causes the outer application to see a function that, despite being a literal `lambda`, it doesn't count as single-use.

gus-massa:
:)

I'll send my idea to burn less fuel with small functions in few days.

PR #1141, opened 2015-11-17
Fix blame error for listof contract

takikawa:
Comparison

Old

````
-> (contract (listof string?) 3 'pos 'neg)
; readline-input:1:0: broke its own contract
;   promised: "list?"
;   produced: 3
;   in: (listof string?)
;   contract from: pos
;   blaming: pos
;    (assuming the contract is correct)
; [,bt for context]
````

New

````
-> (contract (listof string?) 3 'pos 'neg)
; readline-input:1:0: broke its own contract
;   promised: list?
;   produced: 3
;   in: (listof string?)
;   contract from: pos
;   blaming: pos
;    (assuming the contract is correct)
; [,bt for context]
````


rfindler:
Thanks!

Issue #1140, opened 2015-11-16
Floating point literal parsed incorrectly


leppie:
```
> 0e400
0.0
> 0e401
+inf.0
> (system-type)
'unix
> (system-type 'word)
64
> (version)
"6.1" 
```
Tested on 6.2.1 as well. 

Works fine on Windows however, tried both 32/64 bit on 6.2.1.

cky:
I found where the bug is. I'll prepare a pull request for it soon. :-)

cky:
Posted the pull request now: #1145.

PR #1145, opened 2015-11-20
Fix parsing of 0e401 on USE_EXPLICT_FP_FORM_CHECK platforms.

cky:
This is a fix for #1140.

This does not fix `(string->number (string-append "0." (make-string 400 #\0) "1e400"))` (which will still parse as infinity on such Linux, Solaris, HP/UX, etc.), but the common cases should be correct.

(Also, "explict"? Eww. (I worried that I had made a typo, but realised that I actually cut and pasted the macro name which was misspelled.))


mflatt:
Merged – thanks!

PR #1146, opened 2015-11-21
Add a link to a Macro tutorial written by Greg Hendershott

zsh-89:
which helps new learners understand Racket macro better.


mflatt:
merged - thanks!

PR #1149, opened 2015-11-24
add note about sorting with NaN

jbclements:


stamourv:
I'd phrase this as "any comparison involving NaN returns false", which is more concrete, IMO.
Also, you may want to state explicitly what "a suprising result" may be. I.e., it won't be an error, and it won't be a segfault.

rfindler:
I think showing an example is really the best thing to do. Probably also to good to add an expected result and an error if the wrong answer comes back so we notice if the sorting algorithm changes?

samth:
I think the note should use `@margin-note{...}` rather that the text "NB".

stamourv:
Switched to use a margin note and merged.

PR #1153, opened 2015-11-29
remove duplicate examples in doc for `arity=?` and `arity-includes?`

philnguyen:
(Each removed example is a duplicate of the one 2 lines above it.)


stamourv:
Merged, thanks!

PR #1155, opened 2015-12-01
Improve dynamic-place examples to include a reference to quote-module-path for use with submodules.

tonyg:


stamourv:
I think it's nice to have an example relatively early in the section, which you're moving to later.
How about leaving the original example where it is, and referring to it when you introduce the second example? You could even add a link to the original.

PR #1159, opened 2015-12-03
Extend `free-vars` to allow reporting module-bound variables.

stamourv:


stamourv:
This is done as a new mode, to not break code that relies on it only returning let- and lambda- bound variables.

samth:
LGTM, though that `if` might be better written as a couple `cond` cases.

stamourv:
Addressed comments and merged.

PR #1162, opened 2015-12-08
Wait more time before of to abort the tests.

juanfra684:
For slow systems without JIT support (like sparc64), 20 minutes is not enough.


stamourv:
LGTM

juanfra684:
I committed a better patch. `(sleep 1800)` will be only used when the JIT is not enabled.

sorawee:
The current code sleeps for 1 hour, so I believe this PR is obsolete. Closing.

PR #1168, opened 2015-12-14
Don't drop expressions with side effects in eq? reduction

gus-massa:
The expression `(eq? x y)` is reduced to `#f` when the types of
`x` and `y` are known. Reduce this to `(begin x y #f)` when they
have side effects.


mflatt:
Looks good to me.

PR #1171, opened 2015-12-15
Use syntax-property in prop:make-rename-transformer example.

LeifAndersen:


stamourv:
I see what you're trying to do, but I think that would work better as a second example. As you wrote it, it gives more visibility to a more advance usage of rename transformers, at the cost of making the simple use cases harder to understand, because there's more stuff in the way. Do you see what I mean?

LeifAndersen:
Okay, I pushed another version.

LeifAndersen:
Just a ping, did you like this newer version better, or is there something else you still didn't like?

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -224,6 +224,46 @@ an identifier, the @racket[exn:fail:contract] exception is raised.
   (slv-lookup slv-inst-2)
 ]
 
+In the above example, the struct is @racket[free-identifier=?] to the renaming
+target identifier.
+If, however, this is not desired, such as if the struct is mutable, this can be
+prevented by setting the @racket['not-free-identifier=?]
+@racket[syntax-property] to @racket[#t] will prevent this.
```
takikawa: There's a duplicate "prevent" here. You should probably just delete "will prevent this". Also maybe this should refer to the `make-rename-transformer` docs since the `'not-free-identifier=?` feature isn't specific to this property. It's a general rename transformer issue that just happens to be relevant to the example.

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -224,6 +224,46 @@ an identifier, the @racket[exn:fail:contract] exception is raised.
   (slv-lookup slv-inst-2)
 ]
 
+In the above example, the struct is @racket[free-identifier=?] to the renaming
```
stamourv: s/renaming//

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -224,6 +224,46 @@ an identifier, the @racket[exn:fail:contract] exception is raised.
...
+target identifier.
+If, however, this is not desired, such as if the struct is mutable, this can be
```
stamourv: Sentence is convoluted. Try a more direct version: Setting the ... property prevents this, which is useful for ..., for example.

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -224,6 +224,46 @@ an identifier, the @racket[exn:fail:contract] exception is raised.
...
+     (begin
+       (define-values (x y) (syntax-local-value/immediate #'box))
+       (set-string-box-str! x (syntax-e #'value))
+       #'(void))]))
+
+(define-syntax current-str (string-box #t))
+(string-lookup current-str)
+(string-set! current-str #f)
+(string-lookup current-str)
+]
```
stamourv: This example is pretty complicated. Do you have anything simpler? Also, whichever example goes there should probably test `free-identifier=?`, IMO.

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
LeifAndersen: I like the idea of linking to `make-rename-transformer`, in fact, we might be able to do it without an example.

PR #1172, opened 2015-12-16
Extend the information in info-lang.

LeifAndersen:
This way people can install different packages based on there system.

### DIFF
--- a/pkgs/racket-doc/scribblings/raco/info.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/info.scrbl
@@ -62,8 +62,8 @@ grammar of @racket[_info-module]:
                 list* reverse append
                 string-append
                 make-immutable-hash hash hash-set hash-set* hash-remove hash-clear hash-update
-                path->string build-path collection-path
-                system-library-subpath]
+                path->string build-path collection-path if equal?
+                system-library-subpath system-type]
 ]
 
 For example, the following declaration could be the @filepath{info.rkt}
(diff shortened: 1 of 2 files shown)


stamourv:
This should be documented and tested.
Also, *their*.

LeifAndersen:
> This should be documented

It is, this commit extends the documentation

> and tested

So, I don't know if there are any unit tests in the code base itself. But I did test it on some of @soegaard 's poppler packages. And it seemed to work just fine.

stamourv:
Re docs: I should have been clearer. What I meant was that the use case for those should be documented. I.e., "If you want a package that ... then use ..."
Re tests: The pkgs tests are at: `pkgs/racket-test/tests/pkg/`

mflatt:
@LeifAndersen It might be ok to add `if`, etc., but I wonder whether you're trying to set up platform-specific dependencies. If so, use the `#:platform` modifier on a dependency, instead of producing a conditional result.

LeifAndersen:
@mflatt Ah, I didn't see the #:platform keyword. Although that doesn't seem to solve all of the problem, because what if I wanted different packages for if I was on a 32 bit vs a 64 bit system? #:platform seems to only give one of 'unix, 'osx, or 'windows.

mflatt:
No, `#:platform` supports more specific platform strings, which is why the `gui-lib` package can call pull in the right Pango implementation for 32-bit Windows versus 64-bit Windows, and so on.

LeifAndersen:
Oh cool, thanks.

PR #1175, opened 2015-12-21
Add 'combinations' and 'in-combinations'

bennn:
New functions for `racket/list`, to accompany `*permutations`.

Besides converting the input list to a vector, `in-combinations` runs in constant space. For time:

```
(define x* (range 25))

(let ([c* (time (in-combinations x*))])
  (time (for ([x c*]) (void))))
;; cpu time: 1 real time: 1 gc time: 0
;; cpu time: 14581 real time: 15424 gc time: 777

(let ([c* (time (in-combinations x* 5))])
      (time (for ([x c*]) (void))))
;; cpu time: 0 real time: 0 gc time: 0
;; cpu time: 2 real time: 2 gc time: 0
```

If you'd rather the implementation was split across 2 functions `(in-combinations n k)` and `(in-powerset n)` I'll happily do that, but I think the name `combinations` is more intuitive. (If there's any way to redirect searches for "powerset" to return "combinations", we should add that.)


stamourv:
Re search terms: Look up `index` from `scribble/base`.

pkgs/racket-doc/scribblings/reference/pairs.scrbl
```diff
@@ -1255,6 +1255,28 @@ Returns a list with all elements from @racket[lst], randomly shuffled.
   (shuffle '(1 2 3 4 5 6))]}
 
 
+@defproc[(combinations [lst list?] [size (or/c #f exact-nonnegative-integer?) #f])
+         list?]{
+
+Return a list of all combinations of the input list.
+If @racket[size] is given, limit results to combinations of with @racket[size] elements.
```
stamourv: s/with//

pkgs/racket-doc/scribblings/reference/pairs.scrbl
```diff
@@ -1255,6 +1255,28 @@ Returns a list with all elements from @racket[lst], randomly shuffled.
   (shuffle '(1 2 3 4 5 6))]}
 
 
+@defproc[(combinations [lst list?] [size (or/c #f exact-nonnegative-integer?) #f])
+         list?]{
+
+Return a list of all combinations of the input list.
```
stamourv: Maybe a like to WP or a quick explanation of "combination"? It's a heavily overloaded term, so it may not be clear with meaning you're using. I guess if you put these docs after `permutations`, that may be clear from context?

pkgs/racket-doc/scribblings/reference/pairs.scrbl
```diff
@@ -1255,6 +1255,28 @@ Returns a list with all elements from @racket[lst], randomly shuffled.
   (shuffle '(1 2 3 4 5 6))]}
 
 
+@defproc[(combinations [lst list?] [size (or/c #f exact-nonnegative-integer?) #f])
```
stamourv: Nitpick: I'd document it as two separate arities, as opposed to using an optional argument. The function really has two separate modes.

racket/collects/racket/list.rkt
```diff
@@ -588,6 +590,96 @@
...
+  (for/list ([x (in-combinations l k)]) x))
+
+;; Generate combinations of the list `l`.
+;; - If `k` is a natural number, generate all combinations of size `k`.
+;; - If `k` is #f, generate all combinations of any size (powerset of `l`).
+(define (in-combinations l [k #f])
+  (unless (list? l)
+    (raise-argument-error 'in-combinations "list?" 0 l))
+  (when (and k (not (exact-nonnegative-integer? k)))
+    (raise-argument-error 'in-combinations "(or #f exact-nonnegative-integer?)" 1 k))
```
stamourv: Related to the arities comment above. I'd just say `exact-nonnegative-integer?` here. There's no point in ever passing `#f` explicitly.

racket/collects/racket/list.rkt
```diff
@@ -588,6 +590,96 @@
...
+  (let* ([v       (list->vector l)]
+         [N       (vector-length v)]
+         [N-1     (- N 1)])
+    (letrec ([gen-combinations
+              (cond
+                [(not k)
+                 ;; Enumerate all binary numbers [1..2**N].
+                 ;; Produce the combination with elements in `v` at the same
+                 ;;  positions as the 1's in the binary number.
+                 (let* ([num-incrs (- (expt 2 N) 1)]
```
stamourv: Internal defines?

racket/collects/racket/list.rkt
```diff
@@ -588,6 +590,96 @@
...
+                        [b*        (make-vector N #f)] ;; Binary counter
+                        [b*-incr
```
stamourv: Wouldn't the code be simpler if you used integers and bitwise ops?

racket/collects/racket/list.rkt
```diff
@@ -588,6 +590,96 @@
...
+                               #f
+                               (begin (set! num-incrs (- num-incrs 1))
+                                      (let loop ([i 0])
+                                        (if (vector-ref b* i)
+                                            (begin (vector-set! b* i #f) (loop (+ i 1)))
+                                            (vector-set! b* i #t))))))]
+                        [b*-num-true
+                         ;; Count the number of #t in the bit vector
+                         (lambda ()
+                           (let loop ([i N-1] [acc 0])
```
stamourv: `for/fold`?

racket/collects/racket/list.rkt
```diff
@@ -588,6 +590,96 @@
...
+                             (if (< i 0)
+                                 acc
+                                 (loop (- i 1) (+ acc (if (vector-ref b* i) 1 0))))))]
+                        [b*->combination
+                         ;; Make a list of `v` elements at positions marked `#t` in `b*`
+                         (lambda ()
+                           (let loop ([i N-1] [acc '()])
```
stamourv: `for/fold`?

rfindler: Just noticed, reading this email trail that `shuffle` has only one call in
the examples? How about duplicating that call so people see it returns
different results each time?

bennn:
Thanks for the feedback, this was really helpful.
(I'd missed `bitwise-bit-set?` earlier; also I didn't realize that `for/fold` is about as fast as `let loop ...`)

Running time is slightly faster now. Here's the average (and standard deviation) for making combinations for `(range 20)`. (Average of 30 runs.)
- before: 150.77ms, (2.77ms)
- after: 143ms, (2.65ms)

stamourv:
Much better. LGTM.
Also, what happened to Robby's comment? Looks like github ate it?

rfindler:
(I was just saying that two calls to shuffle in the examples might be nice.
Since you're in the neighborhood.)

bennn:
@rfindler I added a call to shuffle, not sure if this is what you meant.
https://github.com/racket/racket/commit/80767a84e970dfe695cc859bd8e1489103a61396#diff-a52ea0d2b3f1810e2018c2a43bdf48beR1268

rfindler:
No, I was talking about the docs for shuffle. The top line of the link
there. Don't worry about it.

Robby

bennn:
You just want more shuffle examples?

rfindler:
I shouldn't have brought it up. Sorry for the noise.

stamourv:
@bennn: The answer is yes. :)
I agree that this would be an improvement.

bennn:
Rebased into a single commit. Ready to merge?

LeifAndersen:
I'll merge this, but @stamourv and @rfindler, @bennn doesn't have commit access yet. ;)

LeifAndersen:
Merged, thanks.

stamourv:
@bennn: You should have commit access now.

bennn:
:balloon: :balloon: :cake:

PR #1176, opened 2015-12-23
Refactor streams to use promises and pairs instead of mutable pairs and stream-pares

AlexKnauth:


elibarzilay:
I'm not sure that I follow this completely, but wanted to see that stuff
move to regular promises forever, so I looked...

* Why does it still have an implementation of something that looks just
  like `lazy`?

* `stream-eager`, at least in its original version, was doing something
  that is exactly like `delay/strict`.

* You should be careful about uses of `delay`, since they generally have
  a higher cost, and they don't compose as `lazy` does (which, AFAICT,
  this code still re-implements).

* Another important issue for performance is whether exceptions are
  caught or not.  `racket/promise` does, and it's a high penalty in
  cases where you don't care about catching them.

At least in some point in the past, switching to plain promises would
have made it much shorter.

(It didn't because Matthew based that code on Bewig's stream code
(srfi-42), and (IIRC) Bewig insisted on having a separate type, which is
why he didn't switch to using his own srfi-45.)

AlexKnauth:
The reason I thought it might be better to use `delay` for the elements instead of `lazy` was for people who might want a stream of promises. When they might want get an element out of the stream, but don't want to force the promise yet.

AlexKnauth:
It still has something that looks like `lazy` because the rest needs to be a stream, not a promise of a stream. A stream could use promises instead of having a `forced?` field with the second field as either the lambda or the value depending on that, but that might be unnecessary higher cost? I'm not sure.

AlexKnauth:
Should there be a lighter-weight version of `delay` that doesn't try to handle multiple values or store exceptions? Would that help this?

elibarzilay:
Re your first two comments: what I wanted to see is a simple type that
gets used uniformly, so streams are just lazy lists --- where just the
list spine is lazy, so the values are used as-is.  IOW, `stream-cons` is
just `(lazy (cons ...))`.  If there is some need to delay the values too
then yes, `delay` is the right thing.  And if there's a need to
distinguish such values from other promises, then a new `delay/...`
struct can be used just for that, but otherwise use the same forcing as
`lazy`.


> Should there be a lighter-weight version of `delay` that doesn't try
> to handle multiple values or store exceptions?

That would be *really* nice if there's a good way to do that.  Where
"good" means that you get something like `lazy*` where it plays nicely
with `lazy` both ways so you can interchange them freely.  (I think that
I started to do that at some point, but got lazy.)


> Would that help this?

That depends on the kind of exception catching that the current code
does.  If it's still similar enough to Bewig's code, then it would help
since he never bothered catching exceptions in either this or in
srfi-45.

(As a side note, the consistency that results from catching exceptions
is that `(lazy (/ 2 (random 2)))` is always going to force to 1 or to an
error, but not change; or

    (let ([c 0]) (delay (begin0 (/ 9 c) (set! c (add1 c)))))

for a less random example.  Yet another variant of `delay`, if you want
to stick your head fully into that bed, is one that would do more work,
and capture parameters too.  Something like that would be very useful
too (for different reasons, of course).)

AlexKnauth:
> On Dec 26, 2015, at 7:40 PM, Eli Barzilay <notifications@github.com> wrote:
> 
> Re your first two comments: what I wanted to see is a simple type that
> gets used uniformly, so streams are just lazy lists --- where just the
> list spine is lazy, so the values are used as-is. IOW, stream-cons is
> just (lazy (cons ...)). If there is some need to delay the values too
> then yes, delay is the right thing.
> 

So then `(stream-cons a b)` could be `(lazy (cons (delay a) b))`, except that streams and promises are not supposed to be the same thing. 

So `(make-stream (lazy (cons (delay a) b)))`. But if you try to call stream-first on that, it will force it and evaluate the `b` expression, which is not what we want, I think.

So then it could be `(make-stream (lazy (cons (delay a) (lazy b)))`, except that `stream-rest` needs to return a stream. 

So `(make-stream (lazy (cons (delay a) (make-stream (lazy b))))`. But then, if the b expression will probably be a stream itself, there will be a `(make-stream (lazy ...))` nested inside that in place of `b`. Using `lazy` here doesn't actually help, because of the extra `make-stream` in between. So that's why we would have to reimplement the nested forcing that `lazy` does anyway. 

Instead of `(make-stream (lazy ...))`, this does `(make-stream #false (lambda () ...))`. 

> (As a side note, the consistency that results from catching exceptions
> 
> is that (lazy (/ 2 (random 2))) is always going to force to 1 or to an
> error, but not change;
> 

Oh, that's interesting:

(define s (stream (/ 2 (random 2))))
> (stream-first s)
. . /: division by zero
> (stream-first s)
. . /: division by zero
> (stream-first s)
. . /: division by zero
> (stream-first s)
2

sorawee:
@AlexKnauth This file has been refactored so many times now that it diverges pretty significantly from the base of your PR. Do you still wanna leave the PR open?

PR #1177, opened 2015-12-23
Better printing for streams

AlexKnauth:
See https://groups.google.com/forum/#!topic/racket-dev/PPGMnd7jKBM

Relies on https://github.com/racket/racket/pull/1176


AlexKnauth:
This doesn't handle cycles of length greater than one yet; I'll do that next.

PR #1182, opened 2015-12-28
add a lighter-weight version of delay

AlexKnauth:
that doesn’t deal with multiple values or exceptions

P.S. Where are the tests for promises?

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/promise.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/promise.scrbl
@@ -173,3 +173,18 @@ TODO: Say something on:
   demand.
 ;}
 }
+
+@defform[(delay* body ...+)]{
+Like @racket[delay], except that it only handles one value, and it
+doesn't deal with exceptions.
+
+If the last @racket[body] expression either produces more than one
+value, forcing it for the first time will raise a result arity
+mismatch error, and subsequent forces will raise a "reentrant promise"
+error.
+
+If it raises an exception, forcing it for the first time will raise
+that exception, and subsequent forces will raise a reentrant promise
+error.
+}
+
--- a/racket/collects/racket/promise.rkt
+++ b/racket/collects/racket/promise.rkt
@@ -212,3 +212,29 @@
                     (cons '#:tick       #'0.2)
                     (cons '#:use        #'0.12))])
     (λ(stx) (make-delayer stx #'delay/idle kwds))))
+
+;; A lighter-weight version of delay that doesn't care about multiple values or
+;; exceptions
+
+;; force/delay* : Delay*-Promise -> Any
+;; Like force/generic from racket/private/promise, except that it
+;; only deals with one value and doesn't care about exceptions
+;; I'm not sure whether the common code could be abstracted out or not
+(define (force/delay* promise)
+  (let ([v (pref promise)])
+    (cond [(pair? v)
+           (car v)]
+          [(procedure? v)
+           (pset! promise (make-running (object-name v)))
+           (let ([v (v)])
+             (pset! promise (list v))
+             v)]
+          [else (error 'force "delay* promise with invalid contents: ~e" v)])))
+
+(define-struct (delay*-promise promise) ()
+  #:property prop:force force/delay*)
+
+(provide (rename-out [delay** delay*]))
+(define delay* make-delay*-promise)
+(define-syntax (delay** stx) (make-delayer stx #'delay* '()))
+
--- a/pkgs/racket-test/tests/racket/promise.rkt
+++ b/pkgs/racket-test/tests/racket/promise.rkt
@@ -182,6 +182,37 @@
           (force p) => 99
           (format "~a" p) => "#<promise!99>")))
 
+(define (test-delay*)
+  ;; normal single value
+  (let ([i 0])
+    (define p (delay* (set! i (add1 i)) i))
+    (test i => 0
+          (force p) => 1
+          i => 1
+          ;; the body is never evaluated more than once
+          (force p) => 1
+          i => 1))
+  ;; multiple values
+  (let ([i 0])
+    (define p (delay* (set! i (add1 i)) (values 1 2)))
+    (test i => 0
+          (force p) =error> "result arity mismatch"
+          i => 1
+          ;; this is what happens when the exception isn't caught
+          (force p) =error> "reentrant promise"
+          ;; but it still doesn't evaluate the body more than once
+          i => 1))
+  ;; exceptions
+  (let ([i 0])
+    (define p (delay* (set! i (add1 i)) (error "I'm an error")))
+    (test i => 0
+          (force p) =error> "I'm an error"
+          i => 1
+          ;; this is what happens when the exception isn't caught
+          (force p) =error> "reentrant promise"
+          ;; but it still doesn't evaluate the body more than once
+          i => 1)))
+
 (provide promise-tests)
 (module+ main (promise-tests))
 (module+ test (promise-tests))
@@ -195,6 +226,7 @@
         do (test-delay/strict)
         do (test-delay/sync)
         do (test-delay/thread)
+        do (test-delay*)
         ;; misc tests
         (let ([x (lazy (delay 1))]) (force x) (force x)) => 1
         ))
--- a/pkgs/racket-test/tests/racket/promise.rkt
+++ b/pkgs/racket-test/tests/racket/promise.rkt
@@ -1,0 +1,200 @@
+#lang racket/base
+
+;; Tests for the various racket promises
+
+(require racket/promise tests/eli-tester (for-syntax racket/base))
+
+;; check that things are `promise?'s or not
+(define (test-types)
+  (for ([v (list 1 '(1) (lambda () 1))])
+    (test (promise? v) => #f))
+  (for ([v (list (delay 1) (lazy 1) (delay (delay 1)) (lazy (lazy 1)))])
+    (test (promise? v) => #t)))
+
+(define (test-syntax)
+  (test (delay) =error> "bad syntax"
+        (lazy)  =error> "bad syntax"
+        (delay #:foo 1 2) =error> "bad syntax"
+        (force (delay/thread #:group #f)) =error> "bad syntax"
+        (force (delay/thread #:group #f 1)) => 1
+        (force (delay/thread 1 #:group #f 2)) => 2
+        (force (delay/thread #:groupie #f 1)) =error> "bad syntax"))
+
+;; basic delay/lazy/force tests
+(define (test-basic-promises)
+  (define thunk1 (lambda () 1))
+  (define promise1 (delay 1))
+  (define ? #f)
+  ;; test a few different values
+  (define-syntax-rule (t (f x ...))
+    (begin (set! ? 1)      (test (f x ...) => ?)
+           (set! ? '())    (test (f x ...) => ?)
+           (set! ? '(1))   (test (f x ...) => ?)
+           (set! ? thunk1) (test (f x ...) => ?)))
+  (define-syntax-rule (t* (f x ...))
+    (begin (t (f x ...)) (set! ? promise1) (test (f x ...) => ?)))
+  ;; `force' is identity for non-promises
+  (t (force ?))
+  ;; basic checks that `delay' works as expected with all kinds of values
+  (t* (force (delay ?)))
+  (t* (force (force (delay (delay ?)))))
+  (t* (force (delay (force (delay ?)))))
+  ;; basic checks that `lazy' works as expected with all kinds of values
+  (t (force (lazy ?)))
+  (t (force (lazy (lazy ?))))
+  (t (force (force (lazy (lazy ?)))))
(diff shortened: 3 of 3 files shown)


pkgs/racket-doc/scribblings/reference/promise.scrbl
```diff
@@ -173,3 +173,9 @@ TODO: Say something on:
   demand.
 ;}
 }
+
+@defform[(delay* body ...+)]{
+Like @racket[delay], except that it only handles one value, and it
+doesn't deal with exceptions.
```
samth: This should specify what happens if the `body` expressions produce more than one value, or if they raise an exception.

racket/collects/racket/promise.rkt
```diff
@@ -212,3 +212,27 @@
...
+
+;; A lighter-weight version of delay that doesn't care about multiple values or
+;; exceptions
+
+;; force/delay* : Delay*-Promise -> Any
+;; only deals with one value and doesn't care about exceptions
+(define (force/delay* promise)
+  (let ([v (pref promise)])
+    (cond [(pair? v)
+           (unsafe-car v)]
```
samth: This use of `unsafe` is not needed.

pkgs/racket-doc/scribblings/reference/promise.scrbl
AlexKnauth: Ok done.

racket/collects/racket/promise.rkt
AlexKnauth: Fixed now. I had used unsafe because the [similar code](https://github.com/AlexKnauth/racket/blob/e08188aeda8c419b741313ef34ef0e49ac6eb6a4/racket/collects/racket/private/promise.rkt#L83) in `private/promise.rkt` used unsafe. Should I change that too?

rfindler:
`reify-result` has an `unsafe-cdr` and it is exported. Seems dangerous and worth changing.

AlexKnauth:
It's not really unsafe if it checks `pair?` first though, is it?

pkgs/racket-doc/scribblings/reference/promise.scrbl
```diff
@@ -173,3 +173,14 @@ TODO: Say something on:
...
+
+If the last @racket[body] expression either produces more than one
+value or raises an exception, forcing it for the first time will
+raise an exception, and subsequent forces will raise a "reentrant
```
samth: Can you make it more specific about what exception is raised in each case, the way the rest of the docs do?

samth:
@rfindler I agree that there are a bunch of no-longer-needed `unsafe` calls in this library, but I think that's a separate change from this PR.

pkgs/racket-doc/scribblings/reference/promise.scrbl
```diff
@@ -173,3 +173,18 @@ TODO: Say something on:
...
+doesn't deal with exceptions.
+
+If the last @racket[body] expression either produces more than one
+value, forcing it for the first time will raise a result arity
+mismatch error, and subsequent forces will raise a "reentrant promise"
+error.
+
+If it raises an exception, forcing it for the first time will raise
+that exception, and subsequent forces will raise a reentrant promise
+error.
```
AlexKnauth: @samth Is this better?

rfindler:
Oh, yes, sorry: I missed the `pair?` check (despite looking multiple times!)

rfindler:
(But I will note that the optimizer transforms car into unsafe-car in this situation already)

rfindler:
And yes, it is separate.

AlexKnauth:
Where are the tests for promises? I can't find them. I can only find the tests in the `lazy` repository. Should I make a pull request there, or should some of those tests be moved to this repository with the other racket tests? Or?

samth:
Yes, it looks to me like https://github.com/racket/lazy/blob/master/tests/promise.rkt should be in this repository.

AlexKnauth:
Should that be part of this pull request? If so, where should I put it?

AlexKnauth:
(In reply to https://github.com/AlexKnauth/racket/commit/69d965a9cb6b7c2bd0a0b442e18bc427d3da668a#commitcomment-15221712)

Ok done. 

However, there's a `link: module mismatch; possibly, bytecode file needs re-compile because dependencies changed` error that I keep getting. At first I thought it would go away like those kinds of errors usually do, but it didn't, and it consistently shows up, even in the travis build.

Any idea what could be causing it?
```
. ../../../../racket/collects/racket/promise.rkt:237:49: link: module mismatch;
 possibly, bytecode file needs re-compile because dependencies changed
  importing module: 'promise
  exporting module: "/Users/Alex/Documents/GitHub/AlexKnauth/racket/racket/collects/racket/promise.rkt"
  exporting phase level: 0
  internal explanation: variable not provided (directly or indirectly) in: delay*
```

AlexKnauth:
Ok, Matthew Flatt found that bug and fixed it, and the tests are now passing.

AlexKnauth:
And now the build is passing since it was fixed on master.

elibarzilay:
@AlexKnauth: one issue is that this duplicates code from the generic force in the private file -- there should be a comment saying that, but it would be much better to move this thing into the private file so both variations can share code.

--BUT--

This is much less important than a cheap *`lazy`* variation, because that where the composition happens.  Having a `lazy*` would make it possible to have nestings of both kinds of promises (`lazy` and `lazy*`) where you use the latter if you know that no exceptions will happen or if you don't care about them.

In fact, I might be that a `delay*` is almost useless, since (IIRC, I'm half guessing based on vague memories) a `delay` could be implemented with a `lazy` wrapped around a `delay/strict`, but kept since it's a name that most people expect.  Assuming that this is true, then a `delay*` is useless since it can be implemented in a similar way, and there's no legacy name to keep in that case.  (But maybe it's still useful to have it.)

As for the implementation, I think that the best plan I had was something like a `composable-promise*` struct which is a sibling of `composable-promise`, then implement `force/composable*` as a copy of `force/composable` except for the exception catching, and then extend both so that each of them calls the other when it gets an instance of the other.  The problem is that `force/composable` has a loop that is supposed to do the flattened kind of forcing through nested (composable) promises, and each of these two variants needs to do the same.

(Guesstisketched code follows.)

So, first change the structs so that `composable-promise` inherits from `composable-promise*` (which inherits from `promise`):

    (define-struct promise ([val #:mutable])
      #:property prop:custom-write promise-printer
      #:property prop:force force/generic)
    ;; Then, a subtype for composable promises that don't catch exceptions
    (define-struct (composable-promise* promise) ()
      #:property prop:force force/composable*)
    ;; And then regular composable promises
    (define-struct (composable-promise composable-promise*) ()
      #:property prop:force force/composable)

Then, `force/composable` is mostly the same.  In the flat loop it checks for `composable-promise?` which catches both kinds -- it doesn't matter since the loop is already in a handler.  The code is the same, with just one new comment:

    (define (force/composable root)
      (let ([v (pref root)])
        (cond
          [(procedure? v)
           ;; mark the root as running: avoids cycles, and no need to keep banging
           ;; the root promise value; it makes this non-r5rs, but the only
           ;; practical uses of these things could be ones that use state to avoid
           ;; an infinite loop.  (See the generic forcer below.)
           ;; (careful: avoid holding a reference to the thunk, to allow
           ;; safe-for-space loops)
           (pset! root (make-running (object-name v)))
           (call-with-exception-handler
            (lambda (e) (pset! root (make-reraise e)) e)
            (lambda ()
              ;; iterate carefully through chains of composable promises
              (let loop ([v (v)]) ; does not handle multiple values!
                (cond [(composable-promise? v)
                       ;; this includes `composable-promise*'s since we're
                       ;; already in a handler anyway
                       (let ([v* (pref v)])
                         (pset! v root) ; share with root
                         (cond [(procedure? v*) (loop (v*))]
                               ;; it must be a list of one value (because
                               ;; composable promises never hold multiple values),
                               ;; or a composable promise
                               [(pair? v*) (pset! root v*) (unsafe-car v*)]
                               ;; note: for the promise case we could jump only to
                               ;; the last `let' (for `v*'), but that makes the
                               ;; code heavier, and runs slower (probably goes over
                               ;; some inlining/unfolding threshold).
                               [else (loop v*)]))]
                      ;; reached a non-composable promise: share and force it now
                      [(promise? v) (pset! root v) (force v)]
                      ;; error here for "library approach" (see above URL)
                      [else (pset! root (list v)) v]))))]
          ;; try to make the order efficient, with common cases first
          [(pair? v) (if (null? (unsafe-cdr v)) (unsafe-car v) (apply values v))]
          ;; follow all sharings (and shortcut directly to the right force)
          [(composable-promise*? v) (force/composable* v)]
          [(composable-promise? v) (force/composable v)]
          [(null? v) (values)]
          [(promise? v) (force v)] ; non composable promise is forced as usual
          [else (error 'force "composable promise with invalid contents: ~e" v)])))

And finally, `force/composable*` is kind of similar, except that it doesn't catch exceptions.  The problem is that the flat loop might encounter a `composable-promise` so in that case do catch exceptions, and arrange for it to happen just once:

    ;; `force/composable*' is similar to `force/composable' ignoring exceptions,
    ;; unless running into a `composable-promise`
    (define (force/composable* root)
      ;; same as `force/composable*', except that...
      (let ([v (pref root)])
        (cond
          [(procedure? v)
           (pset! root (make-running (object-name v)))
           ;; there is no exception handler here
           (let loop ([v (v)] [wrapped? #f])
             (cond [(composable-promise? v)
                    (if (or wrapped? (composable-promise*? v))
                      (let ([v* (pref v)])
                        (pset! v root)
                        (cond [(procedure? v*) (loop (v*) wrapped?)]
                              [(pair? v*) (pset! root v*) (unsafe-car v*)]
                              [else (loop v* wrapped?)]))
                      ;; but if we see a first non-* promise, wrap in a handler
                      (call-with-exception-handler
                       (lambda (e) (pset! root (make-reraise e)) e)
                       (lambda () (loop v #t))))] ; no need for nested wrappers
                   [(promise? v) (pset! root v) (force v)]
                   [else (pset! root (list v)) v]))]
          [(pair? v) (if (null? (unsafe-cdr v)) (unsafe-car v) (apply values v))]
          [(composable-promise*? v) (force/composable* v)]
          [(composable-promise? v) (force/composable v)]
          [(null? v) (values)]
          [(promise? v) (force v)]
          [else (error 'force "composable promise with invalid contents: ~e" v)])))

It's tempting to abstract both things in a single piece of code, but I have no idea if this can work or not.  (Performance is important here, and it's full of tail-context traps.)

"Or something™."

AlexKnauth:
> On Jan 5, 2016, at 6:23 AM, Eli Barzilay <notifications@github.com> wrote:
> This is much less important than a cheap lazy variation, because that where the composition happens. Having a lazy* would make it possible to have nestings of both kinds of promises (lazy and lazy*) where you use the latter if you know that no exceptions will happen or if you don't care about them.
> 
That would be better, but I wanted to see what was involved in the simpler version first. 
> In fact, I might be that a delay* is almost useless, since (IIRC, I'm half guessing based on vague memories) a delay could be implemented with a lazy wrapped around a delay/strict, but kept since it's a name that most people expect. Assuming that this is true, then a delay* is useless since it can be implemented in a similar way, and there's no legacy name to keep in that case. (But maybe it's still useful to have it.)
> 

One of the uses that I was thinking of was for the implementation of streams. I'm not sure whether the elements or rests of streams should use this by default, but if catching exceptions and multiple values becomes enough of a problem, there could be a version of streams that used `delay*`. 

(Streams don't catch exceptions right now, but should that be fixed?)

There would also be uses where you know the immediate evaluation doesn't raise an exception, and it could be a safe optimization. Even if somewhere there is an exception, it would be fine if it's wrapped in one normal delay, with everything else outside of that being delay*. That's assuming that they return something that might contain the promise, instead of forcing it.

An implementation in terms of delay/strict would be a bit slower because there's an extra struct wrapped around it, right?

racket/collects/racket/promise.rkt
```diff
@@ -212,3 +212,29 @@
...
                     (cons '#:use        #'0.12))])
     (λ(stx) (make-delayer stx #'delay/idle kwds))))
+
+;; A lighter-weight version of delay that doesn't care about multiple values or
+;; exceptions
+
+;; force/delay* : Delay*-Promise -> Any
+;; Like force/generic from racket/private/promise, except that it
+;; only deals with one value and doesn't care about exceptions
+;; I'm not sure whether the common code could be abstracted out or not
```
AlexKnauth: (In reply to the first bit of https://github.com/racket/racket/pull/1182#issuecomment-168975715)
@elibarzilay I'm not sure how any common code could be abstracted out. Is this the kind of thing you meant? Or is there any way you can see to abstract anything out?

samth:
How much lighter-weight is this? And can you test that on both BC and CS?

sorawee:
> One of the uses that I was thinking of was for the implementation of streams.

`stream-first` can return multiple values though, right?

AlexKnauth:
Ah, I hadn't considered that. This might be less useful than I thought.

PR #1183, opened 2015-12-29
Add for/stream and for*/stream comprehensions to racket/stream

### LINKED ISSUE #664: no "for/stream"
stream lacks its own for form, to match for/list, for/vector, for/hash, and for/set

lexi-lambda:
This adds `for/stream` and `for*/stream`, which implement lazy stream comprehensions. They are ported almost directly from my collections library’s implementation of `for/sequence`.

Closes #664.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sequences.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sequences.scrbl
@@ -1018,6 +1018,26 @@ stream, but plain lists can be used as streams, and functions such as
   new stream is constructed lazily.
 }
 
+@deftogether[(@defform[(for/stream (for-clause ...) body-or-break ... body)]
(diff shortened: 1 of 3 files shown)


bennn:
Why `for/stream` instead of `for/sequence`?

lexi-lambda:
The name `for/stream` makes considerably more sense in the context of Racket’s other for loops—they’re all named after the concrete data structure they produce, so we have `for/hash` instead of `for/dict`, for example. Plus, Racket “sequences” are a very general, fairly vague notion, so I think the name would be a poor description of what this actually does. Streams are lazy, which is fairly unique to streams within the Racket ecosystem of data structures, so communicating that is important.

stamourv:
LGTM

lexi-lambda:
Merged in f52d43e600acf91839106299c2f53534266f7cfe.

PR #1184, opened 2015-12-29
more *syntax/loc for match forms

andmkent:
I was using match-let and got a syntax error that pointed to this file. After changing the match-let definition to use syntax/loc (line 124) the error pointed to the exact spot causing the problem. Yay!

I changed quite a few vanilla syntax-quotes to the *syntax/loc form... perhaps some do not need to? I'm not sure.


samth:
Looks good to me.

andmkent:
I added the nested syntax/loc you mentioned (in person).

samth:
Looks good.

PR #1188, opened 2016-01-02
Extend `random` to work with ranges and sequences.

stamourv:
Requested by Matthias.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/numbers.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/numbers.scrbl
@@ -842,13 +842,29 @@ both in binary and as integers.
                     [rand-gen pseudo-random-generator?
                                (current-pseudo-random-generator)])
             exact-nonnegative-integer?]
+           [(random [min (integer-in 1 4294967087)]
+                    [max (integer-in 1 4294967087)]
+                    [rand-gen pseudo-random-generator?
+                              (current-pseudo-random-generator)])
+            exact-nonnegative-integer?]
+           [(random [seq sequence?]
+                    [rand-gen pseudo-random-generator?
+                              (current-pseudo-random-generator)])]
            [(random [rand-gen pseudo-random-generator?
                               (current-pseudo-random-generator)]) 
             (and/c real? inexact? (>/c 0) (</c 1))])]{  
 
 When called with an integer argument @racket[k], returns a random
-exact integer in the range @racket[0] to @math{@racket[k]-1}. When
-called with zero arguments, returns a random inexact number between
+exact integer in the range @racket[0] to @math{@racket[k]-1}.
+
+When called with two integer arguments @racket[min] and @racket[max], returns a
+random exact integer in the range @racket[min] to @math{@racket[max]-1}.
+
+When called with a sequence argument @racket[seq], returns a random element of
+the sequence. Like @racket[sequence-length], does not terminate on infinite
+sequences, and extracts elements up to the returned element.
+
+When called with zero arguments, returns a random inexact number between
 @racket[0] and @racket[1], exclusive.
 
 In each case, the number is provided by the given pseudo-random number
--- a/pkgs/racket-test-core/tests/racket/number.rktl
+++ b/pkgs/racket-test-core/tests/racket/number.rktl
@@ -2449,19 +2449,29 @@
 (test (begin (random-seed 23) (list (random 10) (random 20) (random 30)))
       'random-seed-same
       (begin (random-seed 23) (list (random 10) (random 20) (random 30))))
+(test (begin (random-seed 23) (list (random 10 20) (random 20 30) (random 30 40)))
+      'random-seed-same2
+      (begin (random-seed 23) (list (random 10 20) (random 20 30) (random 30 40))))
+(test (begin (random-seed 23) (list (random '(1 2 3)) (random '(4 5 6)) (random '(7 8 9))))
+      'random-seed-same3
+      (begin (random-seed 23) (list (random '#(1 2 3)) (random '#(4 5 6)) (random '#(7 8 9)))))
+(test (begin (random-seed 23) (list (random '#(1 2 3)) (random '#(4 5 6)) (random '#(7 8 9))))
+      'random-seed-same4
+      (begin (random-seed 23) (list (random '#(1 2 3)) (random '#(4 5 6)) (random '#(7 8 9)))))
 (arity-test random-seed 1 1)
-(arity-test random 0 2)
+(arity-test random 0 3)
 (err/rt-test (random-seed "apple"))
 (err/rt-test (random-seed 4.5))
 (err/rt-test (random-seed -1))
 (err/rt-test (random-seed (expt 2 31)))
 (err/rt-test (random-seed big-num))
-(err/rt-test (random "apple"))
+(err/rt-test (random 'apple))
 (err/rt-test (random 0))
 (err/rt-test (random -6))
 (err/rt-test (random 4294967088))
 (err/rt-test (random (expt 2 32)))
 (err/rt-test (random big-num))
+(err/rt-test (random 10 5))
 
 (random-seed 101)
 (define x (list (random 10) (random 20) (random 30)))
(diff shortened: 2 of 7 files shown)


samth:
The extension to sequences seems like it should be in a separate function related to sequences, especially because many sequences could implement optimized versions (such as for vectors).

Also, does this change the performance of using `random` from the current behavior?

stamourv:
Re separate function for sequences: I thought about that. But since `random` was already doing double (and now triple) duty, I figured may as well put sequences there too.
Re optimized versions: All my version does is take `sequence-ref` and `sequence-length`, both of which will be specialized for, e.g., vectors. So I doubt it would make much difference.
Re performance: I just ran some quick tests, and there's no observable difference between the two when taking 10M random floats, and about 5% overhead when taking 10M integers.

rfindler:
Did you consider a new function named `random-ref`? (Not sure which I like better.)

I think there is a lot of useful pieces of random-related functionality spread around in various places (the math library has some). 

Should we collect that into `racket/random`?

bennn:
+1 for `random-ref`, with optional arguments for the number of refs to make and whether sampling should be done without replacement.

stamourv:
Not 100% sold on the name `random-ref`, but can't think of anything much better. (`random-sample`? meh.)

@bennn: Those would be nice additions, but getting close to `math/distributions`. It's a pretty simple utility, though, so the overlap is probably not an issue.

@rfindler: Re `racket/random`: Aside from the math library stuff (which is probably better to leave there, because of the TR dependency), randomness-related functions all seem to be in `racket/base` (or already in `racket/random`). I agree that the new `random-ref` (or whatever we call it) would belong there.

For now, I'll add range support to `random`, and add a new `random-ref` for sequences, with @bennn's suggestions.

stamourv:
Actually, thinking about it more, @bennn's suggestions should be a different function. `random-ref` should return a single element, whereas what @bennn suggests should return a list. I'll implement both, and call the second `random-sample`.

gus-massa:
Perhaps I'm missing some detail, but the implementation of `sequence-length` in [for.rkt/R2035](https://github.com/stamourv/racket/commit/f187ae9408199789f0df40fb183b61225fe1d260#diff-c25597d0d3581a6dcf7a184ec78a24a6R2035) looks like O(N), for example ` (sequence-length (make-vector 1000000))`.

In [pre-base.rkt/R131](https://github.com/stamourv/racket/commit/f187ae9408199789f0df40fb183b61225fe1d260#diff-13eaec578e818e85ad4749dc69a4dd57R131) should it say "either *max* and prng" instead of "either *min* and prng"?

Also, as an easy test, I copied the new function to Racket 6.3

    #lang racket
    ...

    (define-values (-random) ; more featureful than #%kernel's `random`
      ...)

    (collect-garbage)
    (time (for ([i (in-range 10000000)])
            (random 7)))
    (collect-garbage)
    (time (for ([i (in-range 10000000)])
            (-random 7)))

and I got

    cpu time: 13417 real time: 18114 gc time: 0
    cpu time: 17488 real time: 23214 gc time: 0

stamourv:
@gus-massa

Re `sequence-length`: Oh, that's disappointing. I assumed it was doing something smarter than that. I'll file a PR to improve it.

Re timings: Oh, that's more overhead than I observed. With the sequence stuff out of the way, I think I can reduce some of it. I'll report back.

stamourv:
Ok, with the new version, I get 1414 vs 1261 on 6.3. About 12%.

rfindler:
That seems worth trying to find a new name to me. No?

stamourv:
With `begin-encourage-inline` (and on HEAD as opposed to 6.3), it's down to 6%.

I'm not too worried about that level of overhead for `random`. When I started the FSP work, I looked for programs where `random` was responsible for a significant portion of the running time, and did not find any (even for programs that used it heavily).

I'm more inclined to go with the nicer API. IMO, since `range` already has the max-only case, it should also have the max+min case, but I'm open to other opinions.

rfindler:
That seems okay with me. I'm not following the connection to `range`, tho.

stamourv:
Er, that should have been `random`.

I'll merge.

gus-massa:
FYI: Perhaps this post is too late, but I have mostly good news: In the simple case the overhead is reduced during the optimization step.

For this test, I used the definition of `-random` of the initial commit, but I think the result applies to the last version too, especially with the use of `begin-encourage-inline`. 

With this program:

    (for ([i (in-range 30)])
      (-random 5)) ; <--extended version

After decompiling and simplifying the output, I got this program (all the `fx` operations actually are `unsafe`):

    (print-values
     (let ([for-loop ?])
       (#%set!-rec for-loop (lambda (r n)
                              (if (fx< n 30)
                                  (begin
                                    (random 5)
                                    (let ([n (fx+ n 1)])
                                      (if (fx< n 30)
                                        (begin
                                          (random 5)
                                          (let ([n (fx+ n 1)])
                                            (if (fx< n 30)
                                              (begin
                                                (-random 5) ; <-----
                                                (let ([n (fx+ n 1)])
                                                  (if (fx< n 30)
                                                    (begin
                                                      (-random 5) ; <-----
                                                      (for-loop #<void> (fx+ n 1)))
                                                    #<void>)))
                                              #<void>)))
                                        #<void>)))
                                r))
         (random 5)
         (random 5)
         (random 5)
         (random 5)
         (random 5)
         (random 5)
         (-random 5) ; <-----
         (-random 5) ; <-----
         (for-loop #<void> 8)))

So the optimizer reduces most of the uses of `-random` to `random` and in those cases there is no overhead. The problem is that the optimizer unrolls the loop and it has no enough fuel to inline the lasts uses of `-random`. I saw a similar behavior before, but I don't know how to fix it, since the simple solution of increasing the `inline_fuel` would increase the number of unrolls, but would not inline the lasts `-random`.

stamourv:
@gus-massa: Awesome! Thanks for investigating.

elibarzilay:
@stamourv: won't this break with sequences that don't like to be used more than once unless you're doing the trick-whose-name-I-forget for a single pass?  Something like the (very tempting) use case of a n `in-lines` sequence which would be a quick way to get a random line...

stamourv:
@elibarzilay: Yes, as much as `sequence-length` would. Not sure there's anything we can do about it. It's documented, though.

elibarzilay:
@stamourv, well, `sequence-length` by itself won't break, it's the assumption that the sequence is always fine after using it.  IMO, this is a bigger caveat than getting stuck on an infinite seq (which is the only thing I saw in the docs, but TBH I skimmed through it quickly).

The thing that I was referring to is the known way to choose a random element from a sequence of unknown length (such as an input stream): scan the sequence where you choose the first one, then at a 1/2 chance switch to the second, 1/3 to switch to the third, etc.  Overall you're paying for using many more random numbers and you're getting a single scan in return -- and when your input is a general sequence this seems to me like a good tradeoff.  (But I really like that random selection thing, which is why your commit stood out: I immediately wanted to see if you can do that...)

(BTW, the warning about infinite sequences would still apply, since it still needs to do a full scan.)

stamourv:
@elibarzilay: You're right, the breakage is slightly different. I'll update the docs.
As for single-traversal sampling: That wouldn't give a uniform distribution, and I don't think that having different distributions for different kinds of sequences is a good idea.

elibarzilay:
No, it's uniform.  A bit more sensitive to bad random functions, but that's not a problem in this case.  Run the code below, which is actually a special case of [this thing](http://rosettacode.org/wiki/Knuth%27s_algorithm_S#Racket):

    #lang racket
    (define (rand-elt l)
      (let loop ([n 1] [l l] [x #f])
        (if (null? l) x
            (loop (add1 n) (cdr l) (if (zero? (random n)) (car l) x)))))
    (define (test n runs)
      (define l (build-list n values))
      (define counts (make-vector n 0))
      (for ([i runs])
        (define x (rand-elt l))
        (vector-set! counts x (add1 (vector-ref counts x))))
      (for/list ([c (in-vector counts)]) (/ (round (* 10000 (/ c runs))) 100.0)))
    (for ([i (in-range 1 11)]) (eprintf "~s -> ~s\n" i (test i 10000000)))

stamourv:
Oh, interesting. I didn't know about that.
Sure, I'll switch the implementation to use that.

PR #1191, opened 2016-01-05
More O(1) special cases for  sequence-length

gus-massa:
I added more O(1) special cases for `sequence-length`. It doesn't have test cases, because I didn't find where to write them.

I also tried to add O(1) special cases for `sequence-ref`, but I really don't like the final code. It's too complex. So I'd prefer not to merge the second commit.

### DIFF
--- a/racket/collects/racket/sequence.rkt
+++ b/racket/collects/racket/sequence.rkt
@@ -2,6 +2,8 @@
 
 (require "stream.rkt"
          "private/sequence.rkt"
+         "fixnum.rkt"
+         "flonum.rkt"
          racket/contract/combinator
          racket/contract/base
          (for-syntax racket/base)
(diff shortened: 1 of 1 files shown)


racket/collects/racket/sequence.rkt
```diff
@@ -39,32 +41,56 @@
 (define (sequence->list s)
   (for/list ([v s]) v))
 
-(define (sequence-length s)
-  (unless (sequence? s) (raise-argument-error 'sequence-length "sequence?" s))
-  (cond [(list? s) (length s)]
+(define (sequence-length/o1 s)
```
stamourv: Would it make sense to bail out early when `(struct? s)`. Otherwise, sequence structs (which I'd expect to be a common case) will have to go through all these tests before falling back to the general method. For short sequences, that may matter.

samth: But `(struct? s)` will be #f for most of those, since the structs will be opaque, won't they?

stamourv: Bleh. You're right.
/me shakes fist at opaque structs.

gus-massa:
I made a few tests, and the new version of `sequence-length` is 20% slower than the old version. I thought that the overhead of the `for` would cover all the differences. I'm using:

    (define s (sequence-tail "xx1234567890" 2))
    #;(define s (in-range 10))

    (collect-garbage)
    (time (for ([i (in-range 1000000)])
            (sequence-length s)))

stamourv:
20% is a lot, but:
- the overhead is fixed, so it will matter less for longer sequences
- taking the length of short sequences will be quick no matter what, so 20% more may not matter that much.

gus-massa:
Do I rebase and merge the first commit?

stamourv:
Fine by me.

gus-massa:
Done.

I'll left the O(1) cases of `sequence-ref` for the future. Hopefuly I or omeone else can implement it elegantly.

PR #1197, opened 2016-01-07
Fix bug in shifting of types during inlining

gus-massa:
I think this fix the crash that appeared in https://github.com/racket/racket/pull/1076. I also found an example that shows a bug in 6.3.

During inlining, the type information gathered in code that was inside the lambda is copied to the outer context. But the coordinates of the type information were shifted in the wrong direction, so the type was
assigned to the wrong variable.

The important change is the sign in line 1962. With this change the problem is fixed.

I also review all the other uses of `merge_types`, and I found two places where I think it should say `-1` instead of `1`. But both of the versions with `-1` and `1` compile without problems, and I coudn't make an example where this difference causes a problem. Moreover, using `printf` in those places, it looks like it's dead code (or they are used only in very weird code).

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -3630,6 +3630,30 @@
            #f)
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Check that the type information is shifted in the
+;; right direction while inlining.
+;; The first example triggered a bug in 6.3.
+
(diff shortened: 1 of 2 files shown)


mflatt:
This change looks right to me.
It does look like the two places that 1 becomes -1 are currently unreachable, since the `SCHEME_LET_STAR` representation mode has become uncommon. I'm not sure what to do about that, though, since it seems that small changes to the optimizer could make them reachable again, and they cover forms that are allowed by the data structure's constraints.

gus-massa:
Should I add a

    Merge to v6.4

to the comment before merging?

mflatt:
Yes, please do add "Merge to v6.4".

PR #1205, opened 2016-01-13
Fix check-version from version/check

lexi-lambda:
I’m not sure how long this has been an issue for or what changed to make it break, but `check-version` doesn’t currently work on 6.2.1, as discovered by @benklebe. This breaks DrRacket’s “Check for Updates” option. This is a quick fix. There might be a nicer way to restructure some of this, but it probably isn’t worth it.

### DIFF
--- a/racket/collects/version/check.rkt
+++ b/racket/collects/version/check.rkt
@@ -1,27 +1,14 @@
 #lang racket/base
-(require racket/tcp)
+
+(require net/url
+         racket/tcp
+         "utils.rkt")
 
 (define version-url "http://download.racket-lang.org/version.txt")
 (define timeout 30)
 
-(require "utils.rkt")
-
-;; This file can be invoked from an installer, and in case it's
-;; without zo files using `net/url' is extremely slow.  Instead, do
-;; things directly.
-;; (require net/url)
-;; (define (url->port url) (get-pure-port (string->url url)))
-
 (define (url->port url)
-  (define-values [host path]
-    (apply values (cdr (regexp-match #rx"^http://([^/:@]+)(/.*)$" url))))
-  (define-values [i o] (tcp-connect host 80))
-  (fprintf o "GET ~a HTTP/1.0\r\nHost: ~a\r\n\r\n" path host)
-  (flush-output o)
-  (close-output-port o)
-  (unless (regexp-match #rx"^HTTP/[0-9.]+ 200 OK\r\n.*?\r\n\r\n" i)
-    (error 'url->port "bad reply from server: ~a" (read-line)))
-  i)
+  (get-pure-port (string->url url)))
 
 (define error-value
   (case-lambda
@@ -104,3 +91,8 @@
 (provide check-version)
 (define (check-version)
   (with-timeout timeout check-version-raw))
+
+(module+ test
+  (let ([result (check-version)])
+    (unless (eq? result 'ok)
+      (error 'check-version "failed due to non-ok result: ~v" result))))


stamourv:
I'm not sure I understand the fix. Was the problem that closing the port earlier could lead to the input not being sent? A comment explaining why the code has to be this way (as opposed to the "obvious" way it was before) would be nice.

Also, are there tests for this code?

stamourv:
@rfindler and I had a closer look.
Our best theory is that closing the port too early caused some buffer somewhere to block or something, and that closing later fixed it. @elibarzilay, does this look plausible?
It looks like a potentially more robust refactoring would be to have `url->port` return both `o` and `i`, and have the code that reads `i` close `o` after the read to `i`.

lexi-lambda:
There are a number of issues with the existing implementation that this *doesn’t* fix. For example, if the server returns a non-200 status code, the error message is useless: it calls `read-line` without an argument, so it tries to read from whatever `current-input-port` was set to. Even if it was adjusted to `(read-line i)`, it would miss the status line.

I would be willing to try to do a more comprehensive refactor to make some of the code more robust, but since this can’t really depend on anything besides `racket/base` and `racket/tcp`, it’s probably going to have some rough edges, anyway.

lexi-lambda:
(As an aside, it would be very nice to have this fixed in 6.4 so that the DrRacket auto-updater works.)

rfindler:
I'm happy with the fix you suggest, Alexis, and slightly happier with one
that at least appears to avoid the same problem happening again. Other
fixes like you suggest would also be great. I also agree that getting
something into 6.4 seems wise.

lexi-lambda:
@rfindler Do you have something specific in mind when you say you want to “avoid the same problem happening again”? Seems like adding some tests would be an obvious choice, but were you considering anything else?

rfindler:
Sorry: I meant that the fix you suggest is strange and our best (sadly not
very good) theory suggests a slightly different fix and so we should do
that, just in case.

Tests would also be nice. :)

lexi-lambda:
Ah! I see—I’m a little out of the loop. What exactly is that theory and what fix would it imply?

rfindler:
The theory is that there is some buffer that filled up somewhere that
causes blocking and so that's why moving the close later works. But if
that's the case, why not move the close until the point where the data is
actually read?

Even better would be to know why this fix works, tho.

mflatt:
I think closing too early sends an EOF to the server, and if the server gets that EOF early enough it will close its end, too.

rfindler:
I'm not completely sure from reading the code, but it looks like (read i)
producing an eof shouldn't have hung, but we were seeing it hang, not
terminate with an error.

mflatt:
I got "bad reply from server", but I'm not too surprised if the server replies in different ways to an EOFed connection.
See also `tcp-abandon-port`, which closes an output port without sending an EOF.

rfindler:
`tcp-abandon-port` looks like the right fix, I agree.

lexi-lambda:
Sounds like `tcp-abandon-port` is perfect, and it seems to work correctly. I’ve updated the PR to use that, instead.

samth:
I think a comment about why `tcp-abandon-port` is the right thing would be nice there (including a pointer to this issue/discussion).

lexi-lambda:
Ok, added a comment.

elbarzil:
FWIW, this has nothing much to do with version checking...  IIRC, the
`tcp-abandon-port` thing was introduced when some (yahoo?) servers were
misbehaving, and the recent changes to the servers are likely the cause
of this breaking now.

But one thing to note is that I wrote that code assuming that it should
be available in some minimal version of racket where there's none of the
usual libraries for dealing with urls.  Since a minimal racket needs
these things anyway, it might be better to dump that `url->port` thing
for something more complete.  (This would also be more robust against
the no-longer-true quasi-assumption that it works against a specific
server.)

stamourv:
@elbarzil: Are you saying that the definition could be replaced with 
```
(define (url->port url) (get-pure-port (string->url url)))
```
as the comment above the current definition says?

elibarzilay:
Yes, provided that it's not running into any new problems, or the
problem that is mentioned in that comment (that I haven't seen earlier).

It would also be good in making it deal with https, redirections, and
all kinds of other http-isms that might be needed on the way to the web
content now.

(And BTW, "elibarzilay" is much better -- I just still can't remember
that I'm using two github logins...)

stamourv:
@elibarzilay: Sounds good! That's a much more robust solution.

lexi-lambda:
Okay, if that works, sounds like an ideal solution. I’ve updated this to use `get-pure-port`.

stamourv:
Looks good.
Could you add a test submodule, for drdr's benefit?

lexi-lambda:
Ok, added a simple test that actually hits r-l.org.

stamourv:
LGTM

lexi-lambda:
Merged in 3620bae6da0c8d315f9aa924ee3eff55e7bb20ee.

PR #1206, opened 2016-01-13
Add 'fallback key to config.rktd files.

samth:
The 'fallback key points to a directory to find another config.rktd file which
is used if no key is found.

### DIFF
--- a/pkgs/racket-doc/scribblings/raco/config.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/config.scrbl
@@ -153,6 +153,16 @@ directory}:
        @racket[#t] if the installation uses absolute path names,
        @racket[#f] otherwise.}
 
+ @item{@indexed-racket['fallback] --- a path, string, or byte string
+       for a directory containing an additional @filepath{config.rktd}
+       file. The procedures provided by @racketmodname[setup/dirs]
+       automatically consult this file for keys not found in the
+       @filepath{config.rktd} file in the @tech{configuration
+       directory}. The additional @filepath{config.rktd} file may have
+       a @racket['fallback] entry, which is also consulted as needed.
+
+       @history[#:added "6.4.0.1"]}
+
  @item{@indexed-racket['cgc-suffix] --- a string used as the suffix (before
        the actual suffix, such as @filepath{.exe}) for a
        @filepath{CGC} executable. Use Windows-style casing, and the
--- a/pkgs/racket-doc/scribblings/raco/setup.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/setup.scrbl
@@ -1409,6 +1409,15 @@ function for installing a single @filepath{.plt} file.
   absolute path names for executable and library references, 
   @racket[#f] otherwise.}
 
+@defprof[(find-fallback-dir) (or/c path? #f)]{
+  Returns a path to the directory used to find a fallback @filepath{config.rktd}
+  file, used when the @filepath{config.rktd} file via @racket[(find-config-dir)]
+  does not specify a value for a given key. The result is @racket[#f] if no such
+  fallback directory is specified. 
+
+  @history[#:added "6.4.0.1"]
+}
+
 @; ------------------------------------------------------------------------
 
 @section[#:tag "getinfo"]{API for Reading @filepath{info.rkt} Files}
--- a/racket/collects/setup/private/dirs.rkt
+++ b/racket/collects/setup/private/dirs.rkt
@@ -14,20 +14,27 @@
 ;; ----------------------------------------
 ;; config: definitions
 
-(define config-table
+;; read-config-file : path[dir] -> promise[hash]
+(define (read-config-file d)
   (delay/sync
-   (let ([d (find-config-dir)])
-     (if d
-         (let ([p (build-path d "config.rktd")])
+   (let* ([p (build-path d "config.rktd")]
+          [tbl
            (if (file-exists? p)
-               (call-with-input-file* 
-                p
-                (lambda (in) 
-                  (call-with-default-reading-parameterization
-                   (lambda ()
-                     (read in)))))
-               #hash()))
-         #hash()))))
+               (call-with-input-file*
+                   p
+                 (lambda (in)
+                   (call-with-default-reading-parameterization
+                    (lambda ()
+                      (read in)))))
+              #hash())]
+          [fallback (hash-ref tbl 'fallback #f)])
+     (if fallback
+         (hash-set tbl 'fallback (delay/sync (read-config-file fallback)))
+         tbl))))
+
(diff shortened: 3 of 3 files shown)


mflatt:
Unfortunately, there's a second implementation of "config.rktd" reading in "startup.rktl", so 'fallback would need to be implemented there, too.
Can you say more about your use case, though? I think I've implemented this once and decided that it was unnecessary or a bad idea, but I don't immediately recall the details.

mflatt:
There are also sandbox interactions where it's important to read the configuration file early enough, otherwise modules fail when run in a sandbox. So, delaying the read of a fallback configuration file could run into trouble with sandboxes.

samth:
Right, I wanted to have at least a strawman implementation to discuss if this was a good idea before fixing things like that.

The use case is that I'd like to create a tool that implements something like Cabal sandboxes in Haskell. I think all the pieces are already there to make this work, but this would make it nicer. The way it would work is as follows:

* A sandbox would basically just be a directory with a `pkgs` subdirectory that would be used as a directory scope for `raco pkg`, and a `config.rktd` file which lists that directory in the `'pkgs-search-dirs` entry. 

* Then, if you run `racket` with the appropriate `config.rktd` file specified on the command line/environment variable, everything would see the pkgs you installed in the sandbox. Otherwise, it would appear as if they weren't there. (This would probably be automated with a shell script.)

That plan would already work today, but to create the new `config.rktd` file you'd have to copy the existing one, and add an entry to the `'pkgs-search-dirs` key. Furthermore, if someone changed some aspect of the configuration, it wouldn't be reflected in the sandbox configuration. Thus this approach.

mflatt:
If you set `'pkgs-search-dirs`, you almost certainly need to set `'links-search-files`, and then `'lib-search-dirs` for installed foreign libraries, and probably a couple of other things. It may be that you end up copying a lot of the configuration, anyway, which suggest that a single `'fallback` field is too coarse-grained.

I wonder whether a sandbox might be better implemented via `PLTADDONDIR`, so that each sandbox instance is something like a separate user, and so anything installed in user scope is specific to the sandbox. That's the direction I ended up going in the distribution-building framework (via the `-A` uses in the top-level `Makefile`). Or did you have in mind a different interaction with user scope?

samth:
I had imagined that packages installed in user scope would be accessible to programs running in the sandbox, which I think wouldn't work if using PLTADDONDIR the way you suggest. But given that that's not such a big loss, it's probably better to try that direction first, since it wouldn't require core changes.

Can you explain why extending 'pkgs-search-dirs would require also extending 'links-search-files? I think I understand the need to changed 'lib-search-dirs for when pkgs install native libraries.

mflatt:
I think it's unlikely that someone will be able to use packages installed in user scope while switching a wider scope by changing sandboxes. Probably that line of thought is how I gave up on this idea before.

The `'pkgs-search-dirs` configuration field itself doesn't affect the search for collections. The package system updates a file that must be listed in `'links-search-files` when a package is installed.

samth:
@lexi-lambda Since you were asking about this, you might be interested in this discussion.

lexi-lambda:
I don’t currently understand how exactly scopes work, but for what it’s worth, I would *not* want packages installed in a user scope to be available within a sandbox. In fact, ideally, each sandbox should be as isolated as possible. Having deps *appear* to be installed locally, then having them not exist upon deployment because they were globally installed is a frustrating source of dependency hell.

I’m not sure if this is something scopes are capable of solving (as opposed to just having separate installations), but it would be nice if they could.

mflatt:
I've worked through a refinement of the `PLTADDONDIR` idea, but I'm currently ambivalent about it. See
mflatt/racket@c3f1805552a0528354ba415d37cb2c3adab8fe18.

That commit adds `raco setup`-level infrastructure to support something like
```
  raco sandbox init
```
which would set up the current directory as a kind of sandbox for further user-scope package installations, etc., and it would create a "bin" directory to hold variants of `racket`, `raco`, etc., that use the sandbox. Instead of relying on a `PLTADDONDIR` environment variable, which would leak to subprocesses and and have to be wrapperd around any use of the sandbox (including launchers created from the sandbox), the generated `racket` would be more directly tethered to the enclosing directory as its addon directory.

I'm ambivalent, because I'm not yet sure the use case is worth the extra support in `raco setup` and related tools. I also know that there are a few more gaps to fill, such as making `find-exe` locate the tethered `racket` and making `raco exe -l` preserve the addon directory, and probably there are other little gaps I didn't think of.

A small number of packages would need to change to cooperate with `raco setup` and addon tethering, because the packages generate executables manually as an installation step. Here's an example: mflatt/gui@d4b90761725f6a5752dccb7fe6c9a16f47f04b25.

lexi-lambda:
> Instead of relying on a `PLTADDONDIR` environment variable, which would leak to subprocesses 

Could you elaborate on why this would be a bad thing? I could imagine a Racket program that shells out to `raco my-custom-command`, or just shells out to `racket` and assumes the existence of the sandbox-installed dependencies, so my first instinct is that I would *want* that environment to propagate to subprocesses.

mflatt:
@lexi-lambda I think you should fix such a program to use `find-exe` if it wants to run the same Racket as the one running your program. Otherwise, that program doesn't run in a Racket installation that's not in `PATH` or isn't the first one that's in `PATH`.

I suppose that setting environment variables seems bad to me because I'm so frequently working on build tools or shell-like programs that perform externally supplied commands. A shell-like program might be implemented in Racket X, but if it's running commands that say `racket`, then the commands surely intended Racket Y as in my path when I started. In those situations, environment variables don't compose well for the same reason that dynamic binding in general doesn't compose as well as lexical binding.

If someone wants to parameterize a computation with a new `PATH` or `PLTADDONDIR` setting, then that's certainly fine, but `parameterize` should be separate and explicit – perhaps through a `raco sandbox exec` tool.

mflatt:
After thinking about it more, I'm inclined to merge the changes in mflatt/racket@c3f1805 (with a few updates). I think it's probably close enough to the right track to be worth keeping.

lexi-lambda:
> If someone wants to parameterize a computation with a new `PATH` or `PLTADDONDIR` setting, then that's certainly fine, but parameterize should be separate and explicit – perhaps through a `raco sandbox exec` tool.

This makes sense to me. (I think I was imagining that kind of tool, but you’ve made me realize that’s a separate concept.)

mflatt:
@samth Do you want to keep this PR open, or should we close it? The idea of chaining "config.rktd"s still makes sense, and I ended up generalizing tethering in a way that would play well with it, but chaining still looks to me like it would be difficult to work with.

samth:
I'm just going to close it -- I wanted the chaining for the sandbox feature that we're moving toward.

PR #1209, opened 2016-01-15
read/contract

bennn:
(I'm trying to solve a specific use-case. This solution is maybe too general.)

### Goal

Parsing command-line arguments is a little awkward & error-prone right now.
```
(define my-n (make-parameter 0))
(command-line
  [("-n") arg "a number" (let ([n (number->string arg)]) (if n (my-n n) (error "expected number")))]
  ....)
```
I'd like an easy way to read values from a string, assert they match a specification, and raise a helpful message otherwise.

### This PR

Adds a general `read/contract` for reading one datum from an input port & applying a contract. This makes command-line parsing a little easier:

```
....
(with-input-from-string arg (lambda () (my-n (read/contract number?))))
....
```

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1784,6 +1784,16 @@ The @racket[define-struct/contract] form only allows a subset of the
   
 }
 
+@defform[#:literals (current-input-port)
+         (read/contract contract-expr [in (current-input-port)])
+         #:contracts ([in input-port?])]{
+  Calls @racket[read] to retrieve a single @tech{datum} from @emph{in}, then applies
+  @racket[contract-expr] to the result.
+
+  @history[#:added "6.4.0.1"]
+
+}
+
 @defidform[current-contract-region]{
   Bound by @racket[define-syntax-parameter], this contains
   information about the current contract region, used by
--- a/racket/collects/racket/contract/private/base.rkt
+++ b/racket/collects/racket/contract/private/base.rkt
@@ -4,6 +4,7 @@
          (rename-out [-recursive-contract recursive-contract])
          current-contract-region
          invariant-assertion
+         read/contract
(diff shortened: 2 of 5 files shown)


racket/collects/racket/contract/private/base.rkt
```diff
@@ -87,6 +88,21 @@
...
+  (with-syntax ([(ctc port)
+                 (syntax-case stx ()
+                  [(_read ctc) #'(ctc #f)]
+                  [(_read ctc in) #'(ctc in)])])
+    (quasisyntax/loc stx
+      (let* ([in (or port (current-input-port))]
+             [v (read in)])
+        (contract ctc
+                  v
+                  in
```
bennn: Is there a better way to make a blame object?

pkgs/racket-doc/scribblings/reference/contracts.scrbl
```diff
@@ -1784,6 +1784,16 @@ The @racket[define-struct/contract] form only allows a subset of the
   
 }
 
+@defform[#:literals (current-input-port)
+         (read/contract contract-expr [in (current-input-port)])
+         #:contracts ([in input-port?])]{
+  Calls @racket[read] to retrieve a single @tech{datum} from @emph{in}, then applies
+  @racket[contract-expr] to the result.
```
bennn: What's the easiest way to add examples? Create a namespace with `racket/port` and call `with-input-from-string`?

samth:
Why is an API like `read` the best solution here? Why not something like:

```
(checked str number->string #:message "whoops")
```

With this definition

```
(define (checked v conv [pred values] #:message [msg "bad value])
  (define v* (conv v))
  (unless (pred v*)
    (error msg))
  v*)
```

That seems shorter and doesn't involve `read` or contracts that don't have boundaries, and seems like it fits your use case.

bennn:
Something like `checked` is a better fit right here, but I thought this version of `read` could be useful in other places.

Issue #1212, opened 2016-01-16
Make srfi-19 easily found in the documentation


linusbergl:
Hi there!

I have spent some time looking through the documentation trying to find a "configurable" date->string procedure. I was surprised to not find it in the racket time library since it is not something you want to write yourself, and racket in many ways is "ammunition included". 

Now, I knew that I could just use srfi-19, but that is not clear in the documentation which might be confusing to racket beginners. 

My suggestion would be to somehow clarify that there is such a procedure included in the racket distribution, or even make the SRFI-19 one part of racket/date.

Best regards
Linus

takikawa:
You may want to look at the [gregor](https://github.com/97jaz/gregor) package instead of srfi/19 BTW. The srfi/19 library is not very Rackety.

linusbergl:
Well. There is one procedure to convert date structs to strings with control of how the output should look included. Depending on a third party library should not be necessary.

I just think it might be a good idea to make the date->string from srfi-19 a bit more discoverable, or at least provide something as powerful in the standard Racket/date.

rfindler:
I think the future is the gregor library too and so we try to make it more discoverable, perhaps by pointing people to it from racket/date or other things like that.

linusbergl:
Well... I don't doubt you, but currently there is no working documentation for gregor online (403, permission denied) and installing it through raco also does not give a working documentation. 

And isn't it a bit overkill to demand that you install a completely new library (with no less than 7 other deps) to be able to date->string any non-standard string representation of a date? The syntax of gregor for that specific function isn't particularly more clear (from what I gather reading the tests on github) than srfi19 date->string.

Installing an extra 7 packages for just one procedure that is already included seems a bit off to me. I'm just a hobby dev, so what do I know.

jackfirth:
@linusbergl srfi/19 is itself a third party dependency as well, it just happens to have been written before Racket's new package system was created and thus was included with the full distribution by default. So it "already being included" is much more an accident of history than a conscious design decision.

PR #1214, opened 2016-01-18
margin-note for gregor & srfi-19

bennn:
For #1212, until `gregor` is part of a main distribution :smile: 

Should we add an adjective to discourage srfi-19?

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/time.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/time.scrbl
@@ -165,6 +165,9 @@ result is the result of  the last @racket[body].}
 @; ----------------------------------------------------------------------
 
 @section[#:tag "date-string"]{Date Utilities}
+@margin-note{For more date & time operations, see
+  @other-doc['(lib "gregor/scribblings/gregor.scrbl") #:indirect "Gregor: Date and Time"]
+  or @link["../srfi/srfi-19.html"]{srfi/19}}
 
 @note-lib-only[racket/date]
 


stamourv:
Could you use an indirect `other-doc` to point to the Gregor docs if Gregor is already installed?
Re discouraging SRFI 19 (or SRFIs in general): I'm all for it.
Re including Gregor in the main distribution: I talked with @97jaz about that at RacketCon, and he's in favor. We just need to go ahead and do it. :)

rfindler:
I think the name "gregor" is not conducive to people finding this libray.
Could it be named `date`?

jbclements:
> On Jan 20, 2016, at 2:18 PM, Vincent St-Amour <notifications@github.com> wrote:
> 
> Could you use an indirect other-doc to point to the Gregor docs if Gregor is already installed?
> Re discouraging SRFI 19 (or SRFIs in general): I'm all for it.
> Re including Gregor in the main distribution: I talked with @97jaz about that at RacketCon, and he's in favor. We just need to go ahead and do it. :)

Gregor docs are a teensy bit scary. Is there any way to make it slightly less confrontational?

John

rfindler:
Pull request with edits?

bennn:
(I don't know how to answer anything but Vincent's first 2 comments in this PR)

bennn:
@stamourv I haven't been able to figure out `other-doc`. Can you show me what you have in mind?

(How do I reference the local docs for `gregor`? How does `other-doc` know when the local doc doesn't exist? Does the pkg server URL go in the `#:indirect` argument -- if so I haven't figured out how to make it go to the URL on click)

stamourv:
AFAIU, `other-doc` will check whether the package is installed, and point there if it is. I think putting the pkg link in the `indirect` argument should work?

jackfirth:
> Gregor docs are a teensy bit scary. Is there any way to make it slightly less confrontational?

Perhaps Gregor could benefit from a "Gregor Guide / Gregor Reference" split in the docs?

97jaz:
So, a few issues:

If Gregor is going to be part of the main distribution, I agree that it should have a name more like "date." The reason it doesn't have a name like that now is because Racket already has a date library (`racket/date`), and gregor is incompatible with it. (There are a few questions that need to be answered before Gregor becomes part of the main distribution. More on that below.)

The documentation: I'm open to suggestions and pull requests. For my own part, I think it could use a better (and longer) introduction, and the section named "Time Scale" could stand to be rewritten.

As for including Gregor in the main distribution, there are a few things to work out:
- `racket/base` already provides a `date` struct. Does Gregor's `date` need to be renamed? Similarly, `racket/base` provides `time` syntax, while Gregor provides has a `time` struct. JodaTime's solution to this problem in Java was to call its classes `LocalDate` and `LocalTime`. We could do something like that.
- For localization Gregor currently depends on a number of CLDR data files, which are in their own packages. These files are large, and I think there would be a lot of resistance to including them in the distribution. I'm sure we don't need to ship with support for all of these locales: [https://github.com/unicode-cldr/cldr-dates-modern/tree/master/main]. But I could use some help figuring out which we need to ship with and whether and how users might be able to add more.

bennn:
Hm, I think the current revision is the right code, but in Scribble it's not very good.

Code:
```
@other-doc['(lib "gregor/scribblings/gregor.scrbl")
  #:indirect @hyperlink["http://pkgs.racket-lang.org/#[gregor]"]{Gregor}]
```

Rendered:
gregor01

On click, it goes to the local `racket/doc/gregor/index.html`, whether or not `gregor` is installed.

In contrast, leaving out the `#:indirect` argument renders as below, and links to the same location.

- If `gregor` is installed:
screen shot 2016-01-25 at 04 20 37

- Otherwise:
screen shot 2016-01-25 at 04 19 02


Should we change `other-doc` to check if the package is installed, or just do a check in this margin-note?

bennn:
With [cce6235](https://github.com/racket/racket/commit/cce6235a96b0bfc5e866d2e79db60902061b4d38):

- If `gregor` is not installed
g-no

- If `gregor` is installed and the reference is rebuilt:
g-yes

mflatt:
@bennn I don't think you should make documentation rendering depend explicitly on whether a package is installed.

The intent is that you use
```
 @other-doc['(lib "gregor/scribblings/gregor.scrbl") #:indirect "Gregor: Date and Time"]
```
and then when "gregor" is not installed, the link goes to docs.racket-lang.org.

The missing piece is that docs.racket-lang.org doesn't yet have all package documentation, so the link to docs.racket-lang.org doesn't work as well as it should. I expect to finally move all package documentation to docs.racket-lang.org next week, though.

bennn:
Thanks; we're now using the intended `other-doc`.

pkgs/racket-doc/scribblings/reference/time.scrbl
```diff
@@ -1,5 +1,5 @@
 #lang scribble/doc
-@(require "mz.rkt" (for-label racket/date))
+@(require "mz.rkt" pkg/lib (for-label racket/date))
```
stamourv: I don't think you need that anymore.

stamourv:
LGTM modulo comment and commit squashing.

bennn:
squashed

PR #1217, opened 2016-01-20
add `in-hash-set` fast path for hash sets

stchang:
This is a pull request because I'm unsure about the name. 

`in-hash-set` creates a sequence from a hash-table-based set. But is the name confusing? Maybe `in-hashset` as another alternative?

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sets.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sets.scrbl
@@ -153,16 +153,20 @@ Analogous to @racket[for/list] and @racket[for*/list], but to
 construct a @tech{hash set} instead of a list.
 }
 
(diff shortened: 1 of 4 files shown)


stchang:
I named the sequences `in-immutable-set`, `in-mutable-set`, and `in-weak-set`. Will merge unless there are objections.

stchang:
Here's some microbenchmark numbers for `in-set` vs the new sequences for set size = 1000000

```
immutable
cpu time: 216 real time: 214 gc time: 0
cpu time: 84 real time: 83 gc time: 4
mutable
cpu time: 60 real time: 60 gc time: 0
cpu time: 32 real time: 31 gc time: 0
weak
cpu time: 128 real time: 127 gc time: 0
cpu time: 80 real time: 80 gc time: 0
```

PR #1219, opened 2016-01-21
add racket/immutable

AlexKnauth:
For operations that return immutable strings, byte-strings, or vectors.


racket/collects/racket/immutable.rkt
```diff
@@ -0,0 +1,299 @@
...
+(define-syntax-rule
+  (define-immutable-versions ->immutable stuff ...)
+  (begin
+    (define-immutable-version ->immutable stuff)
+    ...))
+
+;; Helper Conversion Functions
+
+(define (maybe-string->immutable-string str)
+  (and str (string->immutable-string)))
```
stamourv: This is missing a `str` argument.

stamourv:
Why have whole new functions for, e.g., upcasing immutable strings. Why not add an `#:immutable?` keyword argument to existing functions? That would also remove the need for a dedicated library.

racket/collects/racket/immutable.rkt
AlexKnauth: Thanks, fixed now.

AlexKnauth:
Re: `#:immutable?` keyword argument:
Since functions like `string-upcase` are provided by the kernel, and the kernel doesn't support keyword arguments, I don't think that would work. I guess there could be a module somewhere below `racket/base` that re-defined them and delegated to the old versions ... would that be better than this?

samth:
That's the usual strategy -- see the keyword-accepting IO functions, for example.

AlexKnauth:
Would it make sense to have an `create-immutable?` parameter (or maybe separate `create-immutable-string?`, `create-immutable-bytes?`, and `create-immutable-vector?` parameters) for this?

AlexKnauth:
I've make a separate pull request here (https://github.com/racket/racket/pull/1341) for adding an optional `#:immutable?` keyword argument to everything.

minad:
I was just about to propose a function vector-immutable! to turn a mutable vector into an immutable one. I would like to use that instead of placeholders. This PR seems to be exactly about that. What is the status here?

AlexKnauth:
That's what @gus-massa's `unsafe-vector*->immutable-vector!` pull request (https://github.com/racket/racket/pull/1237) is about. It would make libraries like this cheaper by getting rid of unnecessary allocations.

I'm not sure what the status is. I don't know whether it would be best to make separate functions (this pull request) or to add optional keyword arguments to the existing ones (https://github.com/racket/racket/pull/1341). There are different problems for both.

gus-massa:
I'm sitll not convinced if `unsafe-vector*->immutable-vector!` is a good idea or not. I prefer this ugly long name, so it's clear that it's a weird operation with strict conditions. It will return the same vector magicaly marked as immutable, but it also may return a new inmmutable vector, or perhaps a cached version???. My proposal was to just set the immutable flag, becaause it's easier and faster.

The problem of using `vector->immutable-vector!` is that it may break some code, like this:

    #lang racket
    
    (define (my-v-immutable? v)
      (eq? (vector-ref v 0) 'immutable))
    
    (define (my-v->immutable! v)
      (vector-set! v 0 'immutable))
    
    (define (my-v-set! v n val)
      (when (my-v-immutable? v)
        (error "nasty error :( :( :( :( :( :( :("))
      (vector-set! v n val))
    
    ;---
    
    (define (something v)
      (sleep (random))
      (when (my-v-immutable? vec)
        (error "nice error :)"))
      (sleep (random))
      (my-v-set! vec 1 7))
    
    (define vec (vector 'mutable 1))
    
    (void (thread (lambda ()
                    (sleep (random))
                    (sleep (random))
                    (my-v->immutable! vec))))
    
    (something vec)
    (vector-ref vec 1)

the function `something` looks good. It has a nice error in case someone use an immutable vector, but if you run this a few time you will get a race condition that makes the program raise the nasty error.

My other concer was that `vector->immutable-vector!` may confuse a future version of the bytecode optimizer. I think the current version ignores immutability of vectors, so there are currently no problems with something like this.

Anyway, the plan is to change the current VM of Racket and use the VM of Chez Scheme. See https://groups.google.com/forum/#!msg/racket-dev/2BV3ElyfF8Y/4RSd3XbECAAJ

So I guess that adding weird new procedures is on hold.

samth:
`unsafe-vector*->immutable-vector` would be useful in several places that I know of, including in the contract library, where the same operation is performed by copying. So adding this odd procedure in the near term seems like a useful idea to me.

minad:
@samth I would like to have it to create immutable cyclic datastructures. but why not a nicer name like vector->immutable-vector! what is unsafe about it? I dont think it matters for the underlying vm or wont matter with chez, or does it?

mfelleisen:
What are the preconditions that make existing code fail in weird ways?

samth:
@minad The reason that it's unsafe is that code can currently rely on vectors not to change to be immutable, and so the new function would break the assumptions of current code. The precondition of the function would presumably be that no one else has a reference to the vector, which the function would not check. Generally, `unsafe` means that the procedure has unchecked preconditions.

samth:
@mfelleisen I'm not sure what you're asking. The bad thing that could happen is just what @gus-massa describes.

mfelleisen:
I understand that. But can we formulate a general pre-condition when these bad things can’t happen?

mfelleisen:
p.s. I guess the pre-condition is “no other thread assumes that this vector is immutable and the upstream flowgraph doesn’t either.”  W/o a Rust like type system, we may not be able to get there.

samth:
I think the precondition is "no one else has a reference to the vector". Which we can't check, and so it's an `unsafe` function.

mfelleisen:
[[ Now that’s a non-sequitor. We can’t check unsafe-vector-ref either but in Andrew’s refinement type system we can. ]]

samth:
I think the definition of `unsafe` in Racket is "could violate system invariants if used without following preconditions, where those preconditions are not enforced".

mfelleisen:
There is 

 — not enforced but enforceable in principle 
 — not enforceable, period 

I thin the two are distinct and this one is of the second kind and thus more worrisome than then first.

samth:
Just as @pnwamk's additions to TR can enforce the contract on `unsafe-vector-ref` (or just a dynamic check), an affine type system or a more expensive dynamic check (such a reference counts) could enforce that precondition.  Similar hard-to-enforce preconditions: the FFI requires lots of things of C code in order to be safe, `unsafe-chaperone-procedure` omits very-hard-to-verify dynamic check, etc.

mfelleisen:
See above.

gus-massa:
I prefer a stricter condition. “There are no other references to this vector anywhere, and I promise to not use this reference again”. 

My idea was to use this in straightforward cases, where the result of a primitive is guaranteed to be a fresh vector (or string) so it’s obvious that there are no other references anywhere, for example to transform 

    (vector->immutable-vector (list->vector a-long-list))

to

    (unsafe-vector*->immutable-vector! (list->vector a-long-list))

I think it’s ok to extend the uses cases to something like 

    (define (make-long-vector n a b c)
      (unsafe-vector*->immutable-vector! 
        (let ([v (make-vector n b)])
          (vector-set! v 0 a)
          (vector-set! v (-n 1) c)
          v)))

because it is applied to the only reference to the vector that is alive.

And by extension to equivalent code like 

    (define (make-long-vector n a b c)
      (let ([v (make-vector n b)])
        (vector-set! v 0 a)
        (vector-set! v (-n 1) c)
        (unsafe-vector*->immutable-vector! v)))

but now is not so obvious what “alive” means. Probably something like: `v` is never used in the continuation.

But I wouldn’t like something like

    (define (make-long-vector n a b c)
      (let ([v (make-vector n b)])
        (vector-set! v 0 a)
        (vector-set! v (-n 1) c)
        (unsafe-vector*->immutable-vector! v)
        v))

I don’t know too much about rust(!!!). but IIRC in rust after you use a reference as an argument you lose the ownership and you can't use it again. Also, I think you can enforce that all the borrowed references are released and you have the only and true reference. 

And for more complex flows, if they can implement reference counting as a library I guess they can extend it to check that there is only one reference alive. I'm not sure if they have a garbage collector to breack cycles, because you have to be sure that all the temporal reference have been collected.

mfelleisen:
Your uses are good. (I could extend the set! calculus and prove them syntactically correct.) But I like to think of unsafe-foobar! as something that comes with pre-conditions that we can check or validate somehow in general and in a reasonably efficient manner. This one appears to impose a large run-time cost and/or a thoroughly complicated/new type system.

mromyers:
Here's a way you could wrap it to safely enforce that condition:
```
(define (make-immutable-vector/pre-ref n [b])
  (let ([v (make-vector n)])
    (case-lambda
     [() (if (immutable? v) v
           (unsafe-vector*->immutable-vector! v))]
     [(i x)(vector-set! v i x)])))

(define (make-long-vector n a b c)
  (define pre-v (make-immutable-vector/pre-ref n b))
  (pre-v 0 a)
  (pre-v (- n 1) c)
  (pre-v))
```
I think the compiler should be able to in-line `pre-v` in most cases.

PR #1224, opened 2016-01-25
add hash-iterate-pair and hash-iterate-key+value

stchang:
- cuts in-hash and in-hash-pairs iteration time in half
- refactor hash_table_index
- add tests


stchang:
The switch to hamt's in 6.3 slows `hash-iterate-key` and `hash-iterate-value`. `in-hash` and `in-hash-pair` were previously making separate calls to `in-hash-key` and `in-hash-value` and thus hash iteration slowed considerably from 6.2 to 6.3. 

This refactoring eliminates the unnecessary lookups.

mflatt:
@stchang It looks like `hash_table_index` lost chaperone handling. Am I misreading that change?

stchang:
@mflatt You are right.  I moved chaperone handling to each scheme_hash_iterate_ function. 

With the change to hash_table_index's signature to add multiple return values, it made it cleaner this way by eliminating a few lookups and checks.

mflatt:
@stchang  Got it, thanks!

The docs need `@history[#:added "6.4.0.5"]`. 

Also, when you update the version number, it also needs to be updated in "pkgs/base/info.rkt". (Unfortunately, having that version updated automatically would defeat the point of the dependency spec, so we manually update the version in both places and use a test to check that they're in sync.)

PR #1226, opened 2016-01-28
Documentation for scheme_register_process_global is fixed.

LeifAndersen:
It was previously both incomplete, and incorrect.


mflatt:
merged

PR #1227, opened 2016-01-29
optimizer: prediates, local types and fixnum reductions

gus-massa:
**First commit: use type predicates to calculate local types**

The functions `expr_implies_predicate` was very similar to `expr_produces_local_type`, and slightly more general. Merging them, it is possible to use the type information in expressions where the optimizer used only the local types that were visible at the definition.

For example, this is useful in this expression to transform `bitwise-xor` to it's unsafe version.

    (lambda (x)
      (when (fixnum? x)
        (bitwise-xor x #xff)))

Previously the reduction was applied to

    (let ([x (vector-length v)])
      (bitwise-xor x #xff))

I left `scheme_expr_produces_local_type` as a wrapper of `expr_implies_predicate`. I used `fuel=10`, because it was the value used before, but in all the other cases `fuel=5`. It's more complicated, because the functions counted `fuel` in a slightly different way. I'm not sure if I should change it to `5` here too, but I think it's probably not very relevant.

**Seccond commit: reductions for expressions with fixnum**

For example, reduce:

    (= <fx> <fx>) ==> (unsafe-fx= <fx> <fx>)
    (fx= <fx> <fx>) ==> (unsafe-fx= <fx> <fx>)
    (max <fx> <fx>) ==> (unsafe-fxmax <fx> <fx>)
    (fxmax <fx> <fx>) ==> (unsafe-fxmax <fx> <fx>)
    (zero? <fx>) ==> (unsafe-fx= <fx> 0)
    (bitwise-not <fx>) ==> (unsafe-fxnot <fx>)
    (fxnot <fx>) ==> (unsafe-fxnot <fx>)

I added `racket/unsafe/ops` to the namespace used in the test, so I removed a lot of mini-modules. I hope it's ok.

In the redution `(zero? <fx>) ==> (unsafe-fx= <fx> 0)`, I put:

    SCHEME_APPN_FLAGS(new) |= (APPN_FLAG_IMMED | APPN_FLAG_SFS_TAIL)

This line was necessary in an old commit with the reduction of `(equal? <fx> <fx>) ==> (eq? <fx> <fx>)`. I think it's not necessary here. Is it correct? Is it necessary? Is it necessary to add something similar to all the other reductions?

I was not sure if it's better to reduce

    (zero? <fx>) ==> (unsafe-fx= <fx> 0)

or

    (zero? <fx>) ==> (eq? <fx> 0)

I think they are equivalent, but the first looked more natural. Perhaps it's possible to change all the `(unsafe-fx= x y) ==> (eq? x y)` everywhere.

It would be straightforward to add `(zero? <fl>) ==> (unsafe-fl= <fl> 0)` and family. Do I add them? Later?


mflatt:
This looks good to me.

Setting `APPN_FLAG_IMMED | APPN_FLAG_SFS_TAIL` looks right.

I'm not sure about `eq?` versus `unsafe-fx=`. Keeping `unsafe-fx=` may communicate to parts of the optimizer that the arguments are expected to be fixnums, but I think that kind of expectation only matters currently for flonums (for unboxing).

gus-massa:
I'll rebase+merge this as is, and later I´ll try to think if it's better to add a few `APPN_FLAG_IMMED | APPN_FLAG_SFS_TAIL` here and there.

PR #1229, opened 2016-01-29
add unsafe hash-iterate- ops

stchang:
- add `unsafe-<hash type>-iterate-<op>` ops
- add `in-<hash type>-<key or val>` sequence form
- add `unsafe-<hash type>-iterate-<op>*` ops (dont check for impersonators and chaperones)
- add `in-<hash type>-<key or val>*` sequence form (dont check for impersonators and chaperones)

- immutable hash iteration time (the slowest of all the hash tables) cut in half
- other hash table iteration improved by 20-30%
(will post some microbenchmarks below)

- refactors `hash_table_next` (in the same manner as `hash_table_index`)

Design questions:
- Are the new primitives in the right collections?
- Are the names ok?
- Is it necessary to have impersonator-supporting and non-impersonator-supporting variants?
(the performance for the non-impersonator "star" variants is only slightly better)

I will squash if everything else looks ok.

*edited to add: I achieve the immutable hash iteration speedups by having `hash-iterate-first` and `-next` return a "pointer" into the tree, rather than a numeric index. I figured this is ok since it's an unsafe op.


stchang:
Here are some numbers for hash tables with 1000000 keys and vals.

```
1st num: non-fast path (current Racket head)
2nd num: (safe) fast path (current Racket head)
3rd num: unsafe fast path
4th num: unsafe fast path (no impersonator check)
immutable --------------------
in-hash
cpu time: 228 real time: 229 gc time: 0
cpu time: 196 real time: 199 gc time: 0
cpu time: 124 real time: 121 gc time: 0
cpu time: 108 real time: 111 gc time: 0
in-hash-pairs
cpu time: 224 real time: 225 gc time: 0
cpu time: 200 real time: 200 gc time: 0
cpu time: 108 real time: 109 gc time: 0
cpu time: 108 real time: 107 gc time: 0
in-hash-keys
cpu time: 212 real time: 209 gc time: 0
cpu time: 184 real time: 185 gc time: 0
cpu time: 84 real time: 86 gc time: 0
cpu time: 84 real time: 85 gc time: 0
in-hash-values
cpu time: 216 real time: 214 gc time: 0
cpu time: 188 real time: 188 gc time: 0
cpu time: 96 real time: 98 gc time: 0
cpu time: 100 real time: 99 gc time: 0
mutable --------------------
in-hash
cpu time: 76 real time: 77 gc time: 0
cpu time: 48 real time: 47 gc time: 0
cpu time: 36 real time: 38 gc time: 0
cpu time: 32 real time: 32 gc time: 0
in-hash-pairs
cpu time: 72 real time: 72 gc time: 0
cpu time: 44 real time: 48 gc time: 0
cpu time: 36 real time: 33 gc time: 0
cpu time: 32 real time: 32 gc time: 0
in-hash-keys
cpu time: 60 real time: 60 gc time: 0
cpu time: 36 real time: 36 gc time: 0
cpu time: 24 real time: 22 gc time: 0
cpu time: 20 real time: 21 gc time: 0
in-hash-values
cpu time: 60 real time: 61 gc time: 0
cpu time: 40 real time: 37 gc time: 0
cpu time: 20 real time: 22 gc time: 0
cpu time: 20 real time: 21 gc time: 0
weak --------------------
in-hash
cpu time: 128 real time: 128 gc time: 0
cpu time: 100 real time: 103 gc time: 0
cpu time: 72 real time: 70 gc time: 0
cpu time: 68 real time: 67 gc time: 0
in-hash-pairs
cpu time: 124 real time: 123 gc time: 4
cpu time: 84 real time: 85 gc time: 0
cpu time: 68 real time: 69 gc time: 0
cpu time: 64 real time: 61 gc time: 0
in-hash-keys
cpu time: 112 real time: 113 gc time: 0
cpu time: 72 real time: 72 gc time: 0
cpu time: 48 real time: 45 gc time: 0
cpu time: 44 real time: 44 gc time: 0
in-hash-values
cpu time: 116 real time: 116 gc time: 0
cpu time: 72 real time: 75 gc time: 0
cpu time: 44 real time: 44 gc time: 0
cpu time: 40 real time: 41 gc time: 0
```

stchang:
This is the test file.

```
#lang racket
(define big-list (build-list 1000000 add1))
(define big-list2 (build-list 1000000 values))

(define big-hash/immut (make-immutable-hash (map cons big-list big-list2)))
(define big-hash/mut (make-hash (map cons big-list big-list2)))
(define big-hash/weak (make-weak-hash (map cons big-list big-list2)))

(define big-hash/immut/seq (in-hash big-hash/immut))
(define big-hash/mut/seq (in-hash big-hash/mut))
(define big-hash/weak/seq (in-hash big-hash/weak))

(define big-hash/immut-pairs/seq (in-hash-pairs big-hash/immut))
(define big-hash/mut-pairs/seq (in-hash-pairs big-hash/mut))
(define big-hash/weak-pairs/seq (in-hash-pairs big-hash/weak))

(define big-hash/immut-keys/seq (in-hash-keys big-hash/immut))
(define big-hash/mut-keys/seq (in-hash-keys big-hash/mut))
(define big-hash/weak-keys/seq (in-hash-keys big-hash/weak))

(define big-hash/immut-vals/seq (in-hash-values big-hash/immut))
(define big-hash/mut-vals/seq (in-hash-values big-hash/mut))
(define big-hash/weak-vals/seq (in-hash-values big-hash/weak))

(collect-garbage)
(collect-garbage)
(collect-garbage)

(displayln "1st num: non-fast path (current Racket head)")
(displayln "2nd num: (safe) fast path (current Racket head)")
(displayln "3rd num: unsafe fast path")
(displayln "4th num: unsafe fast path (no impersonator check)")
(require rackunit)

(displayln "immutable --------------------")
(displayln "in-hash")
(check-true 
  (= (time (for/sum ([(k v) big-hash/immut/seq]) (+ k v)))
     (time (for/sum ([(k v) (in-hash big-hash/immut)]) (+ k v)))
     (time (for/sum ([(k v) (in-immutable-hash big-hash/immut)]) (+ k v)))
     (time (for/sum ([(k v) (in-immutable-hash* big-hash/immut)]) (+ k v)))
     ))
(displayln "in-hash-pairs")
(check-true 
  (= (time (for/sum ([k+v big-hash/immut-pairs/seq]) (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-hash-pairs big-hash/immut)]) 
             (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-immutable-hash-pairs big-hash/immut)])
             (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-immutable-hash-pairs* big-hash/immut)])
             (+ (car k+v) (cdr k+v))))
     ))
(displayln "in-hash-keys")
(check-true
  (= (time (for/sum ([k big-hash/immut-keys/seq]) k))
     (time (for/sum ([k (in-hash-keys big-hash/immut)]) k))
     (time (for/sum ([k (in-immutable-hash-keys big-hash/immut)]) k))
     (time (for/sum ([k (in-immutable-hash-keys* big-hash/immut)]) k))
     ))
(displayln "in-hash-values")
(check-true
  (= (time (for/sum ([v big-hash/immut-vals/seq]) v))
     (time (for/sum ([v (in-hash-values big-hash/immut)]) v))
     (time (for/sum ([v (in-immutable-hash-values big-hash/immut)]) v))
     (time (for/sum ([v (in-immutable-hash-values* big-hash/immut)]) v))
     ))

(displayln "mutable --------------------")
(displayln "in-hash")
(check-true
  (= (time (for/sum ([(k v) big-hash/mut/seq]) (+ k v)))
     (time (for/sum ([(k v) (in-hash big-hash/mut)]) (+ k v)))
     (time (for/sum ([(k v) (in-mutable-hash big-hash/mut)]) (+ k v)))
     (time (for/sum ([(k v) (in-mutable-hash* big-hash/mut)]) (+ k v)))
     ))
(displayln "in-hash-pairs")
(check-true
  (= (time (for/sum ([k+v big-hash/mut-pairs/seq]) (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-hash-pairs big-hash/mut)]) 
             (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-mutable-hash-pairs big-hash/mut)]) 
             (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-mutable-hash-pairs* big-hash/mut)]) 
             (+ (car k+v) (cdr k+v))))
     ))
(displayln "in-hash-keys")
(check-true
  (= (time (for/sum ([k big-hash/mut-keys/seq]) k))
     (time (for/sum ([k (in-hash-keys big-hash/mut)]) k))
     (time (for/sum ([k (in-mutable-hash-keys big-hash/mut)]) k))
     (time (for/sum ([k (in-mutable-hash-keys* big-hash/mut)]) k))
     ))
(displayln "in-hash-values")
(check-true
  (= (time (for/sum ([v big-hash/mut-vals/seq]) v))
     (time (for/sum ([v (in-hash-values big-hash/mut)]) v))
     (time (for/sum ([k (in-mutable-hash-values big-hash/mut)]) k))
     (time (for/sum ([k (in-mutable-hash-values* big-hash/mut)]) k))
     ))

(displayln "weak --------------------")
(displayln "in-hash")
(check-true
  (= (time (for/sum ([(k v) big-hash/weak/seq]) (+ k v)))
     (time (for/sum ([(k v) (in-hash big-hash/weak)]) (+ k v)))
     (time (for/sum ([(k v) (in-weak-hash big-hash/weak)]) (+ k v)))
     (time (for/sum ([(k v) (in-weak-hash* big-hash/weak)]) (+ k v)))
     ))
(displayln "in-hash-pairs")
(check-true
  (= (time (for/sum ([k+v big-hash/weak-pairs/seq]) (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-hash-pairs big-hash/weak)]) 
             (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-weak-hash-pairs big-hash/weak)]) 
             (+ (car k+v) (cdr k+v))))
     (time (for/sum ([k+v (in-weak-hash-pairs* big-hash/weak)]) 
             (+ (car k+v) (cdr k+v))))
     ))
(displayln "in-hash-keys")
(check-true
  (= (time (for/sum ([k big-hash/weak-keys/seq]) k))
     (time (for/sum ([k (in-hash-keys big-hash/weak)]) k))
     (time (for/sum ([k (in-weak-hash-keys big-hash/weak)]) k))
     (time (for/sum ([k (in-weak-hash-keys* big-hash/weak)]) k))
     ))
(displayln "in-hash-values")
(check-true
  (= (time (for/sum ([v big-hash/weak-vals/seq]) v))
     (time (for/sum ([v (in-hash-values big-hash/weak)]) v))
     (time (for/sum ([v (in-weak-hash-values big-hash/weak)]) v))
     (time (for/sum ([v (in-weak-hash-values* big-hash/weak)]) v))
     ))
```

gus-massa:
:)

Visual Studio don´t like [hash-c#L4226](https://github.com/racket/racket/pull/1229/files#diff-90d2fa46e6ddf69604e1a60db1e1dcffR4226). What about:

    - Scheme_Object *obj = argv[0], *res[0];
    + Scheme_Object *obj = argv[0], *res[2];

I don't undetnd what is a "pointer" into the tree. What happens if someone prints it? It creates a segfault? Is the pointer encoded as a fixnum? What happens if there is a gc and the hash is moved (or collected)?

How is this visible? I tried this, but I didn't get an error:

    #lang racket/base
    (require racket/unsafe/ops)

    (define h #hash((1 . 1) (2 . 2) (3 . 3) (4 . 4) (5 . 5) (6 . 6)))
    #;(define h (make-hash '((1 . 1) (2 . 2) (3 . 3) (4 . 4) (5 . 5) (6 . 6))))
    (let loop ([hi (unsafe-immutable-hash-iterate-first* h)])
      (when hi
        (displayln hi)
        (loop (unsafe-immutable-hash-iterate-next* h hi))))

stchang:
>  hash-c#L4226

Oops, typo. Thanks! Strange that the tests didnt segfault

> "pointer"

The signature of `unsafe-immutable-hash-iterate-first*` is `ImmutableHash -> UnsafeIndex` where

An `UnsafeIndex` is one of:
- `(list ImmutableHash ExactNonNegInt)`
- `(cons ImmutableHash (cons ExactNonNegInt UnsafeIndex))`

where the hashes in the list are subtrees of the input and an index in the list is a position in the corresponding subtree.

Printing an UnsafeIndex should print a list of valid Racket hash and int values.

gus-massa:
OK, it's a litte strange but not dangerous. I like this. For the record, the smallest example I found was with:

    (define l (build-list 33 add1))
    (define h (make-immutable-hash (map cons l l)))

Now I have the inverse question: Why can't this be used for the safe version of `hash-iterate-first`? I guess the problem is that `hash-iterate-first` has already the wrong signature.  

Would be it a good idea to define all the new `unsafe-X-hash-iterate-first` as (X . -> . any/c) like `dict-iterate-first`, in case someone wants to change the implementation later?

stchang:
> the smallest example I found was with

Yes, since the nodes are 32 bits wide, any immutable hash of size 32 or less will not have any subtrees.

> Why can't this be used for the safe version of hash-iterate-first

Yes it could, though as you mentioned it would technically be backwards-incompatible. It may not be so bad though since you can't do anything with the result of `hash-iterate-first` and `-next` other than to give it to the other `hash-iterate-` functions.

> Would be it a good idea to define all the new `unsafe-X-hash-iterate-first` as `(X . -> . any/c)` 

Yes I think that's a good idea. I'll change the docs.

gus-massa:
Just another question. For small hashes (<= 32?), is it posible to use only the number? I think that in many ocations the hash has only a 10 or 20 elements, and it would be nice to iterate without producing garbage. Does this change make the implementation much harder?

Perhaps this idea can be extended to the first iterations of any hash. If current code returns `(#hash((1 . 1) ...) n)` then return just `n` instead. (This will make no difference for big hashes, but I guess this is easier to implement than a special case for small hashes.)

stchang:
> small hashes

That might be worth looking into. You bring up a good point in general about space usage. Maybe I could switch to a mutable data structure and reuse it when calling `next` within the same subtree. That would make it log space instead of linear.

rfindler:
Does it make sense to just have a completely separate path for hashes that are below a certain size? You'd have to have some kind of a test at the start of the for loop, but that seems like it could be in the noise, even for small hashes

stchang:
> completely separate path for hashes that are below a certain size

I can do some experiments but my guess is that small hash iteration will be quick enough that it might not matter

rfindler:
I see what you mean. I haven't built your commit, but I'm seeing that small immutable hashes are about 2x slower than vectors right now and if you've already sped them up by 2x, then it seems unlikely that you'd get better results that the vector case. Here's my test code:

```
#lang racket
(define v (make-vector 30 #f))
(define ht (apply hash (apply append (build-list 30 (λ (x) (list 1 #f))))))
(define N 2000000)
(time
 (for ([x (in-range N)])
   (for ([e (in-vector v)])
     e)))

(time
 (for ([x (in-range N)])
   (for ([(k v) (in-hash ht)])
     v)))
```

gus-massa:
A small correction the last test case:

    (define ht (apply hash (apply append (build-list 30 (λ (_x_) (list _x_ #f))))))


I guess that the time difference was only in the large hashes, were the old algorithm needed a lot of pointer chasing to reach the actual vale and calculate the next one. In small hashes I guess the difference was smaller. (I really need to measure this to be sure.)

About mutable structures: I prefer immutable structures. Can an intermediate index be saved and the sequence replayed later (for example, using a continuation)?

I think that adding a special case to replace for `(original-hash pos) ==> pos` will be enough to solve the space problem for small hashes without changing too much the algorithm.

gus-massa:
I wrote my idea and pushed it to one of my branches: https://github.com/gus-massa/racket/tree/16-1-E-stchang%2Bunsafe-hash-iterate-try2

It's an advance draft (I hope all the tests finich correctly). If you want you can copy and squash it and fix the details. Or if you prefer I can send it as a PR after this is finished.

I coudn't run "make export" in Windows. (Is it necesary to add scheme_unsafe_hash_tree_start/next to schemef.h? Isn't it better to add it to schpriv.h? I'm not sure about the difference.)

I think it would be good to add test for bigger hashes in hash.rktl. IIUC it only test a hash with 10 elements. Do the other hashes have some internal magic numbers bigger than 32?

stchang:
@gus-massa Have you run timing experiments with your additions? On my current branch, small immutable hash table iteration takes negligible time, suggesting that it's not worth it to special case them.

> schemef.h vs schpriv.h

I wasn't sure where to put the declarations either but I just followed the other `hash_tree_` functions.

And yes, tests with larger hash tables would be helpful. I'll add them before I merge.

stchang:
Modifying the iteration index structure "in place" shaves another 10-20% off iteration time for immutable hashes. Here are some updated numbers:

```
1st num: non-fast path (current Racket head)
2nd num: (safe) fast path (current Racket head)
3rd num: unsafe fast path
4th num: unsafe fast path (no impersonator check)
immutable --------------------
in-hash
cpu time: 228 real time: 227 gc time: 0
cpu time: 200 real time: 201 gc time: 0
cpu time: 96 real time: 94 gc time: 0
cpu time: 92 real time: 92 gc time: 0
in-hash-pairs
cpu time: 228 real time: 229 gc time: 4
cpu time: 200 real time: 202 gc time: 4
cpu time: 96 real time: 94 gc time: 0
cpu time: 88 real time: 90 gc time: 0
in-hash-keys
cpu time: 208 real time: 209 gc time: 0
cpu time: 184 real time: 186 gc time: 0
cpu time: 68 real time: 68 gc time: 0
cpu time: 68 real time: 69 gc time: 4
in-hash-values
cpu time: 208 real time: 212 gc time: 0
cpu time: 192 real time: 190 gc time: 0
cpu time: 80 real time: 82 gc time: 0
cpu time: 80 real time: 81 gc time: 0
```

rfindler:
Is that safe for continuation grabs?

stchang:
No probably not.

rfindler:
Which means we can't use it, right? :(

stchang:
Right, unless we have an unsafe iterator

samth:
But even if we had that unsafe iterator, we couldn't use it in `for` loops, so it doesn't seem that valuable.

stchang:
Unsafe `for`? :)

stchang:
Another idea could be to return the index and sub tree using multiple values. That would eliminate both the mutation and the excess pair allocations.

gus-massa:
I think it's better to keep the x2 improvement of the first version, and keep it safe for continuations. Using `values` will make it more difficult to use, and actually `values` are slower than `cons`.

I don't have a benchmark for the use with small hashes. It's just a gut feeling. Usually the non allocating version is faster. With small hashes, both run times are negligible, but if you have an `unsafe-immutable-hash-next` inside a hot thigh loop, I guess it would be possible to see a measurable the difference.

(I have some ideas to extend the non-allocating version to medium hashes (with a few hundreds of elements), but I want to think carefully about the details. My idea is to use essentially your implementation so it's fast, but encode the tree differently. I hope to have working prototype and benchmarks for next week.)

stchang:
> `values` will make it more difficult to use

Hmm, it seems like the `for` forms don't support multiple values anyways (because there's no named `let-values`). I guess I'll have to revert.

mflatt:
Lots of good work here!

I don't understand how this is safe for mutable hash tables. It seems
like on thread could use `in-mutable-hash` (which is supposed to be
safe) while another thread mutates the hash table, and then the
`in-mutable-hash` thread can crash. Am I missing a guard somewhere?

Not related to this patch: Does iteration through a weak hash table
currently make sense? It looks like a GC during iteration could cause
`hash-iterate-value` to raise a "bad index" error even if the index was
derived from `hash-iterate-next`.

Related to this patch: Instead of an error, I think you get a seg
fault, because the weak reference is replaced with NULL.


The "for.rkt" changes look like a lot of new and repetitive code. Does
it make sense to abstract more? Specifically, I worry about increasing
the footprint of "for.rkt". Is it much bigger with these changes?
Creating a new macro that expands to `define-sequence-syntax` would not
address that issue, but moving the content of the function that
generates `:do-in` to a compile-time function might make both the
implementation and compiled format smaller.

Please stick to `/* ... */` comments in the C code, so that we can
continue to support old compilers.

I think you meant to fill in the `""`s to `chaperone_hash_key`.

mflatt:
I agree that the new function declarations generally belong in "schpriv.h" instead of "schemef.h". The "schemef.h" file is for functions that we want to make visible via the FFI or to C embeddings/extensions.

stchang:
Managed to cut another 10-20% from immutable hash iteration by decrementing the index instead of incrementing. (compare to the top set of numbers above)

```
1st num: non-fast path (current Racket head)
2nd num: (safe) fast path (current Racket head)
3rd num: unsafe fast path
4th num: unsafe fast path (no impersonator check)
immutable --------------------
in-hash
cpu time: 228 real time: 229 gc time: 0
cpu time: 200 real time: 202 gc time: 0
cpu time: 104 real time: 105 gc time: 0
cpu time: 104 real time: 101 gc time: 0
in-hash-pairs
cpu time: 228 real time: 229 gc time: 0
cpu time: 200 real time: 201 gc time: 0
cpu time: 80 real time: 78 gc time: 0
cpu time: 76 real time: 79 gc time: 0
in-hash-keys
cpu time: 208 real time: 208 gc time: 0
cpu time: 188 real time: 190 gc time: 0
cpu time: 76 real time: 75 gc time: 0
cpu time: 76 real time: 76 gc time: 4
in-hash-values
cpu time: 216 real time: 214 gc time: 0
cpu time: 188 real time: 191 gc time: 0
cpu time: 88 real time: 85 gc time: 0
cpu time: 80 real time: 84 gc time: 0
```

stchang:
> I don't understand how this is safe for mutable hash tables

Yes, you are right. I need to put the guards back in for mutable hash tables and bucket tables.

> Does iteration through a weak hash table currently make sense? It looks like a GC during iteration could cause hash-iterate-value to raise a "bad index" error even if the index was
derived from hash-iterate-next.

Yes this can happen but I think an exception is the best we can do. Unless `iterate-next` returns a promise? 

And I'll add a check to make sure it throws an exception instead of segfaulting.

> The "for.rkt" changes look like a lot of new and repetitive code

I agree it needs abstraction! I was going to work on it after I got the primitives stable.

> Please stick to /* ... */

Ah ok, I didn't realize one was newer.

> I think you meant to fill in the ""

Oops. Thanks.

> I agree that the new function declarations generally belong in "schpriv.h"

I'll try to clean things up a little.

stchang:
@mflatt Do you think the non-chaperone-supporting unsafe variants are necessary here? They don't really improve performance and they add a lot of code and primitives.

mflatt:
@stchang Dropping them sounds fine to me.

stchang:
Does anyone know if I can restart the AppVeyor tests?

samth:
New pushes should cause it to run again.

mflatt:
@stchang JFYI: Looking back at this commit, it looks like you took "new functions in 'schpriv.'h" to mean that all functions should be moved there. It's not backwards-compatible to move functions out of "schemef.h", though, because it means that they functions are not necessarily accessible to embeddings, extensions, or FFI uses. Although that level of compatibility is not super-important in practice, I think the old functions probably should have stayed in "schemef.h", and I will investigate moving them back.

Along similar lines, I think #1248 moves toward restoring the split between implementation details in "hash.c" and Racket interface in "list.c". I imagine moving even more back in that direction.

One concrete thing that went wrong is `hamt_popcount()` as an non-static function. All non-static bindings are supposed to start with `scheme_` (to avoid potential collisions at the level of C linking). That rule probably isn't written down anywhere, and I'll start writing down things like this in the "README" file.

stchang:
Ok sorry for the misunderstanding. I'll try to stick to the conventions in the future.

Adding them to the README would be good. I'll try to think of other things to add if that would be helpful. Relatedly, another thing I wasn't sure about is the flags on primitives. Are these written down somewhere?

For example, what is the difference between between `UNSAFE_OMITABLE` and `UNSAFE_FUNCTIONAL`? From what I can tell `OMITABLE` is used for references to mutable data structures, while `FUNCTIONAL` is for immutable data structures?

mflatt:
Right: a reference to a mutable data structure is itself omittable, since it has no side-effects, but it doesn't always give the same answers for the same inputs. So, unsafe-functional implies unsafe-omittable, but not vice versa.

I think those descriptions are best written next to the flag definitions.

mflatt:
I've added comments for the `SCHEME_PRIM_` flags, to be pushed later.

stchang:
Makes sense, thanks!

stchang:
Sorry, one more question @mflatt. 

My use of the `FUNCTIONAL` flag for mutable hash `-iterate-first` and `-iterate-next` is wrong?  eg, 
https://github.com/racket/racket/blob/master/racket/src/racket/src/list.c#L896

PR #1231, opened 2016-02-02
not sort xml attributes

simmone:
When I use:
 ```
(xml->xexpr 
  (document-element 
    (read-xml (open-input-string 
      "<a c=\"1\" a=\"2\" b=\"3\">inner</a>"))))
```
I expected the result is:
      `'(a ((c "1") (a "2") (b "3")) "inner"))`

but I found the actual result is:
    ` '(a ((a "2") (b "3") (c "1")) "inner")`
attibutes list is sorted.

I want attibutes list order is the original order in xml.

So I change the lex-attributes func, use check-duplicates to find duplicate attributes, then return the original order attributes list.

### DIFF
--- a/pkgs/racket-test/tests/xml/test.rkt
+++ b/pkgs/racket-test/tests/xml/test.rkt
@@ -462,6 +462,12 @@ END
      (test-syntax:read-xml/element
       "<a href=\"#\">inner</a>"
       '(a ([href "#"]) "inner"))
+
+     (test-syntax:read-xml/element
+      "<a c=\"1\" a=\"2\" b=\"3\">inner</a>"
+      '(a ([c "1"] [a "2"] [b "3"]) "inner"))
+
+     (test-syntax:read-xml/element/exn "<a c=\"1\" a=\"2\" c=\"3\">inner</a>" "read-xml: lex-error: at position 1.20/21: duplicated attribute name c")
      
      (test-syntax:read-xml/element
       "<root>&nbsp;</root>"
--- a/racket/collects/xml/private/reader.rkt
+++ b/racket/collects/xml/private/reader.rkt
@@ -270,17 +270,16 @@
 
 ;; lex-attributes : Input-port (-> Location) -> (listof Attribute)
 (define (lex-attributes in pos)
-  (sort (let loop ()
-          (skip-space in)
-          (cond [(name-start? (peek-char-or-special in))
-                 (cons (lex-attribute in pos) (loop))]
-                [else null]))
-        (lambda (a b)
-          (let ([na (attribute-name a)]
-                [nb (attribute-name b)])
-            (cond
-              [(eq? na nb) (lex-error in pos "duplicated attribute name ~a" na)]
-              [else (string<? (symbol->string na) (symbol->string nb))])))))
+  (let* ([result_list 
+          (let loop ()
+            (skip-space in)
+            (cond [(name-start? (peek-char-or-special in))
+                   (cons (lex-attribute in pos) (loop))]
+                  [else null]))]
+         [check_dup (check-duplicates result_list (lambda (a b) (eq? (attribute-name a) (attribute-name b))))])
+    (if check_dup
+        (lex-error in pos "duplicated attribute name ~a" (attribute-name check_dup))
+        result_list)))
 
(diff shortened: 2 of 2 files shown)


jeapostrophe:
Thanks! Do you think it would be better to use `check-duplicates` from racket/list so there's no allocation? Also, do any existing tests break?

simmone:
updated.

One of the CI failed, it report `raco test -l tests/db/all-tests` failed.
But in my branch, it all passed, I don't know why, the db tests unstable?

samth:
That test has been failing intermittently recently, so it's not your fault.

jeapostrophe:
Awesome, thanks so much!

mbutterick:
Am I missing something? Why would duplicate attribute definitions trigger an error?

“an XML processor may at user option issue a warning when … more than one attribute definition is provided for a given attribute, **but this is not an error**.”

https://www.w3.org/TR/xml/#attdecls

simmone:
I think duplicate attributes deliver a ambiguous meaning, it’s a error indeed.

mbutterick:
Well, if the XML spec explicitly says otherwise, there’s nothing to argue.

Still, it's not hard to see why the spec doesn’t insist on unique attributes. Suppose you had a million attributes. Every time you added another, you’d have to search through the whole list looking for a duplicate, and if found, remove it. This could be costly. 

Instead, the XML spec treats the attribute list as a LIFO stack, where you can push new attributes onto the front in the usual `cons` style (without worrying about duplicates). 

Moreover, duplicate attributes are not ambiguous — the spec also says “the first declaration is binding.” Thus, you can read attributes by using a `memq` style first-match heuristic (again, instead of searching the whole list, removing duplicates in advance, etc.)

mbutterick:
Erratum: said `memq`, meant `assq`

simmone:
Thank you.

I understand, find duplicate may be costly.
Yours opinion is ignor the duplicate attribute, right?

I google a few online xml validator, like:

http://codebeautify.org/xmlvalidate

http://www.utilities-online.info/xsdvalidation

http://www.xmlvalidation.com/documentation.html

If I use duplicate attributes, it all report the error.

So I won't change the piece of code and want to see more people to join this argument.

Btw, another interesting question is :
The opensource project like racket, if have different opinion on some feature, who decide to do or not do?

jeapostrophe:
I think Matthew brings up an important point. Let's see if I understand
things correctly...

1) Old versions of `xml` de-duplicate and sort the attributes. This means
that for old code, string->xml and xml->string are not inverses. Changing
either could potentially change existing system behavior.

2) The spec does not require unique attributes, but it also does not seem
to require allowing duplicates because it seems to say that libraries may
only see the first.

3) We could make the parser go faster by following the spec and just
`cons`ing every attribute and not looking at anything

4) A few online checkers flag errors with duplicate attributes... perhaps
because it may be confusing to some people which will take effect.

I feel that we should look at what other popular libraries do. It seems
like the standard may not really be authoritative here. In particular, I'm
curious what libxml and Java's XML libraries do on these points.


On Fri, Feb 5, 2016 at 7:32 PM, Chen Xiao <notifications@github.com> wrote:

> Thank you.
>
> I understand, find duplicate may be costly.
> Yours opinion is ignor the duplicate attribute, right?
>
> I google a few online xml validator, like:
>
> http://codebeautify.org/xmlvalidate
>
> http://www.utilities-online.info/xsdvalidation
>
> http://www.xmlvalidation.com/documentation.html
>
> If I use duplicate attributes, it all report the error.
>
> So I won't change the piece of code and want to see more people to join
> this argument.
>
> Btw, another interesting question is :
> The opensource project like racket, if have different opinion on some
> feature, who decide to do or not do?
>
> > 在 2016年2月6日，上午2:00，Matthew Butterick <notifications@github.com> 写道：
> >
> > Erratum: said memq, meant assq
> >
> > —
> > Reply to this email directly or view it on GitHub <
> https://github.com/racket/racket/pull/1231#issuecomment-180471570>.
>
> >
>
> —
> <https://github.com/racket/racket/pull/1231#issuecomment-180634741>.
>



-- 
Jay McCarthy
Associate Professor
PLT @ CS @ UMass Lowell
http://jeapostrophe.github.io

                          - D&C 64:33

PR #1233, opened 2016-02-07
Fix eq reduction

gus-massa:


mflatt:
This change looks right to me.

PR #1234, opened 2016-02-09
More reductions for (eq? var var) and (if expr var var)

gus-massa:
These are many short commits, I'll probably squash all but the first. It's easier to discuss if they are separated.

**First commit: avoid compiler warning**
Totaly unrelated.

**Second commit: reduce `(if t v v) => (begin t v)` when `v` is a top level**
Is it necessary to check that both instances of the toplevel have the same flags? I think that currently all of them have the same flags in these cases.

**Third commit: reduce `(eq? var var) ==> #t`**
I'm reusing `equivalent_exprs`, I don't like this completely but it looks silly to make a copy of the implementation. (Perhaps, in the future, I'd like to extend this to more complex expressions like `(eq? (car x) (car x))`. But befor that, I want to see if expressions like this appear in real code and look carefully at the details. Because the second `car` would be transformed into a `unsafe-car` ...) Also, I reduced this to `(eq? var var) ==> (begin var var #t) ==> #t` just in case to be future proof.

**Fourth commit: reduce `(if v v #f) => v` when `v` is a top level**
I reused again `equivalent_exprs` but with another pattern. It's slightly different because it changes the number of times `v` is called in each branch and after the optimization.

**Sixth commit: comment (squash)**
I added a warning to `equivalent_exprs`. Is it necesary? Is this a good idea? 

**Fifth commit: move `(if x #t #f)` reduction after branch optimization**
It makes more sense to test this after the branches are optimized, instead of before. It catches more cases, but in some cases it generates a new unnecessary `#<void>`. For example in `(not (if (let ([r (random)]) #t) #t #f)))`.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1042,18 +1042,38 @@
 (test-comp '(lambda (x) (eq? 7 x))
            '(lambda (x) (eqv? 7 x)))
 
-(test-comp #t
-           '(eq? 7 7))
-(test-comp #f
-           '(eq? 9 6))
-(test-comp #t
-           '(eqv? 7 7))
-(test-comp #f
-           '(eqv? 9 6))
-(test-comp #t
-           '(equal? 7 7))
-(test-comp #f
-           '(equal? 9 6))
+; car is a primitive, map is required from another module
+(let ([test-equal?
+       (lambda (e?)
+         (test-comp #t
+                    `(,e? 7 7))
+         (test-comp #f
+                    `(,e? 9 6))
+         (test-comp #t
+                    `(,e? (values 1 2) (values 1 2))
+                    #f)
+         (test-comp '(lambda (x) #t)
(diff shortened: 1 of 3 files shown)


gus-massa:
**Seventh commit: Reduce `(if (let (...) x y #t) K _) => (let (...) x y K)`**

This fix the problem with the unnecessary `#<void>` added in the fifth commit, for expressions like `(not (if (let ([r (random)]) #t) #t #f))`. It's slightly more general so it will make smaller a few additional expressions like `(if (let ([r (random)]) #t) 0 1)`.

Also, I found and fixed a bad interaction between two of the reductions. The problem was that in this example, it recognized the test condition as an expression with a `#f` as a result, and then it reduced

    (if (begin (random) #f) 1 2) 
      => (if (begin (begin (random)]) #f) #f) 1 2)
      => (begin (begin (random) #f) 2)

gus-massa:
I rebased this over the two major changes, and now it's compiling without error over HEAD. I squashed the changes to rebase mere easily, but the questions are still there.

mflatt:
Looks good to me.

PR #1237, opened 2016-02-10
unsafe-vector*-to-immutable-vector! (draft)

gus-massa:
This is an idea for a strange new primitive. I made only a minimal implementation to discuss and get some feedback. It changes a mutable vector into an immutable vector. I think it's a nice complement for `racket/inmmutable`:https://github.com/racket/racket/pull/1219.

Benchmarks with `100000000` elements:

```
#lang racket/base

(define l-100 (for/list ([i (in-range 100)]) i))
(define l-1000 (for/list ([i (in-range 1000)]) i))

(time (for/sum ([i (in-range 1000000)])
        (vector-length (vector->immutable-vector (list->vector l-100)))))

(time (for/sum ([i (in-range 1000000)])
        (vector-length (list->vector l-100))))

(time (for/sum ([i (in-range 100000)])
        (vector-length (vector->immutable-vector (list->vector l-1000)))))

(time (for/sum ([i (in-range 100000)])
        (vector-length (list->vector l-1000))))
```

Results with the change disabled (flag in line 3317 = 0)

```
cpu time: 2808 real time: 2871 gc time: 96
cpu time: 1748 real time: 1874 gc time: 15
cpu time: 2558 real time: 2789 gc time: 110
cpu time: 1451 real time: 1464 gc time: 77
```

Results with the change enabled (flag in line 3317 = 1)

```
cpu time: 1841 real time: 1940 gc time: 46
cpu time: 1716 real time: 1767 gc time: 31
cpu time: 1591 real time: 1648 gc time: 79
cpu time: 1419 real time: 1453 gc time: 61
```

The problem is that it breaks an invariant. I think the invariant is not used in (untyped)Racket, but it may be used in the future or in Typed Racket, so I hid the primitive.

It's applied only in a few cases, where the argument of `vector->inmmutable-vector` is an explicit "fresh" `vector`. Now it only works with `list->vector` and `struct->vector`, because I cut a few corners to get a first draft.

This two functions are marked as `fresh` using an especial function, it may be implemented as a flag in the primitive instead.

Currently it only work with `vector`s, because there was already some support for `vector`s in the optimizer. I think this change would be more useful for `string`s and `bytes`s.


samth:
What's the invariant that this breaks?

samth:
Also, this primitive would be useful in the contract system, even without the automatic optimization.

gus-massa:
Perhaps the right word is not "invariant", but this example will get an strange result if one of the `(immutable? v)` is optimized beause it's the result should not change:

    (define (f v) (unsafe-vector*->immutable-vector! v))
     
    (let ([v (vector 1 2 3)])
      (let ([x (immutable? v)]
            [y (f v)]
            [z (immutable? v)])
        (eq? x z)))

gus-massa:
Just to clarify, my idea is that the previous code is an _unsuported_ use of `unsafe-vector*->immutable-vector!` so the result is unspecified and compiler is free to generate any result (including a segfault).

In any sensible implementation, I expect the result to be `#f` or `#t`.

For example, the naive implementation just run the code as written and the result is `#f`. This is the current behavior in Racket and I guess it will not change during this year, but I wish to keep the possibility of an improvement for later. I don't know if Typed Racket currently does something smarter.

A sufficient smart compiler could realize that `z` is equal to `x`, because the immutability of the vectors never changes, so the result will be `#t`.

gus-massa:
This is superseded by 61b365f4be8b772d1b14ddb18ff74d775381a4be that is already merged.

PR #1241, opened 2016-02-14
change intermediate repersentation for the bytecode optimizer

mflatt:
This pull request is in the hope of getting @gus-massa to poke at the
changes to make sure they're sensible to merge. Any amount of code
review would be appreciated, but probably the changes are too large
for a line-by-line review. (Maybe it helps that there are 1k fewer
LoC, but that sill leaves 2k changed/added LoC.) Trying examples
and/or decompiling may be the only reasonable strategy for review.

It's a big change. Considering the time required to make the changes,
the possibility of bugs or regressions, and the fact that this compiler
will be replaced as soon as possible, I'm not sure it was worthwhile.
Still, I feel good about how the change made some of the worst code in
the compiler become much simpler or go away, and it may allow
worthwhile optimizations in the short run.

See also racket/compiler#12


mflatt:
Note: the CI failures in "modprot" are due to the mismatch between the old "zo-lib" package (as updated by the PR in the "compiler" repo) and these changes.

racket/src/racket/src/schpriv.h
```diff
@@ -1459,6 +1459,104 @@ Scheme_Object *scheme_top_introduce(Scheme_Object *form, Scheme_Env *genv);
...
+   is allocated to represent a separate variable. Different passes
+   in the comiler store different information about the variable. */
+typedef struct Scheme_Compiled_Local
+{
+  Scheme_Object so;
+
+  /* The `mode` determines which of the union cases below
+     is active, corresponding to information for a particular
+     pass: */
+  unsigned int mode : 3;
```
samth: Does `mode` take on values from the `#define`s below (such as `SCHEME_VAR_MODE_LETREC_CHECK`)? If so, maybe a pointer there would be a good addition to this comment, but I'm confused about why there are more than 3 of those modes, since there are 3 cases in the union.

mflatt: I'll extend the comment. Some modes (NONE and COMPILE) don't currently need extra fields.

samth:
Also, I'm curious what the worst design flaw is, now that the second-worst is going away. :)

mflatt:
@samth: ".c"

mflatt:
The revised commit fixes a problem that disabled closure conversion in some cases.

gus-massa:
It's like meeting an old friend after five years ...

I'm still reading here and there, but I like the changes. It's nice to see that all the offsets and deltas disappeared.

I only found that three lines are not necessary: [optimize.c#L4192-L4194](https://github.com/mflatt/racket/blob/72b5206f67b492d03738cf757d3395e9fec0758e/racket/src/racket/src/optimize.c#L4192-L4194)

gus-massa:
Just another nitpick, there is a typo in the title of the commit: repeRsentation -> repResentation

gus-massa:
I think that `shift_fuel` and similar can be removed. They are defined in [optimize.c#L76](https://github.com/mflatt/racket/blob/optimizer/racket/src/racket/src/optimize.c#L76) and [optimize.c#L102](https://github.com/mflatt/racket/blob/optimizer/racket/src/racket/src/optimize.c#L102).

They are copied multiple times from one to another, but they are never used for something interesting now.

stchang:
Unrelated to this specific pr, just playing around with the optimizer. Would it make sense to optimize (binary) `=` to `eq?` for some cases?

gus-massa:
@stchang : I think I added that last week, see [optimize.c#L3875](https://github.com/racket/racket/blob/master/racket/src/racket/src/optimize.c#L3875)

It changes `=` to `unsafe-fx=` when the optimizer knows that both arguments are fixnums. It is not posible to do this if one argument isn't known, because `=` may rise an error.

For example

    (lambda (s) (= (string-length s) 0))  ==>  (lambda (s) (unsafe-fx= (string-length s) 0))

stchang:
@gus-massa Ah nice, ok, makes sense.

mflatt:
Since the latest version passes tests and benchmarks, I plan to merge it (as a squashed commit) tomorrow.

gus-massa:
What about something like

    (check-equal? (number? (equal-hash-code (tuple 5)) #t)
    (check-equal? (equal-hash-code (tuple 5)) (equal-hash-code (tuple 5)))
    [...]

and perhaps also these additional tests that may fail if you re unlucky

    (check-not-equal? (equal-hash-code (tuple 5))) 0)
    (check-not-equal? (equal-hash-code (tuple 5)) (equal-hash-code (tuple 6)))
    [...]


Perhaps you can piggyback a change: `start_simltaneous --> start_simUltaneous`.

I think that `optimize_info_lookup` still does too many things, but the new code is much shorter than the previous, so it will be easier to refactor. (In particular, I'm thinking about splitting the `register_use_at` at the last line, but I still have to think about the details...)

LGTM :)

PR #1242, opened 2016-02-16
fix (combinations n k) bug

bennn:
Now using Gosper's hack to enumerate length k binary numbers.
New implementation is shorter & a little more obviously correct
(if you trust the bit-twiddling)
https://en.wikipedia.org/wiki/Combinatorial_number_system#Applications

### DIFF
--- a/racket/collects/racket/list.rkt
+++ b/racket/collects/racket/list.rkt
@@ -604,69 +604,40 @@
   (define v (list->vector l))
   (define N (vector-length v))
   (define N-1 (- N 1))
-  (define gen-combinations
+  (define (vector-ref/bits v b)
+    (for/fold ([acc '()])
+              ([i (in-range N-1 -1 -1)])
+      (if (bitwise-bit-set? b i)
+        (cons (vector-ref v i) acc)
+        acc)))
+  (define-values (first last incr)
     (cond
-      [(not k)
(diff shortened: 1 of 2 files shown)


samth:
The link to Wikipedia should be in the source code.

bennn:
Added [wiki link](https://github.com/racket/racket/pull/1242/files#diff-607ab65a73af9b2046acf4cfbb443282R624)

racket/collects/racket/list.rkt
```diff
@@ -604,69 +604,40 @@
...
+     [else
+      ;; Enumerate numbers with `k` ones, smallest to largest
+      (define first (- (expt 2 k) 1))
+      (define gospers-hack ;; https://en.wikipedia.org/wiki/Combinatorial_number_system#Applications
+        (if (zero? first)
+          add1
+          (lambda (n)
+            (let* ([u (bitwise-and n (- n))]
+                   [v (+ u n)])
+              (+ v (arithmetic-shift (/ (bitwise-xor v n) u) -2))))))
```
stamourv: Should this `/` be `quotient` to ensure the same behavior as that Wikipedia code?

stamourv:
LGTM
Please close PR15251 when you push.

bennn:
merged :tada:

Issue #980, opened 2015-07-23
string->url doesn't parse literall IPv6 addresses


tim-brown:
The following URL: `ldap://[2001:db8::7]/c=GB?objectClass=one&objectClass=two` does not parse correctly. The `host` should be `[2001:db8::7]`. Instead the URL parses as:
```
URL: "ldap://[2001:db8::7]/c=GB?objectClass=one&objectClass=two"
scheme:         "ldap"
host:           "[2001"
path-absolute?: #f
path  bits:     ("db8::7]" "c=GB")
query:          ((objectClass . "one") (objectClass . "two"))
```
I'm sure it's simple enough to change, but there are a number of pull requests around this area (e.g. https://github.com/plt/racket/pull/948); so maybe this should be done when things settle down.

(Issue arose because I've just tried to implement: http://rosettacode.org/wiki/URL_parser)

Blaisorblade:
:+1:
Any chance things did settle down?
-- A fellow Racket user

Phlosioneer:
This is still an issue. The URL parser should look for square brackets. There are currently no pull requests involving the URL parser. The string->url function is defined [here] (https://github.com/racket/racket/blob/e6a0caa1471820ae1f8a76985858b197a35763a3/racket/collects/net/url-string.rkt). I'll try and see if I can learn enough racket to fix it, but I'm still rather new to the language.

PR #1243, opened 2016-02-16
Added IPv6 support to the string->url function. Passes all tests.

Phlosioneer:
Made the minor edits to the url parser to support IPv6 literals. The net library already supports IPv6 addresses; it just couldn't be fed them via string. Fixes issue #980 . A matching pull request to the racket/net repo will be made in a minute or so, which adds the tests for the new use cases in this code.

### DIFF
--- a/racket/collects/net/url-string.rkt
+++ b/racket/collects/net/url-string.rkt
@@ -176,17 +176,20 @@
            "(?:"              ; | / user-at-opt
            "([^/?#@]*)"       ; | | #2 = user-opt
            "@)?"              ; | \
-           "([^/?#:]*)?"      ; | #3 = host-opt
+           "(\\["             ; | / #3 = ipv6-host-opt
+           "(?:.*)"           ; | | hex-addresses
+           "\\])?"            ; | \
+           "([^/?#:]*)?"      ; | #4 = host-opt
            "(?::"             ; | / colon-port-opt
-           "([0-9]*)"         ; | | #4 = port-opt
+           "([0-9]*)"         ; | | #5 = port-opt
            ")?"               ; | \
            ")?"               ; \
-           "([^?#]*)"         ; #5 = path
+           "([^?#]*)"         ; #6 = path
            "(?:\\?"           ; / question-query-opt
-           "([^#]*)"          ; | #6 = query-opt
+           "([^#]*)"          ; | #7 = query-opt
            ")?"               ; \
            "(?:#"             ; / hash-fragment-opt
-           "(.*)"             ; | #7 = fragment-opt
+           "(.*)"             ; | #8 = fragment-opt
            ")?"               ; \
            "$")))
 
@@ -194,7 +197,7 @@
 ;; Original version by Neil Van Dyke
 (define (string->url str)
   (apply
-   (lambda (scheme user host port path query fragment)
+   (lambda (scheme user ipv6host host port path query fragment)
      (when (and scheme (not (regexp-match? #rx"^[a-zA-Z][a-zA-Z0-9+.-]*$"
                                            scheme)))
        (url-error "Invalid URL string; bad scheme ~e: ~e" scheme str))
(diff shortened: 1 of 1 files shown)


racket/collects/net/url-string.rkt
```diff
@@ -176,17 +176,20 @@
            "(?:"              ; | / user-at-opt
```
Blaisorblade: [The comment above](https://github.com/racket/racket/pull/1243/files#diff-69b1d68b3a51a4a74ecf364822d21099L165) claims this regexp matches some RFC — shouldn't that comment be updated?

racket/collects/net/url-string.rkt
```diff
@@ -176,17 +176,20 @@
...
            "([^/?#@]*)"       ; | | #2 = user-opt
            "@)?"              ; | \
-           "([^/?#:]*)?"      ; | #3 = host-opt
+           "(\\["             ; | / #3 = ipv6-host-opt
+           "(?:.*)"           ; | | hex-addresses
+           "\\])?"            ; | \
+           "([^/?#:]*)?"      ; | #4 = host-opt
```
Blaisorblade: This reads like `(ipv6-host-regexp)?(host-regexp)?`, so allows both to be present and, below, discards the host using the IPv6 host instead. Don't we rather want to use alternation, as in `((ipv6-regexp)|(host-regexp))?` (allowing either one, or nothing since the host is optional, but not both)?

The [PR with tests](https://github.com/racket/net/pull/8/files) lacks negative tests against this.

racket/collects/net/url-string.rkt
Blaisorblade: I've actually found the pointed-to appendix (https://tools.ietf.org/html/rfc3986#appendix-B), and learned that it does support IPv6. In fact, that regexp uses `([^/?#]*)` for the host part; the custom adaption here `([^/?#:]*)` breaks that support by excluding `:`.

Blaisorblade:
Link to PR: https://github.com/racket/net/pull/8

As an outsider fellow user, unfamiliar with the codebase: it feels bad that tests are in another repo, but I'm not familiar enough with the codebase to judge this. But that's certainly not your responsibility.

I've added some comments on the PRs themselves — I'm sure this needs more test. (Again, as an outsider fellow user).

Phlosioneer:
@Blaisorblade It does not, in fact, match IPv6 addresses in the previous implementation. The capture for hosts excludes the ability of having a colon in the host name, which is required for IPv6.

Yeah, I should use the | operator in regex. I had forgotten it existed; still a little new to all of this. I'll update the PR tonight.

samth:
@Blaisorblade Yes, it's unfortunate that some of `net` is in this repo, and some in `racket/net`, and that therefore either the tests have to be split up, or some of the implementation is separate from the tests.

Perhaps there's a better compromise here to be found -- I'm open to suggestions.

Blaisorblade:
> @Blaisorblade It does not, in fact, match IPv6 addresses in the previous implementation. The capture for hosts excludes the ability of having a colon in the host name, which is required for IPv6.

I meant that the regexp from the RFC does match IPv6, unlike the code. The previous code explains the mismatch:
> the user@host:port is parsed here

Parsing that needs to find the `:` separating the host and the port, hence the change, which is only correct for IPv4 but not IPv6.

@samth I moved the discussion to #1245, since it shouldn't block this PR.

Phlosioneer:
@Blaisorblade I'm not sure I follow. Are you saying that I need to put more stuff in my PR to recognize ipv6, or are you just commenting on why the current implementation is the way it is?

Blaisorblade:
Sorry.
> are you just commenting on why the current implementation is the way it is?

Yeah. The point is, that might help to figuring out how to best change it.

In particular: could you please update the comment to match the new code?

> Are you saying that I need to put more stuff in my PR to recognize ipv6

Not necessarily. But we do need more tests to know this actually recognizes IPv6.

Phlosioneer:
Alright. I'll do both those things tonight. Thanks for clarifying ^w^

mflatt:
@Phlosioneer The discussion gives me the impression that you intend to make more changes. Is that right, or is this pull request ready to be merged in your view? Thanks!

Phlosioneer:
It's functionally ready, and follows the IPv6 standard to the letter. It needs better commenting / code style. I lost interest in coding in racket, and thus didn't finish this. Sorry. You can probably fork the request, add the 3-5 comments needed, and pull-request that instead. I have no intention of returning to this.

rfindler:
Thank you very much for your contribution!

mflatt:
Yes, thanks! I have a revision that uses `|` as @Blaisorblade suggests and also makes `url->string` work, and I expect to merge it tomorrow.

PR #1246, opened 2016-02-18
Remove unused variables to eliminate compiler warnings.

samth:
Creating a PR for more testing and to see if I deleted anything important.


mflatt:
Looks good – thanks!

samth:
Note that http://drdr.racket-lang.org/34007/pkg-src/build/make has two other warnings for possibly-uninitialized variables -- they looked ok to me, but I wasn't sure what the best way to silence the warning was.

mflatt:
@samth If you'd like to fix those, I think the right changes are https://github.com/mflatt/racket/commit/6961c3734adacca735d18d102988990e467a49fc

PR #1248, opened 2016-02-19
Special case for small hashes in scheme_unsafe_hash_tree_start/next

gus-massa:
The index of the iteration is like a list of fixnums and internal hashes. The idea is to encode it using only integers to avoid allocations. This is useful for small hashes and makes the iteration ~50% faster. For long hashes, we should keep the index in the list representation because the additional searches make the code slower.

I didn't squash the commits, so each pass can be tested and benchmarked separately.

The first commit adds the change `(#hash i0) --> i0` so all the hashes with less than 32 elements should use only numbers without allocation.

The seccond ommit  adds the change `(#hash i1 #hash i0) --> (+ (<< i1 6) (<< 1 5) i0)`. This is useful for hashes of a few hundred elements, but the internal structure is more complicated, so an unlucky hash may need a deeper tree in spite it is not so big.

The third ommit adds the change `(#hash i2 #hash i1 #hash i0) --> (+ (<< i2 12) (<< 1 11) (<< i1 6) (<< 1 5) i0)`.

I could add more steps, but it's boring and the results are not promising. (See benchmark below.) I think it's better to add only the first two cases, and leave the trees that use `(#hash i2 #hash i1 #hash i0)` unchanged.

This still needs some cleanup and I'm not sure if all the chaperone cases are added in the right commit.

--

I'm using this benchmark. (The first loop is only a baseline to ensure that my machine is not doing something very strange between comparisons.) (I have the times with the internal `when` uncommented. The times are bigger, but the results are equivalent.)

    #lang racket/base
    
    (require racket/unsafe/ops)
    
    (define tot 100000000)
    
    (for ([z (in-range 1)])
      (printf "for ~a: " tot)
      (collect-garbage)(collect-garbage)(collect-garbage)
      (time
       (for/and ([j (in-range 10)])
         (for/and ([i (in-range tot)])
           (> i -1))))

      (for ([h-len (in-list '(10 100 1000 10000 100000 1000000))])
    
      (define old (build-list h-len (lambda (x) (cons x (number->string x)))))
      (define h (make-immutable-hash old))
    
      (printf "~a x ~a: " h-len (/ tot h-len))
      (collect-garbage)(collect-garbage)(collect-garbage)
      (time
       (for ([i (in-range (/ tot h-len))])
         (let loop ([h-idx (unsafe-immutable-hash-iterate-first h)])
           (when h-idx
             #;(when (eq? (unsafe-immutable-hash-iterate-key h h-idx)
                          (unsafe-immutable-hash-iterate-value h h-idx))
                 (display "*"))
             (loop (unsafe-immutable-hash-iterate-next h h-idx)))))))
    )

Results:
 
0 = Master

    for 100000000: cpu time: 16068 real time: 16193 gc time: 0
    10 x 10000000: cpu time: 21373 real time: 21856 gc time: 297
    100 x 1000000: cpu time: 24383 real time: 24554 gc time: 421
    1000 x 100000: cpu time: 30903 real time: 31262 gc time: 720
    10000 x 10000: cpu time: 24616 real time: 25163 gc time: 548
    100000 x 1000: cpu time: 25131 real time: 25506 gc time: 374
    1000000 x 100: cpu time: 26567 real time: 26926 gc time: 63

1 = only `(#hash i0) --> i0`

    for 100000000: cpu time: 16583 real time: 16848 gc time: 0
    10 x 10000000: cpu time: 9610 real time: 9844 gc time: 0
    100 x 1000000: cpu time: 25350 real time: 25724 gc time: 656
    1000 x 100000: cpu time: 32370 real time: 32885 gc time: 744
    10000 x 10000: cpu time: 25538 real time: 25989 gc time: 606
    100000 x 1000: cpu time: 26083 real time: 26598 gc time: 419
    1000000 x 100: cpu time: 27752 real time: 28252 gc time: 202

2 = add `(#hash i1 #hash i0) --> (+ (<< i1 6) (<< 1 5) i0)`

    for 100000000: cpu time: 16224 real time: 16583 gc time: 0
    10 x 10000000: cpu time: 9766 real time: 9874 gc time: 0
    100 x 1000000: cpu time: 10234 real time: 10389 gc time: 0
    1000 x 100000: cpu time: 33166 real time: 33868 gc time: 644
    10000 x 10000: cpu time: 25818 real time: 26286 gc time: 549
    100000 x 1000: cpu time: 26598 real time: 27051 gc time: 452
    1000000 x 100: cpu time: 27705 real time: 28096 gc time: 62

3 = add `(#hash i2 #hash i1 #hash i0) --> (+(<< i2 12) (<< 1 11) (<< i1 6) (<< 1 5) i0)`

    for 100000000: cpu time: 16972 real time: 17644 gc time: 0
    10 x 10000000: cpu time: 10311 real time: 11029 gc time: 0
    100 x 1000000: cpu time: 10905 real time: 11170 gc time: 0
    1000 x 100000: cpu time: 33883 real time: 34898 gc time: 856
    10000 x 10000: cpu time: 26957 real time: 27643 gc time: 669
    100000 x 1000: cpu time: 27612 real time: 28564 gc time: 294
    1000000 x 100: cpu time: 29765 real time: 30685 gc time: 173

### DIFF
--- a/racket/src/racket/src/list.c
+++ b/racket/src/racket/src/list.c
@@ -4118,30 +4118,34 @@ Scheme_Object *unsafe_scheme_hash_tree_iterate_next(int argc, Scheme_Object *arg
   
   if (SCHEME_NP_CHAPERONEP(o)) o = SCHEME_CHAPERONE_VAL(o);
 
-  return scheme_unsafe_hash_tree_next(argv[1]);
+  return scheme_unsafe_hash_tree_next((Scheme_Hash_Tree *)o, argv[1]);
 }
 Scheme_Object *unsafe_scheme_hash_tree_iterate_key(int argc, Scheme_Object *argv[])
 {
   Scheme_Object *obj = argv[0], *args = argv[1], *key;
-  Scheme_Hash_Tree *subtree = (Scheme_Hash_Tree *)SCHEME_CAR(args);
-  int i = SCHEME_INT_VAL(SCHEME_CADR(args));
+  Scheme_Hash_Tree *subtree;
+  int i;
+  
+  scheme_unsafe_hash_tree_subtree(obj, args, &subtree, &i);
   key = subtree->els[i];
   
   if (SCHEME_NP_CHAPERONEP(obj))
-    return chaperone_hash_key("unsafe-weak-hash-iterate-key", obj, key);
+    return chaperone_hash_key("unsafe-immutable-hash-iterate-key", obj, key);
   else
     return key;
 }
 Scheme_Object *unsafe_scheme_hash_tree_iterate_value(int argc, Scheme_Object *argv[])
 {
   Scheme_Object *obj = argv[0], *args = argv[1];
-  Scheme_Hash_Tree *subtree = (Scheme_Hash_Tree *)SCHEME_CAR(args);
-  int i = SCHEME_INT_VAL(SCHEME_CADR(args));
+  Scheme_Hash_Tree *subtree;
+  int i;
+
+  scheme_unsafe_hash_tree_subtree(obj, args, &subtree, &i);
 
   if (SCHEME_NP_CHAPERONEP(obj)) {
     Scheme_Object *chap_key, *chap_val;
-    chaperone_hash_key_value("unsafe-weak-hash-iterate-value",
-			     obj, subtree->els[i], &chap_key, &chap_val, 0);
+    chaperone_hash_key_value("unsafe-immutable-hash-iterate-value",
+                             obj, subtree->els[i], &chap_key, &chap_val, 0);
     return chap_val;
   } else {
     int popcount;
@@ -4151,15 +4155,17 @@ Scheme_Object *unsafe_scheme_hash_tree_iterate_value(int argc, Scheme_Object *ar
 }
 Scheme_Object *unsafe_scheme_hash_tree_iterate_pair(int argc, Scheme_Object *argv[])
 {
-  Scheme_Object *obj = argv[0], *args = argv[1];
-  Scheme_Hash_Tree *subtree = (Scheme_Hash_Tree *)SCHEME_CAR(args);
-  int i = SCHEME_INT_VAL(SCHEME_CADR(args));
-  Scheme_Object *key = subtree->els[i];
+  Scheme_Object *obj = argv[0], *args = argv[1], *key;
+  Scheme_Hash_Tree *subtree;
+  int i;
+
+  scheme_unsafe_hash_tree_subtree(obj, args, &subtree, &i);
+  key = subtree->els[i];
 
   if (SCHEME_NP_CHAPERONEP(obj)) {
     Scheme_Object *chap_key, *chap_val;
-    chaperone_hash_key_value("unsafe-weak-hash-iterate-pair",
-			     obj, subtree->els[i], &chap_key, &chap_val, 0);
+    chaperone_hash_key_value("unsafe-immutable-hash-iterate-pair",
+                             obj, subtree->els[i], &chap_key, &chap_val, 0);
     return scheme_make_pair(chap_key, chap_val);
   } else {
     Scheme_Object *val;
@@ -4171,14 +4177,16 @@ Scheme_Object *unsafe_scheme_hash_tree_iterate_pair(int argc, Scheme_Object *arg
 }
 Scheme_Object *unsafe_scheme_hash_tree_iterate_key_value(int argc, Scheme_Object *argv[])
 {
-  Scheme_Object *obj = argv[0], *args = argv[1], *res[2];
-  Scheme_Hash_Tree *subtree = (Scheme_Hash_Tree *)SCHEME_CAR(args);
-  int i = SCHEME_INT_VAL(SCHEME_CADR(args));
-  Scheme_Object *key = subtree->els[i];
+  Scheme_Object *obj = argv[0], *args = argv[1], *key, *res[2];
+  Scheme_Hash_Tree *subtree;
+  int i;
+
+  scheme_unsafe_hash_tree_subtree(obj, args, &subtree, &i);
+  key = subtree->els[i];
 
   if (SCHEME_NP_CHAPERONEP(obj)) {
-    chaperone_hash_key_value("unsafe-weak-hash-iterate-pair",
-			     obj, subtree->els[i], &res[0], &res[1], 0);
+    chaperone_hash_key_value("unsafe-immutable-hash-iterate-pair",
+                             obj, subtree->els[i], &res[0], &res[1], 0);
   } else {
     Scheme_Object *val;
     int popcount;
--- a/pkgs/racket-test-core/tests/racket/hash.rktl
+++ b/pkgs/racket-test-core/tests/racket/hash.rktl
@@ -212,6 +212,27 @@
   (test-hash-iters-generic lst3 lst4)
   (test-hash-iters-specific lst3 lst4))
 
+
+;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+; Use keys that are a multile of a power of 2 to
+; get "almost" collisions that force the hash table
+; to use a deeper tree.
+
+(let ()
+  (define vals (for/list ([j (in-range 100)]) (add1 j)))
+  (define sum-vals (for/sum ([v (in-list vals)]) v))
+  (for ([shift (in-range 150)])
+    (define keys (for/list ([j (in-range 100)])
+                   (arithmetic-shift j shift)))
+    ; test first the weak table to ensure the keys are not collected
+    (define ht/weak (make-weak-hash (map cons keys vals)))
+    (define sum-ht/weak (for/sum ([v (in-weak-hash-values ht/weak)]) v))
+    (define ht/mut (make-hash (map cons keys vals)))
+    (define sum-ht/mut (for/sum ([v (in-mutable-hash-values ht/mut)]) v))
+    (define ht/immut (make-immutable-hash (map cons keys vals)))
+    (define sum-ht/immut (for/sum ([v (in-immutable-hash-values ht/immut)]) v))
+    (test #t = sum-vals sum-ht/weak sum-ht/mut sum-ht/immut)))
+
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
 (let ()
--- a/racket/src/racket/src/hash.c
+++ b/racket/src/racket/src/hash.c
@@ -2738,80 +2738,189 @@ Scheme_Object *scheme_hash_tree_next_pos(Scheme_Hash_Tree *tree, mzlonglong pos)
 # define HAMT_TRAVERSE_NEXT(i) ((i)+1)
 #endif
 
+#define mzHAMT_MAX_INDEX_LEVEL 4 /* For the compressed form of the index */
+
 /* instead of returning a pos, these unsafe iteration ops */
 /* return a view into the tree consisting of a: */
 /*   - subtree */
 /*   - subtree index */
 /*   - stack of parent subtrees and indices */
+/* For small hashes, it uses a compressed representation as fixnums */
+/*   - (#h i0) --> (+ (<< 1 5) i0) */
+/*   - (#h i1 #h i0) --> (+ (<< 1 10) (<< i1 5) i0) */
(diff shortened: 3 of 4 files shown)


stchang:
Ah nice. I forgot you can still use the original hash argument. 

I haven't looked at the changes in detail, but why do you need the extra `1` flag between the indices, ie, cant you just keep masking and shifting by 5 until you get 0?

gus-massa:
Each index is between 0 and 31 (both extremes inclusive). The flag is necessary to distinguish between 

    (#hash 0 #hash 3)  -->  (+ (<< 0 6) (<< 1 5) 3) = 35
    (#hash 3)  --> 3

I was copying the previous implementation blindly ... I didn't think too much about this, so I'll think out loud: 

In both cases, `scheme_unsafe_hash_tree_next` goes to `(#hash 2)` or if it's necessary to `(#hash <count> #hash 2)` or deeper if needed, so for this function they are equivalent.
  
The difference is in `scheme_unsafe_hash_tree_key/value/...`. In `(#hash 3)` they just read whatever is in the third slot, but in `(#hash 0 #hash 3)` it takes the tree that is in the third slot and reads it's zeroth slot. One possibility is to add a check of `(HASHTR_SUBTREEP(ht->els[i]) || HASHTR_COLLISIONP(ht->els[i]))` to all this functions, but I think that the current implementation with the flag is easier.

With the current proposal that use 5+1 bits, we could group up to 5 levels in a fixnum in all platforms. Without the flag we could group up to 6. Anyway, I think that it's better not to add support for more levels, because the I don't see any improvement and I expect that at some point it will be worst to go from the original hash that using the last subtree.

I can try to implement the version for the next level and see if there is an improvement in the hash with 1000 elements. I guess that when the hashes are big the structure is unpredictable and this additional deep will not help too much.

stchang:
You are right, thanks. I was not accounting for the 0 case.

racket/src/racket/src/list.c
```diff
@@ -4118,13 +4118,28 @@ Scheme_Object *unsafe_scheme_hash_tree_iterate_next(int argc, Scheme_Object *arg
   
   if (SCHEME_NP_CHAPERONEP(o)) o = SCHEME_CHAPERONE_VAL(o);
 
-  return scheme_unsafe_hash_tree_next(argv[1]);
+  return scheme_unsafe_hash_tree_next((Scheme_Hash_Tree *)o, argv[1]);
```
stchang: Oops, looks like the chaperone check above was not needed. But with your changes it's needed again.

racket/src/racket/src/list.c
```diff
@@ -4118,13 +4118,28 @@ Scheme_Object *unsafe_scheme_hash_tree_iterate_next(int argc, Scheme_Object *arg
...
+      subtree = (Scheme_Hash_Tree *)obj;
+    i = SCHEME_INT_VAL(args);
+    while (i & (1<<5)) {
+      subtree = (Scheme_Hash_Tree *)subtree->els[i & ((1<<5)-1)];
+      i = i >> 6;
+    }
+  }
   key = subtree->els[i];
   
   if (SCHEME_NP_CHAPERONEP(obj))
```
stchang: Another typo in my code (if you would be so kind to fix), the calls to `chaperone_hash_key` in `unsafe_scheme_hash_tree_` fns should use name `"unsafe-immutable-hash-iterate-"` instead of `"unsafe-weak-"`.

racket/src/racket/src/list.c
```diff
@@ -4135,8 +4150,23 @@ Scheme_Object *unsafe_scheme_hash_tree_iterate_key(int argc, Scheme_Object *argv
...
+  } else {
+    if (SCHEME_NP_CHAPERONEP(obj))
+      subtree = (Scheme_Hash_Tree *)SCHEME_CHAPERONE_VAL(obj);
+    else
+      subtree = (Scheme_Hash_Tree *)obj;
+    i = SCHEME_INT_VAL(args);
+    while (i & (1<<5)) {
+      subtree = (Scheme_Hash_Tree *)subtree->els[i & ((1<<5)-1)];
+      i = i >> 6;
+    }
```
stchang: Maybe define `#define`s for these constants?

racket/src/racket/src/list.c
```diff
@@ -4135,8 +4150,23 @@ Scheme_Object *unsafe_scheme_hash_tree_iterate_key(int argc, Scheme_Object *argv
...
+  }
```
stchang: And a separate function to compute the subtree?

gus-massa:
> Maybe define `#defines` for these constants?

I agree, but I can't find where is defined the original `32`. Where is it?

stchang:
I had a hard time finding it as well. I believe it's `mzHAMT_LOG_WORD_SIZE` and `mzHAMT_WORD_SIZE` in hash.c.

https://github.com/racket/racket/blob/master/racket/src/racket/src/hash.c#L2402

mflatt:
@gus-massa @stchang I think part of the problem is that anything that relies on the representation of a hash-table variant was originally meant to be in "hash.c", while "list.c" combined those variants with generic operations and Racket primitives. That's why `mzHAMT_WORD_SIZE` is defined only for "hash.c".

gus-massa:
I added the next two levels, and for my surprise the time of the benchmarks decreased (except for the hash with 1000000 elements that is too big and needs more levels). The code was very repetitive, so I reimplemented it using real loops instead of the hand unrolled version. The final version has too many changes, so I rebased and squashed all of them.

Following the suggestions, I created a function `scheme_unsafe_hash_tree_subtree` to reconstruct the tree from the arguments. I put it in "hash.c" so it can use the constant `mzHAMT_LOG_WORD_SIZE`. I marked it as `XFORM_NONGCING`, is that correct? I'm happy with this because now all the logic to save and load the subtrees is very close in case someone wants to improve it.

One detail that I don't like is that this functions takes a `Scheme_Object *` as first argument to extract the tree from a chaperones, but the other functions use a `Scheme_Hash_Tree *` and the chaperone logic is in "list.c". It's a small inconsistency I don't like.

I created a new internal constant `mzHAMT_MAX_INDEX_LEVEL` in "hash.h". It regulates the longest lists that are compressed to fixnums. I finally defined it as `4` because it the maximal number that guaranty that the result enters in a fixnum in 32 bits. For some of my tests, it was useful to change it to `2`, to get small examples of hash that need the index represented as a link. I'm slightly worried that this code will be not tested under normal circumstances.

[Suggestions about better names are welcome.]

samth:
Presumably some large test hash tables will exercise the code you're worried about?

gus-massa:
The easy solution is to add the case of a hash with `1000000` elements in [pkgs/racket-test-core/tests/racket/hash.rktl](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/hash.rktl) . I just added it. But I estimate that each test adds .1 seccods, and it has many tests with each hash. (I think it's good to have keep the small `10` test to use other code paths. I added the `100`, `1000`, `10000` because they are almost free.)

One alternative solution is to keep the current test in core, and add the bigger test in a new file [pkgs/racket-test/tests/racket/hash-big.rkt](https://github.com/racket/racket/blob/master/pkgs/racket-test/tests/racket/hash-big.rkt)

rfindler:
If there is some structure to this repetitive code, could you think of
designing a little Racket language and a compiler from it to (a small
subset of) C? Then you could control which sizes are the special ones and
you could design the code in such a way that it is unlikely to have a bug
in a large table unless there is also a bug in a small table?

stchang:
> why do you need the extra 1 flag between the indices?

Sorry, still thinking about this. Could you use only 1 bit overall instead of 1 bit between each index? 

Specifically, add `1 << (mzHAMT_LOG_WORD_SIZE * level)` before returning the index.

And then when extracting the indices, you could do `while(i)` instead of `while (i & (1<<mzHAMT_LOG_WORD_SIZE))`.

gus-massa:
@rfindler : It was a straightforward repetition, like:

    i1 = something(i0)
    i2 = something(i1)
    i3 = something(i2)
    i4 = something(i3)

where `something` was actually a few lines of code with `if`s and some logic, so I replaced it with

    #define MAX 4
    for (n =0; n < MAX, n++)
      i[n+1] = something(i[n])

I initially expected that `MAX` will be `2` or `1` or `3`, so it was easier to cut and paste, especially because the first and last case are special. But if `MAX` is `4` it becomes a bad idea.

--

The code is doing a depth first traversal of the internal representation of the hash, and saves the "stack" in a `list` to implement the `start/next` operations and be able to continue the traversal later. Here I'm replacing the `list` with local `C` variables to avoid allocating the list in the `C` heap, and the result of the `start/next` operations is an encoding of the values in the bits of a fixnum.

So the internal storage are very different and I think it's not possible to make a code that automate this. Moreover, the transition from variables to list is very error prone, and I have to fix it many times. (The transition from list to variables is also problematic, but am I only using the simplest case, that is when the traversal goes to the upper node. For this aplication, it's enough and easy to implement.) 

For testing, I changed the value of the constant `MAX`, so the transition from the short representation to the long representation is with hashes of a few hundreds or thousand of elements, so it's easy to test. With `MAX=4` the transition is with hashes of about one millon elements. (It depends on the internal details and collisions, so the number is only an aproximation.)

samth:
@gus-massa Could you create a custom hash table, or a struct with a custom hash function, to generate more collisions and thus test the code with fewer elements?

gus-massa:
I'm don't know too much of the details of the hashs, perhaps @stchang or @mflatt can build an example of a table with more problems. Perhaps they can use as keys an `struct` with a very bad election of `gen:equal+hash`?

samth:
I think this structure will always collide, no matter what instances you give it: 

```
(struct x (a)
  #:property prop:equal+hash
  (list (λ _ #f) (λ _ 1) (λ _ 1)))
```

Is that what you need?

gus-massa:
With this bad hashing function, all of them go to the first slot in the first node of the tree. So it's like reducing one level in the test. So probably the `1000000` case would be enought, if all those structs don't use too much memory. 

(But I'm thinking about @stchang idea to reduce the flags, this can add room for an aditional level. I'll try a litle and reply later.)

gus-massa:
I was thinking about a more evil list of keys. After skiming over the code and some time minimizing the initial idea, I got this:

    (define evil (for/list ([j (in-range 1 3)])
                   (arithmetic-shift j (* 6 5))))
    (define old (map (lambda (x) (cons x (number->string x))) evil))   
    (define h (make-immutable-hash old))

The hash code are different, but they colides at all the first levels, so the internal tree is very deep.

I'll try to cleanup it and add it to [pkgs/racket-test-core/tests/racket/hash.rktl](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/hash.rktl)

gus-massa:
@stchang : Sorry for the delay I was testing your idea of using only one flag, and I think I have a version of it, but it still has a small bug.  I had to use `while(i >= (1 << (mzHAMT_LOG_WORD_SIZE * 2)))` instead of `while (i)` because the last value of `i` is the position in the internal array. I feel that this new repreentation is slighty faster than my representation, but the difference is small (2% or 3%) and I didn't have enought patience to measure it carefully. I'll finish this and send an update.

The not so good new is that with this representation is not posible to increase the number of levels stored in a fixnum in 32 bits, because they only have 30 bits + 1 sign bit. If we compress a list of 6 levels and the stop bit, it would use `6*5+1=31` bits. And I am not confident that `SCHEME_INT_VAL(scheme_make_integer(1<<31))` does the right thing. So this change doesn't increase the `mzHAMT_MAX_INDEX_LEVEL` constant.

I also tried to add some test with almost-collisions, so the hash has to use a deeper representation. I get some segfaults (in window, 32 bits) in the master branch. I guess it's a problem with the weak hashes when one of the elements is garbage collected at the wrong moment, but I'm not sure. I made a branch with the additional tests in case you want to take a look [gus-massa/16-2-W-Hash-Error](https://github.com/gus-massa/racket/tree/16-2-W-Hash-Error) (See: [Travis](https://travis-ci.org/gus-massa/racket/jobs/111002075#L1711), [AppVeyor](https://ci.appveyor.com/project/gus-massa/racket/build/1.0.405#L3327) )

gus-massa:
I updated the commit with the new representation that uses only one flag, some tests an some tweaks in `scheme_unsafe_hash_tree_start`.

mflatt:
This is ready to merge, right?

gus-massa:
I think so. 

Do I rabase and merge this now?

mflatt:
Yes

PR #1250, opened 2016-02-20
Use normalize-path instead of custom trailing slash remover

takikawa:
This makes the following interaction not produce an internal error:

    raco pkg install -n foo /

I think normalizing the string does hit the filesystem, but the code right above checks if the directory exists anyway so I think it shouldn't error. All the pkg tests seem to pass.

### DIFF
--- a/racket/collects/pkg/private/orig-pkg.rkt
+++ b/racket/collects/pkg/private/orig-pkg.rkt
@@ -19,7 +19,9 @@
     [(link static-link) `(,type
                           ,(path->string
                             (find-relative-path (pkg-installed-dir)
-                                                (simple-form-path src)
+                                                ;; normalize with ending slash
+                                                (path->directory-path
+                                                 (simple-form-path src))
                                                 #:more-than-root? #t)))]
     [(clone) 
      (define-values (transport host port repo branch path)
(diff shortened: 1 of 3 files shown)


takikawa:
I tried using `simple-form-path` but it doesn't actually remove trailing slashes on directory names. I suppose another solution is to change the regexp in the helper (say, to require at least on character before the slash) but I worry there are more corner cases.

mflatt:
I agree that the regexp-based path manipulation is bogus. (I never noticed that function.)

Digging back through the history at https://github.com/jeapostrophe/galaxy, it appears that the slash-removal was added to make `file-name-from-path` work to get the package name, but I think that wasn't the right choice, either, and it's not how things are parsed now. So, maybe the right repair is to get rid of `directory-path-no-slash`?

takikawa:
I also tried getting rid of the slash removal too (by taking out that let expression there) but one of the package update tests fail. Basically it fails to detect that an update with the same package and checksum should be a no-op, because in one case the path has a slash and doesn't in the other. Probably this can be fixed, but I didn't investigate too closely yet.

mflatt:
My next guess: normalize via `path->directory-path` (which makes the slash always there, instead of trying to remove it).

takikawa:
I tried @mflatt's suggestion to add `path->directory-path`, but I added it to `desc->orig-path` instead and took out the normalization in the staging code entirely. This now passes all the tests.

(I would like to add a test for `raco pkg install -n foo /` but I'm not sure there is a good way to test that without using chroot)

mflatt:
This looks like the right strategy to me.

PR #1253, opened 2016-02-21
Add optional failure thunk for regexp and friends

takikawa:
I noticed that the regexp API doesn't provide any easy way to test if a string input is actually a valid regular expression (i.e., without installing an exception handler). This PR adds a failure thunk to the constructors for this purpose.

I ended up adding some new thread local variables to track the failure thunk and also its result. Is this a reasonable approach or is it discouraged to add more data to threads?

Also does the API addition seem reasonable? Or is there a better interface?

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/regexps.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/regexps.scrbl
@@ -198,7 +198,9 @@ Returns @racket[#t] if @racket[v] is a @tech{regexp value} created by
 otherwise.}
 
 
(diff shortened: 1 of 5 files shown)


mflatt:
I think it would make sense to give the handler the string describing the problem with the given string, so that tools like `raco pkg show` can report it back to the user.

The thread-local variables are not great, but they're consistent with the way things are implemented now. The `regerrorproc` and `regerrorval` hold a GCable values, so they need to be registered (once per place) with `REGISTER_SO`, like `regparsestr`.

mflatt:
Merged - thanks!

PR #1254, opened 2016-02-22
add setter variant to saved-errno

rmculpepper:
This makes it easier to make mock foreign functions for testing.


mflatt:
Looks good to me. A test added to "pkgs/racket-test-core/tests/racket/foreign-test.rktl" would make it even better.

rmculpepper:
Looks like `#:saved-errno 'posix` doesn't work on Windows. Shouldn't it?

rmculpepper:
Merged

PR #1260, opened 2016-02-24
sizeof(char) is always 1

mmcco:
POSIX and ANSI specify that char is always 1 byte, and I'm almost
certain that no systems violate this. Regardless, the SIZEOF_CHAR macro
is never used.


mmcco:
I need to change my name and email in the commit. Doing so now. Please don't merge until I get a chance to.  :-)

mflatt:
Merged

PR #1262, opened 2016-02-24
margin-note to say #:auto-value is evaluated once

bennn:
Response to: https://groups.google.com/forum/#!topic/racket-users/-jbLzjygMpg

Sample:
screen shot 2016-02-24 at 17 17 04


bennn:
I'm not sure if the second sentence helps without an example. Maybe better to delete it?

stamourv:
I like it.

bennn:
merged :tada:

PR #1270, opened 2016-02-29
Burn less fuel when inlining small functions (draft)

gus-massa:
Use a simple heuristic to burn a different amount of fuel when the inlined function is small or big.

For "big" procedures the inline fuel is multiplied by `1/2` as before, but for "small" procedures the fuel is multiplied only by `3/4`. The threshold between "big" and "small" is at one half of the maximal threshold for inlining. 

This cut is arbitrary, I'm not sure if it would be better to pick another value. Moreover, perhaps it would be better to use a more complex function, like multiplying by `1 - 1/2 size/max`, but I'm worried about rounding and other details. (Perhaps instead of a linear function it's better to use a logarithmic scale or something.)

Anyway, I choose this function because it's simple to program and simple to prove that it has the minimal properties. If `x`is the fuel, 

* `f(0) = 0`
* `0 <= f(x) < x` (if `x` is positive)
* `x<y  ==>  f(x)<=f(y)`

### DIFF
--- a/racket/src/racket/src/optimize.c
+++ b/racket/src/racket/src/optimize.c
@@ -1809,9 +1809,9 @@ static Scheme_Object *no_potential_size(Scheme_Object *v)
 }
 
 static Scheme_Object *apply_inlined(Scheme_Lambda *lam, Optimize_Info *info,
-				    int argc, Scheme_App_Rec *app, Scheme_App2_Rec *app2, Scheme_App3_Rec *app3,
+                                    int argc, Scheme_App_Rec *app, Scheme_App2_Rec *app2, Scheme_App3_Rec *app3,
                                     int context, Scheme_Object *orig, Scheme_Object *le_prev,
-                                    int single_use)
+                                    int single_use, int size, int threshold)
 /* Optimize the body of `lam` given the known arguments in `app`, `app2`, or `app3` */
 {
   Scheme_IR_Let_Header *lh;
@@ -1827,8 +1827,13 @@ static Scheme_Object *apply_inlined(Scheme_Lambda *lam, Optimize_Info *info,
   if (!expected) {
     /* No arguments, so no need for a `let` wrapper: */
     sub_info = optimize_info_add_frame(info, 0, 0, 0);
-    if (!single_use || lam->ir_info->is_dup)
-      sub_info->inline_fuel >>= 1;
+    if ((!single_use || lam->ir_info->is_dup)
+        && (sub_info->inline_fuel >= 0)) {
+      if (size < (threshold >> 1))
+        sub_info->inline_fuel = (sub_info->inline_fuel >> 1) + (sub_info->inline_fuel >> 2);
+      else
+        sub_info->inline_fuel >>= 1;
+    }
     p = scheme_optimize_expr(p, sub_info, context);
     info->single_result = sub_info->single_result;
     info->preserves_marks = sub_info->preserves_marks;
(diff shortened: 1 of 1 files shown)


mflatt:
I find it hard to evaluate this change. Do you have some programs where this revised heuristic improves performance? Does it have any effect on the traditional benchmarks (in "common")?

gus-massa:
Nope, just gut feelings. I'll try the benchmarks ...

gus-massa:
I got this semi-articficial case:

    With differential fuel burning:
      cpu time: 30295 real time: 31214 gc time: 0
    With constant fuel burning:
      cpu time: 40342 real time: 41069 gc time: 0

Program: It's a variation of [collatz-q.sch](https://github.com/racket/racket/blob/master/pkgs/racket-benchmarks/tests/racket/benchmarks/common/collatz-q.sch)

    #lang racket/base
    (require racket/unsafe/ops)
    
    (define (cycle-even n)
      (unsafe-fx+ 1 (cycle-length (unsafe-fxrshift n 1))))
    
    (define (cycle-odd n)
      (unsafe-fx+ 1 (cycle-length (unsafe-fx+ 1 (unsafe-fx* 3 n)))))
    
    (define (unsafe-fxodd? n)
      (unsafe-fx= 1 (unsafe-fxand n 1)))

    ; Max value for 32bits, but divided by 10 to get some warnings
    (define almost-outside-fixnum 100000000)
    
    (define (cycle-length n)
      (when (unsafe-fx> n almost-outside-fixnum)
        (displayln n))
      (cond
        [(unsafe-fx= n 1) 1]
        [(unsafe-fxodd? n) (cycle-odd n)]
        [else (cycle-even n)]))
    
    ; modify to get a rutime of half a minute
    (define repetitions 1000)
    
    ; keep this constant so all the numbers are fixnums
    (define bigest-case 10000)
    
    (time (let loop ([r 0] [t #f])
            (if (unsafe-fx= r repetitions)
              t
              (loop (unsafe-fx+ r 1)
                    (let loop ([i 1] [v #f])
                      (if (unsafe-fx= i bigest-case)
                        v
                        (loop (unsafe-fx+ i 1)
                              (cycle-length i))))))))

Idea/history of the modifications:

* split the main function into many smaller functions to burn fuel unnecesarily
* use `unsafe-fx` so the time differences look bigger
* segfault???
* Add a `displayln` to try to understand where is the problem.
* I know the Collatz conjeture, but the sequence was more unexpected than what I expected (Obligatory xkcd: https://xkcd.com/710/ )
* Reduce the number so it enters in a fixnum
* Add a second loop to get a run time of 10-20 secconds.

After this, I only added a name and explanation for each constant. I didn't modify the code to increase the difference, so it's posible to get a higer ratio.

mflatt:
I appreciate the example that speeds up, but none of the benchmarks I've tried speed up without modifications. Meanwhile, the total size of bytecode files in "collects/racket/private" grows by 5% (and I think those sources tend to representative). Since a larger bytecode size tends to have a corresponding effect on run-time footprint, and since the performance benefits so far seem elusive, this change isn't looking like a good direction.

gus-massa:
I still think that this is useful, especially for code with a lot of small functions. For example functions that are only a thin wrapper or error check of the real function that does the heavy lift.

So I tried the original `collatz` example in `r6rs `and `htdp-advanced` (with minimal changes) and ... the difference is negligible. Probably less than a 5% of the run time, but I'm not sure in which direction because I was too lazy to repeat this until I get a statistically significant result, and it's too noisy to be sure.

So I'll close this, and I'll try again in a few years if I get some new idea.

PR #1276, opened 2016-03-07
Reduce (let ([x <expr>]) #f) ==> (begin <expr> #f)

gus-massa:
Sometimes the optimizer removes all the references to a variable but it doesn't detect that the variable is unused, so it keeps the definition.

Later, the sfs detects the unused variable so it marks it, but it doesn't remove the let form.

And I think it's enough to reduce `(let ([x <expr>]) x)` to `(values <expr>)` instead of `(begin0 <expr>)` or `(values (begin0 <expr>))`.

Also, more special cases for `(values <expr>)` in `finish_application2`.


mflatt:
This looks good to me.

gus-massa:
> Is it important/intentional that setting these flags was lost in the case of list*? Maybe list* is replaced by values somewhere else?


I think that with `list*` those flags are setted automatically, so I think it was not necessary to set them again. But `values` doesn't set them because it not true when it has more than one argument. I´ll set them for both functions anyway.

mflatt:
It makes sense that the flags are already set for `list*` – so, whichever variant you think is best is fine with me.

PR #1285, opened 2016-03-23
Add list-pair?

gus-massa:
This predicate is useful in the optimizer to track simultaneously the `list?` and `pair?` types of an expression.

The first commit is almost unrelated, but it's necesary for the test cases. In some cases, complex variants of `(if (not <expr>) tb fb)` are not reduced. Extract the type information of the tests in `<expr>` to use it in `tb` and `fb`.

The seccond commit is the main part. I made the new primitive visible, but an alternative is to hide it and remove the docs and JIT.

The third commit is a small change to `comp=?`, because sometimes it's difficult to find the test that is failing.

The fourth commit remove a few unnecesary checks of mutability, because `add_type` checks it anyway.


mflatt:
This looks generally good to me.

I don't have a strong opinion on whether to export `list-pair?`. I would probably leave support for it in the JIT but remove it from the docs and the set of things in `#%kernel` that are exported by `racket/private/pre-base`, but I'm also ok with leaving it as exported by `racket/base` (and documented).

gus-massa:
I hide it. 

In optimize.rktl and basic renamed it as `k:list-pair?` to so it's clear that it's defined in the `#%kernel`, but I'm not sure if this is a good idea.

mflatt:
Looks good, and I think importing as `k:list-pair?` for tests is a good idea.

PR #1287, opened 2016-03-25
Improve Guide example on `flat-named-contract'

leafac:
Previously, on the example, the function was not anonymous, so no
`tempN` would appear on the error message.

The fixed example makes use of `flat-named-contract` on an anonymous
function, which resembles the snippet above it and demonstrates the
purpose of `flat-named-contract` better.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/contracts/simple-function.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/contracts/simple-function.scrbl
@@ -404,11 +404,14 @@ the contract so that error messages become intelligible:
 @interaction[#:eval 
              contract-eval
              (module improved-bank-server racket
-               (define (amount? x) (and (number? x) (integer? x) (>= x 0)))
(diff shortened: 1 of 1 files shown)


stamourv:
@rfindler any thoughts?

rfindler:
I agree that this change should be made. Probably good that someone checks the output, just in case.

leafac:
I manually checked the output, but double checking wouldn't hurt :smile:

stamourv:
I'm working in the contract docs right now. Will push this when I push the rest.

rfindler:
Excellent thanks!

stamourv:
Merged, thanks!

leafac:
Thanks for the feedback and for merging my pull request.

For the reference, here's the commit https://github.com/racket/racket/commit/05292b7e6926e39ff78b4f9d98dc32013aeca41d.

Best.

stamourv:
Thank *you* for improving Racket! :)

PR #1299, opened 2016-04-03
optimizer: clone K in (if (if x tb #f) a K) reduction

gus-massa:
I think this is necessary to use clone in case `K` is a once used variable. Anyway, at this time the value should have been already propagated. I'm worried that for some reason this is not propagated in the first pass, and then `K` is duplicated by this reduction, and then a second optimization pass breaks the code. I couldn't trigger a bug using this, perhaps it's impossible, but I think it's too subtle.

[I was reviving this because I want to include `toplevels `in `scheme_ir_duplicate_ok`, but it's used in a few places and I have to be sure what else would change. I think it's better to wait after the new branch is created.]

In the second commit, I think the check was necessary to avoid problems with the shifting of the local variables, but it's not necessary with the new representation.


mflatt:
These changes make sense to me.

Issue #1295, opened 2016-03-31
Providing other forms with unicode lambda 


jasonhemann:
Racket helpfully provides unicode lambda (`λ`) as a synonym for `lambda`. This is especially helpful in emacs when the input-method is set to racket-unicode. But for similar forms, we're restricted to those that use  `lambda`. As an example: 

```
> ((lambda (x) x) 5)
5
> ((λ (x) x) 5)
5
> ((match-lambda (x x)) 5)
5
> ((match-λ (x x)) 5)
; match-λ: undefined;
;  cannot reference an identifier before its definition
```

One could obviously create one's own version of these identifiers. Or change input methods. But I think it might be helpful to have them available by default. It would keep folk that would want them from needing to roll their own versions, and make it less likely that folk would take these identifiers for some other purpose, and what have you.

stamourv:
Sounds reasonable to me. Want to give it a shot?

jasonhemann:
Sure.

sorawee:
Closing, per #1301.

PR #1301, opened 2016-04-04
adding unicode-lambda forms for match

jasonhemann:
This adds forms `match-λ`, `match-λ*`, and `match-λ**`. Further reflecting upon #1295 the Racket docs pull up 141 entries for `lambda`. That's an awful lot to add. The three identifiers `match-lambda`, `match-lambda*`, and `match-lambda**` are the only ones by default in #lang racket that don't coincide with versions defined elsewhere (unlike `case-lambda`). 

Let's see how these go and maybe take others on an as-desired basis. First pull request to racket, so extra scrutiny warranted.

### DIFF
--- a/racket/collects/racket/match/define-forms.rkt
+++ b/racket/collects/racket/match/define-forms.rkt
@@ -16,8 +16,9 @@
 (provide define-forms)
 
 (define-syntax-rule (define-forms parse-id
-                      match match* match-lambda match-lambda*
-		      match-lambda** match-let match-let*
+                      match match* match-lambda match-λ 
+                      match-lambda* match-λ* match-lambda** match-λ** 
+                      match-let match-let* 
                       match-let-values match-let*-values
 		      match-define match-define-values
                       match-letrec match-letrec-values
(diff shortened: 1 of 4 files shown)


racket/collects/racket/match/define-forms.rkt
```diff
@@ -16,8 +16,9 @@
 (provide define-forms)
 
 (define-syntax-rule (define-forms parse-id
-                      match match* match-lambda match-lambda*
-		      match-lambda** match-let match-let*
+                      match match* match-lambda match-λ
```
bennn: Is it possible to use `(rename-out match-lambda match-λ)` etc here? That would be simpler than using `define-syntaxes`

jasonhemann: Yes. Quite.

sorawee: I think `match-λ` here is correct already. `rename-out` should be used below instead (in `provide`).

@jasonhemann do you still want to make the change?

SWDev2020: @sorawee no, please feel free.

jasonhemann: > I think `match-λ` here is correct already. `rename-out` should be used below instead (in `provide`).
> 
> @jasonhemann do you still want to make the change?
@sorawee no, please feel free.

sorawee: In that case, I propose that we close this PR. Is that OK?

jasonhemann:
Closed, as per discussion.

PR #1302, opened 2016-04-05
Fixs for the optimizer

gus-massa:
The first commit fixs some tests for `eqX?`, they worked anyway, but this version is better.

The second commit moves the update of the `rand `variable after all the possible changes.


mflatt:
LGTM

PR #1305, opened 2016-04-09
optimizer: allow duplication of toplevels

gus-massa:
Reduce `(if (if x y #f) z <top>) => (if x (if y z <top>) <top>)`.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1945,6 +1945,19 @@
                           (if r #t (something-else))))
            '(lambda (x) (if (something) #t (something-else))))
 
+(let ([test-if-if-reduction
+       (lambda (dup)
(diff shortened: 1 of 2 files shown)


mflatt:
Is there any issue with mutable toplevels and `scheme_ir_duplicate_ok()` being used by `scheme_ir_propagate_ok()`?

gus-massa:
I think there is no direct problem with mutable toplevels, but perhaps I'm missing some case.

For `scheme_ir_propagate_ok()`, I moved `scheme_ir_duplicate_ok()` to the bottom of the function and I added a `return 0;` (in [R5820](https://github.com/racket/racket/pull/1305/commits/a6cd58a95dc0c0c9160875a3c4f3662e46e256e4#diff-5e09d7df9102fa81a4736e357a0c3e1aR5820)) to ensure that all the `scheme_ir_toplevel_type`s are handled by the special case in the `if` and they don't fall through to the `scheme_ir_duplicate_ok()`.

mflatt:
Ah, ok – looks good to me.

PR #1306, opened 2016-04-11
More reductions uing types

gus-massa:
**First commit:** Only wrap the argument `x` as `(values x)` when the optimizer is not sure that `x` is a single valued expression. This allows to remove the checks in the calling sites.

**Second commit:** Reduce `(unbox (box x)) => (values x)`. And extend the reductions for `cXr` to the unsafe versions, for example reduce `(unsafe-car (cons x y)) => x`.

**Third commit:** _(don't merge)_ check and save types in unsafe operations. I think this is technicaly correct, but it enables many surprising unexpected reductions, so I think it's better to not merge this. Also, it's very difficult to find examples where this aditional information is useful.
 
For example reduce:

    (lambda (x) (unsafe-car x) (pair? x)) ;infer type
    => (lambda (x) (unsafe-car x) #t)
    => (lambda (x) #t)

and

    (lambda (y) (+ (unsafe-car y) (unbox y)) ;infer error
    => (lambda (y) (unsafe-car y) (unbox y))
    => (lambda (y) (unbox y))

**Fouth commit:** The previous commit breaks one test that uses `(unsafe-car void)`. It's not clear whether  the optimizer is allowed to realize that it's undefined behavior and do strange things after it, but just in case I propose to change the test to `(unsafe-car var)`.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1228,6 +1228,27 @@
            '(lambda (w z) (list w z))
            #f)
 
+(test-comp '(lambda (u v) (car (cons u v)))
+           '(lambda (u v) u))
+(test-comp '(lambda (u v) (unsafe-car (cons u v)))
+           '(lambda (u v) u))
+(test-comp '(lambda (u v) (car (unsafe-cons-list u v)))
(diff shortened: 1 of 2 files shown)


mflatt:
This all looks fine to me, including the third commit.

PR #1309, opened 2016-04-14
Add #:extra-id-name to struct

soegaard:
Don't commit. This is a suggestion.
See  https://groups.google.com/forum/#!topic/racket-dev/ctG3yOUeQcU

### DIFF
--- a/racket/collects/racket/private/define-struct.rkt
+++ b/racket/collects/racket/private/define-struct.rkt
@@ -239,7 +239,8 @@
                            (#:reflection-name . #f)
                            (#:only-constructor? . #f)
                            (#:omit-define-values . #f)
-                           (#:omit-define-syntaxes . #f))]
+                           (#:omit-define-syntaxes . #f)
+                           (#:extra-id-name . #f))]
                  [nongen? #f])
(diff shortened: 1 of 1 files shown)


mflatt:
This looks like a good change, but "id" and "name" seem redundant in `#:extra-id-name`. How about just `#:extra-name`? Should we also add `#:name` (analogous to `#:constructor-name` and `#:extra-construct-name`)?

soegaard:
> This looks like a good change, but "id" and "name" seem redundant in #:extra-id-name. 
> How about just #:extra-name? 

Agree #:extra-name is better.

> Should we also add #:name (analogous to #:constructor-name and #:extra-construct-name)?

I thought about that, but since the name is specified in `(struct name ...)` it seems to
that #:name  is redudant?

mflatt:
I have in mind that `#:name` would override the binding of `name` to static information. I now see that you had made the combination of `#:extra-id-name` and `#:omit-define-syntaxes` provide an alternate name, but I think it's more consistent to add `#:name` and make `#:omit-define-syntaxes` suppress all bindings to static information.

PR #1310, opened 2016-04-16
Rename define-simple-macro from syntax/parse/define to define-syntax-parse-rule

lexi-lambda:
In [a mailing list thread from last year](https://groups.google.com/d/msg/racket-dev/e4WfuZ9JWUk/StzZW7lBCQAJ), it was mentioned that `define-simple-macro` might not be the best name for the `define-syntax-rule` analog exported from `syntax/parse/define`. I was not super satisfied with any of the proposed alternate names, but after considering it for a while, I think `define-syntax-pattern` is a fair one.
1. It is pretty close to `define-syntax-rule`, but distinct enough to not be any more confusing than `syntax-case` vs `syntax-parse`.
2. The term `pattern` is already used a lot within the `syntax/parse` docs, so it is a familiar term (probably more than `rule`, really).
3. Even in isolation, the name pretty accurately describes what the form does.

Obviously, I’ve left the old name in to avoid breaking backwards compatibility, but I’ve updated the docs to mention that the other name is preferred.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
@@ -12,7 +12,7 @@
 
 @defmodule[syntax/parse/define]
 
-@defform[(define-simple-macro (macro-id . pattern) pattern-directive ...
+@defform[(define-syntax-parse-rule (macro-id . pattern) pattern-directive ...
            template)]{
 
 Defines a macro named @racket[macro-id]; equivalent to the following:
@@ -25,11 +25,11 @@ Defines a macro named @racket[macro-id]; equivalent to the following:
 
 @(the-eval '(require syntax/parse/define))
 @examples[#:eval the-eval
-(define-simple-macro (fn x:id rhs:expr) (lambda (x) rhs))
+(define-syntax-parse-rule (fn x:id rhs:expr) (lambda (x) rhs))
 ((fn x x) 17)
 (fn 1 2)
 
-(define-simple-macro (fn2 x y rhs)
+(define-syntax-parse-rule (fn2 x y rhs)
   #:declare x id
   #:declare y id
   #:declare rhs expr
@@ -38,7 +38,7 @@ Defines a macro named @racket[macro-id]; equivalent to the following:
 (fn2 a #:b 'c)
 ]
 
-}
+@history[#:added "6.6.0.2"]}
 
 @defform[(define-syntax-parser macro-id parse-option ... clause ...+)]{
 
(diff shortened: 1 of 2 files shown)


stamourv:
I'm not sold on the name. `pattern` describes only half of what `define-simple-macro` does (`template` being the other half).
Some more ideas: `define-syntax-rewrite`, `define-syntax-clause`.

jackfirth:
Would `define-syntax-parse-rule` make sense?

mbutterick:
> I'm not sold on the name.

Nor I. A syntax pattern is a distinct thing in Racket (and it's already nettlesome that `syntax/parse`  supports syntax patterns that `syntax-case` does not, so to introduce a third meaning of the phrase ...)

> `define-syntax-parse-rule`

= best, by analogy to `define-syntax-rule`

lexi-lambda:
I’d be fine with `define-syntax-parse-rule`. It’s a bit wordy, but it’s probably the only name I’ve seen so far without any significant problems. Any opposed?

stamourv:
While we're at it, how about we make this new form implicitly quasisyntax its template? (as opposed to syntax, which is what `define-simple-macro` does).
I just wrote a bunch of code that used `define-simple-macro` and explicitly quasisyntaxed the template, which it turns out does not work. Now I have to use the full-fledged `syntax-parse` instead, which is a bummer. Quasisyntaxing would have solved my problem.

lexi-lambda:
Why not take it a step further? What about using `template` from `syntax/parse/experimental/template` instead of `syntax` or `quasisyntax`? And while we’re at it, what about getting that out of its `experimental` directory?

jeapostrophe:
+1

On Wednesday, July 6, 2016, Alexis King <notifications@github.com> wrote:

> Why not take it a step further? What about using template from
> syntax/parse/experimental/template instead of syntax or quasisyntax? And
> while we’re at it, what about getting that out of its experimental
> directory?
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/racket/racket/pull/1310#issuecomment-230921279>, or mute
> the thread
> <https://github.com/notifications/unsubscribe/AAOM-u2rSemxCSXHYgXpFejW-4IKNA9Xks5qTCZ1gaJpZM4IIz39>
> .
>


-- 
Jay McCarthy
Associate Professor
PLT @ CS @ UMass Lowell
http://jeapostrophe.github.io

                          - D&C 64:33

stamourv:
Sounds great!
@rmculpepper: Any blockers to moving templates out of experimental?

sorawee:
Should we revisit this PR? The "moving templates out of experimental" is already done since the initial release of Racket 7 IIUC. 

There's a conflict that must be resolved though.

jackfirth:
I think we should revisit this, yeah. Every time I write `define-simple-macro` my eye twitches a little.

@sorawee It might be easier to just make a new PR than to try and resolve this one's conflicts. It's not exactly a complex change.

PR #1311, opened 2016-04-19
Replace // ... with /* ... */

gus-massa:
Replace // ... with /* ... */ to support old C compilers.

I think it should be merged in 6.5.

### DIFF
--- a/racket/src/racket/src/optimize.c
+++ b/racket/src/racket/src/optimize.c
@@ -2563,9 +2563,10 @@ static Scheme_Object *rator_implies_predicate(Scheme_Object *rator, int argc)
       return scheme_list_pair_p_proc;
     else if (SAME_OBJ(rator, scheme_mcons_proc))
       return scheme_mpair_p_proc;
-    // XXX This could be implemented
-    // else if (SAME_OBJ(rator, scheme_make_byte_string_p))
-    //  return scheme_byte_string_p_proc;
+    /* XXX This could be implemented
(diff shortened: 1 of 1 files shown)


racket/src/racket/src/optimize.c
```diff
@@ -2563,9 +2563,10 @@ static Scheme_Object *rator_implies_predicate(Scheme_Object *rator, int argc)
       return scheme_list_pair_p_proc;
     else if (SAME_OBJ(rator, scheme_mcons_proc))
       return scheme_mpair_p_proc;
-    // XXX This could be implemented
-    // else if (SAME_OBJ(rator, scheme_make_byte_string_p))
-    //  return scheme_byte_string_p_proc;
+    /* This could be implemented
```
stamourv: Nitpick: Your changes dropped the `XXX`. Is that a keyword that someone may grep for?

rfindler: I think Jay uses that?​

gus-massa:
I restored the `XXX`.

mflatt:
(I don't know that "XXX" has any meaning here, but I'm interested to hear if someone specifically wants to keep it.)

I thought that `//` comments would create trouble for the MSVC build, but they didn't. Maybe it was a problem back with MSVC 2005 (instead of 2008). So, even though I still prefer using only `/* ... */`, maybe enough time has passed that we should start allowing `//`.

gus-massa:
OK, let's close this.

PR #1315, opened 2016-04-25
Add `procedure-impersonator*?`.

stamourv:
@mflatt, @rfindler: In case you want to review.


mflatt:
I think the docs should be a little clearer: `chaperone-procedure*?` returns #t for any impersonator of a value produced by `impersonate-procedure*` or `chaperone-procedure*` (or something like that).

Otherwise, this looks good to me.

stamourv:
Merged, with improved docs.

PR #1316, opened 2016-04-26
Implement mpairs outside of the core.

samth:
- [x] Implement mpair library
- [x] Support unsafe ops
- [x] Remove runtime mpair procedures
- [x] Figure out why the `peval` benchmark produces different answers
- [ ] Measure performance
- [ ] Correct printing


racket/collects/racket/private/mpair.rkt
```diff
@@ -0,0 +1,41 @@
...
+                    (current-inspector)
+                    #f
+                    null
+                    #f
+                    'mcons))
+
+(define mcar (procedure-rename (make-struct-field-accessor accessor 0) 'mcar))
+(define mcdr (procedure-rename (make-struct-field-accessor accessor 1) 'mcdr))
+
+(define set-mcar! (procedure-rename (make-struct-field-mutator setter 0) 'set-mcAr!))
```
jeapostrophe: typo on this line

rfindler: Last time I tried, `procedure-rename` was slower than writing a let and a wrapper lambda.

mflatt: Struct accessors are treated specially, so there's no extra wrapper at run time.

samth:
Unfortunately, this is significantly slower.

On dynamic:
```
[samth@huor:~/.../racket/benchmarks/common (mpair) plt] r dynamic.rkt 
cpu time: 633 real time: 632 gc time: 181
[samth@huor:.../racket/benchmarks/common racket-6.5.0.5] r dynamic.rkt 
cpu time: 460 real time: 459 gc time: 207
```

On peval:
```
[samth@huor:~/.../racket/benchmarks/common (mpair) plt] r peval.rkt 
cpu time: 2760 real time: 2758 gc time: 40
[samth@huor:.../racket/benchmarks/common racket-6.5.0.5] r peval.rkt 
cpu time: 1181 real time: 1181 gc time: 45
```

This is therefore on hold for the time being.

samth:
@mflatt improved this with e887fa56d1f4c, but there now appears to be a performance issue with GC. 

This program shows the problem:

```racket
#lang racket
(define v (make-vector 1000000))

(struct mp (a d) #:transparent #:mutable)
'cons
(collect-garbage)
(collect-garbage)
(collect-garbage)
(time
 (for ([i (in-range (vector-length v))])
   (vector-set! v i (cons 1 2))))

(collect-garbage)
(collect-garbage)
(collect-garbage)
'struct
(time
 (for ([i (in-range (vector-length v))])
   (vector-set! v i (mp 1 2))))  
```

On my machine, I get about 9 ms for the first look, and about 45 for the second, with the difference almost all GC time.

samth:
However, despite that issue, there's still something else going wrong for peval, which spends only 3% of its time in GC.

PR #1318, opened 2016-04-30
Improve with-disappeared-uses to support implicit begin and add record-disappeared-use

lexi-lambda:
This makes `with-disappeared-uses` surround its arguments with `let`, so it can contain multiple body expressions or internal definitions. It also adds a `record-disappeared-use` function, which is like `record-disappeared-uses` but for a single identifier.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/syntax-util.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax-util.scrbl
@@ -141,13 +141,15 @@ the parameter has a non-false value. This is done automatically by
 forms like @racket[with-disappeared-uses].
 }
 
-@defform[(with-disappeared-uses stx-expr)
+@defform[(with-disappeared-uses body-expr ... stx-expr)
          #:contracts ([stx-expr syntax?])]{
 
-Evaluates the @racket[stx-expr], catching identifiers looked up using
-@racket[syntax-local-value/record]. Adds the caught identifiers to the
-@racket['disappeared-uses] syntax property of the resulting syntax
-object.
+Evaluates the @racket[body-expr]s and @racket[stx-expr], catching identifiers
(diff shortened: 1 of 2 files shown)


pkgs/racket-doc/scribblings/reference/syntax-util.scrbl
```diff
@@ -141,13 +141,15 @@ the parameter has a non-false value. This is done automatically by
 forms like @racket[with-disappeared-uses].
 }
 
-@defform[(with-disappeared-uses stx-expr)
+@defform[(with-disappeared-uses body-expr ... stx-expr)
```
stamourv: This makes it look like there's an asymmetry between `body-expr`s and `stx-expr`, but the prose seem to consider them indistinguishable. Any reason not to have one name and `...+`? Same with the `#:contracts` below.

lexi-lambda: There *is* a difference, though: the `body-expr`s can evaluate to anything, but the `stx-expr` must evaluate to a syntax object. I can clarify the prose a little bit more to make that clear, though (the last bit should probably read “Adds the caught identifiers to the `'disappeared-uses` syntax property of the syntax object produced by `stx-expr`.”).

rmculpepper:
I'd prefer changing `record-disappeared-uses` to take either a list of identifiers or a single identifier rather than adding a new procedure.

lexi-lambda:
@stamourv Updated the docs with the tweak I mentioned.

@rmculpepper Okay, I removed `record-disappeared-use` and implemented the change you suggested. I was mostly just following the pattern set by `generate-temporaries` / `generate-temporary`, but I guess this is different since `record-disappeared-uses` already comes from `racket/syntax`.

stamourv:
Merged, thanks!

PR #1319, opened 2016-05-01
Add examples for _enum and _bitmask usage. (Close #1317)

### LINKED ISSUE #1317: FFI | More examples for _bitmask and _enum
It was not immediately obvious to me on how to actually call a ffi procedure with a bitmask or enum parameter; it was only when I found this [stackoverflow question](http://stackoverflow.com/questions/1890853/using-bitmask-from-pltscheme-ffi) that I understood how to use them.

Maybe adding an example of a procedure call with an enum or bitmask in the documentation would help?

I would also gladly provide a patch for this if you deem it useful.

stonecauldron:
In reference to issue #1317


stamourv:
Merged, thanks!

PR #1320, opened 2016-05-03
expand docs for `syntax-local-introduce`

mbutterick:
Add more detailed explanation, and code sample showing characteristic use.


mbutterick:
found a typo ... will resubmit

PR #1321, opened 2016-05-03
expand docs for `syntax-local-introduce`

mbutterick:
Add more detailed explanation, and code sample showing characteristic use.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
@@ -923,11 +923,37 @@ definition context.
 
 @defproc[(syntax-local-introduce [stx syntax?]) syntax?]{
 
-Produces a syntax object that is like @racket[stx], except that the
-presence of @tech{scopes} for the current expansion---both the
-macro-introduction scope and the use-site scope, if any---is flipped
-on all parts of the syntax object. See @secref["transformer-model"] for information
-on macro-introduction and use-site @tech{scopes}.
+Produces a syntax object that is like @racket[stx], but with a new @tech{scope} added that flips the apparent introduction site for all parts of the syntax object. Meaning, parts of the syntax object that start with a macro-introduction scope will end up with a use-site scope, and vice versa. (See @secref["transformer-model"] for information on macro-introduction and use-site @tech{scopes}.)
+
+Consistent with its name, this procedure is commonly used to introduce an unhygienic binding at the use site, by making a macro-introduced binding appear as if it came from the use site. It is important, however, that the syntax object passed to @racket[syntax-local-introduce] be free of any extraneous scopes. Otherwise, the behavior of the introduced identifier can be unpredictable. Using @racket[(datum->syntax #f _id)] to prepare the input syntax is good policy.
+
+@examples[#:eval stx-eval
+
+(module sli racket
+  (provide sli-x)
+  (define-syntax (sli-x stx)
+    (syntax-case stx ()
+      [(_) (with-syntax ([x (syntax-local-introduce (datum->syntax #f 'x))])
+             #'(define x 'syntax-local-introduce-x))])))
+
+(require 'sli)
+(let ()
+  (sli-x)
+  x)
+
+(module sli-broken racket
+  (provide sli-x-broken)
+  (define-syntax (sli-x-broken stx)
+    (syntax-case stx ()
+      [(_) (with-syntax ([x (syntax-local-introduce #'x)])
+             #'(define x 'syntax-local-introduce-x))])))
+
+(require 'sli-broken)
+
+(eval:error (let ()
+              (sli-x-broken)
+              x))
+]
 
 @transform-time[]}
 


pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -923,11 +923,48 @@ definition context.
 
 @defproc[(syntax-local-introduce [stx syntax?]) syntax?]{
 
-Produces a syntax object that is like @racket[stx], except that the
-presence of @tech{scopes} for the current expansion---both the
-macro-introduction scope and the use-site scope, if any---is flipped
-on all parts of the syntax object. See @secref["transformer-model"] for information
-on macro-introduction and use-site @tech{scopes}.
+Produces a syntax object that is like @racket[stx], but with a new 
+@tech{scope} added that flips the apparent introduction site for all
```
samth: I think it's inaccurate to say that it's a "new scope". What's flipped are two existing scopes.

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -923,11 +923,48 @@ definition context.
...
+and vice versa. (See @secref["transformer-model"] for information on 
+macro-introduction and use-site @tech{scopes}.)
+
+Consistent with its name, this procedure is commonly used to introduce 
+an unhygienic binding at the use site, by making a macro-introduced 
+binding appear as if it came from the use site. It is important, however, 
+that the syntax object passed to @racket[syntax-local-introduce] be free 
+of any extraneous scopes. Otherwise, the behavior of the introduced 
+identifier can be unpredictable. Using @racket[(datum->syntax #f _id)] 
+to prepare the input syntax is good policy.
```
samth: I think we should be more specific about what happens than "unpredictable".

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
mbutterick: I was adapting your comment that "`(syntax-local-introduce stx)` adds a single scope, using the equivalent of `'flip` mode."

samth: Right, but that scope isn't "new" -- it's `the
 -macro-introduction scope and the use-site scope, if any`, as the old docs say.

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
mbutterick: The whole premise of this pull request is that I don't understand what this function does. We can therefore agree that I'm unqualified to document it ;)

mbutterick:
Updated to include @mflatt’s simple explanation & example from [“Binding as Sets of Scopes”](http://www.cs.utah.edu/plt/scope-sets/implementation.html).

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -923,11 +923,92 @@ definition context.
...
+ 
+(invoke stashed-foo)
+]
+
+Along these lines, @racket[syntax-local-introduce] can also be used
+to introduce an unhygienic binding at the use site,
+by making a macro-introduced binding appear as if it originated at the
+use site. It is important, however, that the syntax object passed
+to @racket[syntax-local-introduce] be free of any extraneous scopes.
+Using @racket[(datum->syntax #f _id)]
```
samth: How about "Here, we use `(datum->syntax #f id)` to produce an identifier with no scopes."?

samth:
This looks very nice!

pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -923,11 +923,101 @@ definition context.
...
+For instance, the @racket[sli-x-broken] macro below won't work because we're 
+using @racket[#'x] as an identifier. The @racket[#'] prefix creates a syntax
+object that has the scopes of the current context (in this case, inside the 
+macro). Even though this identifier's macro-introduction scope is canceled by 
+@racket[syntax-local-introduce], these other scopes prevent it 
+from binding the @racket[x] at the use site.
+
+But in the @racket[sli-x] macro, we use @racket[(datum->syntax #f _id)] to 
+produce an identifier with no scopes, which will make the unhygienic 
+binding work.
```
AlexKnauth: Should you clarify that it needs to have no scopes because then `syntax-local-introduce` will start with a clean slate and add exactly the scopes needed to make the unhygienic binding work?

mbutterick: OK.

AlexKnauth:
This build is only failing because of what looks like random planet failure on one of the seven Travis builds. Should this build be restarted?

samth:
I restarted that build.

samth:
Now passed. @mflatt, any thoughts on these additions?

mbutterick:
As a way of creating unhygienic bindings, would it be worthwhile to contrast `syntax-local-introduce` against the more traditional `(datum->syntax calling-site-stx 'id-name)`? 

After further experiments, it seems to me that the identifier created by `(syntax-local-introduce (datum->syntax #f 'id-name)` is in a sense “more” unhygienic because it sits higher in the scope hierarchy (and thereby captures more uses)

AlexKnauth:
From what I understood from discussions like [this racket users list message](http://lists.racket-lang.org/users/archive/2013-October/059737.html) is that the `calling-site-stx` usually isn't what you want, particularly because another `outer` macro could expand to a use of your `inner` macro, and then the `calling-site-stx` would point to the parens in the definition of the other `outer` macro. 

So, in examples like [this](http://lists.racket-lang.org/users/archive/2013-October/059736.html), `syntax-local-introduce` can make it work even in cases where other macros expand to uses of `inner` or `outer`.

mbutterick:
Fair enough — probably past time for me to give up on this mystery. FWIW here’s an example of how a `syntax-local-introduce` identifier seems to be able capture an identifier in another macro in a way that a `datum->syntax` style identifier cannot. But notice that this unhygienic identifier doesn’t work inside `rackunit`. I have to conclude this is some kind of accidental scope leakage, not a by-design attribute of `syntax-local-introduce` — if `(invoke-use-site-id-macro)` fails in one circumstance, it ought to fail in the other.

```racket
#lang racket

(module macros racket
  (provide (all-defined-out))
  
  (define-syntax (make-use-site-id-macro stx)
    (syntax-case stx ()
      [(_)
       (with-syntax ([use-site-id (datum->syntax stx 'use-site-id)])
         #'(define (use-site-id) 'use-site-id-success))]))
  
  (define-syntax-rule (invoke-use-site-id-macro)
    (use-site-id)) ; a default macro-introduced identifier
  
  (define-syntax (make-sli-id-macro stx)
    (syntax-case stx ()
      [(_)
       (with-syntax ([sli-id (syntax-local-introduce (datum->syntax #f 'sli-id))])
         #'(define (sli-id) 'sli-id-success))]))
  
  (define-syntax-rule (invoke-sli-id-macro)
    (sli-id)))  ; a default macro-introduced identifier

(require 'macros rackunit)
(make-use-site-id-macro)
(check-equal? (use-site-id) 'use-site-id-success)
(check-exn exn:fail? (λ () (eval #'(invoke-use-site-id-macro))))

(make-sli-id-macro)
(check-equal? (sli-id) 'sli-id-success)
;; works here, despite relying on a macro-introduced identifier, but ...
(invoke-sli-id-macro) 
;; ... not inside rackunit
(check-exn exn:fail? (λ () (eval #'(check-equal? (invoke-sli-id-macro) 'sli-id-success)))) 
```

PR #1322, opened 2016-05-03
Adding remove-impersonators and chaperone-opaque-struct functions

thinkmoore:
This pull request adds two new primitive functions to Racket in the module ```racket/unsafe/chaperones```. These functions are necessary to efficiently implement bounded polymorphic contracts: [github.com/thinkmoore/bounded](github.com/thinkmoore/bounded). Documentation and examples of bounded polymorphic contracts are available at [thinkmoore.net/bounded](thinkmoore.net/bounded).

```remove-impersonators``` allows impersonators to be removed from the chain of impersonators wrapping a value as long as the caller has access to a value underneath the impersonators being removed.

```chaperone-opaque-struct``` can be used to create a chaperone of any struct, even an opaque struct for which the caller does not have a struct-type object. However, it does not allow any redirects to be specified, and so can only be used to tag values with impersonator properties.

I put these primitives in the unsafe collection because they slightly change the semantics of chaperones. That said, I think they are "safe" in that they should not interfere with any abstractions that were previously available to clients of chaperoned values. If it makes sense, they could be moved elsewhere.

I bumped the version number to reflect the new primitive operations, and also added a new procedure ```scheme_unsafe_init_struct``` to install unsafe functions related to structs.

### CODE REVIEW

**Comment by samth** [comment]
`chaperone-opaque-struct` seems like it would be better as a modification to `chaperone-struct`, or as a different version with the same API but fewer restrictions.

Do you need impersonator versions of these?

**Comment by thinkmoore** [comment]
Yes, I think a nice alternative to ```chaperone-opaque-struct``` would be to relax the restrictions in ```chaperone-struct``` to allow you to create chaperones/impersonators of any struct as long as you only add properties. As far as chaperones vs impersonators, I thought a chaperone was always usable where an impersonator was expected. But IIRC that isn't actually the case?

**Comment by thinkmoore** [comment]
Looking into this further, it looks like the only change that would be necessary in ```chaperone-struct``` is to remove the explicit witness check. It looks like that check (and thus the optional ```struct-type``` argument) is only included to explicitly rule out the functionality I want to introduce with ```chaperone-opaque-struct```. As far as I can tell, removing this check has no other effect, since supplying redirects also require additional arguments that serve as a witness

Should chaperones and impersonators require a witness for adding properties but not otherwise interposing on structs?

**Comment by rfindler** [comment]
Should (chaperone-struct 3 ...) work?

**Comment by thinkmoore** [comment]
Not sure what you're proposing Robby, but I would say no since 3 is not a struct. 

I'm proposing that ```(chaperone-struct someStruct someProp #t)``` return a chaperone for ```someStruct``` that does not change its behavior in any way but adds the specified impersonator property, as long as ```someStruct``` is a struct (of any structure type).

**Comment by rfindler** [comment]
Are you referring to the `struct?` predicate? Are you saying that chaperone-struct should work only when that returns #t?

This program prints `#f`

```
#lang racket
(struct s (x))
(struct? (s 1))
```

Stepping back, what I'm saying is that ALL values in Racket are structs. The only question is if the accessors (and whatnot) are available to you. You don't have access to the struct selectors for 3 (and indeed, they are specially optimized representations in the runtime for that particular value, but for this one: `73786976294838206464` you will find something very close to what you think of as struct if you could poke around in the guts of the pointers and whatnot in the runtime system).

So if you want this to work for arbitrary values, then there is the monumental task of adapting all of the primitive operations in the runtime system so they accept values with chaperone properties on them.

thinkmoore:
Thanks for the clarification Robby. It sounds like even the ```chaperone-opaque-struct``` primitive I suggest is in fact unsafe, even for user-created structure types (which we could presumably treat differently from built-in structures): those user-created structure types may have operations implemented in native code that do not support chaperones.

I'm not sure of the best way forward here. The functionality I actually need for the contracts is to add an additional property to structs that already have a chaperone or impersonator. In that case, it's clearly safe to add a chaperone or impersonator since the value already has one.

Does it make sense to weaken the witness requirement in the case where the value being impersonated is already an impersonator?

rfindler:
As a matter of principle, there are no non-struct values in Racket and we cannot make that distinction. I think that probably it would be doable, however.

I think that would make sense, but it would be weird. 

Can you say more about the structs that you actually to be chaperoned? Presumably you never need to chaperone the number 3. Is there something more specific, that can be phrased positive to characterize the structs that you do need to add properties to?

thinkmoore:
In the implementation of bounded polymorphic contracts, the user may specify a "bound" contract that is a contract for a particular type of opaque struct (pseudo-code here):
```(bounded-> ((forall X <: (object/c ...)) X X)```

Under the hood, when the a value passes into the function through the first ```X```, I apply the projection of the object contract to it. In addition, I need to record what the original value and the contracted value in a property. Then, when the value is returned from the function through the second ```X```, I can retrieve this information from the property and use it to remove this contract, leaving any additional chaperones/impersonators that were added during the body of the function.

In this scenario, it is safe to add a chaperone or impersonator since we must have already added one when attaching the contract bound. However, we don't want to break abstractions by making it possible for the implementation of these contracts to get access to the struct-type info for objects, for example.

Thinking about it some more, I could probably implement this with clever use of ```make-weak-hasheq``` at the cost (probably small) of performance and memory. Thoughts?

Any thoughts about ```remove-impersonators```? Because it removes all impersonators between the one specified and the witness provided, I think it avoids the possible issues we discussed earlier related to patterns that use multiple layers of impersonators.

rfindler:
I think that you really want the ability to do (chaperone-struct 3 ...), after all, right? Otherwise you couldn't specify something like (between/c 0 10) as a bound?

Hm. Maybe we should add a `can-be-chaperoned?` predicate (with a better name, of course) and just say something like "this procedure returns #f only when the runtime system is, sadly, limited". As we move more and more stuff up into Racket maybe we can even decrease this set. 

I'm not really sure we would want to commit to a strategy like this, where we are essentially promising a feature that we probably won't ever be able to actually implement tho. :)

thinkmoore:
The mechanism described above is what I do for bounds that specify higher-order contracts. For flat contracts, the wrapping and unwrapping steps are not needed, you just do the first-order-check on both the way in and out. (I haven't actually implemented that yet, currently, I just reject non-higher-order contracts as bounds.)

rfindler:
Oh, of course. (I think you must have explained this to me once before. It seems quite elegant.)

I have some thoughts, but I think it's time to move to email.

samth:
If the further thoughts on this are about the changes to Racket proposed by @thinkmoore, can we keep them in this discussion so that they're archived/accessible to others of us who are interested?

mflatt:
@thinkmoore Would it help to allow an impersonator-property predicate as a witness?

(As you suggest, if you can demonstrate that an impersonator property was added before, then that should be enough to know that another can be added.)

rfindler:
@samth they are not directly about this. Sorry for the tantalizing note. I'll delete my comment. And then delete this one.

PR #1323, opened 2016-05-04
make match exception transparent

jbclements:
Per discussion on racket-users mailing list, making exn:misc:match transparent to better match other exceptions raised by racket.


samth:
Seems good to me, but maybe have more of the information about this change in the commit message (such as the reasoning that's in the PR, and a link to the discussion).

jbclements:
okay, added info in commit message. I'll wait for the automatic tests to run, then merge this.

PR #1326, opened 2016-05-19
Make relative directory argument for resolve-module-path-* have a def…

LeifAndersen:
…ault to #f


mflatt:
This seems fine to me.

Issue #1325, opened 2016-05-14
Potentially incorrect indentation


Earnestly:
In https://github.com/racket/racket/blob/master/racket/src/racket/src/read.c#L3136 the `if` statement seems to suggest that `scheme_read_err` is part of the `else` block:

```c
  else
    scheme_tell_all(port, &line2, &col2, &pos2);
    scheme_read_err(port, stxsrc, line, col, pos, pos2-pos, -1, indentation, "read: expected fixnum, got %V", n);
  return NULL;
```

The indention is misleading if this isn't the case.

gus-massa:
Looking at the previous function [read.c#L3117](https://github.com/racket/racket/blob/master/racket/src/racket/src/read.c#L3117), I think that the curly brackets are missing. 

The code is equivalent, because the other branch has a `return`, but it's confusing. 

Perhaps you can send this as a PR. (If Matthew doen't fix this inmeditely.)

gus-massa:
Fixed in https://github.com/racket/racket/commit/9978696991d04f8bc8aed34aa02fa39cd7b9b3f3

Thanks for the report.

PR #1328, opened 2016-05-21
Fix confusing indentation

gus-massa:
Add a `{...}` because the indentation is confusing. The code is equivalent, because the other branch has a `return`,

Reported by @Earnestly in https://github.com/racket/racket/issues/1325


stamourv:
Confusing indeed.
Dropping the else clause altogether and moving its contents in the function body would be clearer, IMO, and more in line with the style used elsewhere in the runtime.

gus-massa:
I agree, but the code of the previous function [read.c#L3117](https://github.com/racket/racket/blob/master/racket/src/racket/src/read.c#L3117) is very similar and has this format.

Perhaps I can change both.

stamourv:
Sounds like a good plan.

gus-massa:
New version, without the `else` branchs.

mflatt:
LGTM

PR #1330, opened 2016-05-23
add example of negative phase level

mbutterick:


mbutterick:
Extend [this doc page](http://docs.racket-lang.org/guide/stx-phases.html?q=phase) about compile- and run-time phases to include an example of the negative phase levels that are alluded to (but not demonstrated) at the end.

PR #1334, opened 2016-05-28
Fix description of the `#x20` X-expression

cky:
An X-expression of `#x20` converts to `&#x20;` (actually `&#32;`), not `&#20;`. An X-expression of `20` would indeed convert to `&#20;`, however.


jeapostrophe:
Thank you!

PR #1340, opened 2016-06-12
improve vertical positioning & paren shaping in `syntax->string`

mbutterick:
Two improvements:

1) Currently, `syntax->string` collapses multiple newlines into one. But the docs say that it “Builds a string with newlines and indenting according to the source locations”. So it seems more accurate to preserve the newlines.

2) Currently, `syntax->string` ignores the `'paren-shape` property. (Perhaps it predates 'paren-shape.) It seems more accurate to observe it.

### DIFF
--- a/racket/collects/syntax/to-string.rkt
+++ b/racket/collects/syntax/to-string.rkt
@@ -3,7 +3,7 @@
            syntax/stx)
   
   (require racket/list)
-
+  
   (define (syntax->string c)
     (let* ([s (open-output-string)]
            [l (syntax->list c)]
@@ -14,7 +14,7 @@
         (let ([c (syntax-column c)]
               [l (syntax-line c)])
           (when (and l (l . > . line))
-            (newline)
+            (for-each (λ (_) (newline)) (range (- l line)))
             (set! line l)
             (init-line!))
           (when c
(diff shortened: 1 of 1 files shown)


racket/collects/syntax/to-string.rkt
```diff
@@ -55,10 +55,14 @@
                  ((loop init-line!) i))]
               [(pair? (syntax-e c))
                (advance c init-line!)
-               (printf "(")
+               (define c-paren-shape (syntax-property c 'paren-shape))
+               (printf (format "~a" (or c-paren-shape #\()))
```
AlexKnauth: `printf` already handles format strings, so this could be `(printf "~a" (or c-paren-shape #\( ))`.

mflatt:
I didn't know that this function exists.

@stamourv, "optimization-coach" has the only use of syntax->string` that I find. Ok with you?

stamourv:
Yes.

AlexKnauth:
What happened to this?

stamourv:
Merged, thanks!

AlexKnauth:
As https://github.com/racket/racket/commit/9422d6660103494e6c79f9f3e1c8944de66c4901

PR #1346, opened 2016-06-20
Add #:username and #:password arguments to net/git-checkout

lexi-lambda:
This is an initial implementation of allowing authentication through the `net/git-checkout` interface, which sets up allowing the package system to clone packages with private git sources, as discussed on the mailing list.


lexi-lambda:
Subsumed by #1472.

PR #1347, opened 2016-06-21
add lang-reader-module-paths to factor out copy-pasted code

AlexKnauth:
This pull request adds `lang-reader-module-paths` to `syntax/module-reader` to go with `make-meta-reader`. A lot of uses of `make-meta-reader` use the exact same code for the third argument. This factors it out so that it can be re-used instead of copy-pasted.

This exact code (or pretty close to it) has been copy-pasted in at-exp, errortrace, and 2d, as well as in packages like afl, colon-kw, curly-fn, debug, exact-decimal-lang, hygienic-quote-lang, hygienic-reader-extension, lispi, no-vert-bar-lang, pollen, postfix-dot-notation, quote-bad, repltest, semilit (although this uses `bytes->string/utf-8` instead of `bytes->string/latin-1`), and readtable.

One small question, should this use `bytes->string/latin-1` like at-exp uses, or `bytes->string/utf-8` like semilit uses?


AlexKnauth:
Hello?

AlexKnauth:
Should this be added for version 6.6?

stamourv:
That looks like a pretty conservative change to me, so I'm fine with it going into 6.6.
@mflatt? Others?

AlexKnauth:
Thanks! Should I now make pull requests to use this in `errortrace` and `2d`?

rfindler:
that would be great, thanks!

PR #1349, opened 2016-06-21
Track disappeared uses of match expanders within match forms

lexi-lambda:
Match expanders used when expanding `match` forms should be added to the resulting syntax object’s `'disappeared-uses` syntax property so they are picked up by DrRacket’s import analysis and binding arrows. This implements that functionality using `with-disappeared-uses` and `syntax-local-value/record`.


samth:
Awesome. I'll merge when the CI passes.

PR #1353, opened 2016-06-22
Skip superfluous whitespace in PLTSTDERR environment variable contents

tonyg:
Of the following commands,

    PLTSTDERR=" " racket
    PLTSTDERR="  " racket
    PLTSTDERR="debug" racket
    PLTSTDERR=" debug" racket
    PLTSTDERR="  debug" racket
    PLTSTDERR="debug " racket
    PLTSTDERR="debug  " racket
    PLTSTDERR="debug@foo info@bar" racket
    PLTSTDERR="debug@foo  info@bar" racket
    PLTSTDERR=" debug@foo  info@bar" racket

only numbers 3, 6 and 8 run without a printed warning. All should be acceptable. This pull request skips the superfluous whitespace, allowing all to be accepted without warning.

### DIFF
--- a/racket/src/racket/cmdline.inc
+++ b/racket/src/racket/cmdline.inc
@@ -755,6 +755,10 @@ static Scheme_Object *get_log_level(char *prog, char *real_switch, const char *e
   l =  scheme_make_null();
 
   while (1) {
+    while (*str && *str == ' ') {
+      str++;
+    }
+
     if (!*str) {
       if (default_lvl == -1) default_lvl = 0;
       if (last)
--- a/racket/src/racket/cmdline.inc
+++ b/racket/src/racket/cmdline.inc
@@ -746,6 +746,8 @@ static Scheme_Object *reverse_path_list(Scheme_Object *l, int rel_to_cwd)
 # define GC_CAN_IGNORE /**/
 #endif
 
+#include <ctype.h>
+
 static Scheme_Object *get_log_level(char *prog, char *real_switch, const char *envvar, const char *what, GC_CAN_IGNORE char *str)
 {
   int k, len, default_lvl = -1;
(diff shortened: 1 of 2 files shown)


racket/src/racket/cmdline.inc
```diff
@@ -755,6 +755,10 @@ static Scheme_Object *get_log_level(char *prog, char *real_switch, const char *e
   l =  scheme_make_null();
 
   while (1) {
+    while (*str && *str == ' ') {
```
samth: Should this skip tab characters as well?

tonyg: Probably, but `' '` is hardcoded elsewhere in the same routine and I don't want to get into heavily refactoring it. That can be a separate issue.

samth:
Should this be mentioned in the docs?

tonyg:
Perhaps; but again, I think improving the documentation can be a separate issue.

I'd like one more set of eyes on this (given it's horrible C string-parsing code) before I merge this. @mflatt, could you take a quick look?

samth:
Since we're changing the semantics, I think it should be mentioned in the docs.

tonyg:
OK, I've updated the PR:
 - docs describe the old and new behaviours
 - the code now uses `isspace` (C89, `ctype.h`) instead of directly comparing against `' '`
 - (minor doc typo fix bundled in too)

@samth, look ok to you? If so I will merge.

tonyg:
Incidentally, I discovered this because I was using the following in a script:

```sh
export PLTSTDERR="$PLTSTDERR debug@syndicate-broker"
exec racket server.rkt 2>&1
```

... which fails when `PLTSTDERR` is empty. This didn't seem right to me.

samth:
I would write just the current behavior, and use @history to note the change. Otherwise, looks good.

tonyg:
Closed by 1ba42bb70d.

PR #1354, opened 2016-06-22
Document this-syntax and #:auto-nested-attributes

lexi-lambda:
Both `this-syntax` and `#:auto-nested-attributes` are features of `syntax/parse` that appear to be undocumented. This is my best attempt to explain their behavior.


rmculpepper:
I added docs, but I rewrote them to be more precise about what `this-syntax` can produce and to emphasize that `#:auto-nested-attributes` is deprecated.

lexi-lambda:
Wonderful; thank you. :)

PR #1355, opened 2016-06-25
Add H: search box option, to search for #lang providers

bennn:
Adds two new search options:
- `H: str` shows only matches from a module `str` documented with `defmodulelang` 
- `R: str` same for `defmodulereader`

Searching for `H:` or `R:` alone lists all `#lang` and `#reader` modules.

Here's a screenshot:
screen shot 2016-06-24 at 23 49 47

### DIFF
--- a/pkgs/racket-index/scribblings/main/private/make-search.rkt
+++ b/pkgs/racket-index/scribblings/main/private/make-search.rkt
@@ -168,7 +168,14 @@
           [(exported-index-desc? desc)
            (let ([libs (map lib->name (exported-index-desc-from-libs desc))])
              (string-append* `("[" ,@(add-between libs ",") "]")))]
-          [(module-path-index-desc? desc) "\"module\""]
+          [(module-path-index-desc? desc)
+            (cond
+             [(language-index-desc? desc)
+              "\"language\""]
+             [(reader-index-desc? desc)
+              "\"reader\""]
+             [else
+              "\"module\""])]
           [else "false"]))
       (and href
            (string-append "[" (quote-string text) ","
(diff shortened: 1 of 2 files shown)


bennn:
@mflatt ping, can you look at this and https://github.com/racket/scribble/pull/45 before the v6.6 release?

The build is failing because https://github.com/racket/scribble/pull/45 isn't merged. The command `racket -l tests/pkg/test -- -q` exits cleanly on my fork.

bennn:
Added search option `R:` for `#reader` modules. Cancelled the travis build because CI will fail until https://github.com/racket/scribble/pull/48 is merged.

bennn:
ping; anyone who wants to try the search options, here's a copy of the docs that supports `H:` and `R:`
http://www.ccs.neu.edu/home/types/racket-doc/index.html

stamourv:
I see the interest in being able to search for languages.
If you search for, say `H:eopl`, you get the full list of exports from eopl, whereas if you search fo `H:eop`, you only get the language. Is that intentional?
The `H:` prefix is unfortunate, but `L:` is already taken...

bennn:
Yes that's intentional. It's the same fall-through behavior as `L:` ... if there's an exact match, show all results from the module, if not just show module names that partially match.

The reason I think it should be like `L:` and not like `M:` is so `H:` by itself doesn't just list "all identifiers whose language partially matches the empty string". Instead you get a list of language names.

stamourv:
Right, the empty string behavior makes sense, and is nice. It's the "string that matches a language" behavior that I was finding odd.
In any case, this looks like a useful feature, and the implementation looks reasonable.

bennn:
1dd934c8cb78ba038235eab9659bc9c4c08c07df

PR #1360, opened 2016-07-01
Fixed location of traced procedures. Closes #1300

### LINKED ISSUE #1300: functions defined with trace-define (and trace-lambda) don't retain original source location
Maybe current behavior is in fact the desired behavior, but it surprised me. I love how Racket lets me know where a function was defined, and it's name if available. Often that's the sort of information I'd want to know when tracing. But `trace`d (`trace-define`d, `trace-lambda`ed) procedures instead return the new procedure created within the `trace` macro. 

```
 > (require racket/trace)
 ﻿> (define (const4 x) 4)
 ﻿> (trace-define (const5 x) 5)
 ﻿> const4
 #<procedure:const4>
﻿ > const5
 #<procedure:...racket/trace.rkt:286:44>
﻿ > (const5 const4)
 >(const5 #<procedure:const4>)
 <5
 5
 > (const5 const5)
 >(const5 #<procedure:...racket/trace.rkt:286:44>)
 <5
 5
```

wilbowma:


stamourv:
Could you add some tests that would fail with the current implementation?

wilbowma:
Can do

wilbowma:
Good thing you made me do that. I missed a case.

stamourv:
Merged, thanks!

Issue #1115, opened 2015-10-30
`raco pkg install --skip-installed P` errors if P is installed in a wider scope


samth:
If you have package P installed in installation scope, and then you do:

```
$ raco pkg install --skip-installed P
```

This will error, even though P is of course installed.

samth:
This was discovered by @sabauma

vimniky:
I have the same problem . thanks @samth

mflatt:
This appears to have been fixed by the merged PR, although there's a related question in a newer issue.

PR #1361, opened 2016-07-01
skip-installed from all scopes, not current

wilbowma:
closes issue #1115


samth:
This looks nice and simple to me. @mflatt or @jeapostrophe?

wilbowma:
Investigating the failure

wilbowma:
Okay, this should work. Also, added a test case for it.

wilbowma:
Looks like previous failure had to do with hashes, not something I changed; rebasing to master.

wilbowma:
All tests pass! Going to drop the whitespace changes, per concern about `git blame`.

rfindler:
(FWIW, I wish that whitespace wasn't there in the first place and I remove
it when I edit lines and thanks to Max, DrRacket doesn't include it anymore
.... sigh)

jeapostrophe:
LGTM

PR #1362, opened 2016-07-02
Optimizations for ignored if and vector->immutable-vector

gus-massa:
**optimizer: reduce ignored `if`s**
The optimizer tries to reduce the `if` assuming that the result will be used.
In case it later detects that the result will be ignored, it can try to apply some additional reductions to the branches and to the whole expression. 
For example

```
(void (if x (box 1) (random)))  =>  (void (if x #f (random)))
```

**optimizer: add types for vector->immutable-vector and vector->list**
Some straighforward code to add the internal types for these two functions, and more test for old functions.
For laziness, in `rator_implies_predicate` I recognized them by name instead of using a global `scheme_vector_to_whatever` variable, but I can change that if necessary.
## 

**optimizer: try to mark vector->immutable-vector as omitable (don't merge)**
Mark it as omitable when the optimizer can prove that the argument is a `vector?`
that is not a `chaperone`/`impersonator`.
I'm still **not** happy with this commit, and it's too complicated to be merged now. I prefer to wait until next week after the Racket 6.7 branch is created. Any comments are welcome.

This version tries to recognize the vectors that are not chaperones, because chaperones may cause an error in innocent looking expressions like `(vector->immutable-vector v)`. I added an additional parameter in `do_expr_implies_predicate` because the (untyped) type system is not strong enough to understand this, because it has no intersection types.

An alternative is to give this parameter a more restrictive meaning, so `do_expr_implies_predicate` recognizes only fresh vectors that are not referenced elsewhere like in https://github.com/racket/racket/pull/1237. Or perhaps something more complicated like
- `no_chap=0` recognize any vector (old behavior)
- `no_chap=1` recognize vectors without chaperones
- `no_chap=2` recognize fresh vectors without other references

But there are some corner cases when this is extended to other types like flonums. (Is the result of `(fl+ x y)` fresh when `x` is `0.0`?)

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -1518,6 +1518,20 @@
                       #f
                       v v v2 v2))))
 
+(test-comp '(lambda (x) (vector x x x) (void))
+           '(lambda (x) (void)))
+(test-comp '(lambda (x) (vector-immutable x x x) (void))
+           '(lambda (x) (void)))
+(test-comp '(lambda (x) (vector->immutable-vector (vector x x x)) (void))
+           '(lambda (x) (void)))
+; v may be a chaperone 
+(test-comp '(lambda (v) (when (vector? v)
+                          (vector->immutable-vector v)
+                          (display v)))
+           '(lambda (v) (when (vector? v)
+                          (display v)))
+           #f)
+
 (test-comp '(lambda (w z)
               (if (list w z (random 7))
                   (let ([l (list (random))])
--- a/racket/src/racket/src/optimize.c
+++ b/racket/src/racket/src/optimize.c
@@ -128,9 +128,11 @@ static void merge_types(Optimize_Info *src_info, Optimize_Info *info, Scheme_Has
 static Scheme_Object *lookup_constant_proc(Optimize_Info *info, Scheme_Object *rand);
 
 static Scheme_Object *expr_implies_predicate(Scheme_Object *expr, Optimize_Info *info);
+static Scheme_Object *expr_implies_predicate_no_chap(Scheme_Object *expr, Optimize_Info *info, int no_chap);
 static Scheme_Object *do_expr_implies_predicate(Scheme_Object *expr, Optimize_Info *info,
                                                 int *_involves_k_cross, int fuel,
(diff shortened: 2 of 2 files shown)


mflatt:
First two commits seem fine. I'm not sure the third is a worthwhile direction; is there a specific motivation for it?

gus-massa:
I merged the two first commits, and left the last one here.

The idea for the third commit is to minimize the overhead of the immutable package https://github.com/racket/racket/pull/1341 when it's merged as a part of Racket or used as an independent package. It will add a lot of implicit `vector->immutable-vectors` everywhere. 

It's not very optimization friendly, but I think the optimizer can be improved to fix most of the spots. (And `keyword-lambda` needs some improvements too.)

Anyway, I think that marking the `vector->immutable-vectors` as ommitable as in this commit is not very useful. I think that I can write some microbenchmarks that show a big difference in the run time, but I don't imagine too many realistic programs with a big difference in run time.

I think that https://github.com/racket/racket/pull/1237 is better, because it can decrease the allocations and reduce the run time in realistic programs. But the idea is too weird, so I still prefer to wait and think about it.

sorawee:
Is it still worth optimizing BC? If not, we can close the PR.

Note that the first optimization is already done in CS:

```
#lang racket

(define (test x)
  (void (if x (box 1) (random))))
```

is CP0'ed to:

```
  (letrec ([test.9 (lambda (x_1.10)
                     (if x_1.10
                         (#2%void)
                         (#3%pseudo-random-generator-next!
                           (current-pseudo-random-generator)))
                     (#2%void))])
     ...)
```

samth:
The first and second commits were merged a long time ago.

gus-massa:
I think the only interesting remaining part are the new tests https://github.com/racket/racket/pull/1362/commits/479ef204060bbca52a028279dd042c07045dd6db#diff-1e72f2f913feb37a602ba0782f413a27ccd4a974d7957b06984dc316cfe223b2R1521-R1534 But IIRC `vector-immutable` and other similar functions in CS are not primitives at the Chez Scheme level, so it's difficult to optimize them away. So I don't expect to "fix" them soon.

The test are easy to write again (and the closed PR are visible anyway) so I'm fine with closing this. (I had this as a remainder, but I can move it to my secret wish list.)

Issue #1342, opened 2016-06-13
`command-line` #:once-each doesn't error with rest args


florence:
How to reproduce:

Given the program "once-each.rkt":

```racket
#lang racket
(displayln 
 (command-line
  #:once-each
  [("-p") "thing" (displayln "-p")]
  #:args a a))
```

Run the program with the command line arguments "-p a -p"

Expected behavior:
Error about -p being provided more than once. (this happens with the arguments "-p -p")

Actual behavior:

The program outputs:
```
-p
(a -p)
```

This problem was noted in florence/cover/issues/122

wilbowma:
So it's not clear to me that your expected behavior is in fact expected, since the final "-p" isn't considered a flag, but is an arg.

Nevertheless, I created a patch that fixes this.

wilbowma:
See also the discussion in #1364

wilbowma:
So all tests pass, but I can still imagine situation where this is not the desired behavior. Robby gave a good example on the pull request. What if instead we added something like `#:args-checked-flags`, or `#:args-no-flags` that gets this new behavior?

rfindler:
I like that approach.

rfindler:
Also: it may make sense to go poke around at the various `raco XYZ` things we have and change some of them to one of those new keywords.

wilbowma:
Actually, I think a better interface might be to restrict individual flags, e.g., `#:once-each/args` would prevent the flag from showing up more than once even in the args list; that might introduce a combinatorial problem, as you might want the "no args" modifier for every kind of flag.

jackfirth:
Doesn't the built in `--` flag cover the case where you want to pass a flag as an argument rather than a flag? `command -- foo -b bar -p` would interpret "foo", "-b", "bar", and "-p" as the list of arguments.

rfindler:
Yes. But there may be scripts out there that don't already have the -- in
them.

spdegabrielle:
is this still an issue?  Please consider adding the label `good first issue` so it is findable: 

[Issues labeled `good first issue` in in Racket GitHub repositories](https://github.com/search?q=type:issue+org:racket+state:open+label:%22good+first+issue%22)

sorawee:
Correct me if I'm wrong, but I don't think this is a "good first issue" since the conclusion is not reached yet in the above discussion.

spdegabrielle:
@sorawee my apologies - removed

mflatt:
[...]

PR #1364, opened 2016-07-06
Check "flags" in rest args against once-set table

wilbowma:
Closes issue #1342

### DIFF
--- a/racket/collects/racket/cmdline.rkt
+++ b/racket/collects/racket/cmdline.rkt
@@ -422,10 +422,48 @@
                 (map
                  (lambda (line) (cons 'final line))
                  (cdr spec))])))]
+         [check-invalid-flag
+          (lambda (table orig-multi flag)
+            (when (mpair? (caar table))
+              (let ([set (caar table)])
+                (if (mcar set)
+                    (let ([flags (mcdr set)])
+                      (raise-user-error
+                       (program-name program)
+                       (let ([s (if (= 1 (length flags))
+                                    (format "the ~a option can only be specified once" (car flags))
+                                    (format "only one instance of one option from ~a is allowed" flags))])
+                         (if (and orig-multi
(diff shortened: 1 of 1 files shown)


rfindler:
Does this affect any of the uses of command-line in our various tools?

wilbowma:
Not that I saw when testing, but I'll let Travis do a more thorough test.

--

rfindler:
Well, it would have to be something where you'd look at the uses, I think.
From the discussion on the PR this closes, it sounds like it could make it
so that, eg 'raco exe -o "My Program" -o' would no longer treat '-o' as a
source file. Maybe we don't care about that? But maybe there is a situation
where something similar would turn out to matter?

I'm not sure how to check for this kind of thing in an automated way, tho.
Ideas welcome. :)

(also, please don't change the whitespace in a line if that's the only
change you're making as I find the 'git blame' handy and that obscures it)

wilbowma:
> On Jul 5, 2016, at 20:48, Robby Findler <notifications@github.com> wrote:
> 
> Well, it would have to be something where you'd look at the uses, I think.
> From the discussion on the PR this closes, it sounds like it could make it
> so that, eg 'raco exe -o "My Program" -o' would no longer treat '-o' as a
> source file. Maybe we don't care about that? But maybe there is a situation
> where something similar would turn out to matter?
Yeah I wasn't quite sure this wouldn't break something, but no one had voiced concerns yet and I wanted to write some Racket code so I wrote the patch. :)

> (also, please don't change the whitespace in a line if that's the only
> change you're making as I find the 'git blame' handy and that obscures it)
Sorry; artifact of my editor config. I separated it into a different commit afterwards in case of this concern. I can remove it before merging.

rfindler:
Writing Racket code: I LOVE IT. :)

PR #1365, opened 2016-07-06
optimizer: add string? to list of relevant predicates

gus-massa:
Add `string?` to list of relevant predicates and a few minimal reductions.

I didn't inlude the part about `make-string` and `make-bytes` because these functions are used very few times.

I also changed `bytes?` from `SCHEME_PRIM_IS_NARY_INLINED` to `SCHEME_PRIM_IS_[U]NARY_INLINED`.


gus-massa:
I replaced the original code that recognices `make-bytes` with code that recognices `bytes-append`, because I think it's much mure used. I also added the ability to check/save the types of the arguments of `bytes-append`.

mflatt:
LGTM

samth:
This change caused the following error: http://drdr.racket-lang.org/35906/pkgs/net-test/tests/net/ftp.rkt

To see it, this should be enough:

```
$ racket -l tests/net/ftp
UNKNOWN: : read (compiled): ill-formed code [../../../racket/gc2/../src/validate.c:1389]
  context...:
   (submod /home/samth/sw/plt/extra-pkgs/net/net-test/tests/net/ftp.rkt main): [running body]
```

gus-massa:
After this commit the optimizer marks in some cases string-append as omittable, for example in

    (lambda (x y)
      (when (and (string? x) (string? y))
         (string-append x y)))

If that is removed, then the ftp test pass: https://travis-ci.org/gus-massa/racket/builds/143739213

I think the problem is that the validator don't agree, because the validator doesn't check the types of the arguments. But I don't have a minimal test yet, so I'm guessing. I'll try to see if it's possible to improve the validator.

gus-massa:
An intermediate update. I made a minimized version of _ftp.rkt_. It's difficult to recognize because the functions are renamed, and for example `thread `is replaced by a function `call `defined in the module. 

    #lang racket/base

    (define (call f)
      (display (list 1 2 3 4 5)) ;something to avoid inlining
      (f))
    
    (define (indirect-show x)
      (call (lambda () x)))
    
    (define (test)
      (indirect-show proc-0))
    
    (define ignored-AA (string-append "A" "A"))
    (define ignored-BB (string-append "B" "B"))
    (define (proc-0) 0)
    
    (module+ test (test))

To get the error I run:

    raco make ftp2.rkt
    raco test ftp2.rkt

I think that it's necesary to write the bycode and read it again to trigger the bug.

Any suggestions to reduce this example or aboid the _.zo_ are wellcome.

samth:
Here's a further-reduced program:

```racket
(module ftp2 '#%kernel
  (define-values (test) (lambda () proc-0))
  (define-values (ignored-BB) (string-append "B" "B"))
  (define-values (proc-0) (lambda () 0))
  (module* main #f (proc-0)))
```

mflatt:
I agree that the problem is a mismatch between the optimizer and the validator. This problem hasn't shown up before, because the new functions are the first ones to be omittable only under certain circumstances, but also non-foldable. (All other functions that are treated as omittable are either always omittable or foldable. And when I try to construct an example with a foldable operation but non-foldable arguments, I can't do it at the level of a module definition, in part because of the `NULL` on line 7928 of "optimize.c" that is there to prevent a mismatch between the optimizer and validator.)

I think tweaking the validator to recognize `string-append` and `bytes-append` on literals would solve the immediate problem, due to many little constraints. That seems fragile, because small improvements to the optimizer could again confuse the validator, but I don't have a better idea.

FWIW, I'm certainly tempted again to throw out the validator; keeping it in sync with the optimizer has been a pain. But it has also been a valuable tool for maintaining the optimizer, and keeping the validator on all the time is the only way to ensure that it stays in sync.

Assuming @gus-massa hasn't already done it, I'll try changing the validator.

mflatt:
After further thought: slightly inhibiting the optimizer's decision for module-level definitions is a more direct and general repair, and it only barely affects the optimizer's results.

gus-massa:
:)

I reached a similar conclusion and I was trying to understand more the details and fix it, but I was unsuccessful.

I agree that adding a special case for `(string-append "X" "Y")` in the validator is a bad idea. In particular, I still don't understand why `(string-append "X" (string-append "Y" "Z"))` doesn't generate the error.

I tried to trigger the bug with the other foldable functions, but again I couldn't get the error. I though that

    (define exp2-120 (* (expt 2 30) (expt 2 30) (expt 2 30) (expt 2 30))

was very promising (or the version with `1073741824`), but luckily it doesn't generate an error.

PR #1368, opened 2016-07-08
Add a way to recognize cpointer predicates.

samth:
Similar to `struct-predicate-procedure?`. Allows Typed Racket
(and the contract system generally) to avoid chaperone wrapping for cpointer
predicates.

Open questions:

* Is this a performance problem? I doubt it's critical.
* Should this be provided by a library not called "unsafe"?
* Do we want predicates for the other kinds of `define-cstruct` functions?

This will help resolve the regression caused by racket/typed-racket#385.

Could be merged to 6.6 if we also want to take that Typed Racket change.


mflatt:
Seems ok to me. I agree that it's unlikely to be a performance problem.

AlexKnauth:
What about `cpointer?` itself? Should that be a safe predicate?

samth:
Yes, I have some code to fix that as well.

PR #1372, opened 2016-07-11
typo: description of module-compiled-name argument

bennn:
I'm sure this is a typo, but not sure this suggested fix is right.


bennn:
@rmculpepper ping, can you proofread (or close this & make your own change) ?

rmculpepper:
Looks good to me.

PR #1374, opened 2016-07-12
edit: make object/c-helpers private

bennn:
Remove `object/c` helpers from `class-internal.rkt`.
(And add a comment that TR needs these helpers in their defining file: `class-c-old.rkt`)

Fixes the TR "undocumented public identifiers" DrDr error
http://drdr.racket-lang.org/35945/racket/share/pkgs/typed-racket-test/test-docs-complete.rkt

### DIFF
--- a/racket/collects/racket/private/class-c-old.rkt
+++ b/racket/collects/racket/private/class-c-old.rkt
@@ -22,9 +22,10 @@
          just-check-existence just-check-existence?
          build-internal-class/c internal-class/c-late-neg-proj
          class/c-internal-name-clauses
+         dynamic-object/c
+         ;; needed by TR
          base-object/c? build-object/c-type-name object/c-width-subtype?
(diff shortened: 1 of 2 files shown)


stamourv:
LGTM
Also, since those identifiers were (accidentally) being exported by TR, I would recommend including this as part of the release. That way, no one can start depending on that. Any objections?

samth:
They are in fact accidentally exported by `racket` as well.

bennn:
The accident started just 5 days ago, with https://github.com/racket/racket/commit/4650a123505e5cc37954c913c944c7e89166909b. These were never exported as part of a release -- so I don't think an announcement is necessary.

samth:
@stamourv the commit this fixes is not on the release branch.

stamourv:
@samth: Ah, no problem then.

PR #1375, opened 2016-07-14
Guard FFI union functions with type checks

takikawa:
Adds error checking for FFI union functions and also some tests.

### DIFF
--- a/racket/collects/ffi/unsafe.rkt
+++ b/racket/collects/ffi/unsafe.rkt
@@ -1162,6 +1162,8 @@
          (protect-out union-ref union-set!))
 
 (define (_union t . ts)
+  (unless (and (ctype? t) (map ctype? ts))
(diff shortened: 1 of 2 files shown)


racket/collects/ffi/unsafe.rkt
```diff
@@ -1162,15 +1162,35 @@
          (protect-out union-ref union-set!))
 
 (define (_union t . ts)
+  (unless (and (ctype? t) (map ctype? ts))
```
bennn: Shouldn't this be `andmap` ?

takikawa: Erm yes, good point. Clearly my tests aren't good enough.

LeifAndersen:
Is there any reason we haven't merged this yet @takikawa or @bennn ?

PR #1378, opened 2016-07-17
typo: syntax/modcode 'path-string' => 'path'

bennn:
Fix a few places where the docs said `path-string?` but contracts in the implementation said `path?`.


stamourv:
How about making the implementation accept path strings? That would be more flexible.

bennn:
This was an easier PR to make, that's all. If we like #1382 better we can close this.

bennn:
we like #1382

PR #1379, opened 2016-07-17
optimizer: more types anotations related with list?

gus-massa:
**First commit:**
The original idea was to refactor the latest test for the type inference of the arguments in some popular functions that use `vector?`, `string?` or `bytes?`. But the easy to write test make more visible the missing cases. In particular, many of the popular functions that use `list?` were added to the optimizer before it can recognize `list?`. (For example in an application of `map `the optimizer could recognize that the first argument was a `procedure?` but not that the second was a `list?` neither that the result is a `list?`.)

So this grows out of scope, and I think most of the changes involve `list?` but there are a few small changes related to other types.

I also added a prefix for the identifiers in `#%kernel` to make it easy to test `map` from _map.krt_ and `k:map` from the `#%kernel`.

**Second commit:**

The second commit is smaller, and I'm not sure it is correct.

In the documentation, some arguments like the position in  `(vector-ref <vector> <position>)` are documented as `exact-nonnegative-integer?` but due to the implementation details they are in a subset of `fixnum?`. 

In particular I'm worried about `length`. How long can be a list?

In the first commit I left them as `real?` because it's the smaller type that the optimizer recognizes that include `exact-nonnegative-integer?`. But I think they can be hanged to `fixnum?`

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -10,22 +10,23 @@
          racket/unsafe/ops
          compiler/zo-parse
          compiler/zo-marshal
-         ;; `random` from `racket/base is a Racket function, which makes
-         ;; compilation less predictable than a primitive
-         (only-in '#%kernel random
-                            (list-pair? k:list-pair?)))
+         (prefix-in k: '#%kernel))
+         ;; Some primitives like `random` are shadowed by Racket functions in
+         ;; `racket/base` and other modules. Using the primitive makes the
+         ;; compilation more predictable and removes the reference to the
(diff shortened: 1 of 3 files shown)


mflatt:
I see no problem with these changes.

A list's length is always a fixnum, because there are no encoding tricks that allow a list of length N to be represented by less than N pairs, and a fixnum can always hold the number of words (and therefore pairs) available in the virtual address space.

gus-massa:
The problem was with `(-)`, because `-` has a weird combination of flags that I think is the only way to trigger a bug in my commit.

I made a PR with a fix: https://github.com/racket/racket/pull/1381

Thank for the report.

PR #1382, opened 2016-07-19
change syntax/modcode signatures

bennn:
Alternative to #1378. Instead of changing the docs, this changes the functions to take `path-string?` instead of `path?`.


bennn:
@stamouv ping, can we merge this and close #1378 ?

stamourv:
It looks fine to me, but I'd prefer if @mflatt or @rmculpepper could comment.

rmculpepper:
Looks good to me.

mflatt:
I see no problems, and I agree that this is the better alternative

bennn:
thanks! https://github.com/racket/racket/commit/ba8b848f9425e385b5c698a1d14018d8ec369929

PR #1383, opened 2016-07-19
Make xrepl the default repl if it is installed

LeifAndersen:


stamourv:
:tada:
LGTM

LeifAndersen:
Okay, will merge as soon as the CI tests pass.

racket/collects/racket/init.rkt
```diff
@@ -12,3 +12,7 @@
 (provide (all-from-out racket
                        racket/enter
                        racket/help))
+
+;; Set the default replt to XREPL
+(with-handlers ([exn:fail? void])
```
samth: Couldn't this use the test that I suggested: `(collection-file-path "main.rkt" "xrepl" #:fail (lambda _ #f))`?

LeifAndersen: Yes. although FWIW, this didn't seem any better than your other suggestion, and so I went with that one. I don't really have much of a preference and would be happy to change it if you feel strongly one way or the other.

samth: The major difference is that your version will swallow other errors, such as not having readline or editline available.

LeifAndersen: That it will, is that a problem though?

samth: Depends what you want the behavior to be.

LeifAndersen: Fair. Well also iirc, xrepl actually swallows up the error itself if libedit or libreadline are not installed, and works normally, just without readline support.

Either way though, this doesn't really matter, so I'll just change it to the other way (the second one you suggested in slack using `collection-file-path` and call it good.

LeifAndersen:
merged

rfindler:
FWIW, I think it is better not to swallow errors. (cf: my dissertation and lots of pain over the years in the DrRacket REPL implementation)

PR #1384, opened 2016-07-19
Fix bugs in type errors for data/queue

schuster:
Some of the functions in `data/queue` were reporting their type errors with actual procedures rather than function names, so type errors led to errors in the error-handling code itself. I fixed that, and also fixed the "expected" description for `enqueue-front!`


stamourv:
Merged, thanks!

PR #1387, opened 2016-07-23
Add racket/interactive

LeifAndersen:


mflatt:
@LeifAndersen Thanks for doing this! Here are a few issues that I noticed:

 * `racket/interactive` starts with `#lang racket`; a library module
   should always start `#lang racket/base`, instead

 * `racket/gui/interactive` should also use `#lang racket/base`

 * Need a cast to `(Scheme_Hash_Tree *)` on line 146 of "main.c"; there
   shouldn't be any warnings when compiling Racket on Unix

 * The "config.rktd" shouldn't have to exist to get the default start
   path, and `'interactive-file` shouldn't have to be in the
   configuration table; make `'racket/interactive` the default
 
 * Docs: scheme_make_hash_tree actually takes 0 or `eq?`, 1 for
   `equal?`, or 2 for `eqv?`; I've been meaning to add `#define`s for
   those
 
 * In the `history` note for the REPL, use `@elem{...}` instead of
   `"..."` so that you can write `@racketmod[racket/interactive]`
   and `@filepath{.mzschemerc}` --- but
   don't write `@filepath{.mzschemerc}`, and instead use
   `(find-system-path 'init-file)`, since the answer varies among
   platforms; also, use a period at the end of the note

 * The `racket/gui/interactive` module is documented in the wrong
   place; it belongs in the "gui-doc" package, and the `history` note
   there should refer to that package's version (which should be
   incremented)

 * The docs should specify the symbol key expected in "config.rktd";
   consider putting that information in both the startup description
   and the "config.rktd" docs

LeifAndersen:
@LeifAndersen Thanks for doing this! Here are a few issues that I noticed:

 * `racket/interactive` starts with `#lang racket`; a library module
   should always start `#lang racket/base`, instead

Done

 * `racket/gui/interactive` should also use `#lang racket/base`

Done

 * Need a cast to `(Scheme_Hash_Tree *)` on line 146 of "main.c"; there
   shouldn't be any warnings when compiling Racket on Unix

Done

 * The "config.rktd" shouldn't have to exist to get the default start
   path, and `'interactive-file` shouldn't have to be in the
   configuration table; make `'racket/interactive` the default
 
Fair enough, done.

 * Docs: scheme_make_hash_tree actually takes 0 or `eq?`, 1 for
   `equal?`, or 2 for `eqv?`; I've been meaning to add `#define`s for
   those
 
Ah, okay, thanks. Done Also `#define`s added.

 * In the `history` note for the REPL, use `@elem{...}` instead of
   `"..."` so that you can write `@racketmod[racket/interactive]`
   and `@filepath{.mzschemerc}` --- but
   don't write `@filepath{.mzschemerc}`, and instead use
   `(find-system-path 'init-file)`, since the answer varies among
   platforms; also, use a period at the end of the note

Done

 * The `racket/gui/interactive` module is documented in the wrong
   place; it belongs in the "gui-doc" package, and the `history` note
   there should refer to that package's version (which should be
   incremented)

While I agree the documentation is a bit odd there, I put it there because that's where racket/gui/init is also documented. Do you think we should move that as well?

Also, I'm not sure what you mean by the package's version should be incremented? Is the gui-lib on version 6.8 rather than 6.7?

Thanks

 * The docs should specify the symbol key expected in "config.rktd";
   consider putting that information in both the startup description
   and the "config.rktd" docs

Done

Thanks for the feedback.

pkgs/racket-doc/scribblings/raco/config.scrbl
```diff
@@ -178,4 +178,11 @@ directory}:
        @racket[find-config-tethered-console-bin-dir] and
        @racket[find-config-tethered-gui-bin-dir].}
 
+  @item{@indexed-racket['interactive-file] and
+        @indexed-racket['gui-interactive-file] --- a path to the interactive module
```
mflatt: "path" should be "module path"

pkgs/racket-doc/scribblings/raco/config.scrbl
```diff
@@ -178,4 +178,11 @@ directory}:
...
+        that runs when the REPL runs on startup, unless the
+        @Flag{q}/@DFlag{no-init-file} is provided. Defaults to
+        @racketmodname[racket/interactive] and
+        @racketmodname[racket/gui/interactive].}
```
mflatt: Technically, I think these should be `@racket['racket/interactive]` and `@racket['racket/gui/interactive]`, since they're symbols that are used as module paths.

pkgs/racket-doc/scribblings/reference/startup.scrbl
```diff
@@ -102,7 +102,8 @@ thread, unless the @Flag{V}/@DFlag{no-yield} command-line flag is
 specified. Requiring @racketmodname[racket/gui/base] sets this parameter call
 @racket[(racket 'yield)].
 
-@history[#:changed "6.7" "Run racket/interactive file rather than directly running .racketrc"]
+@history[#:changed "6.7" @elem{Run @racketmodname[racket/interactive] file
+         rather than directly running @racket[(find-system-path 'init-file)]}]
```
mflatt: Needs a terminating period in the `@elem{}`

racket/src/racket/cmdline.inc
```diff
@@ -672,7 +672,7 @@ static int finish_cmd_line_run(FinishArgs *fa, Repl_Proc repl)
     Scheme_Object *a[2];
     a[0] = get_init_filename(fa->global_env, INIT_FILENAME_CONF_SYM, DEFAULT_INIT_MODULE);
     if (a[0]) {
-      a[1] = scheme_false;
+      a[1] = scheme_make_integer(0);
```
mflatt: Was this to avoid problems on Windows? I think `scheme_make_false()` is the right choice.

mflatt:
Thanks! I added some small comments to the commits.

Good point about `racket/gui/init` docs. I do think those should move. The current version of the "gui-lib" package is 1.26.

pkgs/racket-doc/scribblings/raco/config.scrbl
LeifAndersen: Okay done

pkgs/racket-doc/scribblings/raco/config.scrbl
LeifAndersen: okay done.

racket/src/racket/cmdline.inc
LeifAndersen: Yes it was, outside of the technical difference the docs give, I still don't have a good intuition when to use 0 or #f. Anyway, `scheme_make_false()` works. Do you know why the literal `scheme_false` doesn't work on windows?

pkgs/racket-doc/scribblings/reference/startup.scrbl
LeifAndersen: Woops, thanks.

LeifAndersen:
Great, thanks for the additional feedback. Sorry about missing the period, I should have caught that. I'd be happy to move the `racket/gui/interactive` and `racket/gui/init` docs to the `gui-lib` package.

LeifAndersen:
Hmm...actually, it's not entirely obvious where the best place to put the documentation for `racket/gui/interactive` and `racket/gui/init` is. Do you have any suggestions?

LeifAndersen:
Actually, never mind. I just decided the right thing to do here was to create a new section.

mflatt:
Ok – push when you're ready.

LeifAndersen:
Okay, I also added a PR to https://github.com/racket/gui/pull/37

I'll push them both as soon as the tests finish here. Thanks.

lexi-lambda:
(Can this PR be closed, given that it looks like the relevant commits have been merged?)

PR #1395, opened 2016-07-27
Fix for-label requires for some struct documentation

lexi-lambda:
Commit f549724e36c141989f4d1d2d103f5b65e740865e accidentally incorrectly updated the `for-label` requires for a couple places in the documentation; this adjusts them to be correct.


stamourv:
Merged, thanks!

Issue #911, opened 2015-04-01
racket/vector should include vector-sort and vector-sort!


lexi-lambda:
Currently, `racket/vector` doesn't include any sorting implementations, which is especially annoying when wanting a performant, in-place sort.

The `rnrs/sorting-6` module actually *does* provide these two functions, but they're pretty worthless.

```racket
(define (vector-sort < v)
  (list->vector (sort (vector->list v) <)))

(define (vector-sort! < v)
  (let ([v2 (vector-sort < v)])
    (vector-copy! v 0 v2)))
```

Not only is the vector to list conversion unnecessary, but the in-place sort isn't in-place at all, which sort of defeats the point. Ideally, `vector-sort` would just be implemented in terms of `vector-sort!`, just copying the vector first.

lexi-lambda:
Closing, as I believe this was addressed by https://github.com/racket/racket/commit/42f478473583519048602b6e7563dbb57b3c51d9.

PR #1398, opened 2016-07-27
add vector-sort to racket/vector

andmkent:
Adds ```vector-sort``` and ```vector-sort!``` to racket/vector, addressing #911 (and I want to use this in Typed Racket... =)

It seems a bit faster in general than the merge sort implementation that racket/list's ```sort``` uses, but you don't get the 'stable sort' property ```sort``` gives you (however, that implementation has to allocate another vector to work in, while this implementation (introsort) works in place). Also, it has the same worst-case time complexity as ```sort``` (n log n).

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/pairs.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/pairs.scrbl
@@ -528,7 +528,7 @@ which takes two elements of @racket[lst] and returns a true value if the
 first is less (i.e., should be sorted earlier) than the second.
 
 The sort is stable; if two elements of @racket[lst] are ``equal''
-(i.e., @racket[proc] does not return a true value when given the pair in
+(i.e., @racket[less-than?] does not return a true value when given the pair in
 either order), then the elements preserve their relative order from
 @racket[lst] in the output list.  To preserve this guarantee, use
 @racket[sort] with a strict comparison functions (e.g., @racket[<] or
--- a/pkgs/racket-doc/scribblings/reference/vectors.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/vectors.scrbl
@@ -366,5 +366,56 @@ Like @racket[vector-member], but finds an element using @racket[eq?].
 (vector-memq 9 (vector 1 2 3 4))
 ]}
 
+@defproc[(vector-sort [vec vector?]
+                      [less-than? (any/c any/c . -> . any/c)]
+                      [start exact-nonnegative-integer? 0]
+                      [end exact-nonnegative-integer? (vector-length vec)]
+                      [#:key key (any/c . -> . any/c) (λ (x) x)]
+                      [#:cache-keys? cache-keys? boolean? #f])
+         vector?]{
+
+ Like @racket[sort], but operates on vectors; a 
+ @emph{fresh} vector of length @racket[(- end start)] is
+ returned containing the elements from indices 
+ @racket[start] (inclusive) through @racket[end] (exclusive)
+ of @racket[vec], but in sorted order (i.e., @racket[vec] is
+ not modified). This sort is stable (i.e., the order of ``equal''
+ elements is preserved).
+ 
+@mz-examples[#:eval vec-eval
+(define v1 (vector 4 3 2 1))
+(vector-sort v1 <)
+v1
+(define v2 (vector '(4) '(3) '(2) '(1)))
+(vector-sort v2 < 1 3 #:key car)
+v2]
+
+@history[#:added "6.6.0.5"]{}
+}
+
+@defproc[(vector-sort! [vec (and/c vector? (not/c immutable?))]
+                       [less-than? (any/c any/c . -> . any/c)]
+                       [start exact-nonnegative-integer? 0]
+                       [end exact-nonnegative-integer? (vector-length vec)]
+                       [#:key key (any/c . -> . any/c) (λ (x) x)]
+                       [#:cache-keys? cache-keys? boolean? #f])
+         void?]{
+
+ Like @racket[vector-sort], but @emph{updates} indices
+ @racket[start] (inclusive) through @racket[end] (exclusive)
+ of @racket[vec] by sorting them according to the @racket[less-than?]
+ procedure.
+
+@mz-examples[#:eval vec-eval
+(define v1 (vector 4 3 2 1))
+(vector-sort! v1 <)
+v1
+(define v2 (vector '(4) '(3) '(2) '(1)))
+(vector-sort! v2 < 1 3 #:key car)
+v2]
+
+@history[#:added "6.6.0.5"]{}
+}
+
 
 @close-eval[vec-eval]
--- a/pkgs/racket-test-core/tests/racket/vector.rktl
+++ b/pkgs/racket-test-core/tests/racket/vector.rktl
@@ -229,4 +229,197 @@
   (err/rt-test (vector-map (lambda (x) x) #() #() #()) (check-regs #rx"vector-map" #rx"mismatch between procedure arity")))
 
 
+;; ---------- vector-sort basic ----------
+(test #("a" "b" "c" "c" "d" "e" "f")
+      vector-sort
+      #("d" "f" "e" "c" "a" "c" "b")
+      string<?)
+(test #("a" "c" "e")
+      vector-sort
+      #("d" "f" "e" "c" "a" "c" "b")
+      string<?
+      2
+      5)
+(test #(("a") ("c") ("e"))
+      vector-sort
+      #(("d") ("f") ("e") ("c") ("a") ("c") ("b"))
+      string<?
+      2
+      5
+      #:key car)
+(test #(("a") ("c") ("e"))
+      vector-sort
+      #(("d") ("f") ("e") ("c") ("a") ("c") ("b"))
+      string<?
+      2
+      5
+      #:key car
+      #:cache-keys? #t)
+
+(let ()
+  (define (car< x y) (< (car x) (car y)))
+  (define (random-vec n range)
+    (build-vector n (λ _ (list (random range)))))
+  ;; sort a vector, then sort it with a #:key and with #:cache-keys?
+  ;; and make sure they're all doing the same thing
+  (define (vector-sort* v [start 0] [end (vector-length v)])
+    (let ([v1 (vector-sort v car< start end)]
+          [v2 (vector-sort v < start end #:key car)]
+          [v3 (vector-sort v < start end #:key car #:cache-keys? #t)])
+      (test #t equal? v1 v2)
+      (test #t equal? v1 v3)
+      v1))
+  (define (test-sort len times)
+    (or (zero? times)
+        ;; build a random vector, sort it, check that it's sorted
+        (and (let* ([rand (random-vec len (if (even? times) 1000000 10))]
+                    [sorted (vector-sort* rand)])
+               (and (= len (vector-length sorted))
+                    (or (<= len 1)
+                        (let loop ([i 0] [i+1 1])
+                          (or (= i+1 len)
+                              (and (let ([ival   (car (vector-ref sorted i))]
+                                         [i+1val (car (vector-ref sorted i+1))])
+                                     (<= ival i+1val))
+                                   (loop (add1 i) (add1 i+1))))))))
+             (test-sort len (sub1 times)))))
+  (test #t test-sort    1  10)
+  (test #t test-sort    2  20)
+  (test #t test-sort    3  60)
+  (test #t test-sort    4 100)
+  (test #t test-sort    5 100)
+  (test #t test-sort   10 100)
+  (test #t test-sort  100 100)
+  (test #t test-sort 1000 100)
+  ;; test stability for vector-sort
+  (test #((1) (2) (3 a) (3 b) (3 c)) vector-sort* #((3 a) (1) (3 b) (2) (3 c)))
+  ;; test short lists (+ stable)
+  (test #() vector-sort* #())
+  (test #((1 1)) vector-sort* #((1 1)))
+  (test #((1 2) (1 1)) vector-sort* #((1 2) (1 1)))
+  (test #((1) (2)) vector-sort* #((2) (1)))
+  (for-each (λ (v) (test #((0 3) (1 1) (1 2)) vector-sort* v))
+            '(#((1 1) (1 2) (0 3))
+              #((1 1) (0 3) (1 2))
+              #((0 3) (1 1) (1 2))))
+  (for-each (λ (v) (test #((0 2) (0 3) (1 1)) vector-sort* v))
+            '(#((1 1) (0 2) (0 3))
+              #((0 2) (1 1) (0 3))
+              #((0 2) (0 3) (1 1))))
+  ;; exhaustive tests for 2 and 3 item lists
+  (for-each (λ (v) (test #((1 x) (2 y)) vector-sort* v))
+            '(#((1 x) (2 y))
+              #((2 y) (1 x))))
+  (for-each (λ (v) (test #((1 x) (2 y) (3 z)) vector-sort* v))
+            '(#((1 x) (2 y) (3 z))
+              #((2 y) (1 x) (3 z))
+              #((2 y) (3 z) (1 x))
+              #((3 z) (2 y) (1 x))
+              #((3 z) (1 x) (2 y))
+              #((1 x) (3 z) (2 y))))
+  (test #((2)) vector-sort* #((4) (2) (1) (3)) 1 2)
+  (test #((1) (2)) vector-sort* #((4) (2) (1) (3)) 1 3)
+  (test #((1) (2) (3)) vector-sort* #((4) (2) (1) (3)) 1 4)
+  (test #((1) (2) (4)) vector-sort* #((4) (2) (1) (3)) 0 3)
+  (test #((1) (2) (3) (4) (5) (6) (7) (8)) vector-sort*
+        #((4) (2) (1) (3) (6) (5) (8) (7) (10) (9))
+        0 8)
+  (test #((1) (3) (5) (6) (7) (8) (9) (10)) vector-sort*
+        #((4) (2) (1) (3) (6) (5) (8) (7) (10) (9))
+        2 10)
+
+  (define ((check-regs . regexps) exn)
+    (and (exn:fail? exn)
+         (andmap (λ (reg) (regexp-match reg (exn-message exn)))
+                 regexps)))
+
+  (err/rt-test (vector-sort! (list 1) <) (check-regs #rx"vector-sort!" #rx"vector" #rx"immutable"))
+  (err/rt-test (vector-sort! #(1) <) (check-regs #rx"vector-sort!" #rx"vector" #rx"immutable"))
+  (err/rt-test (vector-sort (list 1) <) (check-regs #rx"vector-sort" #rx"vector"))
+  (err/rt-test (vector-sort! (vector 1) 1) (check-regs #rx"vector-sort!" #rx"any/c any/c . -> . any/c"))
+  (err/rt-test (vector-sort (vector 1) 1) (check-regs #rx"vector-sort" #rx"any/c any/c . -> . any/c"))
+  (err/rt-test (vector-sort! (vector 1) (λ (x) x)) (check-regs #rx"vector-sort!" #rx"any/c any/c . -> . any/c"))
+  (err/rt-test (vector-sort (vector 1) (λ (x) x)) (check-regs #rx"vector-sort" #rx"any/c any/c . -> . any/c"))
+  (err/rt-test (vector-sort! (vector 1) < #:key 42) (check-regs #rx"vector-sort!" #rx"any/c . -> . any/c"))
+  (err/rt-test (vector-sort! (vector 1) < #:key <) (check-regs #rx"vector-sort!" #rx"any/c . -> . any/c"))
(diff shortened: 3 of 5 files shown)


rfindler:
The implementation in racket/private/sort also used an inplace vector sorting algorithm (in some cases). Should these perhaps be compared directly and then use the best one for both?  (There may be some dependency problems that make this too painful, tho.)

racket/collects/racket/private/vector-sort.rkt
```diff
@@ -0,0 +1,404 @@
...
+
+;; - - - - - - - - - - - - - - - - - - -
+;; Magic Numbers
+;; - - - - - - - - - - - - - - - - - - -
+
+;; Constants that determine which sort is used when
+
+;; at what size should quicksort resort to insertion sort?
+(define insertion-sort-cutoff 17)
+;; how deep should quicksort go before switching to
```
samth: "switching to" ...

andmkent:
@rfindler I perused racket/private/sort again, I don't think there's much interesting code to be shared. They both use insertion sort as a helper for small vector subsections, but they're not very large (code wise) or complex.

I do think, however, It would be nice at some point to use (pretty much verbatim) the code in racket/private/sort to implement and provide a ```vector-stable-sort``` function from racket/vector (at which point ```sort``` could be defined in terms of the internal implementation of ```vector-stable-sort```).

stamourv:
The non-uniformity between list and vector sort re stability is unfortunate. Is there no way around it?
As is, if someone switches from using lists to vectors in their program, its results may change. And with generic sequences, where your code doesn't even know/care whether it's processing lists or vectors, that would be even more surprising.

andmkent:
The solution to this is to provide a vector-stable-sort.

It's unfortunate that stability was baked into the spec for the one and
only standard list sorting algorithm.

andmkent:
I can port the vector-based stable sort over to racket/vector today if
people feel this is important.

rfindler:
I don't think it is unfortunate that stability is part of the spec of the built-in `sort` function. I think it is the right default.

I also agree with @stamourv that vector-sort should be stable. Perhaps you want to add a `vector-unstable-sort` function?

andmkent:
_Skippable, possibly unneeded tangent:_ IMHO, if stability is a part of the correctness criteria of your program (or is desired for some other reason), (1) it should be made explicit and call a function named something like```stable-sort``` rather than eliding that detail and allowing someone to unknowingly introduce a subtle bug at some point in the future, and (2) users who do not require stability should by default get a faster sort that uses less memory -- obviously different languages (and people) have different opinions about this =) (a perusal of respectable languages sees both defaults, for example).

Perhaps Racket---in specifying in the docs that ```sort``` is stable for quite some time now---has _already_ committed to ```*-sort``` functions being stable by default?

Anyway, that's all I've got to say about that and I'll shut up =)

So then, how does the room feel? Would people be most happy if ```vector-sort``` is a stable sort and the standard library also provides ```unstable-sort``` and ```vector-unstable-sort``` as list/vector sorts which provide no stability guarantee but are free to attempt to work in less space and be faster in general?

andmkent:
If that's the case, I'll go ahead and change this PR to just be the unstable-sort additions, and the stable ```vector-sort``` will be a different PR.

rfindler:
Your argument "if something is part of the requireents for correctness, it should be explicitly documented" along the lines you write suggests that we should write "safe-will-signal-an-error-on-non-numbers-+" instead of "+" and other such abominations.

andmkent:
I think sort stability and throwing an error when trying to add non-numbers are apples and bananas. To each there own fruit I suppose.

rfindler:
(in reality it is a judgment call exactly what guarantees get put into what primitive operations in a language design. It is my judgment that stable sorting is a good choice and my main evidence for that is that large number of subtle bugs that have crept in due to hash-table ordering implicit dependencies. My guess is that unstable sorting would also tend to produce such subtle bugs.)

andmkent:
Fair enough =)

rfindler:
Your comments did not seem to be specific to sorting in any way.

andmkent:
The comments were not meant to be only about sorting, but were _also_ not meant to be a universal truth equally applicable to all program features =)

Anyway! I am happily making this PR implement ```vector-unstable-sort``` -- thanks for all your feedback and insight. It is much appreciated.

elibarzilay:
> I perused racket/private/sort again, I don't think there's much interesting code to be shared.

This is a weird thing to say, given that the private/sort converts lists to arrays for sorting:

* For the `vector-sort` case, using private/sort would be very easy:
  - the trivial 0/1 cases are the same
  - the `cache-keys?` case needs different (and simpler) conversion steps
  - the already sorted case is also similar except for a different (simpler) loop
  - the <=3 case could be implemented with almost the same code (I'd use an internal macro which abstracts over pulling the values with car/cadr/caddr and over the constructor for the result)
  - and the last case is also similar with a different (simpler) pre/post conversion bits

* Related nit: regardless of how it's implemented, it's wrong for a functional `vector-sort` to be documented as "returns a new vector".  Instead, the documentation should just say that it does not mutate the input vector — this way it is still possible to return the same vector in case it's already sorted.

* I'm guessing that you'd have hard time beating the performance of the result of doing this.

* For `vector-sort!` the story is probably different — both due to very different performance when you can mutate the input, and also because of the human factor that makes it something that is rarely used (in racket).  Therefore, a practical approach might be building on the above to keep it simple, and on the conversion back to an output vector simply bang the elements into the input vector.  It will probably be slower, but probably not as much, and possibly within a factor of "hukairs" given that it's the destructive version.  Or have some traditional destructive (but stable) sort done in that case.


> They both use insertion sort as a helper for small vector subsections, but they're not very large (code wise) or complex.

The code is not too large, but the result of compiling it is pretty big since a few different versions are compiled to make it perform better, and it takes care to optimize some common cases in a jit-friendly way.  That part of the implementation is much trickier than the actual sorting, so IMO is's definitely worth it to avoid doing it all from scratch.


As for the stability issue:

> allowing someone to unknowingly introduce a subtle bug at some point in the future

The thing is that a stable sort is *unsurprising*, or IOW, it's more defined, and therefore it cannot introduce subtle bugs.  It's the other direction (having an unstable sort) that can lead to subtle bugs.

The main challenge is speed, and when I looked at a whole pile of sorting algorithms, among the generic ones (ie, not very specialized ones like radix sort) the runtime cost was negligible to non-existent.

Finally, an even bigger reason that the plain `sort` is stable (and explicitly so!  IIRC, the previous might have had both a sort and a stable sort) is that it's common across the scheme and the bigger lisp world to have stable sorts.  And the reason for *that* is similar to what Robby said: a stable sort is more defined, and therefore less surprising, and therefore a better default.


---

(BTW, `i≤` is easy to confuse with `i<=` so IMO it's over-unicoded.  Also, the `while`s make my eyes tear.)

gus-massa:
> Related nit: regardless of how it's implemented, it's wrong for a functional vector-sort to be documented as "returns a new vector". Instead, the documentation should just say that it does not mutate the input vector — this way it is still possible to return the same vector in case it's already sorted.

I'm worried bout this possibility. If it's an immutable vector, I don't mind that the result is the same vector. But if it is a mutable vector I think that the possibility of returning the same vector can be confusing. (If the input vector is immutable but not sorted, should the return vector be immutable?)

(This is not a problem with lists, because lists are always immutable.)

For example, is this code correct?

    ;imagine something like a genetic algorithm
    (lambda (make-new-generation old-generation)
      (define new-generation (vector-sort old-generation > #:key points))
      (vector-set! new-generation
                   (sub1 (vector-length new-generation))
                   (mix (vector-ref new-generation 0)
                        (vector-ref new-generation 1))))

rfindler:
I think it makes the most sense to have functions that return mutable datastructures be clear about when they return the same (in the sense of a mutation in one will mutate the other) vector.

andmkent:
Here's the initial performance I'm seeing sorting lists with the vector-backed ```sort``` and ```unstable-sort```:

https://gist.github.com/andmkent/71e4ada23d8e221ab254f07b68869546

So yah, not hugely different (some of the improvement are a few things I could tweak in private/sort as well).

andmkent:
whoops (didn't mean to close -- buttons are so clickable =)

elibarzilay:
@gus-massa

> I'm worried bout this possibility. If it's an immutable vector, I don't mind that the result is the same vector. But if it is a mutable vector I think that the possibility of returning the same vector can be confusing. (If the input vector is immutable but not sorted, should the return vector be immutable?)

This would only be a problem when you're mixing mutable and immutable things...

> For example, is this code correct?

... as in this example, where you're using the functional `sort` but then expect the result to be a copy and therefore fine to mutate.  But yeah, there's a point, so maybe it should also say that it (`sort`) always produces an immutable result and that makes it possible to return the input in case it's sorted and immutable.  (I'm thinking about a loop that keeps sorting the same thing which is already sorted.)

But in any case, that was a minor nit.

@rfindler

> I think it makes the most sense to have functions that return mutable datastructures be clear about when they return the same (in the sense of a mutation in one will mutate the other) vector.

So I think that the above should work for this.  The thing that seems wrong to me is to guarantee a fresh copy, since there are plenty of cases where you're sorting a list that can already be sorted, so it's a big win to allow retuning it without copying it.

@andmkent

> Here's the initial performance I'm seeing sorting lists with the vector-backed sort and unstable-sort:

1. I don't know how you measured it, or which versions you compared, but some of these numbers seem wrong.  For example, there are large differences in the sorted section, where the two should have the same behavior (since they both should test that the input is sorted).  But in any case, if you take from this that the difference is minor enough to not keep an unstable version then I'm happy.

2. However, the first point was more important: I think that this is a case where it is much better to keep a single version of the code.  Not because the sorting code is complicated, but because stuff around it (as the multiple precompiled-via-macro versions, and the unsafe bindings) are very important since that's where most of the optimizations happen -- and it's not trivial and it's likely to change with things like new optimizations and jit strategies.

andmkent:
There is still some tidying up to do, testing, etc... but I just wanted to push now to show how I have shared the sort and unstable sort code to get any feedback and that aspect.

andmkent:
After tinkering around with different options a bit, I decided for now to just put a vector front end on the code already in private/sort.rkt, which adds ```vector-sort``` and ```vector-sort!``` to racket/vector.

andmkent:
(just fixed a typo and deleted a file I didn't mean to add -- I believe it's ready unless more tweaks are desired from the general audience)

rfindler:
I think it would be good to go back and make sure that the changes in the diff are all substantive (i.e., no reformatting changes). If the reformatting is important, maybe make it a separate commit? 

Just in general, when going back over the history of some file I find it very very helpful to correlate certain version and certain changes with specific lines when I'm trying to figure out what's going on and when there are lots of whitespace changes, it makes it very difficult to do that effectively.

andmkent:
I'll see if I can tidy up sort.rkt w.r.t. white space. I was trying to stick with DrRacket indenting defaults but the file was indented in a way to minimize right creep... I need to work on my white
space foo.

rfindler:
Thanks. I know it is a pain to keep this old, crappy whitespace around.

andmkent:
There shouldn't be any white space changes for the sake of white space changes in the second commit now. The first commit moved an encompassing "(let () ..." to be on it's own line so the body of the file had sane indenting defaults even with the no-indenting style the top of the file uses (without moving the let, every time DrRacket attempts to place the cursor in the correct place it was way off, and that got real old real quick -- I think the current spacing is a nice compromise)

elibarzilay:
I wasn't sure that I'm looking at the right thing, so I ended up going to your repo, so this is based on what I see there, after commit 33307ab.  (Also, this came out very long, I think that it's better to start from the existing file and add the new functionality instead of trying to backtrack things.)

1. You added a dependency on `qqstx.rkt`, which looks fishy to me.  All of the uses of `` #` `` and `` #, `` are simple enough to be easily eliminated.  Having minimal dependencies was important to be able to use sorting early enough, so I'd try to avoid adding dependencies even if it looks like it works now.  (More on these uses below.)

2. "note, these are the raw interfaces to --- user-facing definitions" -- something's wrong with this sentence.  Maybe there's a missing `sorting` before the `---`?

3. You dropped the "Note that there is no error checking on the arguments" comment, which seems wrong, since the new code still doesn't check things (and more on your keyword arguments later).

4. The code is still shifted by two spaces.  I've seen your description of this being a compromise, but the result that Robby is complaining about is that by default this commit looks like the whole file is rewritten.  (The reasons for the weird indentation is that the `define-values` and `let` expressions are kind of global annotations for being able to use local macros that are not needed in the resulting zo after expansion.  Whether that is a good idea or not is an unrelated discussion...)  (Another unrelated discussion is how DrR indents these -- it looks like it's aligning each `(define` with where it thinks it should be wrt the surrounding expression rather than adopting the preceding `(define` which is what Emacs does -- that's probably the source of the editing annoyances.)

5. `;; end of module` at the end is redundant.

6. You've kept my helper definition of `define-syntax-rule`, but there's a bunch of macros that should use it instead of the more verbose form: `sort-internal-body`, `precomp`, `list-sort-body`, `vector-sort-body`.

7. A repeated conception mistake (IMO!) that I see is seen in `sort-internal-body`: you define a pattern with a `#:maybe-key` "keyword argument".  But this is not a keyword argument, just part of a pattern that must appear exactly in that place.  This is even worse when there are more than one of these things.  So (again -- IMO) it would be much better to drop it to avoid confusion that these are keywords in the usual sense.

8. Also, there are macro uses that are not needed.  An example of this is the `<?` helper: in your version you're using a nested (unquoted) macro code to apply key or not -- compare that with my version that just has plain code that is wrapped in a `(if has-getkey? ...)`.  The idea is to let the Racket optimizer deal with the fact that `has-getkey?` is a literal that is known at compile-time.  I've spent considerable time making these things work, so in general I think that you should be very careful in removing them only if you know what it's doing now, and that the optimizer indeed fails on some things that require using more complicated macros.  (I'm guessing that in this case there's no need for the more complicated code.)  That's the complex part that I was referring to earlier, BTW.

9. Also, in the same dept as changing the indentation, you have lots of litte name changes, like `n` → `index`, `v` → `A`, and more.  (My naming scheme was intentional, BTW, I intensly hate long-but-trivial names like `index`.)  I'm no longer maintaining this code -- but if you want to be in that position, then I think that it would be much better to have a bunch of no-op commits.  For example, one commit that will only change the indentation, and another that will change the names.  (But only if you want to become the parent of this code...)

10. There is a new `swap!` macro definition that is not used at all.

11. Then there's `let*` for `n/2-` and `n/2+` which is at a weird place (and a redundant change from my code, again), with a comment above it that is now duplicated from the one left at the beginning of the file.

12. Then you swapped `copying-insertionsort` and `merge` for some reason (more unnecessary changes).  (BTW, not that at this point you're already way more indented than my code -- due to various things mentioned above.)

13. As a side-note, I also dislike big ";; Foo" comments that precede a "(define foo ...)" without adding more information.  (You're actually reducing information from `copying-mergesort` to `Mergesort`.)  Also, `start the sorting!` is kind of a similarly redundant `main body` comment.

14. Another reundant change is `(i>> n 1)` → `(i/2 n)` (in the same spirit as above, I'd categorize this in the renames dept).

15. The `for standard data types` comment is misleading -- it sounds like implementing sorting for multiple types of sequences, but it's only for multiple types that are being sorted over.

16. You've added a bunch of precompiled sorts there -- any reason?  I was specifically stingy on these since they'd all generate code in the zo, so I wanted to keep it minimal, and have it resort to the usual runtime in exceptions.  For example, I avoided `char`s since IME sorting a list of characters is extremely rare in practice.  For the same reason I avoided the `string>` variants.  Are you sure that you want to add them?

17. You're converting the whole thing into an immutable hash table at the end -- did you time it and it was faster?  Otherwise this is mutability in a place that is not user-visible, so speed should be preferred.

18. I don't get the point behind `generic-sort` and `generic-sort/key`.  All they seem to do is translate the old no-keyword style to the new "keyword-like argument" style.  And if so, then the keywords are not doing anything at all (besides making the code hard to read...).

19. The `if we know statically` comment seems to be something that you wanted to do but didn't...  That `syntax-case` is always going to end up in the second branch.  (BTW, it looks like what you wanted to do is again something that the racket optimizer is better at doing, as in the above case.)

20. You lost the `(< n 2)` case??  Instead, you check it in the `(i<= n 3)` case where it couldn't happen (since it's always sorted).  (And, of course, you end up redundantly checking that it's sorted...)

21. You've moved/copied the `<?` helper further down which makes for more indentation -- it's a macro so there's no need for that (other than, I'm guessing, macro-based optimization that racket could do).

22. In the `General note` comment I had some clever comment that Jay took on and implemented a macro-expanded sort, it would be a shame to lose it.  (Also, you have a case of a bunch of trailing spaces at the end of that comment block.)

23. Comment about *my* code: looks like I ended up with a duplicated `let` for the same bindings (`([a (car lst)] [b (cadr lst)] [c (caddr lst)])`).

24. Ummm...  So I reach the end of this, and I see that there's nothing on vectors, and then I see that the next section duplicates a bunch of code...  I'm too tired to follow that, but my intention in one of the earlier comments was to use the same toplevel function to deal with both cases instead of duplicating it.

25. Also, I didn't check the current code, but at least in the PR it looks like you copied the sorting tests to the vector tests.  I think that it's better to create a new test suite that will do both at the same time -- just run each test once as a list and once as a vector.

andmkent:
I don't have a much free time to tinker with this any more and I've made due without it for my own purposes. I'd be happy to make some minor changes at some point in the future, but if it's nowhere near what we want for Racket's standard ```vector-sort``` then someone else is welcome to tackle the need when they see fit (starting from this code or from scratch, whatever is easier).

@elibarzilay -- IMHO, using the inlined commenting feature that Github provides is very helpful when you have comments about specific places in code.

samth:
The docs for the new functions should have a `@history` annotation saying when they were added. Otherwise, looks good to me.

elibarzilay:
@samth, I'm assuming that you have seen my detailed comments, which I spend a good amount of time writing to make them useful and easy to follow, as well as the "not-gonna-do-it" reply.

Using a four letter word would have been more polite.

samth:
Almost all of your comments were addressed in the final commit.

rfindler:
It looks like maybe the commit that got merged is this one? I'm sure we are all open to continuing to work on Racket here so (I hope) that further feedback would be read and thought about and probably acted on?

https://github.com/racket/racket/commit/42f478473583519048602b6e7563dbb57b3c51d9

And maybe even a high-level explanation of what did & didn't get followed from Eli's earlier comments?

andmkent:
As a closing remark on this PR, here is a brief high level summary of 
the changes:

- new ```vector-sort``` and ```vector-sort!``` functions were added to
  racket/vector (w/ supporting code added to racket/private/sort.rkt).
  these functions, like ```sort```, provide a _stable_ sort.

- The vector and list sorting algorithms use the same back end, with
  slightly different front ends (the front ends are similar but are
  specialized to handle lists/vectors before calling the common back
  end).

- The set of precompiled sorts was expanded to include all of the
  standard comparators from racket/base


Thank you to all those who helped w/ suggestions and comments
(especially @elibarzilay). I tried to incorporate all the suggestions
in the final product (only intentionally ignoring some
cosmetic/stylistic suggestions that seemed arbitrary in the big
picture, and I didn't do anything fancy to shrink the tests down).
More specifically, all of the following suggestions were incorporated:

- no unnecessary dependencies (e.g. qqstx.rkt were added)
- several redundant/unnecessary macros/special casing on constants
   from the earlier draft were removed
- the generic-sort/key functions were removed
- other small corrections were made

elibarzilay:
@samth:

> Almost all of your comments were addressed in the final commit.

Apologies!  I completely missed the additional commit.


@rfindler:

> I'm sure we are all open to continuing to work on Racket here

I certainly hope that this code won't become orphan...  (And unfortunately, the kind of time requirement for suggesting a pull request for such things is way more than I can afford...)


@andmkent:

> As a closing remark [...]

Thanks for updating it; there are still some things that are bugging me, below is quick list up to the point where I gave up...

But one major thing which you list as a feature is the addition of precompiled versions.  Are these actually needed?  I'm guessing that sorting characters is pretty rare and not a performance bottleneck, and the same goes for things like `string-ci>?` and `i>`.  The resulting zo file that I see is about twice bigger than the previous one, and this library is loaded very early and therefore probably used in a bunch of places where load times and bytecode sizes are important.

I wasn't sure that I'm looking at the right thing, so I ended up going to your repo, so this is based on what I see there, after commit 33307ab.  (Also, this came out very long, I think that it's better to start from the existing file and add the new functionality instead of trying to backtrack things.)

The rest of the list, in case you want to improve it:

1. The code is still shifted by two spaces and still has different structure that leads to a much deeper nesting, and still has a bunch of redundant renamings.  But the damage from all of these is already done, too late to seperate these into no-op commits.

2. `;; end of module` at the end, and some other comment headers that are redundant (ie, a big "foo" comment before a definition of "foo").

3. Macros are still bogusly using keywords which look like regular keyword arguments but are not: `sort-internal`, `list-sort-body`, `vector-sort-body`, `vector-sort-case-lambda`.

4. `list-sort-body` and `vector-sort-body` should use `define-syntax-rule`.

5. There are still macros that are not needed like `<?`, as well as slight changes in the code that might lead to changes in how its optimized (the `<?` comes with a higher-order mention of the key function, and I vaguely remember this having a negative effect).

6. `i*2` and `i/2` are unsafe, but they're not mentioned in the list of things to uncomment to get a safe version of the code.

7. The `swap!` macro is redundant since it's used just once, so the code could just move.

8. The `for standard data types` is still misleading.

PR #1404, opened 2016-07-29
Fix bug with #:defined-table in define-generics

rohinmshah:
Added a test so that the issue does not happen again.


stamourv:
Merged, thanks!

PR #1406, opened 2016-07-31
optimizer: don't move APPN_FLAG_OMITTABLE inside lambdas

gus-massa:
Some expressions are omittable only when the arguments have certain types. In this case the application is marked with `APPN_FLAG_OMITTABLE` instead of relaying on the flags of the primitive.

The optimizer can't use this flag to move the expression inside a `lamba `or across a potential continuation capture, unlike other omittable expressions. They can be moved only in more restricted conditions.

For example, in this program

    #lang racket/base
    
    (define n 10000)
    (define m 10000)
    (time
     (define xs (build-list n (lambda (x) 0)))
     (length xs)
     (define ws (list->vector xs)) ; <-- omittable
     (for ([i (in-range m)])
       (vector-ref ws 0)))     ; <-- ws is used once

If the optimizer moves the expression in the definition of `ws` inside the recursive
`lambda `that is created by the `for`, then the code is equivalent to:

    #lang racket/base
  
    (define n 10000)
    (define m 10000)
    (time
     (define xs (build-list n (lambda (x) 0)))
     (length xs)
     (for ([i (in-range m)])
       (vector-ref (list->vector xs) 0)))     ; <-- moved here

And the new code is `O(n*m)` instead of `O(n+m)`. This example is a minimized version
of the function `kde` from the `plot` pakage, where `n=m` and the bug changed the run time
from linear to quadratic.

Additional tests: https://travis-ci.org/gus-massa/racket/builds/148608283

This fix the error deteted by DrDr and reported by @samth in https://github.com/racket/racket/pull/1379#commitcomment-18388854

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -2111,6 +2111,18 @@
            '(lambda (z) (lambda ()
                           (lambda () z)))
            #f)
+;; Don't move omittable expressions that keep a reference:
+(test-comp '(lambda (z) (let ([r (pair? z)])
+                          (lambda () r)))
+           '(lambda (z) (lambda ()
+                          (lambda () (pair? z))))
+           #f)
+(test-comp '(lambda (z) (when (list? z)
+                          (let ([r (list->vector z)])
+                            (lambda () r))))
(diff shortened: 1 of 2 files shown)


mflatt:
LGTM

PR #1409, opened 2016-08-02
update #lang tutorial to use reader submodule

mbutterick:
Rather than the older `lang/reader.rkt` fandango.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/hash-languages.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/hash-languages.scrbl
@@ -66,8 +66,11 @@ Unlike @racketmodname[racket], @racketmodname[s-exp] cannot be used as a
 module path with @racket[require]. Although the syntax of
 @racket[_language] for @hash-lang[] overlaps with the syntax of module
 paths, a @racket[_language] is not used directly as a module
-path. Instead, a @racket[_language] is suffixed with
-@racketidfont{/lang/reader} to obtain a module path, and the resulting
+path. Instead, a @racket[_language] obtains a module path by trying two 
+locations:  first, it looks for a @racketidfont{reader} submodule of the 
+main module for  @racket[_language]. If this is not a valid module path, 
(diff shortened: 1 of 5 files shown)


mbutterick:
I would also recommend improving (though I will not attempt it myself) the error message that arises when you try to use a `#lang foobar` and no reader is found:

```text
standard-module-name-resolver: collection not found
  for module path: racket/foobar/lang/reader
  collection: "racket/foobar/lang"
  in collection directories:
   /Users/MB/Library/Racket/development/collects
   /Users/MB/git/racket/racket/collects
   ... [207 additional linked and package directories]
  no packages suggestions are available .
```

I understand that Racket looks in two locations for a reader — `(submod foobar/main reader)`, and then `foobar/lang/reader` — so this error message is not wrong. 

But here's the problem. Let's suppose you're trying to make a `reader` submodule for your language and you goof it up somehow, e.g., you forget to `provide` your `read-syntax`. This is the error message you get, which suggests that your problem has something to do with a file you haven't made (and indeed, the `lang` subdirectory is now something of an obsolete technique). 

In my experience, you see this error a lot when you're starting out with languages in Racket, and it is confusing. The failure to find a reader would ideally have its own special message, not just a generic error from `standard-module-name-resolver`, for instance:

```text
#lang: no read-syntax function provided
  for language: foobar
  by either (submod foobar/main reader)
  or foobar/lang/reader.rkt
```

mbutterick:
(per discussion with @mflatt and @LeifAndersen)

LeifAndersen:
Mostly looks good to me. Although do you think it would be a bad idea to keep some of the information about the legacy way to do it in there? (Maybe with a note saying that is legacy?)

Otherwise it looks great to me.

mbutterick:
> bad idea to keep some of the information about the legacy way to do it in there?

Fine with me, I just didn’t know whether the `foobar/lang/reader` approach is now considered alternative, obsolete, or deprecated. AFAICT it’s something that was necessary only because submodules hadn’t been invented.

Issue #1414, opened 2016-08-14
Socket file-descriptor leak when using `http-sendrecv` with method `HEAD`


tonyg:
To reproduce:

At a repl,

```racket
(require net/http-client)
(http-sendrecv "google.com" "http://google.com/" #:method #"HEAD")
```

Check `/proc/XXXXX/fd`. Notice there is a socket fd still open. Repeat the `http-sendrecv` command a ferw times. More dangling fds will result.

Contrast with what happens when you remove the `#:method #"HEAD"` from the `http-sendrecv` call: no sockets are leaked in that case.

PR #1415, opened 2016-08-14
Explicitly close the http-conn on http-sendrecv with method HEAD.

tonyg:
Fix for #1414.


jeapostrophe:
I looked for this one but failed to find it. Good job!

PR #1423, opened 2016-08-16
Record disappeared uses from syntax/parse/experimental/template

lexi-lambda:
This adds disappeared use tracking to `syntax/parse/experimental/template`, which marks disappeared uses of pattern variables (like `syntax` already does) and template metafunctions.


rmculpepper:
Looks good, but could you de-indent the body of with-disappeared-uses to minimize the "changed" lines? Or introduce a helper function (`do-template*`?), if misindented code would bother you.

lexi-lambda:
@rmculpepper Yep, that makes sense; I’ve de-indented the body.

PR #1425, opened 2016-08-19
JIT: Allow the jitter to reverse the arguments of string=?

gus-massa:
Also, make a few tweaks to the implementation of string=?/bytes=? in the JIT.

A posible alternative is that when one of the arguments is an explicit string, put it always as the second argument to simplify the code in jitinline.c (using if necesary the reversed version to raise the right errors).


mflatt:
LGTM

PR #1429, opened 2016-08-23
Support an empty (values) range in ->i

btlachance:
Without the `positive?` check, an empty vector gets through to a usage of `in-vector` that ends up raising a range error.

(Beyond having an explicit name, is there a big difference between `test/no-errors` and `test/spec-passed`?)


rfindler:
Looks like `test/no-error` is strictly worse than `test/spec-passed` to me.

rfindler:
Thanks: 

https://github.com/racket/racket/commit/07718022b4815b774b8c04a5ccd6e747ee4a25c1

https://github.com/racket/racket/commit/2e5b4a65150afc9517d5e344aaa6abaa63751e91

PR #1430, opened 2016-08-24
add location info to the match error msg

andmkent:
Currently, running the following program "foo.rkt" from the terminal

```racket
#lang racket

(match-let ([(cons a b) 42])
  (+ a b))
```

produces the following error message:

```
match-let-values: no matching clause for 42
  context...:
   /Users/amk/Repos/plt/6.6/collects/racket/match/runtime.rkt:24:0: match:error
   /Users/amk/Desktop/foo.rkt: [running body]
```

With these suggested changes, you instead get this error:

```
% racket foo.rkt      
foo.rkt:3:0 match-let-values: no matching clause for 42
  context...:
   /Users/amk/Repos/plt/match-error/racket/collects/racket/match/runtime.rkt:24:0: match:error
   /Users/amk/Desktop/foo.rkt: [running body]
```

In this particular example this isn't that big of a deal -- but in large code bases not specifying which  ```match``` threw the error is very painful.

### DIFF
--- a/racket/collects/racket/match/runtime.rkt
+++ b/racket/collects/racket/match/runtime.rkt
@@ -22,10 +22,13 @@
 
 
 (define (match:error val srclocs form-name)
-  (raise (make-exn:misc:match (format "~a: no matching clause for ~e" form-name val)
-                              (current-continuation-marks)
-                              val
-                              srclocs)))
+  (raise (make-exn:misc:match
+          (format "~a: no matching clause for ~e\n  location: ~a"
+                  form-name val
+                  (srcloc->string (car srclocs)))
+          (current-continuation-marks)
+          val
+          srclocs)))
 
 (define-syntax-parameter fail
   (lambda (stx)


samth:
Shouldn't we be able to have this in the stack trace, rather than in the message?

rfindler:
Why does this say `match-let-values` when there is no `match-let-values` in
the program?

andmkent:
@rfindler it seems match-let is defined in terms of match-let-values in a way that causes this to happen

rfindler:
You make it sound like this isn't software or something. You know we can
change programs to make them do what we want, right?   :)

In this case, one might have both match-let-value and match-let expand into
a third form that accepts a "name to use in error messages" argument or
something like that.

andmkent:
@rfindler I couldn't agree more. Are you suggesting that I add that fix to this PR? Or that I submit another one to fix that issue next? I'm not sure what your suggesting with your comment TBH =)

rfindler:
I had interpreted this PR as coming from a desire to fix the error message associated with match-let and was pointing out another way that it was wrong. Your time is your own, as far as I'm concerned and I'm grateful for whatever contributions you choose to make to Racket. Sorry if I stepped over the line.

andmkent:
@rfindler I used match-let here, but in practice the (avoidable) lack of location information is a frustration with _every_ match form (they all use the same internal code eventually). I agree with your suggestion that it's dumb for it to report an error for ```match-let-values``` in this case, but I'm still developing a feel for how narrow/broad PRs to ```racket``` should be and wasn't sure what would be a better approach for the issue you mentioned. It's a trivial fix I'm happy to do, I just wanted to clarify if you thought it fit in this PR or should appear in a different one.

andmkent:
@samth pointed out that ```class``` has a similar issue to deal with. I'm changing the match message to look more like theirs.

andmkent:
now the error message looks more like what ```class``` does in similar circumstances (where a dynamic error can provide precise source location information):

```racket
% racket foo.rkt
match-let-values: no matching clause for 42
  location: foo.rkt:3:0
  context...:
   /Users/amk/Repos/plt/match-error/racket/collects/racket/match/runtime.rkt:24:0: match:error
   /Users/amk/Desktop/foo.rkt: [running body]
```

I'll address the match-let-values issue in another PR.

rfindler:
That's a surprisingly easy change! What are the other srclocs in that list?

andmkent:
that's the only entry in the list -- match:error is given the result of syntax-srclocs

which is defined here:
https://github.com/racket/racket/blob/master/racket/collects/racket/match/runtime.rkt#L64

and called here: 
https://github.com/racket/racket/blob/master/racket/collects/racket/match/gen-match.rkt#L43

endobson:
I think this is the wrong approach, but I'm disagreeing with with the racket documentation here. (https://docs.racket-lang.org/reference/exns.html?q=exn#%28part._err-msg-conventions%29)

There already is [prop:exn:srclocs](https://docs.racket-lang.org/reference/exns.html?q=exn#%28def._%28%28quote._~23~25kernel%29._prop~3aexn~3asrclocs%29%29) which exceptions must set to get good integration with DrRacket (link to the problematic code). The issue is that the error display handler in the command line program doesn't display it.

Previous discussion:
http://lists.racket-lang.org/dev/archive/2014-March/013988.html

I currently have a custom error handler which deals with this nicely enough for me and never got the time to improve it and push it upstream.
```
(let ((old-error-display-handler (error-display-handler)))
    (error-display-handler
      (λ (str exn)
        (when (exn:srclocs? exn)
          (for ((srcloc ((exn:srclocs-accessor exn) exn)))
            (displayln (source-location->string srcloc))))
        (old-error-display-handler str exn))))
```

andmkent:
The solution @endobson suggests does seem nicer.

In the meantime, though, this change would be extremely useful and is in line with what ```class``` does today.

PR #1432, opened 2016-08-25
report locations in default error handler

pnwamk:
Based on comments from @endobson in PR #1430 I threw together an initial, naive addition to the default error handler that displays source location info when available.

I'm not sure how close this is to whatever would be ideal, but perhaps this can generate some thoughts/comments/etc even if we scrap this initial attempt? =)

This initial change has the following effects on this program (ignoring the changes from #1430, which cause the location to be printed twice at the moment with this change).

Program:
```racket
#lang racket

(match 42
  [(cons a b) "success"])
````

Racket 6.6 command line:
```
% racket foo.rkt
match: no matching clause for 42
  context...:
   /Users/amk/Repos/plt/racket/racket/collects/racket/match/runtime.rkt:24:0: match:error
   /Users/amk/Desktop/foo.rkt: [running body]
```

Racket 6.6 w/ these changes, command line:
```
% racket foo.rkt
match: no matching clause for 42
  location:
   foo.rkt:3:0
  context...:
   /Users/amk/Repos/plt/racket/racket/collects/racket/match/runtime.rkt:24:0: match:error
   /Users/amk/Desktop/foo.rkt: [running body]
```

(Before I had mistakingly suggested DrRacket's printing was altered, but that is not the case fortunately)

### DIFF
--- a/racket/collects/racket/match/runtime.rkt
+++ b/racket/collects/racket/match/runtime.rkt
@@ -23,9 +23,8 @@
 
 (define (match:error val srclocs form-name)
   (raise (make-exn:misc:match
-          (format "~a: no matching clause for ~e\n  location: ~a"
-                  form-name val
-                  (srcloc->string (car srclocs)))
+          (format "~a: no matching clause for ~e"
+                  form-name val)
           (current-continuation-marks)
           val
           srclocs)))
--- a/racket/src/racket/src/error.c
+++ b/racket/src/racket/src/error.c
@@ -3158,8 +3158,8 @@ def_error_display_proc(int argc, Scheme_Object *argv[])
   s = scheme_char_string_to_byte_string(argv[0]);
 
   scheme_write_byte_string(SCHEME_BYTE_STR_VAL(s),
-			   SCHEME_BYTE_STRTAG_VAL(s),
-			   port);
+                           SCHEME_BYTE_STRTAG_VAL(s),
+                           port);
 
   /* Print context, if available */
   if (SCHEME_CHAPERONE_STRUCTP(argv[1])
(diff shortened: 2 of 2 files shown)

### CODE REVIEW

**Comment by andmkent** [comment]
For reference I think this is the last discussion on this issue: http://lists.racket-lang.org/dev/archive/2014-March/013988.html

**Comment by andmkent** [comment]
@samth I was wrong about it changing DrRacket printing -- that was the change from #1430 (doh!). DrRacket printing is unaffected by this change (at least as far as I can tell).

**Comment by andmkent** [comment]
To do list pieced together from previous [mailing list discussion](http://lists.racket-lang.org/dev/archive/2014-March/013988.html):

- [ ] DrRacket prints correctly (i.e. is unaffected by this change) when
debugging is on.

- [ ] Address how DrRacket should print errors when debugging/profiling
is off.


- [ ] Are there other program whose harness is worse off with the default display
error handler starts showing source locations?

- [ ] Investigate whether the handler should not print the first returned
source location for `exn:fail:read`, `exn:fail:syntax`, and
`exn:fail:contract:variable` exceptions ()because that location is
supposed to be built into the error-message text, unless the
`error-print-source-location` parameter has a false value).

**Comment by rfindler** [comment]
Also: it would be nice to find a way to get multiple source locations
printed when there are multiple to show.

Robby

**Comment by andmkent** [comment]
Should it print ```location...:``` instead of ```location:``` like the context? (to account for the possibility of there being more than 1?)

**Comment by rfindler** [comment]
Oh, yes! The ellipsis is a good idea. That'll help smooth the DrRacket integration.

The ellipsis doesn't mean "many", btw. Here's some info about it and other related things that you might want to check out:

  http://docs.racket-lang.org/reference/exns.html#%28part._err-msg-conventions%29

**Comment by andmkent** [comment]
FYI/update: at the moment, things like ```raise-syntax-error``` end up getting some location info double printed =(

```
test.rkt:4:2: cond: bad syntax (`else' clause must be last)
  at: (else 1)
  in: (cond (else 1) (else 2))
  location...:
   test.rkt:4:2
  context...:
   /Users/amk/Repos/plt/racket/racket/collects/racket/private/cond.rkt:31:18: loop
   /Users/amk/Repos/plt/racket/racket/collects/racket/private/cond.rkt:21:13: go
   /Users/amk/Repos/plt/racket/racket/collects/syntax/wrap-modbeg.rkt:46:4
   standard-module-name-resolver
```

**Comment by andmkent** [comment]
Okay, so for the exceptions affected by [error-print-source-location](http://docs.racket-lang.org/reference/exns.html?q=error-print-source-location#%28def._%28%28quote._~23~25kernel%29._error-print-source-location%29%29) there is special casing to not print the source location they are already printing in their error message.

This means the above error message (for cond) now prints like this:

```
test.rkt:4:2: cond: bad syntax (`else' clause must be last)
  at: (else 1)
  in: (cond (else 1) (else 2))
  context...:
   /Users/amk/Repos/plt/racket/racket/collects/racket/private/cond.rkt:31:18: loop
   /Users/amk/Repos/plt/racket/racket/collects/racket/private/cond.rkt:21:13: go
   /Users/amk/Repos/plt/racket/racket/collects/syntax/wrap-modbeg.rkt:46:4
   standard-module-name-resolver
```

Happy to make tweaks and incorporate further suggestions -- if others are happy with it I think it might be good to ship.

andmkent:
@mflatt is this similar to what you had in mind (from your previous [comments](http://lists.racket-lang.org/dev/archive/2014-March/013992.html) on the topic)?

pnwamk:
ping

mflatt:
This all seems plausible, so I'm fine with merging to try it out.

pnwamk:
Commits are now squashed into 1. (I don't have push access to this repo - happy to tweak things further if needed)

stamourv:
Merged, thanks!

PR #1433, opened 2016-08-25
update lookup-errno with more error codes, relax contract

rmculpepper:
Recognize all of the symbols from POSIX.1 (if defined by errno.h).

lookup-errno now returns #f when given an unknown symbol instead of raising a contract error. It should not return #f for any symbol that it previously accepted.


mflatt:
This looks ok to me.

Do we know whether MSVC provides all of these, too, so that the functionality will be the same with both MSVC and MinGW?

rmculpepper:
VS 2010 and later define the following codes that MinGW does not:

  - `ENODATA, ENOSR, ENOSTR, ETIME`: marked as optional in POSIX.1
  - `ENOLINK`: described as "reserved" in POSIX.1
  - `ENOTRECOVERABLE`: not in the 2004 version of POSIX.1
  - `EBADMSG, EIDRM, ENOMSG, ETXTBSY`: no explanation I found

VS pre-2010 defined a much smaller list.

PR #1436, opened 2016-08-25
struct is preferred

spdegabrielle:
changed 'make-struct' to struct 
struct is preferred form according to documentation: https://docs.racket-lang.org/reference/define-struct.html


stamourv:
Merged, thanks!

PR #1438, opened 2016-08-25
Allow dotted list syntax for match expanders 

jsmaniac:
The following code could not be compiled previously, as the pattern handling match-expanders was `(expander args ...)`, and so rejected the form `(some-match-expander a b c . d)`. I changed it to `(expander . args)` and added a test.

``` racket
(match v [(some-match-expander a b c . d) (displayln (list a b c d))])
```

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -646,6 +646,30 @@ expander and @racket[len] always returns @racket[0].
   (len nil)
   (len (cons 1 nil))
   (len (cons 1 (cons 2 nil)))]
+
+Match expanders accept any syntax pair whose first element is an
+@racket[identifier?] bound to the expander. The following example
+shows a match expander which can be called with an improper syntax
+list of the form @racket[(expander a b . rest)].
+@examples[#:label #f
+  #:eval match-eval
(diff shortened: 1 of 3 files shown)


samth:
Does the documentation need to change for this?

samth:
Did you mean to push the changes to the `syntax-parse` docs?

jsmaniac:
I pushed an update to the docs, and included a small fix in the syntax/parse docs.

I noted the change using  `@history[#:changed "6.6.0.5" "description"]`, since the current version is 6.6.0.4. Is that okay, or should I indicate the change for 6.6.0.4?

samth:
Usually we put the current version (unless the commit changes the version as well).

jsmaniac:
Updated :)

jsmaniac:
I bumped the version number in the docs to 6.6.0.5 (the new "current" version of Racket), and rebased the commits on top of the latest HEAD.

pmatos:
Amazingly this PR is clean without conflicts with base branch, even though it's from 2016. @samth  you were involved in this discussion, can we merge or shall we close? What's the status here?

sorawee:
Note: the `#:changed "6.9.0.2"` needs to get bumped.

samth:
Now merged.

Issue #1443, opened 2016-08-30
Is `read-cdot` supposed to be left- or right- associative?


AlexKnauth:
@jeapostrophe 

With `read-cdot` turned on, should `a.b.c` read as `(#%dot (#%dot a b) c)` or `(#%dot a (#%dot b c))`? I would have expected `(#%dot (#%dot a b) c)`, because that's how field references normally behave. 

In a language where parens mean grouping, `a.b.c` would be `(a.b).c`.

However with the current implementation, it's the other one.

bennn:
+1 for `(a.b).c`

jeapostrophe:
The interpretation is on purpose. You can see that in the default implementation of `#%dot` --- https://github.com/jeapostrophe/remix --- the `a` binding is a "dot-transformer" that looks over the rest of the binding and interprets it. In fact, the default binding flattens any sequences of dots into `(#%dot a b c)` so that `a` can see the whole context and expand appropriately.

AlexKnauth:
So this is okay because remix flattens it? But what about other languages that would want to use this?

For example languages that define `#%dot` to transform `a.f` into `(f a)` or something would want `a.string->number.factorial.write` to transform into `(write (factorial (string->number a)))`. Other languages that define `#%dot` to transform `a.x` into `(hash-ref a 'x)` would want `a.x.y.z` to transform into `(hash-ref (hash-ref (hash-ref a 'x) 'y) 'z)`. 

And `remix`'s `#%dot` form would just need a rule where `(#%dot (#%dot a x) y)` would recognize that `(#%dot a x)` isn't a dot-transformer, so it would flatten it to `(#%dot a x y)`, which it would have ended up doing anyway.

bennn:
If it is on purpose, then it should be documented. C-style is left-to-right associativity.
(e.g. Table 2-1 on p51 of [K&R](https://hassanolity.files.wordpress.com/2013/11/the_c_programming_language_2.pdf))

jeapostrophe:
It just so happens to flatten it for struct accessors. Other languages can expand it however they want.

It is documented.

http://docs.racket-lang.org/reference/reader.html?q=cdot#%28part._parse-cdot%29

AlexKnauth:
That doesn't specify whether it's left- or right- associative, it only gives a low-level process that may or may not produce something that makes sense. *Other* language implementors will be expecting left-associativity, and that low-level explanation won't undo that expectation for them.

jeapostrophe:
[...]

PR #1446, opened 2016-09-01
make cdot reader left associative

AlexKnauth:
So that with the `read-cdot` option turned on,
```racket
X.Y.Z
```
Reads equal to 
```racket
(#%dot (#%dot X Y) Z)
```
This is what most language implementers will expect when they are writing their `#%dot` macros.

Resolves https://github.com/racket/racket/issues/1443.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/reader.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/reader.scrbl
@@ -955,13 +955,19 @@ equivalent prefix as discussed in @secref["parse-number"]. If these
 numbers are followed by a @litchar{.} intended to be read as a C-style
 infix dot, then there must be separating whitespace.
 
-Finally, after reading any value, @racket[_x], the reader will seek
-over whitespace until it reaches a non-whitespace character. If the
-character is not @litchar{.}, then the value, @racket[_x], is returned
-as usual. If the character is @litchar{.}, then another value,
-@racket[_y], is read and the result @racket[(list '#%dot _x _y)] is
-returned. In @racket[read-syntax] mode, the @racket['#%dot] symbol has
-the source location information of the @litchar{.} character and the
+Finally, after reading any datum @racket[_x], the reader will seek
+through whitespace and look for zero or more sequences of a
+@litchar{.} followed by another datum @racket[_y]. It will then group
+@racket[_x] and @racket[_y] together in a @racket[#%dot] form so that
+@racket[_x.y] reads equal to @racket[(#%dot _x _y)].
+
+If @racket[_x.y] has another @litchar{.} after it, the reader will
+accumulate more @litchar{.}-separated datums, grouping them from
+left-to-right. For example, @racket[_x.y.z] reads equal to
+@racket[(#%dot (#%dot _x _y) _z)].
+
+In @racket[read-syntax] mode, the @racket[#%dot] symbol has the
+source location information of the @litchar{.} character and the
 entire list has the source location information spanning from the
 start of @racket[_x] to the end of @racket[_y].
 
--- a/pkgs/racket-test-core/tests/racket/read.rktl
+++ b/pkgs/racket-test-core/tests/racket/read.rktl
@@ -1242,6 +1242,27 @@
   (test 17t0 readstr "#x11.0t0"))
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; read-cdot
+
+(parameterize ([read-cdot #true])
+  (test '(#%dot a b)                     readstr "a.b")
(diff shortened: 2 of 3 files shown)


LeifAndersen:
Yes, very much this.

pkgs/racket-doc/scribblings/reference/reader.scrbl
```diff
@@ -956,12 +956,16 @@ numbers are followed by a @litchar{.} intended to be read as a C-style
...
 
 Finally, after reading any value, @racket[_x], the reader will seek
-over whitespace until it reaches a non-whitespace character. If the
-character is not @litchar{.}, then the value, @racket[_x], is returned
-as usual. If the character is @litchar{.}, then another value,
-@racket[_y], is read and the result @racket[(list '#%dot _x _y)] is
-returned. In @racket[read-syntax] mode, the @racket['#%dot] symbol has
-the source location information of the @litchar{.} character and the
+through whitespace and look for zero or more sequences of a
+@litchar{.} followed by another value @racket[_y], to put each
```
bennn: While we're here, I think "value" should be "datum", re Jay's comment on the issue.

pkgs/racket-doc/scribblings/reference/reader.scrbl
```diff
@@ -956,12 +956,16 @@ numbers are followed by a @litchar{.} intended to be read as a C-style
...
+@racket[_y] together with @racket[_x] in a @racket[#%dot] form.
+@racket[_x.y] will be read as @racket[(#%dot _x _y)].
+
+It does this in a left-associative way so that @racket[_x.y.z]
+will be read as @racket[(#%dot (#%dot _x _y) _z)].
```
bennn: Suggestion: replace this sentence with "If `y` contains a `.`, the reader will group `.`-separated datums from left-to-right. For example, `x.y.z` reads as `(list (list #%dot x y) z)`.

pkgs/racket-doc/scribblings/reference/reader.scrbl
AlexKnauth: Yes, definitely.

pkgs/racket-doc/scribblings/reference/reader.scrbl
jeapostrophe: +1

jeapostrophe:
You should update `read-cdot`'s existing clients because this is a backwards-incompatible change.

pkgs/racket-doc/scribblings/reference/reader.scrbl
AlexKnauth: It's not really that `y` contains a `.`, it's that there's another `.` after `y`. So would "If `y` has another `.` after it, the reader will group `.`-separated datums from left-to-right" be better?

jeapostrophe:
Once you are happy with the docs, it is fine to launch by me

LeifAndersen:
Hey @jeapostrophe , since we're breaking backwards compatibility, do you have any opinions on changing the name to something other than cdot? As cdot doesn't make sense as a name unless you know much about c-style languages. Also, cdot means something completely different in say latex, which might confuse things.

bennn:
@LeifAndersen how about `read-#%dot` ?

AlexKnauth:
That can be a separate change. I would rather do one thing at a time

bennn:
@AlexKnauth ok. Are you planning to update the existing clients in this PR?

AlexKnauth:
@bennn Yes, there's a pull request ready here: https://github.com/jeapostrophe/remix/pull/1

LeifAndersen:
@bennn I'm fairly ambivalent on read-#%dot.  I'm not aware of anything else in our standard library that follows a convention like that though.

LeifAndersen:
Some suggestions (with help from Ben) might be: `read-dot-notation`, also `read-dot-operator` would be nice too, although I guess that very much goes against Jay saying dot is not an operator, so maybe not the second one.

AlexKnauth:
Since dots already mean 3 different things in racket without this, those would all be confusing. The name `cdot` at least differentiates it from the more lisp-based uses of dots.

LeifAndersen:
Ya, that's fair.

jeapostrophe:
I don't think we should change the name and I don't think we should stress about the backwards-incompat given that there is one use and no one uses it. I'm good to commit.

AlexKnauth:
Okay. Can you look over the doc changes and see if there's anything else that should be changed?

LeifAndersen:
Ya, I'm inclined to agree with @jeapostrophe here. As sad as it is, I can't think of anything better then cdot, oh well. :( We should probably not change the name, as its not really that important.

jeapostrophe:
I think everything is good and I can merge it.

PR #1449, opened 2016-09-03
fix procedure-rename to not convert procedures into methods or methods into normal procedures

AlexKnauth:
So that programs like
```racket
#lang racket
(define (f a b c d e #:x [x 5]) a)
(define f* (procedure-rename f 'f))
(f*)
```
Report
```
. f: arity mismatch;
  expected: 5 plus an optional argument with keyword #:x
  given: 0
```
Instead of
```
. f: arity mismatch;
  expected: 4 plus an optional argument with keyword #:x
  given: 0
```

Fixes https://github.com/AlexKnauth/kw-utils/issues/3

### DIFF
--- a/pkgs/racket-test-core/tests/racket/procs.rktl
+++ b/pkgs/racket-test-core/tests/racket/procs.rktl
@@ -363,8 +363,8 @@
 
 ;; ----------------------------------------
 ;; Check error for non-procedures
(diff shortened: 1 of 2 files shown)


rfindler:
LGTM

AlexKnauth:
(I don't have push access)

rfindler:
I've pushed the commit. Thanks!

stamourv:
@AlexKnauth: Re commit access: that should be fixed now.

PR #1450, opened 2016-09-05
Make `sql-null?` a struct predicate.

samth:
This allows Typed Racket to know that it's pure and safe to use as an opaque
value.


samth:
@rmculpepper Is this ok with you?

PR #1457, opened 2016-09-09
optimizar: add strict-true? and boolean? as relevant predicates

gus-massa:
The idea is to add a new primitive that is equivalent to

    (define (strict-true? x)
      (eq? x #t))

And add this new primitive and `boolean?` as relevant predicates for the optimizer.

**optimizer: fix reduction to handle boolean? as a relevant predicate**

This fix two reductions that caused no problems when the only predicate that recognizes `#f` is `not`.

**add hidden strict-true? primitive**

Is this the right name for the primitive? It's hidden from the Racket code to discourage it's direct use.

This include the implementation for the JIT and some optimizations.

**Use compiled startup again (squash)**

I'll squash this before merging.

**add boolean? to the list of relevant predicates**

It's more difficult to add `boolean?` as a relevant predicate due to the intersections with `not `and `strict-true?`. It was necesary to fix two problems with previous reductions that were fine until now (see first commit).

I added many usual predicates to a list to mark the result of them as `boolean?`. The list is copied from optimiza.rktl.

I put code to recognize them by name with `IS_NAMED_PRIMITIVE`, but many of them can be recognized comparing them with a global variable using `SAME_OBJ`. See  [optimize.c#L2947](https://github.com/gus-massa/racket/blob/2ba24c37a812dfc33d6bd97259f958c65461409f/racket/src/racket/src/optimize.c#L2947) I like more the seccond method with  `SAME_OBJ`, but I think that the "by-name" method will make it easier to port this code to Racket someday. I can change them if it's better to use the other method.
 
**optimizer: add reduction for the expansion of `or` with a boolean argument**

This removes the intermediate variable that creates the expansion of `or` with a `boolean?` as first argument, so an expression like 

    (or (symbol? x) (something))

is reduced to 

    (if (symbol? x) #t (something))

The check is done before optimizing the argument, so this may miss some cases where it's obvious that the argument is a boolean after optimizing it. Anyway, the idiom is very common, so I think that this special case is useful.

I tried to extend this to the predicates that recognize structs and struct properties, but I failed. It would be a nice future adition.


gus-massa:
In the new commit, I marked the struct and struct property predicates as `boolean?`s. I'm still not sure about the details, in particular if it's necesary to check that `SCHEME_PROC_SHAPE_IDENTITY(shape)` is a  pair in [optimize.c#L2998](https://github.com/gus-massa/racket/blob/39ea2dce5cae4ed1d31a3d2065d9e939fddbaeeb/racket/src/racket/src/optimize.c#L2998).

(I'll squash this with "add boolean? to the list of relevant predicates".)

gus-massa:
I still can’t believe that I forgot to add `eq?` to the list of primitives that return a `boolean?`. 

Currently one of the tests is failing because the optimizer reduces:

    (boolean? (equal? x y))  ==>  (begin (values (equal? x y)) #t)

I think that the `values` is unnecessary, but perhaps I’m missing a corner case.

The easiest solution is to add a `values` in the test to make it pass.

A more difficult solution is to add a special case for `equal?` in `ensure_single_value`.

An even more difficult solution is to fix `ensure_single_value` because it calls `single_valued_noncm_expression` so it checks unnecessarily that it’s `non_cm`, but in this case it will be necessary to check all the use places of `ensure_single_value` to be sure that they only care about the single value and they don’t expect the `non_cm` check too.

mflatt:
[Sorry for the very slow response!]

I think `equal?` should be marked as non-cm, and that should solve the problem with the `values` wrapper. (Probably I just didn't think hard enough about it before, other than observing that `equal?` can't be folding like `eq?` and `eqv?`.)

The rest looks ok to me. I slightly prefer the name `true-object?` (following `eof-object?`) instead of `strict-true?`, but I'm happy either way.

gus-massa:
I rebased al the commits, and renamed `strict-true?` to `true-object?`.

I also added two small tweaks. One is an extension of the reduction for the expanded version of `or` and the other is a similar modification for the expanded version of `and`. (I think it’s more natural to squash and merge them now, than waiting to add the later.)

mflatt:
That all sounds good to me.

gus-massa:
I've pushed the squashed version: https://github.com/racket/racket/commit/f159295e55b2e480eaba453dbd405cd7185137d5

PR #1458, opened 2016-09-09
Add a #%namespaced form to racket/base for use with language extensions

lexi-lambda:
This is a set of changes designed to make implementation of language extensions, aka “meta-languages”, easier. It is related to an email conversation I had with @mflatt and @AlexKnauth a couple months ago.

Currently, simple language extensions like `s-exp` and `at-exp` work well with the Racket `#lang` model, but more complex ones like Alex’s `afl`, `debug`, and `hygienic-quote`, my own `curly-fn`, and DrRacket’s graphical syntax rely on a set of hacks to work properly. Here’s a summary of the problem:

  1. Historically, `read` and `read-syntax` are supposed to have identical behavior, the only difference is that `read-syntax` includes source location information. Importantly, syntax objects produced by `read-syntax` are *not* supposed to include lexical context.

  2. Implementation of new syntactic features is easy in most `#lang`s: the reader simply produces some kind of tagged s-expression, and the module language provides a macro bound to that tag. This allows most logic to remain in macro-expansion, instead of the reader, which is nice. It also permits adding arbitrary syntactic features without adding the dimension of lexical scopes to reader functions.

  3. Language extensions/meta-languages that want to introduce new syntactic features have a problem: they do not have any control of the lexical environment being provided by the module language. I will use `curly-fn` as an example, since it is my package and I am most familiar with it, but the below comments also apply to many of Alex’s packages and “graphical syntax”, and I’m sure it has other applications as well.

Currently, `curly-fn` and similar packages use a hacky workaround that involves using `make-syntax-introducer` to apply a scope to the entire module, then selectively removing it for identifiers that should be hygienically introduced by the language extension. This is to avoid ambiguous binding errors caused by inappropriately returning syntax objects with lexical context from `read-syntax`. This solution sort of works, but it is not without significant problems:

  1. Doing this completely breaks `read`. Given that a proper interpretation of a module now *requires* additional lexical context, this introduces an extra dimension to module reading that currently didn’t exist.

  2. Languages implemented this way are difficult to compose or reuse. Even though `curly-fn` is implemented via readtable, it’s impossible to use the `curly-fn` readtable in another `#lang`, because the scope introduced by `make-syntax-introducer` won’t exist in the other language.

  3. For similar reasons to the above two points, these kinds of extensions don’t really work in the REPL, at least not without a lot of extra work to create a separate set of hacks for handling top-level interactions.

  4. It isn’t well-supported by the compiler, so it leads to [things like this][1] to get namespacing to work properly. This is pretty brittle, so it’s definitely a bad pattern to see.

There are two obvious ways to fix this problem. The first would be to enhance the compiler and macro expander to properly support syntax objects with lexical produced by `read-syntax` and do away with `read` entirely. This would involve a lot of work, and it’s not entirely clear what it would take. I don’t actually personally think that obsoleting `read` for reading languages would be much of a loss, but it would definitely be a break from things.

The alternative approach is to provide a hook in base languages that language extensions can use to safely introduce hygienic syntactic extensions. This pull request implements such a hook, in the form of a new `#%namespaced` syntax transformer that is provided by `#lang racket/base`. This is a form of simple, explicit identifier namespacing implemented as an entirely derived concept within the existing Racket macro system—the only reason it needs to go in `base` is so that `racket/base` can provide it.

There’s no doubt that `#%namespaced` is an easy way out of a larger problem: it effectively reintroduces a limited solution that other languages use as a “hygiene substitute”. However, here it is only used as a hook for language extensions to offload as much work as possible to the hygienic macro expander, so it’s not nearly as insidious as it might first appear. It also solves a practical problem that various people seem to have run into, myself included, while the more general approach would involve an unknown effort that nobody really has the time to research and implement.

To see the compromise solution in action, I have updated my `curly-fn` package to demonstrate what a use of `#%namespaced` looks like in the implementation of a language extension. You can see the [relevant lines here][2], though the existing implementation is a little bit cluttered because it supports `#lang`s that don’t support `#%namespaced` using the older approach. However, for languages that *do* support `#%namespaced`, `curly-fn` can both be easily mixed in using its readtable extension *and* works in the REPL without any additional effort. I would imagine this sort of solution would also work with other packages as well as DrRacket’s graphical syntax.

My apologies for the long pull request description—I want to make the motivations as clear for this change as possible. Please let me know if you have any questions, suggestions, or concerns.

[1]: https://github.com/lexi-lambda/racket-curly-fn/blob/cef01874a9a5282d0a72e3acc2630c7605359c93/curly-fn-lib/curly-fn/lang/reader.rkt#L3-L9
[2]: https://github.com/lexi-lambda/racket-curly-fn/blob/cef01874a9a5282d0a72e3acc2630c7605359c93/curly-fn-lib/curly-fn/lang/reader.rkt#L59-L63

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/syntax.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax.scrbl
@@ -17,6 +17,8 @@
                      (only-in compiler/cm-accomplice
                               register-external-module)
                      racket/performance-hint
+                     (only-in syntax/module-reader
+                              make-meta-reader)
                      syntax/parse))
 
 @(define require-eval (make-base-eval))
@@ -1753,6 +1755,36 @@ expander introduces @racketidfont{#%app} identifiers.
 
 Like @racket[#%app], but without support for keyword arguments.
 As a special case, @racket[(#%plain-app)] produces @racket['()].}
+
+@;------------------------------------------------------------------------
+@section[#:tag "namespaced-transformer"]{Explicitly Namespaced Identifiers: @racket[#%namespaced]}
+
+The @racket[#%namespaced] form adds a form of explicit identifier namespacing, designed to be used in
+contexts where syntax objects must be created without lexical scope, such as when implementing
+@racket[read-syntax] for a language. This is especially useful for “language extensions”, also known
+as “meta languages”, such as languages created using @racket[make-meta-reader]. Since language
+extensions do not have control over the lexical environment that the produced syntax objects will be
+evaluated in, it is difficult to create hygienic extensions that adjust semantics. If the “host”
+language provides @racket[#%namespaced], then language extensions can use it to safely hook into
+the result by defining ordinary syntax transformers.
+
+For other use cases, @racket[#%namespaced] is usually the wrong choice. See @racket[local-require] for
+a more flexible, alternative way to scope imported identifiers without the need for explicit
+namespacing.
+
+@defform[(#%namespaced module-path form)
+         #:grammar ([form id (id . args)])]{
+Equivalent to @racket[form], except that @racket[id]’s lexical context is adjusted to refer to an
+identifier imported from @racket[module-path]. All source locations and syntax properties in
+@racket[form] are preserved, and if @racket[form] is a pair, then @racket[args] are passed through
+entirely unmodified.
+
+@mz-examples[
+(#%namespaced racket/math pi)
+(#%namespaced racket/list (take '(1 2 3) 2))
+(#%namespaced racket/match (match '(1 2 3)
+                             [(list _ x _) x]))
+]}
 
 @;------------------------------------------------------------------------
 @section[#:tag "lambda"]{Procedure Expressions: @racket[lambda] and @racket[case-lambda]}
--- a/pkgs/racket-test/tests/racket/namespaced-transformer.rkt
+++ b/pkgs/racket-test/tests/racket/namespaced-transformer.rkt
@@ -1,0 +1,14 @@
+#lang racket/base
+
+(require rackunit
+         syntax/macro-testing)
+
+(check-= (#%namespaced racket/math pi) 3.14159 0.00001)
+(check-equal? (#%namespaced racket/list (take '(a b c) 2)) '(a b))
+(check-equal? (#%namespaced racket/match (match '(a b c)
+                                           [(list _ x _) x]))
+              'b)
+
+(check-exn #rx"^quote: unbound identifier"
+           (λ () (convert-syntax-error
+                  (#%namespaced racket/list '(1 2 3)))))
--- a/racket/collects/racket/private/base.rkt
+++ b/racket/collects/racket/private/base.rkt
@@ -12,6 +12,7 @@
              "cert.rkt"
              "submodule.rkt"
              "generic-interfaces.rkt"
+             "namespaced-transformer.rkt"
              (for-syntax "stxcase-scheme.rkt"))
 
   (#%provide (all-from-except "pre-base.rkt"
@@ -38,6 +39,7 @@
              (all-from "cert.rkt")
              (all-from "submodule.rkt")
              (all-from "generic-interfaces.rkt")
+             (all-from "namespaced-transformer.rkt")
              (for-syntax syntax-rules syntax-id-rules ... _)
              (rename -open-input-file open-input-file)
              (rename -open-output-file open-output-file)
--- a/racket/collects/racket/private/namespaced-transformer.rkt
+++ b/racket/collects/racket/private/namespaced-transformer.rkt
@@ -1,0 +1,53 @@
+(module namespaced-transformer "pre-base.rkt"
+
+  (require (for-syntax "pre-base.rkt"
+                       "stx.rkt"
+                       "stxcase-scheme.rkt"))
+
+  (provide #%namespaced)
+
+  (define-for-syntax (strip-context e)
+    (cond
+      [(syntax? e)
+       (datum->syntax #f (strip-context (syntax-e e)) e e)]
+      [(pair? e) (cons (strip-context (car e))
+                       (strip-context (cdr e)))]
+      [(vector? e) (list->vector
+                    (map (lambda (e) (strip-context e))
+                         (vector->list e)))]
+      [(box? e) (box (strip-context (unbox e)))]
+      [(prefab-struct-key e)
+       => (lambda (k)
+            (apply make-prefab-struct k
+                   (strip-context (vector->list (cdr (struct->vector e))))))]
+      [else e]))
+
+  (define-syntax (#%namespaced stx)
+    (syntax-case stx ()
+      [(_ module-path form)
+       (unless (module-path? (syntax->datum #'module-path))
+         (raise-syntax-error #f "bad module path" stx #'module-path))
+       (cond
+         [(identifier? #'form)
+          (syntax-local-introduce
+           (syntax-local-lift-require
+            (syntax-local-introduce (strip-context #'module-path))
+            (syntax-local-introduce (strip-context #'form))))]
+         [(stx-pair? #'form)
+          (unless (identifier? (stx-car #'form))
+            (raise-syntax-error #f "namespaced form must start with identifier" stx (stx-car #'form)))
+          (datum->syntax #'form
+                    (cons (syntax-local-introduce
+                           (syntax-local-lift-require
+                            (syntax-local-introduce (strip-context #'module-path))
+                            (syntax-local-introduce (strip-context (stx-car #'form)))))
+                          (stx-cdr #'form))
+                    #'form #'form)]
+         [else
+          (raise-syntax-error #f "namespaced form must be pair or identifier" stx #'form)])]
+      [(_ _)
+       (raise-syntax-error #f "missing namespaced form" stx)]
+      [(_)
+       (raise-syntax-error #f "missing module path and namespaced form" stx)]
+      [(_ _ _ . _)
+       (raise-syntax-error #f "too many sub-forms" stx)])))


pkgs/racket-doc/scribblings/reference/syntax.scrbl
```diff
@@ -1755,6 +1757,36 @@ Like @racket[#%app], but without support for keyword arguments.
...
+
+For other use cases, @racket[#%namespaced] is usually the wrong choice. See @racket[local-require] for
+a more flexible, alternative way to scope imported identifiers without the need for explicit
+namespacing.
+
+@defform[(#%namespaced module-path form)
+         #:grammar ([form id (id . args)])]{
+Equivalent to @racket[form], except that @racket[id]’s lexical context is adjusted to refer to an
+identifier imported from @racket[module-path]. All source locations and syntax properties in
+@racket[form] are preserved, and if @racket[form] is a pair, then @racket[args] are passed through
```
jsmaniac: What happens to the source location and syntax properties of the pair itself? In other words, if `id` is a macro, what does `(syntax-source stx)` produce within the macro? If I read you code correctly, I think the source location and properties of the syntax pair (or syntax list) itself are preserved.

Is this what "All source locations and syntax properties in `@racket[form]` are preserved" implies?

lexi-lambda: Yes, that’s all correct. The implementation of `#%namespaced` is very careful to preserve *all* source locations and properties, even the ones “on the parentheses”, that is, the srclocs and properties of the pair itself.

mflatt:
@lexi-lambda After a first read (finally!), I don't get how `#%namespaced` differs from `local-require`. Can you say more about the difference, or maybe point me to a part of the description that I missed?

lexi-lambda:
@mflatt Sure. For `racket/base` specifically, the difference isn’t super relevant. However, for other `#lang`s, it might be.

The idea is that `#%namespaced` is much “lower level” than something like `local-require`. It’s the absolute minimum functionality required to add a way for meta-languages to ask for specific bindings without doing things like producing syntax from `read-syntax` with scopes. Using `local-require` for this would require meta-languages to make two assumptions:

  1. All `#lang`s that support language extensions provide `local-require` under the same name.
  2. All `#lang`s that support language extensions provide a block-scoping form under the same name (probably `let`).

These are much stronger requirements than asking all languages to reprovide `#%namespaced` from `racket/base`, which is conceptually much closer to reproviding similar hooks like `#%datum` or `#%top`. Additionally, `local-require` + `let` probably wouldn’t be a good replacement for `#%namespaced`, anyway, since that would make things like this fail to work:

```racket
#lang curly-fn racket/base
(require (only-in racket/base [let racket-let])
         (for-syntax syntax/parse))

; use a Clojure-like let
(define-syntax let
  (syntax-parser
    [(_ [(~seq x:id v:expr) ...] body:expr ...+)
     #'(let ([x v] ...) body ...)]))

#{+ % 2} ; oops, now `let` has different semantics
```

Basically, it seems extremely likely that `let` will be rebound in different `#lang`s, so it wouldn’t make sense to use it as part of an implicit protocol all `#lang`s must support. The `#%namespaced` form makes the protocol as simple and explicit as possible.

mflatt:
@lexi-lambda Thanks – that makes sense.

Sorry for delaying this so long, but I think we should wait at least until after the v6.8 branch before merging.

Then, I wonder how bad it would be to delay even a little longer:

Not only am I back to Racket work (my semester dominated by teaching is over, happily), but I expect to rewrite the Racket reader as my next task. It seems possible that a fresh look at the reader and reader–expander stack will suggest a more general solution, especially with bytecode-level changes as fair game. It would be nice to avoid `#%namespaced` if a more general replacement is in reach.

Then again, I don't know whether the reader implementation itself is a 1-week, 1-month, or longer task. It may be too much of a delay, considering that I don't even know whether we'll find a more general solution, and considering we have a useful intermediate solution.

lexi-lambda:
That all makes sense to me. `#%namespaced` is a hack. It’s a hack that solves a problem, but it’s still a hack. If you think that there’s even a *possibility* that it can be done better, then I think it’s worth putting it off to try that first. I would only suggest going this route if it’s really unclear how to do it otherwise.

LeifAndersen:
Has this idea died completely? Or have we found a better approach.

lexi-lambda:
My understanding is that we are waiting for the Racket rewrite of the macroexpander to become *the* macroexpander, which is currently waiting on the Chez rewrite, so this has been put on hold until that takes place. Is there something you would like it for?

LeifAndersen:
Nothing urgent. I just think it would be nice to add new (not entirely s-expressiony) syntax to Video, and I would (ideally) like to do it in a composable way.

mbutterick:
I’ve been wanting to add a tutorial about metalanguages to _Beautiful Racket_. But this issue has puzzled me.

> There are two obvious ways to fix this problem. The first would be to enhance the compiler and macro expander to properly support syntax objects with lexical [context] … The alternative approach is to provide a hook in base languages that language extensions can use to safely introduce hygienic syntactic extensions.

I find it confusing that a metalanguage hooks into only the reader but would expect to behave as both a reader (that is, converting bytes from an input stream into syntax objects) **and** an expander (that is, introducing bindings for these syntax objects). 

The idea of inventing a new “hygienically expandable reader” layer — perhaps it would work, but it seems to merge two ideas that are separate for good reasons (much like [ligers](http://ligerfacts.org/)). This would create a conceptual disjoin with the standard language-making workflow.

Why not give the metalanguage two bites at the apple, just like any other language? That is, let the metalanguage have a `reader` submodule (that implements its reader hacking) and then an expander (to handle the syntax it introduced).

So instead of this:

source code → metalanguage reader (that maybe sneaks in some bindings early) → main reader → metalanguage expander

The flow would look like this:

source code → metalanguage reader → main reader → metalanguage expander → main expander

Thereby preserving the firewall between reader world (where no bindings tread) and expander world.

AlexKnauth:
How would it know which "expander layer" it should use to look up bindings for any given form?

The status quo answers that question with: "use the main expander for everything."

This pull request answers that question with: "use the module given in the `#%namespaced` form for that."

What your comment seems to suggest to me is: "have each reader layer tag its outputs with the expander layer that goes with it, and when racket's expander is processing a form, it should use the expander layer shown by the tag on that form."

Is that right? Or were you thinking of something else?

lexi-lambda:
> I find it confusing that a metalanguage hooks into only the reader but would expect to behave as both a reader (that is, converting bytes from an input stream into syntax objects) **and** an expander (that is, introducing bindings for these syntax objects).

I disagree with this part of the premise. It’s true that it’s been Lisp tradition to have a “reader” that produces raw, “flat” data, but Racket is not a traditional Lisp, and its reader already produces syntax objects (albeit syntax objects with no lexical context). Why do we force our readers to produce unhygienic data to be at the mercy of the module language? Would it not be more elegant for `#lang`s to be able to produce syntax objects fully enriched with lexical context so that users do not have to worry about mangling the names of their core forms with things like `#%my-form`?

The current approach of producing syntax objects with no lexical context works well for `#lang racket`, which has very little syntax to begin with, and all of its types of syntax are essentially represented in the sorts of things that may be 2D syntax objects. Other `#lang`s with more syntax do not have this luxury. For example, consider my own `#lang tulip` implementation, which uses a non-s-expression syntax. A lambda in `#lang tulip` is written `[ x => add 1 x ]`. A user of this `#lang` should not be able to shadow the binding that implements the lambda syntax! So my reader produces something like `(#%lambda [(x) (add 1 x)])`, and fortunately, identifiers beginning with `#%` are not legal in Tulip. However, it’s annoying that this is even a problem—macros do not have this issue, since they can use lexical scope to hide their implementation details.

It seems to me that the nicest solution to this problem is to make readers *more* hygienic, not less. Instead of forcing `#lang` authors to do a complicated dance of producing a magical symbolic name and arranging for that name to be bound, just let them produce a syntax object with some preset lexical context. It solves the `#lang` problem and the meta-language problem, and it is a strict superset of existing functionality. (The only potential gotcha is that some tools might assume using `read` instead of `read-syntax` is okay if they don’t want source locations, but existing meta-languages *already* violate that assumption by depending on the undefined behavior of producing syntax objects with scopes from the reader.)

mbutterick:
FWIW I think this PR edges into the broader question of how to compose `#lang`s — which is interesting and largely unexplored. Should it be possible? Seems like it. How? Building upon the current metalanguage technique seems iffy, given that many of the interesting things that metalangages want to do contradict longstanding Racket policies. IOW, perhaps better to put metalanguages on the shelf next to PLaneT, and start over with a better idea.

> How would it know which "expander layer" it should use to look up bindings for any given form? … What your comment seems to suggest to me is: "have each reader layer tag its outputs with the expander layer that goes with it, and when racket's expander is processing a form, it should use the expander layer shown by the tag on that form."

Mark the relevant identifiers with syntax properties? If you want to maintain a correspondence between which syntax objects are handled by which expander. That may not always be the case however.

> Why do we force our readers to produce unhygienic data to be at the mercy of the module language? 

So it’s possible to reason about how the expander will affect its input. “No bindings” is a strong, simple, verifiable guarantee. “Maybe bindings, sometimes, depending” is not.

I could flip your question around: why should `#lang` authors be at the mercy of whatever metalanguages users add to the `#lang` line? Given that today’s `#lang`s are a bread-and-butter feature in Racket, and today’s metalanguages something of a sideshow, they can’t be put on the same footing.

lexi-lambda:
> So it’s possible to reason about how the expander will affect its input. “No bindings” is a strong, simple, verifiable guarantee. “Maybe bindings, sometimes, depending” is not.

Yes, it would certainly be necessary to come up with a well-defined, internally consistent semantics for what scopes on reader-produced syntax *mean*. If that turned out to be way harder than anticipated, I would not recommend it, but it seems feasible to me.

> I could flip your question around: why should `#lang` authors be at the mercy of whatever metalanguages users add to the `#lang` line?

I’m not sure I understand how your proposal would intrinsically give `#lang` authors any more control over metalanguages than adding scopes to reader syntax would.

That said, yes, it’d be nice to make sure `#lang`’s invariants can’t be broken by bad actors. Currently, meta-languages that introduce new forms unfortunately already have the potential to do that. The general question here of how to safely compose languages seems like a rich area of research, but existing meta-languages are already exploiting undefined behavior to do what they need to do, and I think we mostly agree what they do is useful and desirable. There’s certainly a discussion to be had over how much we are willing to do something quick to support metalanguages or if we ought to wait until we know how to do The Right Thing.

This pull request is clearly *not* The Right Thing. Both my proposal and your proposal seem a little closer to The Right Thing, but I’d bet they’re still pretty far off if we care about enforcing language invariants and making sure languages actually compose. I imagine it’s trivial to show that languages do not compose in general, so there would likely need to be a restricted but sufficiently expressive language to express language extensions… but that seems well into the bounds of new research and is probably out of scope for a discussion on a GitHub pull request.

mbutterick:
Sorry to be a moron, but if you refined the idea behind `#%namespaced` to use, say, identifier renaming rather than scopes — wouldn’t you end up with a solution compatible with today’s `#lang` convention that `read-syntax` ought to return a syntax object with no bindings?

Suppose you have `#lang mu hostlang`. Roughly:

1. `mu` is a metalanguage that wraps the reader as usual and relies on certain bindings. Let’s say `let` and `local-require` from `racket/base`. 

2. Obviously, `mu` can’t assume that `let` or `local-require` is available in `hostlang` with the expected meanings. So `mu` also provides its `let` and `local-require` (as an expander module would).

3. When the `mu` reader extension runs,  its uses of `let` and `local-require` are _✨waves hands:sparkles:_ replaced with `mu:let` and `mu:local-require`.

4. The `read-syntax` in `hostlang` can do `strip-context` (or equivalent) on its syntax object, and these identifier names will remain intact.

5. Then the `hostlang` expander runs. Everything is good except there’s no binding for `mu:let` and `mu:local-require`. This would ordinarily trigger an unbound-identifier error, but ...

6. …  _✨waves hands, more vigorously:sparkles:_ Racket has shimmed `#%top` to import the identifiers from `mu`, prefixed as before. So they are bound to the values `mu` intended.

7. Repeat for other metalanguages on the `#lang` line.

lexi-lambda:
That scheme doesn’t work (nor does *any* name-mangling scheme), since it will capture user-defined bindings named `mu:let` (since `#%top` only gets introduced if the name is unbound). Also, what if the language provides its own version of `#%top`? Also also, it’s not immediately clear to me why making `#%top` do this magically is any easier than just making the expander accept syntax from the reader with scopes on it.

Maybe the first two points are intentional? But it’s not immediately clear to me why they would be advantageous.

jackfirth:
There was a little discussion at RacketCon 2017 about the problem this PR tackles and one interesting proposed solution was to reframe the problem as one of reader _datums_ rather than reader _bindings_. Basically, instead of producing a list reader extensions could produce syntax objects containing a _Q value_: an entirely new kind of primitive value that self-describes how to macroexpand it. Q values would contain the following:

- Some other Racket value representing its "datum", which might contain nested syntax objects within it just like list syntax objects.
- A module binding reference (symbol + module path + phase) for a function that defines how to macroexpand the datum.

It would need to be a new kind of value instead of, say, structs with some special property because transparent structs aren't cross-phase persistent and prefab structs can't have struct properties.

In this brave new world, reading and expanding `#lang curly-fn racket/base` would work like this:

- The `curly-fn` reader reads `racket/base` to determine the host language reader, then overrides the host reader to read `#{list 1 % 3}` as a Q value containing the list `'(list 1 % 3)`, the symbol `curly-fn-expand`, and the module path `(lib "curly-fn")`.
- The `racket/base` reader does what it normally does to produce a syntax object without any lexical information.
- The Racket expander does what it normally does _except_ the `#{list 1 % 3}` Q value is expanded by dynamically requiring `curly-fn-expand` from `curly-fn` and calling it with two arguments: the Q value's datum `'(list 1 % 3)` and a procedure for recursively expanding nested syntax objects.
- The `curly-fn-expand` procedure implements the expansion by inserting a `#%curly-fn` identifier that's bound as a macro in the `curly-fn` module (_not_ an identifier with no binding like `#%app`).
- That `#%curly-fn` macro does normal macro-y things, possibly using syntax parameters to communicate with the host language and imported libraries it's being used with.

This preserves clean separation between the reader and expander: the former produces only structured data describing what the programmer _wrote_, while the latter resolves bindings and expands transformers to produce structured data describing what the programmer _meant_. It also gives reader extensions control over how languages and modules can configure or override them. It's essentially if you took `#%namespaced` and made it an entirely separate kind of value that the Racket core expander treats specially instead of a magic hook macro that a host language has to implement with special scoping logic.

I'm pretty sure the name "Q value" was chosen without much significance and I might not even be remembering the name correctly. Ask @willghatch for more details (since it was his idea).

lexi-lambda:
Another thought, similar to @jackfirth’s comment: we might already have a way to do this in `#%datum`. If readers produce 3D syntax, the macroexpander will wrap it in `#%datum`. Currently, Racket’s `#%datum` blindly expands to `quote`, but it would be possible to create a protocol certain values could implement (structure type properties or generic interfaces, perhaps?) that would be handled differently by `#%datum`. This would still involve tweaking the `#%datum` form from `racket/base`, so it isn’t free, but it might be less complicated than changing the macroexpander. On the other hand, I don’t know how the reader interacts with the separate module instantiations introduced by phasing, so the structs would likely need to be cross-phase persistent, which I think currently rules out generic interfaces (but structure type properties would work).

PR #1460, opened 2016-09-14
add xml-attribute-encode function

jbclements:
Per discussion with @jeapostrophe , adds `xml-attribute` encode function (and test, and docs)


jeapostrophe:
Docs?

jbclements:
Oh ... yep, wrote docs, didn't include them. Hang on.

jbclements:
okay, docs added. Thanks for catching that!

jeapostrophe:
Thanks for catching this!

PR #1464, opened 2016-09-18
add file/glob

bennn:
API for Unix-style globs on path-strings.

### DIFF
--- a/pkgs/racket-doc/file/scribblings/glob.scrbl
+++ b/pkgs/racket-doc/file/scribblings/glob.scrbl
@@ -1,0 +1,125 @@
+#lang scribble/doc
+@require["common.rkt"
+         scribble/examples
+         (for-label racket/sequence)]
+
+@title[#:tag "glob"]{Globbing}
+
+@defmodule[file/glob]{The @racketmodname[file/glob] library implements
+@hyperlink["https://en.wikipedia.org/wiki/Glob_(programming)"]{globbing}
+for @racket[path-string?] values. A @emph{glob} is a path string that matches
+a set of path strings using the following @emph{wildcards}:
+@itemlist[
+@item{
+  A sextile (@tt{*}) matches any sequence of characters in a file or directory
+  name.
+}
+@item{
+  Two sextiles (@tt{**}) match any sequence of characters and any number of
+  path separators.
+}
+@item{
+  A question mark (@tt{?}) matches any single character in a file or directory
+  name.
+}
+@item{
+  Square bracket-delimited character groups, e.g. @tt{[abc]}, match any
+  character within the group. The square brackets have the same meaning in globs
+  as in regular expressions, see
+  @secref["regexp-syntax" #:doc '(lib "scribblings/reference/reference.scrbl")].
+}
+@item{
+  If the glob ends with a path separator (@tt{/} on any @racket[(system-type)],
+  additionally @tt{\} on @racket['windows])
+  then it only matches directories.
+}
+]
+
+By default, wildcards will not match files or directories whose name begins
+with a period (aka "dotfiles").  To override, set the parameter
+@racket[glob-capture-dotfiles?] to a non-@racket[#f] value or supply a similar
+value using the @racket[#:capture-dotfiles?] keyword.
+}
+
+@defthing[glob/c (or/c path-string? (sequence/c path-string?))]{
+A @tech[#:doc '(lib "scribblings/reference/reference.scrbl")]{flat contract}
+that accepts a glob or a sequence of globs.
+}
+
+All @racketmodname[file/glob] functions accept @racket[glob/c] values.
+These functions also recognize braces (@tt|{{}}|) as a @emph{meta-wildcard} for
+describing multiple globs.
+@margin-note{Braces are interpreted @emph{before} any other wildcards.}
+
+@itemlist[
+@item{
+Brace-delimited, comma-separated character groups, e.g. @tt|{{foo,bar}}|,
+expand to multiple globs before the @racketmodname[file/glob] module begins
+matching. For example, the @racket[glob/c] value @racket|{{foo,bar}.rkt}|
+has the same meaning as @racket['("foo.rkt" "bar.rkt")].
+}
+]
+
+@defproc[(glob [pattern glob/c] [#:capture-dotfiles? capture-dotfiles? boolean? (glob-capture-dotfiles?)]) (listof path-string?)]{
+  Builds a list of all paths on the current filesystem that match any glob
+  in @racket[pattern]. The order of paths in the result is unspecified.
+
+  If @racket[pattern] contains the wildcard @tt{**}, then @racket[glob]
+  recursively searches the filesystem to find matches.
+  For example, the glob @racket{/**.rkt} will search the @emph{entire filesystem}
+  for files or directories with a @racket{.rkt} suffix (aka, Racket files).
+
+Examples:
+@codeblock{
+> (glob "*.rkt")
+;; Lists all Racket files in current directory
+
+> (glob "*/*.rkt")
+;; Lists all Racket files in all sub-directories of the current directory.
+;; (Does not search sub-sub-directories, etc.)
+
+> (glob (build-path (find-system-path 'home-dir) "**" "*.rkt"))
+;; Recursively searches home directory for Racket files, lists all matches.
+
+> (glob "??.rkt")
+;; Lists all Racket files in current directory with 2-character names.
+
+> (glob "[a-z0-9].rkt")
+;; Lists all Racket files in current directory with single-character,
+;; alphanumeric names.
+
+> (glob '("foo-bar.rkt" "foo-baz.rkt" "qux-bar.rkt" "qux-baz.rkt"))
+;; Filters the list to contain only files or directories that exist.
+
+> (glob "{foo,qux}-{bar,baz}.rkt")
+;; Same as above, returns at most 4 files.
+}
+}
+
+@defproc[(in-glob [pattern glob/c] [#:capture-dotfiles? capture-dotfiles? boolean? (glob-capture-dotfiles?)]) (sequence/c path-string?)]{
+  Returns a stream of all paths matching the glob @racket[pattern],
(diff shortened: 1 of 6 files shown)


pkgs/racket-doc/file/scribblings/glob.scrbl
```diff
@@ -0,0 +1,53 @@
...
+  @racket{*} matches any sequence of characters.
+}
+@item{
+  @racket{?} matches any single character.
+}
+@item{
+  Square bracket-delimited character groups, e.g. @racket{[abc]}, match any character
+  within the group. The square brackets have the same meaning in globs as in
+  regular expressions, see @secref["regexp-syntax" #:doc '(lib "scribblings/reference/reference.scrbl")].
+}
```
jeapostrophe: You should also add {a,b}-{c,d}.rkt as a way to write a-c.rkt a-d.rkt b-c.rkt b-d.rkt

jeapostrophe:
This is great!

elibarzilay:
1. "implements Unix-style wildcards" -- globs are mostly independent of unix.  Maybe just the dot-prefix thing...  (Also at the top of the code.)

2. The documentation uses a bunch of `@racket`s that look wrong and should probably be `tt` or similar.

3. It would also be *very* useful to have `**` (which can match path separators too).

4. But note about {}s: they are *not* a globbing tool.  In shells they get expanded before the globbing stage.

5. Using globs in path strings to represent globs sounds dangerous to me.  I'm guessing that you don't want "?:\\" on windows to expand to all drives or do local network searches for "\\\\*\\stuff".

6. I really dislike the global `*glob-capture-dotfiles?*` parameter -- it would be much nicer if it were a property of the glob itself.

7. I'm not sure that `in-glob` is the only useful thing -- stuff like `glob->list` etc could still be useful.

8. But in any case, it's often useful to specify several patterns so they get `or`ed (roughly) -- so it would be useful for `in-glob` to get multiple inputs to do that.  Then it might make sense to add braces as a shorthand for multiple strings, similar to what shells do.

9. Either the implementation or the description of `glob-match?` seems problematic: it looks like `(glob-match "foo" "foo")` will only be true if there is a file/directory named "foo" in the current directory.

10. It's not clear to me whether something like "*/*/*" would work -- I hope it does, but then the docs should say that.

11. In case it helps -- I used to have some old globbing code for meta-stuff which was working pretty well but it probably died with the move from svn to git.  I put it at http://tmp.barzilay.org/tree.ss (there are also some tests for it if you're interested).  The main thing there is a function that turns a glob to a regexp (and from that to a predicate), and the test cases try a bunch of edge cases.

12. The code has a mistaken comment saying "? makes the previous character/pattern optional,"

pkgs/racket-doc/file/scribblings/glob.scrbl
bennn: I don't get it -- can you explain more what the `{}` should do? (Right now `"[ab]-[cd].rkt"` matches the same 4 files.)

bennn:
(Found the tests at http://svn.plt-scheme.org/plt/trunk/ )

pkgs/racket-doc/file/scribblings/glob.scrbl
jeapostrophe: braces work on whole strings

{foo,bar}-{zog,zag}.rkt

is

foo-zog.rkt foo-zag.rkt bar-zog.rkt bar-zag.rkt

Your shell does this :)

gus-massa:
> I really dislike the global `*glob-capture-dotfiles?*` parameter -- it would be much nicer if it were a property of the glob itself.

What about replacing it with a keyword arguent for `glob` / `in-glob` / `glob-match?` ?

Also, I think that any of the current parameters in racket have the `*___*` in their names.

bennn:
The latest commit implements `**` and `{foo,bar,{baz,qux}}` etc. & addresses most of @elibarzilay 's comments (thanks again Eli & Jay & Gustavo).

Things that we might still want to change:
- I added keyword arguments, but kept the parameter (as the keywords' default, also without the `**`)
- globs are still path strings -- because I think it's a convenient interface & couldn't think of a "less dangerous" option.
- I don't understand point `7` about `glob->list`. I think `glob` is doing what `glob->list` would do.

Docs here: http://www.ccs.neu.edu/home/types/racket-doc/file/glob.html

jeapostrophe:
I say merge!

bennn:
2 technical issues:
- Is it possible to use `rackunit` in the `file` collection or in `racket-test/tests/file` ? The CI failed when I tried `(require rackunit)`.
- [L40](https://github.com/racket/racket/pull/1464/files#diff-751266cc6e8f06cdc77958951f21cd0eR40) in the doc has a broken cross reference to "flat contract", but I don't know how to fix it.

stamourv:
You should be able to use `rackunit` from `racket-test/tests/file`, but not from inside the `file` collection proper.
To fix the cross-reference, use the top-level of the reference as the `doc`, instead of the specific section on contracts.

elibarzilay:
I'll enumerate, as usual...

1. Re my earlier point 7: it looked to me like the only interface you provide is `in-glob`, but I can think of at least two additional useful things: a `glob->list` that returns the list (instead of going through an extra step to get it, in the many cases where a list is fitting); maybe some `glob->pred` that returns a predicate (or, if you're doing a similar hack to what I did: some `glob->regexp`).

2. I see now that I fell into the GH markdown trap -- in point 10 the thing that was unclear to me was whether things like `*/*/*` work: I'm sure that most people expect it to work, but there are cases where you can only have a `*` in the last element, so it would be good to mention that.

3. Re the `foo{x,y}bar` thing: this is a good thing in general, but note that there could be two ways to interpret it.  The more obvious one (IMO) is "paths whose name matches `foo` followed by `x` or `y` followed by `bar`".  But what shells do is subtly different: they do the brace expansion *before* expanding globs.  For example, in a new directory do a `touch fooxbar` and then try `echo foo{x,y}bar*` -- it behaves differently in bash and in zsh, and in each one can behave differently based on some shell configuration.  IMO, the former interpretation is more reasonable when you deal with file globs (shells have a problem where they can't tell what's intended to be a glob and what not), and therefore at some point I had used a `foo(x|y)bar` syntax instead.

   In any case, I'm not criticizing what you did (I didn't even try it, so I really don't know) -- but it's something that should be mentioned explicitly in the docs either way.  From a brief readin, it looks to me like you're doing the shell thing, but because you just use it to match, you get the good behavior, and in that case describing it as first doing the `{}` thing might be misleading.


4. I see this in the docs:

   > If the glob ends with a path separator (`/` on `'unix` or `'macosx` systems, `\` on `'windows` then it only matches directories.

   This looks wrong because the convention is that `/` can also be used on Windows to separate directories, so I think that a suffix of `/` should mean the same across all platforms, and `\` works also if you're on windows.  This is especially true since you're keeping the paths-or-strings thing, which means that I would expect a path of `foo/` to match a `foo` directory when it's used as a glob on windows too.

5. Finally, having both a keyword and a parameter means that either it's complictaed (as in having the parameter have three values: include dots, don't include them, and unset; the first two overriding a keyword, and the last being the default), or it's simple and then the parameter can mess up otherwise-fine code.  IMO the complexity is not worth it, and the simple case is too fragile.

bennn:
Thanks,

1. [`(glob ....)`](https://github.com/racket/racket/pull/1464/files#diff-751266cc6e8f06cdc77958951f21cd0eR61) builds the list of matching files. `(λ (x) (glob-match? g x))` [(doc)](https://github.com/racket/racket/pull/1464/files#diff-751266cc6e8f06cdc77958951f21cd0eR109) does what `(glob->pred g)` would do. There is a `glob->regexp` in the implementation, but I would rather just export `glob-match?` because `glob-match?`'s implementation is a little smarter than just calling `regexp-match?` on the whole pattern.

2. `*` and `**` can be anywhere in the pattern, there's 2 examples in the docs for this: [L75](https://github.com/racket/racket/pull/1464/files#diff-751266cc6e8f06cdc77958951f21cd0eR75) [L79](https://github.com/racket/racket/pull/1464/files#diff-751266cc6e8f06cdc77958951f21cd0eR80)

3. Brace expansion happens before any globs are expanded -- so `(glob "{foo,bar}*")` gives the same result as `(glob '("foo*" "bar*"))`. I tried to explain this in the docs with an [example](https://github.com/racket/racket/pull/1464/files#diff-751266cc6e8f06cdc77958951f21cd0eR92) and calling braces a [meta-wildcard](https://github.com/racket/racket/pull/1464/files#diff-751266cc6e8f06cdc77958951f21cd0eR50), but I'll add another line to make it clear.

4. Fixed, thanks

5. The parameter has 2 values, `#f` and anything else, and changes the default value of the keyword argument `#:capture-dotfiles?`. The use-case I have in mind is that someone writes a script with a few calls to glob functions, then realizes they wanted to match dotfiles. They can just set the parameter instead of modifying the call sites. (And if you do specify `#:capture-dotfiles?`, that call ignores the parameter.) Can you explain how it might be fragile?

About `{a,b}` vs. `{a|b}`, I think we should stick to `,` because it'll be familiar to (some) shell, [perl](http://perldoc.perl.org/functions/glob.html), and [php](http://php.net/manual/en/function.glob.php) programmers. But I do like the `|` --- anyone else have thoughts about switching?

elibarzilay:
1. Sounds good.

2. -"-

3. Given that the overall effect is to leave only strings that match paths, I'm saying that you could downplay the two stages.

4. -"-

5. Oh, I missed that the parameter sets the default value for the keyword.  That sounds perfectly fine then.

6. Yeah, I gave my `(|)` choice as an example, I don't really have an opinion on it, and preferring `{,}` has valid reasons...

bennn:
Just squashed the commits. Let's merge in T-minus 2 days?

stamourv:
T-minus 2 days?
I'd recommend merging now; branch day is tomorrow.

bennn:
👍  97c65102b3ca4aaacf0674fe691e3c0783224bc0

Danl2620:
(I filed this issue for some specifics: https://github.com/racket/racket/issues/4000)

PR #1466, opened 2016-09-20
raco pkg new: include v6.5 and v6.6 and v6.7 in .travis.yml

jsmaniac:
Also, by default, `- raco setup --check-pkg-deps --pkgs <<name>>` is in the `after_success` section, which means that it can fail and the build will still pass. Should I move it up into the `script` section, so that dependency errors fail the build by default on new packages?


jsmaniac:
I just updated this to make Travis build new packages on v6.7 by default. Ping @stamourv : should this be included in the release branch?

stamourv:
Hmm, that's an interesting one.
Yes, it would be nice to have this in the release. But, until the release actually happens, this change "breaks" raco pkg new, in that it will ask Travis to look for a version that doesn't exist (yet).
I guess I could merge it to the release branch, but only merge it to master after 6.7 is released?

In any case, any additions to the release at this point need consensus from the release managers. I'll bring it up.

jsmaniac:
@stamourv yes, it breaks the Travis build for packages created via `raco pkg new` until 6.7 is out. I thought 6.7 would be out soon enough that it wouldn't really matter (plus the Travis error is easy enough to spot), but your idea of merging on `release` first then on `master` when 6.7 is out sounds way better :) .

stamourv:
@jsmaniac: After consultation with release management, we'll go with merging to release, then merging to master once the release is out.

In general, though, hard-coding release numbers like that is a terrible idea (not blaming you).
A better solution would be to have `raco pkg new` write down the version it's using itself (if it's a release version), maybe defaulting to the latest (there's code for latest version checking somewhere) if it's not. Want to give it a shot?

jsmaniac:
@stamourv I'll think about it. I suppose something along these lines would work:

```
(require version/utils)
(string-join
 (for*/list ([major (in-range 6 (add1 (car (version->list (version)))))]
             [minor (in-range 0 (add1 (cadr (version->list (version)))))])
   (format "- RACKET_VERSION=~a.~a" major minor))
 "\n")
```

It will definitely require some testing (I'm not sure if `(version)` always returns `major.minor.x.y`, for starters), so I'll put that on my todo-list for later.

stamourv:
That general approach should work. A couple comments.
* Do we want to list all these versions? That doesn't seem to be a great use of Travis, no?
* This won't catch releases like 6.1.1, which was a regular (i.e., not a quick bug fix) release, AFAIR. We've basically moved to just using major.minor, but I expect that bug fix releases may still end up using the 3rd number. And we'd definitely want to test those.

jsmaniac:
Yeah, I thought the `6.1.1` and future bug-fix releases would still need to be hard-coded. I don't really see how this could be done otherwise: if there is a 6.7.1 and a 6.7.2 release, and we move on to 6.8, the 6.8 release does not contain any reference to the 6.7.1 and 6.7.2 releases.

In the end, the candidate versions are those present on http://mirror.racket-lang.org/installers/, but I don't really like the idea of fetching this within `raco pkg new` (I'd expect that command to work in the same way, regardless of the network connectivity).

As for whether to include all the versions by default, or a specific set + the latest / the last 3 / …, I'd rather leave the decision to people who are more knowledgeable about the specificities of each relase (i.e. I probably should ask on the dev mailing list).

stamourv:
Personally, I'd be fine with just having `raco`'s own version (or latest, if it's not a release) in the list.
`raco pkg new` is not the be-all end-all of good testing practices. It's there to give you a starting point, and show you what you can do. Having a Travis config at all is enough for.

stamourv:
Merged into master. Should I leave this PR open for when you add the improvements we discussed?

jsmaniac:
@stamourv I'm closing this PR, and I opened a separate issue to choose which versions to include, as it's not quite clear for me what is the best automatic strategy.

PR #1467, opened 2016-09-21
optimizer: char?, small-char? and more reductions like (equal? x y) => (eq? x y)

gus-massa:
The first part consist in some new type annotation for the optimizer, in particular a new primitive `small-char?` that recognize chars with a `value < 256`. This is useful for the second part that has some extension for reductions like `(equal? x y) => (eq? x y)`. The idea is to classify the relevant predicates, and use that classification to enable more optimizations.

**All strict-true? branch + fix (ignore)**
These commits use some of the new code from https://github.com/racket/racket/pull/1457, so I copied a squashed version here.

**add symbol? and keyword? to relevant predicates** 
Nothing surprising here. Also, includes tweaks to the types infered for the arguments of other functions like `box` and `unbox`.  

**add char? and small-char? to relevant predicates**
A new primitive `small-char?` that recognize chars with a `value < 256`, because if they are `equal?` then they are `eq?`. The other "bigger" chars are only `eqv?` when `equal?`.

**Use compiled startup again (squash)**
I’ll squash this before merging.

**extend the reductions like (equal? x y) => (eq? x y)**
Classify the relevant predicates in 4 categories. In particular, if `<expr>` satisfy `pred?` we can use this classification to enable the correct reduction:

```
(equal? <expr> y) ==> [no reduction, unless y has a different type]
(equal? <expr> y) ==> (eqv? <expr> y)
(equal? <expr> y) ==> (eq? <expr> y)
(equal? <expr> y) ==> (begin <expr> (pred? y))
```


mflatt:
These changes are good as far as I can tell.

Maybe `latin-1-char?` instead of `small-char?`? Or `interned-char?`, which more directly expresses the relevant property, but grates slightly with `symbol-interned?` (which only works on symbols).

gus-massa:
I like `interned-char?`, specially because perhaps some day more chars will be interned. (I don't have any unicode block in mind now, but it doesn't seam to be an impossible idea.) 

But I agree that it sounds too similar to `symbol-interned?`. I'll wait a few days, and if no one propose a better name I'll merge this with `interned-char?`.

mflatt:
Ok!

gus-massa:
I found two small bugs in the jit code, but they are easy to fix.

In particular 

    (define (s-c? x) (small-char? x))
    
    (s-c? #\a)        ;==> #f   (jit)
    (small-char? #\a) ;==> #t   (no-jit)

They are not discovered by the tests because most of the tests run the not jitted code. Is it posible to force all the code to be jitted? (This is more a feature request than a real question.)

I tried to rebase all of this, but I'm geting segfault when I reenable cstartup.inc with the 1154 expected primitives . I'll try to fix whatever is broken and if in a few days I still have problems I'll ask for help.

rfindler:
Could you use this macro to jit things?

  (define-syntax-rule (jit-it e) ((lambda () e)))

Robby

gus-massa:
I’ll merge this in a minute.

The error was very silly. I was adding a new primitive `interned-char?`, but at the same time I was removing a duplicated version of `char-title-case?`, so the total count of primitives wasn’t changing. I forgot about `char-title-case?` and while rebasing I was trying to increase the total count of primitives, so I got an error.

I finally left the `syntax?` primitive classified  as `RLV_IS_RELEVANT` instead of `RLV_EQ_TESTEABLE`. This classification is a little overcautious an it disables the reduction `(equal? stx-x y) => (eq? stx-x y)`, where `stx-x` is a `syntax?`.  I think that it would be correct to enable this reduction, but `syntax?` is a very complex type and it’s a little weird that it doesn’t have some kind of custom criteria for `equal?`.

PR #1469, opened 2016-09-24
raco: show all the matching commands for an ambiguous prefix

gus-massa:
When a an ambiguous command is used in _raco_, it's difficult to guess how many additional letter to add to break the ambiguity. So the idea is to show all the matching commands. For example:

```
>raco p
raco.exe: Ambiguous command prefix: p
Usage: raco <command> <option> ... <arg> ...
Matching commands:
  pkg            manage packages
  pack           pack files/collections into a .plt archive
Frequently used commands:
  make           compile source to bytecode
  setup          install and build libraries and documentation
  pkg            manage packages
  exe            create executable
  test           run tests associated with files/directories
A command can be specified by an unambiguous prefix.
See `raco help' for a complete list of commands.
See `raco help <command>' for help on a command.
```

To add this, I had to refactor a lot of code, so the changes appear to be bigger than the real changes.

(Moreover, perhaps I should have splitted the `let*` at the bottom in a bunch of `defines`.)

### DIFF
--- a/racket/collects/raco/raco.rkt
+++ b/racket/collects/raco/raco.rkt
@@ -1,27 +1,38 @@
 #lang racket/base
 (require "command-name.rkt"
-         "all-tools.rkt")
+         "all-tools.rkt"
+         racket/string)
 
 (module test racket/base)
 
-(define (find-by-prefix hash str)
-  (let ([trie (make-hash)])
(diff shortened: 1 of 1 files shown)


mflatt:
Merged after adding a repair for `raco` without arguments. Thanks!

PR #1472, opened 2016-09-29
Add support for git-backed packages that require authentication

lexi-lambda:
This set of changes implements support for packages backed by private git repositories that support authentication via HTTP Basic Authentication (which GitHub supports, for example). It is divided into three parts:

  1. It adjusts `net/git-checkout` to support `#:username` and `#:password` options when cloning a package.

  2. A new configuration option, `git-checkout-credentials`, is added to `raco pkg config`. This is a set of credentials that the package system will attempt to supply when installing a package.

  3. The package system itself is adjusted to supply the aforementioned credentials. It will always attempt to download a package without any authentication first, but if that fails, it will retry with credentials. If none of the sets of credentials are successful, then the package will fail to install as normal.

This permits consuming packages backed by private GitHub repositories, which is a feature my team would need to have available if we want to use Racket for some of our internal tooling.

### DIFF
--- a/racket/collects/pkg/private/checkout-credentials.rkt
+++ b/racket/collects/pkg/private/checkout-credentials.rkt
@@ -1,0 +1,24 @@
+#lang racket/base
+
+(require net/git-checkout
+         racket/list
+         "config.rkt")
+
+(provide call-with-git-checkout-credentials)
+
+(define (call-with-git-checkout-credentials thunk)
+  (let loop ([credentials-list (cons #f (get-git-checkout-credentials))])
+    (define credentials (first credentials-list))
+    (with-handlers ([exn:fail:git? (λ (exn)
+                                     (if (empty? (rest credentials-list))
+                                         (raise exn)
+                                         (loop (rest credentials-list))))])
+      (define c (make-custodian))
+      (parameterize ([current-custodian c]
+                     [current-git-username (and credentials (hash-ref credentials 'username))]
+                     [current-git-password (and credentials (hash-ref credentials 'password))])
+        (dynamic-wind
+         void
+         thunk
+         (lambda ()
+           (custodian-shutdown-all c)))))))
--- a/pkgs/net-doc/net/scribblings/git-checkout.scrbl
+++ b/pkgs/net-doc/net/scribblings/git-checkout.scrbl
@@ -36,7 +36,9 @@ for information on command-line arguments and flags.
                                                                      [(git) 9418]
                                                                      [(http) 80]
                                                                      [(https) 443])]
-                       [#:strict-links? strict-links? any/c #f])
+                       [#:strict-links? strict-links? any/c #f]
+                       [#:username username (or/c string? #f) (current-git-username)]
+                       [#:password password (or/c string? #f) (current-git-password)])
          string?]{
 
 Contacts the server at @racket[hostname] and @racket[port]
@@ -96,6 +98,20 @@ If @racket[strict-links?] is true, then the checkout fails with an
 error if it would produce a symbolic link that refers to an absolute path
 or to a relative path that contains up-directory elements.
 
+If both @racket[username] and @racket[password] are non-@racket[#f]
+@emph{and} @racket[transport] is @racket['http] or @racket['https], then
+the provided credentials are passed to the remote server using HTTP Basic
+Authentication.
+
 @history[#:added "6.1.1.1"
          #:changed "6.3" @elem{Added the @racket[initial-error] argument.}
-         #:changed "6.2.900.17" @elem{Added the @racket[strict-links?] argument.}]}
+         #:changed "6.2.900.17" @elem{Added the @racket[strict-links?] argument.}
+         #:changed "6.6.0.5" @elem{Added the @racket[username] and @racket[password] arguments.}]}
+
+@deftogether[(@defparam[current-git-username username (or/c string? #f)]
+              @defparam[current-git-password password (or/c string? #f)])]{
+Parameters used by @racket[git-checkout] as the default values of the
+@racket[_username] and @racket[_password] arguments to control
+authentication with the remote server.
+
+@history[#:added "6.6.0.5"]}
--- a/pkgs/net-doc/net/scribblings/git-checkout.scrbl
+++ b/pkgs/net-doc/net/scribblings/git-checkout.scrbl
@@ -106,12 +106,19 @@ Authentication.
 @history[#:added "6.1.1.1"
          #:changed "6.3" @elem{Added the @racket[initial-error] argument.}
          #:changed "6.2.900.17" @elem{Added the @racket[strict-links?] argument.}
-         #:changed "6.6.0.5" @elem{Added the @racket[username] and @racket[password] arguments.}]}
+         #:changed "6.6.0.5" @elem{Added the @racket[username] and @racket[password] arguments.}
+         #:changed "6.6.0.5" @elem{Changed to raise @racket[exn:fail:git] exceptions
+                                   instead of @racket[exn:fail].}]}
 
 @deftogether[(@defparam[current-git-username username (or/c string? #f)]
               @defparam[current-git-password password (or/c string? #f)])]{
 Parameters used by @racket[git-checkout] as the default values of the
 @racket[_username] and @racket[_password] arguments to control
 authentication with the remote server.
+
+@history[#:added "6.6.0.5"]}
+
+@defstruct[(exn:fail:git exn:fail) () #:transparent]{
+Raised by @racket[git-checkout] due to errors parsing or communicating with the git protocol.
 
 @history[#:added "6.6.0.5"]}
--- a/pkgs/racket-doc/pkg/scribblings/pkg.scrbl
+++ b/pkgs/racket-doc/pkg/scribblings/pkg.scrbl
@@ -944,6 +944,9 @@ for @nonterm{key}.
         documentation; an empty string, which is the default, disables
         the URL so that the local filesystem is used. This key can be
         set only in @exec{installation} scope.}
+  @item{@exec{git-checkout-credentials} --- A list of git credentials in the form
+        @nonterm{username}@litchar{:}@nonterm{password} that are tried when downloading
+        packages with git sources using the HTTP or HTTPS protocols.}
   @item{@exec{trash-max-packages} --- A limit on the number of package implementations
         that are kept in a trash folder when the package is removed or updated.}
   @item{@exec{trash-max-seconds} --- A limit on the time since a package is removed or
@@ -955,7 +958,8 @@ for @nonterm{key}.
  ]
 
 @history[#:changed "6.1.1.6" @elem{Added @exec{trash-max-packages} and @exec{trash-max-seconds}.}
-         #:changed "6.3" @elem{Added @exec{network-retries}.}]}
+         #:changed "6.3" @elem{Added @exec{network-retries}.}
+         #:changed "6.6.0.5" @elem{Added @exec{git-checkout-credentials}.}]}
 
 
 @subcommand{@command/toc{catalog-show} @nonterm{option} ... @nonterm{package-name} ...
--- a/pkgs/racket-doc/pkg/scribblings/pkg.scrbl
+++ b/pkgs/racket-doc/pkg/scribblings/pkg.scrbl
@@ -946,7 +946,8 @@ for @nonterm{key}.
         set only in @exec{installation} scope.}
   @item{@exec{git-checkout-credentials} --- A list of git credentials in the form
         @nonterm{username}@litchar{:}@nonterm{password} that are tried when downloading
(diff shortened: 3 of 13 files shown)


racket/collects/pkg/private/checkout-credentials.rkt
```diff
@@ -0,0 +1,24 @@
...
+         "config.rkt")
+
+(provide call-with-git-checkout-credentials)
+
+(define (call-with-git-checkout-credentials thunk)
+  (let loop ([credentials-list (cons #f (get-git-checkout-credentials))])
+    (define credentials (first credentials-list))
+    (with-handlers ([exn:fail:git? (λ (exn)
+                                     (if (empty? (rest credentials-list))
+                                         (raise exn)
```
lexi-lambda: Hmm, that makes sense, but it would be a really bad idea to print passwords in an error message. Maybe it should just print the usernames?

bennn: Oh right. Yeah. Something like "failed to log in with usernames ...., last exception was: ...."

jeapostrophe:
I think we should merge this once you put in a test.

lexi-lambda:
@jeapostrophe What kind of test do you think would be most useful? It seems like it would be difficult to test the actual authentication mechanism without actually running a git server that requires authentication.

jeapostrophe:
You could use pltgit. You could make a burner github account with an app-specific password. Clearly it would be good to have tests, right?

lexi-lambda:
Yes, absolutely, but I don’t want core Racket tests to depend on some piece of *my* configuration that only I control. If you can give me instructions on how to test this against some PLT-controlled service, though, I’d be happy to write the relevant tests.

jeapostrophe:
I think the problem is that our pltgit only supports SSH keys :/

Can you make a little fake shim server that looks for a correctly formatted request appear?

lexi-lambda:
Possibly, yes. It looks like [`git http-backend`](https://git-scm.com/docs/git-http-backend) implements the HTTP server as a CGI program, and it would be possible to wrap that with something that implements HTTP basic auth. Is it okay for a test to depend on `git` being on the system? And is it possible to use the web server in the tests, or would I have to implement it using only what’s in `base`?

jeapostrophe:
On Wed, Oct 5, 2016 at 3:57 PM, Alexis King <notifications@github.com>
wrote:

> Possibly, yes. It looks like git http-backend
> <https://git-scm.com/docs/git-http-backend> implements the HTTP server as
> a CGI program, and it would be possible to wrap that with something that
> implements HTTP basic auth. Is it okay for a test to depend on git being
> on the system?
>
Yes

> And is it possible to use the web server in the tests, or would I have to
> implement it using only what’s in base?
>
You can use the Web Server

Jay

-- 
Jay McCarthy
Associate Professor
PLT @ CS @ UMass Lowell
http://jeapostrophe.github.io

                          - D&C 64:33

lexi-lambda:
@jeapostrophe: I pushed a commit that adds a test using a proxy server, which is in `pkgs/racket-test/tests/pkg/git-http-proxy.rkt`. It still uses a predefined port, though, like the other services in `pkgs/racket-test/tests/pkg/util.rkt`. If you have a suggestion for a way to use an arbitrary port without a *lot* of restructuring, I’d be open to it, but this seems to work alright for now.

jeapostrophe:
It looks great. We should merge.

elibarzilay:
1. It looks like this catches any `exn:fail:git` and retries with the next set of credentials, but that exception can be raised for other failures too?  So if there's some random error and I have three sets of credentials it'll re-try stuff four time --- and if some error happens after a minute's timeout I'll be waiting four minutes...

2. Credentials should allow forms other than username/password since there are other ways of authenticating that other people will want.

3. IMO, any piece of software that deals with passwords should be extra noisily verbose about storing passwords, even more if they're unencrypted.

4. Encrypting them *does* have an advantage: pulling the password out is easy if you know what you're doing, but in most cases you're a hacker harvesting strings for passwords and you know nothing about racket.

5. Finally, something like this should really have some way to specify credentials to specific urls (or url templates/regexps/whatever).  There are systems that will happily blacklist your IP or username (if multiple credentials have the same username) when they see more than N authentication failures in a short period.  We had this issue with the PLT git server blacklisting people who had more than 3 identity files specified that would be tried for all hosts, resulting in being blacklisted before getting to the correct key.  (And I see companies that *do* have such restrictions, BTW.)

stamourv:
@elibarzilay: Aside from maybe the encryption issue, it looks like all your comments could be addressed later without breaking backwards compatibility (and maybe the encryption issue too).
Given that @lexi-lambda would like this change to be included in the release (and branch day is tomorrow (which is not the last chance to merge, but I digress)), I'd be in favor of going forward with the merge now, and improving the feature later.

rfindler:
@stamourv: bullet point 1 seems likely to be a trivial fix and worth doing today?

stamourv:
Could be. I'd leave it up to @lexi-lambda. She's offering us a very nice contribution already. If she wants to do more work before merging, that's nice, but if not that's ok too.

rfindler:
Sure, of course!

lexi-lambda:
> It looks like this catches any exn:fail:git and retries with the next set of credentials, but that exception can be raised for other failures too? So if there's some random error and I have three sets of credentials it'll re-try stuff four time --- and if some error happens after a minute's timeout I'll be waiting four minutes...

The tricky thing here is that git does not really specify authentication very closely—it clearly permits using basic auth, but it doesn’t describe what responses should be or what should happen when things go wrong. Since the smart HTTP protocol makes a number of requests, it’d probably require more thought to figure out how to more gracefully treat authentication failures (that is, it should look for status codes and headers, such as `WWW-Authenticate`, which the current code doesn’t do at all). For that reason, it wouldn’t be trivial, but it’s certainly doable.

> Credentials should allow forms other than username/password since there are other ways of authenticating that other people will want.

Probably, yeah, but I think those are outside the scope of this pull request, and nothing in these changes should preclude adding more authentication schemes later.

> IMO, any piece of software that deals with passwords should be extra noisily verbose about storing passwords, even more if they're unencrypted.

Yeah. I am uncomfortably aware that the way this code handles passwords is terrible, perhaps bordering on irresponsible. I am no security expert, though, and I’m not sure what the right thing to do is here. For GitHub, at least, you can use personal auth tokens as “passwords”, so you’re not storing your real password as plaintext, but that’s a totally GH-specific thing.

Is there any facility in the Racket core that could encrypt/decrypt the passwords? Also, separately, what do you consider “noisily verbose” in this particular context? I know the `raco pkg config` interface isn’t very good for this, and I’d like to improve it, but I really don’t have the time to overhaul that right now.

> Finally, something like this should really have some way to specify credentials to specific urls (or url templates/regexps/whatever). There are systems that will happily blacklist your IP or username (if multiple credentials have the same username) when they see more than N authentication failures in a short period. We had this issue with the PLT git server blacklisting people who had more than 3 identity files specified that would be tried for all hosts, resulting in being blacklisted before getting to the correct key.

If this becomes a problem for people, it sounds like a good idea to implement it. One thing I’d definitely *like* to do, at the very least, is do some caching about which set of credentials succeeded for a particular package to be a little more intelligent when trying to clone the repository next time, but I definitely think that’s something that can be added later. In the meantime, given that this won’t impact people at all unless they explicitly opt-in to this feature, I’m not sure it’s worth the effort to do up front (and I *do* think it would be a lot of effort, not because it’s technically complicated, but because the `raco pkg config` command-line interface is not very sophisticated, and providing a good UI for these kinds of options would require making it a lot more capable than it is now).

elibarzilay:
> The tricky thing here is that git does not really specify authentication very closely—it clearly permits using basic auth, but it doesn’t describe what responses should be or what should happen when things go wrong. Since the smart HTTP protocol makes a number of requests, it’d probably require more thought to figure out how to more gracefully treat authentication failures (that is, it should look for status codes and headers, such as `WWW-Authenticate`, which the current code doesn’t do at all). For that reason, it wouldn’t be trivial, but it’s certainly doable.

I would assume that the authentication failures would be covered by standard HTTP response codes (maybe even done completely by the web server in some cases), and therefore would be easy to catch.

> > Credentials should allow forms other than username/password since there are other ways of authenticating that other people will want.
> 
> Probably, yeah, but I think those are outside the scope of this pull request, and nothing in these changes should preclude adding more authentication schemes later.

The only thing is that you're saving a list of username/passwords, so there's no real place to extend it.  What I meant in the above is that it should be wrapped by something that indicates the contents (eg cons a 'basic in front), and a simple check for that (check that 'basic is there), and that's enough to have different things in the future.

> Is there any facility in the Racket core that could encrypt/decrypt the passwords? Also, separately, what do you consider “noisily verbose” in this particular context? I know the `raco pkg config` interface isn’t very good for this, and I’d like to improve it, but I really don’t have the time to overhaul that right now.

There might be something in ssl, but the main thing is some visible warning, in the docs & command line, wherever possible.  If the cli interface doesn't make it possible to add printouts, then maybe the description could include "(UNENCRYPTED)", or something...

> If this becomes a problem for people, it sounds like a good idea to implement it. One thing I’d definitely *like* to do, at the very least, is do some caching about which set of credentials succeeded for a particular package to be a little more intelligent when trying to clone the repository next time, but I definitely think that’s something that can be added later. In the meantime, given that this won’t impact people at all unless they explicitly opt-in to this feature, I’m not sure it’s worth the effort to do up front (and I *do* think it would be a lot of effort, not because it’s technically complicated, but because the `raco pkg config` command-line interface is not very sophisticated, and providing a good UI for these kinds of options would require making it a lot more capable than it is now).

Same as above: just a small hook to make it possible.

lexi-lambda:
I’m going to merge this because, frankly, I’d like to have this functionality, and I don’t have the time to implement anything significantly better right now. Hopefully it shouldn’t be too hard to revisit this later if people need stronger guarantees or more features.

stamourv:
Thanks for contributing!

elibarzilay:
@lexi-lambda: if you pushed it as-is, then you *are* making it unnecessarily hard to revisit for more features etc.  Just cons a symbol in the front, and you'll make it easy again.

This is not an imaginary problem, BTW.  The handin server still uses MD5 checksum -- it's pretty hard to update in a backward compatible way because there is no place for a type or version specification.  Another example of a similar problem: look at a linux password file (usually `/etc/shadow`) and you'll see passwords that start with things like `$n$` -- that's a hack since there was no other easy way to enhance the file format without breaking compatibility.

lexi-lambda:
I pushed what I had at the suggestion of three other maintainers, not least in part because I simply do not have more time to dedicate to this right now. If you’d like to revert it, I will not complain, but unless someone else is willing to take this on, it seems unlikely it will get done at all.

There is an argument to be made that doing things right eventually is better than doing things wrong now, but I am willing to defer to the judgement of others here.

elibarzilay:
@lexi-lambda: I'm not going to revert it.  To make it clear, all I'm suggesting is that you cons up one symbol before the username/password list.  It's a tiny ~1min change that would go a long way to make it easy to extend in the future, since it'll be possible to make the current code deal with configurations that start with that symbol.

stamourv:
@elibarzilay: If the change is this simple (which I can believe), please point me to the relevant parts of the code, and I'll make the changes.
Asking @lexi-lambda to do more work after she's already contributed a valuable improvement is not the right things to do here.

elibarzilay:
@stamourv, I'm talking about a very minimal change that will avoid an ugly hack when it's extended in the future.

Looking through the diff, it looks like only two files are relevant:

1. In `pkgs/racket-doc/pkg/scribblings/pkg.scrbl`:
   ```
     @item{@exec{git-checkout-credentials} --- ...
   ```
   Update and say that it starts with a format specification datum which is currently only `'basic`.

2. In `racket/collects/pkg/private/config.rkt`:
   ```
     (or (current-pkg-git-checkout-credentials)
         (read-pkg-cfg/def 'git-checkout-credentials)))
   ```
   Minimal: add a `cdr` around the whole thing.  Optional: throw an error if it's not 'basic.

   (It *looks* like `current-pkg-git-checkout-credentials` is intended to fake a setup, but I'm not sure since I don't see it used anywhere in the diff.)

   In the definition of `get-default`:
   ```
         ['git-checkout-credentials '()]
   ```
   Change the value to `'(basic)`.

   In the definition of `pkg-config`:
   ```
           (update-pkg-cfg! 'git-checkout-credentials
                            (for/list ([val (in-list vals)])
   ```
   wrap a `(cons 'basic ...)` around the `for/list`

   A bit later in `pkg-config`:
   ```
                (for ([creds (in-list (read-pkg-cfg/def 'git-checkout-credentials))])
   ```
   Add a `cdr` around the result of `read-pkg-cfg/def`

stamourv:
I've pushed the changes. Thanks @eli!

elibarzilay:
@stamourv -- thanks!

PR #1473, opened 2016-10-01
Have id-table-ref! call its failure thunks.

stamourv:
Closes 15346.

This is not bugwards compatible, and could conceivably break some programs.
But, the fixed behavior is consistent with the docs, as well as with all similar operations elsewhere in Racket.


stamourv:
Merged.

PR #1474, opened 2016-10-01
doc: sequence/c elements vs. values

bennn:
Clarify that `sequence/c` works for sequences with arbitrarily many elements.

The current wording made me think `(list 1 2)` would not satisfy `(sequence/c integer?)`.


bennn:
65a69417cc4d1bc1d5c8cfef2b3ab6bb830852d4

Issue #1239, opened 2016-02-12
"error: 'JIT_LOCAL4' undeclared" on OpenBSD/powerpc


juanfra684:
When I try to build Racket 6.4 on OpenBSD/powerpc with GCC 4.9.3 and jit/places/futures enabled, it shows this error during the build:

```
cc -O2 -pipe    -I./.. -I/usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/../include -I/usr/local/include -I/usr/X11R6/include -pthread -I/usr/local/include  -DUSE_SENORA_GC  -c /usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/hash.c -o hash.o
cc -O2 -pipe    -I./.. -I/usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/../include -I/usr/local/include -I/usr/X11R6/include -pthread -I/usr/local/include  -DUSE_SENORA_GC  -c /usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/jit.c -o jit.o
/usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/jit.c: In function 'scheme_generate':
/usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/jit.c:2469:32: error: 'JIT_LOCAL4' undeclared (first use in this function)
           mz_ld_runstack_base_alt(JIT_RUNSTACK);
                                ^
/usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/jit.c:2469:32: note: each undeclared identifier is reported only once for each function it appears in
/usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/jit.c: In function 'generate_function_getarg':
/usr/ports/pobj/racket-minimal-6.4/racket-6.4/src/racket/src/jit.c:3465:26: error: 'JIT_LOCAL4' undeclared (first use in this function)
     mz_set_local_p(JIT_RUNSTACK, JIT_RUNSTACK_BASE_LOCAL);
                          ^
*** Error 1 in racket/src (Makefile:217 'jit.o')
*** Error 1 in racket/src (Makefile:136 'all')
```

Unlike the default compiler on OpenBSD (GCC 4.2), GCC 4.9.3 supports thread-local storage (emulated).

juanfra684:
[...]

PR #1477, opened 2016-10-02
Add support for PowerPC on FreeBSD

### LINKED ISSUE #1471: Racket on PowerPC (FreeBSD)?
Is there a reason why Racket currently doesn't support the PowerPC architecture on FreeBSD? 
I noticed that PowerPC is supported on other operating systems. If it's simply a matter of configuration I can also try to fix it myself.

I know that FreeBSD doesn't have that many users so I understand if porting is not a priority.

alehed:
I'm not sure if this is all that is needed, but it compiles now with freebsd on powerpc.
Closes #1471

### DIFF
--- a/racket/src/racket/sconfig.h
+++ b/racket/src/racket/sconfig.h
@@ -477,6 +477,7 @@
 #  define SCHEME_PLATFORM_LIBRARY_SUBPATH "arm-freebsd"
 # elif defined(__powerpc__)
 #  define SCHEME_PLATFORM_LIBRARY_SUBPATH "ppc-freebsd"
+#  define MZ_USE_JIT_PPC
 # else
 #  error Unported platform.
 # endif
--- a/racket/src/racket/sconfig.h
+++ b/racket/src/racket/sconfig.h
@@ -475,6 +475,8 @@
 #  define FLUSH_SPARC_REGISTER_WINDOWS
 # elif defined(__arm__)
 #  define SCHEME_PLATFORM_LIBRARY_SUBPATH "arm-freebsd"
+# elif defined(__powerpc__)
+#  define SCHEME_PLATFORM_LIBRARY_SUBPATH "ppc-freebsd"
 # else
 #  error Unported platform.
 # endif


alehed:
Why is the build failing?

juanfra684:
Probably, you want also to add the line: `#define MZ_USE_JIT_PPC`.

Did you try to run DrRacket? Everything works fine?.

If possible, can you run the tests and to copy/paste the output?:
```
$ racket -e '(eval-jit-enabled)'
$ raco pkg install --auto -u racket-test-core
$ cd ~/.racket/$(racket -e '(display (version))')/pkgs/racket-test-core/tests/racket/
$ racket -f quiet.rktl
```

alehed:
Compiling it with `#define MZ_USE_JIT_PPC` gives me the error

```
./../src/jit.c: In function 'scheme_generate':
./../src/jit.c2473: error: 'JIT_LOCAL4' undeclared (first use of this function)
```

alehed:
Without jit enabled the build works. Installing packages works but it complains about not being able to install some documentation. 

The output of `racket -f quiet.rktl` is:
```
Section(basic)


TIMEOUT -- ABORTING!
```

juanfra684:
Build racket with `--disable-places --disable-futures`.

alehed:
Should I disable places and futures by default for the powerpc architecture?

stamourv:
FWIW, those at least used to work on PPC on Mac OS. I last used that combination ~5 years ago, though, and I think we got rid of those build machines some time ago, so it may have bit-rotten.

alehed:
So I'll just disable them in my build script not in the racket configuration file?

juanfra684:
@alehed yes, both are not supported on powerpc. With both disabled in the configure script, you can build the interpreter with JIT.

@stamourv look #1239.

stamourv:
Ah, looks like I was out of date then. :)

alehed:
I cloned racket, changed freebsd on ppc in the sconfig to contain:
```
#define SCHEME_PLATFORM_LIBRARY_SUBPATH "ppc-freebsd"
#define MZ_USE_JIT_PPC
```
And built it with `make`. I get the error: `http-client: connection ended early`. Any thoughts on this?
Note that by default both futures and places are disabled on ppc.

alehed:
Building it with the ports system (minima-racket) with the above sconfig and without places or futures I get an error in racket3m:
```
/usr/ports/lang/racket-minimal/work/racket-6.6/src/racket/.libs/racket3m: Undefined symbol "GC_variable_stack"
```

alehed:
I finally got a build working with JIT enabled (I didn't give it any JIT flags like msse, msse2 or mfpmath=sse). The output of eval-jit-enabled ist true. 
I have to say, my racket interpreter is pretty slow. When I execute the given test, the output is still the same as when I ran it the first time (see above).
BTW: I installed racket-minimal and I'm running freebsd without a window system so I don't know if DrRacket would work or not.

alehed:
I was also mistaken about the default build. On PowerPC you specifically have to specifically disable places and futures, otherwise it doesn't work.

juanfra684:
The reason why I asked for the tests is because racket crashes on OpenBSD/macppc. I installed FreeBSD 10.2 and I see the same crash in the same test (`racket -f unsafe.rktl`). Something is broken in the powerpc JIT or GC code (the same test works fine without the JIT but other different tests fail).

That said, I think that the PR is correct and someone should commit the changes.

alehed:
Yes, for me it also crashed on that test. 

Btw: Is it normal that a simple hello world in racket takes about 2s on an old computer (~700 according to geekbench)?

juanfra684:
Try this: `echo '#lang racket/base\n(displayln "hello world")' > hello.rkt && raco make hello.rkt`.  Then run `racket hello.rkt`, it will avoid the slow cold start. BSDs on powerpc are usually slow, I don't know how Apple made OSX fast on these machines.

rfindler:
If you run `raco make -v hello.rkt` (adding the `-v` flag) and you see lots and lots of output, that means that probably you missed a step in the build process. (You would see all this output only the first time you run the command; but you know it was there if the command gets a lot faster the second time.)

alehed:
Compiling reduces the two seconds to about one and a half.

On `raco make -v` the output shows nothing unexpected. On `raco make --vv` I get almost 300 lines of `checking: ...` but also that seems quite expected.

But to be honest, it's not such a big deal for me. As long as it works, I'm happy.

PR #1487, opened 2016-10-10
Add #:logger keyword argument to with-intercepted-logging.

### LINKED ISSUE #1486: with-intercepted-logging only intercepts logs sent to current-logger
Because of [L45](https://github.com/racket/racket/blob/master/racket/collects/racket/logging.rkt#L45) in `racket/logging.rkt`, the function `with-intercepted-logging` can only intercept logs sent to the `(current-logger)` parameter.

And so the following program (adapted from the [docs](http://docs.racket-lang.org/reference/logging.html#%28def._%28%28lib._racket%2Flogging..rkt%29._with-intercepted-logging%29%29)) returns 0, but I was expecting 2.

```
#lang racket
(require racket/logging)

(let ([warning-counter 0]
      [l (current-logger)])
 (with-intercepted-logging
  (lambda (l)
   (when (eq? (vector-ref l 0)
          'warning)
    (set! warning-counter (add1 warning-counter))))
  (lambda ()
   (log-message l 'warning "Warning!" (current-continuation-marks))
   (log-message l 'warning "Warning again!" (current-continuation-marks))
   (+ 2 2))
  'warning)
 warning-counter)
;; 0
```

Also, this means `with-intercepted-logging` won't intercept any logs from functions created by `define-logger`.

Is it possible to change the implementation of `with-intercepted-logging` to intercept any log event that would be received by a `(make-log-receiver (current-logger) ....)` declared _outside_ the dynamic extent of `proc`?

stamourv:
Closes #1486.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/logging.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/logging.scrbl
@@ -351,13 +351,17 @@ Returns @racket[#t] if @racket[v] is a valid logging level (@racket['none],
                               (or/c symbol? #f))
                              any)]
            [proc (-> any)]
+           [#:logger logger logger? #f]
            [level log-level/c]
            [topic (or/c #f symbol?) #f]
            ... ...)
          any]{
 
-Runs @racket[proc], calling @racket[interceptor] on any log event that would
-be received by @racket[(make-log-receiver (current-logger) level topic ... ...)].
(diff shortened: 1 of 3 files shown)


stamourv:
@bennn: If you want to give it a shot.

stamourv:
Oh, good. I wasn't sure about the docs.
Is it ok if this doesn't go into the release?

bennn:
Yeah that's fine, no rush.

PR #1489, opened 2016-10-13
passing get-timestamp rather than sys-type to write-central-directory (zip.rkt)

cjallen88:
It's worth noting that this hasn't caused me an issue, I came across it as I wanted to see what sys-type actually did. I couldn't say what the effect of this change would be and don't have a use case for it, it just looks wrong!


mflatt:
Thanks!

PR #1490, opened 2016-10-14
Some small fixes to the documentation

jsmaniac:
* Wrong contract for syntax-local-value in the documentation.
* Clarified signature in documentation for expand-import, expand-export and pre-expand-export
* Corrected typo in documentation for "for".
* Fixed error message for function which seems to have been renamed in the docs
* Fixed typo in a comment in the tests
* Fixed a typo in the documentation for set-subtract.
* Use double ellipses for the free-id-table-set*, free-id-table-set*!, bound-id-table-set* and bound-id-table-set*! operations

This commit is a squashed rebase of several commits, if you prefer the individual commits, see https://github.com/jsmaniac/racket/tree/doc-changes-2 which contains exactly the same changes.

@stamourv I think this can safely be merged in the 6.7 `release` branch, since it only contains changes to the documentation, to a comment and to an error message which is inside an example in the docs.


stamourv:
Merged. Will look into getting it in the release.

PR #1499, opened 2016-10-28
faster in-*-id-table

pnwamk:
I put together a more efficient version of the `in-*-id-table` forms using `define-sequence-syntax`. (in a benchmark I was tinkering with this version was 6x faster than the previous definition)

I haven't done thorough testing yet (is there a test suite sitting right under my nose somewhere I'm not seeing?), but anecdotally it seems to work great! =)

Comments? (cc @rmculpepper)

Also, there are several other functions which iterate over hash tables. If this definition is ironed out and still significantly faster, would it make sense to rewrite some of those as simple `for` loops using this iteration mechanism?

### DIFF
--- a/pkgs/racket-test-core/tests/racket/id-table-test.rktl
+++ b/pkgs/racket-test-core/tests/racket/id-table-test.rktl
@@ -47,7 +47,10 @@
     (test 2 bound-id-table-ref d3 a)
     (test 3 bound-id-table-ref d3 b)
     (test 4 bound-id-table-ref d3 b2)
-    (test 5 bound-id-table-ref d3 b3))
+    (test 5 bound-id-table-ref d3 b3)
+    (test 4 sequence-length (in-bound-id-table d1))
+    (test (for/list ([(k v) (in-bound-id-table d1)]) (cons k v))
+          (λ () (for/list ([(k v) (values (in-bound-id-table d1))]) (cons k v)))) )
 
   (let ()
     ;; Test in-dict, iteration methods for mutable id-tables
--- a/racket/collects/syntax/id-table.rkt
+++ b/racket/collects/syntax/id-table.rkt
@@ -341,8 +341,6 @@
              (-> idtbl? (listof identifier?))]
             [idtbl-values
              (-> idtbl? list?)]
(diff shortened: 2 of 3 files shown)


samth:
Does the relocate function get used when there's concurrent mutation?

pnwamk:
Yes, relocate is an attempt to mimic how @rmculpepper handled what happened if the bucket in the underlying mutable hash table changes between iterations within that bucket.

rmculpepper:
First, you should keep the old implementation around so the new one can fall back to it. The implementation of `in-idtbl/proc` is wrong; it produces a list (a sequence that yields one value per step) rather than a sequence that yields two values per step. You might want to add tests of the following forms to the existing tests in `pkgs/racket-test-core/tests/racket/id-table-test.rkt`:

    (sequence->length (in-free-id-table _)) = _
    (for ([(k v) (in-free-id-table _)]) _) = (for ([(k v) (values (in-free-id-table _))]) _)

Second, avoid macros that expand into (big) macro definitions like `define-in-tbl`. I recommend defining a compile-time transformer-producing function first:

    (begin-for-syntax
      ;; make-in-table-transformer : Indentifier -> Syntax -> Syntax/#f
      (define ((make-in-table-transformer pred?-id) stx)
        (syntax-case stx ()
          [[(key val) (_ table)]
           (with-syntax ([pred? pred?-id])
             #'___)]
          [_ #f])))

and then calling it inside of the `make-code` template:

    (define (in-tbl* d)
      (in-id-table 'in-idtbl d identifier->symbol identifier=?))
    (define-sequence-syntax in-tbl
      (lambda () #'in-tbl*)
      (make-in-table-transformer #'tbl?)) ;; might need other args too

Yes, redefining the existing functions like `idtbl-map` and `idtbl-for-each` to use `in-idtbl` sounds like a good idea. I would avoid adding new higher-order looping functions if the new `in-idbtl` sequence is fast enough to cover their use cases.

pnwamk:
I believe I made all the suggested changes -- happy to tweak things further if needs be.

rmculpepper:
Looks good to me. Thanks!

PR #1505, opened 2016-10-31
Track the origin of modules produced by module+ forms

lexi-lambda:
Since `module+` creates modules via `syntax-local-lift-module-end-declaration`, the syntax that is eventually produced does not contain information about the forms that produced them. This means that Check Syntax does not draw arrows to `module+` identifiers, and it also won’t draw arrows to macro invocations that _expand_ to `module+` forms.

These changes ensure the `'origin` property is propagated by manually calling `syntax-track-origin` using each of the `module+` forms that contribute to the resulting module.


lexi-lambda:
Gentle ping on this, given that I think this change is relatively harmless. @mflatt, you’re probably the person most qualified to look at this, but I’m sure @jeapostrophe or @rmculpepper would be more than capable, too.

jeapostrophe:
It looks good to me.

PR #1510, opened 2016-11-08
add more plugins to emacs section of guide

willghatch:
Due to the recent mailing list conversation about emacs plugins for working with Racket, I thought I would add a little to the guide section about it.  In particular this adds some packages that I find useful for evil-mode users.  Perhaps there should be a separate bullet list of packages that are useful specifically for evil-mode users?

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/other-editors.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/other-editors.scrbl
@@ -85,6 +85,10 @@ popular among Racketeers as well.
        Debian and Ubuntu packages for Paredit are available under the
        name @tt{paredit-el}.}
 
+ @item{@hyperlink["https://github.com/Fuco1/smartparens"]{Smartparens}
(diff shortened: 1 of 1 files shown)


stamourv:
Would it make sense to combine those into one `evil-racket-support` emacs package or something? Otherwise, it's kind of a scattering of stuff.

willghatch:
Perhaps.  I can see the appeal of keeping a short list.  But at the same time I'm generally not keen on emacs packages that just configure some other packages.  Personally I have some extra configuration for each of these packages, and any configuration in an `evil-racket-support` package would just be something I would be fighting.  I think it would be useful to have just a long list of editor plugins/configuration/etc that people find useful, but maybe the guide could just point to it.  Maybe somewhere in a wiki.

stamourv:
How about a general `evil-mode` bullet? Puts everything related to it in the same place, and easy to skip for people who don't use it.

willghatch:
Agreed.  How about this update?

stamourv:
LGTM

PR #1517, opened 2016-11-15
Extend reductions for expressions like (let ([x (let ([y M]) N)]) P) 

gus-massa:
The expressions like `(let ([x (let ([y M]) N)]) P)` are reduced to `(let ([y M]) (let ([x N]) P))`.

Extend this reduction to expressions where the outer `let` has more than one clause, for example

    (let ([x (let ([y M]) N)]
          [z _])
      P)

It's only one character change :), I hope I'm not missing a corner case.

The second commit is straightforward, and I plan to squash it and piggyback it with the other commit. I think that this additional argument was necessary with the old coordinate system.


mflatt:
Looks fine. I imagine the constraint to a single clause was also related to the old coordinate system, and the constraint doesn't seem necessary now.

gus-massa:
I think so. This change was was almost impossible with the old coordinate system and totaly trivial now.

PR #1518, opened 2016-11-15
Extend reductions for expressions like (if (if X Y #f) Z K)

gus-massa:
The expresions like `(if (if X Y #f) Z K)` are reduced to `(if X (if Y Z K) K)`,  when `K` is a constant,

Extend this to other similar expressions where the inner if is 

    (if X Y #f) ----> [1] (current)
    (if X #f Y) ----> [2] 
    (if X Y #t) ----> [3] 
    (if X #t Y) ----> [4]

I used these number in a preliminay test to see if the other versions appear in real code. `[1]` is the most popular, but `[4]` also appears many times`. https://travis-ci.org/gus-massa/racket/jobs/175386247#L625


mflatt:
LGTM

PR #1519, opened 2016-11-15
Remove argument of finish_optimize_application

gus-massa:
The value of `rator_flags` was calculated in `optimize_application` and used in `finish_optimize_application`, but it is possible to calculate it directly in `finish_optimize_application`, and then remove some internal coupling.

This also simplifies other locations where the `rator ` of an application was changed and then it was necessary to recalculate the value of `rator_flags` to complete the optimization steps.


mflatt:
LGTM

PR #1521, opened 2016-11-18
Add missing properties to contract chaperone.

stamourv:
@robby: Does this look good to you?


stamourv:
Oops, I meant @rfindler

rfindler:
I no longer remember this code. But if you add a test and it passes, then sure.

stamourv:
Added a test and merged.

PR #1524, opened 2016-11-20
Avoid traversing large vectors in contract checking.

samth:
This is a 2x improvement in performance for configuration 010001 of the
suffixtree benchmark from Takikawa et al, POPL 2016.

This change has two potential drawbacks:

* Some contract errors no longer happen at all. Some contract
  errors are now delayed to when the element is accessed, rather
  than when the contract is applied. This seems unlikely to
  break existing programs, but it's possible.

* The "less than 10" rule is ad-hoc and might be confusing.

@rfindler, I'd appreciate your thoughts on whether this is
a good idea.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -420,7 +420,8 @@ passes its arguments to @racket[or/c].
 
 @defproc[(vectorof [c contract?]
                    [#:immutable immutable (or/c #t #f 'dont-care) 'dont-care]
-                   [#:flat? flat? boolean? #f])
+                   [#:flat? flat? boolean? #f]
+                   [#:eager eager (or/c #t #f exact-nonnegative-integer?) #t])
          contract?]{
 Returns a @tech{contract} that recognizes vectors. The elements of the vector must
 match @racket[c].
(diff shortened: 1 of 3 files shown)


rfindler:
It doesn't seem like a good idea to me, but then again, neither does the
original design (at lease ever since we got chaperones) for these
combinators. Maybe we should think about what a better design is instead
and promote that?

Meanwhile, I don't mind a keyword argument that causes this functionality
to kick in when asked for. Would that work for now?

samth:
That's certainly reasonable, and would be easy for TR to generate. I'd rather think harder about a better design, though, and generate that.

We could go with an always-lazy design, similar to how mutable vectors work. Unfortunately, I think this slows down some other programs (I've seen this on the `synth` benchmark from the same paper, which uses lots and lots of length-1 immutable vectors).

We could have a keyword argument for how many elements to check eagerly. Then TR and other users of the contract system could choose, but that seems like a choice that we as authors of the contract system could make more effectively.

Other suggestions we might want to think about? I'm happy to implement whatever we decide here.

rfindler:
I'm not sure the contract system is always the best place to make this
guess​. I think that, in general, the client knows more than the contract
system, so it does make sense for there to be a way to let the clients make
the decision.

I also feel like a few more keyword arguments are a good way to go here.

How about a keyword argument that says "copy only if larger than N
elements" where, if N is supplied as a nat, that is taken, but N can also
be supplied as a symbol that is something like "you make a guess, contract
system" and we write that that's currently 10, but we make up a micro
benchmark that we maybe reevaluate periodically to choose a different
number? (And document the benchmark too.) We could also allow +inf.0 as the
argument, which means to never copy? Maybe that will unify some of the
other behavior internally?

samth:
@rfindler I've now re-implemented this along the lines you suggested (but without the "contract system picks something" option so far).

rfindler:
LGTM

PR #1527, opened 2016-11-28
replace references to "[Mac] OS X" with "macOS"

mbutterick:
Because what’s old is new again.

### DIFF
--- a/pkgs/net-doc/net/scribblings/osx-ssl.scrbl
+++ b/pkgs/net-doc/net/scribblings/osx-ssl.scrbl
@@ -3,11 +3,11 @@
           (for-label net/osx-ssl
                      openssl))
 
-@title[#:tag "osx-ssl"]{OS X Native SSL: Secure Communication}
+@title[#:tag "osx-ssl"]{macOS Native SSL: Secure Communication}
 
 @defmodule[net/osx-ssl]{The @racketmodname[net/osx-ssl] module
 offers a fraction of the functionality of @racketmodname[openssl] and
-works only on OS X, but it has the advantage that it works before
+works only on macOS, but it has the advantage that it works before
 OpenSSL libraries are installed.}
 
 @history[#:added "6.3.0.12"]
--- a/pkgs/net-doc/net/scribblings/osx-ssl.scrbl
+++ b/pkgs/net-doc/net/scribblings/osx-ssl.scrbl
@@ -3,7 +3,7 @@
           (for-label net/osx-ssl
                      openssl))
 
-@title[#:tag "osx-ssl"]{macOS Native SSL: Secure Communication}
+@title[#:tag "osx-ssl"]{OS X Native SSL: Secure Communication}
 
 @defmodule[net/osx-ssl]{The @racketmodname[net/osx-ssl] module
 offers a fraction of the functionality of @racketmodname[openssl] and
--- a/pkgs/racket-doc/openssl/openssl.scrbl
+++ b/pkgs/racket-doc/openssl/openssl.scrbl
@@ -28,10 +28,10 @@ or with the Racket distribution. In particular:
 @filepath{libeay32.dll} and @filepath{ssleay32.dll}, which are
 included in the Racket distribution for Windows.}
 
-@item{For Mac OS X, @racketmodname[openssl] depends on
+@item{For macOS, @racketmodname[openssl] depends on
 @filepath{libssl.dylib} and @filepath{libcrypto.dylib}. Although those
-libraries are provided by Mac OS X 10.2 and later, their use is
-deprecated, so the Racket distribution for Mac OS X includes newer
+libraries are provided by macOS 10.2 and later, their use is
+deprecated, so the Racket distribution for macOS includes newer
 versions.}
 
 @item{For Unix, @racketmodname[openssl] depends on
@@ -462,7 +462,7 @@ immediately. Only supported on Windows.}
 
 @item{If @racket[src] is @racket[(list 'macosx-keychain _path)], then
 the certificates from the keychain stored at @racket[_path] are loaded
-immediately. Only supported on Mac OS X.}
+immediately. Only supported on macOS.}
 
 ]
 
@@ -493,7 +493,7 @@ on the platform:
 @tt{SSL_CERT_FILE} and @tt{SSL_CERT_DIR} environment variables, if the
 variables are set, or the system-wide default locations otherwise.}
 
-@item{On Mac OS X, the default sources consist of the system keychain
+@item{On macOS, the default sources consist of the system keychain
 for root certificates: @racket['(macosx-keychain
 "/System/Library/Keychains/SystemRootCertificates.keychain")].}
 
--- a/pkgs/racket-doc/scribblings/guide/scripts.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/scripts.scrbl
@@ -4,12 +4,12 @@
 @title[#:tag "scripts"]{Scripts}
 
 Racket files can be turned into executable scripts on Unix and Mac
-OS X.  On Windows, a compatibility layer like Cygwin support the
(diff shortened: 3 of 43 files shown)


mbutterick:
PS. In `scrbl` files only, not `rkt` files.

pkgs/net-doc/net/scribblings/osx-ssl.scrbl
```diff
@@ -3,11 +3,11 @@
           (for-label net/osx-ssl
                      openssl))
 
-@title[#:tag "osx-ssl"]{OS X Native SSL: Secure Communication}
+@title[#:tag "osx-ssl"]{macOS Native SSL: Secure Communication}
```
stamourv: With this, someone searching for "OS X SSL" will not get this result, whereas they would now. Could you add the old title to the index separately?

pkgs/racket-doc/openssl/openssl.scrbl
```diff
@@ -28,10 +28,10 @@ or with the Racket distribution. In particular:
 @filepath{libeay32.dll} and @filepath{ssleay32.dll}, which are
 included in the Racket distribution for Windows.}
 
-@item{For Mac OS X, @racketmodname[openssl] depends on
+@item{For macOS, @racketmodname[openssl] depends on
 @filepath{libssl.dylib} and @filepath{libcrypto.dylib}. Although those
-libraries are provided by Mac OS X 10.2 and later, their use is
-deprecated, so the Racket distribution for Mac OS X includes newer
+libraries are provided by macOS 10.2 and later, their use is
```
stamourv: Nitpick: can one speak of macOS 10.2? 10.2 was before the name change.

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
```diff
@@ -115,7 +115,7 @@ by @racket[kind], which must be one of the following:
  environment variable or flag is specified, or if the value is not a
  legal path name, then this directory defaults to
  @filepath{Library/Racket} in the @tech{user's home directory} on Mac
- OS X and @racket['pref-dir] otherwise.  The directory might not
+ macOS and @racket['pref-dir] otherwise.  The directory might not
```
stamourv: "Mac macOS"

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
```diff
@@ -405,7 +405,7 @@ identity of the referenced file or directory (if any).}
 @defproc[(file-size [path path-string?]) exact-nonnegative-integer?]{
 
 Returns the (logical) size of the specified file in bytes. On Mac
-OS X, this size excludes the resource-fork size. On error (e.g., if no
+macOS, this size excludes the resource-fork size. On error (e.g., if no
```
stamourv: idem

pkgs/racket-doc/scribblings/reference/futures.scrbl
```diff
@@ -13,7 +13,7 @@
 @note-lib[racket/future]
 
 @margin-note{Currently, parallel support for @racket[future] is enabled
-  by default for Windows, Linux x86/x86_64, and Mac OS X x86/x86_64. To
+  by default for Windows, Linux x86/x86_64, and macmacOS86/x86_64. To
```
stamourv: idem

pkgs/racket-doc/scribblings/reference/places.scrbl
```diff
@@ -26,7 +26,7 @@ hardware threads.
 
 @margin-note{Currently, parallel support for places is enabled
   only for Racket 3m (which is the main variant of Racket), and only
-  by default for Windows, Linux x86/x86_64, and Mac OS X x86/x86_64. To
+  by default for Windows, Linux x86/x86_64, and macmacOS86/x86_64. To
```
stamourv: idem

pkgs/racket-doc/scribblings/reference/startup.scrbl
```diff
@@ -19,7 +19,7 @@ The core Racket run-time system is available in two main variants:
 
  @item{Racket, which provides the primitives libraries on which
        @racketmodname[racket/base] is implemented. On Unix and Mac
-       OS X, the executable is called
+       macOS, the executable is called
```
stamourv: idem

pkgs/racket-doc/scribblings/raco/launcher.scrbl
```diff
@@ -471,7 +471,7 @@ are as follows:
 @itemize[
 
  @item{@filepath{.icns} @'rarr @racket['icns] file for use on Mac
-       OS X}
+       macOS}
```
stamourv: "Mac macOS"

pkgs/racket-doc/scribblings/raco/launcher.scrbl
```diff
@@ -170,7 +170,7 @@ tethering.
          void?]{
 
 Like @racket[make-gracket-launcher], but for starting Racket. On Mac
-OS X, the @racket['exe-name] @racket[aux] association is ignored.}
+macOS, the @racket['exe-name] @racket[aux] association is ignored.}
```
stamourv: "Mac macOS"

pkgs/racket-doc/scribblings/guide/scripts.scrbl
```diff
@@ -4,12 +4,12 @@
 @title[#:tag "scripts"]{Scripts}
 
 Racket files can be turned into executable scripts on Unix and Mac
-OS X.  On Windows, a compatibility layer like Cygwin support the
+macOS.  On Windows, a compatibility layer like Cygwin support the
```
stamourv: "Mac macOS"

mflatt:
Even with repairs, I would like to consider this more before merging, because there must be other places (like ".txt" files) to make consistent.

mbutterick:
I'll make these edits (duh, I forgot about the possibility of hard line breaks) and also comb through the `txt` files.

pkgs/racket-doc/openssl/openssl.scrbl
mbutterick: Since the usage here is “10.2 __and later__”, I think it's better with “macOS”. If it were a specific reference to only the 10.2 release, then "OS X 10.2". Also, American corporations are the worst.

pkgs/net-doc/net/scribblings/osx-ssl.scrbl
mbutterick: I’ll revert this one because “OS X” is baked into the whole naming scheme of the package & its functions. The package maintainer should decide what happens.

mbutterick:
> because there must be other places (like ".txt" files) to make consistent.

I pushed an update for the main `INSTALL.txt` file. Otherwise AFAICT the references to "Mac OS X" in `txt` files are part of changelogs, history files, and other things that I assume should remain accurate to the time they were written, not the confusing times we live in today.

mflatt:
I don't know if I like this idea, but just as a poll: What if we changed "Mac OS X" to just "Mac OS" instead of the correct (for now!) "macOS"?

I know it shouldn't bother me, but because I'm so old, "macOS" hurts my eyes – especially at the beginning of a sentence. "Mac OS" has a perfectly clear meaning, it side-steps the issue of past versions, and it has the advantage of staying equally incorrect when when Apple changes the OS name yet again. "Mac OS" has the drawback of being wrong.

mbutterick:
It’s not bothersome because you’re old. It’s bothersome because it’s objectively awful.

Your argument for “Mac OS” is persuasive.

rfindler:
+1

mfelleisen:
+1 for Matthew and +omega for Matthew. Max OS all the way (no typo :-)

mflatt:
Ok, let's try out "Mac OS"

PR #1532, opened 2016-12-06
Fix bug in hash-table key-value match pattern

jrslepak:
Match pattern allowed `(hash-table)' to match on non-empty hashes


samth:
This looks right to me. Note that you don't have to add tests in all those different places. :)

jrslepak:
Failing case was allowing the match clause `[(hash-table) expr]` to fire when given `#hash((x . 3))`, when it should only match a key-value hash pattern with one key-value pair.

jrslepak:
Ok, wasn't sure if it was worth checking the mzlib versions of match, but it wasn't much work to add those in.

PR #1534, opened 2016-12-13
Space-performance adjustments to `for/list` and `for .... in-list ....`

mflatt:
These two commits seem like good ideas at the moment, but they have some important caveats, so I'd like to get opinions on them.

The first commit compiles simple `for/list` forms to `cons` onto a recursive call instead of using `reverse` – cutting the pair allocations in half, and making it more like `map`. The second commit is about avoiding an implicit reference to the value that is being processed in a loop iteration, which can prevent that value from being GCed.

If we go with these changes, the second commit needs to be augmented with doc changes and tests to check and explain the intended guarantees.

Both of these experiments were motivated by the linklet branch. The first commit doesn't improve the expander's performance in my tests, but it doesn't hurt. The second commit reflects an important and pervasive need in the expander to not retain the original of an expression that is in the process of being expanded. (The expander currently has its own macro to replace several uses of `for/list`.)


mflatt:
Beware of trying to build with these commits. Something is wrong and the documentation step of the build starts using all of memory.

rfindler:
The space commit seems clearly like a good idea and I think it makes sense to document this explicitly as a limitation (in a tone that suggests that if someone wants to fix it with a more complex `make-sequence` protocol, that fix would be welcome).

The other one is hard for me to judge. Is the performance improvement by avoiding the reverse something that is well-understood, or is this something where making some plots might help clarify the choice?

mflatt:
The performance difference for building a list is the same as the following two programs in current Racket:
```
#lang racket

(define N 5000)
(define L 5000)

(collect-garbage)
(time
 (void
  (for/fold ([v 0]) ([i (in-range N)])
    (for/list ([j (in-range L)])
      j))))

(collect-garbage)
(time
 (void
  (for/fold ([v 0]) ([i (in-range N)])
    (let loop ([j 0])
      (if (= j L)
          null
          (cons j (loop (add1 j))))))))
```
At the moment, I don't find an `L` that makes the latter one slower. Back when we changed `map`, I thought that it wasn't always a win, but maybe things have changed.

Meanwhile, the bug in the revised `for/list` is that it doesn't handle 0 sequences (by stopping after 1 iteration). And I think I was confused about the difficulty of making `for/list` always avoid `reverse`; it now looks easy enough to generalize to all `for/list` forms, although I haven't yet done that.

rfindler:
I ran the program below on a few machines I have easy access to (a macbook air with 1.7 ghz core i7, mac pro 2.6 ghz intel xeon e5, and a red hat Intel(R) Xeon(R) CPU E31270 @ 3.40GHz) and none of them could find a value of `L` where the latter was slower. (I also tried without the jit on the air.)

```
#lang racket
(require compiler/find-exe)

(define me (let-syntax ([m (λ (stx) (with-syntax ([f (syntax-source stx)]) #''f))]) m))

(let/ec done
  (for ([ignored (in-range 10000)])
    (define sp (open-output-string))
    (parameterize ([current-output-port sp])
      (system* (find-exe) "-e" (~s `(require (submod (file ,(~a me)) timing)))))
    (unless (equal? "" (get-output-string sp))
      (display (get-output-string sp))
      (done (void)))
    (display ".")
    (flush-output))
  (printf "never slower\n"))

(module timing racket/base
  (define L (+ 1 (random 10000)))
  (define N (ceiling (/ (* 5000 5000) L)))

  (collect-garbage)
  (define-values (results1 cpu1 real1 gc1)
    (time-apply
     (λ ()
       (for/fold ([v 0]) ([i (in-range N)])
         (for/list ([j (in-range L)])
           j)))
     '()))

  (collect-garbage)
  (define-values (results2 cpu2 real2 gc2)
    (time-apply
     (λ ()
       (for/fold ([v 0]) ([i (in-range N)])
         (let loop ([j 0])
           (if (= j L)
               null
               (cons j (loop (add1 j)))))))
     '()))

  (when (< cpu1 cpu2)
    (printf "slower @ L = ~s\n" L)))
```

mflatt:
Thanks! After looking at this a lot more, I've become convinced that the `for/list` change is a good idea. Partly, I'm more encouraged because I now have a general solution that ends up being just a little more code, instead of a solution that's a big chunk of duplicated code that only handles a subset of `for/list` forms.

Adjusting the sequence API to make `in-list` generally work also isn't that bad, although I don't have it right yet. (So far, I've found one optimizer bug. Space safety is hard.)

rfindler:
Am I right that because the code below prints out a list (and not `#f`) that the `#%kernel` version of `map` has to be adjusted too? (Probably you already realized this, and if so and decided not to bother then that's okay with me :)).

```
#lang racket
(let ()
  (define l1 (cons 2 '()))
  (define lb (make-weak-box l1))
  (define l2 (cons 1 l1))
  (set! l1 #f)
  (void
   (map (λ (i j k)
          (when (= i 2)
            (collect-garbage)
            (printf "~s\n" (weak-box-value lb))))
        l2 l2 l2)))

(let ()
  (define (my-map f l1 l2 l3)
    (cond
      [(null? l1) null]
      [else (let ([r1 (cdr l1)]
                  [r2 (cdr l2)]
                  [r3 (cdr l3)])
              (cons (f (car l1) (car l2) (car l2))
                    (my-map f r1 r2 r3)))]))

  (define l1 (cons 2 '()))
  (define lb (make-weak-box l1))
  (define l2 (cons 1 l1))
  (set! l1 #f)
  (void
   (my-map (λ (i j k)
             (when (= i 2)
               (collect-garbage)
               (printf "~s\n" (weak-box-value lb))))
           l2 l2 l2)))

```

PR #1535, opened 2016-12-13
doc: fix types in regexp constructors

bennn:
- change return types of `pregexp` etc. to use the right predicate
- change input of `byte-regexp` and `byte-pregexp` from `string?` to `bytes?`


stamourv:
LGTM

PR #1540, opened 2016-12-19
Remove the documentation mentioning `chaperon-procedure*?`

shhyou:
The function `chaperon-procedure*?` mentioned in the documentation does not exist (or is out-dated).


stamourv:
Oops. This looks like an earlier sketch for the docs for `impersonate-procedure*?` which I left there by mistake. Thanks for catching it!

PR #1542, opened 2016-12-20
Detect and propagate errors from tar to result of tar-gzip.

tonyg:
The `pkg-index` server periodically runs a task that calls `get-pkg-content` from `pkg/lib`. The task is guarded by a semaphore and a use of `call-with-semaphore`.

The thunk passed to `call-with-semaphore` is supposed to either return or raise an error. Either way, the semaphore would then be released, and subsequent executions guarded by the same semaphore could continue.

Instead, it appeared as if the thunk was never returning. This prevented all subsequent update-checks from running until the server was restarted.

What was happening was this:

 - During updating, a call to `get-pkg-content` was made for package `racket-lang-org` hosted at `git://github.com/racket/racket-lang-org` and at revision `461e03f57bed723629273f4f3dbe9fba6dd1ca15` (though the current head of that repository, `fb310a712f96650331d86f5ba44c3b1b457693fe`, also triggers the bug)

 - `get-pkg-content` was calling `download-repo!` from `pkg/private/download`

 - `download-repo!` was calling `tar-gzip` with the results of a successful git checkout operation

 - `tar-gzip` was spawning a new thread, which called `tar->output`

 - `tar->output` was failing with the following error message:

         tar: path too long for USTAR: blog/_src/posts/2007-12-19-your-security-hole-is-my-fun-hack-or-computing-factorial-in-drscheme-with-a-click-powered-loop.md
           context...:
            /home/tonyg/src/racket/racket/collects/file/tar.rkt:46:0
            /home/tonyg/src/racket/racket/collects/racket/private/map.rkt:57:19: loop
            /home/tonyg/src/racket/racket/collects/racket/private/more-scheme.rkt:261:28
            /home/tonyg/src/racket/racket/collects/file/tar.rkt:184:16

 - Because `tar->output` was running in a separate thread, the exception was being lost.

 - The main thread that was waiting for the completion of `tar->output` ended up waiting for ever because no call to `close-output-port` was ever made by the crashed thread.

There are several issues here:

 1. Lack of any kind of error-handling for the cross-thread interaction in `tar-gzip`.
 2. Problems with very long pathnames given to `tar`.
 3. Lack of any kind of timeout in the `call-with-semaphore` thunk.

This pull request adds a `with-handlers` around the call to `tar->output` in `tar-gzip` to catch any exception; if an exception is caught by the new `with-handlers`, it is re-raised in the main thread. This allows the thunk given to `call-with-semaphore` to proceed, which in turn allows subsequent updates to run.

### DIFF
--- a/pkgs/racket-test/tests/file/gzip.rkt
+++ b/pkgs/racket-test/tests/file/gzip.rkt
@@ -77,6 +77,26 @@
   ;; should be around 6 times smaller
   (id* (file->bytes big-file) 4))
 
+(define (test-hang-on-long-filename)
+  ;; `tar-gzip` was crashing in a background thread due to a very long
+  ;; filename, leading to a hang in the main thread. Trigger the bug
+  ;; with a long filename, and check that progress is made in a
+  ;; reasonable length of time.
+  (local-require file/tar)
+  (define long-filename
+    "extremely-long-file-name-that-exceeds-the-internal-length-limits-on-USTAR-because-of-arcane-decisions-and-bad-choices-of-fixed-width-records-and-so-on-and-so-on")
+  (define temp-dir (find-system-path 'temp-dir))
+  (define long-path (build-path temp-dir long-filename))
+  (with-output-to-file long-path (lambda () (write long-filename)) #:exists 'replace)
+  (define tgz-file (build-path temp-dir "long-filename-test.tgz"))
+  (when (file-exists? tgz-file) (delete-file tgz-file))
+  (define ch (make-channel))
+  (thread (lambda ()
+            (with-handlers [(exn:fail? (lambda (_exn) (channel-put ch 'success)))]
+              (tar-gzip tgz-file long-path)
+              (channel-put ch 'didnt-fail-in-tar-gzip))))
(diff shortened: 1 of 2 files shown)


rfindler:
Probably you don't want `(with-handlers [(values`, unless you're trying explicitly to handle all things that could be passed to `raise` except `#f`. I think you want to use `exn:fail?` as the predicate instead (because catching break exceptions and propagating them is strange and confusing).

tonyg:
I would argue against `exn:fail?`, and specifically for propagating breaks, because the use of a thread is unimportant here, and any failure or interruption should be signalled on the main thread.

rfindler:
How would a break get queue'd for this private thread?

rfindler:
It really doesn't matter, I suppose. We can see all the code that runs on this private thread and none of it raises anything other than `exn:fail?`.

I hope you squash the commits before pushing.

mflatt:
This looks fine to me, although it would be great to have a test in "pkgs/racket-test/tests/file".

tonyg:
Thanks @rfindler and @mflatt for the review comments. @mflatt, I have added a test case. On the assumption that all is OK, I'll merge this now.

PR #1544, opened 2016-12-21
optimizer: merge lookup_constant_proc and optimize_for_inline

gus-massa:
The objective of `lookup_constant_proc` and the first part of `optimize_for_inline` was to find out if the value of an expression was a procedure and get it to analyze its properties or try to inline it. Both were called together in a few places, because each one had some special cases that were missing in the other.

So, move the lookup and special cases from `optimize_for_inline` to `lookup_constant_proc`, and keep only the code relevant to inlinig in `optimize_for_inline`.

The changes are asier to undertand if you imagine that `optimize_for_inline` was splited and renamed as `lookup_constant_proc`, and then a few missing details of the old `lookup_constant_proc` were copied to the new `lookup_constant_proc`.

When there is an arity mismatch, it returns `scheme_true`. Another posibility would be to use `scheme_false` but I guess that when this code will be translated from C to racket, the `NULL` will be translated to `scheme_false` and casue a collision.

When not trying to inline, it will go inside `will_be_lambda`s. Please double check that this is correct.
[R2600](https://github.com/racket/racket/compare/master...gus-massa:16-11-Inline-NNN?expand=1#diff-5e09d7df9102fa81a4736e357a0c3e1aR2600)

I'm using directly `[var]->mode` and `[var]->optimize.known_val` instead of `optimize_info_lookup_lambda` to avoid the error when the mode is incorrect. See [2544](https://github.com/racket/racket/compare/master...gus-massa:16-11-Inline-NNN?expand=1#diff-5e09d7df9102fa81a4736e357a0c3e1aR2544) This is necesary in some code like:

    (call/cc 
     (let ([p (lambda (_) 0)
        p))

I don't want to move the optimization for call/cc after the argument is optimized because I exper it to be more useful in cases were the lambda is direct, like

    (let/cc exit
      (do-domething-ignoring-exit))

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -340,6 +340,24 @@
                  g
                  g))))
 
+;; Check reduction of single-use lambdas
+;; this test uses that a lambda with a '(1) can't be duplicated
+(test-comp '((lambda (x) '(1)) 5)
+           ''(1))
+(test-comp '((case-lambda [(x) '(1)] [(x y) 0]) 5)
+           ''(1))
+(test-comp '(let ([f (lambda (x) '(1))])
+              (f 5))
+           ''(1))
+(test-comp '(let ([f (case-lambda [(x) '(1)] [(x y) 0])])
+              (f 5))
+           ''(1))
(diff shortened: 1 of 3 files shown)


mflatt:
Looks good to me

PR #1545, opened 2016-12-24
Typo - "describes" should be "described"

arrdem:
A simple enough fix, but one that stuck out to me as I was reading the live `_enum` docs.


stamourv:
Thanks!

PR #1548, opened 2016-12-26
Make PLTDISABLEGC work again

jamesbornholt:
It looks like `PLTDISABLEGC` hasn't been working for a while.

A git bisect points at e59f888 as the commit where this setting broke. In that change, new GCs no longer inherit the `avoid_collection` value, which is what PLTDISABLEGC increments. So as soon as the new GC is created for the master place ([GC_switch_out_master_gc](racket/src/racket/gc2/places_gc.c#372)), that setting is lost.

I'm not 100% sure this is the right fix because the invariants here are tricky; in particular, the old code in b24c387 incremented `avoid_collection` just before creating the new GC, and then decremented it when inheriting. But it didn't break my places-using code, and does indeed disable the GC now :-)


mflatt:
Thanks!

PR #1553, opened 2016-12-30
Need parentheses

sorawee:
This needs parentheses. Otherwise, it's going to be just `id` which will "match anything, bind identifier", and give

```
0
0
0
```

while the expected output is

```
0
1
2
```

@rfindler (who originally wrote this part): does this seem correct?


rfindler:
You are right that, if you want a correct function, you need the parens. But my intention was to demonstrate the pitfall.

I see that I didn't explain that very well! I'll push a clarification (that closes the PR; if you think it can be improved, however, please do re-open).

PR #1557, opened 2017-01-03
document raco/all-tools, fix typos in raco documentation

bennn:
The point of this PR was to put `raco/all-tools` in the docs, but it turns out most of the edits are typo fixes.

### DIFF
--- a/pkgs/racket-doc/scribblings/raco/api.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/api.scrbl
@@ -106,7 +106,7 @@ collection.  The following fields are used:
 
  @item{@indexed-racket[compile-omit-paths] : A list of immediate file
        and directory paths that should not be compiled. Alternatively,
-       this field's value @racket['all], which is equivalent to
+       this field's value is @racket['all], which is equivalent to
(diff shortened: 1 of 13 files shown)


mflatt:
Thanks! This is clearly a big improvement.

I didn't try building the document, but I don't think that using `defstruct*` instead of another binding's description will work right, since it will interrupt the enclosing descrpition. Also, it will make the declaration appear to be provided by the module, but it isn't.

FWIW, I'm not a fan of using a predicate like `path-string?` as a noun that stands for a value that satisfies the predicate, but that's minor.

mflatt:
Looks ready to merge to me — thanks!

bennn:
haha thanks, I was just building a copy of the docs:
http://www.ccs.neu.edu/home/types/racket-doc/raco/command.html?q=asdf#%28tech._command._specification%29

Since "command specifications" is now a `deftech` [(link)](https://github.com/racket/racket/pull/1557/files#diff-40d6c8457967df7241159a8c9aa091a8R14), do you think it should be "raco command specifications" ?

mflatt:
I think "command specifications" is fine.

PR #1558, opened 2017-01-03
Make range from racket/list act like in-range when used with for

lexi-lambda:
As mentioned on the dev mailing list, this make `range` act like `in-range` when used directly within `for` forms to improve performance. Unlike `in-range`, it treats arity-related misuses as a runtime error instead of a syntax error to ensure complete backwards compatibility with the existing `range` function.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/pairs.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/pairs.scrbl
@@ -1264,12 +1264,19 @@ predecessor until @racket[end] (excluded) is reached.  If no starting
 point is provided, @racket[0] is used. If no @racket[step] argument is
 provided, @racket[1] is used.
 
+Like @racket[in-range], a @racket[range] application can provide better
+performance when it appears directly in a @racket[for] clause.
+
 @mz-examples[#:eval list-eval
   (range 10)
   (range 10 20)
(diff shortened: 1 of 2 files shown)


mflatt:
LGTM. I'm impressed that you thought to preserve the `range` name and appreciate that you included a `history` note.

rfindler:
Was the existing test suite sufficient?

lexi-lambda:
Given that the only observable difference in behavior is the amount of time iteration takes, I wasn’t sure how to write a test that ensures the good behavior produced by this change. I *could* add a test to ensure the cases work properly, though…? I’d admittedly be a bit skeptical of the value of such a test, but I’ll add it if you like.

rfindler:
Oh, I was thinking of a test case like ((values range) 1 2) or something like that, where there is a new case to handle in the implementation than when `range` is defined as a normal function, but maybe the way that `define-sequence-syntax` works it doesn't matter.

rfindler:
(indeed, I would say that if you think there isn't a need for a test like that, then I'm happy. I was just curious if you had considered that.)

lexi-lambda:
Sure, understood. I think this particular case doesn’t warrant the test, but I didn’t mean to be dismissive of the suggestion.

PR #1565, opened 2017-01-09
Replace arithmetic with bit operations

mromyers:
quotient/remainder is replaced with word-index and bit-index, expt with arithmetic-shift.
Closes https://github.com/racket/data/issues/2


stamourv:
Thanks!

PR #1566, opened 2017-01-09
optimizer: use flags of lambdas if they can't be inlined

gus-massa:
In the first commit https://github.com/racket/racket/commit/989e21a7484af05a6eea726296d260c1f167b6fe the optimizer tries harder to find a `lambda `to get the flags to determine if the expression is single valued or preserves marks. This restores the behavior before https://github.com/racket/racket/commit/85eee2bbbcc075fbf97cf56d8d9d59ffd03b3a0d . It’s small and it is some kind if bug fix, so it can be merged to the release version of 6.8. (The current code provides the correct result, but it’s slightly pessimistic.)

The second commit https://github.com/racket/racket/commit/1ced6f23cd84b33e4eb86023b1e98e82c64ad7fb is an idea to move this code to the bottom of each `finich_optimize_application function`, so the flags are calculated after all the other optimizations, and the indirect `lambda `can be used for calculating the flags for the application and whether the clocks should advance. 

This second commit is big, and I think I still have to fix a few corner cases, so I’ll resubmit this next week.  If the first commit is not included in 6.8, I think it would be better to squash them.

### DIFF
--- a/racket/src/racket/src/optimize.c
+++ b/racket/src/racket/src/optimize.c
@@ -2473,7 +2473,9 @@ int scheme_check_leaf_rator(Scheme_Object *le)
 
 int scheme_get_rator_flags(Scheme_Object *le)
 {
-  if (SCHEME_PRIMP(le)) {
+  if (!le) {
+    return 0;
+  } else if (SCHEME_PRIMP(le)) {
     int opt;
     opt = ((Scheme_Prim_Proc_Header *)le)->flags & SCHEME_PRIM_OPT_MASK;
     if (opt >= SCHEME_PRIM_OPT_NONCM) {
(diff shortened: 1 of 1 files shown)


mflatt:
I don't see any problem with the first commit, so merging that one is fine with me – although it would be nice to have a test case to expose the regression, if possible. (I couldn't' construct one right away, but I didn't try hard.)

gus-massa:
I only could test this visually using decompile, and I think it’s possible to test it using something like

    (match m
       [(compilation-top max-let-depth binding-namess prefix code)
     ...])

But I wanted to test this with something like

    (lambda ()
      (define (f) ‘(123)) ; f not clonable
      (display (list f f)); f not single use
      (values (f))

In this example, `f` is correctly marked as single valued and mark preserving, so it the `values` in the last line could disappear. But the reduction for `values` uses `single_valued_noncm_expression` that doesn’t look at the flags of the `lambdas`. My idea is to extend  `single_valued_noncm_expression` later (next week?), I think that this reduction will enable to test easily more similar cases.

PR #1569, opened 2017-01-10
delete unnecessary linebreak

mbutterick:
To obviate this problem on the `docs` server where there ends up being an empty paragraph in the margin note, which creates extra space at the bottom of the box:

screen shot 2017-01-10 at jan 10 11 41 54 am

Without the linebreak, it will look like so:

screen shot 2017-01-10 at jan 10 11 45 44 am

The linebreak, and the `p` tag it impliedly imposes, is unnecessary because the lower half of this margin note is inside a `span` that has a `display:block` CSS style explicitly imposed in the tag, so it’s still spaced properly when it’s revealed.

screen shot 2017-01-10 at jan 10 11 43 36 am


stamourv:
Thanks!

PR #1571, opened 2017-01-11
doc: switch "A is the result of B" to make more sense

AlexKnauth:
This changes
> The result of `proc` is the result of the `call-with-continuation-prompt` call.

to

> The call to `call-with-continuation-prompt` returns the result of `proc`.


stamourv:
Thanks!

PR #1574, opened 2017-01-13
Check ADLER32 after inflate

lehitoskin:
After prompting from @tonyg I decided to add an ADLER32 check in `zlib-inflate`.

### DIFF
--- a/racket/collects/net/git-checkout.rkt
+++ b/racket/collects/net/git-checkout.rkt
@@ -1081,6 +1081,20 @@
        (arithmetic-shift (read-number-by-bits i (arithmetic-shift n -1))
                          8))]))
 
+;; ADLER32 implementation
+;; https://www.ietf.org/rfc/rfc1950.txt
+(define (bytes-adler32 bstr)
+  (define ADLER 65521)
+  (define-values (s1 s2)
+    (for/fold ([s1 1]
+               [s2 0])
+              ([bits (in-bytes bstr)])
+      (define a (modulo (+ s1 bits) ADLER))
+      (define b (modulo (+ s2 a) ADLER))
+      (values a b)))
+  ; (s2 << 16) | s1
+  (bitwise-ior (arithmetic-shift s2 16) s1))
+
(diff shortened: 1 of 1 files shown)


tonyg:
(By the way, @lehitoskin, thank you for making the PR!)

tonyg:
Sure enough, the failing CI builds are from not having an `open-output-bytes` port in `o`.

lehitoskin:
I added an intermediate string port. Should be fine now.

lehitoskin:
It appears the appveyor check failed because of a networking issue:
```
Section(foreign)
Section(cstruct)
Section(uni-norm)
ERROR: tcp-connect: connection failed
  address: www.unicode.org
  port number: 80
  step: 6
  system error: A connection attempt failed because the connected party did not properly respond after a period of time, or established connection failed because connected host has failed to respond.; errno=10060
Command exited with code 2
```
And I don't know quite why that one build failed in Travis. I looked around in the logs and it doesn't appear to be related to my code changes.

lehitoskin:
The error in Travis for the one test doesn't really make sense to me, either: https://travis-ci.org/racket/racket/jobs/191791547#L1648-L1671

samth:
@tonyg Is your review comment addressed?

samth:
@lehitoskin can you rebase this, so that the CI runs again?

lehitoskin:
@samth Okay. I hope I did this correctly...

samth:
@mflatt any thoughts on this PR?

mflatt:
Do we care that this keeps a copy of a potentially large file in memory, instead of always streaming to an output file? Probably we don't deal with any files big enough for this to matter, so I'm ok with ignoring that issue for now, but but the alternative would be a thread that performs the checksum concurrently.

samth:
That was my worry as well here. I think we probably want to avoid decompressing the whole thing in memory.

mflatt:
I'll merge and adjust the implementation to streaming

PR #1577, opened 2017-01-16
Optimizer: More single valued and mark preserving reductions

gus-massa:
**consider the flags of unclonable lambdas for reductions**

Transform `single_valued_noncm_function` into a `get_rator_flags` wrapper to get a single implementation.

The lambdas can be marked as single valued and/or mark preserving. With this information is possible to remove unnecessary wrapping like the `values` in

    (let ([f (lambda () '(1))])
      (display f f)
      (values (f)))

or in reductions like

    (car (list (f))) ==> (values (f)) ==> (f)

**advance effect clocks after reductions**

After some reductions, the new rator advance less the effect
clocks than the original rator. For example in

    (equal? x 7) ==> (eq? x 7)
    (my-struct? x) ==> #t or #f [when the optimizer kmows the type of x]

**mark struct operations as single valued and mark preserving**

Also, the optimizer recognizes struct operations as procedures,
so it will reduce

  (procedure? my-struct?) ==> #t
  (procedure? prop:my-prop?) ==> #t

To log an arity error, the optimizer calls `scheme_get_proc_name` that doesn't recognize the internal representation of the struct procedures, so I wrapped the all in a check. [gus-massa/.../optimize.c#L2801](https://github.com/gus-massa/racket/blob/67e66ba3d240d960b9db6e728e5dfa331d4a4e70/racket/src/racket/src/optimize.c#L2801)

A possible alternative is to modify `scheme_get_proc_name` of try harder to create a name for the procedure.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -357,6 +357,58 @@
 (test #t (lambda () (let ([f (case-lambda [(x) '(1)] [(x y) 0])])
                       (eq? (f 5) (f 5)))))
 
+;; Check that lambdas are marked as single valed and mark preserving
+(test-comp '(let ([f (lambda () '(1))])
+              (display (list f f))
+              (values (f))) 
+           '(let ([f (lambda () '(1))])
+              (display (list f f))
+              (f)))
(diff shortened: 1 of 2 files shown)


mflatt:
These look good as far as I can tell

PR #1581, opened 2017-01-18
update docs to use `configure-runtime` submodule rather than obsolete…

mbutterick:
The docs for module-handling configuration [1] discuss the use of a syntax property to bring in a `get-language-info` function, which in turn points at a `runtime-config` module. (One confusion here is that the docs seem contradictory: above the code sample on this page it says "The module form must be annotated with a `'language-info` syntax property" but in the actual sample, it's a `'module-language` property.)

Elsewhere in the docs, however, this `'module-language` property is described as “an older protocol” [2] and suggests that the use of a `configure-runtime` submodule is preferred. Further evidence that the approach is obsolete: “I haven't yet re-written the Guide's chapter 17 to use submodules, but I'm planning to do that next.” [3] — May 2013

Based on this, I infer that the approach described in [1] is now deprecated. I have updated this section of the docs to use a `runtime-config` submodule, as described in [2].

[1] http://docs.racket-lang.org/guide/module-runtime-config.html
[2] http://docs.racket-lang.org/reference/running-sa.html#%28part._configure-runtime%29
[3] https://lists.racket-lang.org/dev/archive/2013-May/012318.html

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/hash-languages.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/hash-languages.scrbl
@@ -381,15 +381,15 @@ run-time configuration cannot be handled by the
 parser module.
 
 Instead, it will be handled by a new
-@racket['configure-runtime] submodule that we'll add inside
+@racket[configure-runtime] submodule that we'll add inside
 the parsed @racket[module] form. When a module is run directly
 with @exec{racket}, @exec{racket} looks for a
-@racket['configure-runtime] submodule. If it exists, @exec{racket}
+@racket[configure-runtime] submodule. If it exists, @exec{racket}
 runs it. But if the module is imported into another module,
 the @racket['configure-runtime] submodule is ignored. (And if the
-@racket['configure-runtime] submodule doesn't exist, @exec{racket}
+@racket[configure-runtime] submodule doesn't exist, @exec{racket}
 just evaluates the module as usual.) That means that the
-@racket['configure-runtime] submodule can be used for any special
+@racket[configure-runtime] submodule can be used for any special
 setup tasks that need to happen when the module is run directly.
 
 Going back to the @racket[literal] language (see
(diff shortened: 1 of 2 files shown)


pkgs/racket-doc/scribblings/guide/hash-languages.scrbl
```diff
@@ -362,150 +362,77 @@ The @filepath{kiddo.rkt} example illustrates how the format for
 printing a result value can depend on the main module of a program
 instead of the language that is used to implement it.
 
+More broadly, certain features of a language are only invoked when
+a module written in that language is run directly with @exec{racket}
+(as opposed to being imported into another module). One example is
+result-printing style (as shown above). Another example is REPL
+behavior. These features are part of what's called the
+@emph{run-time configuration} of a language.
```
samth: This should probably be either `@tech` or `@deftech`.

mbutterick: Wasn’t sure about that, or whether it should link to this:

http://docs.racket-lang.org/reference/running-sa.html?q=run-time%20configuration#%28part._configure-runtime%29

samth: In general you don't want to use `@emph` for technical terms. But I'm not sure if that's the right place to link to -- it seems like maybe this is the definition.

pkgs/racket-doc/scribblings/guide/hash-languages.scrbl
```diff
@@ -362,150 +362,77 @@ The @filepath{kiddo.rkt} example illustrates how the format for
...
+the @emph{source text} representing the module.
+For that reason, the run-time configuration for a
+module needs to be available even if the module is compiled
+to bytecode form and the source is unavailable. Therefore,
+run-time configuration cannot be handled by the
+@racketidfont{get-info} function we're exporting from the language's
+parser module.
+
+Instead, it will be handled by a new
+@racket['configure-runtime] submodule that we'll add inside
```
samth: Why the `'` here?

Issue #1556, opened 2017-01-02
"take" and "drop" report misleading error in case of bad first argument


hashimmm:
For example: 

> (take 123 2)
. . take: index reaches a non-pair
  index: 2
  in: 123

and:

> (drop "asdf" 2)
. . drop: index reaches a non-pair
  index: 2
  in: "asdf"

Something like the usual contract violation message, showing what was provided and what was expected, would be much nicer.

bennn:
#1582

PR #1582, opened 2017-01-18
clarify error message for take and drop

bennn:
#1556

@hashimmm what do you think?

### DIFF
--- a/racket/collects/racket/list.rkt
+++ b/racket/collects/racket/list.rkt
@@ -144,16 +144,13 @@
 (define (drop* list n) ; no error checking, returns #f if index is too large
   (if (zero? n) list (and (pair? list) (drop* (cdr list) (sub1 n)))))
 (define (too-large who list n)
-  (raise-arguments-error
-   who
-   (if (list? list)
-       "index is too large for list"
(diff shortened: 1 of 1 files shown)


hashimmm:
Nice, works for me. I suppose any wording other than "chain of pairs" would be either be inaccurate or less user-friendly.

AlexKnauth:
I don't think you need the "index reaches a non-pair" language at all. Would it make sense to replace it with something that expresses "expected a (possibly improper) list with at least ~a pairs"?

bennn:
Okay, here's another try:
```
> (take 2 2)
; take: contract violation
;   expected: a chain of at least 2 pairs
;   given: 2
> (drop '(2) 2)
; drop: contract violation
;   expected: a list with at least 2 pairs
;   given: '(2)
```

(maybe should say "list .... elements" instead of "list .... pairs")

bennn:
Posted one more update, based on feedback from alex

bennn:
ping @hashimmm  @AlexKnauth , okay if I merge this tonight?

AlexKnauth:
This looks much clearer

PR #1590, opened 2017-01-24
fix for examples

pnwamk:
the previous examples would break with usages of things like the `#:break` keyword

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/for.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/for.scrbl
@@ -379,13 +379,15 @@ source for all syntax errors.
 @mz-examples[#:eval for-eval
 (define-syntax (for/digits stx)
   (syntax-case stx ()
-    [(_ clauses . defs+exprs)
+    [(_ clauses body ... tail-expr)
      (with-syntax ([original stx])
        #'(let-values
-           ([(n k)
-             (for/fold/derived original ([n 0] [k 1]) clauses
-               (define d (let () . defs+exprs))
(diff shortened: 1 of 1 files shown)


stamourv:
Looks good, but you may want to add some explanation re why this particular way of doing it is preferable (i.e., your PR comment).

pnwamk:
PR comment updated to describe the details of the issue

stamourv:
Sorry. I meant in the docs proper.

pnwamk:
I'm not sure what we would add -- I think they docs are fine, the example was just wrong, and is now consistent with the rest of the docs in that area.

In the docs the declared specification for `for/fold/derived` is the following:

```racket
(for/fold/derived orig-datum
  ([accum-id init-expr] ...) (for-clause ...)
  body-or-break ... body)
```

Which is correct! I just fixed the _example_ following that specification in the docs to actually expand into that form that we already declared things should be. Previously the example was just wrong in a silly way (i.e. it didn't do what the docs said to do) that happened to work for some, but not all, of the ways a `for` loop can be used.

If you think the prose in the docs could be improved in a particular way I'm happy to make that change. But from where I sit the docs are fine, the example just broke it's own rule.

stamourv:
Here's my reasoning: the code in the docs was wrong, therefore it's not unbelievable that someone writing their own `for` form could make the same mistake (and end up with a `for` form that can't accept all the bells and whistles), and so explaining that mistake explicitly, and how to avoid it, may be a good idea.

If you think having an example that works fine is enough to avoid that, no problem. Just bringing it up.

PR #1591, opened 2017-01-24
Implemented and provided prop:template-metafunction

jsmaniac:
It is currently not possible to make an identifier behave as a template metafunction and, say, as a match expander.

This pull request adds the structure type property prop:template-metafunction which can be used to define template metafunctions.


jsmaniac:
The second OSX job timed out just before finishing, could someone restart just that job? Thanks!

racket/collects/racket/private/prop-template-metafunction.rkt
```diff
@@ -0,0 +1,69 @@
+(module prop-template-metafunction '#%kernel
+  (#%provide (protect metafunction
```
jsmaniac: Note: I'm not sure if it is okay to protect these identifiers in a separate module (previously they were defined & protected in `template.rkt`).

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
```diff
@@ -346,4 +346,37 @@ the context above; instead, @racket[let-values] would report an
 invalid binding list.
 }
 
+@deftogether[
+ ((defthing prop:template-metafunction struct-type-property?)
+  (defthing template-metafunction? (-> any/c boolean?))
+  (defthing template-metafunction-accessor (-> template-metafunction? identifier?)))]{
```
lexi-lambda: I think it might be better to only export `prop:template-metafunction`, since that’s what most of the other similar properties do, and it hides details of the implementation. Also, it doesn’t look like these are actually exported from `racket/base` anywhere.

racket/collects/racket/private/prop-template-metafunction.rkt
```diff
@@ -0,0 +1,69 @@
...
+                      prop:metafunction
+                      (rename prop:metafunction? metafunction?)
+                      metafunction-accessor))
+  
+  ;; The prop:metafunction structure type property can contain an
+  ;; identifier bound to the run-time metafunction procedure, or the index of a
+  ;; field containing such an identifier.
```
AlexKnauth: Could you make it so that `prop:metafunction` can be a function that takes an instance of that structure and produces an identifier? Where you could say
```racket
(struct foo [id]
  #:property prop:metafunction
  (lambda (self) (foo-id self)))
```
Properties like `prop:rename-transformer`, `prop:require-transformer`, `prop:provide-transformer`, `prop:pattern-expander`, and `prop:code-transformer` all allow something like this.

jsmaniac: Sure, will do!

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
jsmaniac: Can one extract the template-metafunction from an identifier with just the `prop:template-metafunction` or should I export a higher-level `identifier? → procedure?` extractor? (in order to manually expand the metafunction, I think this can be useful if you want to interleave metafunction expansion with some custom rewriting of the stx object.)

I haven't written struct-type-properties for a while, so I don't remember :).

lexi-lambda: You can’t do that, but I think it would be better to provide a separate `expand-template` function if you really wanted, which would handle things like `~?` and `~@` properly. I don’t think that’s necessary, though, since you can already manually expand a metafunction by `local-expand`ing a use of `syntax`, so it seems unnecessarily complicated (and I think expanding a metafunction by hand seems potentially indicative of something maybe being a bad idea…).

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
```diff
@@ -346,4 +346,37 @@ the context above; instead, @racket[let-values] would report an
...
+               #:property prop:template-metafunction
+                          (struct-field-index proc-id)))
+
+           (define-syntax mf (mf-struct #'my-metafunction-proc))
+
+           (template (mf 3))
+           (with-syntax ([(x ...) #'(1 2 3)])
+             (template ((mf x) ...)))]
+ 
+ @history[#:added 6.9]}
```
AlexKnauth: Versions should be strings, not floating-point numbers, and also we're past 6.9 now.

AlexKnauth:
@jsmaniac I've started some small fixes on a branch here, which you can use if you want:
https://github.com/AlexKnauth/racket/commits/prop-template-metafunction-2
This includes fixing the `history` annotation, renaming `metafunction` -> `template-metafunction`, and actually providing the new values from `syntax/parse/experimental/template`.

racket/collects/racket/private/prop-template-metafunction.rkt
AlexKnauth: I've started on a commit that does this here on a `prop-template-metafunction-2` branch based on yours:
https://github.com/AlexKnauth/racket/commit/da9e4f06b9b0323c7f5562dbee6b29ac4860ef58

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
AlexKnauth: They shouldn't be exported from `racket/base` but from `syntax/parse/experimental/template`. However, currently they're not exported from there either, and they're called `prop:metafunction` and `metafunction?` when they should be called `prop:template-metafunction` and `template-metafunction?`.

I started a branch to fix that
https://github.com/racket/racket/compare/master...AlexKnauth:prop-template-metafunction-2

jsmaniac:
@AlexKnauth Thanks a lot for the excellent patches, and sorry it took me so long to review them!

jsmaniac:
I just realized that we changed the name of a couple of identifiers exported by `racket/collects/racket/template.rkt`: `metafunction` and `metafunction?` became `template-metafunction` and `template-metafunction?`.

This is likely to break backwards compatibility, should I revert the change (and remove `template-` from all names, for consistence)?

jsmaniac:
Also I'll ask again: is it okay to `protect` the identifiers `template-metafunction`, `prop:template-metafunction`, `template-metafunction?` and `template-metafunction-accessor` in a separate module (previously they were defined & protected in `template.rkt`).

I really don't understand `protect`, so I'm clueless :)

racket/collects/racket/private/prop-template-metafunction.rkt
```diff
@@ -0,0 +1,88 @@
...
+                               " an identifier")
+                val))))))
+  
+  (define-values (prop:template-metafunction
+                  prop:template-metafunction?
+                  template-metafunction-raw-accessor)
+    (make-struct-type-property 'template-metafunction
+                               prop:template-metafunction-guard))
+  
+  (define-values (template-metafunction-accessor)
```
rmculpepper: Since the accessor isn't exported (and shouldn't be), there's no need for an optional fallback argument, so this code can be simplified.

racket/collects/racket/private/prop-template-metafunction.rkt
```diff
@@ -0,0 +1,88 @@
...
+            (let-values ([(accessor) (make-struct-field-accessor make-struct-accessor val)])
+              (λ (instance)
+                (let-values ([(metafunction-internal-id) (accessor instance)])
+                  ;; Check that the value contained in the field is correct
+                  ;; (if (identifier? metafunction-internal-id) ...)
+                  (if (if (syntax? metafunction-internal-id)
+                          (symbol? (syntax-e metafunction-internal-id))
+                          #f)
+                      metafunction-internal-id
+                      (raise-argument-error
```
rmculpepper: `raise-argument-error` should be used with a contract, not a statement. Maybe use `raise-arguments-error` instead (note plural).

racket/collects/racket/private/prop-template-metafunction.rkt
```diff
@@ -0,0 +1,88 @@
+(module prop-template-metafunction '#%kernel
```
rmculpepper: I would prefer this whole module's contents to be put in racket/private/template.rkt instead of a separate file. And that means this code can use the "small-scheme.rkt" language, including `define` and `cond` and so on.

racket/collects/racket/private/prop-template-metafunction.rkt
```diff
@@ -0,0 +1,88 @@
...
+  (#%provide (protect template-metafunction
```
rmculpepper: These don't need to be protected. (I'm not sure why I protected `metafunction` and `metafunction?` originally.)

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
```diff
@@ -346,4 +346,42 @@ the context above; instead, @racket[let-values] would report an
...
 
+@deftogether[
+ ((defthing prop:template-metafunction struct-type-property?)
+  (defthing template-metafunction? (-> any/c boolean?)))]{
+
+ A structure type property, and the associated predicate. The property value is
+ one of:
+
+ @itemlist[
+   @item{an identifier bound to a run-time metafunction}
```
rmculpepper: I think the phrase "run-time metafunction" is a little confusing. Maybe "a run-time procedure implementing the metafunction", and then below "The metafunction's implementation should accept ...".

It might be worth adding a paragraph somewhere about that clarifies that the "run time" of the metafunction is typically the "compile time" of the enclosing module.

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
```diff
@@ -346,4 +346,42 @@ the context above; instead, @racket[let-values] would report an
...
+   @item{the index of a field containing such an an identifier}
+   @item{a procedure that takes an instance of the structure and produces
+         such an identifier}]
+
+ The identifier's run-time metafunction should accept a syntax object
+ representing its use, and produce a new syntax object as a result.
+
+ When an identifier is bound as syntax to a structure instance with this
+ property, it is treated as a template metafunction as if the identifier had
+ been declared with define-template-metafunction.
```
rmculpepper: `@racket[define-template-metafunction]`

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
lexi-lambda: At first, I thought your comment was describing a different behavior—I interpreted the documentation as requiring the identifier to actually be bound to a metafunction (i.e. something ultimately bound by `define-template-metafunction`), not just any ordinary procedure. Upon closer inspection of the PR, though, you’re right: that’s not what the implementation does.

So, obviously, I agree with you: the docs as-written aren’t right. But it’s also worth thinking about which implementation is right. As implemented right now, without exporting the accessor (as both you and I have suggested), this means you can’t use `prop:template-metafunction` to easily indirect to an existing metafunction (assuming you don’t already have the metafunction procedure). Though, on the other hand, you could do it indirectly by making the metafunction procedure produce ``#`(other-metafunction-id . #,(stx-cdr input-stx)))``, so maybe it doesn’t matter.

pmatos:
ping! This PR has one reviewer approval but has never been merged. Can someone look into the current situation so this is not forgotten?

samth:
I think the current situation is that a small doc change is needed, plus moving the contents of one file into another. @jsmaniac I'm happy to do that if you don't have time.

rmculpepper:
I'm not sure this is the right approach. I want to take a closer look again, and I'll try to make time tonight.

jsmaniac:
> I think the current situation is that a small doc change is needed, plus moving the contents of one file into another. @jsmaniac I'm happy to do that if you don't have time.

Yes, please! I'm not working much with Racket anymore, unfortunately :-( , so I'm a bit rusty on this codebase.

samth:
This would be good to add but I think @rmculpepper would need to be the one to fix the merge conflicts.

rmculpepper:
There's a conflict between this code and the current design of template metafunctions. Currently, the implementation of hygiene for metafunctions is in `define-template-metafunction`. This code would allow the definitions of metafunctions that bypass `define-template-metafunction`, so they would be unhygienic (bad). 

I think the best fix is for me to move the implementation of hygiene from definition to application, and then I can merge this (and fix any remaining issues mentioned above).

jsmaniac:
I rebased my PR and fixed the conflicts (original can still be found at https://github.com/jsmaniac/racket/tree/prop-template-metafunction-original-pr) but haven't done any of the suggested&requested changes (yet, though I'm not that active on Racket these days so could take a long while).

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
jsmaniac: > I think the phrase "run-time metafunction" is a little confusing. Maybe "a run-time procedure implementing the metafunction", and then below "The metafunction's implementation should accept ...".

Done :)

I'm marking this as resolved based on the assumption that the difference between passing directly a procedure and passing an identifier previously defined using define-template-metafunction isn't too significant (and IIRC it's more common in this sort of prop:xyz to pass a procedure?). But @lexi-lambda please to reopen this conversation if you'd like the other implementation :)

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
jsmaniac: Done :)

pkgs/racket-doc/syntax/scribblings/parse/experimental.scrbl
jsmaniac: @AlexKnauth's branch has been merged into this PR a while ago, just double-checked and it seems that all is up-to-date so I'm marking this discussion as resolved --- please do reopen if I was mistaken :)

jsmaniac:
I added two commits that would need a review:

- [ ] https://github.com/racket/racket/pull/1591/commits/c4c7ddc6ede19b358b7cfe12869f2b19c6255039
- [ ] https://github.com/racket/racket/pull/1591/commits/32ddaecd6ff5248c730cf7f881f9eeef8a05dfab

Those changes were pending in my copy of the repo for quite some time, and I'm really not sure if they are the right thing to do.

Other changes that need to be done for this PR:

- [ ] "move the implementation of hygiene from definition to application" (I'm not sure what implementing hygiene at the call-site exactly means @rmculpepper ? https://github.com/racket/racket/pull/1591#issuecomment-618381330
- [ ] Move the contents of racket/collects/racket/private/prop-template-metafunction.rkt to racket/private/template.rkt https://github.com/racket/racket/pull/1591#discussion_r261272302

PR #1596, opened 2017-01-27
optimizer: advance the vclock for a `values` with 0 or more than 1 argument

gus-massa:
To avoid moving expressions that may have a side effect, the optimizer must recognize that in this position this will cause an error and advance the virtual clock.

Currently the only primitive that is flagged as SCHEME_PRIM_IS_OMITABLE and may have multiple return values is `values`.

Thanks to Robby for finding the original version of the test.


mflatt:
LGTM

Issue #1599, opened 2017-01-31
Inconsistent file-position with peeking-input-port, read and utf-8


jsmaniac:
```
(let* ([p (open-input-bytes (string->bytes/utf-8 "(…)abcdef"))]
       [p (peeking-input-port p)])
  (read p)
  (displayln (file-position p))
  (port->string p))
;; => "abcdef"
```

In the code above, `(displayln (file-position p))` shows `7`, but it should show `5`, as `(bytes-length (string->bytes/utf-8 "(…)"))` gives `5` (the `…` character takes 3 bytes).

Another way to check that the `7` is incorrect is to create a new port and skip `7` bytes. This indeed goes too far:

```
(let* ([p (open-input-bytes (string->bytes/utf-8 "(…)abcdef"))]
       [pk (peeking-input-port p 'p 7)])
  (port->string pk))
;; => "cdef" instead of "abcdef"
```

PR #1600, opened 2017-01-31
Test for 69d7636 fix position counting with internal `read` shortcut

jsmaniac:
@mflatt Thanks for fixing issue #1599 !

This is a test for it (I haven't tried it with the fixed implementation yet, but it should pass with your patch).


mflatt:
Hi @jsmaniac – The bug-repair commit does include a test, although it more directly tests the problem primitives instead of going through `peeking-input-port`.  Did you see that test, but would also like to have one at the `peeking-input-port` level?

jsmaniac:
@mflatt Yes, I just thought it would be good to also test it through peeking-input-port. Feel free to reject the PR if you think the lower-level test is enough, of course :) !

mflatt:
Ok, thanks for the clarification!

Issue #1602, opened 2017-02-01
syntax-parse: datum->syntax called with list as context argument


AlexKnauth:
In Racket version 6.8.0.2, if I run the program below, I get the following error message:
```
datum->syntax: contract violation
  expected: (or/c syntax? #f)
  given: '(#<syntax 2> #<syntax wrong>)
  argument position: 1st
  other arguments...:
   '(#<syntax wrong>)
   '(#<syntax 2> #<syntax wrong>)
  context...:
   /Applications/Racket/2017-01-17/Racket v6.8.0.2/collects/syntax/parse/private/runtime-report.rkt:672:0: context-prose-for-expect
   /Applications/Racket/2017-01-17/Racket v6.8.0.2/collects/syntax/parse/private/runtime-report.rkt:577:0: report/expectstack
   /Applications/Racket/2017-01-17/Racket v6.8.0.2/collects/syntax/parse/private/runtime-report.rkt:58:0: handle-failureset
   /Applications/Racket/2017-01-17/Racket v6.8.0.2/collects/syntax/parse/private/runtime-report.rkt:28:0: call-current-failure-handler
   /Users/Alex/Desktop/match-patterns-as-values/syntax-parse-bug.rkt: [running body]
```

The program that produces it has a weird configuration of syntax classes. This is a simplification of the syntax classes used in the Turnstile package.
```racket
#lang racket/base
(require syntax/parse)

(define-syntax-class stuff
  [pattern (2 :three)])

;; a splicing syntax class for just 3
(define-splicing-syntax-class three
  [pattern 3])

;; like stuff, but with an extra attribute
(define-syntax-class stuff*
  [pattern :stuff #:with random-attr 'whocares])

(syntax-parse #'(1 2 wrong)
  [(1 . :stuff*)
   'body])
```

Expected behavior:

If I try to simplify the syntax-classes slightly, for instance if I take away the extra `random-attr` attribute, then I get the good error message I expected, saying `expected the literal 3 at: wrong`.

AlexKnauth:
@rmculpepper 

The bad call to `datum->syntax` happens [*here*](https://github.com/racket/racket/blob/master/racket/collects/syntax/parse/private/runtime-report.rkt#L689-L693), and the bad value is `cx`, which is a list instead of a syntax object:

```racket
(define (stx+index->stx stx+index)
  (let*-values ([(stx) (car stx+index)]
                [(index) (cdr stx+index)]
[...]

PR #1604, opened 2017-02-01
syntax-parse: ps->stx+index returns pairs with syntax objects inside

AlexKnauth:
Fixes https://github.com/racket/racket/issues/1602


AlexKnauth:
Travis cancelled the two MacOS builds because their MacOS queue is backed up. However, all the other builds are passing and I ran all the syntax-parse tests on my Mac as well.

AlexKnauth:
@rmculpepper Should this be merged?

rmculpepper:
The full fix was more complicated, because a `'car` progress frame can also mean looking inside a vector or prefab struct. I included a condensed version of your test case and added more for vectors and structs.

PR #1609, opened 2017-02-03
Use correct base path for finding gracket.

samth:
Fixes error in
http://drdr.racket-lang.org/38697/pkgs/compiler-test/tests/compiler/embed/test.rkt#output1


mflatt:
This change makes sense to me

samth:
I'll merge it after all the CI tests pass.

Issue #1616, opened 2017-02-08
“git-checkout” fails when it can’t reuse the connection


leafac:
For context, see https://github.com/tonyg/racket-pkg-website/issues/36.

My investigation led me to believe that `git-checkout` is to blame for the issue. It [reuses the HTTP connection](https://github.com/racket/racket/blob/b3223ad8d2ef285e11394fcd4e42af0a13e4775c/racket/collects/net/git-checkout.rkt#L742-L764) for multiple `http-conn-sendrecv!`. But, as the following script demonstrates, `http-conn-sendrecv!` (in particular, `http-conn-recv!`) might have the side effect of closing the connection:

```racket
#lang racket
(require net/http-client)

(define conn (http-conn))
(http-conn-open! conn "git.leafac.com" #:ssl? 'secure #:port 443)
(http-conn-sendrecv! conn "/")
(http-conn-live? conn) ; => #f
(http-conn-sendrecv! conn "/") ; => Raise contract violation
```

I believe the correct move is to change `read-dumb-objects` (as well as helper functions, for example, `get-packfile-list`) in `git-checkout.rkt` to use a fresh HTTP connection for each request. For example, using `http-conn-sendrecv` instead of `http-conn-sendrecv!`.

I’d like some feedback before working on a pull request. In particular, from the people that worked on `git-checkout` (@mflatt, @florence, @rfindler and @lexi-lambda) and @jeapostrophe which helped me in the investigation.

jeapostrophe:
Excellent. At least `git-checkout.rkt` could confirm that the connection is still live before trying to use it

mflatt:
Continuing @jeapostrophe's comment: A fresh connection could be noticeably slower for servers that support reusing the connection. Instead of using a fresh connection every time, probably the best approach is to check whether the connection is closed and then reopen it.

Maybe the `conn` should be wrapped with another layer that remembers the host and port, and then a wrapper on `http-conn-sendrecv!` would check whether the connection is still open and reopen as needed each time.

leafac:
Thank you for the feedback. I feel like a wrapper around the `conn` is the best compromise between performance and correctness. But I think it wouldn’t belong in `git-checkout`, but as a separate widget in `http-client`. What’s your opinion?

jeapostrophe:
I agree, `http-conn-open/reuse`?

PR #1618, opened 2017-02-09
Fix “git-checkout”

leafac:
Closes https://github.com/racket/racket/issues/1616

I added `http-conn-sendrecv!/reuse` to `net/http-client`. Using it, the following snippet works:

```racket
#lang racket
(require net/http-client)

(define conn (http-conn))
(http-conn-open! conn "git.leafac.com" #:ssl? 'secure #:port 443)
(http-conn-sendrecv!/reuse conn "git.leafac.com" "/" #:ssl? 'secure #:port 443)
(http-conn-live? conn) ; => #f
(http-conn-sendrecv!/reuse conn "git.leafac.com" "/" #:ssl? 'secure #:port 443)
```

I’m asking for your (@mflatt, @jeapostrophe, and anyone else that is interested) feedback on this approach before I proceed with the following tasks:

**EDIT**: The following tasks became stale, as we decided to go with a different approach. See conversation below.

- [x] Adapt `net/git-checkout` to use `http-conn-sendrecv!/reuse` in `read-dumb-objects`.
- [x] Write automated tests for `http-conn-sendrecv!/reuse`.
- [x] Document `http-conn-sendrecv!/reuse`.

------

@jeapostrophe: I decided to go with `http-conn-sendrecv!/reuse` instead of `http-conn-open/reuse` because it is a smaller change to `http-client`. In particular, to implement `http-conn-open/reuse` I would have to change the `http-conn` struct itself, to keep around the `host` (which is lost on `http-conn-close!`) and `ssl?` (which is never stored in `http-conn` in the first place).

The benefit of this approach is that users **have control** over how the reconnection might happen. For example, they might want change the policy of SSL validation upon reconnection. The problem with this approach is that users **have to control** how the reconnection might happen. In `net/git-checkout`, for example, this means keeping `host`, `ssl?` and `port` around, even after `http-conn-open!`.

------

Thank you for your feedback.

### DIFF
--- a/racket/collects/net/http-client.rkt
+++ b/racket/collects/net/http-client.rkt
@@ -47,17 +47,23 @@
 
 ;; Core
 
-(struct http-conn (host port port-usual? to from abandon-p) #:mutable)
+(struct http-conn (host port port-usual? to from abandon-p
+                        auto-reconnect? auto-reconnect-host auto-reconnect-ssl?) #:mutable)
 
 (define (make-http-conn)
-  (http-conn #f #f #f #f #f #f))
+  (http-conn #f #f #f #f #f #f #f #f #f))
 
 (define (http-conn-live? hc)
   (and (http-conn-to hc)
        (http-conn-from hc)
        #t))
 
-(define (http-conn-open! hc host-bs #:ssl? [ssl? #f] #:port [port (if ssl? 443 80)])
+(define (http-conn-liveable? hc)
+  (or (http-conn-live? hc)
+      (http-conn-auto-reconnect? hc)))
+
+(define (http-conn-open! hc host-bs #:ssl? [ssl? #f] #:port [port (if ssl? 443 80)]
+                         #:auto-reconnect? [auto-reconnect? #f])
   (http-conn-close! hc)
   (define host (->string host-bs))
   (define ssl-version (if (boolean? ssl?) 'auto ssl?))
@@ -97,10 +103,15 @@
   ;; (thread (λ () (copy-port log-i to (current-error-port))))
 
   (set-http-conn-to! hc to)
-  (set-http-conn-from! hc from))
+  (set-http-conn-from! hc from)
+
+  (set-http-conn-auto-reconnect?! hc auto-reconnect?)
+  (set-http-conn-auto-reconnect-host! hc host-bs)
+  (set-http-conn-auto-reconnect-ssl?! hc ssl?))
 
 (define (http-conn-close! hc)
-  (match-define (http-conn host port port-usual? to from abandon) hc)
+  (match-define (http-conn host port port-usual? to from abandon
+                           auto-reconnect? auto-reconnect-host auto-reconnect-ssl?) hc)
   (set-http-conn-host! hc #f)
   (when to
     (close-output-port to)
@@ -111,10 +122,16 @@
   (set-http-conn-abandon-p! hc #f))
 
 (define (http-conn-abandon! hc)
-  (match-define (http-conn host port port-usual? to from abandon) hc)
+  (match-define (http-conn host port port-usual? to from abandon
+                           auto-reconnect? auto-reconnect-host auto-reconnect-ssl?) hc)
   (when to
     (abandon to)
     (set-http-conn-to! hc #f)))
+
+(define (http-conn-enliven! hc)
+  (when (and (not (http-conn-live? hc)) (http-conn-auto-reconnect? hc))
+    (http-conn-open! hc (http-conn-auto-reconnect-host hc) #:ssl? (http-conn-auto-reconnect-ssl? hc)
+                     #:port (http-conn-port hc) #:auto-reconnect? (http-conn-auto-reconnect? hc))))
 
 (define (write-chunk out data)
   (let ([bytes (->bytes data)])
@@ -129,7 +146,9 @@
                          #:headers [headers-bs empty]
                          #:content-decode [decodes '(gzip)]
                          #:data [data #f])
-  (match-define (http-conn host port port-usual? to from _) hc)
+  (http-conn-enliven! hc)
+  (match-define (http-conn host port port-usual? to from _
+                           auto-reconnect? auto-reconnect-host auto-reconnect-ssl?) hc)
   (fprintf to "~a ~a HTTP/~a\r\n" method-bss url-bs version-bs)
   (unless (regexp-member #rx"^(?i:Host:) +.+$" headers-bs)
     (fprintf to "Host: ~a\r\n" 
@@ -240,9 +259,10 @@
 
 ;; Derived
 
-(define (http-conn-open host-bs #:ssl? [ssl? #f] #:port [port (if ssl? 443 80)])
+(define (http-conn-open host-bs #:ssl? [ssl? #f] #:port [port (if ssl? 443 80)]
+                        #:auto-reconnect? [auto-reconnect? #f])
   (define hc (make-http-conn))
-  (http-conn-open! hc host-bs #:ssl? ssl? #:port port)
+  (http-conn-open! hc host-bs #:ssl? ssl? #:port port #:auto-reconnect? auto-reconnect?)
   hc)
 
 (define (http-conn-CONNECT-tunnel proxy-host proxy-port target-host target-port #:ssl? [ssl? #f])
@@ -260,7 +280,8 @@
       (error 'make-ports "HTTP CONNECT failed: ~a" tunnel-status)))
 
   ;; SSL secure the ports
-  (match-define (http-conn _ _ _ t:to t:from _) hc)
+  (match-define (http-conn _ _ _ t:to t:from _
+                           auto-reconnect? auto-reconnect-host auto-reconnect-ssl?) hc)
   (cond [(not ssl?) ; it's just a tunnel... no ssl
          (define abandon-p (lambda (p) ((http-conn-abandon-p hc) p)))
          (values ssl? t:from t:to abandon-p)]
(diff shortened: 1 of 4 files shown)


mflatt:
This is ok with me, but it doesn't seem all that convenient relative to defining a specific `http-conn-sendrecv!/reuse` in modules like `net/git-checkout`.

Another idea to consider, in case you didn't consider something like it already: an `#:auto-reconnect?` argument to `http-conn-open[!]` that would case the host and SSL to be saved in the connection (even after its gets closed) and make `http-conn-send!` and derived functions automatically connect when given a closed connection.

leafac:
First, thank you for your feedback, @mflatt.

> This is ok with me, but it doesn't seem all that convenient relative to defining a specific `http-conn-sendrecv!/reuse` in modules like `net/git-checkout`.

Do you think that `net/git-checkout` would be the only user of `http-conn-sendrecv!/reuse`, so it would be better to have it as internal function in `net/git-checkout`, instead of public API in `net/http-client`?

> Another idea to consider, in case you didn't consider something like it already: an `#:auto-reconnect?` argument to `http-conn-open[!]` that would case the host and SSL to be saved in the connection (even after its gets closed) and make `http-conn-send!` and derived functions automatically connect when given a closed connection.

I considered this design, yes. As far as I understand, that was @jeapostrophe’s suggestion. I didn’t pursue it yet because it’s more invasive and it’s more difficult for me to anticipate all the consequences. But let me try it now:

- [ ] Add fields to `http-conn` struct.
  - `auto-reconnect?`: Whether to automatically reconnect upon `http-conn-send!` and `http-conn-recv!`.
  - `auto-reconnect-host`: The `host` argument with which to call `http-conn-open!` upon automatic reconnection.
  - `auto-reconnect-ssl?`: The `ssl?` argument with which to call `http-conn-open!` upon automatic reconnection.
- [ ] Add an optional argument `#:auto-reconnect?` argument to `http-conn-open[!]`. The default is `#f`, to preserve the current behavior. If `#t`, then populate the `http-conn` with the fields above.
- [ ] Define `http-conn-liveable?` to mean `(or http-conn-live? http-conn-auto-reconnect?)`.
- [ ] Change the contracts of `http-conn-send!`, `http-conn-recv!` and `http-conn-sendrecv!` to require a `http-conn-liveable?` instead of a `http-conn-live?`.
- [ ] Change `http-conn-send!`, `http-conn-recv!` and `http-conn-sendrecv!` to call `http-conn-open!` if necessary.

How does that sound?

jeapostrophe:
Thanks for getting on this @leafac

I think you should put this code in `net/http-client`.

Your plan is good, but a few comments:
- You'll need the port too.
- I suggest making a function like `http-conn-enliven!` that takes a `http-conn-liveable?` and returns a `http-conn-live?` and have `send!`, `recv!`, and so on call that.

leafac:
@jeapostrophe: Thank you for your feedback. I’ll work on my proposal and get some code for you to review soon.

I don’t think I’ll need to handle `port` in any special way. Because `http-conn-open!` [saves the port](https://github.com/racket/racket/blob/master/racket/collects/net/http-client.rkt#L94) in the `http-conn`, and no other function mutates it—including [`http-conn-close!`](https://github.com/racket/racket/blob/master/racket/collects/net/http-client.rkt#L102-L111). Of course, I could store a redundant field just for clarity. What do you think is best?

jeapostrophe:
You're right. I just meant that it will need to use the saved host/port/ssl?/etc, all the arguments to open!

leafac:
@jeapostrophe, @mflatt: I implemented the approach we discussed, it’s on this pull request (the original proposal is still available on https://github.com/racket/racket/compare/master...leafac:fix-git-checkout-1 for comparison). Please let me know if you like it—after your approval I’ll move on to automated tests and documentation.

With this patch, the following snippet works:

```racket
#lang racket
(require net/http-client)

(define conn (http-conn))
(http-conn-open! conn "git.leafac.com" #:ssl? 'secure #:port 443 #:auto-reconnect? #t)
(http-conn-sendrecv! conn "/")
(http-conn-live? conn) ; => #f
(http-conn-sendrecv! conn "/") ; => Works :)
```

And, more importantly, the original issue no longer happens:

```
$ ./racket -l- net/git-checkout --https git.leafac.com pollen-component.git test
Contacting git.leafac.com
Getting refs [dumb protocol]
Server does not support `shallow`
Downloading loose objects
Extracting tree to test
Cleaning up
"8668ea3dc787b1a842baf65e8e10f60e6f5cbc8b"
```

-----

As always, thank you for your time reviewing my contributions.

racket/collects/net/http-client.rkt
```diff
@@ -97,10 +103,15 @@
   ;; (thread (λ () (copy-port log-i to (current-error-port))))
 
   (set-http-conn-to! hc to)
-  (set-http-conn-from! hc from))
+  (set-http-conn-from! hc from)
+
+  (set-http-conn-auto-reconnect?! hc auto-reconnect?)
+  (set-http-conn-auto-reconnect-host! hc host-bs)
```
jeapostrophe: I think we can actually just use the `host` field and don't need `auto-reconnect-host`. Is there something I'm missing?

racket/collects/net/http-client.rkt
```diff
@@ -111,11 +122,17 @@
...
 
 (define (http-conn-abandon! hc)
-  (match-define (http-conn host port port-usual? to from abandon) hc)
+  (match-define (http-conn host port port-usual? to from abandon
+                           auto-reconnect? auto-reconnect-host auto-reconnect-ssl?) hc)
   (when to
     (abandon to)
     (set-http-conn-to! hc #f)))
 
+(define (http-conn-enliven! hc)
```
jeapostrophe: I think this should check for `auto-reconnect?`

racket/collects/net/http-client.rkt
```diff
@@ -409,21 +431,27 @@
   [http-conn-live?
    (-> any/c
        boolean?)]
+  [http-conn-liveable?
```
jeapostrophe: This should be added to the docs

racket/collects/net/http-client.rkt
```diff
@@ -409,21 +431,27 @@
...
         (#:ssl? base-ssl?-tnl/c
-                #:port (between/c 1 65535))
+                #:port (between/c 1 65535)
+                #:auto-reconnect? boolean?)
         void?)]
   [http-conn-close!
    (-> http-conn? void?)]
   [http-conn-abandon!
    (-> http-conn? void?)]
+  [http-conn-enliven!
```
jeapostrophe: This should be added to the docs

racket/collects/net/http-client.rkt
```diff
@@ -409,21 +431,27 @@
...
+   (-> http-conn-liveable? void?)]
   [http-conn-send!
    (->*
-    (http-conn-live? (or/c bytes? string?))
+    (http-conn-liveable? (or/c bytes? string?))
```
jeapostrophe: The contracts need to be updated too

jeapostrophe:
Good work. I left some comments. Is it possible to add a test case with a synthetic server that closes things?

racket/collects/net/http-client.rkt
leafac: I believe you’re missing this: https://github.com/racket/racket/blob/6632beeca98de0dda9b56121ee7ebed517a3a1ba/racket/collects/net/http-client.rkt#L104

The function `http-conn-close!` overwrites the `host` and then it’s lost—we don’t have access to it when trying to reconnect.

racket/collects/net/http-client.rkt
leafac: It does, via its contract. Do you think I should add redundancy?

racket/collects/net/http-client.rkt
leafac: 👍 This is my next task, now that you reviewed the approach.

racket/collects/net/http-client.rkt
leafac: 👍 This is my next task, now that you reviewed the approach.

racket/collects/net/http-client.rkt
leafac: Do you mean, in the contracts in the documentation?

leafac:
@jeapostrophe: Thanks a lot for reviewing my contribution.

I’m going to try to add a test with a synthetic server that closes things now that you confirm my solution is sound.

racket/collects/net/http-client.rkt
leafac: Note, however, that `http-conn-close!` does **not** overwrite `port`, so I was able to reuse that field without having to add a new `auto-reconnect-port` field to `http-conn`.

jeapostrophe: You're right

racket/collects/net/http-client.rkt
jeapostrophe: I do, because it is not only called externally.

jeapostrophe: Specifically, it shouldn't error but should just not do anything

racket/collects/net/http-client.rkt
jeapostrophe: Ya, I imagine you planned on that :)

jeapostrophe:
Excellent

racket/collects/net/http-client.rkt
leafac: Got it. Done.

leafac:
I wrote documentation and automated tests. This pull request is now ready for review.

jeapostrophe:
Excellent work. Thank you!

leafac:
Thank you for your review and for merging my contribution.

PR #1623, opened 2017-02-28
Numeric Contracts

LeifAndersen:
Add the following for contracts to `racket/contracts`

positive-integer?
negative-integer?
nonnegative-integer?
nonpositive-integer?

These are exactly like their `exact-*` counterparts, but also work on inexact? numbers, like their typed counterparts.

Only two packages currently seem to use any of these 4 predicates, rsound and video, both of which are easily fixed.


samth:
Does the last comment mean that adding these breaks `rsound` and `video`?

LeifAndersen:
Adding any new function to a Racket generally breaks any packages that have an define an identifier with the same name. The `rsound` and `video` packages are the only ones (on the package server) that use those identifiers, both for this exact purpose. And honestly, it only takes changing 3 lines of code to fix the packages to make them work both with versions of racket before and after those contracts are added.

LeifAndersen:
In fact, its so easy I have fixed both of them, and going to submit a PR for `rsound` momentarily.

LeifAndersen:
I mean, I wouldn't be opposed to putting them in a new library, like `racket/contract/numeric` or something like that. I just put them there so they would be at the same scope as >/c </c etc.

rfindler:
I don't think these belong in racket/contract.

LeifAndersen:
@rfindler okay, why?

rfindler:
I don't think they have anything to do with racket/contract so
racket/contract/X doesn't seem right either.

They go in a number related library, to me.

I mean, predicates for other datastructures go in the library where those
other data structures go, right?

Robby

LeifAndersen:
@rfindler So, on the one hand I agree with you, it feels kind of odd to have them in `racket/contract`. On the other hand though, we have `</c`, `>/c`, `<=/c`, `>=/c`, `between/c`, etc. already in `racket/contract`, so it makes sense to have them there for consistency.

jackfirth:
A math-related contracts library could reprovide those combinators, with the eventual goal of removing them from `racket/contract`.

LeifAndersen:
I mean, we 'could' add them to `racket/base` somewhere, like their `exact-*` counterparts, and that would probably break no code. I just didn't do that because currently the `exact-*` counterparts are implemented in C.

LeifAndersen:
@jackfirth In principle I agree. Although I'm usually weary of removing exports from libraries.

rfindler:
If you write a new list predicate, should it go in racket/contract? If you write a new function predicate that works with arities, does it go in racket/contract? If you write a new predicate on classes, does it go in racket/contract? No to all of those, in my opinion. The contract library has a few convenient combinators for primitive things like numbers, but that's different than putting in random predicates for every possible base value we might have in some other library. They can define the predicate themselves and those predicate will work fine with the racket/contract library. 

Think of it this way: if you were looking for functions related to numbers, why would you look in the contract library?

LeifAndersen:
> a new list predicate

Depends on the predicate. We already have list/c and listof and list*of etc. So it might make sense to have others in there, such as list-assoc/c or something like that. So that analogy fails.

> new function predicate that works with arities

No, there are neither contracts combinations for functions nor flat contracts dealing with functions (that are documented) in racket/contract. So it would be silly to put a predicate there.

On the other hand,, the contract library already *does* have both contract combinators that handles numbers, as well as direct flat contracts for numbers, such as natural/c. So it is not surprising to have flat contracts for those cases as well. (Although maybe it would make sense to change the name to positive-integer/c instead, which the more I think about it, the more I think that is the correct choice here, so I'm going to make a commit changing those names).

> a new predicate on classes

No, of course not. Contracts for the class library are in `racket/class`, so from a design perspective, it would make sense to put predicates for classes there.  From an implementation perspective I think the class library relies on the contract one, so we couldn't do that even if we wanted to.

> Think of it this way: if you were looking for functions related to numbers, why would you look in the contract library?

I'm not looking for functions related to numbers, I'm looking for *contracts* related to numbers. I want this because I have several libraries that have contracts for all of the functions that require nonnegative integers, both exact and inexact. So much so that it makes sense to have them in a common library. And given that I'm not the only person who has run into this, and it seems to be a common contract pattern, it makes sense to put it in the contract library, as the contract library clearly handles number related contracts, both conbinators and flat contracts.

LeifAndersen:
Okay. Switching the names to *-integer/c seems to fit a lot better now, both in terms of the rest of the `racket/contract` api, as @rfindler pointed out. Also, it does not break any of the projects which gets rid of the concern @samth had.

rfindler:
Thank you for your interest in racket and contributions to it. I remain unconvinced about this particular change, however, so I will close this PR now.

I will note that I find both your arguments for why it belongs unconvincing and I also find your arguments why it could not go elsewhere unconvincing. (I didn't talk about them much, because that's not what this PR is doing, however.)

LeifAndersen:
So then, explain to me why natural-number/c belongs in racket/contract but positive-integer/c does not?

rfindler:
I wrote before you asked (I'm not sure why you didn't get it) that natural-number/c was a mistake.

LeifAndersen:
Ah, okay. Yes, that appears to have gotten lost...yay github and their crappy interface. :(

So, @rfindler , if your argument is that natural-number/c was a mistake and is really on there because of backwards compatibility, than I agree we should not add these contracts here, but somewhere else. (Probably in either base or math....but that's for a different PR.)

Anyway, thanks.

rfindler:
Thank you.

(I do wish we had a `natural?` or maybe even `nat?` predicate.)

PR #1626, opened 2017-03-04
relax contract for (random min max)

bennn:
Relax contract for `(random min max)` to accept any pair of integers such that:
- `(< min max)`
- `(- max min)` is between 1 and 4294967087

### DIFF
--- a/racket/collects/racket/private/pre-base.rkt
+++ b/racket/collects/racket/private/pre-base.rkt
@@ -98,18 +98,6 @@
   (define-values (double-flonum?) ; for symmetry with single-flonum?
     (lambda (x) (flonum? x)))
 
-  (define-values (enforce-random-int-range)
-    (lambda (x)
-      (unless (and (exact-positive-integer? x)
-                   (<= x 4294967087))
-        (raise-argument-error 'random "(integer-in 1 4294967087)" x))))
-  (define-values (enforce-greater)
-    (lambda (x y)
-      (unless (> y x)
-        (raise-argument-error
-         'random
-         (string-append "integer greater than " (number->string x))
-         y))))
   (begin-encourage-inline
    (define-values (-random) ; more featureful than #%kernel's `random`
      (let ([random ; to get the right name
@@ -121,26 +109,54 @@
                (random x)]
               [(x y)
                ;; two args, either max and prng, or min and max
-               (cond [(exact-positive-integer? y) ; min and max case
-                      (enforce-random-int-range x)
-                      (enforce-random-int-range y)
-                      (enforce-greater x y)
-                      (+ x (random (- y x)))]
+               (cond [(exact-integer? y) ; min and max case
+                      (unless (exact-integer? x)
+                        (raise-argument-error 'random "exact-integer?" 0 x y))
+                      (unless (< x y)
+                        (raise-argument-error
+                         'random
+                         (string-append "(>/c " (number->string x) ")")
+                         1 x y))
+                      (let ([d (- y x)])
+                        (unless (<= d 4294967087)
+                          (raise-arguments-error
+                           'random
+                           "difference between arguments is greater than 4294967087"
+                           "min" x "max" y))
+                        (+ x (random d)))]
                      [(pseudo-random-generator? y) ; int and prng case
-                      (enforce-random-int-range x)
+                      (unless (and (exact-integer? x)
+                                   (<= 1 x 4294967087))
+                        (raise-argument-error
+                         'random "(integer-in 1 4294967087)" 0 x y))
                       (random x y)]
                      [else
+                      (unless (exact-integer? x)
+                        (raise-argument-error 'random "exact-integer?" 0 x y))
                       (raise-argument-error
                        'random
-                       "(or/c (integer-in 1 4294967087) pseudo-random-generator?)"
-                       y)])]
+                       "(or/c exact-integer? pseudo-random-generator?)"
+                       1 x y)])]
               [(min max prng) ; three args: min, max, and prng
-               (enforce-random-int-range min)
-               (enforce-random-int-range max)
-               (enforce-greater min max)
-               (unless (pseudo-random-generator? prng)
-                 (raise-argument-error 'random "pseudo-random-generator?" prng))
-               (+ min (random (- max min) prng))])])
+               (unless (exact-integer? min)
+                 (raise-argument-error 'random "exact-integer?" 0 min max prng))
+               (unless (exact-integer? max)
+                 (raise-argument-error 'random "exact-integer?" 1 min max prng))
+               (unless (< min max)
+                 (raise-argument-error
+                  'random
+                  (string-append "(>/c " (number->string min) ")")
+                  1 min max prng))
+               (let ([d (- max min)])
+                 (unless (<= d 4294967087)
+                   (raise-arguments-error
+                    'random
+                    "difference between first and second arguments is greater than 4294967087"
+                    "min" min "max" max "rand-gen" prng))
+                 (unless (pseudo-random-generator? prng)
+                   (raise-argument-error
+                    'random "pseudo-random-generator?" 2 min max prng))
+                 (+ min (random d prng)))])])
        random)))
 
   (define-values (new:collection-path)
--- a/pkgs/racket-doc/scribblings/reference/numbers.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/numbers.scrbl
@@ -842,8 +842,8 @@ both in binary and as integers.
                     [rand-gen pseudo-random-generator?
                                (current-pseudo-random-generator)])
             exact-nonnegative-integer?]
-           [(random [min (integer-in 1 4294967087)]
-                    [max (integer-in 1 4294967087)]
+           [(random [min exact-integer?]
+                    [max (integer-in (+ 1 min) (+ 4294967087 min))]
                     [rand-gen pseudo-random-generator?
                               (current-pseudo-random-generator)])
             exact-nonnegative-integer?]
--- a/pkgs/racket-test-core/tests/racket/number.rktl
+++ b/pkgs/racket-test-core/tests/racket/number.rktl
@@ -2462,9 +2462,15 @@
 (test (begin (random-seed 23) (list (random 10) (random 20) (random 30)))
       'random-seed-same
       (begin (random-seed 23) (list (random 10) (random 20) (random 30))))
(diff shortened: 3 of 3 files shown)


bennn:
Here's a sample error message:
```
> (random -1 (expt 2 40))
; random: (- max min) difference is not an (integer-in 1 4294967087)
;   min: -1
;   max: 1099511627776
; [,bt for context]
```

Here's another error message:
```
> (random 1 "three")
; random: contract violation
;   expected: (or/c exact-integer? pseudo-random-generator?)
;   given: "three"
```

Here's a picture of the docs:
screen shot 2017-03-03 at 20 47 07

I wasn't sure what to do about the contract. I want the requirement on `(- max min)` to be in contract (rather than just the text) but didn't want to make an `->i` contract.

rfindler:
Just refer to the variable: (<=/c (- bignumber min)) is Max's contract.

I would not refer to the variables like that in the error tho. "Fist and
second argument" is better there.

Robby

jbclements:
LGTM. Robby, you click the button, is that right?

rfindler:
Two requests:

- please call `raise-arguments-error` with all supplied arguments when any of them are wrong.

- please squash the commits.

@bennn can you push? If so, please go ahead after that.

bennn:
@mflatt @gus-massa would it be ok to replace calls to `enforce-random-int-range` with calls to the new function with the offset? (By passing 0 as the second argument, or changing `enforce-random-int-range/offset` to make the second argument optional.) I'm worried because inlining.

Also I just noticed this PR doesn't check `exact-integer?` enough, fixing now.

bennn:
Nevermind about inlining, the function's only called once so I'm just going to make the change (call `enforce.../offset` instead)

@robby yep I'll use `raise-arguments-error`. Will plan to push late tonight, unless anything else comes up.

bennn:
@rfindler do you also want the `rand-gen` argument in the error message? Or is the idea: "show all values involved in the check that led to the error message"

rfindler:
Show all argument supplied to the function that was actually called (in the error message) is the right generalization. That is, if I call `(random 0)` I should see one value and if I call `(random 0 (current-pseudo-random-generator))` I should see two.

Of course, sometimes this is too much of a pain to be worth it. In that case, call `raise-argument-error`.

bennn:
Made all the error messages use `raise-arguments-error` _except_ when given 2 arguments and the second is neither an `exact-integer?` or a PRNG. Because in that case, I didn't know what name to give the arguments.

Here's some example error messages. Suggestions?
```
> (define rng (current-pseudo-random-generator))

> (random "1")
; random: contract violation
;   expected: (or/c (integer-in 1 4294967087) pseudo-random-generator?)
;   given: "1"

> (random -1)
; random: contract violation
;   expected: (or/c (integer-in 1 4294967087) pseudo-random-generator?)
;   given: -1

> (random (add1 4294967087))
; random: contract violation
;   expected: (or/c (integer-in 1 4294967087) pseudo-random-generator?)
;   given: 4294967088

> (random "1" 2)
; random: first argument is not exact-integer?
;   min: "1"
;   max: 2

> (random "1" "2")
; random: contract violation
;   expected: exact-integer?
;   given: "1"

> (random 1 "2")
; random: contract violation
;   expected: (or/c exact-integer? pseudo-random-generator?)
;   given: "2"

> (random -1 "2")
; random: contract violation
;   expected: (or/c exact-integer? pseudo-random-generator?)
;   given: "2"

> (random 2 1)
; random: second argument is not greater than first argument
;   min: 2
;   max: 1

> (random -1 -2)
; random: second argument is not greater than first argument
;   min: -1
;   max: -2

> (random 0 (add1 4294967087))
; random: difference between first and second arguments is greater than
;   4294967087
;   min: 0
;   max: 4294967088

> (random "1" rng)
; random: first argument must be (integer-in 1 4294967087)
;   k: "1"
;   rand-gen: #<pseudo-random-generator>

> (random -1 rng)
; random: first argument must be (integer-in 1 4294967087)
;   k: -1
;   rand-gen: #<pseudo-random-generator>

> (random (add1 4294967087) rng)
; random: first argument must be (integer-in 1 4294967087)
;   k: 4294967088
;   rand-gen: #<pseudo-random-generator>

> (random "1" 2 rng)
; random: first argument is not exact-integer?
;   min: "1"
;   max: 2
;   rand-gen: #<pseudo-random-generator>

> (random "1" "2" rng)
; random: first argument is not exact-integer?
;   min: "1"
;   max: "2"
;   rand-gen: #<pseudo-random-generator>

> (random 1 "2" rng)
; random: second argument is not exact-integer?
;   min: 1
;   max: "2"
;   rand-gen: #<pseudo-random-generator>

> (random -1 "2" rng)
; random: second argument is not exact-integer?
;   min: -1
;   max: "2"
;   rand-gen: #<pseudo-random-generator>

> (random (add1 4294967087) "2" rng)
; random: second argument is not exact-integer?
;   min: 4294967088
;   max: "2"
;   rand-gen: #<pseudo-random-generator>

> (random 2 1 rng)
; random: second argument is not greater than first argument
;   min: 2
;   max: 1
;   rand-gen: #<pseudo-random-generator>

> (random -1 -2 rng)
; random: second argument is not greater than first argument
;   min: -1
;   max: -2
;   rand-gen: #<pseudo-random-generator>

> (random 0 (add1 4294967087) rng)
; random: difference between first and second arguments is greater than
;   4294967087
;   min: 0
;   max: 4294967088
;   rand-gen: #<pseudo-random-generator>
```

(cc @rfindler @jbclements @danprager )

danprager:
This is looking good.

Strictly speaking, is the maximum range constant a property of the particular random-number generator employed?

bennn:
Yep, see section 1.1 here: https://www.iro.umontreal.ca/~lecuyer/myftp/papers/streams00.pdf

danprager:
Thanks. I see now that the class of rng's is fixed (L’Ecuyer), so the constant is ok.

[I was imagining that it might be possible to plug in other rng algorithms.]

rfindler:
@bennn I feel stupid; I wrote `raise-arguments-error` but what I meant was `raise-argument-error` where you use the second alternative (with `bad-pos` and all of the arguments supplied). 

What you've done is certainly better! But in the case you found problematic, you could just take the `bad-pos` approach.

bennn:
Ok, I updated some errors* to use `bad-pos`. (I didn't realize that version of `raise-argument-error` existed!)
"LGTM"

(* Some errors = the errors where it was awkward to name arguments)

```
(random "1" "1")
;; ==> 
;; random: contract violation
;;   expected: exact-integer?
;;   given: "1"
;;   argument position: 1st
;;   other arguments...:
;;    "1"
```

Here's a few more examples in the toplevel:
```
> (random "1" 1)
; random: contract violation
;   expected: exact-integer?
;   given: "1"
;   argument position: 1st

> (random 1 "1")
; random: contract violation
;   expected: (or/c exact-integer? pseudo-random-generator?)
;   given: "1"
;   argument position: 2nd

> (random 3 2)
; random: contract violation
;   expected: (>/c 3)
;   given: 2
;   argument position: 2nd

> (random 3 2 (current-pseudo-random-generator))
; random: contract violation
;   expected: (>/c 3)
;   given: 2
;   argument position: 2nd

> (random 1 2 "rng")
; random: contract violation
;   expected: pseudo-random-generator?
;   given: "rng"
;   argument position: 3rd
```

racket/collects/racket/private/pre-base.rkt
```diff
@@ -121,26 +109,54 @@
...
+                        (+ x (random d)))]
                      [(pseudo-random-generator? y) ; int and prng case
-                      (enforce-random-int-range x)
+                      (unless (and (exact-integer? x)
+                                   (<= 1 x 4294967087))
+                        (raise-argument-error
+                         'random "(integer-in 1 4294967087)" 0 x y))
                       (random x y)]
                      [else
+                      (unless (exact-integer? x)
```
bennn: This check is new, and maybe unnecessary, and incidentally the only duplicate call to `raise-...-error` (all the others have unique error messages or arguments). 

But it's here because it makes sense to me that `(random "a" "b")` blames its first argument.

rfindler:
Are you passing all the arguments? I would have expected them to all show up in the error message, not just the first one.

bennn:
Yes, I'm passing all the arguments in every error message. But they don't all appear when I run examples in the toplevel.

Example: `(random "1" "1")` (error raised on [L135](https://github.com/racket/racket/pull/1626/files#diff-13eaec578e818e85ad4749dc69a4dd57R135) of `pre-base.rkt`)
in the toplevel:
```
; random: contract violation
;   expected: exact-integer?
;   given: "1"
;   argument position: 1st
```

when I save to a file and run `racket file.rkt`
```
random: contract violation
  expected: exact-integer?
  given: "1"
  argument position: 1st
  other arguments...:
   "1"
  context...:
   /Users/ben/code/racket/fork/racket/collects/racket/private/pre-base.rkt:104:12: random
   /Users/ben/code/racket/plan-b/model/file.rkt: [running body]
```

rfindler:
Disable your repl customizations and see if they come back. This sounds
like a bug or a poor design decision (imo).

Robby

mflatt:
With xrepl (now on by default in REPL mode), you have to use `,bt` to see any message field that end "...". The output only says "[,bt for context]", though.

bennn:
(Guess I shouldn't have deleted those `[,bt for context]` lines from my examples)

rfindler:
I guess the casual user will not realize that more information is available to them. And we almost certainly won't get that very useful information in our bug reports.

Can we do something slightly differently? Maybe could we have the lines with the ellipses show up and say ",bt for more information, including the elided parts of the message" or something like that?

bennn:
Do you think I should change uses of `raise-argument-error` to be `raise-arguments-error .... "first-arg" x "second-arg" y)` etc.? I would rather keep the current messages.

fwiw, DrRacket also doesn't show the context. I need to click the red "triple X" to see what the "other arguments" are.

rfindler:
On Mon, Mar 6, 2017 at 5:22 PM, Benjamin Greenman <notifications@github.com>
wrote:

> Do you think I should change uses of raise-argument-error to be raise-arguments-error
> .... "first-arg" x "second-arg" y) etc.? I would rather keep the current
> messages.
>

I like what you currently have and think you've done a great service and it
is time to stop belaboring this! Please push away.


> fwiw, DrRacket also doesn't show the context. I need to click the red
> "triple X" to see what the "other arguments" are.
>
>
Believe it or not, I knew that. :)

Robby

Issue #1622, opened 2017-02-20
Docs for max / argmax should link to each other


jackfirth:
If one doesn't know about [`argmax`](http://docs.racket-lang.org/reference/pairs.html?q=argmax#%28def._%28%28lib._racket%2Flist..rkt%29._argmax%29%29), it's hard to find from the docs for [`max`](http://docs.racket-lang.org/reference/generic-numbers.html?q=max#%28def._%28%28quote._~23~25kernel%29._max%29%29) as they're in completely separate sections and don't mention each other. A few "see also" notes in the docs for max/min/argmax/argmin would improve discoverability.

stamourv:
Excellent idea. Would you like to submit a PR?

jackfirth:
When I get time :) I have to go through some internal bureaucracy for PRs to ensure I own the copyright instead of my employer. Makes it difficult to find the time for simple one-off changes like this.

bennn:
Wow.
Would you be ok with someone submitting a PR for this on your behalf?

jackfirth:
It is what it is. Go right ahead!

bennn:
https://github.com/racket/racket/commit/10aaff692b55877952ece746b1fb2c05ef9c97dd

PR #1632, opened 2017-03-11
[doc] link max <--> argmax

bennn:
This is the right look, but is it possible to use a `seclink` instead of the `for-label` import?

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/numbers.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/numbers.scrbl
@@ -5,7 +5,8 @@
                      racket/fixnum
                      racket/unsafe/ops
                      racket/require
-                     racket/random))
+                     racket/random
+                     racket/list))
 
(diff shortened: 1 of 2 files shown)


bennn:
cc #1622 

previews:
screen shot 2017-03-11 at 13 35 24
screen shot 2017-03-11 at 13 35 44

rfindler:
I think we usually put the "see also"s into the description, not in the
sidebars.

Robby

bennn:
New commit, new previews:

screen shot 2017-03-11 at 14 00 28
screen shot 2017-03-11 at 14 04 16

bennn:
One more, in standard "see also" style.

screen shot 2017-03-11 at 14 10 16
screen shot 2017-03-11 at 14 10 27

bennn:
@rfindler ping, do you like the most recent revision?

rfindler:
Yes

PR #1634, opened 2017-03-15
Add procedures like port->X but close the port

m4burns:
I have found these useful.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/port-lib.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/port-lib.scrbl
@@ -25,29 +25,38 @@ on @racket[in] until it produces @racket[eof].
 (port->list read-number (open-input-string "12345"))
 ]}
 
-@defproc[(port->string [in input-port? (current-input-port)]) string?]{
+@defproc[(port->string [in input-port? (current-input-port)]
+                       [#:close? close? any/c #f])
+         string?]{
 
 Reads all characters from @racket[in] and returns them as a string.
+The input port is closed unless @racket[close?] is @racket[#f].
+
 @examples[#:eval port-eval
 (port->string (open-input-string "hello world"))
 ]}
 
-@defproc[(port->bytes [in input-port? (current-input-port)]) bytes?]{
+@defproc[(port->bytes [in input-port? (current-input-port)]
+                      [#:close? close? any/c #f])
+         bytes?]{
 
 Reads all bytes from @racket[in] and returns them as a @tech{byte string}.
+The input port is closed unless @racket[close?] is @racket[#f].
 
 @examples[#:eval port-eval
 (port->bytes (open-input-string "hello world"))
 ]}
 
 @defproc[(port->lines [in input-port? (current-input-port)]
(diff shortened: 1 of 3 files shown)


samth:
I think these might be nicer as optional keyword arguments to the existing functions. What do you think?

m4burns:
Switched to optional keyword arguments (rebase)

pkgs/racket-doc/scribblings/reference/port-lib.scrbl
```diff
@@ -25,41 +25,52 @@ on @racket[in] until it produces @racket[eof].
 (port->list read-number (open-input-string "12345"))
 ]}
 
-@defproc[(port->string [in input-port? (current-input-port)]) string?]{
+@defproc[(port->string [in input-port? (current-input-port)]
+                       [#:close? close? any/c #f])
+         string?]{
 
 Reads all characters from @racket[in] and returns them as a string.
+When @racket[close?] is @racket[#t], the input port is closed.
```
bennn: Should these say `true` instead of `@racket[#t]`?

Most places I looked in the docs say "If .... is `#f` .... otherwise ....", but [`copy-directory/files`](http://docs.racket-lang.org/reference/Filesystem.html#%28def._%28%28lib._racket%2Ffile..rkt%29._copy-directory%2Ffiles%29%29) says `true`

samth: I would say "The input port is closed unless `close?` is `#f`."

m4burns: Thank you, updated to @samth 's suggestion

samth:
Looks good. Can you squash?

pkgs/racket-test-core/tests/racket/portlib.rktl
```diff
@@ -14,14 +14,29 @@
   (test '(#\h #\e #\l #\l #\o #\return #\newline #\t #\h #\e #\r #\e)
 	port->list read-char (p))
   (test "hello\r\nthere" port->string (p))
+  (test "hello\r\nthere" port->string/close (p))
```
bennn: Is `port->string/close` defined here? (I was waiting for the build to finish to see, but might as well ask)

m4burns: Oh, it's not defined there. Tests passed locally (run with `racket test.rkt` in this directory), so something's up. I'll make sure I'm actually running these tests.

m4burns: I was using the wrong racket executable. Removed the offending lines and confirmed tests pass locally. Thanks!

m4burns:
Squashed.

samth:
Looks good once CI finishes.

m4burns:
All the Travis jobs have terminated with "The job exceeded the maximum time limit for jobs, and has been terminated."

samth:
Ah, that's because of the web server test issue, but this is ok anyway.

PR #1635, opened 2017-03-15
(Attempt to) fix license and acknowledgements links on docs.racket-lang.org.

### LINKED ISSUE #1619: License and Acknowledgement page broken links
When going to the License page on https://docs.racket-lang.org/, the links to License and Acknowledgement are broken. Same if you go to Acknowledgements.

stamourv:
I'm not sure that's the right solution, though. Also can't really test locally.

May close #1619.


mflatt:
I don't think this is the right repair, because docs rendered to user use space should point to the installation's license and acknowledgements.

Something is wrong with the installers (both release and snapshots) where full paths end up in "license/index.html" and "acknowledgements/index.html". I'm looking into that problem.

PR #1642, opened 2017-03-23
gzip: use path-add-extension as default

bennn:
Instead of `string-append`-ing the extension, because that doesn't work for all path strings

### DIFF
--- a/pkgs/racket-doc/file/scribblings/gzip.scrbl
+++ b/pkgs/racket-doc/file/scribblings/gzip.scrbl
@@ -8,13 +8,17 @@ utilities to create archive files in @exec{gzip} format, or simply to
 compress data using the @exec{pkzip} ``deflate'' method.}
 
 @defproc[(gzip [in-file path-string?]
-               [out-file path-string? (string-append in-file ".gz")])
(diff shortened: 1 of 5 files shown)


mflatt:
LGTM

bennn:
Right. (I should have said that in a comment.)

What about adding an optional byte string argument to `path-add-extension` to  replace `_`?

mflatt:
That addition sounds ok to me.

bennn:
- changed `path-add-extension`to accept an optional 3rd argument
- `gzip` uses the 3rd argument, so zipping `x.tar` gives `x.tar.gz`
- added `@history` annotations for both changes

Docs building now, will post screenshots when they're finished

bennn:
Is `6.8.0.2` the right version to cite?

screen shot 2017-03-25 at 20 31 58
screen shot 2017-03-25 at 20 34 11

mflatt:
This looks ready to push to me.

bennn:
Thanks for looking it over; merged via https://github.com/racket/racket/commit/2cbd44d64d815faeb939421cf448fbe2ae7635d0 and https://github.com/racket/racket/commit/30aa951db886dd4e6ac4d1630a0d2e42d035085b

PR #1644, opened 2017-03-23
generics: optional scope argument for private macros

AlexKnauth:
To make them more friendly to macros that expand to generics.

This is related to my [email on the racket-dev list][racket-dev-email] on macros expanding to generics, and would allow me to solve my problem there.

Currently, the `gen:interface` identifier serves two purposes: getting the info for the generic interface, and setting the scope for the method implementations. This pull request separates these two purposes into two separate identifiers, so that macro implementors can choose to pass a scope identifier with different scopes from the generic-interface identifier.

  [racket-dev-email]: https://groups.google.com/forum/#!topic/racket-dev/JAXrLTEOrpE

### DIFF
--- a/pkgs/racket-test/tests/generic/method-table.rkt
+++ b/pkgs/racket-test/tests/generic/method-table.rkt
@@ -1,0 +1,78 @@
+#lang racket/base
+
+(require racket/generic
+         racket/private/generic-methods
+         rackunit)
+
+;; ---------------------------
+;; Without #:scope argument
+
+(check-equal? (generic-method-table gen:equal+hash
+                (define equal-proc 'a)
+                (define hash-proc 'b)
+                (define hash2-proc 'c))
+              '#(a b c))
+
+;; missing implementation for hash2-proc filled in with false
+(check-equal? (generic-method-table gen:equal+hash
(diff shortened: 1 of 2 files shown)


AlexKnauth:
Some of the Travis builds are failing because they're having trouble connecting to `www.unicode.org`.

mflatt:
This looks like a fine change to me (and just needs docs and tests to merge).

AlexKnauth:
This only modifies undocumented internal macros, so what would I do for docs?

I'll add some tests.

stchang:
> what would I do for docs?

Perhaps a better question would be, there's a few examples now that use forms from `private/generic-methods`. Would some of those forms be worth moving to `racket/generic` and documenting them?

mflatt:
I agree with @stchang. I didn't catch that a private module is being used, so maybe it's ok to just have tests for now, but it sounds like something needs to be made public and documented.

AlexKnauth:
Should it be made public in a separate pull request?

mflatt:
Yes, a separate PR sounds good

AlexKnauth:
So should this one be merged?

mflatt:
Yes

Issue #1656, opened 2017-04-02
build-flat-contract-property rejects #:generate keyword


LiberalArtist:
This program:
```
#lang racket
(struct 42/c ()
  #:property prop:flat-contract
  (build-flat-contract-property
   #:first-order (λ (c) (λ (v) (equal? 42 v)))
   #:generate (λ (c) (λ (fuel) (λ () 42)))))
```
raises the following error:
```
application: procedure does not expect an argument with given keyword
  procedure: build-flat-contract-property
  given keyword: #:generate
  arguments...:
   #:first-order #<procedure>
   #:generate #<procedure>
```

bennn:
Should be fixed in the next nightly build, and on Racket 6.9. Thanks for the report!

https://github.com/racket/racket/commit/3bb131ecb230a8670465c9af8de89c8283e41005

PR #1657, opened 2017-04-02
add keywords to (the public) `build-flat-contract-property`

bennn:
#1656

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/random-generate.rkt
+++ b/pkgs/racket-test/tests/racket/contract/random-generate.rkt
@@ -364,3 +364,49 @@
            (λ (x) (if x 'fail 11))
            'pos
            'neg))
+
+(let () ;; test generate / exercise for `build-flat-contract-property contracts
+  (define even-list/c
+    (let ()
+      (struct ctc ()
+        #:property
+        prop:flat-contract
+        (build-flat-contract-property
+         #:name (λ (c) 'even-list/c)
+         #:first-order (λ (c) (λ (v) (and (list? v) (andmap even? v))))
+         #:late-neg-projection
+         (λ (c)
+           (λ (b)
+             (λ (v neg-party)
+               (unless (and (list? v) (andmap even? v))
+                 (raise-blame-error b v
+                                    #:missing-party neg-party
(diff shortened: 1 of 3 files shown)


pkgs/racket-test/tests/racket/contract/random-generate.rkt
```diff
@@ -364,3 +364,53 @@
...
+             (λ () '(2))))
+         #:exercise
+         (λ (c)
+           (λ (fuel)
+             (values (lambda (c) (void)) '())))))
+      (ctc)))
+  (check-exn cannot-generate-exn? (λ () (test-contract-generation even-list/c)))
+  (check-not-exn (λ () (test-contract-generation even-list/c/generate)))
+  (check-exercise 2 void? even-list/c)
+  (check-exercise 2 void? even-list/c/generate))
```
bennn: This last check is boring, because I didn't see how to make an interesting `#:exercise` for `even-list/c`. Any suggestions?

rfindler: there is nothing interesting to do for exercise with flat contracts.

rfindler:
I think a better thing to do here is to add the `#:generate` argument to the function and remove the `#:exercise` argument from the docs. Otherwise, this looks great to me.

Thanks!

bennn:
removed `#:exercise`, added a `@history` note
https://github.com/racket/racket/commit/3bb131ecb230a8670465c9af8de89c8283e41005

stamourv:
Just to confirm, are you requesting that this commit go into 6.9?

bennn:
Yes,

but Robby emailed me that the `@history` may be misleading (as the keyword was only removed from the docs, not the function). So that commit may require a follow-up.

stamourv:
Ok. Let me know which commits to cherry-pick when you're ready.

bennn:
@stamourv the commits are:
- 3bb131ecb230a8670465c9af8de89c8283e41005
- 50f67cf6e48c4e9fdfa9ce09117b4dde13cf4da1
thank you!

stamourv:
Done. Thanks!

PR #1658, opened 2017-04-03
doc: raise-blame-error's #:missing-party

bennn:
Add documentation for the `#:missing-party` optional argument.

Should I take out the parenthetical?


bennn:
ping @stamourv, I think this and #1657 should be in Racket v6.9

rfindler:
I will get there. Chill.

rfindler:
The missing party is not necessarily negative.

PR #1660, opened 2017-04-06
Provide srcloc when constructing `define` in `define/contract`.

samth:


samth:
Does this look right @rfindler?

rfindler:
What are you trying to achieve? Is there some program that behaves badly?

Robby

samth:
This program:

```racket
#lang racket
(define/contract ((f x) y) any/c (+ x y))
((f 1) 2 3)
```

now names the second procedure correctly, producing this error message:

```
/tmp/b.rkt:3:0: arity mismatch;
  expected: 1
  given: 2
  arguments...:
```

Old error message:
```
.../norm-define.rkt:53:83: arity mismatch;
  expected: 1
  given: 2
  arguments...:
```

samth:
Unfortunately, I haven't been able to add something to the contract test suite to effectively test this, perhaps because the good behavior only happens when the function is defined in a saved file.

rfindler:
Maybe something like this?

```
(parameterize ([current-namespace (make-base-namespace)])
  (namespace-require 'racket/base)
  (namespace-require 'racket/contract)
  (eval (datum->syntax #f
                       '(define/contract ((f x) y) any/c (+ x y))
                       (vector "x.rkt" 1 0 1 0)))
  (printf "~s\n" (eval '(object-name (f 1))))
  (eval (datum->syntax #f '(define ((g x) y) y) (vector "x.rkt" 1 0 1 0)))
  (printf "~s\n" (eval #`(object-name (g 1)))))
```

samth:
Thanks, I added that as a test in `test/racket/contract/define-contract`.

rfindler:
Maybe the last two lines of the test case aren't needed?

Otherwise, looks good to push to me.

Thanks!

PR #1663, opened 2017-04-11
Corrected Racket Reference for get/build-late-neg-projection …

LiberalArtist:
and `get/build-val-first-projection`.

Both had incorrectly shown that the returned function took a `contract?` as an extra first argument.


rfindler:
It looks to me like the `c` argument is the contract. In other words, there is only one, not contracts?

LiberalArtist:
I think so, if I understand the question correctly.

rfindler:
Oh, stupid me. I misread the diff. Sorry.

PR #1664, opened 2017-04-11
Revise Racket Guide 7.8 Building New Contracts

LiberalArtist:
to use late neg projections in the examples, rather val first projections.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/contracts/new-combinators.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/contracts/new-combinators.scrbl
@@ -10,7 +10,9 @@
 Contracts are represented internally as functions that
 accept information about the contract (who is to blame,
 source locations, @|etc|) and produce projections (in the
-spirit of Dana Scott) that enforce the contract. A
+spirit of Dana Scott) that enforce the contract.
+
+In a general sense, a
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/guide/contracts/new-combinators.scrbl
```diff
@@ -179,23 +182,21 @@ to use this API looks like this:
...
+          (check-int ans rng-blame neg-party)
+          ans)
+        (raise-blame-error
+         blame #:missing-party neg-party
+         f
+         '(expected "a procedure of one argument" given: "~e")
+         f))))]
 The advantage of this style of contract is that the @racket[_blame]
-and @racket[_v] arguments can be supplied on the server side of the
+and @racket[_f] arguments can be supplied on the server side of the
```
LiberalArtist: I think this part might need to change: it looks like the `f` argument has to be supplied for each client, but I'm not very confident.

rfindler:
Welp, looking this over it occurs to me that it is building wrapper functions instead of using chaperones, so probably that should get edited too. I'll take a look into it.

rfindler:
I've pushed your commit and some follow on edits. If something should be done differently, then I'm happy to discuss more (and so please feel free to re-open).

PR #1670, opened 2017-04-17
in-range: guard against step size of 0 (infinite loop).

stamourv:
c.f. mailing list discussion.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sequences.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sequences.scrbl
@@ -180,6 +180,12 @@ each element in the sequence.
   
   @examples[#:label "Example: sum of even numbers" #:eval sequence-evaluator
     (for/sum ([x (in-range 0 100 2)]) x)]
+
+  Note that @racket[in-range] accepts floating-point numbers, in which case
+  the usual caveats for floating-point addition apply. For example, the sequence
+  @racket[(range (- 1 1e-16) 1.0 1e-17)] reaches a point where adding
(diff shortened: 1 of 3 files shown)


pkgs/racket-doc/scribblings/reference/sequences.scrbl
```diff
@@ -180,6 +180,12 @@ each element in the sequence.
   
   @examples[#:label "Example: sum of even numbers" #:eval sequence-evaluator
     (for/sum ([x (in-range 0 100 2)]) x)]
+
+  Note that @racket[in-range] accepts floating-point numbers, in which case
+  the usual caveats for floating-point addition apply. For example, the sequence
+  @racket[(range (- 1 1e-16) 1.0 1e-17)] reaches a point where adding
+  @racket[1e-17] to the previous element of the sequence results in the same
+  floating-point number, which results in an infinite loop.
 }
```
bennn: The Scribble contract for  the`step` argument should be updated.
(Also, should that contract say `zero?` instead of `(=/c 0)`? )

mflatt:
I don't think the implementation should disallow 0: it's a backward-incompatible change, it isn't general (hence the documentation note about inexact arithmetic), and the existing behavior is both well-defined and potentially useful (in the same way that infinite streams are useful).

A documentation note to warn about 0 or small numbers seems good, though.

stamourv:
Good points.
I'll remove the changes to the implementation, then adapt the doc changes.
Thanks for the feedback!

stamourv:
Pushed a revised version.

PR #1674, opened 2017-04-25
doc: explain blame object passed to late-neg

bennn:
cc @thinkmoore 

Question for @rfindler : do you think it might be a good idea to add a `@deftech{}` for "blame object that `blame-missing-party?` returns `#true` for"?

If your answer is at least "maybe", then I'll edit this PR to define/use a `@deftech` and we can talk more.


rfindler:
Thanks!

I don't think a second blame object deftech is needed.

Robby

PR #1675, opened 2017-04-26
Add chaperone-vector*, unsafe-chaperone-vector, and property-only vec…

stamourv:
…tor chaperones.

By analogy with the procedure chaperone equivalents.

@mflatt: Any thoughts?


samth:
What contracts do you plan to use `unsafe-chaperone-vector` for?

stamourv:
Space-efficient vector contracts.

samth:
I understand how chaperone-vector* helps there, but how does that use the
unsafe version?

stamourv:
Its use is to allow "peeling back" an existing contract while preserving identity.
The idea is to use the existing contract machinery when a value gets its first contract, but if it gets a second, we switch to the (more expensive) space-efficient machinery and "replace" the original contract with a merged version of the two.

samth:
In that situation, the two arguments to unsafe-chaperone-vector are both
chaperones of the same underlying vector, right? Would it be worth
enforcing that with a dynamic check? Or maybe mentioning it in the docs.

stamourv:
They are, in that particular case. But I see that more as a property of the particular use case, rather than of `unsafe-chaperone-vector` in general. It should, IMO, stick as closely as possible to `unsafe-chaperone-procedure`, which doesn't check the analogous property.

pkgs/racket-doc/scribblings/reference/chaperones.scrbl
```diff
@@ -86,59 +86,19 @@ the same reason that they should avoid @racket[impersonator?].}
...
 impersonators to @racket[v2].
 
+Any two values that are @racket[eq?] to one another are also @racket[impersonator-of?].
 For values that include no impersonators, @racket[v1] and @racket[v2] can
 be considered impersonators of each other if they are @racket[equal?].
 
-Otherwise, impersonators within @racket[v2] must be intact within
-@racket[v1]:
+Otherwise, if @racket[v1] impersonates @racket[v1*] then @racket[(impersonator-of? v1 v2)]
+is @racket[#t] if and only if @racket[(impersonator-of? v1* v2)] is @racket[#t].
```
mflatt: What does it mean to say that `v1` impersonates `v1*`? I think the part that was cut below was trying to define "impersonates"; is "impersonates" defined somewhere else?

pkgs/racket-doc/scribblings/reference/chaperones.scrbl
```diff
@@ -397,30 +357,53 @@ after @racket[redirect-proc] (in the case of a mutator).
...
+                             [set-proc (or/c (vector? exact-nonnegative-integer? any/c . -> . any/c) #f)]
                              [prop impersonator-property?]
                              [prop-val any] ... ...)
           (and/c vector? impersonator?)]{
 
 Returns an impersonator of @racket[vec], which redirects the
 @racket[vector-ref] and @racket[vector-set!] operations.
 
-The @racket[ref-proc] must accept @racket[vec], an index passed to
+@racket[ref-proc] and @racket[set-proc] must either both be procedures
```
mflatt: Nit: don't start a sentence with a lowercase variable name. (See http://docs.racket-lang.org/style/reference-style.html#%28part._.Typesetting_.Prose%29)

pkgs/racket-doc/scribblings/reference/chaperones.scrbl
```diff
@@ -397,30 +357,53 @@ after @racket[redirect-proc] (in the case of a mutator).
...
 must produce a replacement for the value, which is used
 with @racket[vector-set!] on the original @racket[vec] to install the
 value.
 
 Pairs of @racket[prop] and @racket[prop-val] (the number of arguments
 to @racket[impersonate-vector] must be odd) add impersonator properties
-or override impersonator-property values of @racket[vec].}
+or override impersonator-property values of @racket[vec].
+
+@history[#:changed "6.9.0.1"]{Added non-interposing vector impersonators.}
```
mflatt: Turns out to be "6.9.0.2"

mflatt:
I don't see any problems in the implementation just by looking. So, except maybe for the documentation issues, this seems ok to me.

pkgs/racket-doc/scribblings/reference/chaperones.scrbl
bennn: Should this PR include the changes from #1666 ?

mflatt: Ok, more comments over there. Looking again, I see that part of the problem with the old documentation is that it's trying to centralize the definition of "impersonates" in `impersonator-of?`, but it works better to distribute that definition over things that construct impersonators. The main missing bit is the handling of compound values that contain impersonators, plus some details for immutable hash tables.

stamourv: @bennn: It looks like those changes may have gotten squashed into this PR by accident. There was a lot of rebasing going on, so that wouldn't be surprising. Let's address the comments from #1666, then see what it looks like for this one.

stamourv:
Detangled the doc changes from #1666 and merged, with the requested changes.

PR #1684, opened 2017-05-08
Improve docs for FFI custodian API.

LeifAndersen:
The docs were unclear on when an object goes from weakly to strongly
held.

### DIFF
--- a/pkgs/racket-doc/scribblings/inside/custodians.scrbl
+++ b/pkgs/racket-doc/scribblings/inside/custodians.scrbl
@@ -52,7 +52,8 @@ zero, the value is allowed to be garbage collected (and automatically
 removed from the custodian).
 
 Independent of whether @var{strong} is zero, the value @var{o} is
-initially weakly held. A value associated with a custodian can
+initially weakly held, until the garbage collector would have otherwise
+collected it. A value associated with a custodian can
 therefore be finalized via will executors.
 
 The return value from @cpp{scheme_add_managed} can be used to refer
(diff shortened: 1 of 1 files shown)


bennn:
fwiw this is unclear to me (just reading the new doc, and none of the context), maybe change to:

> Independent of whether @var{strong} is zero, the value @var{o} is
> initially weakly held, but becomes strongly held if the garbage collector
> attempts to collect it.

but I'm not sure my edit is correct

LeifAndersen:
If we replace `if` with `when` that sounds pretty close to my understanding. I'll make that change.

pkgs/racket-doc/scribblings/inside/custodians.scrbl
```diff
@@ -51,16 +51,18 @@ be remembered until either the custodian shuts it down or
 zero, the value is allowed to be garbage collected (and automatically
 removed from the custodian).
 
-Independent of whether @var{strong} is zero, the value @var{o} is
-initially weakly held. A value associated with a custodian can
-therefore be finalized via will executors.
+Independent of whether @var{strong} is zero, the value @var{o}
+is initially weakly held, but becomes strongly held when
```
bennn: change `but` to `and` ?

LeifAndersen: Hmmm...can you give me your reasoning for changing `but` to `and` here?

bennn: Because you changed `if` to `when`.

I'm thinking, "the lifetime of `o` is predictable, initially it's weakly held and once the GC finds it it's strongly held"

LeifAndersen: This makes sense. I'll make that change.

LeifAndersen:
@mflatt Should we merge this?

mflatt:
Yes, looks good.

PR #1689, opened 2017-05-12
add `prop:authentic` and `(struct .... #:authentic ....)`

mflatt:
Implementation of the `#:authentic` idea discussed on the dev list.


mfelleisen:
Just to make this clear. My remarks about "channels of communication" are not an objection but a timely observation. We have other holes like this one. We just need to be aware of them.

PR #1691, opened 2017-05-15
Adds built-in syntax classes which match regular expressions

jsmaniac:
I'm not very satisfied with the names, suggestions are welcome:

```
rx
px
rx-not-px
byte-rx
byte-px
byte-rx-not-px
```

I included `rx-not-px` and `byte-rx-not-px` Should these be discarded? They can be made rather easily with `{~and :rx {~not :px}}` and will probably be very rarely used.

Also, should I add a `rx-or-byte-rx` syntax class, which matches all regular expressions (byte or other)?

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
@@ -37,7 +37,9 @@ actually a valid expression.
 @defstxclass[integer]
 @defstxclass[exact-integer]
 @defstxclass[exact-nonnegative-integer]
-@defstxclass[exact-positive-integer])]{
+@defstxclass[exact-positive-integer]
+@defstxclass[regexp]
+@defstxclass[byte-regexp])]{
 
 Match syntax satisfying the corresponding predicates.
 }
@@ -59,6 +61,7 @@ and bytes, respectively.
 @defstxclass[id]{ Alias for @racket[identifier]. }
 @defstxclass[nat]{ Alias for @racket[exact-nonnegative-integer]. }
 @defstxclass[str]{ Alias for @racket[string]. }
+@defstxclass[character]{ Alias for @racket[char]. }
 
 @defstxclass[(static [predicate (-> any/c any/c)]
                      [description (or/c string? #f)])]{
--- a/racket/collects/syntax/parse/private/lib.rkt
+++ b/racket/collects/syntax/parse/private/lib.rkt
@@ -14,7 +14,7 @@
          exact-integer
          exact-nonnegative-integer
          exact-positive-integer
-
+         
          id
          nat
          char
@@ -37,6 +37,8 @@
 (define exact-integer-stx? (stxof exact-integer?))
 (define exact-nonnegative-integer-stx? (stxof exact-nonnegative-integer?))
 (define exact-positive-integer-stx? (stxof exact-positive-integer?))
+(define regexp-stx? (stxof regexp?))
+(define byte-regexp-stx? (stxof byte-regexp?))
 
 
 ;; == Integrable syntax classes ==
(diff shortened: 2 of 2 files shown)


wilbowma:
Based on what `racket` provides in terms of regexp predicates, and the availability of syntax-class combinators, I would suggest including only `rx`, `px`, `byte-rx` and `byte-px` (names aside).

samth:
I think the most common one to use would be something that match all forms of regular expressions (px/rx/byte/pbyte).

wilbowma:
I could see Sam's suggestion being widely used as well.

greghendershott:
It looks like you're also fixing a bug where `character` was the syntax class defined/provided, but the docs said `char` -- and you want to provide and doc both, now?

If so:

- That's great!

- I don't see where `char` is defined?

- A bit nit-picky: Because `char?` is the actual Racket predicate, I think `char` should stay in the main list, and `character` should be documented as the thing that's an alias (for `char`)?

rmculpepper:
How about I add the special power to attach syntax classes to existing bindings, and we call them `regexp`, `pregexp`, etc? (With this new power we can also rename `str` to `string` for the syntax class that recognizes literal strings, which has been a long-standing minor irritation to me.)

My inclination is to make this power available only to `syntax/parse/private/lib`, not as a user-level feature.

See also http://macrologist.blogspot.com/2013/06/define-vs-attach.html

samth:
@rmculpepper I like that idea.

jsmaniac:
@greghendershott `char` is defined as an alias of `character` a bit below my changes in the `lib.rkt` file, [line 60](https://github.com/jsmaniac/racket/blob/8bc9cef7a9d3c118438b5436af8ec020d363ea7a/racket/collects/syntax/parse/private/lib.rkt#L60) of the original. That section also contains the `nat` and `id` aliases. I agree that `char` is more intuitive, though.

@rmculpepper That sounds much better than re-inventing names — `str` always felt like a bit of a hack.

jsmaniac:
@wilbowma, @samth concerning the availability of a single px/rx/byte/pbyte predicate, I made a few quick tests and it seems that it is possible to match strings with byte-regexps (which implicitly convert the string, I suppose using utf-8?), and byte-strings with regexps. However, if either the string or the regexp is a `byte-`, then the resulting match object contains byte-strings.

The behaviour when mixing the types is therefore a bit iffy, and is likely to cause havoc further down the line. It will also be unlikely be unexpected by macro-writers, so after reconsidering it, it might be better to force macro-writers to be explicit when mixing different kinds of regexps (especially since the point of having a `:regexp` pattern will be to perform some operations with it at compile-time, so the macro-writer should likely know against which type of strings the (byte-)regexp will be used).

rmculpepper:
I've added internal support for attaching syntax classes to existing bindings, and I've done so for `string` and `bytes`.

My current preference is to add only `regexp` to `syntax/parse/private/lib` (and thus `syntax/parse`) for now. It's not clear to me that making finer distinctions between regexp types *in patterns* (as opposed to in compile-time helper functions) will be useful. If you try it and think they're useful, we could create a `syntax/parse/lib/regexp` library for them.

jsmaniac:
@rmculpepper What should that single `:regexp` pattern match, then? Both regular expressions and byte regexp? It would be semantically different then (as functions, `regexp?` vs `byte-regexp?`, and as a pattern, `:regexp` matching both).

I'm fine with not making the distinction between `pregexp?` and `regexp?`, especially since the `regexp?` predicate does not make the distinction either. Merging `regexp` and `byte-regexp` seems more iffy to me though. As I said earlier, `regexp-match` changes its return type from `string?` to `byte-string?` if matching a plain string against a `byte-regexp?`.

Here's an example where a macro might want part of its syntax to be a `regexp?`, and feeding a `byte-regexp?` instead ostensibly causes issues:

```racket
#lang racket
(require (for-syntax syntax/parse syntax/stx))
(define-syntax shorten-names
  (syntax-parser
    [(_ rx long-name:id ...)
     #:when (or (regexp? (syntax-e #'rx))
                (byte-regexp? (syntax-e #'rx))) ;; oops
     #:with (short-name ...)
     (stx-map (λ (n)
                (datum->syntax n
                               (string->symbol
                                (car
                                 (regexp-match (syntax-e #'rx)
                                               (symbol->string
                                                (syntax-e n)))))))
              #'(long-name ...))
     #'(begin
         (define-syntax short-name
           (make-rename-transformer #'long-name))
         ...)]))

(define foo 1)
(define → 2)
(shorten-names #px"^." foo →) ;; works fine

(define bar 3)
(define … 4)
(shorten-names #px#"^." foo →) ;; error:
;; string->symbol: contract violation
;;   expected: string?
;;   given: '(#"\342\206\222")
```

jsmaniac:
@rmculpepper I updated my patch based on your overloading of `string` as a syntax class. Before I push, one quick question:

Should `string`, `regexp` and `byte-regexp` be provided by `lib.rkt`?

The rationale being that they are documented as part of syntax/parse, so it should export them, especially since several other languages have their own binding for `string`, so you'd want the name clash to occur rather than a failure to recognise the `:string` as a syntax pattern). Or should I leave them out of the `provide` (the rationale being that the identifiers are already exported by `racket/base`, so there's no need to re-export them)?

rmculpepper:
Whoops, for some reason I thought `regexp?` recognized byte-regexps too, but it doesn't. You're right, let's add both `regexp` and `byte-regexp` to `syntax/parse/private/lib` and keep the meanings consistent with the corresponding predicates.

Regarding whether they should be exported: that's a good question, and I'm not sure which side of the trade-off is better for new modules. In this case, though, providing them would risk breaking compatibility, so I think we shouldn't.

PR #1695, opened 2017-05-17
Changes the "Report a Bug" url in the documentation

jsmaniac:
The sidebar for the main page on docs.racket-lang.org contains a "Report a Bug" link which points to http://bugs.racket-lang.org/, which seems to be the old page for reporting bugs (and GitHub is preferred nowadays).

A separate issue would be to update the URL used by DrRacket's "Send a bug report" window (https://github.com/racket/drracket/issues/114), but since I have basically no knowledge of the GitHub API beyond its existence, I won't do it myself.

### DIFF
--- a/pkgs/racket-index/scribblings/main/config.rkt
+++ b/pkgs/racket-index/scribblings/main/config.rkt
@@ -4,7 +4,8 @@
 
 ;; Configuration of various parts of the main pages
 
-(define bug-url "http://bugs.racket-lang.org/")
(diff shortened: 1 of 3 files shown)


samth:
Unfortunately it's not really possible to make "Report a bug" work in general because of GitHub's authentication model, but I have some plans/code to work on that that I need to finish up.

samth:
Merged in ee73a550aabd.

PR #1698, opened 2017-05-22
changed make-variable-like-transformer to create #%app form

iitalics:
The docs claim that "Uses of the macro in operator position are
interpreted as an application", yet the current implementation
confusingly wraps #%expression around the operator instead of just
creating #%app.

Worse, the #%expression isn't in the phase level of the application,
meaning that languages can't easily use make-variable-like-transformer
if they intend to reimplement #%app. Maybe I'm mistaken and there is
rationale behind doing it this way / code that depends on the current
behavior. I find my self needing this macro but just rewriting it this
way since the bad phase-level #%expression just gets in the way and I
wanted #%app instead.

### DIFF
--- a/racket/collects/syntax/transformer.rkt
+++ b/racket/collects/syntax/transformer.rkt
@@ -24,5 +24,5 @@
               [else
                (raise-syntax-error #f "cannot mutate identifier" stx #'id)])]
        [(id . args)
-        (let ([stx* (cons #'(#%expression id) (cdr (syntax-e stx)))])
-          (datum->syntax stx stx* stx))]))))
+        (with-syntax ([app (datum->syntax stx '#%app)])
+          (syntax/loc stx (app id . args)))]))))


rmculpepper:
`make-variable-like-transformer` is intended to cooperate with redefinitions of `#%app`, and in the simple examples I've tried, it does. I don't understand what you mean by the `#%expression` being at the wrong phase level. Do you have a small example that doesn't work the way you expect with the current definition of `make-variable-like-transformer`?

iitalics:
@rmculpepper  https://gist.github.com/iitalics/458fc2d765d3da57d2d96a6c8c930fde
This code is attempting to read syntax properties off of the function in an application. Then, we attach properties to a function and make it look like an identifier using make-variable-like-transformer. The issue arises when #%expression is introduced and obscures the desirable syntax properties. "Being at the wrong phase level" is evident by the fact that I can't get rid of the #%expression (and therefore expose the syntax properties) by simply writing my own implementation of it. The only solution I see is pattern matching on the expression to manually remove #%expression.

Is there any reason why #%expression is preferred over #%app? As I see it, #%expression just forces the #%app to be made anyways.

rmculpepper:
@iitalics Okay, it sounds like you would like to change the expansion to make the code slightly easier to analyze. I don't have a strong preference, if no one else is currently depending on the current expansion.

I like the `#%expression` expansion because it makes explicit that even in operator position the variable-like macro is treated just as an independent expression. It also prevents expansion from going through the used-as-operator path more than once, such as in the cases of `#%app` transformers that expand into implicit applications rather than explicit applications, such as

    (define-syntax-rule (my-app f a ...) (begin (do-something) (f a ...)))

On the other hand, there's no real harm if that happens.

The fact that your redefinition of `#%expression` doesn't affect the expansion of the variable-like macro has nothing to do with a "wrong phase level"; it's simple hygiene. The reference to `#%expression` refers to the binding in the context of the syntax template. In contrast, the transformer goes out of its way to get the binding of `#%app` from the use site, using `datum->syntax` to transfer it to the pair that represents the implicit application.

Nitpick: your gist example shouldn't be seeing the used-as-operator expansion anyway, since your `my-app` local-expands the operator as an independent expression.

iitalics:
>it's simple hygiene. The reference to #%expression refers to the binding in the context of the syntax template.

Well yes, but which syntax template? My impression is that its the template in `(cons #'(#%expression id) (cdr (syntax-e stx)))`, which means that #%expression is always the one imported from racket/base (due to `(require (for-template racket/base))`). I don't think this is very desirable for writing #lang's that aim to override many of racket/base's other imports.

>Nitpick: your gist example shouldn't be seeing the used-as-operator expansion anyway

I'm sorry, I'm not exactly sure what you meant by this. 

I think that only introducing #%app is easier to work with, and follows the documentation more closely without the level of indirection of #%expression. At the very least, #%expression should have a phase level that the user has some control over.

PR #1700, opened 2017-05-23
add neg-party arg to id-set-contract-check

stchang:
Fixes bug from e92b8610f2799602425410fa494a3cd6b022e3cf

### DIFF
--- a/racket/collects/syntax/id-set.rkt
+++ b/racket/collects/syntax/id-set.rkt
@@ -81,23 +81,23 @@
   (lambda (s)
     (and (id-set? s) (idsettype? s) (mutability? s))))
 
-(define (id-set-contract-check idsettype mutability b s)
+(define (id-set-contract-check idsettype mutability b s neg-party)
   (unless (id-set? s)
-    (raise-blame-error b s "expected either a free or bound identifier set"))
+    (raise-blame-error b #:missing-party neg-party s "expected either a free or bound identifier set"))
   (case idsettype
     [(free)
      (unless (free-id-set? s)
-       (raise-blame-error b s "expected a free-identifier set"))]
+       (raise-blame-error b #:missing-party neg-party s "expected a free-identifier set"))]
     [(bound)
      (unless (bound-id-set? s)
-       (raise-blame-error b s "expected a bound-identifier set"))])
+       (raise-blame-error b #:missing-party neg-party s "expected a bound-identifier set"))])
   (case mutability
     [(mutable)
      (unless (mutable-id-set? s)
-       (raise-blame-error b s "expected a mutable id set"))]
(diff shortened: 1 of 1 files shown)


rfindler:
I can't seem to find any existing test cases for `id-set/c`. Do you see any?

rfindler:
Looks like we need some too. Unless I'm running things wrong, this produces blame and the docs say it shouldn't:

```
#lang racket/base
(require racket/contract
         syntax/id-set)
(contract (id-set/c integer?)
          (mutable-free-id-set)
          'pos 'neg)
```

rfindler:
Here's one way to add tests. This doesn't seem like quite the right place, but it isn't a terribly horrible place, either?

```
☕  git diff 
diff --git a/pkgs/racket-test/tests/racket/contract/syntax.rkt b/pkgs/racket-test/tests/racket/contract/syntax.rkt
index d2a625e..0d628a7 100644
--- a/pkgs/racket-test/tests/racket/contract/syntax.rkt
+++ b/pkgs/racket-test/tests/racket/contract/syntax.rkt
@@ -2,7 +2,8 @@
 
 (require "test-util.rkt")
 (parameterize ([current-contract-namespace
-                (make-basic-contract-namespace)])
+                (make-basic-contract-namespace
+                 'syntax/id-set)])
   
   (test/pos-blame
    'syntax/c1
@@ -20,4 +21,24 @@
   
   (test/no-error '(syntax/c (list/c #f)))
   (contract-error-test 'syntax/c-non-flat '(syntax/c (vector/c #f))
-                       (λ (x) (regexp-match? #rx"flat-contract[?]" (exn-message x)))))
+                       (λ (x) (regexp-match? #rx"flat-contract[?]" (exn-message x))))
+
+
+  (test/pos-blame
+   'id-set/c.1
+   '(contract (id-set/c integer?)
+              5
+              'pos 'neg))
+
+  (test/pos-blame
+   'id-set/c.2
+   '(contract (id-set/c integer?)
+              (mutable-free-id-set)
+              'pos 'neg))
+
+  (test/pos-blame
+   'id-set/c.3
+   '(contract (id-set/c integer?)
+              (mutable-free-id-set)
+              'pos 'neg))
+  )
```

stchang:
There seems to already be tests here https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/id-set-test.rktl#L639

So I went ahead and merged

rfindler:
I wasn't getting email from drdr about failing tests. And the test case I wrote above suggests that either the implementation or the docs are wrong.

Can we fix those two things?

rfindler:
(I was wrong about the docs/implementation mismatch. I misread the docs.)

Issue #1702, opened 2017-05-24
`db` collection in Racket raises incorrect error message when executing certain kinds of Sqlite3 queries.


abmclin:
The following programs:

```
#lang racket/base
(require db)

(define conn (sqlite3-connect #:database 'memory))
(query-exec conn "Select ,* from nonexistenttable") ; Has an extra comma that doesn't belong there.

```

```
#lang racket/base
(require db)

(define conn (sqlite3-connect #:database 'memory))
(query-exec conn "Create Table 'test' ('notes' Text)")
(query-exec conn "Create Table 'test' ('notes' Text)") ; Executed twice after same table already created.

```

```
#lang racket/base
(require db)

(define conn (sqlite3-connect #:database 'memory))
(query-exec conn "Create Table 'test' ('notes' Text)")
(query-exec conn "Select * from test; ") ; An extra space after the terminating semi-colon.

```

all raise ***query-exec: multiple statements given*** exception.

What should have occurred instead is ***query-exec: near ",": a syntax error*** and ***query-exec: table 'test' already exists*** exceptions in the first two programs and no exception raised in the third one.

abmclin:
This issue has been resolved by the PR merge.

PR #1704, opened 2017-05-24
PR fix for Issue #1702

abmclin:
Please see Issue #1702 for an explanation of the reasons behind this pull request.

The tests are being committed to the `db` repository, I will submit a separate pull request there.

Some notes for your consideration.

In the commit dealing with connection.rkt, my changes are within the atomic execution region in the prepare1* method. As far as I can tell, the changes, including the call to `get-error-message` are safe with respect to atomic execution but I'm not 100% sure.

For the specific commit relating to the ffi.rkt file, which aims to fix the issue of semi-colon terminated statement with extra space throwing an exception. I think it's user friendly for the `db` collection to proactively trim all strings from callers so they don't have to do it themselves before calling query-exec and its friends.

I use `string-trim` from racket/string in the copy-buffer procedure because it's a convenient central location which captures all SQL strings given to the connection object. Are there any performance concerns with bringing another dependency into ffi.rkt? I checked and it seems racket/string is a small library.

### DIFF
--- a/racket/collects/db/private/sqlite3/connection.rkt
+++ b/racket/collects/db/private/sqlite3/connection.rkt
@@ -212,10 +212,16 @@
                       ;; entry of stmt in table.
                       (A (let-values ([(prep-status stmt tail?)
                                        (sqlite3_prepare_v2 db sql)])
-                           (when tail?
-                             (when stmt (sqlite3_finalize stmt))
-                             (error* fsym "multiple statements given"
-                                     '("given" value) sql))
+                           (cond
+                             [(not (= 0 prep-status))
+                              (when stmt (sqlite3_finalize stmt))
+                              (error* fsym (get-error-message)
+                                      '("given" value) sql)]
+                             [else
+                              (when tail?
+                                (when stmt (sqlite3_finalize stmt))
+                                (error* fsym "multiple statements given"
+                                        '("given" value) sql))])
                            (when stmt (hash-set! stmt-table stmt #t))
                            (values prep-status stmt))))])
         (when DEBUG?
--- a/racket/collects/db/private/sqlite3/connection.rkt
+++ b/racket/collects/db/private/sqlite3/connection.rkt
@@ -208,22 +208,20 @@
       (let*-values ([(db) (get-db fsym)]
                     [(prep-status stmt)
                      (HANDLE fsym
-                      ;; Do not allow break/kill between prepare and
-                      ;; entry of stmt in table.
-                      (A (let-values ([(prep-status stmt tail?)
-                                       (sqlite3_prepare_v2 db sql)])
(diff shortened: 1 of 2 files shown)


abmclin:
Hello @rmculpepper I was wondering if you would be able to review this pull request and the other related PR at https://github.com/racket/db/pull/9

Thank you.

bennn:
FWIW looks good to me. I think the use of `string-append` and the new code in the atomic block are fine.

rmculpepper:
@abmclin, thanks for the pull request, and sorry for taking so long to get to it. I have two comments:

I think the sql strings should be trimmed only on the right; otherwise, if sqlite reported a syntax error with a location, the location would be off. (I don't know if it currently does, but I think the smallest change that achieves the goal would be best.)

Either string-trim should be called in the ffi functions before copy-buffer (rather than being called within copy-buffer), or copy-buffer should be renamed to something like trim-and-copy-buffer.

Otherwise, it looks good to me, including the atomic section.

abmclin:
@rmculpepper Good point that trimming on the left would shift the string leftward and throw off locations. Right-only trimming seems to be the safest conservative approach. I'll make the changes.

I am leaning towards renaming `copy-buffer` to `trim-and-copy-buffer` approach.

abmclin:
@rmculpepper @bennn Suggested changes have been made, I also did a grep search of the Racket repository to be sure the `copy-buffer` procedure defined in sqlite3/ffi.rkt wasn't being used elsewhere which would be broken by the rename to `trim-and-copy-buffer`. So far it seems the rename is safe.

rmculpepper:
I tried building with the PR commits locally merged, and I got lots of "database is locked" (SQLITE_BUSY) errors from raco setup. The fix is to let the outer `HANDLE` take care of any non-zero `prep-status` result, so that error reporting is handled uniformly but more importantly so that "busy" failures get properly identified and automatically retried. Sorry for missing that before. The code should look like this:

    (HANDLE fsym
     ;; Do not allow break/kill between prepare and
     ;; entry of stmt in table.
     (A (let-values ([(prep-status stmt tail?)
                      (sqlite3_prepare_v2 db sql)])
          (cond [(not (zero? prep-status))
                 (when stmt (sqlite3_finalize stmt))
                 (values prep-status #f)]
                [tail?
                 (when stmt (sqlite3_finalize stmt))
                 (error* fsym "multiple statements given"
                         '("given" value) sql)]
                [else
                 (when stmt (hash-set! stmt-table stmt #t))
                 (values prep-status stmt)]))))

abmclin:
@rmculpepper thank you for catching that.  I wish it had occurred to me to perform a fresh build with the changes.

I'll be sure to do that next time I make any changes in the Racket repository.

PR #1713, opened 2017-06-04
Add scheme_atexit c function.

LeifAndersen:
This way programs can actually call atexit. (Otherwise atexit
is frequently not provided in libc as a symbol.)

### DIFF
--- a/racket/src/racket/include/mzscheme.exp
+++ b/racket/src/racket/include/mzscheme.exp
@@ -63,6 +63,7 @@ scheme_close_managed
 scheme_schedule_custodian_close
 scheme_add_custodian_extractor
 scheme_add_atexit_closer
+scheme_add_atexit
 scheme_add_evt
 scheme_add_evt_through_sema
 scheme_is_evt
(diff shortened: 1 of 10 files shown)


mflatt:
Calling `atexit()` is not always the right thing. Probably the code in `scheme_add_atexit_closer()` in "thread.c", which sometimes calls atexit() but sometimes does other things, should be moved into `scheme_atexit()` and called from `scheme_add_atexit_closer()`.

LeifAndersen:
Good idea. I have moved the relevant code to scheme_atexit, thanks.

LeifAndersen:
Would it be a good idea to also bump the version number, or do you think this is too minor for that?

mflatt:
If you need to depend on it from a package, then bump the version number. Otherwise, I'd just leave it.

LeifAndersen:
Okay thanks. Finally, do you think its worth documenting on: http://docs.racket-lang.org/inside/Custodians.html?q=atexit#%28idx._%28gentag._582._%28lib._scribblings%2Finside%2Finside..scrbl%29%29%29

Otherwise I think its good to merge. (It doesn't seem like there are any tests for thread.c)

mflatt:
I think it's worth documenting, and I agree that it's otherwise ready to go.

LeifAndersen:
Thanks for your feedback. Merged.

LeifAndersen:
(I've also pushed the docs.)

PR #1714, opened 2017-06-05
Add mzrt_sema_try_wait to the mzrt library.

LeifAndersen:
This commit add mzrt_sema_try_wait to the functions that operate
on semaphores. The existing ones are:

* int mzrt_sema_create(mzrt_sema **sema, int init);
* int mzrt_sema_post(mzrt_sema *sema);
* int mzrt_sema_wait(mzrt_sema *sema);
* int mzrt_sema_destroy(mzrt_sema *sema);

### DIFF
--- a/racket/src/racket/src/mzrt.c
+++ b/racket/src/racket/src/mzrt.c
@@ -598,6 +598,18 @@ int mzrt_sema_wait(mzrt_sema *s)
   return 0;
 }
 
+int mzrt_sema_try_wait(mzrt_sema *s)
+{
+  int locked = 1;
+  pthread_mutex_lock(&s->m);
+  if(s->ready) {
+    --s->ready;
+    locked = 0;
+  }
+  pthread_mutex_unlock(&s->m);
+  return locked;
+}
+
 int mzrt_sema_post(mzrt_sema *s)
 {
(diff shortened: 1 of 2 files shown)


mflatt:
Unless I'm misreading, I recommend inverting the return value: 1 (= true) for a successful wait, 0 (= false) f the semaphore isn't ready. That would be most consistent with `semaphore-try-wait?`.

LeifAndersen:
This is fine with me.

I made it that way to match the pthread's sema_wait and Window's WaitForSingleObject semaphore call. (Both of which return 0 when the semaphore was decremented, and something non-zero otherwise.)

LeifAndersen:
Anyway, Since none of the thread functions in mzrt are documented should I document them, or would we like to leave them undocumented?

mflatt:
Good point about 0 normally meaning success in C, so choose whichever variant you think is most consistent.

I should really make the `mzrt_...` functions properly public (which involves adding them to "schemef.h" et al.) and document them, but I don't think that needs to be your responsibility for this commit.

LeifAndersen:
All right then. I will push the version that matches the mzrt_mutex_trylock function. (Which also matches the pthreads and windows api).

Thanks.

PR #1716, opened 2017-06-07
style: document '/', as in 'call/cc'

bennn:
Add row to the table in the section on "Names".

Also clarify that the nearby margin note is about `#%`


bennn:
Preview:
screen shot 2017-06-07 at 09 35 45

stamourv:
LGTM

PR #1727, opened 2017-06-20
Add #:lib-suffix parameter to ffi-lib call.

LeifAndersen:
This is useful if you have a library that doesn't use the usual extension (namely so on os x).

Still need to add docs.

### DIFF
--- a/racket/collects/ffi/unsafe.rkt
+++ b/racket/collects/ffi/unsafe.rkt
@@ -107,18 +107,26 @@
 ;; ----------------------------------------------------------------------------
 ;; Getting and setting library objects
 
-(define lib-suffix (bytes->string/latin-1 (subbytes (system-type 'so-suffix) 1)))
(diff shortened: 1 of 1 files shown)


mflatt:
`ffi-lib` tries the path that you provide without adding any suffix. For your case, can you just include the suffix in the `ffi-lib` path argument?

LeifAndersen:
Oh, interesting. That's not what the docs say:

> path is a path without a version or suffix (i.e., without ".dll", ".so", or ".dylib"); and

Should I update them to say that the path _can_ potentially include the suffix?

mflatt:
That bit of documentation is after "Normally", and there are several constraints in that itemization that aren't always true, as detailed in the rest. The rest is long and complex, which is why there's a "Normally" to explain the typical use. Maybe you can see a rewrite to make the "Normally" aspect more clear.

LeifAndersen:
Oh hey...I completely missed that. Thanks. :)

PR #1736, opened 2017-06-28
bug: add guard against string-paths in 'find-relative-path'

bennn:
Add guard, if first argument to `find-relative-path` is not a path,
 then don't call `path-convention-type` on it.

Thanks to Jack Firth for reporting:
https://github.com/racket/rackunit/pull/46


mflatt:
LGTM - thanks!

PR #1753, opened 2017-07-21
Typos for counting numbers in German and French

catull:
In English, "one" can be used both when counting (1, 2, 3, 4, 5, ...) as well as an article in front of a noun (one tree, one bridge, ...).

In German the two use cases are distinct.
When counting it is "eins", but as an article it would be "ein" (for male or neutral nouns) or "eine" (for female nouns).
Thus in the text the correct term would be "eins" in German.

For the French counting, "une" is not quite correct.
It is much more common to count "un", "deux", "troix", "quatre", "cinq", ....


mflatt:
Thanks!

Issue #1754, opened 2017-07-21
Using 'integer->integer-bytes' with 8-bit integers?


HiPhish:
Hello everyone,

I am curious as to why the `integer->integer-bytes` function only works for integers of two, four or eight bytes length, but not those of one byte length. For positive integers
```racket
(integer->integer-bytes n 1 #t)
```
would be the same as
```racket
(bytes n)
```
but in order to account for negative integers I need to do bit-fiddling like this:
```racket
(bytes (if (< n 0) (bitwise-ior #b10000000 (abs n)) n))
```
On top of that I have to add all sorts of extra boilerplate to figure out whether the integer is eight or more bits long. It would be much simpler if `integer->integer-bytes` offered a unified interface for all integers up to a certain size (i.e. eight in this case).

Is there a technical reason for not supporting 8-bit integers?

mflatt:
There was no good reason. Thanks to @rain-1 for the patch to add the missing functionality.

HiPhish:
This didn't make it into 6.10, did it?

mflatt:
Correct. The change went in well after the branch date for v6.10 (which was July 7).

HiPhish:
OK, thanks. I guess I can expect 6.11 in about another three months then.

PR #1755, opened 2017-07-21
Add supporting for 1 byte numbers to "integer->integer-bytes" and "in…

rain-1:
Add supporting for 1 byte numbers to "integer->integer-bytes" and "integer-bytes->integer" in racket/src/racket/src/numstr.c

https://github.com/racket/racket/issues/1754

### DIFF
--- a/racket/src/racket/src/numstr.c
+++ b/racket/src/racket/src/numstr.c
@@ -2145,6 +2145,19 @@ int scheme_check_long_double(const char *where, long_double d, const char *dest)
 Scheme_Object *scheme_bytes_to_integer(char *str, int slen, int sgned, int rshft, int mask)
 {
   switch(slen) {
+  case 1:
+    if (sgned) {
+      char val;
+      memcpy(&val, str, sizeof(char));
+      return scheme_make_integer(val);
+    } else {
+      unsigned char val;
+      memcpy(&val, str, sizeof(unsigned char));
+      val >>= rshft;
+      if (mask < 8) { val &= (((unsigned char)1 << mask) - 1); }
+      return scheme_make_integer(val);
+    }
+    break;
   case 2:
     if (sgned) {
       short val;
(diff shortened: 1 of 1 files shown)


rain-1:
tested it out with a few values:

```
(integer->integer-bytes #x235E 2 #f)
(integer->integer-bytes #x22 1 #f)
(integer->integer-bytes #xFE 1 #f)
(integer->integer-bytes #x22 1 #t)
(integer->integer-bytes #x7E 1 #t)
(integer->integer-bytes (- #x7E) 1 #t)

(integer-bytes->integer (integer->integer-bytes #x235E 2 #f) #f)
(integer-bytes->integer (integer->integer-bytes 34 1 #f) #f)
(integer-bytes->integer (integer->integer-bytes #xFE 1 #f) #f)
(integer-bytes->integer (integer->integer-bytes #x22 1 #t) #t)
(integer-bytes->integer (integer->integer-bytes #x7E 1 #t) #t)
(integer-bytes->integer (integer->integer-bytes (- #x7E) 1 #t) #t)
```

rain-1:
I wanted to ask if the code is correct, when `sgned` is true it coerces to an unsigned short. 

```c
  case 2:
    {
      if (sgned) {
        unsigned short value = val;
        memcpy(str, &value, sizeof(unsigned short));
      } else {
        short value = val;
        memcpy(str, &value, sizeof(short));
      }
    }
    break;
```

scheme_bytes_to_integer does the opposite:

```c
  case 2:
    if (sgned) {
      short val;
      memcpy(&val, str, sizeof(short));
      return scheme_make_integer(val);
    } else {
      unsigned short val;
      memcpy(&val, str, sizeof(unsigned short));
      val >>= rshft;
      if (mask < 16) { val &= (((unsigned short)1 << mask) - 1); }
      return scheme_make_integer(val);
    }
    break;
```

I may be misunderstanding the code but I thought I'd ask!

mflatt:
@rain-1 I think you're right that the test on `sgned` is backward for the action. The compiler likely generates the same code for both branches, but it doesn't have to.

I'll fold that repair and your tests cases into your commit to merge, and I'll add documentation.

Thanks!

mflatt:
Merged

PR #1757, opened 2017-07-25
should be "requiring" module invokes the function

arbipher:
Before this line, it says the function flows from providing module to requiring module.
After this line, it says the argument travels back from requiring moduleo to providing module.
I believe that argument supplying and function invoking should happen at requiring module, rather than providing module.


mflatt:
Thanks!

PR #1758, opened 2017-07-27
gitignore: vim swapfiles

bennn:
Ignore the first 3 levels of Vim-generated swapfiles.


mflatt:
This looks fine to me.

PR #1767, opened 2017-08-08
Fix the contract for with-intercepted-logging

LeifAndersen:
The contract was missing the required log-level/c. This commit
adds it.


stamourv:
LGTM

PR #1770, opened 2017-08-11
Point make-meta-reader to the document of read and read-syntax

shhyou:
- Point read, read-syntax and make-meta-reader to
  the documentation explaining the arguments of
  read and read-syntax (the ones discussing x and x+4
  arguments).

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/module-reader.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/module-reader.scrbl
@@ -417,6 +417,8 @@ succeeds, then the loaded module's @racketidfont{read},
 @racketidfont{read-syntax}, or @racketidfont{get-info} export is
 passed to @racket[convert-read], @racket[convert-read-syntax], or
 @racket[convert-get-info], respectively.
+See @secref["parse-reader" #:doc ref-doc] for information on
+the protocol of @racketidfont{read} and @racketidfont{read-syntax}.
 
 @margin-note{The @racketmodname[at-exp] language supplies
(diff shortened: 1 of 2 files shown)


shhyou:
A different change could be making `reader` a `deftech` at the beginning of Section 1.3 and move the third paragraph of **1.3.18 Reading via an Extension** there.

pkgs/racket-doc/syntax/scribblings/module-reader.scrbl
```diff
@@ -416,7 +416,9 @@ If loading the module produced by @racket[module-path-parser]
 succeeds, then the loaded module's @racketidfont{read},
 @racketidfont{read-syntax}, or @racketidfont{get-info} export is
 passed to @racket[convert-read], @racket[convert-read-syntax], or
-@racket[convert-get-info], respectively.
+@racket[convert-get-info], respectively. See
+@secref["parse-reader" #:doc '(lib "scribblings/reference/reference.scrbl")]
```
mflatt: You can use `refman` here (as defined in "common.rkt") instead of `'(lib "scribblings/reference/reference.scrbl")`

mflatt:
This looks like a good change. If my `refman` comment is correct (I didn't try it to make sure), that would be a little better, but I'm happy to merge either way.

I can easily believe that rearranging the docs could be even better, since I have trouble following the cross-references myself.

pkgs/racket-doc/syntax/scribblings/module-reader.scrbl
shhyou: Changed. Thanks!

Issue #1784, opened 2017-08-28
Recursive contract


zaoqi:
```racket
(define (sizedstream/c t) (or/c null? (cons/c t (sizedstream/c t)) (promise/c (sizedstream/c t)))
```
How to define it?

bennn:
Just put `recursive-contract` around the part that needs to be delayed:

```
#lang racket/base

(require racket/contract)

(define (sizedstream/c t)
  (or/c null?
       (cons/c t (recursive-contract (sizedstream/c t)))
       (promise/c (recursive-contract (sizedstream/c t)))))

(define/contract s1
  (sizedstream/c integer?)
  (cons 1 (cons 2 (cons 3 '()))))
```

PR #1785, opened 2017-08-28
doc: example recursive-contract

bennn:
Thanks to @zaoqi (#1784) for the idea.

Is there a simpler / clearer example we could use instead of `int-list/c` ?

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -3132,6 +3132,17 @@ value to be checked against the contract is supplied to the contract.
 Without it, the @racket[contract-expr] is evaluated earlier. This option
 is supported only when @racket[type] is @racket[#:flat].
 
+@examples[#:eval (contract-eval)
+  (define even-length-list/c
+    (or/c null?
+          (cons/c any/c
+                  (cons/c any/c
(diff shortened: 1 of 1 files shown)


zaoqi:
@bennn 
```racket
(define (promise+/c t)
  (or/c
   t
   (promise/c (recursive-contract (promise+/c t)))))
```

rfindler:
I think that a better example is probably a function one that accepts two contracts and returns a contract on lists that alternate those elements, one after the other. And then use it in a `#:rest` argument on an `->` contract.

This would require some English to motivate and explain, tho. 

You could start with an even-length list of integers and work your way up?

bennn:
I like `even-length-list/c`, that seems simple & useful to me.

I'm thinking we should leave it at this:
- I don't want to add more English, because I don't want to distract from the main point of "how do I use `recursive-contract` at all?"
- I'd rather the contract isn't a function, because then it's really obvious when you've got it wrong. If you remove `recursive-contract` from this example then `define` raises an exception.

bennn:
screen shot 2017-08-28 at 11 51 24

rfindler:
Okay, that all makes sense. And I do think that something in the guide section on contracts is where to discuss adding a contract like `hash` has.

rfindler:
When you are happy with it, please go ahead and push the commit. Thanks!

PR #1794, opened 2017-09-02
require racket/contract/base in generic.scrbl to fix label links

iitalics:
The contracts in this page weren't linked, because the contract library was not `require`d.


stamourv:
Thanks!

PR #1803, opened 2017-09-15
merge syntax/parse's template into core (syntax, syntax/loc, etc)

rmculpepper:
Preliminary version for discussion. Don't merge yet.

This PR replaces the existing implementation of `syntax` and derived forms (`syntax/loc`, `quasisyntax`, the right-hand sides of `syntax-rules` clauses, etc) with one supporting the features of syntax/parse's `template` form.

It adds the following exports to racket/base: `??` and `?@`. Support for template metafunctions is included in `syntax` et al, but I haven't added `define-template-metafunction` to racket/base. I'm ambivalent about metafunctions; they haven't been as useful to me as I expected, but I'd like to hear if others find them useful.

Still to do: update documentation, move tests


rfindler:
Does this mean attributes are going to be available that lower-layer too? (And I could easily believe this is a confused question; apologies in advance if so.)

rmculpepper:
Attributes have always acted as syntax pattern variables. They should continue to behave the same. (Including support for promises, behavior on absent matches, etc.)

One new feature is that if `x` is bound as an attribute, an occurrence of `x.y` is an error until `x.y` is also bound as a pattern variable. (Because it was probably intended as an attribute reference rather than a literal.)

Did that answer your question?

rfindler:
Yes, I think so. Thanks!

LeifAndersen:
Has there been any progress on this PR?

(I ask because I'm quite excited to have this out of an `experimental` collect.)

rmculpepper:
I think this is ready. @mflatt, could you review?

mflatt:
@rmculpepper I did not read the code in depth, but it look good in almost every dimension I looked at, including expansion overhead, load time, and bytecode size.

The "template.rkt" module breaks the expander build and (for the same reason) the RacketCS build. I think I just need to teach the linklet flattener that it's ok to discard an unreferenced call to `gensym`. So, let me do that and double-check RacketCS before you merge.

Should there be any tests added in "racket-test-core"? Moving the full set of template tests into the ".rktl" style seems unnecessary and painful, but maybe it's worth a couple of tests in "stx.rktl" to check that `??` is at least present and similar.

mflatt:
I pushed the flattener improvement, so the expander now builds and RacketCS works with these changes.

rmculpepper:
@mflatt, thanks! Just to check my understanding, the gensym in question is `absent-pvar-escape-key`, which is unreferenced because there are no uses of the procedures that refer to it given the syntax templates that occur in the expander code and its dependencies, so they get dropped. Is that right?

Adding basic tests to "stx.rktl" sounds like a good idea. I'll do that.

mflatt:
@rmculpepper Yes, that's right.

PR #1807, opened 2017-09-18
Improve `namespace-attach-module` example

michaelballantyne:
The current example fails to illustrate the complete use, as the final evaluation of `apple` would succeed even without the `namespace-attach-module` call. This updated example demonstrates that `namespace-require` is still needed after the module is attached, and also shows how the behavior differs from `namespace-attach-module-declaration`.


LeifAndersen:
Merged, thanks.

PR #1809, opened 2017-09-21
doc: fix hash-union contracts

bennn:
Changed to match `hash-set` and `hash-set!` contracts

- `hash-can-functional-set?` => `immutable?
- `hash-mutable?` => `(not/c immutable?)`


stamourv:
LGTM

PR #1814, opened 2017-09-22
Make compute-offsets public.

LeifAndersen:


samth:
Should this have some tests?

LeifAndersen:
Probably, and added to `pkgs/racket-test-core/tests/racket/cstruct.rktl`

pkgs/racket-test-core/tests/racket/cstruct.rktl
```diff
@@ -156,5 +156,15 @@
   (test #f equal? b (cast b _B-pointer _B-pointer))) ; cast forces new wrapper
 
 ;; ----------------------------------------
+;; Check to ensure offsets are computed correctly
+
+(let ()
+  (test '(0 4 8) compute-offsets (list _int _bool _string))
+  (test '(0 4 5 8) compute-offsets (list _int _byte _byte _long))
```
mflatt: `_long` is not always 8 bytes (not even on 64-bit platforms, thanks to Windows).

LeifAndersen:
Okay @samth tests added.

pkgs/racket-doc/scribblings/foreign/types.scrbl
```diff
@@ -1402,6 +1402,34 @@ expects arguments for both the super fields and the new ones:
...
 
+@defproc[(compute-offsets [types (listof ctype?)]
+                          [alignment (or/c #f 1 2 4 8 16) #f]
+                          [declare (listof (or/c #f exact-integer?)) '()])
+         (listof exact-integer?)]{
+                                  
+ Given a list of types in a C struct type, return the offset
+ of those types.
+
+ @racket[types] is a list of types of the C struct. The same
```
mflatt: Please don't start a sentence with a lowercase identifier (http://docs.racket-lang.org/style/reference-style.html section 5.3). The second sentence isn't complete.

pkgs/racket-doc/scribblings/foreign/types.scrbl
```diff
@@ -1402,6 +1402,34 @@ expects arguments for both the super fields and the new ones:
...
+ as @racket[types] in @racket[make-cstruct-type].
+
+ @racket[alignment] sets the C struct padding. If
```
mflatt: Nit: I don't think "padding" is the right word here.

pkgs/racket-doc/scribblings/foreign/types.scrbl
```diff
@@ -1402,6 +1402,34 @@ expects arguments for both the super fields and the new ones:
...
+ @racket[alignment] is @racket[#f], then the natural
+ alignment is used. Otherwise the given value is used. The
+ behavior is identical to @racket[make-cstruct-type].
```
mflatt: Maybe it's better to leave the description to the `make-cstruct-type` docs (and improve those docs as needed)?

racket/collects/ffi/unsafe.rkt
```diff
@@ -1410,10 +1410,11 @@
 ;; ----------------------------------------------------------------------------
 ;; Struct wrappers
 
-(define (compute-offsets types alignment declared)
+(define* (compute-offsets types [alignment #f] [declared '()])
```
mflatt: A public function should check arguments. I know that many functions here don't, because the original author didn't include them, but we've been trying to fill them in.

mflatt: I'm not seeing a `provide` of `compute-offsets`. Am I overlooking something?

pkgs/racket-test-core/tests/racket/cstruct.rktl
LeifAndersen: Good catch, thanks.

racket/collects/ffi/unsafe.rkt
LeifAndersen: Fair. I just didn't because I noticed none of the other ones did so I suspected there was a reason for that. I'll fix that. Thanks.

pkgs/racket-doc/scribblings/foreign/types.scrbl
LeifAndersen: Fixed, thanks.

pkgs/racket-doc/scribblings/foreign/types.scrbl
LeifAndersen: Err...ya,thanks.

pkgs/racket-doc/scribblings/foreign/types.scrbl
LeifAndersen: I honestly could go either way.

racket/collects/ffi/unsafe.rkt
LeifAndersen: The module (for some reason) defined the `define*` macro to be a define and provide. The other cstruct parts of that module used it for define, which is why I simply added the `*`.

pkgs/racket-test-core/tests/racket/cstruct.rktl
mflatt: Looking again, I guess it doesn't matter since the `_long` is at the end, but I like the change to `_int`.

racket/collects/ffi/unsafe.rkt
mflatt: Ah, right.

mflatt:
LGTM

LeifAndersen:
(The failing tests are because unicode.org was down during the time travis was building. Once it goes back up the tests run as expected.)

PR #1818, opened 2017-09-25
Add make-serialize-info example.

LeifAndersen:


pkgs/racket-doc/scribblings/reference/serialization.scrbl
```diff
@@ -583,6 +583,38 @@ This directory path is used as a last resort when
...
+      (define pie-pattern (pie 'transporter-error))
+      (values pie-pattern
+              (λ (type)
+                (set-pie-type! pie-pattern type))))))
+ (define original-pie
+   (pie 'apple))
+ original-pie
+ (define pie-in-transit
+   (serialize original-pie))
+ pie-in-transit
```
bennn: Does this (and the `(pie-type ....)` call below) look good when rendered?

(For me, "good" is "takes up less than 1 line".)

LeifAndersen: I think so. At least this is how it looks like when I render it.
screen shot 2017-09-26 at 4 13 02 pm

LeifAndersen:
Everything looks checked out to me, okay to merge?

PR #1832, opened 2017-10-04
informal RFC: return expression for for/fold?

pnwamk:
TL;DR it would be great if a `for/fold` user could provide an expression that determined what value(s) is(are) returned from a for-loop upon completion.

Longer description:

On numerous occasions I have wished that I could have a little more control over what value(s) are returned by my `for/fold`s.

In particular, sometimes I need to perform a final computation before I'm truly done (maybe reverse a list I've been accumulating) and other times I'm only actually interested in a subset of the fold variables values persisting after the loop completes.

This PR was me just tinkering around seeing how easy it would be to add such a feature.

Here is my first ghetto version in use:

```racket
(define sorted-nums
  '(1 1 2 3 4 4 4 5 6 7 7 8 9 10 10))

;; returns a de-duped version of sorted-nums
;; that preserves the order from sorted-nums
(define deduped-sorted-nums
  (for/fold ([acc '()]
             [seen (hash)])
            #:return (reverse acc)
    ([x (in-list sorted-nums)])
    (cond
      [(hash-ref seen x #f) (values acc seen)]
      [else (values (cons x acc) (hash-set seen x #t))])))

deduped-sorted-nums
;; ==> '(1 2 3 4 5 6 7 8 9 10)
```

In that example, I have a hash I want to use during the for loop but am not interested in afterwards, and I want my accumulator list to be reversed once the for loop is done before being returned.

With the `#:return` expression I can specify the final result should only return one value, namely `(reverse acc)`.

I would find a feature like this really useful... thoughts?

### DIFF
--- a/racket/collects/racket/private/for.rkt
+++ b/racket/collects/racket/private/for.rkt
@@ -1541,6 +1541,8 @@
 
   (define-syntax for/fold/derived
     (syntax-rules ()
+      [(_ orig-stx ([fold-var finid-init] ...) #:return ret-expr . rest)
+       (for/foldX/derived/final [orig-stx #f] ([fold-var finid-init] ...) ret-expr . rest)]
       [(_ orig-stx ([fold-var finid-init] ...) . rest)
        (for/foldX/derived/final [orig-stx #f] ([fold-var finid-init] ...) (values* fold-var ...) . rest)]))
 


mflatt:
I like this idea.

pnwamk:
Quick questions:

1. Is the `#:return` and it's location in the `for/fold` form okay as is? (i.e. is there a better keyword and/or location?)

2. Is the current approach to adding it reasonable? I was happy to learn I'm really just exposing hooks that are already part of the `for/fold` implementation -- does that look like the right way to add this feature?

If, in its current form, people are more or less happy with it -- I'd be happy to add it to `for*/fold` as well and add some tests over the next few days (maybe get it in for 6.11?).

mflatt:
The `#:return` keyword feels imperative, but I have no strong opinion.

The implementation seems right to me.

I don't think this should go in 6.11, though, since we're in the "don't add new stuff" week.

I think something like this has been proposed in the past, and we didn't make the change then – but I don't remember why. Maybe it was that keywords weren't so established at the time, so we didn't have a lightweight syntax for adding a return option.

samth:
How about `#:result`?

pnwamk:
or `#:returns` so it reads more like a description and not a command?

rfindler:
This seems like a nice improvement to me.

Lets talk indentation. The example can't be right, right? How do you want DrRacket to indent this? (Please do not answer in a way that requires DrRacket to detect the present of a keyword, unless you plan to follow up with a new design of the indenter ;) )

samth:
What about putting it inside the sequence of accumulators?

pnwamk:
Like so?

```racket
(for/fold ([acc '()]
           [seen (hash)]
           #:returns (reverse acc))
          ([x (in-list sorted-nums)])
  (cond
    [(hash-ref seen x #f) (values acc seen)]
    [else (values (cons x acc) (hash-set seen x #t))]))
```

rfindler:
I'm not completely sold on my own thought, but what if it is not indented differently than the body but is the first (or last) thing in the body?

rfindler:
Oh, now that I see Sam's suggestion in action, I like it a lot, too.

pnwamk:
I was worried about putting it after the seq-exprs (e.g. in the body) because those bindings wouldn't be in scope for the return expression (right?).

rfindler:
Ah, right. It would make no sense to people if it were in the body.

mfelleisen:
+ω for the idea. (I might have asked for a bad version of this before.) 

I definitely think including the #:results-in expression in the accumulator section of the loop. And I’d almost go with another clause shape so that the indenter does not have to be changed: 

 (for/fold ([acc1 e10][acc2 e20][#:results-in (zip acc1 acc2)]) ([x (loop-over-this)] . . ) body . .)

samth:
I still prefer `#:result` and the placement I described. @mfelleisen I'm not sure what the extra `[]` helps -- DrRacket already indents my suggestion correctly.

Also I noticed that the example produces an error in terms of `for/fold/derived` with current Racket :)

pnwamk:
since the idea is popular, I've implemented some of the suggested changes and moved the PR here: https://github.com/racket/racket/pull/1834 in order to delete this accidental branch on the racket repo.

PR #1834, opened 2017-10-05
add for/fold #:result clause

pnwamk:
Initially introduced and discussed here https://github.com/racket/racket/pull/1832.

Allow users to specify the final return expression of a `for/fold`:

```racket
(for/fold ([acc '()]
           [seen (hash)]
           #:result (reverse acc))
          ([x (in-list '(0 1 1 2 3 4 4 4))])
  (cond
    [(hash-ref seen x #f)
     (values acc seen)]
    [else (values (cons x acc)
                  (hash-set seen x #t))]))
;; ==> '(0 1 2 3 4)
```

I've made some of the suggested changes, added documentation, and made it so the `#:result` clause does not duplicate code during macro expansion.

Also, I accidentally made the first PR from a branch on the racket repo instead of my own fork (doh!) -- so this fixes that =)

To do:
- [x] improve syntax error messages for `for/fold`, `for*/fold`, `for/fold/derived`, and `for*/fold/derived` (since the other features in this PR add more ways syntactic error messages could be poorly reported and expose implementation details, etc)
- [x] add tests
- [x] document new syntax

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/for.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/for.scrbl
@@ -309,8 +309,10 @@ result of the last evaluation of @racket[body]. If the
   (error "doesn't get here"))
 ]}
 
-@defform[(for/fold ([accum-id init-expr] ...) (for-clause ...)
-           body-or-break ... body)]{
+@defform/subs[(for/fold ([accum-id init-expr] ... maybe-result) (for-clause ...)
+                body-or-break ... body)
+              ([maybe-result (code:line)
+                             (code:line #:result result-expr)])]{
 
 Iterates like @racket[for]. Before iteration starts, the
 @racket[init-expr]s are evaluated to produce initial accumulator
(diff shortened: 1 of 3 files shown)


pnwamk:
Implemented, tested, and documented -- ready for review.

pkgs/racket-doc/scribblings/reference/for.scrbl
```diff
@@ -319,8 +321,11 @@ for each @racket[accum-id], and the corresponding current accumulator
 value is placed into the location. The last expression in
 @racket[body] must produce as many values as @racket[accum-id]s, and
 those values become the current accumulator values. When iteration
-terminates, the results of the @racket[for/fold] expression are the
-accumulator values.
+terminates, if a @racket[result-expr] is provided then the result of the
+ @racket[for/fold] is the result of evaluating @racket[result-expr]
+ (with the accumulator ids in scope and bound to their final values),
```
samth: I would explicitly reference `accum-id`s here.

pkgs/racket-test-core/tests/racket/for.rktl
```diff
@@ -441,6 +441,51 @@
         (define k i) #:final (= k 2)
         (list i j)))
 
+;; extra tests for #:result
```
samth: I think we should have a test where `#:result` produces multiple values.

samth:
This looks good to me.

samth:
Note that there's an optimization that would apply to most loops using this to avoid ever creating multiple values, but neither Racket nor Chez seems to do it currently.

pnwamk:
@mflatt when you have a sec, a sanity check on the implementation (and any other feedback of course) would be great!

mflatt:
The only issue I see is that a `history` note should be added to the docs. Otherwise, this looks good to me.

PR #1840, opened 2017-10-08
Adds optional keyword argument to check-duplicates to determine retur…

### LINKED ISSUE #1092: `check-duplicates` is ambiguous when returning `#f`
Function `check-duplicates` was added lately from `unstable/list` (used to be named `check-duplicate`). It's behavior is ambiguous: a result of `#f` either means the list has no duplicate, or the duplicate value is `#f`. In fact, the contract's range is `(or/c any/c #f)`, which is essentially `any/c`.

As I see no optional argument to help disambiguating these situations, I guess this was overlooked rather than intended?

hzafar:
…ned value when no duplicate found.

Documentation updated to match that of hash-ref, which has a similar default argument.

(Fixes #1092)


pkgs/racket-test-core/tests/racket/list.rktl
```diff
@@ -336,6 +336,8 @@
 (test 'a check-duplicates '(a a b))
 (test '(a 3) check-duplicates '((a 1) (b 2) (a 3)) #:key car)
 (test 4 check-duplicates '(1 2 3 4 5 6) (lambda (x y) (equal? (modulo x 3) (modulo y 3))))
+(test #f check-duplicates '(#t #f #f) #:default "no dups")
+(test "no dups" check-duplicates '(#t #f) #:default "no dups")
```
bennn: Can you add a test where `#:default` is a procedure

pkgs/racket-doc/scribblings/reference/pairs.scrbl
```diff
@@ -1173,13 +1173,26 @@ traversal.
 
 @defproc[(check-duplicates [lst list?]
                            [same? (any/c any/c . -> . any/c) equal?]
-                           [#:key extract-key (-> any/c any/c) (lambda (x) x)])
-         (or/c any/c #f)]{
+                           [#:key extract-key (-> any/c any/c) (lambda (x) x)]
+                           [#:default failure-result (failure-result/c any/c)])
```
bennn: `#:default` is missing a `default-expr` (it needs a `lambda` like `#:key` has)

stamourv:
Looks like @bennn's comment was addressed. Sorry it too so long to notice! And thank you for your contribution!

PR #1841, opened 2017-10-10
Clarify some behavior of the default (compiled-)load handler.

shhyou:
Reflect the default behavior of the load handler and
the compiled-load handler when the source file does
not exist.


mflatt:
LGTM

PR #1844, opened 2017-10-13
Add #:tag to the docs for define-cpointer-type

LeifAndersen:


mflatt:
Sorry for being late here, but the docs should have `tag-expr`, not `tag-id`. That is, the optional `#:tag` form has an expression that should produce a tag value, not necessarily an identifier.

LeifAndersen:
Woops, my apologies. I would be happy to change that. While I'm at it, anything else that you would like changed?

mflatt:
Nothing else - thanks for the improvement!

PR #1847, opened 2017-10-15
Make splicing-syntax-parameterize recur into module* submodules with #f instead of a module path

lexi-lambda:


pkgs/racket-doc/scribblings/reference/splicing.scrbl
```diff
@@ -81,7 +86,12 @@ export specifications will expand as if they had not been inside of the
 (splicing-syntax-parameterize ([place (lambda (stx) #'"Oz")])
   (define here (where)))
 here
-]}
+]
+
+@history[
+ #:changed "6.11.0.1"
+ @elem{Modified to syntax parameterize @racket[module*] submodules with
+       @racket[#f] instead of a @tech{module path}.}]}
```
mflatt: Nit: It sounds like the change was to stop parameterizing  submodules that have a module path. Something like `@racket[#f] (as opposed to a @tech{module path})` would be clearer to me.
But that's minor, and I'll merge as-is unless you feel like adjusting the text in the next few minutes.

lexi-lambda: Thanks, that’s a good point. I’ve tweaked the wording slightly, though in a different way from your suggestion, which I hope is less ambiguous.

PR #1852, opened 2017-10-17
Fix curry input contract

iitalics:
`curry` now immediately raises a contract exception when immediatly given a non-procedure argument.


bennn:
Merged!
27ec348a62a8dbae617ea99d0d465494ad310b08

PR #1853, opened 2017-10-17
Improve how splicing-syntax-parameterize interacts with module* forms

lexi-lambda:
This is a refinement on the previous PR related to this, which introduced commit 00d438cfbe8ea4d8db8277fc712d3315cf4a9e5f. This commit does it in a much more careful way, manually emulating how the macroexpander expands `module*` forms in order to allow `splicing-syntax-parameterize` to apply even within `#%module-begin` forms introduced by the expander.

### DIFF
--- a/racket/collects/racket/splicing.rkt
+++ b/racket/collects/racket/splicing.rkt
@@ -343,9 +343,13 @@
                (syntax/loc/props body
                  (begin-for-syntax (wrap-param-et e (orig-id ...) (temp-id ...)) ...))]
               [(module . _) body]
-              [(module* n #f form ...)
-               (syntax/loc/props body
-                 (module* n #f (expand-ssp-body (sp-id ...) (temp-id ...) (orig-id ...) form) ...))]
+              [(module* name #f form ...)
+               (datum->syntax body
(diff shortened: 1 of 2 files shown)


racket/collects/racket/splicing.rkt
```diff
@@ -343,9 +343,14 @@
                (syntax/loc/props body
                  (begin-for-syntax (wrap-param-et e (orig-id ...) (temp-id ...)) ...))]
               [(module . _) body]
-              [(module* n #f form ...)
-               (syntax/loc/props body
-                 (module* n #f (expand-ssp-body (sp-id ...) (temp-id ...) (orig-id ...) form) ...))]
+              [(module* name #f form ...)
+               (with-syntax ([#%module-begin (datum->syntax body '#%module-begin)])
```
mflatt: Is this binding used? I think it would have to appear syntactically in the `with-syntax` body to be relevant.

lexi-lambda: Whoops, you’re right—that’s vestigial from an older version of the code. I’ll remove it.

PR #1854, opened 2017-10-17
remove precomiled sorts

pnwamk:
TL;DR the cache of precompiled sort functions is rarely ever hit... would it be better to get rid of it?

The current implementation of `sort` dispatches to precompiled versions of `sort` when they use common comparison operators (e.g. `keyword<?`).

Perusing Racket code bases, it seems _extremely rare_ for someone to sort a list/vector using a _primitive comparison operator_ (e.g. string<?, keyword<?, etc) with no `#:key` argument.

Since providing a `#:key` argument or using a custom comparison function means we don't use one of the precompiled versions, it appears usages of `sort` _rarely_ benefit from any of the precompiled code.

In a quick experiment, I added a `printf` to `sort` and ran `raco setup` on a fresh install and it appears there were 40,311 cache misses and 33 cache hits (30 for keyword<?, 3 for string<? -- a hit rate of about 0.08%).

One last note, the size of the .zo file for `sort` is obviously impacted by this caching. Back in Racket v6.5 it was ~25k, recently after adding `vector-sort` and some additional precompiled versions the size was brought up to ~45k (i.e. the current size -- sorry, that's my fault!). Removing precompiled sorts altogether brings the .zo size down to ~15k.

Should we remove the precompiled versions (this PR does that)? Should we remove all but the most commonly (but still rather rarely) used comparators, e.g. `keyword<?`? Should we do nothing?

### DIFF
--- a/racket/collects/racket/private/sort.rkt
+++ b/racket/collects/racket/private/sort.rkt
@@ -17,8 +17,7 @@ redundant pointer arithmetic in that paper is dealing with cases of uneven
 number of elements.)
 
 The source uses macros to optimize some common cases (eg, no `getkey'
-function, or precompiled versions with inlinable common comparison
-predicates) -- they are local macros so they're not left in the compiled
+function) -- they are local macros so they're not left in the compiled
 code.
 
 |#
@@ -151,34 +150,6 @@ code.
           (unless (zero? n/2-) (copying-mergesort Alo Amid2 n/2-))
           (merge #f B1lo (i+ B1lo n/2+) Amid2 Ahi Alo)))))
   
-  ;; - - - - - - - - - - - - - - - - - - - - - - - -
-  ;; Precompiling of standard comparison functions
-  ;; for standard data types
-  ;; - - - - - - - - - - - - - - - - - - - - - - - -
-  (define precompiled-sorts
-    (let ([sorts (make-hasheq)])
-      (define-syntax-rule (precomp less-than? more ...)
-        (let ([sort-proc
-               (λ (A n) (sort-internal-body A less-than? n #f))])
-          (hash-set! sorts less-than? sort-proc)
-          (hash-set! sorts more sort-proc) ...))
-      ;; for comparison ops provided by racket/base we build
-      ;; fast precompiled versions
-      (precomp unsafe-fl< unsafe-fl<=)
-      (precomp unsafe-fl> unsafe-fl>=)
-      (precomp i< i<=)
-      (precomp i> i>=)
-      (precomp < <=)
-      (precomp > >=)
-      (precomp string<? string<=?)
-      (precomp string>? string>=?)
-      (precomp string-ci<? string-ci<=?)
-      (precomp string-ci>? string-ci>=?)
-      (precomp char<? char<=?)
-      (precomp char>? char>=?)
-      (precomp keyword<?)
-      (make-immutable-hasheq (hash-map sorts cons))))
-  
   (define (generic-sort A less-than? n)
     (sort-internal-body A less-than? n #f))
   
@@ -188,12 +159,7 @@ code.
   (define-syntax (sort-internal stx)
     (syntax-case stx ()
       [(_ vec less-than? n #:key #f)
-       #'(let ([precomp (hash-ref precompiled-sorts less-than? #f)])
-           (if precomp
-               ;; use a precompiled function if found
-               (precomp vec n)
-               ;; otherwise, use the generic code
-               (generic-sort vec less-than? n)))]
+       #'(generic-sort vec less-than? n)]
       [(_ vec less-than? n #:key key)
        #'(generic-sort/key vec less-than? n key)]))
   


rfindler:
Another option is to detect a commonly used comparator in the call site and
at that point construct the specialized version and deposit it into the
client module.

Also: check the shootout benchmarks.

And how much faster is the specialized version these days?

pnwamk:
One note: after chatting w/ @samth I realized running `raco setup` may not necessarily reflect all common use cases of `sort` "in the wild" since it is just building all the modules, etc. So I'll have to do more sampling and some benchmarking, as @rfindler suggests to further clarify what might make sense.

rfindler:
PS: I think it is great you're looking into this. This code isn't very recent and the compiler tradeoffs have changed since then (and are about to change again in a big way, of course!). I expect you'll find a point that makes a lot more sense nowadays.

mfelleisen:
Would it make any sense to run all the benchmarks and tests we have with counting turned on? 

Would it make even more sense to equip drdr with ‘requests’ that collect bespoke statistics for the next N runs?

pnwamk:
Shootout with Racket HEAD:
https://gist.github.com/pnwamk/4a4e34a9bf7f5f831def95f7984d300b

Shootout with this PR:
https://gist.github.com/pnwamk/e0acfafd6bdadf6198562d8c28c4b960

Only those shootout results that explicitly use `sort` in the source code:

```
moments (no obvious cache hits from source code)

with precompiled sorts
----------------------
[racket moments (2288 2285 320) 12]
[racket moments (2292 2288 320) 12]
[racket moments (2308 2304 328) 12]

without precompiled sorts
-------------------------
[racket moments (2316 2312 332) 15]
[racket moments (2284 2284 316) 15]
[racket moments (2292 2292 324) 15]






k-nucleotide (no obvious cache hits from source code)

with precompiled sorts
----------------------
[racket k-nucleotide (2936 2936 56) 18]
[racket k-nucleotide (2940 2937 52) 18]
[racket k-nucleotide (2916 2910 52) 18]


without precompiled sorts
-------------------------
[racket k-nucleotide (2924 2921 52) 16]
[racket k-nucleotide (2940 2938 52) 16]
[racket k-nucleotide (2916 2913 48) 16]






meteor (should hit a cached version of sort for `>`)

with precompiled sorts
----------------------
[racket meteor (296 298 0) 47]
[racket meteor (300 298 0) 47]
[racket meteor (300 300 0) 47]


without precompiled sorts
-------------------------
[racket meteor (300 299 0) 49]
[racket meteor (304 304 0) 49]
[racket meteor (296 299 0) 49]



wordfreq (should hit a cached sort for `string>?`)

with precompiled sorts
----------------------
[racket wordfreq (2748 2749 20) 69]
[racket wordfreq (2812 2809 16) 69]
[racket wordfreq (2744 2741 16) 69]


without precompiled sorts
-------------------------
[racket wordfreq (2824 2825 20) 68]
[racket wordfreq (2856 2856 24) 68]
[racket wordfreq (2808 2806 20) 68]
```

pnwamk:
a quickly thrown together microbenchmark of just sorting with a comparison operator that has a precompiled version in the current `sort` implementation:

```racket
#lang racket/base


(define short-lists '(()
                      (1)
                      (1 2)
                      (2 1)
                      (1 2 3)
                      (3 2 1)
                      (1 2 3 4)
                      (5 4 3 2 1)
                      (1 2 3 4 5 6 7 8 9 10)
                      (10 9 8 7 6 5 4 3 2 1)))

(define list100 (build-list 100 (λ _ (random 100))))
(define list1000 (build-list 1000 (λ _ (random 1000))))
(define list10000 (build-list 10000 (λ _ (random 10000))))


(printf "sorting short lists:\n")
(time (for* ([_ (in-range 100000)]
             [l (in-list short-lists)])
        (sort l <)))

(printf "sorting list of length 100:\n")
(time (for* ([_ (in-range 10000)])
        (sort list100 <)))

(printf "sorting list of length 1000:\n")
(time (for* ([_ (in-range 10000)])
        (sort list1000 <)))

(printf "sorting list of length 10000:\n")
(time (for* ([_ (in-range 10000)])
        (sort list10000 <)))
```

results with precompiled sorts:

```
> racket sorting.rkt      
sorting short lists:
cpu time: 140 real time: 140 gc time: 4
sorting list of length 100:
cpu time: 56 real time: 55 gc time: 0
sorting list of length 1000:
cpu time: 1100 real time: 1100 gc time: 4
sorting list of length 10000:
cpu time: 15148 real time: 15145 gc time: 52
```


results without precompiled sorts:

```
> racket sorting.rkt      
sorting short lists:
cpu time: 164 real time: 165 gc time: 0
sorting list of length 100:
cpu time: 128 real time: 126 gc time: 0
sorting list of length 1000:
cpu time: 1812 real time: 1811 gc time: 20
sorting list of length 10000:
cpu time: 23948 real time: 23940 gc time: 52
```

pnwamk:
Obviously in the microbenchmarks, replacing `<` with `(λ (x y) (< x y))` causes both to run in the same time (since the comparison function has no precompiled version).

Also, obviously the microbenchmarks ignore the cache hit rate (and I still don't have numbers on what that looks like outside of just running `raco make`, but I worry it is common to use a custom comparison function or a #:key argument -- but I could be wrong).

mflatt:
I offer no conclusion, but FWIW: These precompiled sorts were a hassle for the linklet flattener, so I considered removing them at the time. I decided to keep them after seeing the difference in microbenchmarks, but only because that seemed safer. If no real code hits the cache, that's a good argument for getting rid of it.

pnwamk:
I threw together a quick and dirty script that checks for usages of `(sort ...)` and tries to determine if they would be a cache hit or miss based on whether or not keyword args were present, and what the literal symbol for the sort function was:

https://gist.github.com/pnwamk/5f507da57a3276da81f58eb3a24e170f

Here are the results from running that on all the .rkt files (18,058) from pkgs.racket-lang.org:

```
total usages of sort found: 915
confirmed cache misses (i.e. used #:key): 342
probable cache misses (i.e. symbol didn't match precompile list): 317
cache hits (symbol matched a precompiled function): 234
unknown (i.e. didn't match syntax for normal sort usage): 22
hit counts for precompiled versions:
((char<? . 3)
 (< . 96)
 (string<? . 72)
 (<= . 2)
 (string>? . 6)
 (>= . 2)
 (keyword<? . 14)
 (string-ci<=? . 3)
 (string<=? . 22)
 (> . 13)
 (string-ci<? . 1))
```

Again, this doesn't really indicate how _often_ the cache is being hit or missed during execution, and it doesn't exclude usages that are uninteresting (i.e. some are trivial tests that merely sort a list of symbols so the test can compare two lists w/o worrying about order, etc).

bennn:
@pnwamk I think you should email the racket-dev list about this pull request. (Maybe someone there has strong options / requirements, but hasn't seen this yet.)

mflatt:
@pnwamk Thanks for all the investigation here! I'm now convinced that we should get rid of the precompiled sorts.

For Racket-on-Chez, the results for your benchmark are essentially the same with precompiled sorts. Without precompiled sorts, Racket-on-Chez performs in between the performance of Racket with precompiled sorts and without. That's about what we should expect: Racket is going to produce similar code for simple loops with unsafe operations, while Chez Scheme is going to do better with higher-order programs.

I also checked the performance of the Racket expander without precompiled sorts, since the sorting code gets pulled into the expander's implementation. I expected a 5% decrease in the expander's code size, and that happened. I didn't expect a 5-10% speedup from removing the precompiled sorts, but that also happened (and it's not just a decrease in GC due to less code being around, although that's a small part). The effect is consistent across racket7 with the expander JITted, racket7 with the expander compiled to C, and Racket-on-Chez. It's probably worth exploring more why the expander is sensitive to sorting overhead, but meanwhile...

Given your surveys and measurements and the effect on the expander, I think we should merge this PR.

pnwamk:
These changes were incorporated by 59ef2543182e0445979a2245dbaa06d1cdca35fd.

Issue #1866, opened 2017-10-24
Add vector-cas! operation


minad:
Hi,

I would like to use a cas operation on vectors, but only found `box-cas!`.
Would you accept a PR which adds this operation?

~~~ scheme
(vector-cas! vec pos old new) → boolean?
~~~

samth:
Yes, a PR for this would be great.

minad:
Cool! I will look into that!

PR #1867, opened 2017-10-25
add vector-cas! and unsafe-vector*-cas!

minad:
See #1866. The implementation is similar to `box-cas!`. There is no support for chaperones and impersonators. The operation is not jitted.


pkgs/racket-doc/scribblings/reference/unsafe.scrbl
```diff
@@ -246,10 +246,11 @@ Unsafe versions of @racket[unbox] and @racket[set-box!], where the
 @defproc[(unsafe-vector*-length [v (and/c vector? (not/c impersonator?))]) fixnum?]
 @defproc[(unsafe-vector*-ref [v (and/c vector? (not/c impersonator?))] [k fixnum?]) any/c]
 @defproc[(unsafe-vector*-set! [v (and/c vector? (not/c impersonator?))] [k fixnum?] [val any/c]) void?]
+@defproc[(unsafe-vector*-cas! [v (and/c vector? (not/c impersonator?))] [k fixnum?] [val any/c] [val any/c]) boolean?]
```
samth: This should probably have a history note too.

samth:
This looks good to me, although review from @mflatt would be good too.

Another nice thing to add would be an `unsafe-struct-cas!` operation.

pkgs/racket-doc/scribblings/reference/unsafe.scrbl
minad: done

minad:
@samth Thank you for your review! A `unsafe-struct*-cas!` operation would look very much like the `unsafe-vector*-cas!` operation. Would it make sense to also add some safe cas operations for structs? But I suggest to address struct cas in a separate issue/PR.

pkgs/racket-doc/scribblings/reference/vectors.scrbl
```diff
@@ -72,6 +72,17 @@ slot is position @racket[0], and the last slot is one less than
 
 Updates the slot @racket[pos] of @racket[vec] to contain @racket[v].}
 
+@defproc[(vector-cas! [vec (and/c vector? (not/c immutable?))]
```
mflatt: I think the contract here needs `(not/c impersonator?)`

racket/src/racket/src/vector.c
```diff
@@ -633,6 +645,25 @@ scheme_checked_vector_set(int argc, Scheme_Object *argv[])
...
 }
 
+Scheme_Object *
+scheme_checked_vector_cas(int argc, Scheme_Object *argv[])
+{
+  Scheme_Object *vec = argv[0];
+  intptr_t i, len;
+
+  if (!SCHEME_MUTABLE_VECTORP(vec))
+    scheme_wrong_contract("vector-cas!", "(and/c vector? (not/c immutable?))", 0, argc, argv);
```
mflatt: The error-message contract here also needs `(not/c impersonator?)`

racket/src/racket/src/vector.c
```diff
@@ -1076,6 +1107,28 @@ static Scheme_Object *unsafe_vector_star_set (int argc, Scheme_Object *argv[])
...
+static Scheme_Object *unsafe_vector_star_cas (int argc, Scheme_Object *argv[])
+{
+  Scheme_Object *vec = argv[0];
+  Scheme_Object *idx = argv[1];
+  Scheme_Object *ov = argv[2];
+  Scheme_Object *nv = argv[3];
+
+#ifdef MZ_USE_FUTURES
+  return mzrt_cas((volatile size_t *)(SCHEME_VEC_ELS(vec) + SCHEME_INT_VAL(idx)),
+                  (size_t)ov, (size_t)nv)
```
mflatt: Instead of `size_t`, I'd prefer to use `uintptr_t`, since the rest of the implementation uses that type for pointer-sized integers.

minad: I agree. I changed it in the box-cas! implementation too.

mflatt:
Merged - thanks!

rfindler:
Not sure if this is the right place to note this, but the openbsd  builds last night failed with an error message that suggests that this change may be relevant:

https://plt.eecs.northwestern.edu/snapshots/20171026-dfa5d48092/

(follow the "build failed" link for more information)

minad:
@rfindler yes there is a minor copy paste error in unsafe_struct_star_cas (ping @mflatt).
@mflatt, btw thanks for adding the jit implementation!

mflatt:
Fixed in 5f1b70788113a05309c50710ee8ad13f1a3b2154. There was an extra bug in the JIT part.

PR #1877, opened 2017-11-04
Support string keys in hashs destined for json serialization.

### LINKED ISSUE #1876: `#hasheq(("turnip" . 82))` not valid `jsexpr?`
`json` seems to have an arbitrary definition of what can be written out as JSON. I assume this is so that there's some consistency in a round trip of `(string->jsexpr (jsexpr->string foo))`. However, this behavior is odd, especially considering that JSON object keys *must* be strings by specification.

There's an incoming PR for this.

apg:
fixes #1876


apg:
Closing this PR as it presents a backwards incompatible change.

PR #1878, opened 2017-11-04
JSON serialize structs with prop:json-serialize

apg:
Yesterday on #irc, someone asked about JSON serializing arbitrary structs easily. I had suggested that I didn't know of a way (confirmed by others) but thought it'd be a great addition.

I went to @samth with an idea to do the `to-json` part via a struct property, and quickly made a working prototype. This is that prototype, cleaned up a little bit, presented with the intention of starting a discussion around it's inclusion. Is it an alright idea given there's no parsing story? Are there problems with this basic implementation?

Possible enhancements:
- `make-json-serializable-info` would be nicer, probably, with a contract.

This builds on top of #1877, a bug fix to allow string keys in hashes to be considered `jsexpr?`, but does not actually depend on the fix, itself.

### DIFF
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -35,7 +35,7 @@ the @rfc for more information about JSON.
     @item{@racket[string?]}
     @item{@racket[(or/c exact-integer? inexact-real?)]}
     @item{@racket[(listof jsexpr?)]}
-    @item{@racket[(and/c hash-eq? (hash/c symbol? jsexpr?))]}]
+    @item{@racket[(and/c hash-eq? (hash/c (or/c symbol? string?) jsexpr?))]}]
 
 @examples[#:eval ev
   (jsexpr? 'null)
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -35,7 +35,8 @@ the @rfc for more information about JSON.
     @item{@racket[string?]}
     @item{@racket[(or/c exact-integer? inexact-real?)]}
     @item{@racket[(listof jsexpr?)]}
-    @item{@racket[(and/c hash-eq? (hash/c (or/c symbol? string?) jsexpr?))]}]
+    @item{@racket[(and/c hash-eq? (hash/c (or/c symbol? string?) jsexpr?))]}
+    @item{@racket[(json-serializable-struct?)]}]
 
 @examples[#:eval ev
   (jsexpr? 'null)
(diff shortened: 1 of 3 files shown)


samth:
Would it be better to use generics here?

apg:
That’s a good question, that you’d be better equipped to answer. From a user’s perspective they’ll still need to provide a function—likely the same exact function as the `prop` version. 

I don’t see an actual advantage with generics in deserialization, and suspect we’d probably need to introduce another generic if someone figured out a decent approach to it down the line, to avoid suddenly not being fully compliant with the json generic.

I have to read up more, but if we could use generics to extend builtin, primitive types (I think that’s what `#:defaults` would allow?), then that might be an advantage as we could refactor the whole of `json` to break up the large `cond` in `write-json*` and `jsexpr?`, which would actually make it completely worth while, in my opinion.

I guess we could call it `gen:jsexpr`, too, and get the predicate for free!

Note: the current version was cribbed off of `prop:serializable-struct` (or whatever the name is?).

LeifAndersen:
In the case of serialization, I think struct properties make more sense than a generic, because you are attaching a piece of information (a serialize function AND the module/function name for where to find it) to a struct definition, rather than defining a generic method. (I know you _could_ in principle do it in either, it just feels cleaner as a property.)

I mean, it would be nice to do it as a generic with two functions (serialize and deserialize), but the deserialize function needs to be bound to a variable at the top level.

apg:
@LeifAndersen IIUC, you are in favor of the property, then?

I’m beginning to think the property makes sense here, too, for reasons of setting the stage for appropriate deserialization, later. It’d be much more natural to, say, provide a mapping from JSON keys to the appropriate `set!` functions (or something) than to try to figure out a way to do it with generics.

Though, perhaps that deserialization mapping could be provided as a property, and used by the generic deserialize function...

apg:
The backwards incompatibility presented in the discussion of #1876 exists in this PR, too.

pmatos:
This is an interesting PR. What's missing?

rmculpepper:
I see a few problems, but they seem fixable.

Adding variants to `jsexpr?` may break existing code. One solution is to leave `jsexpr?` alone and add a separate `jsexpr-convertible?` predicate (and operations, like `jsexpr-convert`).

I wouldn't call this "serialization" since there's no deserialize operation. It's more like a JSON version of `prop:custom-write` (or `gen:custom-write`, if you prefer). I see that deserialization was mentioned in the discussion above, but it's not addressed by the PR.

There's no need for the `json-serialize-info` struct wrapper around the conversion procedure.

Aside from that, though, I think it might be better to make this a separate little library, though. The main benefit I see to adding it to the core `json` library is that `write-json` could convert structs incrementally rather than doing conversion first (possibly allocating a large new data structure) and then writing. But the corresponding cost is that `write-json` could encounter a bad converted value in the middle of a write.

PR #1880, opened 2017-11-05
doc: more links in 'json' docs

bennn:
* Change 'any?' to 'any/c'
* Add for-label require so 'with-XXX-from-string' is a link in the examples


bennn:
Build failed because file tests are still hanging

PR #1882, opened 2017-11-05
filesystem-change-evt? typo

spdegabrielle:
changed
Returns #f if v is a filesystem change event, #f otherwise.
to 
Returns #t if v is a filesystem change event, #f otherwise.


stamourv:
Thanks!

PR #1891, opened 2017-11-15
Quoting URLs in sendurl

### LINKED ISSUE #1890: View documentation doesn't work
Consider the code:

```
#lang racket

+
```

Right click on `+` and click "View documentation..."

A Racket internal error dialog pops up with the following message:

```
browser-run: process execute failed: '("vivaldi file:///usr/share/doc/racket/reference/generic-numbers.html#(def._((quote._~23~25kernel)._%2B))")
/bin/sh: 1: Syntax error: "(" unexpected
```

Version 6.10.1, Ubuntu 17.10

sorawee:
This fixes #1890. Note that because `"` is not a valid character in URL, it won't appear in `url`.

### DIFF
--- a/pkgs/net-lib/net/sendurl.rkt
+++ b/pkgs/net-lib/net/sendurl.rkt
@@ -247,7 +247,7 @@
     [(not browser)
      (error 'send-url "Couldn't find a browser to open URL: ~e" url)]
     [(custom-browser? browser)
-     (browser-run #:shell #t (string-append (car browser) "\"" url "\"" (cdr browser)))]
+     (browser-run #:shell #t (format "~a~s~a" (car browser) url (cdr browser)))]
     ;; if it's a known browser, then it must be an existing one at this point
     [(not exe) (error 'send-url "internal error")]
     ;; if it's gone throw an error (refiltering will break assumptions of
--- a/pkgs/net-lib/net/sendurl.rkt
+++ b/pkgs/net-lib/net/sendurl.rkt
@@ -247,7 +247,7 @@
     [(not browser)
      (error 'send-url "Couldn't find a browser to open URL: ~e" url)]
     [(custom-browser? browser)
-     (browser-run #:shell #t (string-append (car browser) url (cdr browser)))]
+     (browser-run #:shell #t (string-append (car browser) "\"" url "\"" (cdr browser)))]
     ;; if it's a known browser, then it must be an existing one at this point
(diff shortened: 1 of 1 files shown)


jeapostrophe:
I think you should do (format "~a~s~a" ....) instead so that Racket will add the "s and escapes inside if necessary. (I.e. for the string "foo\"bar"

mflatt:
Just to make sure it's clear in the thread here: I currently think this change shouldn't be made, as explained in my comment for #1890

samth:
Maybe a different approach would be to have a new preference with a better way of specifying it, as well as keeping this preference for the reasons @mflatt mentions.

rfindler:
Don't different shells have different quoting mechanisms and don't they already have people putting in work to make good error messages? Perhaps a better approach is to try to get more information into the error message and to give more immediate feedback (in the preferences setting) somehow?

samth:
I believe that different shells do have subtly different quoting rules, although I don't know if we'll run afoul of them. I just tried a few shells to see what the error messages are, and I wouldn't call any of them good.

rfindler:
Well, perhaps I spoke too softly :). I don't think we should be in the business of producing good error messages for incorrectly formatted commands, nor should we be in the business of finding a common quoting format.

What about a checkbox (whose layout makes it easy to find, near the current GUI) whose effect is to add quotes in a few key places? That is-- we do not make a change in this repo at all, but do it at the drracket (or framework?) level?

samth:
I think we should take @mflatt's implicit advice in #1890 and come up with a preference that doesn't involve `string-append`.

A simple one, that would have a preference similar to the current one, would be a command name (must be findable via `find-executable-path`), a list of before-arguments, and a list of after-arguments. The two lists could be specified by writing them space-separated in the preferences dialog, but would in most cases (such as the one in #1890) be entirely unneeded, and all you'd have to enter is the name of the command.

samth:
Note that that suggestion is agreeing with @rfindler about not doing any common quoting format or error handling, but would need another line for handling the new format, which would look like:

```
[(custom-browser2? browser) 
 (apply browser-run `(,(custom-cmd browser) ,@(custom-before browser)  ,url ,@(custom-after browser)))]
```

mflatt:
A note on shells: While different shells do have different quoting rules, `system` always uses "/bin/sh" on Unix.

sorawee:
Abandon

PR #1903, opened 2017-12-02
CRLF terminate get-pure-port/headers; fixes #1901

### LINKED ISSUE #1901: get-pure-port/headers differs from purify-port breaking extract-field
```racket
#lang racket/base

(require net/head
         net/url)

(define u (string->url "https://www.racket-lang.org"))

;; purify-port returns the \r\n terminated headers including the final
;; \r\n (i.e. the blank line between the headers and the body
;;
;; extract-field expects this and can extract all the headers
;; including the final one -- in this case the CF-RAY header

(define h0 (purify-port (get-impure-port u)))
h0 ;"HTTP/1.1 200 OK\r\nDate: Fri, 01 Dec 2017 17:37:45 GMT\r\nContent-Type: text/html; charset=utf-8\r\nTransfer-Encoding: chunked\r\nConnection: keep-alive\r\nSet-Cookie: __cfduid=d157472d41d231e12d4881d2e765456f71512149865; expires=Sat, 01-Dec-18 17:37:45 GMT; path=/; domain=.racket-lang.org; HttpOnly\r\nx-amz-id-2: iKQuWfsvLZKqfAGiCDShxiw7nTT7kiGjJGJf0Q8nalPt0qRxxz4ccOyyrvhRckGo3f28UPpB+K0=\r\nx-amz-request-id: 0B5BAA0D0F1A4539\r\nCache-Control: max-age=0, no-cache\r\nLast-Modified: Mon, 06 Nov 2017 15:32:34 GMT\r\nx-amz-storage-class: REDUCED_REDUNDANCY\r\nServer: cloudflare-nginx\r\nCF-RAY: 3c67d0f2fa740ecd-EWR\r\n\r\n"
(extract-field "Date" h0)   ;"Fri, 01 Dec 2017 17:37:46 GMT"
(extract-field "CF-RAY" h0) ;"3c67d0f2fa740ecd-EWR"


;; get-pure-port/headers returns the \r\n terminated headers -- but
;; does NOT include the final \r\n
;;
;; extract-field doesn't expect this; although it will extract the
;; headers, the final header gets a spurious \r\n included in its
;; value

(define-values (_ h1) (get-pure-port/headers u #:status? #t))
h1 ;"HTTP/1.1 200 OK\r\nDate: Fri, 01 Dec 2017 17:37:46 GMT\r\nContent-Type: text/html; charset=utf-8\r\nTransfer-Encoding: chunked\r\nConnection: keep-alive\r\nSet-Cookie: __cfduid=d6f908755e154a2f9a02c18b8edb3b5d01512149865; expires=Sat, 01-Dec-18 17:37:45 GMT; path=/; domain=.racket-lang.org; HttpOnly\r\nx-amz-id-2: RR4uLE2k9Lf13jZ80vVAR5LPxq2wddoyBuWtQyWTOkHtIZMAoXTUVoJck9CAFevXLZB0CSZV2es=\r\nx-amz-request-id: ECD76869C640565A\r\nCache-Control: max-age=0, no-cache\r\nLast-Modified: Mon, 06 Nov 2017 15:32:34 GMT\r\nx-amz-storage-class: REDUCED_REDUNDANCY\r\nServer: cloudflare-nginx\r\nCF-RAY: 3c67d0f418c3219e-EWR\r\n"
(extract-field "Date" h1)   ;"Fri, 01 Dec 2017 17:37:46 GMT"
(extract-field "CF-RAY" h1) ;"3c67d0f418c3219e-EWR\r\n"
```

`extract-field` seems to get confused only on the _last_ header -- so, you might test and things work fine but another server happens to order headers differently and then you get bit. At least, that's what happened to me, where the header was `Content-Encoding` and I was matching on the value `"gzip"` but got `"gzip\r\n"`, and I didn't gunzip the bytes.

Probably `get-pure-port/headers` should change and append a final `\r\n`?

Or if we worry that would break existing code, then maybe its documentation should just get a new, "Tip: You should `string-append` a `"\r\n"` to the header string before using it with `extract-field`"?

greghendershott:
Supply a final CRLF just like purify-port.

### DIFF
--- a/pkgs/net-test/tests/net/url-port.rkt
+++ b/pkgs/net-test/tests/net/url-port.rkt
@@ -102,25 +102,25 @@
     "HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\nTransfer-Encoding: chunked\r\n\r\n24\r\nThis is the data in the first chunk \r\n1A\r\nand this is the second one\r\n0\r\n")
    =>
    (values "This is the data in the first chunk and this is the second one"
-           "Content-Type: text/plain\r\nTransfer-Encoding: chunked\r\n")
+           "Content-Type: text/plain\r\nTransfer-Encoding: chunked\r\n\r\n")
 
    (get-pure/headers
     "HTTP/1.0 200 OK\r\nContent-Type: text/plain\r\n\r\nThis is the data in the first chunk and this is the second one")
    =>
    (values "This is the data in the first chunk and this is the second one"
-           "Content-Type: text/plain\r\n")
+           "Content-Type: text/plain\r\n\r\n")
 
    (get-pure/headers
     "HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\nTransfer-Encoding: chunked\r\n\r\n20\r\nThis is the data in the first ch\r\n21\r\nand this is the second oneXXXXXXX\r\n0\r\n")
    =>
    (values "This is the data in the first chand this is the second oneXXXXXXX"
-           "Content-Type: text/plain\r\nTransfer-Encoding: chunked\r\n")
+           "Content-Type: text/plain\r\nTransfer-Encoding: chunked\r\n\r\n")
 
    (get-pure/headers
     "HTTP/1.1 200 OK\r\nContent-Type: text/plain\r\nTransfer-Encoding: chunked\r\nAnother-Header: ta-daa\r\n\r\n20\r\nThis is the data in the first ch\r\n21\r\nand this is the second oneXXXXXXX\r\n0\r\n")
    =>
    (values "This is the data in the first chand this is the second oneXXXXXXX"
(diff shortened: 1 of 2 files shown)


jeapostrophe:
Can you add a test that fails in the old world and succeeds in the new? (Like from the issue?)

greghendershott:
Oops. I read the email from GitHub and pushed a commit to add a test. Coming here to GitHub, and reading the Travis errors from my first commit, I now see that:

- I put the test in the wrong file. Should be `url-port.rkt` not `url.rkt`.
- The existing tests in `url-port.rkt` were already failing, correctly, because they're checking for single terminating `"\r\n"`. IOW I can/should simply update those existing tests.

I'll work on this and force-push to replace the test commit.

greghendershott:
1. I squashed to a single commit with the change and the updated tests.

2. Tests pass on Appveyor and Travis, except for the two macOS jobs -- but from reading the logs on one of them I think the test failures are unrelated to my commit?

jeapostrophe:
Alright, sounds good. I'll merge.

PR #1904, opened 2017-12-05
remove scheme_complex_izi_type from docs

gus-massa:
I can't find `scheme_complex_izi_type` anywhere in the code base. I think it was used in a previous version. Also, this fix some explanation of the behavior of `real?` with complex that have an inexact `0.0` imaginary part.


mflatt:
Thanks!

PR #1912, opened 2017-12-13
Fix the document of struct-type-make-constructor

shhyou:
The function `struct-type-make-constructor` actually takes an optional second argument.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/struct-inspectors.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/struct-inspectors.scrbl
@@ -129,11 +129,14 @@ Returns eight values that provide information about the structure type
 If the type for @racket[struct-type] is not controlled by the current inspector,
 the @exnraise[exn:fail:contract].}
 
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/struct-inspectors.scrbl
```diff
@@ -129,11 +129,14 @@ Returns eight values that provide information about the structure type
 If the type for @racket[struct-type] is not controlled by the current inspector,
 the @exnraise[exn:fail:contract].}
 
-@defproc[(struct-type-make-constructor [struct-type struct-type?])
+@defproc[(struct-type-make-constructor [struct-type struct-type?]
+                                       [constructor-name symbol? #f])
```
bennn: `symbol?` should be `(or/c symbol? #f)`

shhyou: Oh! Thanks for catching this!

PR #1926, opened 2018-01-10
update copyright year to 2018

mbutterick:


mbutterick:
@stamourv: this should be merged into v6.12.

stamourv:
Absolutely. I'm still sorting out some kinks in the pre-release builds, but once the release is on good footing, I'll merge.

PR #1939, opened 2018-01-26
Add splicing-parameterize to racket/splicing

lexi-lambda:


mflatt:
This commit generally looks great.

The new dependency on the contract system increases the cost of using just this library with `racket/base` from around 1 MB to around 10 MB. Obviously, that extra cost does not apply if anything else in a program uses `racket/contract/base`, but it's worth considering whether a more primitive check is appropriate in this case — since the check is really simple, and since `raise-argument-error` will produce a pretty good error message.

lexi-lambda:
Reasonable. I’ve pushed a new version that removes the dependency on the contract system, and it also makes a couple other minor tweaks to prune the added scope from nested `(module ....)` declarations and `(#%require ....)` forms.

mflatt:
LGTM

PR #1942, opened 2018-01-30
typo: local-redirect 'page that is' => 'page is'

bennn:
I _ think_ the redirect page is supposed to say:

> This page is intended to redirect to the result of a search request. Since you're reading this, it seems that the redirection did not work.

If so, we should merge this. If not, I can change it to the right wording.


mflatt:
Yes, we should merge.

PR #1945, opened 2018-02-05
doc: change input/output contracts for 'current-command-line-arguments'

bennn:
The parameter accepts a vector of string & converts the input to a
vector of immutable strings.

Example:

```
#lang racket/base
(require rackunit)

(define str (make-string 3 #\B)) ;; mutable string
(check-false (immutable? str)) ;; yes of course

(define arg
  (parameterize ([current-command-line-arguments (vector str)]) ;; this is OK
    (vector-ref (current-command-line-arguments) 0)))
(check-true (immutable? arg)) ;; interesting

(check-equal? str arg)
```


bennn:
merged c00340c5f0cd11db0a38f126422685ed32e2fb37

PR #1947, opened 2018-02-06
Copy source location to generate runtime paths.

samth:
Fix found by @lexi-lambda. Bug reported by dbenoit@fedoraproject.org.

### DIFF
--- a/racket/collects/racket/runtime-path.rkt
+++ b/racket/collects/racket/runtime-path.rkt
@@ -186,7 +186,8 @@
                                         (path-of
                                          #,(datum->syntax
                                             #'orig-stx
-                                            `(,#'this-expression-source-file))))
+                                            `(,#'this-expression-source-file)
+                                            #'orig-stx)))
                                     #'void)])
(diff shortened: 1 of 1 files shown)


racket/collects/racket/runtime-path.rkt
```diff
@@ -186,7 +186,8 @@
                                         (path-of
                                          #,(datum->syntax
                                             #'orig-stx
```
lexi-lambda: I guess it doesn’t really matter in this case, since `this-expression-source-file` is a macro and the lexical context on the surrounding syntax object is irrelevant, but it seems slightly suspect to adjust the lexical context here at all. It’d work fine to just use `#'here` for the lexical context and `#'orig-stx` for the source location information.

dbenoit17:
I have tested and can confirm this removes many of the references to build-time rpaths in compiled files.  Thanks so much!

samth:
@mflatt Is this ok to merge?

mflatt:
LGTM

PR #1949, opened 2018-02-07
define-simple-macro now works with an ~or macro.

LeifAndersen:
The previous definition of define-simple-macro would fail (with
a very poor error message) if the user made a simple macro like:

(define-simple-macro (~or a b)
  (let ([tmp a])
    (if tmp tmp b)))

While the define-simple-macro form does allow syntax-parse patterns,
this should not be enabled for the initial head pattern, as described
in the define-simple-macro documentation.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
@@ -20,7 +20,7 @@ Defines a macro named @racket[macro-id]; equivalent to the following:
 @racketblock[
 (define-syntax (macro-id stx)
   (syntax-parse stx
-    [(macro-id . pattern) pattern-directive ... (syntax template)]))
+    [((~var macro-id id) . pattern) pattern-directive ... (syntax template)]))
 ]
 
 @(the-eval '(require syntax/parse/define))
@@ -37,6 +37,8 @@ Defines a macro named @racket[macro-id]; equivalent to the following:
 ((fn2 a b (+ a b)) 3 4)
 (fn2 a #:b 'c)
 ]
+
+@history[#:changed "6.12.0.3" "Change generated pattern head to (~var macro-name id) from macro-name"]
 
 }
 
--- a/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
@@ -38,8 +38,10 @@ Defines a macro named @racket[macro-id]; equivalent to the following:
 (fn2 a #:b 'c)
 ]
 
-@history[#:changed "6.12.0.3" "Change generated pattern head to (~var macro-name id) from macro-name"]
-
+@history[#:changed "6.12.0.3" @elem{Changed pattern head to @racket[(~var macro-id id)] from
+                                    @racket[macro-id], allowing tilde-prefixed identifiers or
+                                    identifiers containing colons to be used as @racket[macro-id]
+                                    without producing a syntax error.}]
 }
 
 @defform[(define-syntax-parser macro-id parse-option ... clause ...+)]{
--- a/pkgs/racket-test/tests/stxparse/test.rkt
+++ b/pkgs/racket-test/tests/stxparse/test.rkt
@@ -810,7 +810,7 @@
     (define-simple-macro (~or . _)
       #:with stx (syntax/loc this-syntax (void))
       stx)
-    (check-eq? (~or) void)
+    (check-eq? (~or) (void))
     ))
 
 ;; from Jay McCarthy (4/2016)
--- a/pkgs/racket-test/tests/stxparse/test.rkt
+++ b/pkgs/racket-test/tests/stxparse/test.rkt
@@ -807,6 +807,10 @@
       #:with stx (syntax/loc this-syntax (void))
       stx)
     (check-eq? (x) (void))
+    (define-simple-macro (~or . _)
+      #:with stx (syntax/loc this-syntax (void))
+      stx)
+    (check-eq? (~or) void)
     ))
 
 ;; from Jay McCarthy (4/2016)
--- a/racket/collects/syntax/parse/define.rkt
+++ b/racket/collects/syntax/parse/define.rkt
@@ -8,11 +8,11 @@
 
 (define-syntax (define-simple-macro stx)
   (syntax-parse stx
-    [(define-simple-macro (~and (macro:id . _) pattern) . body)
+    [(define-simple-macro (macro:id . pattern) . body)
      #`(define-syntax macro
          (syntax-parser/template
           #,((make-syntax-introducer) stx)
-          [pattern . body]))]))
+          [(_ . pattern) . body]))]))
 
 (define-simple-macro (define-syntax-parser macro:id option-or-clause ...)
   (define-syntax macro
--- a/racket/collects/syntax/parse/define.rkt
+++ b/racket/collects/syntax/parse/define.rkt
@@ -12,7 +12,7 @@
      #`(define-syntax macro
          (syntax-parser/template
           #,((make-syntax-introducer) stx)
-          [(_ . pattern) . body]))]))
+          [((~var macro id) . pattern) . body]))]))
 
 (define-simple-macro (define-syntax-parser macro:id option-or-clause ...)
   (define-syntax macro


lexi-lambda:
While I agree that this is, in theory, the better/more obvious behavior, this is a breaking change. It’s also *not* what the current documentation claims, which says this:

> Defines a macro named `macro-id`; equivalent to the following:
> ```
> (define-syntax (macro-id stx)
>   (syntax-parse stx
>     [(macro-id . pattern) pattern-directive ... #'template]))
> ```

This change would need to update the documentation to reflect that the new pattern is `(_ . pattern)`, not `(macro-id . pattern)`. Similarly, there should be a `@history` annotation to document the change so that people writing libraries that should be compatible with older Racket versions aren’t bitten.

That’s assuming that making this breaking change is acceptable. Personally, I think this change walks the line between being too likely to change the behavior of existing programs and unlikely to have been used for anything useful. I’ll defer to @rmculpepper and @AlexKnauth on that.

pkgs/racket-test/tests/stxparse/test.rkt
```diff
@@ -807,6 +807,10 @@
       #:with stx (syntax/loc this-syntax (void))
       stx)
     (check-eq? (x) (void))
+    (define-simple-macro (~or . _)
+      #:with stx (syntax/loc this-syntax (void))
+      stx)
+    (check-eq? (~or) void)
```
AlexKnauth: This should be `(void)` right?

AlexKnauth:
I haven't seen any situations where the macro-id pattern is used intentionally for something where the behavior would be different if the head of the pattern was replaced with `_`.

I've seen a few situations where the current behavior gets in the way, which this pull request would fix.

The most likely situation where this would change behavior is one where it would refer to the original name of the macro in an error message instead of some renamed version
```racket
;; lib.rkt
(provide (rename-out [orig-macro new-name]))
(define-simple-macro (orig-macro x)
  (macro-helper (orig-macro x) x))
;; macro-helper checks that x is a valid argument to orig-macro,
;; and uses its first sub-form for the error message
...

;; use.rkt
(require "lib.rkt")
(new-name invalid-x)
;; current behavior: reports the error as a bad use of `new-name`
;; this PR's behavior: reports the error as a bad use of `orig-macro`
;; both: report the error using the source location from inside the definition of orig-macro,
;;      it would have been better to use `this-syntax`, which is the same for both current and PR
```

Or one where the macro uses `quote` on its name, which results in an actual different run-time value. In most cases this value will be used in either (1) a run-time error message, in which case the `new-name` is better, or (2) a run-time analyzer or interpreter, in which case the `orig-macro` name is probably what they wanted. An example of (1) is:
```racket
;; lib.rkt
(provide (rename-out [orig-macro new-name]))
(define-simple-macro (orig-macro x)
  (if (integer? x)
      (number->string x)
      (error 'orig-macro "expected a string, given ~v" x)))

;; use.rkt
(require "lib.rkt")
(new-name "invalid x")
;; current behavior: reports error in terms of new-name
;; this PR's behavior: reports the error in terms of orig-macro
;; both: report the error using the wrong source location, it would be bettor to use `expr/c`,
;;       which is also consistent between current and PR
```

An example of (2) is:
```racket
;; lib.rkt
(provide orig-macro)      ; not renamed in lib
(define (analyze expr)
  (match expr
    [`(orig-macro ,x) ...]
    ...))
(define-simple-macro (orig-macro x)
  (analyze '(orig-macro x)))

;; use.rkt
(require (only-in "lib.rkt" [orig-macro new-name]))
(new-name x)
;; current behavior: pattern in analyze fails to match because the lib author expected 'orig-macro but got 'new-name
;; this PR's behavior: pattern in analyze succeeds because it was quoted as 'orig-macro
```

rfindler:
The pkg server has a lot of code one could use to get a sense of what kidn of breakage this might cause.

pkgs/racket-test/tests/stxparse/test.rkt
LeifAndersen: Eep, yes, good catch.

LeifAndersen:
@AlexKnauth @lexi-lambda 

Okay, if I replace `_` with either just `macro`, or even `(~var macro id)`, would that be a problem?

LeifAndersen:
Like in the push I just fixed.

Otherwise, I can't really think of a case that would still break it.

lexi-lambda:
The documentation still needs to be updated, and I still think a `@history` annotation is warranted, but this does dispel any backwards-compatibility fears, so I think it’s a good compromise.

LeifAndersen:
@lexi-lambda I'm happy to add a history. But I honestly don't see why the documentation needs to be updated otherwise. As you pointed out, the docs say:

```
(define-syntax (macro-id stx)
  (syntax-parse stx
    [(macro-id . pattern) pattern-directive ... #'template]))
```

But this new behavior is significantly closer to that behavior than what was originally there. So I'm not sure what you want me to change it to. (Unless you are now saying to only add a `@history` note?)

lexi-lambda:
> But this new behavior is significantly closer to that behavior than what was originally there.

No, you’re wrong. The current behavior aligns with the documentation precisely; the behavior in this PR diverges from it. The code in the documentation does not work when `macro-id` is a tilde-prefixed identifier or contains a colon because `macro-id` is used directly within the pattern. The new behavior should be updated accordingly to this:

```racket
(define-syntax (macro-id stx)
  (syntax-parse stx
    [((~var macro-id id) . pattern) pattern-directive ... #'template]))
```

I’m not sure where you’re getting the idea that the documentation is in any way inconsistent with the current behavior.

LeifAndersen:
Huh?

```
(define-simple (~or stx)
  (syntax-parse stx
    [((~var ~or id) a b)
     #'(begin
         (displayln 'a)
         (displayln 'b))]))

(~or 1 2)
```

Gives the error message:
```
~or: keyword used out of context in: (~or 1 2)
```

While the old definition of `define-simple-macro` and:
```
(define-simple-macro (~or a b)
  (begin
    (displayln 'a)
    (displayln 'b)))

(~or 1 2)
```
Gives this error message:

```
b: bad attribute value for syntax template
  attribute value: #f
  expected for attribute: syntax
  sub-value: #f
  expected for sub-value: syntax in: b
```

So if we do want to keep `(~or ...)` from being an acceptable pattern, then we at the very least need to fix that error message.

LeifAndersen:
With that being said, I am happy to change the docs to be `(~var macro-id id)`

lexi-lambda:
According to the existing documentation, this code:

```racket
(define-simple-macro (~or a b)
  (begin
    (displayln 'a)
    (displayln 'b)))

(~or 1 2)
```

…should be equivalent to this code:

```racket
(define-syntax (~or stx)
  (syntax-parse stx
    [(~or . (a b))
     #'(begin
         (displayln 'a)
         (displayln 'b))]))

(~or 1 2)
```

…which indeed it currently is (it produces the same exact error message you describe). So the existing behavior aligns with the current documentation, but the new behavior does not. I’m not sure which part of this you are taking issue with.

LeifAndersen:
Ah, okay, I had a typo...sigh (`define-simple` -> `define-syntax`....woops...yeah for macro related error messages...).

Anyway, fine. In that case I now see how you want me to change the docs. Thanks. :)

pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
```diff
@@ -38,6 +38,8 @@ Defines a macro named @racket[macro-id]; equivalent to the following:
 (fn2 a #:b 'c)
 ]
 
+@history[#:changed "6.12.0.3" "Change generated pattern head to (~var macro-name id) from macro-name"]
```
lexi-lambda: I’d just tweak this note a little bit. You can use Scribble formatting here, so you can use things like `@racket`, and even if you don’t change anything else, `macro-name` should be `macro-id`. Still, I’d probably change this to something like:

```racket
@history[#:changed "6.12.0.3" @elem{Changed pattern head to @racket[(~var macro-id id)] from
                                    @racket[macro-id], allowing tilde-prefixed identifiers or
                                    identifiers containing colons to be used as @racket[macro-id]
                                    without producing a syntax error.}]
```

LeifAndersen: Works for me. pushed.

LeifAndersen: err....pushed that version.r

LeifAndersen:
Alright cool. I'll go ahead and merge it then. Thanks. :)

PR #1951, opened 2018-02-14
Fix source of package catalog server

LiberalArtist:
`(collection-file-path "pkg-catalog" "meta")` raises an exception.


jeapostrophe:
Ya, this is better. Thanks.

PR #1952, opened 2018-02-14
Fallback to `_strdup` to try to fix AppVeyor build

samth:


samth:
This didn't work, but the error message is somewhat inscrutable:

```
Section(foreign)
Command exited with code -1073741819
```

mflatt:
Repair merged. The problem with this patch is that the `_strdup` function accessed dynamically from the C library uses a different allocator than the `free` linked into Racket.

PR #1957, opened 2018-02-20
syntax/transformer: add id-transformer

AlexKnauth:
When writing identifier macros, people tend to "want to" write something like
```racket
(define-syntax my-id-macro
  (lambda (stx)
    (syntax-parse stx
      [_:id (... some arbitrary expression producing syntax ...)])))
```

But when this identifier macro is used in a function application position, they get the error `my-id-macro: expected identifier`, even if `my-id-macro` could produce a function expression.

To fix this, people can do this for all their identifier macros:
```racket
(define-syntax my-id-macro
  (lambda (stx)
    (syntax-parse stx
      [_:id (... some expression ...)]
      [(my-id-macro:id . rst) #`(#,(... some expression ...) . rst)])))
```

The `id-transformer` function gets rid of this pattern and allows them to write this instead:
```racket
(define-syntax my-id-macro
  (id-transformer
    (lambda (stx) (... some expression ...))))
```

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/transformer.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/transformer.scrbl
@@ -37,4 +37,34 @@ op
 @history[#:added "6.3"]{}
 }
 
+@defproc[(id-transformer [id-trans (-> identifier? syntax?)])
+         (-> syntax? syntax?)]{
+
+Creates a transformer for an identifier macro that calls
+@racket[id-trans] on the identifier, whether the macro is
+used as a bare identifier or within parentheses.
+
+@examples[#:eval the-eval
+(code:comment
+ "simple example, could be done with make-variable-like-transformer")
+(define-syntax my-add1
+  (id-transformer
+   (lambda (stx) #'add1)))
+my-add1
+(my-add1 5)
+(code:comment
+ "more complex example, taking advantage of being a transformer")
+(define-syntax whereami
+  (id-transformer
+   (lambda (stx)
+     (unless (even? (syntax-line stx))
+       (raise-syntax-error #f "must be used on an even-numbered line" stx))
+     (syntax-property #'add1 'line (syntax-line stx)))))
+(eval:error whereami)     ; line 11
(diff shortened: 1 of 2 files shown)


jackfirth:
Naming nit: I think `make-id-transformer` is a better name, for consistency with other transformer constructing functions like `make-provide-transformer`, `make-variable-like-transformer`, `make-code-transformer`, etc.

AlexKnauth:
Unlike those, it doesn’t produce a struct for a set-transformer, provide-transformer, etc.

rmculpepper:
The name is uninformative. I recommend extending `make-variable-like-transformer` to take `(or/c syntax? (-> identifier? syntax?))` as its first argument and add an option to suppress the `set!-transformer` wrapper. Maybe even change the handling of a `#f` second argument to omit the wrapper (or if that's too backward-incompatible, make it default to no wrapper if given a procedure for the first argument).

A cpp-like `LINE` macro might be a simpler and more compelling example.

AlexKnauth:
`make-variable-like-transformer` is doing a different "job" than `id-transformer` with respect to what it produces. In `make-variable-like-transformer` it's supposed to always produce an expression, and it even wraps it in `#%expression` to do this. But `id-transformer` is meant to be simpler and general to "all" consistent identifier macros. An id-transformer could expand to a definition, or a module declaration, or another macro identifier that expands to one of those.

In addition, `id-transformer` can be used to modify other types of transformers, like syntax-parse's `pattern-expander`s. Those are actually where most of the cases I've come across where I would want to use this, since `(a b c ~my-transformer d e f)` calls the transformer with the syntax object `(~my-transformer d e f)`, even if I, the programmer, meant `~my-transformer` to be an identifier macro. I as a programmer want something simple that I know wont insert an `#%app` or an `#%expression` like a variable-like transformer would.

rmculpepper:
Can you turn your last answer into documentation and examples? In particular, show how this works with other kinds of expanders. Otherwise it feels like a minor variation on make-variable-like-transformer.

PR #1959, opened 2018-02-21
Update function name

pmatos:
Function reference `supervise-named-dynamic-place-at` should be `supervise-place-at`.


mflatt:
Merged as 80ae3394aad01169cc5d29ea25f7a6e5f245ac0a

PR #1966, opened 2018-02-28
make macro debugger work with new expander

rmculpepper:
This pull request does not update startup.inc.
It relies on the corresponding macro-debugger PR.


mflatt:
This looks good to me, although I'd really prefer a newline after `cond` in "log.rkt".

PR #1993, opened 2018-03-15
Fix function name imap-get-messages

pmatos:
The function is called `imap-get-messages` but the example refers to `imap-get-message`, which does not exist.


stamourv:
Thanks!

PR #2000, opened 2018-03-21
match: check duplicate identifiers across list-no-order patters

### LINKED ISSUE #1964: 'match' signals error on list-no-order + dupl. bindings
This program:

```racket
#lang racket
(match '((0 1 2) (3) (0 4) (5 6 7 8))
  ((list-no-order `(,a ,_) `(,a ,b ,_) _ ...)
    b)
  (else #f))
```

... signals the error 

```
a10: unbound identifier;
 also, no #%top transformer is bound in: a10
```

which simply can't be the right behavior.

AlexKnauth:
The docs for `list-no-order` still need to be changed to say that duplicate identifiers aren't allowed across different sub-patterns.

This pull request fixes #1964 by changing the error message from this:
```
a10: unbound identifier;
 also, no #%top transformer is bound in: a10
```
To this: (where DrRacket highlights the the `a` in the second sub-pattern)
```
. list-no-order: unexpected duplicate identifier in: a
```

The way this pull request does the duplicate identifier check, programs like this get the "unexpected duplicate identifier" error:
```racket
(match (list 1 1)
  [(list-no-order x x)
   x])
```
But programs like this are fine:
```racket
(match (list (list 1 1))
  [(list-no-order (list x x))
   x])
```
```racket
(match (list (list 1 1) (list 2 2))
  [(list-no-order (list x x) (list y y))
   x])
```
Since they don't have duplicates that span *across multiple* sub-patterns.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -155,7 +155,12 @@ In more detail, patterns match as follows:
        #:eval match-eval
        (match '(1 2 3)
          [(list-no-order 3 2 x) x])
-       ]}
+       ]
+
+       Unlike other patterns, @racketidfont{list-no-order} doesn't
+       allow duplicate identifiers between subpatterns. For example
+       the pattern @racket[(list-no-order x 1 x)] produces a syntax
+       error.}
 
  @item{@racket[(#,(racketidfont "list-no-order") _pat ... _lvp)] ---
        generalizes @racketidfont{list-no-order} to allow a pattern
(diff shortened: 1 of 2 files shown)


samth:
This looks good to me, thanks!

AlexKnauth:
I finally added a mention of this in the documentation. With this PR it now says:

> Unlike other patterns, `list-no-order` doesn’t allow duplicate identifiers between subpatterns. For example the pattern `(list-no-order x 1 x)` produces a syntax error.

pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -166,7 +171,12 @@ In more detail, patterns match as follows:
        #:eval match-eval
        (match '(1 2 3 4 5 6)
          [(list-no-order 6 2 y ...) y])
-       ]}
+       ]
+
+       Unlike other patterns, @racketidfont{list-no-order} doesn't
```
bennn: Instead of putting this comment in 2 places, what about putting it in a margin note in between the relevant bullet points?

AlexKnauth: Done.

AlexKnauth: The margin note looks like this:
margin-note

AlexKnauth: Or this with a slightly smaller window size:
margin-note-2

AlexKnauth:
The margin note looks like this:
margin-note

Or this with a slightly smaller window size:
margin-note-2

Does this look good?

PR #2007, opened 2018-03-22
fasl: use longer fasl prefix

bennn:
With the old fasl library, a file could have a comment at the top that explained its purpose --- like how GRacket files start with a comment.

The new fasl format doesn't allow comments like that. This is not a big deal to me, but I do think a fasl file should explain itself enough so if someone finds one of these binary files, they can use google to learn more.

This pull request is one idea for a fix. It makes every fasl file start with `racket/fasl`.


rfindler:
Maybe `read-bytes/exactly` needs to be passed a number that might not be `4`?

rfindler:
test case?

mflatt:
I'm a little sad to add overhead on small values (e.g., 15 bytes instead of 7 to encode a `0` by itself), but this is probably the right thing.

bennn:
Just making sure: is it ok to merge this?

PR #2014, opened 2018-03-24
Fix typo

mromyers:
Incidentally, I think this is the smallest possible PR.
Edit: On second thought, I retract that. The smallest would probably be a single bit changed, while 'o' (6F) to 'i' (69) changes 2 bits.


stamourv:
Thanks!

PR #2030, opened 2018-04-04
Improve the local-expand documentation around the stop list

lexi-lambda:
This PR rewrites a portion of the `local-expand` documentation to clarify and correct some of the details of how the stop list controls `local-expand`’s behavior.

First of all, it replaces `lambda` with `#%plain-lambda` and `#%app` with `#%plain-app` in the list of identifiers that are automatically added to the stop list, since the former refer to `lambda` and `#%app` from `#lang racket/base`, but `lambda` and `#%app` from `'#%kernel` are the identifiers actually added to the stop list. This change is consistent with typesetting conventions used elsewhere in Racket’s documentation.

Second, it clarifies when and how the presence of `#%app`, `#%datum`, and `#%top`’s inclusion in the stop list affects expansion. The existing documentation refers to hyperlinked versions of `#%app` and `#%datum`, implying that the identifiers from `#lang racket/base` are somehow blessed by `local-expand`, but this is not the case. This PR changes the wording to make it clear that the `#%app`, `#%datum`, and `#%top` included in the stop list must match the identifiers that would be implicitly introduced by the expander to prevent the introduction from taking place. It also removes a misleading incorrect comment about the introduction of `#%top`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
@@ -248,44 +248,44 @@ information on the form of the list is below. If @racket[stx] is not
 already a @tech{syntax object}, it is coerced with
 @racket[(datum->syntax #f stx)] before expansion.
 
-When an identifier in @racket[stop-ids] is encountered by the expander
-in a sub-expression, expansions stops for the sub-expression. If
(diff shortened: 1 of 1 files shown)


mflatt:
LGTM - thanks!

PR #2032, opened 2018-04-05
Update the documentation for syntax/loc.

LeifAndersen:
This is to help it match Ryan's changes from a few days back.


mflatt:
I believe the "unless" clause in the current document is supposed to be about when the location is transferred, not when `stx-expr` must produce a syntax object. Those exceptions to transferring a location still apply, I believe, so prbably the right fix is to move the "(which must be a syntax object)" to a separate sentence, and to make the history note that the requirement is enforced even when the target is just a pattern variable.

rmculpepper:
I fixed the conditions under which the loc argument's source location information is transferred to the template to match the previous behavior (that change was a mistake) and I updated the docs and added a history note. Let me know if you think they need more clarification.

LeifAndersen:
This seems to work for me. Thanks.

PR #2037, opened 2018-04-10
 schemify: fix boolean-or-fixnum?

gus-massa:
The first commit is the interesting one, and it fix the ignored `boolean?` test.

The seccond commit is silly, but I've wrote too much assembly when I was young and that `fixnum?` range makes me uneasy. Fell free to ignore it. Anyway, `(expt 2 16)` is too conservative. What about `24` bits? But you have more knowdledge about the landscape of the scheme implementations.

The third one is guaranted by the scheme standard (IIRC). I'm not sure if in that point the `null`s must be quoted or not. So look at it carefuly.

The three commits are very related, so it would be good to squash them before merging.

### DIFF
--- a/racket/src/schemify/equal.rkt
+++ b/racket/src/schemify/equal.rkt
@@ -15,18 +15,19 @@
      (let ([val (unwrap val)])
        (or (symbol? val)
            (keyword? val)
+           (null? val)
            (boolean-or-fixnum? val)))]
     [`,val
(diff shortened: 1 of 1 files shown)


mflatt:
These look good to me, and changing 16 bits to 24 bits also sounds good.

gus-massa:
Merged

PR #2039, opened 2018-04-11
fix dynamic-require with symbol to not make available

michaelballantyne:
to match behavior in docs and racket 6.

This change in behavior is breaking something in one of our Turnstile languages (which is otherwise unwise, but needs improved behavior for syntax properties to fix properly)


mflatt:
Thanks!

PR #2041, opened 2018-04-12
Handle symbol changes between openssl-1.0 and 1.1

hitchiker42:
In openssl-1.1 (specifically libcrypto) the functions sk_num, sk_value and sk_pop_free are prefixed by 'OPENSSL_'.
Now both symbol names looked for to support both version 1.0 and 1.1.

I've tested this with both versions on my machine (x86_64 linux) and it works fine. Previously when linking with openssl-1.1 using any of these functions would cause an error.


jeapostrophe:
I hit the same problem on my OS X machine and tested this patch.

PR #2046, opened 2018-04-19
Fix dynamic-place argument name in description

pmatos:
The description refers to `start-proc`, but argument is `start-name`.


pmatos:
Surprised the tests failed given my change was so simple. Somehow I don't think it's related to my change.

gus-massa:
Thanks. Merged https://github.com/racket/racket/commit/ede4279b14f4029d92d01b734a72e17f812ad41a

PR #2051, opened 2018-04-19
Document and generalize some things for first-class definition contexts

lexi-lambda:
This commit adds a section to the reference to document how the expander tracks information about local bindings, and it extends some syntax-local functions to allow them to accept multiple definition contexts instead of just one. In addition, it improves the documentation on how first-class definition contexts interact with `local-expand`, `syntax-local-value`, and `syntax-local-bind-syntaxes`, and it also clarifies what it means to create a child definition context.


pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -474,7 +478,11 @@ can use bindings previously introduced into @racket[intdef-ctx].
 
 @defproc[(syntax-local-bind-syntaxes [id-list (listof identifier?)]
                                      [expr (or/c syntax? #f)]
-                                     [intdef-ctx internal-definition-context?])
+                                     [intdef-ctx internal-definition-context?]
+                                     [extra-intdef-ctxs (or/c internal-definition-context?
+                                                              (listof internal-definition-context?)
+                                                              #f)
```
mflatt: Is it worth allowing `#f` in addition to `null`? I ave no objection if it seems better to allow `#f` for consistency with other functions that historically accept `#f`, although the "extra"ness of this argument seems different.

pkgs/racket-doc/scribblings/reference/syntax-model.scrbl
```diff
@@ -155,15 +155,12 @@ relevant.
...
-source-location information, @tech{syntax properties}, and
-@tech{tamper status}. In particular, an @tech{identifier} is
-represented as a syntax object that combines a @tech{symbol} with scope sets
-and other information. The @deftech{lexical information} of a
-@tech{syntax object} is its @tech{scope set} combined with the portion
-of the global table of bindings that is relevant to the syntax
-object's set of scopes.
+A @deftech{syntax object} combines a simpler Racket value, such as a symbol or pair, with
+@tech{lexical information}, source-location information, @tech{syntax properties}, and @tech{tamper
+status}. The @deftech{lexical information} of a @tech{syntax object} is comprised of a set of
```
mflatt: Nit: I think "is comprised of" can be just "comprises".

PR #2057, opened 2018-04-27
Add interned scopes and make-interned-syntax-introducer

lexi-lambda:
This adds interned scopes to support Hackett’s type/value namespace capability, as discussed on the slack channel. Interned scopes use symbolic identifiers instead of numeric ones, and if those identifiers are interned symbols, the scope is consistent across both instantiations and serialization to/deserialization from bytecode.


racket/src/expander/syntax/scope.rkt
```diff
@@ -290,6 +294,20 @@
...
+(define interned-scopes-table (make-weak-hasheq))
+
+(define (make-interned-scope sym)
+  (let ([make (λ () (make-ephemeron sym (scope sym 'interned empty-binding-table)))])
+    (or (ephemeron-value
+         (hash-ref! interned-scopes-table sym make))
+        (let ([new (make)])
+          (hash-set! sym new)
+          (ephemeron-value new)))))
+
```
mflatt: Maybe this is going to be harder than I thought. The check for an existing scope and adding it to the table must be atomic, otherwise two different scopes can be returned (with the most recent one added to the table).

racket/src/expander/syntax/scope.rkt
```diff
@@ -306,9 +324,23 @@
...
 (define (scope<? sc1 sc2)
-  ((scope-id sc1) . < . (scope-id sc2)))
+  (let ([id1 (scope-id sc1)]
+        [id2 (scope-id sc2)])
+    (cond [(and (symbol? id1) (symbol? id2))
+           (string<? (symbol->string id1) (symbol->string id2))]
+          [(and (exact-integer? id1) (exact-integer? id2))
+           (< id1 id2)]
+          [else
+           (symbol? id1)])))
```
mflatt: It's probably be better to make a subtype of `scope` to hold an identity for an interned scope, and leave `id` alone. One benefit is that `scope<?` and `scope>?` can stay simple and reliably fast. Another is that the ordering-by-timestamp property of scopes (which is tied to a binding heuristic, which could be important for avoiding space leaks – although that would be a small corner case) is not tangled with the interned-versus-uninterned distinction.

If the symbol `id` is really better, use `symbol<?` instead of `string->symbol`, and consider defining `scope>?` in terms of `eq?` ad `scope<?` (which may depend on making `new-scope-id!` atomic with a `box-cas!`, which is probably a good idea anyway).

Minor (and may go way with other adjustments): please always use a newline after `cond`.

racket/src/expander/syntax/scope.rkt
lexi-lambda: Yes, that makes sense. I thought about that briefly while writing this, but didn’t fully consider the implications. Is the right thing to do here to manually lock the table with a semaphore?

racket/src/expander/syntax/scope.rkt
lexi-lambda: I thought about this, but it seemed harder, and I figured you’d tell me to do it if you thought it was the right thing. :) For a while I was also worried about attaching bindings to interned scopes, since I didn’t fully understand what I was doing, but I don’t think that’s a problem anymore, so I can certainly change this to be a subtype.

racket/src/expander/syntax/scope.rkt
mflatt: No, the expander can't have any locks that could get broken by killing a thread. You may have to resort to atomic mode, like I just did for the other table.

lexi-lambda: Gotcha. I didn’t realize that `call-with-semaphore` would be broken by killing a thread. I guess there’s no way at all to clean up when a thread is killed short of using custodians, since `dynamic-wind` doesn’t invoke its outbound thunks on a killed thread? In any case, I can emulate what you did using atomic mode.

lexi-lambda:
@mflatt I’ve updated this with some changes that I think address your comments. I do wonder a little bit about one thing, though: might it make sense to always make interned scopes’ ids negative, like normal deserialized scopes are? The way things stand, due to the way scopes are interned, the id for an interned scope will always be positive, since the scope will be created and interned, and subsequent uses of the scope will use the same number.

It probably doesn’t matter, since I don’t think bindings can ever be added to an interned scope anyway—new binding forms will always introduce their own non-interned scope, which will always be newer than the interned one—but I could be missing a lot of the details about how it works in practice.

mflatt:
It does sound safer to negate the id. Otherwise, this looks good to me.

lexi-lambda:
Thank you for your help!

PR #2058, opened 2018-04-30
json: fix handling of single-precision irrational numbers; other small fixes

LiberalArtist:
* Fix handling of single-precision infinities and nan.

* Document that non-`hash-eq?` hash tables are accepted by `jsexpr?`.

* Document that the value of `json-null` is recognized using `eq?`.

* Use `case` instead of `assoc`.

* Use contracts.

### DIFF
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -30,12 +30,13 @@ the @rfc for more information about JSON.
   @deftech{JS-Expression}, or @deftech{jsexpr}, is one of:
 
   @itemize[
-    @item{the value of @racket[jsnull], @racket['null] by default}
+    @item{the value of @racket[jsnull], @racket['null] by default,
+  which is recognized using @racket[eq?]}
     @item{@racket[boolean?]}
     @item{@racket[string?]}
-    @item{@racket[(or/c exact-integer? inexact-real?)]}
+    @item{@racket[(or/c exact-integer? (and/c inexact-real? rational?))]}
     @item{@racket[(listof jsexpr?)]}
-    @item{@racket[(and/c hash-eq? (hash/c symbol? jsexpr?))]}]
+    @item{@racket[(hash/c symbol? jsexpr?)]}]
 
 @examples[#:eval ev
   (jsexpr? 'null)
@@ -46,6 +47,7 @@ the @rfc for more information about JSON.
   (jsexpr? #hasheq((turnip . 82)))
   (jsexpr? (vector 1 2 3 4))
   (jsexpr? #hasheq(("turnip" . 82)))
+  (jsexpr? +inf.0)
 ]
 }
 
@@ -55,7 +57,11 @@ the @rfc for more information about JSON.
   In some cases a different value may better fit your needs, therefore
   all functions in this library accept a @racket[#:null] keyword
   argument for the value that is used to represent a JSON ``@tt{null}'',
-  and this argument defaults to @racket[(json-null)].}
+  and this argument defaults to @racket[(json-null)].
+
+  Note that the value of @racket[(json-null)] (or an explicitly-provided
+  @racket[#:null] argument) is recognized using @racket[eq?].
+}
 
 @section{Generating JSON Text from JS-Expressions}
 
--- a/pkgs/racket-test/tests/json/json.rkt
+++ b/pkgs/racket-test/tests/json/json.rkt
@@ -38,6 +38,12 @@
         (not (jsexpr? (/ 1.0 0.0)))
         (not (jsexpr? (/ -1.0 0.0)))
         (not (jsexpr? (/ 0.0 0.0)))
+        (not (jsexpr? +inf.0))
+        (not (jsexpr? -inf.0))
+        (not (jsexpr? +nan.0))
+        (not (jsexpr? +inf.f))
+        (not (jsexpr? -inf.f))
+        (not (jsexpr? +nan.f))
         )
   ;; other `null' values
   (parameterize ([json-null #\null])
--- a/racket/collects/json/main.rkt
+++ b/racket/collects/json/main.rkt
@@ -6,6 +6,13 @@
 ;; edited: Matthias, organization in preparation for pretty-print
 
 ;; -----------------------------------------------------------------------------
+;; DEPENDENCIES
+
+;; racket/contract must come before provide
+(require syntax/readerr
+         racket/contract)
+
+;; -----------------------------------------------------------------------------
 ;; SERVICES
 
 (provide
@@ -14,27 +21,37 @@
  
  ;; Any -> Boolean 
  jsexpr?
-
- #;
- (->* (Output-Port) ([#:null Any][#:encode (U 'control 'all)]))
- ;; #:null (json-null)
- ;; #:encode 'control
- write-json
  
- #;
- (->* (Input-Port) ([#:null Any]))
- ;; #null: (json-null)
- read-json
- 
- jsexpr->string
- jsexpr->bytes
- string->jsexpr
- bytes->jsexpr)
-
-;; -----------------------------------------------------------------------------
-;; DEPENDENCIES
-
-(require syntax/readerr)
+ (contract-out
+  [write-json
+   (->* (any/c) ;; jsexpr? but dependent on #:null arg
+        (output-port? ;; (current-output-port)
+         #:null any/c ;; (json-null)
+         #:encode (or/c 'control 'all)) ;; 'control
+        any)]
+  [read-json
+   (->* (input-port?)
+        (#:null any/c) ;; (json-null)
+        any)] ;; jsexpr?
+  [jsexpr->string
+   (->* (any/c) ;; jsexpr? but dependent on #:null arg
+        (#:null any/c ;; (json-null)
+         #:encode (or/c 'control 'all)) ;; 'control
+        any)] ;; string?
+  [jsexpr->bytes
+   (->* (any/c) ;; jsexpr? but dependent on #:null arg
+        (#:null any/c ;; (json-null)
+         #:encode (or/c 'control 'all)) ;; 'control
+        any)] ;; bytes?
+  [string->jsexpr
+   (->* (string?)
+        (#:null any/c) ;; (json-null)
+        any)] ;; jsexpr?
+  [bytes->jsexpr
+   (->* (bytes?)
+        (#:null any/c) ;; (json-null)
+        any)] ;; jsexpr?
+  ))
 
 ;; -----------------------------------------------------------------------------
 ;; CUSTOMIZATION
@@ -48,7 +65,7 @@
 (define (jsexpr? x #:null [jsnull (json-null)])
   (let loop ([x x])
     (or (exact-integer? x)
-        (real-real? x)
+        (inexact-rational? x)
         (boolean? x)
         (string? x)
         (eq? x jsnull)
@@ -56,8 +73,8 @@
         (and (hash? x) (for/and ([(k v) (in-hash x)])
                          (and (symbol? k) (loop v)))))))
 
-(define (real-real? x) ; not nan or inf
-  (and (inexact-real? x) (not (member x '(+nan.0 +inf.0 -inf.0)))))
+(define (inexact-rational? x) ; not nan or inf
+  (and (inexact-real? x) (rational? x)))
 
 ;; -----------------------------------------------------------------------------
 ;; GENERATION  (from Racket to JSON)
@@ -69,24 +86,28 @@
 (define (write-json* who x o jsnull enc)
   (define (escape m)
     (define ch (string-ref m 0))
-    (define r
-      (assoc ch '([#\backspace . "\\b"] [#\newline . "\\n"] [#\return . "\\r"]
-                                        [#\page . "\\f"] [#\tab . "\\t"]
-                                        [#\\ . "\\\\"] [#\" . "\\\""])))
-    (define (u-esc n)
-      (define str (number->string n 16))
-      (define pad (case (string-length str)
-                    [(1) "000"] [(2) "00"] [(3) "0"] [else ""]))
-      (string-append "\\u" pad str))
-    (if r
-        (cdr r)
-        (let ([n (char->integer ch)])
-          (if (n . < . #x10000)
-              (u-esc n)
-              ;; use the (utf-16 surrogate pair) double \u-encoding
-              (let ([n (- n #x10000)])
-                (string-append (u-esc (+ #xD800 (arithmetic-shift n -10)))
-                               (u-esc (+ #xDC00 (bitwise-and n #x3FF)))))))))
+    (case ch
+      [(#\backspace) "\\b"]
+      [(#\newline) "\\n"]
+      [(#\return) "\\r"]
+      [(#\page) "\\f"]
+      [(#\tab) "\\t"]
+      [(#\\) "\\\\"]
+      [(#\") "\\\""]
+      [else 
+       (define (u-esc n)
+         (define str (number->string n 16))
+         (define pad (case (string-length str)
+                       [(1) "000"] [(2) "00"] [(3) "0"] [else ""]))
+         (string-append "\\u" pad str))
+       (define n
+         (char->integer ch))
+       (if (n . < . #x10000)
+           (u-esc n)
+           ;; use the (utf-16 surrogate pair) double \u-encoding
+           (let ([n (- n #x10000)])
+             (string-append (u-esc (+ #xD800 (arithmetic-shift n -10)))
+                            (u-esc (+ #xDC00 (bitwise-and n #x3FF))))))]))
   (define rx-to-encode
     (case enc
       ;; FIXME: This should also encode (always) anything that is represented
@@ -101,7 +122,7 @@
     (write-string (regexp-replace* rx-to-encode str escape) o)
     (write-bytes #"\"" o))
   (let loop ([x x])
-    (cond [(or (exact-integer? x) (real-real? x)) (write x o)]
+    (cond [(or (exact-integer? x) (inexact-rational? x)) (write x o)]
           [(eq? x #f)     (write-bytes #"false" o)]
           [(eq? x #t)     (write-bytes #"true" o)]
           [(eq? x jsnull) (write-bytes #"null" o)]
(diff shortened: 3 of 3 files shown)


LiberalArtist:
As far as I can tell, the AppVeyor build failure has something to do with `racket\raco.exe test -l tests/net/url`: it doesn't seem to be related to this change.

samth:
That was another error in the windows build, which is now fixed. I'm rerunning the build.

LiberalArtist:
@mflatt @mfelleisen I wanted to follow up about whether someone could review this in time for Racket 7.

samth:
Most of this looks good. My assumption is that the library originally did not use actual contracts there for some reason, and so we should figure out if that's the case prior to changing it.

LiberalArtist:
I don't know if this is the reason, but one complication of using contracts (as I noted in the comments) is that using `jsexpr?` as a contract would be incorrect in the presence of a `#:null` argument. Since the implementation already does that checking, I opted to use `any/c` rather than `->i` to avoid any potential performance penalty.

I think the contracts still have value, even with that limitation. I noticed that `read-json` and `write-json` don't check that their output port argument is actually an `output-port?`, so if we remove the contracts, or remove them for now, we should add those manual checks.

mfelleisen:
Eli wrote the initial version and I fixed a couple of things before. 

1. I suspect contracts were not a part of Eli's interests and we all think a lot in terms of performance. 
2. If we use contracts, I think we should use ->i not ->* to make them precise. But good enough for now. 
3. I would have preferred assoc  over case just in case someone comes up with the idea of adding another element to this table. BUT, the table should be lifted to the top as a constant definition. 

I am not an expert in things JSON and so I must simply trust the numeric aspect of this commit.

samth:
`->i` contracts are substantially more expensive in many cases, so we probably want to avoid them here.

There are several reasons why @elibarzilay may have avoided contracts here:

- performance of the actual functions (might be worth a microbenchmark to see what the impact is)
- wanting to use `json` somewhere that using `racket/contract` is not appropriate, such as where the additional loading time is problematic, or where fewer dependencies are needed for some other reason -- see the expander for an example.

But maybe he can answer for sure.

mfelleisen:
My preference would be to write a 

  `json-with-contracts.rkt`

module that imports from the 'raw' module.

rfindler:
In this case, I would lean towards `json` and `json/unsafe` as the name of the two libraries (one with contracts and whatever else checked and the other without). That's roughly the convention we've followed elsewhere.

mfelleisen:
Let’s do that.

rmculpepper:
Please, let's not overload the word "unsafe" with yet another meaning. ("Unsafe" usually means "this feature can violate Racket's invariants and make the process crash", but IIRC it has also been used to mean "this operation makes a future stop executing in parallel".) How about "unchecked" or "no-contracts"?

samth:
I agree with Ryan in general, but I think we should not make two versions
of this library -- that will just make everything more confusing for users.
We should figure out why those contracts are in comments currently, and if
those reasons no longer apply, we should use contracts as the patch already
does.

rfindler:
I don't agree with this definition of unsafe. Unsafe means something closer to what undefined behavior in the C++ standard means, roughly "we aren't checking things here, and if you violated the unchecked invariants, all bets are off". I think this captures precisely what `racket/unsafe/ops`'s `unsafe-fx+` does when you do things like this (I may be wrong, but I believe this is guaranteed NOT to crash):

```
> (unsafe-fx+ #\a 1)
70365521396457
```

and it also captures what happens when you turn off contract checking. You might get a violation of some other library that's used internally, or you might get a bogus result. For example, here's how you get a violation of some other library that's used internally:

```
> (unsafe-fl+ #\a 1)
Segmentation fault: 11
```

This is no "new meaning" and no change to the meaning. This is the meaning it has had all along.

rmculpepper:
It is worth distinguishing an ordinary module's invariants from *Racket's* invariants, and we should reserve "unsafe" to talk things that can violate the latter. Unless I'm missing something, the unchecked version of the json library would not be capable of violating Racket's invariants, and claiming that if you misuse it then "all bets are off" is hyperbole. Some bets are off, but not all.

I also agree with Sam: we should be sure there is a good reason before complicating things.

rfindler:
Ryan: you write "It is worth distinguishing an ordinary module's invariants from Racket's invariants," but where exactly do you draw the line? Is racket/contract in the blessed portion? syntax/parse? #%kernel (which is now in a different set of modules that it used to be).

As for "all bets are off", I don't think it is hyperbole. Ultimately what people care about it how their apps work (or don't work). If we return `70365521396457` because there was an unsafe-fx+ that wasn't correctly used, or because we didn't check some invariant at the json library library and some conditional went the wrong way in a client of json due to a bad result it produced. The rest is internal details and which set of internal details went wrong is useful for debugging and fixing problems, but that's not a concern of the actual users of the system.

As for Matthias's comment, I don't agree either. If you remove a (racket/contract) contract check from a function and it produces a bad value, well, the contract might have been `fixnum?` and the result might now flow into `unsafe-fl+` and then we get a segfault, but `unsafe-fl+` was completely and totally fine and it shouldn't be changed. It was the other library that wrong, not the one with `unsafe-fl+` in it.

bennn:
Please don't add `json/unsafe` here.

In general I agree that we should reserve "unsafe" for things where "the entire system can crash or become corrupted" in addition to giving the wrong result. That's an important distinction to me (though it's not exactly the ordinary module vs. Racket distinction).

I might have some json-parsing code that I can dig up to help benchmark this.

rfindler:
@bennn I fail to understand how "the entire system can crash or become corrupted" cannot happen with anything where some check is not done and the result of some computation is bogus. I hesitate to trot out the old pacemaker example, but certainly we don't want that to produce incorrect heart rate recommendations / electrical pulses, and that'll happen with any missing check (when that check would have resulted in "the entire system" sending a text message to your doctor or whatever).

I mean, if there is an error in DrRacket, you'll get the annoying message box saying "Internal error", possibly thousands of times all in row. How is are those thousands of dialogs better than getting the usual segfault dialog?  Seems worse to me and, indend, my response would be to drop my shell and forcibly terminate the process, which is what a segfault also does, so clearly a segfault was preferred in that case. That's for "crash". For "corrupted": which is worse, DrRacket silently not telling your about some corrpution in the internals of the text% data structures that leads to every time you save it being garbage on the disk (or, even worse, it starts picking other files to overwrite on each save), or the GC having a bug which corrupts memory which leads to much much more visible errors?

This "low level" nonsense is makes sense only if you think that the runtime of Racket is somehow special  compared to the rest of the language and that's an entirely un-LOP, un-Rackety line of thought.

It is abstractions all the way down. Why single out one in this way? We should single them out when they have lots of engineering effort put into them because they are likely better and whatever. And clearly we want to put lots of engineering effort into our GC and our runtime system, but we should also put the macro system in this exalted class. For errors there can be catastrophic. But y'alls definitions excludes it. Why?

mfelleisen:
Here is an error that a lack of programmer-defined contracts cannot generate: 

	add the bits of a function pointer to an integer and used the result as an integer 

With “error” I mean “behavior that fails to live up to the intended level of abstraction”. 
I do not mean “signals a contract failure” or any failure for that matter. 

I think the rest of us believe that this kind of error is somehow worse than the others. 

This yields two levels of errors. Now I think from here on up, there are more layers:

 — safety contracts 
 — programmer contracts 
 — programmer types 
 
etc. As I said in my first response, there’s research to be done. The PL world is 
all wrong about soundness. It’s a spectrum not a binary thing. 

Now in practice Racket has conflated the layers, and I am sad about it. But 
there’s nothing I can do.

rmculpepper:
@rfindler Re drawing the line: The invariants I'm talking about belong to the Racket VM. Ideally, "the line" coincides with things provided with `protect-out`. When that fails to be true, I would consider it a bug. More specifically:
- The `#%kernel` module does not (directly) export any operations capable of violating them; the `#%unsafe` and `#%foreign` modules do, and these primitive modules provide their unsafe operations with the protected flag set. 
- If a non-primitive module has access to unsafe features, it *should* either (1) only use them in safe ways, or (2) provide its unsafe derived features using `protect-out`. An example of option 1 is `match`: it might generate a reference to `unsafe-car`, but only in positions guarded by an earlier pair check. An example of option 2 is `ffi/unsafe`: it requires `#%foreign` and builds higher-level (but still unsafe) features on top of it, and consequently it protects most of its exports. A module that provides, without protection, an export that can be used to violate the Racket VM's invariants is buggy (or untrustworthy, depending on your perspective).

Re values flowing to unsafe operations: I think the standard blame story applies. If module M uses `json/unchecked` and gets an unexpected value and passes it to an unsafe operation, it's M's fault.

Re why "unsafe" is important when there are other bad user experiences too: The point is that in the *absence* of unsafe features, if something bad happens, you can rely on the rules of the language to narrow the source of the problem (and possibly recover). Yes, a bad value might flow along a complicated path before it causes problem, but at least you can still rely on the notion of "value". If `unchecked-jsexpr->json` returns a bad value, I can catch it by calling `string?`. If `unsafe-jsexpr->json` misbehaves, I can't be sure of *anything at all*.

rfindler:
But, @rmculpepper , why are the equations provided by the raw language the only important ones? My program might rely on equations like `(from-nat natural/e 3)` = `3` and I might be reasoning about that when I modify / write my program.

Are you saying that it is reasonable for users of some libraries to be forced to read the bodies of the functions they use and then reason in terms of the definitions of those functions, but not other libraries? Why do we have only ONE important layer of libraries/implementations?

(I'm responding only to the latter portion of your message. The first part I like a lot more but need to think more about it -- my instinct is to wonder why this mechanism applies only to some operations that may be abused, but not other operations that may be abused.)

rmculpepper:
@rfindler I wouldn't say they're the only important ones, but they are by far the most important ones, and the raw language is the basis for all other protection/isolation mechanisms (eg contracts and everything in the Revenge of the Son of the Lisp Machine).

I think the *mechanism* (protected exports) does scale in a way; IIUC `protect-out` means "only modules whose declaration-time code inspector is at least as powerful as mine can use these exports". The Racket VM unsafe/safe boundary is the only really coherent one I know of, though. DrRacket is close, but as I understand it, DrRacket enforces a weaker boundary between itself and user programs (it doesn't prohibit user code from using operations), and it uses different mechanisms.

rfindler:
As for the first paragraph you write, @rmculpepper , I agree that there are some central abstractions that the lowest-level language provides that are the basis for other mechanisms and that might form the basis for a predicate to include some operations and exclude others from a specific set.

The word "unsafe" does not have this connotation to me (and I see the definition of it is "dangerous", which is more about abstraction breaking than "important abstraction" breaking). I suppose we could come up with a small set of libraries that meet this bar and then have two words for this same concept, saving one for these important invariants and only allowing the lesser libraries a lesser word. (That feels kind of sad to me, tho, as the importance seems orthogonal -- as I've argued a lot in this thread, ultimately the app can have catastrophic consequences for any of these, not just these most important ones.)

As for your second paragraph, I disagree completely, perhaps even more than with the disagreement that started this discussion :). In particular I think there are MANY libraries that provide new forms of data and new operations and these operations come with invariants that are coherent. I even gave you an example of one that's not used very much (although it is used fairly centrally in redex's test case generator), namely `data/enumerate`. I would say that all of the libraries you wrote in the `data/` collection come with just as coherent a layer of abstraction as anything else in Racketdom, too. There are many of them (and, to get back to LOP, if we didn't have the ability to make many many more of them, LOP would be doomed).

lexi-lambda:
While I feel like the discussion here is interesting and valuable in general (I tend to agree with bits and pieces of the arguments from all parties), I question how much it is relevant to this particular pull request. Racket strives to be a safe language; we already compromise performance for safety in various places *without* unsafe “escape hatches”.

If using the contract system in our *JSON parsing library*, as has been done in this pull request, warrants providing an unsafe API because the contracts might be prohibitively expensive, then contracts have failed, and the contract system should include a way to arbitrarily (and unsafely) disable contract checking. I do not want a proliferation of `/unsafe` modules that simply disable contracts.

I second @bennn’s call for benchmarking. If the contracts are cheap, this discussion is irrelevant, and we should just merge the changeset with the contracts. If we think `->i` is really the right thing here *and* that contract is slow, I’d still be opposed to adding a `json/unsafe`, which just punts the problem to users instead of figuring out how to provide a safe API with decent performance (with or without the contract system). It would make me sad to give up our principles on something as trivial as JSON parsing.

rfindler:
Nice point about there being a language design problem in here, @lexi-lambda ! Woudln't it be cool if we could write the best, safest library we know how to and just get one that's faster mostly for free by automatically unsafitizing it?

bennn:
@LiberalArtist here are the json files I had on my computer. Unzip this & run `main.rkt` to see how long it takes to parse them.

[some-json.tar.gz](https://github.com/racket/racket/files/2070166/some-json.tar.gz)

I can send a longer version of `house.json` if you think it'd help, but I think we'd have more luck searching online for existing json-parsing benchmark suites. Let me know if you want help with this.

LiberalArtist:
Here are my results from a few runs, where "unchecked" is the current version and "contracts" is as modified by the pull request. Since `json` is self-contained, I copied both versions so that they could run in otherwise the exact same Racket 6.12. I put the exact code I ran in https://github.com/LiberalArtist/json-pr-2058-benchmark, though I've grouped the output across runs below.
```
python.json
-----------
Unchecked
cpu time: 42 real time: 42 gc time: 18
cpu time: 41 real time: 40 gc time: 17
cpu time: 42 real time: 42 gc time: 18
cpu time: 41 real time: 41 gc time: 18
cpu time: 40 real time: 40 gc time: 18
Contracts
cpu time: 25 real time: 25 gc time: 0
cpu time: 23 real time: 24 gc time: 0
cpu time: 24 real time: 24 gc time: 0
cpu time: 24 real time: 24 gc time: 0
cpu time: 24 real time: 24 gc time: 0
------------------------------------------
house.json
----------
Unchecked
cpu time: 1308 real time: 1311 gc time: 39
cpu time: 1334 real time: 1337 gc time: 40
cpu time: 1316 real time: 1320 gc time: 42
cpu time: 1310 real time: 1315 gc time: 39
cpu time: 1278 real time: 1281 gc time: 39
Contracts
cpu time: 1298 real time: 1301 gc time: 2
cpu time: 1321 real time: 1324 gc time: 3
cpu time: 1278 real time: 1281 gc time: 4
cpu time: 1280 real time: 1283 gc time: 4
cpu time: 1255 real time: 1259 gc time: 1
```

I don't know why the "contracts" version seems to be slightly faster, and I assume it means this isn't actually measuring the cost of contracts. (Maybe `case` could be faster than `assoc` or something, though that wasn't my motivation for the change.) 

On the other hand, it does seem to suggest that contracts are not prohibitively expensive.

elibarzilay:
Re the origins of the code: IIRC, I actually started from @dherman's code (I think that it was on planet), and I *think* that I rewrote large parts of it, but kept it roughly the same, with the exception of the null handling.  (Again, this is all "IIRC", it's been a while.)  I don't think that I considered adding contracts, but it makes sense that I wouldn't do that for those reasons, especially since it's not uncommon to handle large amounts of data.

mfelleisen:
Wait.

samth:
@mfelleisen, given the upcoming release schedule, do you have a timeline for how long we should wait here? The benchmarks posted suggest that we shouldn't worry about the cost of the contracts added in this PR.

mfelleisen:
Let’s merge. 

[[ @samth ~~ I wrote down contracts because I didn’t have tests and wanted to recover knowledge from untyped code. I wrote down lightweight contract because I am intensely aware of the cost. 

@LiberalArtist ~~ I think the examples/runs are too small to measure the contract overhead. But yes this is weird. But I am disappointed that Python is so much faster. ]]

LiberalArtist:
@mfelleisen Just to be clear, "python.json" is just a data file: I didn't test against a Python JSON parser. You'd have to ask @bennn about the name, but it looks like it might be data _about_ a Python program.

bennn:
Yep, `python.json` is data about python programs. Zeina and I used it to make Figure 4 [here](http://www.ccs.neu.edu/home/types/resources/pdf/gm-pepm-2018.pdf) :)

@stamourv is there any hope of cherry-picking this onto the release branch?

bennn:
I guess not. Testing started today.

elibarzilay:
[

Sidenote: I missed the fact that this is also changing `case` to
`assoc`.  FWIW, I think that that change was bad: one reason is the one
that Matthias raised (in general, `assoc` means that a bit of
functionality is encoded in a value that is easier to move around).  But
another reason that is more important to me is that it makes code that
is slightly longer in general --- and when combined with aversion to
writing multiple cases on the same line, that "slightly" turns to
"much".  I think that the style of overly wasting vertical space (adding
empty lines "for clarity" (with an extreme that I've seen once: some
code where more than 50% of the lines were followed by an empty lines),
insisting on each element of a sequence of short things be on its own
line (as happened here), dignifying each brace with its own private line
(in *those* languages), etc) --- this style is a plague against making
code more readable.  I keep being surprised that there's so little
awareness to the negative effects of such styles.  (Combined in some
cases with "stylecop" and other automated style checkers that makes
things worse.)

In any case, back to the "FWIW" part, the original use of `assoc`
together with the multiple-cases-on-one-line[*] was very intentional.

[*] Which later got mangeled into the badly-indented result of DrR's
indenter (or maybe pretty-print is the one that leads to the broken
indentation in DrR).

]

LiberalArtist:
I suspect this basically comes down to taste, but here was my thinking for the record.

When I first read this code and tried to figure out what was going on, the use of `assoc` was one of the things that made it less obvious. It certainly wasn't a big issue, but it seemed worth trying to improve for the next person to read the file. This pattern of using `assoc` to approximate `(dict-ref lst k #f)` is certainly common (I've done it myself), but the problem is it isn't the *only* way of using assoc. In particular, I initially thought the code might have used both parts of the returned pair, not just the `cdr`.

The use in `read-json*` was less objectionable and I wouldn't have changed it for that alone, though even there I took a second to convince myself that there was an implicit `pair?` check before the `cdr`. In `write-json*`, though, it coincided with nested uses of `if` and the need for local helper definitions—`n` and `u-esc`, which were needed only if the key wasn't in the table, and `r`, which was needed only to make `assoc` work like `dict-ref`—to make the logical intent of the control flow less obvious to me. I could have addressed most of these concerns by using `assoc` with `cond`, but I thought `case` was even clearer.

The multiple-cases-on-one-line aspect I don't have strong feelings about one way or the other.

samth:
The new contract for `read-json` is too restrictive, breaking existing code. See https://plt.eecs.northwestern.edu/pkg-build/server/built/test-fail/parsack.txt for an example.

mfelleisen:
I see the problem. I will take a look at it and get it fixed for the release.

lexi-lambda:
I think the contract just needs to be corrected from `(->* (input-port?) (#:null any/c) any)` to `(->* () (input-port? #:null any/c) any)`.

mfelleisen:
Yes, and I will add a test and do the cleanup after lunch.

mfelleisen:
@ stchang ~~ my fix to JSON should resolve the problem with your parser package.

PR #2059, opened 2018-05-01
A few small fixes.

samth:


mflatt:
LGTM

samth:
@mflatt I just added another commit here -- can you take a look?

mflatt:
I'm less enthusiastic about that one. Would it make sense to move `read-eval-print-loop` to its own module, so you can `#%require` it in a module that starts with `#%kernel`?

samth:
Isn't that what the change does? I think using `delete-trailing-whitespace` in emacs produced too big a diff -- I'll fix that.

samth:
Ok, thanks. I reworded the commit message to avoid the misunderstanding, and I'll merge when CI completes.

samth:
The AppVeyor failure is something else, and the Travis failure is something transient with weak tables.

PR #2064, opened 2018-05-03
Use `racket/private/repl` now that it was added in #2059.

samth:


mflatt:
Seems ok.

I wonder whether `racket/private/repl` should be made public, so that the documentation can be clear about what module it `dynamic-require`s to get `racket/private/repl`.

I also wonder whether GRacket should `(dynamic-require 'racket/gui/base ....)` instead of `'racket/gui/init`. The difference matters mainly when `-I` is used.

PR #2070, opened 2018-05-11
fix: allow missing fields in `object/c-common-fields-stronger?`

bennn:
Yesterday the Typed Racket unit tests failed with this message:

```
contract (object/c-opaque (h (->m (-> boolean? boolean? boolean?) integer?)) (field (x any/c))) is unexpectedly weaker than (object/c)
```

This PR fixes that failure by changing `object/c-common-fields-stronger?` to only compare the fields that are in both objects (instead of returning `#false` if the second object doesn't have a field from the first).

I think this is the right thing ... `object/c-stronger` is already checking `object/c-width-subtype` to guard against missing fields.

But, a different fix would be to implement this directly in Typed Racket.

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/stronger.rkt
+++ b/pkgs/racket-test/tests/racket/contract/stronger.rkt
@@ -454,6 +454,9 @@
          (object/c (field (f (<=/c 4))))
          (object/c (field (f (<=/c 4)))))
   (ctest #t contract-stronger?
+         (object/c (field (f (<=/c 4))))
+         (object/c))
+  (ctest #t contract-stronger?
          (object/c (m (-> any/c (<=/c 3)))
                    (n (-> any/c any/c)))
          (object/c (m (-> any/c (<=/c 4)))))
--- a/racket/collects/racket/private/class-c-old.rkt
+++ b/racket/collects/racket/private/class-c-old.rkt
@@ -1641,16 +1641,17 @@
                (names-sel this))
        (for/and ([this-name (in-list (names-sel this))]
                  [this-ctc (in-list (ctcs-sel this))])
-         (for/or ([that-name (in-list (names-sel that))]
-                  [that-ctc (in-list (ctcs-sel that))])
-           (and (equal? this-name that-name)
-                (contract-equivalent?
-                 (if (just-check-existence? this-ctc)
-                     any/c
-                     this-ctc)
-                 (if (just-check-existence? that-ctc)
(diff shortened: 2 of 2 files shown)


samth:
Which change broke this part of TR?

bennn:
Here's the change:
https://github.com/racket/racket/commit/8ec3edaa9597223f152f5ca12799c8c669e471bb

(Thanks Sam, I meant to put this commit in the original message)

rfindler:
Thanks!

If you want to push, please add this test case:

```
--- a/pkgs/racket-test/tests/racket/contract/stronger.rkt
+++ b/pkgs/racket-test/tests/racket/contract/stronger.rkt
@@ -453,6 +453,9 @@
   (ctest #t contract-stronger?
          (object/c (field (f (<=/c 4))))
          (object/c (field (f (<=/c 4)))))
+  (ctest #t contract-stronger?
+         (object/c (field (f (<=/c 4))))
+         (object/c))
   (ctest #t contract-stronger?
          (object/c (m (-> any/c (<=/c 3)))
                    (n (-> any/c any/c)))
```

alternatively, I can do it.

bennn:
Added the test & pushed.

I'm not sure why travis is timing out, but this command passes on my machine:
```
racket -l tests/racket/contract/all
```

bennn:
Ok then maybe this isn't the right change --- because if changes _why_ equivalent? will return false in some cases.

Question: there are 2 functions `check-one-object` and `check-one-object/equivalent`. I think the difference between these is that one compares with `contract-stronger?` and the other compares with `contract-equivalent?`. Is that true?

- if it's true, then this change is ok
- if it's not, and `check-one-object/equivalent` is not supposed to only check the names in both, then I'll revert this and change the implementation of `common-fields-stronger`

(I figured the answer to the question above was "yes" because `object/c-equivalent` checks that its arguments have `equal?` method and field names)

bennn:
> I think the problem is that object/c-common-fields-stronger? should not just call check-one-object/equivalent

Maybe --- but this depends on the purpose statement of `check-one-object/equivalent`. If its purpose is to compare the common names (just like `check-one-object` compares common names), then its ok.

(The `(names-sel this)` does look redundant. `class/c` looks okay to me though.)

rfindler:
Well, it is being used in contract-equivalent. That's two different purposes statements, and yet only one function, which seems unlikely to work. Unless I'm missing something?

PR #2071, opened 2018-05-11
Make stx.rkt cross-phase persistent.

samth:
This eliminates more than 70 linklet instance creations.

I'm not sure if this is a good idea. Linklet instance creation
is measured in pycket, which is the equivalent of `-c`.

### DIFF
--- a/racket/collects/racket/private/cond.rkt
+++ b/racket/collects/racket/private/cond.rkt
@@ -3,7 +3,7 @@
 ;; cond
 
 (module cond '#%kernel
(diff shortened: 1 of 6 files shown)


samth:
Actually looking at memory use after loading the repl with `-c` and `-I racket/base`, this seems to save about 40k (after repeated GC).

mflatt:
Seems ok. I think "gen-temp.rkt" is missing from the PR (but it was obvious to reconstruct to try out the patch).

samth:
This doesn't feel like quite enough to be worth doing yet, but I want to explore more in this direction.

PR #2076, opened 2018-05-13
Extended documentation for racket/trace for syntax transformers

wilbowma:
Improved the existing explanation of using `racket/trace` to debug syntax transformers, and added an example of combining `racket/trace` with `debug-scopes` to trace scopes.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/trace.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/trace.scrbl
@@ -5,6 +5,10 @@
         (ev '(require racket/trace))
         (ev '(require (for-syntax racket/base))))
 
+@(begin (define ev1 (make-base-eval))
+        (ev1 '(require racket/trace))
+        (ev1 '(require (for-syntax racket/base))))
+
 @title{Tracing}
 
 @note-lib-only[racket/trace]
@@ -47,6 +51,26 @@ The result of a @racket[trace] expression is @|void-const|.
 (f 10)
 ]
 
+@racket[trace] can also be used to debug @tech{syntax transformers}.
+This is verbose to do directly with @racket[trace]; refer to @racket[trace-define-syntax] for a
+simpler way to do this.
+
+@examples[#:eval ev
+(require (for-syntax racket/trace))
+(begin-for-syntax
+  (define _let
+    (syntax-rules ()
+      [(_ ([x v]) e) ((lambda (x) e) v)]))
+  (trace _let))
+(define-syntax let _let)
+
+(let ([x 120]) x)
+]
+
+When tracing syntax transformers, it may be helpful to modify @racket[current-trace-print-args] and
+@racket[current-trace-print-result] to make the trace output more readable; see
+@racket[current-trace-print-args] for an extended example.
+
 }
 
 @defform*[((trace-define id expr)
@@ -69,7 +93,7 @@ function then tracing it. This form supports all @racket[define] forms.
            (trace-define-syntax (head args) body ...+))]{
 
 The @racket[trace-define-syntax] form is short-hand for first defining a
-macro then tracing it. This form supports all @racket[define-syntax] forms.
+syntax transformer then tracing it. This form supports all @racket[define-syntax] forms.
 
 For example:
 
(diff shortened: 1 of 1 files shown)


wilbowma:
Hm. I can't figure out why these failed.

wilbowma:
After forcing a rerun (without changing any of my PR), most of travis passes, but one fails with `/Users/travis/.travis/job_stages: line 166: shell_session_update: command not found` which doesn't seem to be my fault. The appveyor shows all tests pass, but `Command exited with code -1073740940`; not sure what that means.

bennn:
travis passes, 100%

pkgs/racket-doc/scribblings/reference/trace.scrbl
```diff
@@ -165,6 +175,104 @@ traced call. It receives the name of the function, the function's
...
+  (trace-define-syntax fact
+    (syntax-rules ()
+      [(_ x) 120]))
+  (fact 5)]
+
+In this example, we update the printers by storing the current printers, @racket[ctpa] and
+@racket[ctpr], which we call after converting the arguments @racket[l] and results @racket[r] to
+datum.
+
+Note that this modification of @racket[current-trace-print-args] and
```
bennn: Why not use `maybe-syntax->datum` in the first example, and remove the second codeblock?

(I think the note about the imperative update is useful)

pkgs/racket-doc/scribblings/reference/trace.scrbl
```diff
@@ -165,6 +175,104 @@ traced call. It receives the name of the function, the function's
...
+      [(_ x:nat) (precompute-fact syn (syntax->datum #'x))]
+      [(_ x) #'(run-time-fact x)]))
+  (fact 5)
+  (fact (+ 2 3))]}
+
+Finally, you can modify these parameters to print additional information about syntax objects.
+For example, we can use @racketmodname[debug-scopes] to trace how scopes change through syntax
+transformers.
+
+@; manually typeset to avoid depending on debug-scopes
```
bennn: Hmm, I think it would be better to add this example to the `debug-scopes` docs and just point to it from here --- so the example doesn't need to be manually typeset

pkgs/racket-doc/scribblings/reference/trace.scrbl
```diff
@@ -81,25 +105,11 @@ For example:
...
-
-  (trace-define-syntax fact
-  (syntax-rules ()
-    [(_ x) #'120]))
-  (fact 5)]}
+syntax transformer. This can result in too much output if you do not need to see,
+e.g., source information.
+To get more readable output by printing syntax objects as datums, we can modify the
+@racket[current-trace-print-args] and @racket[current-trace-print-results].
+See those @racket[current-trace-print-args] for an example.
```
bennn: typo: ~those~

pkgs/racket-doc/scribblings/reference/trace.scrbl
```diff
@@ -47,6 +51,26 @@ The result of a @racket[trace] expression is @|void-const|.
...
+    (syntax-rules ()
+      [(_ ([x v]) e) ((lambda (x) e) v)]))
+  (trace _let))
+(define-syntax let _let)
+
+(let ([x 120]) x)
+]
+
+When tracing syntax transformers, it may be helpful to modify @racket[current-trace-print-args] and
+@racket[current-trace-print-result] to make the trace output more readable; see
```
bennn: should be `current-trace-print-results`

pkgs/racket-doc/scribblings/reference/trace.scrbl
```diff
@@ -165,6 +175,104 @@ traced call. It receives the name of the function, the function's
 ordinary arguments, its keywords, the values of the keywords, and a
 number indicating the depth of the call.
 
+Modifying this is useful to get more readable or additional output when tracing syntax transformers.
+For example, we may only care about the shape of the syntax object and not the source location
+information.
+By modifying @racket[current-trace-print-args], we can print the argument to (and, using
+@racket[current-trace-print-result], the result of) the syntax transformer as a datum using
```
bennn: should be `current-trace-print-results`

pkgs/racket-doc/scribblings/reference/trace.scrbl
```diff
@@ -165,6 +175,104 @@ traced call. It receives the name of the function, the function's
...
+    (syntax-rules ()
+      [(_ x) 120]))
+  (fact 5)]
+
+In this example, we update the printers by storing the current printers, @racket[ctpa] and
+@racket[ctpr], which we call after converting the arguments @racket[l] and results @racket[r] to
+datum.
+
+Note that this modification of @racket[current-trace-print-args] and
+@racket[current-trace-print-result] is an imperative update, and will affect all traced identifiers.
```
bennn: should be `current-trace-print-results`

pkgs/racket-doc/scribblings/reference/trace.scrbl
wilbowma: I think it's valuable to see the failure to understand how one should transform `l`.

pkgs/racket-doc/scribblings/reference/trace.scrbl
wilbowma: That's not a bad idea. I'll see if I can get this snippet merged into that package first, then update this PR.

wilbowma:
The debug-scopes example has been merged into that package, so unless there is further feedback, this can be merged now.

PR #2078, opened 2018-05-15
syntax/parse: Add #:track-literals and syntax-parse-track-literals

lexi-lambda:
This PR makes three changes to `syntax/parse`:

  1. It adds the `syntax-parse-track-literals` function, which takes advantage of the unwindable state added in Racket 6.12. It adds the contents of the state under the `'literals` key to the `'disappeared-use` property of syntax objects, taking care to apply `syntax-local-introduce` as necessary.

  2. It adds the `#:track-literals` option to `syntax-parse`, which implicitly applies `syntax-parse-track-literals` to any syntax-valued result of a use of `syntax-parse`. This avoids the need for users to call `syntax-parse-track-literals` themselves, and is intended as the primary interface to this functionality.

     The primary reason not to enable this functionality by default is because it moves the final expression in a clause body out of tail position. Macros defined with `define-simple-macro` always use `#:track-literals`, since the body is never a function call, anyway.

  3. It adjusts the `static` syntax class to add to the literals list using `syntax-parse-state-cons!`, which ensures it is counted as a use even if it doesn’t appear in the macro’s expansion.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
@@ -20,6 +20,7 @@ Defines a macro named @racket[macro-id]; equivalent to the following:
 @racketblock[
 (define-syntax (macro-id stx)
   (syntax-parse stx
+    #:track-literals
     [((~var macro-id id) . pattern) pattern-directive ... (syntax template)]))
 ]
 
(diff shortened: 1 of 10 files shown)


lexi-lambda:
> If `#:track-literals` is specified, then why not require the result to be syntax? (The PR code already requires the result to be a single value---that should be documented explicitly.)

That’s reasonable. I’ll make that change and update the documentation later today.

> Automatically applying `syntax-local-introduce` is helpful for ordinary macros, but IIUC it will do the wrong thing for eg match expanders, etc. The `syntax-parse-track-literals` function should have an argument for skipping that step, but I don't think there's a simple fix for `#:track-literals`. I think the right answer will involve creating a common abstraction for custom expanders (and I've been looking into that for other reasons), but that won't happen immediately.

Reasonable. I’ll make that change as well.

Tangentially, I am pretty sure I know a way to make match expanders cooperate with `syntax-local-introduce` without much effort, so maybe I’ll try and look into that soon. It could be easily made backwards-compatible by making `syntax-local-match-introduce` just call `syntax-local-introduce`. I’ll open a PR if it works.

lexi-lambda:
@rmculpepper I’ve updated this PR with some changes to address your comments.

PR #2082, opened 2018-05-17
Eliminate the need for syntax-local-match-introduce, syntax-local-require-introduce, syntax-local-provide-introduce, and syntax-local-syntax-parse-pattern-introduce

lexi-lambda:
This PR changes `match` expanders, `require` expanders, `provide` expanders, and `syntax/parse` pattern expanders to reuse more macro expansion machinery, making them capable of reusing `syntax-local-introduce` instead of the current proliferation of custom functions. The bindings are still provided for backwards-compatibility, but they are made equivalent to `syntax-local-introduce`.

It does this by adding a new `syntax/apply-transformer` module, which implements a `local-apply-transformer` function that uses `local-expand` internally to apply a transformer exactly once.

I’m not sure who is best suited to review this PR, but I imagine @mflatt, @rmculpepper, and @samth might be interested parties.

### DIFF
--- a/racket/collects/racket/match/parse-helper.rkt
+++ b/racket/collects/racket/match/parse-helper.rkt
@@ -2,10 +2,9 @@
 
 (require (for-template racket/base)
          syntax/boundmap
+         syntax/apply-transformer
          racket/struct-info
-         ;macro-debugger/emit
-         "patterns.rkt"
-         "syntax-local-match-introduce.rkt")
+         "patterns.rkt")
 
 (provide ddk? parse-literal all-vars pattern-var? match:syntax-err
          match-expander-transform trans-match trans-match* parse-struct
@@ -170,20 +169,16 @@
                                   error-msg)
   (let* ([expander* (syntax-local-value expander)]
          [transformer (accessor expander*)]
-         ;; this transformer might have been defined w/ `syntax-id-rules'
-         [transformer (if (set!-transformer? transformer)
-                          (set!-transformer-procedure transformer)
-                          transformer)])
+         [transformer-proc (if (set!-transformer? transformer)
+                               (set!-transformer-procedure transformer)
+                               transformer)])
     (unless transformer (raise-syntax-error #f error-msg expander*))
-    (define introducer (make-syntax-introducer))
-    (parameterize ([current-match-introducer introducer])
-      (let* ([mstx (introducer (syntax-local-introduce stx))]
-             [mresult (if (procedure-arity-includes? transformer 2)
-                          (transformer expander* mstx)
-                          (transformer mstx))]
-             [result (syntax-local-introduce (introducer mresult))])
-        ;(emit-local-step stx result #:id expander)
-        (parse result)))))
+    (parse (local-apply-transformer
+            (λ (stx) (if (procedure-arity-includes? transformer-proc 2)
+                         (transformer-proc expander* stx)
+                         (transformer-proc stx)))
+            stx
+            'expression))))
 
 ;; raise an error, blaming stx
 (define (match:syntax-err stx msg)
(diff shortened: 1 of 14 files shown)


racket/collects/racket/match/parse-helper.rkt
```diff
@@ -169,21 +168,9 @@
...
-         ;; this transformer might have been defined w/ `syntax-id-rules'
-         [transformer (if (set!-transformer? transformer)
-                          (set!-transformer-procedure transformer)
-                          transformer)])
+         [transformer (accessor expander*)])
     (unless transformer (raise-syntax-error #f error-msg expander*))
-    (define introducer (make-syntax-introducer))
-    (parameterize ([current-match-introducer introducer])
-      (let* ([mstx (introducer (syntax-local-introduce stx))]
-             [mresult (if (procedure-arity-includes? transformer 2)
```
samth: The arity-2 code seems not supported by this change.

lexi-lambda: Thanks, good catch. I’ve added a test case for that situation and pushed a fix.

rmculpepper:
This seems like a good idea. I was initially worried about the effect on syntax protection (ie, taints and dye-packs), but after thinking about it some more I discovered a hole that currently exists that I expected this PR to patch. But it doesn't seem to, and I don't know why. Here's the code:

    (module secret racket/base
      (provide m)
      (define secret 42)
      (define-syntax-rule (m)
        (positive? secret)))

    ;; racket/match uses lazy-require; its dependencies must be forced to
    ;; get the right declaration-time inspectors before the code-inspector
    ;; is changed. (This is a separate issue that also needs fixing.)
    (require racket/match)
    (match 5 [_ (void)])

    ;; Downgrade the code inspector; subsequent modules are less privileged.
    (define old-code-inspector (current-code-inspector))
    (current-code-inspector (make-inspector old-code-inspector))

    ;; The following is prevented from working by the syntax protection system.
    #;
    (module insufficient-evil racket/base
      (require racket/match 'secret (for-syntax racket/base))
      (provide)

      (begin-for-syntax
        (define the-secret-id #f))

      (let ()
        (define-syntax (extract stx)
          (define estx (local-expand #'(m) 'expression '()))
          (syntax-case (syntax-disarm estx #f) (#%plain-app)
            [(#%plain-app _pos? secret-id)
             (set! the-secret-id #'secret-id)])
          #'(void))
        (extract))

      (define-syntax (reveal stx) the-secret-id)
      (printf "the secret is ~s\n" (reveal)))

    ;; The following works because the `evil` match expander inherits
    ;; the declaration-time inspector of the `match` macro.

    (module better-evil racket/base
      (require racket/match 'secret (for-syntax racket/base))
      (provide)

      (begin-for-syntax
        (define the-secret-id #f))

      (let ()
        (define-match-expander evil
          (lambda (stx)
            (define estx (local-expand #'(m) 'expression '()))
            (syntax-case (syntax-disarm estx #f) (#%plain-app)
              [(#%plain-app _pos? secret-id)
               (set! the-secret-id #'secret-id)])
            #'(quote 5)))
        (void (match 5 [(evil) 'mwahaha])))

      (define-syntax (reveal stx) the-secret-id)
      (printf "the secret is ~s\n" (reveal)))

    (require 'better-evil)

IIUC, the hole is in the way `syntax-arm` and `syntax-disarm` handle a `#f` argument for the inspector; they use the code inspector associated with the declaration of the module for the current macro transformer (see the docs for more fallback cases). Previously, that meant that every `match` expander, for example, would effectively run with the declaration-time privileges of `match` itself. I expected @lexi-lambda's changes to fix the hole, because now the transformer would be wrapped as a local macro of the module being expanded and get the lower privilege. @mflatt ?

Another reason I would like this change is that currently the macro profiler is blind to custom expanders; this would help improve its accuracy.

We should also use this for `for` and sequence syntax, for the same reasons. (That might require splitting out an implementation that doesn't use keywords and an interface that adds them.)

PR #2085, opened 2018-05-18
adding -B option into the expander to write the bytecodes of the ...

cderici:
…extracted linklet


samth:
You should probably add something to the README describing the new option.

PR #2086, opened 2018-05-19
Tweaks and fixes for strings.rkt

gus-massa:
**fix cache of regexp in `string-{split,trim,normalize-spaces}`**

When the separator is a string, these function construct a regexp that is cached to make repeated calls faster. But when the string is mutated it is necessary to recalculate the regexp. See the examples in the tests.

I'm reusing the cache for `mutable->immutable` that was created for `string-replace`. 

There is a minor difference in the cache `immutable->regexp` of `string-replace` and `string-split`. For `string-replace` it's a `weak-hash` and for `string-split` it is a `weak-hasheq`. Both are correct, the first uses a little less memory but it is slightly slower. Perhaps I should change one of them for consistency. (Moreover I may unify the `immutable->regexp` caches.)

**avoid quadratic time in `string-contains?`**

Use a regexp instead of the straightforward search.

I'm also extending the possible values of the `contained `argument so it can be a regexp, but it's easy to revert this change. I've updated the docs, but I didn't test the format of the docs so there is a high chance that it is wrong. _Please review carefully._

**improve error messages in string-join**

Previously, the error was detected and reported by `string-append`.

**use unsafe-undefined for missing arguments in string-join (?!?!)**

I'm still not sure if this is a good idea. This is similar to the recent change of functions with optional values. Using `unsafe-undefined` instead of a `gensym `makes it easier to avoid the check of the missing argument. But with this change there are many `unsafe-undefined` floating around in the internal functions, so it may not be a good idea.


mflatt:
Most of these changes seem good, but I worry about the chage to `string-contains?`. A program like

```
(time 
 (for ([i 100000])
   (string-contains? (make-string 100 #\x)
                     (make-string 100 #\y))))
```

runs much more slowly. Would it make sense to keep the API the same and switch the implementation to something like Knuth–Morris–Pratt, instead?

rfindler:
Here is an implementation of that here that may or may not be useful:

https://github.com/racket/gui/blob/c8417e8e0b134d057946129c895b9402e1eb2190/gui-lib/mred/private/wxme/text.rkt

(Sorry for the crappy link.)

gus-massa:
I remember the discussion of `string-prefix` and `string-suffix` where the author noted that the direct implementation was faster than my version with regexp's. 

I'll try to copy the version in wxme, but I have to remove the specific parts of wxme. (I never remember the details of KMP, so I have to read the details again each time.) I think I'll add some caching of the internal table and perhaps use a minimal threshold to use the old implementation for small strings. Avoiding the allocation of the internal table may be useful in short strings.

[Probably the other functions in this module can be made faster using a custom version when the pattern is a string. In particular, it would be nice to make `string-trim` faster when it is called preemptive and it doesn't have to trim anything. But this is much longer, so I'll left this possibility for later or someone else.]

gus-massa:
I made an update. Main changes:

**fix cache of regexp in `string-{split,trim,normalize-spaces}`**

In the cache of `string-replace` I changed the `weak-hash` to a `weak-hasheq`. I think it's a good trade off.

**use `unsafe-undefined` for missing arguments in `string-join`**

I'm still using a rename. Should I just search and replace all the `none` to make the code more explicit?

**avoid quadratic time in `string-contains?`**

I copied the KPM algorithm and renamed the arguments and internal variables and made a few small changes. But I also keep the old algorithm too as an internal function called `string-contains?/small`. The main function now selects which version to use according to the length of the strings.

I made a few test with different kind of strings.

* Random string of different length with only two characters (like `"aaaAAAaaAaAA"` and `"aAa"` but longer): `contains?/kmp` is 10% slower.

* Two long strings of the same length, with a single repeated different character each (like `"aaaaaa"` and `"AAAAAA"` but longer): `contains?/kmp` is 100% slower (yep, time is x2)

* One string is a repeated character with length `2L`, and the other has length `L` and the last character is different (like `"aaaaaaaaaa"` and `"aaaaA"` but longer): If `L` is 5, `contains?/kmp` is slower, if `L` is 10 they are almost equal, if `L` is 15 or more `contains?/kmp` is much faster.

I expect that most real use cases are more similar to the first example, where there is not too much repetition. So I decided to use `contains?/small` when the search string has less than 20 characters or the difference of  length of the two strings is less than 20, because the runtime of `contains?/small` is `O(L2*(L1-L2))` and it avoids the allocation of the auxiliary tables.

I'm still thinking if it would be better to increase the threshold from 20 to 50 or something like that.

rfindler:
Is there a way to share your improvement with the code in text%?

gus-massa:
I just send a PR https://github.com/racket/gui/pull/109.

1) Check that `start|m` is small enough instead of waiting until the last character to exit (this is useful for long strings). See [racket/racket#2086/R245](https://github.com/racket/racket/pull/2086/commits/31d71729c2da673d2965be6d67486a65295c3592#diff-b03d9637f37ba61e4a6b84aa459b5461R245) and [racket/gui#109/R3784](https://github.com/racket/gui/pull/109/commits/a1630c65862563b1c43d7a88905cf16dfb6c7792#diff-62180fd67008808e32b6dca8ff11815aR3784 )

2) When a `#f` is found in the kmp-table, advance the position of `start|m` to the current position instead of just advancing it 1. See [racket/racket#2086/R257](https://github.com/racket/racket/pull/2086/commits/31d71729c2da673d2965be6d67486a65295c3592#diff-b03d9637f37ba61e4a6b84aa459b5461R257) and [racket/gui#109/R3807](https://github.com/racket/gui/pull/109/commits/a1630c65862563b1c43d7a88905cf16dfb6c7792#diff-62180fd67008808e32b6dca8ff11815aR3807)

I'm still worried that these changes break something of the multiple overlapping results, or the reverse order. This version has fewer cases to consider.

gus-massa:
The tables generated by `build-kmp-table` have too few `#f`. For example, using the first example of Wikipedia `"abcdabd"` I get `#(#f 0 0 0 0 1 2)` instead of `#(#f 0 0 0 #f 1 2)`. This will not cause an error, but a tiny inefficiency. I'll try to take a look and fix this. Note: The Wikipedia example has also this problem, moreover it has a bug that generate wrong tables. I'll fix it later.

gus-massa:
I updated the part that builds the KMP tables, but I didn't merge it. Now it produces the same tables as the Wikipedia article. [It uses the same algorithm. I now think Wikipedia is right, and I was confused by the combination of strange indentation and blank lines.]

I tried to use a single `loop` to build the table, but I had to add another argument to the `loop` that is a flag to write the table only once in each possition. Something like

    (let loop ([pos 1] [cnd 0] [write? #t]) 
      [...]
      (when write?
        (vertor-set! t pos cnd))
      [...])

The code with the additional flag is shorter but it is more difficult to understand, so I left the version with two `loop`s.

The tables have an extra last coefficient that is the value to use in case you want to get all the (maybe overlapping) matches or count them. It's currently unused, but I left it there in case someone needs the extended version.

I think this is mergeable as is, but running more tests and another look at the code will be helpful.

----

One possible improvement is to add some test for the private functions, in particular the one that builds the KMP table. It's not possible to test it in string.rktl because the functions are not provided. Creating a new private module is an overkill. Is it possible to add something like this?

    (module+ private
      (provide build-kmp-table))

Is this a good idea? Is there something similar in another test file?

bennn:
> Is this a good idea? Is there something similar in another test file? 

Sounds good, `racket/require` has a `for-testing` submodule:
https://github.com/racket/racket/blob/master/racket/collects/racket/require.rkt#L135

gus-massa:
I just added some tests in a separate commit, so it's easy to remove them in case it's not  good idea. (In case the decision is to merged both, I'd squash them.)

I used `(module+ for-testing ...)` for consistency. Is that pattern used in other places? Is that documented? I still prefer `(module+ private ...)` because it is more clear that it is similar to the `private` folder where it is clear that the functions are undocumented, unstable, unsupported and perhaps unsafe. 

In this case the three functions are safe to call. The only problem is that they don't raise a nice error. (The result of `build-kmp-table` is a mutable vector, but it creates a new vector each time the it is called. The caching is done inside `string-contains?/kmp`. If the caching were done inside `build-kmp-table` then it would be necessary to make an immutable vector to avoid someone messing it.)

bennn:
> I used (module+ for-testing ...) for consistency. Is that pattern used in other places?

I tried a grep and 2 `for-testing`s

```
$ ack "\+ for-test" ~/code/racket/fork

..../rackunit/rackunit-lib/rackunit/private/check-info.rkt
30:(module+ for-test

..../rackunit/rackunit-lib/rackunit/private/format.rkt
11:(module+ for-test

..../redex/redex-lib/redex/private/ambiguous.rkt
700:(module+ for-tests 

..../typed-racket/typed-racket-lib/typed-racket/utils/opaque-object.rkt
42:(module+ for-testing

..../racket/collects/racket/require.rkt
135:(module+ for-testing
```

gus-massa:
Intermediate update: 

I've done some test and I think it's better to not use a cache of the KMP table. The lookup in the hash is many times slower than building the table again. 

I'm also thinking about using an hybrid approach, that tries to use the straightforward algorithm, and builds the table only if something interesting is found.

I'll post my new proposal later (probably in 2 or 3 weeks).

gus-massa:
Updated version (after a long time).

I removed the cache completely, because it added a relative big constant time.

I made a variation of the KMP method. It initially tries to find a match without the kmp-table until it finds a partial match of at least 4 characters. This helps to avoid the creation of the table when the two strings are very different. 

It also doesn't create the table when there are less than 4 characters left. This is in particular useful when both strings have almost the same length.

The first table that is created has a size that is similar to the partial match, and if it must be extended the whole table is created.

To run the tests, I exported the old version and the function to create the kmp-table in a submodule called `private`. These functions are safe to call, but they are unsupported and may change/disappear in the future. I think that `private` express this better because it's more similar to the `private` subdirectories.

PR #2091, opened 2018-05-21
Support -l for raco make.

samth:


samth:
@mflatt This is ready for review.

rfindler:
Should it be `-t` to match `racket`?

samth:
`-t` is analogous to the current default behavior of `raco make`. The new flag works like `-`l for `racket`.

rfindler:
Oh, okay. Maybe the docs should be updated to match `racket`s? Right now the term "library path" confused me.

samth:
Unfortunately the [Reference docs](https://docs.racket-lang.org/reference/running-sa.html#%28part._mz-cmdline%29) don't provide a good term to use here. The [Guide](https://docs.racket-lang.org/guide/racket.html#%28part._start-module-mode%29) uses the term "`lib` module path", which suggests that @mflatt's suggestion of "collection-based module path" is the right thing to use in both the help text and the docs.

pkgs/compiler-lib/compiler/commands/make.rkt
```diff
@@ -21,11 +21,16 @@
...
+(define coll-paths (make-parameter null))
+
 (define mzc-symbol (string->symbol (short-program+command-name)))
 
-(define source-files
+(define source-file-paths
   (command-line
    #:program (short-program+command-name)
+   #:multi
+   [("-l") path "Compile <path> interpreted as a collection path"
```
samth, line 32:
```suggestion
   [("-l") path "Compile <path> interpreted as a collection-based module path"
```

pkgs/compiler-lib/compiler/commands/make.rkt
```diff
@@ -48,7 +53,18 @@
    [("--vv") "Very verbose mode"
     (verbose #t)
     (very-verbose #t)]
-   #:args (file . another-file) (cons file another-file)))
+   #:args files
+   (when (and (null? files)
+              (null? (coll-paths)))
+     (raise-user-error (format "~a: expects at least one file path or collection path"
```
samth, line 59:
```suggestion
     (raise-user-error (format "~a: expects at least one file or module path"
```

pkgs/racket-doc/scribblings/raco/make.scrbl
```diff
@@ -36,7 +36,10 @@ The @exec{raco make} command accepts a few flags:
 
 @itemlist[
 
- @item{@Flag{j} @nonterm{n} --- Compiles argument modules in parallel,
+ @item{@Flag{l} @nonterm{path} --- Compiles @nonterm{path} interpreted
+       as a library path, as for @racket[require].}
```
samth, line 40:
```suggestion
       as a collection-based module path, as for @racket[require].}
```

rfindler:
Yeah, I agree. Perhaps also worth considering is making the `racket -l` help text match too?

mflatt:
LGTM

PR #2092, opened 2018-05-21
`format` error message fixes

samth:


mflatt:
I think the current `format` error message limits reporting to a case with 50 arguments or less, and it divides `(error-print-width)` by the number arguments to print.

samth:
I think this is ready now.

mflatt:
LGTM

(Probably BC and CS should both use the newer field-style error format, but that's for a different day.)

PR #2094, opened 2018-05-21
doc fix for chaperone-struct-type

LiberalArtist:
The `guard-proc` argument must accept one more argument than the constructor for the chaperoned struct type, not the same number of arguments.


LiberalArtist:
@samth This looks like the same build failure as in https://github.com/racket/racket/pull/2058 (thanks for your help with that!). I don't know how to get the build to run again (or even if it's possible as a non-maintainer).

samth:
If you just update the PR (for example, by rebasing), the build will restart. I did it manually just now, though.

samth:
The build passed, but I'd like to make sure someone who's used those functions recently reviews this patch: @mflatt @rfindler @pnwamk

mflatt:
This correction looks right to me.

PR #2100, opened 2018-05-24
Add pretty-printing support to write-json and friends.

cwebber:
New #:indent keyword which, if provided with an integer, will pretty-print output at that indentation level.

Licensed under MIT (Expat) / Apache V2 / LGPLv3+, Racket maintainers' pick.

### DIFF
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -61,7 +61,8 @@ the @rfc for more information about JSON.
 
 @defproc[(write-json [x jsexpr?] [out output-port? (current-output-port)]
                      [#:null jsnull any/c (json-null)]
-                     [#:encode encode (or/c 'control 'all) 'control])
+                     [#:encode encode (or/c 'control 'all) 'control]
+                     [#:indent indent (or/c exact-integer? #f) #f])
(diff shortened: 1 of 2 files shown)


samth:
This has a couple small conflicts now -- @cwebber can you resolve them?

pkgs/racket-doc/json/json.scrbl
```diff
@@ -74,18 +75,26 @@ the @rfc for more information about JSON.
...
+  If @racket[indent] is supplied as an integer, output will be "pretty
+  printed" at the specified indentation level.
+
 @examples[#:eval ev
   (with-output-to-string
     (λ () (write-json #hasheq((waffle . (1 2 3))))))
   (with-output-to-string
     (λ () (write-json #hasheq((와플 . (1 2 3)))
                       #:encode 'all)))
+  (displayln
```
sorawee, line 87: If you really want to display, I think you can use `write-json` directly without `displayln` and `with-output-to-string`. For consistency though, I think it probably should evaluate to a string value.

sorawee:
There probably should be tests? The test file is at https://github.com/racket/racket/blob/master/pkgs/racket-test/tests/json/json.rkt.

PR #2101, opened 2018-05-25
bump version for contract-equivalent

bennn:


samth:
Thanks! I wonder how hard it would be to write a script that bumps the version.

PR #2105, opened 2018-05-26
typo: make-file-or-directory-link error message

### LINKED ISSUE #2097: make-file-or-directory-link's misleading error
In a directory, create a file named `a.txt`. Then, link `a.txt` to `b.txt` via `make-file-or-directory-link`. This results in an error as expected, since `a.txt` already exists:

```
$ touch a.txt
$ racket -e '(make-file-or-directory-link "b.txt" "a.txt")'
make-file-or-directory-link: cannot make link;
 the path already exists
  path: /home/sorawee/play/b.txt
```
However, the above error is misleading. It says that the path already exists, which actually refers to `a.txt`. But the given "path:" is `b.txt` instead.

bennn:
Change error message to blame the symlink, not its destination

Closes #2097

### DIFF
--- a/pkgs/racket-test-core/tests/racket/file.rktl
+++ b/pkgs/racket-test-core/tests/racket/file.rktl
@@ -1966,6 +1966,7 @@
   (unless (eq? 'windows (system-type))
     (delete-file tf)
     (make-file-or-directory-link "other.txt" tf)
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks!

PR #2124, opened 2018-06-08
struct-out: get correct list of super-struct mutators

bennn:
struct-out was putting the super-struct's accessors into two parts of a
struct-info: the accessor list and the mutator list

this commit puts the accessors only in the accessor list and the
mutators in the mutator list


gus-massa:
It would be nice to add the example in the issue to the test suit to avoid regressions.

bennn:
Thanks for the reminder, I added a test.

PR #2125, opened 2018-06-12
Allow local-expand stop list to opt-out of implicit extension and adjust the expansion of #%plain-module-begin to reinstate the stop list after initial partial expansion

lexi-lambda:
This pull request makes two changes:

  1. It adds a way to request that `local-expand` *not* implicitly extend the stop list when given a non-empty list.

  2. It adjusts the expansion of `#%plain-module-begin` to avoid throwing away the stop list after performing partial expansion of the module body (as long as `module*` is in the stop list). This makes `#%plain-module-begin` more consistent with other forms that perform partial expansion, such as local binding forms that support internal definition contexts and most user macros.

These changes support a currently-unmerged pull request to Hackett, lexi-lambda/hackett#80, which implements a configurable elaborator and functional dependencies using these changes.

While this PR proposes a particular implementation of these two changes, I don’t necessarily think that this is the best way to implement them. There are some alternative mechanisms that could be used instead of the ones proposed here:

  - This PR allows callers of `local-expand` to pass a list containing the symbol `'only` as `stop-ids`. This is kind of awkward. I think the ideal interface would be to add an optional keyword argument to `local-expand`, something like `#:extend-stop-list?`, but my understanding is that the way the bootstrapping process works currently prohibits exports with keyword arguments.

    An alternative would be to just add an extra optional positional argument, but I think that wouldn’t be any better than the hack this PR implements, and I’d like to avoid adding new positional arguments to `local-expand` if at all possible.

    In general, I would appreciate being able to create a new interface to `local-expand`, one that subsumes the four existing variants (not including `syntax-local-expand-expression`) using keyword arguments like `#:stop-ids`, `#:intdefs`, `#:transformer?`, and `#:capture-lifts?`  and has more reasonable default behavior for the stop list. But that’s a bigger question.

  - The way the adjustment to `#%plain-module-begin` works is a little bit awkward, since it requires `module*` be in the stop list. I decided to take this approach since it seemed least likely to have backwards-compatibility problems, but an alternative design would be to add `module*` implicitly if a non-empty stop list is provided.


samth:
On the first point, why not add an additional positional argument to `local-expand` in the expander, and then wrap that with a keyword-argument (I agree that this would be nicer as an API) in part of the implementation of `racket/base`?

lexi-lambda:
I tried that before a couple weeks ago, but it still caused the bootstrapping process to fail. [According to Matthew](https://racket.slack.com/archives/C06V96CKX/p1525729057000337):

> It looks like the flattener can't infer that the new module is free of side effects. […] The expansion involves `make-optional-keyword-procedure`, and a later pass of the flattener can figure out that calling `make-optional-keyword-procedure` has no side effect, but the check for a reference to expander functions happens before that (and didn't seem easy to reorder, for reasons that I forget). So, adding keywords for functions below the expander level can work this way, but not functions from the expander layer. The right workaround isn't immediately obvious.

samth:
I don't think what you did and what I'm suggesting are the same thing. What I'm suggesting is your change + a function defined in `racket/private/pre-base` (for example) with some keyword arguments.

samth:
Ah, indeed, a little more trickery is needed. See https://github.com/racket/racket/commit/ed71354ec9385513971c42d311c8738817dd4b02

mflatt:
A new `++pure` may need to be added in multiple makefiles, though. I think we shouldn't do that, and we should instead improve the flattener.

samth:
So, I just looked at the flattener, and basically the issue is as follows:

- Checking for dependencies on the expander happens in `extract/check-and-report`, which works on a collection of linklets before flattening. 
- That makes use of a side-effect analysis in the compiler, which doesn't figure out that creating a keyword function is side-effect-free.
- There's a subsequent more sophisticated side effect analysis in the flattener, but that works on one single flattened linklet.

I see a few options:
 - just add `++pure` to the various makefiles
 - make the implementation of `compile` smarter (this has other benefits, of course, but seems hard)
 - run the check for runtime dependencies on the flattened linklet (this makes the error messages much worse)
 - run the check twice, and use the error from the former if the latter also fails (seems like a hack but I'm going to try it)

mflatt:
The last is closest to what I had in mind: delaying the error that the flattener currently reports until after the flattened-linklet pass, and then report the error only if the relevant code persists. But I didn't try it to see where it goes wrong.

samth:
Ok, see https://github.com/lexi-lambda/racket/pull/1

lexi-lambda:
> It's not immediately obvious to me that resetting the stop list under `begin-for-syntax` is the right thing. A programmer might put a `module*` under a `begin-for-syntax` to shift its phase relative to the enclosing module, and probably that `module*` should not be expanded if `module*`s are supposed to be left alone. Maybe the current stop list should be phase-shifted instead.

Here was my reasoning. `#%plain-module-begin` needs to do some partial expansion in order to expand module bodies. There’s no getting around that. I don’t think it’s a good (or particularly useful) idea to try and constrain that partial expansion further, since we still want predictable module expansion. There are certain invariants that module expansion maintains, which support things like recursive bindings and order of phase 1 evaluation.

I don’t want to weaken any of those invariants. So evaluating `begin-for-syntax` forms in the order they appear is still important, and in order to evaluate a `begin-for-syntax`, it needs to be fully-expanded. So I don’t think it’s safe to weaken module expansion in that way, perhaps unless we did a lot more work to track if expansion stopped early and perform evaluation if it didn’t.

> This change also makes `#%plain-module-begin` special in a new way. Should the current stop list be exposed to macros by `syntax-local-stop-list`, or something like that? (Would exposing the current stop list be any sort of information or capability leak?)

Fun fact: when putting together this PR, I actually added a `syntax-local-stop-ids` function. It was easy, it made `#%plain-module-begin` less special, and it made things simpler to document. But then I thought about the capability leak, and I backed the changes out since I didn’t actually need them just yet.

The problem with `syntax-local-stop-ids` is that, theoretically, some macro could grab one of the stop ids that was supposed to be private and stick it somewhere it isn’t supposed to go. I think the answer here is to make `syntax-local-stop-ids` explode dye packs that the current code inspector doesn’t have access to on each identifier in the stop list, but this would still require people sticking private identifiers in the stop list to arm each identifier, which they might not have currently done. So that would involve making `local-expand` implicitly call `syntax-protect`, but then some people might wish to disable that implicit arming… and the whole thing became too complicated for me to feel like working all the way through.

That said, with @samth’s changes, maybe we can add a `#:protect-stop-ids?` keyword argument to handle that behavior, too. It doesn’t do much to help with the complexity of all the things `local-expand` implicitly does, but at least it would keep things safe, and it would allow users who didn’t want the implicit behavior to opt out of them.

mflatt:
On `begin-for-syntax`: That makes sense, except for `module*`, which can't contribute to the evaluation of compile-time code. Maybe just keep (and shift) `module*` if it's present in the current stop list?

On `syntax-local-stop-ids`: Yep, that's what I started to think. So maybe punting for now is ok.

lexi-lambda:
> On `begin-for-syntax`: That makes sense, except for `module*`, which can't contribute to the evaluation of compile-time code. Maybe just keep (and shift) `module*` if it's present in the current stop list?

Yes, that makes sense. I’ll see if I can plumb that through somehow (and I’ll check what the Racket 6 expander does with a stop list that includes just `module*` when it encounters a `module*` submodule inside `begin-for-syntax`).

> On `syntax-local-stop-ids`: Yep, that's what I started to think. So maybe punting for now is ok.

It’s definitely more subtle than it seems. Thinking on that last bit a little bit more, tainting identifiers returned from `syntax-local-stop-ids` isn’t really a perfect solution. It allows some of the specialness of `#%plain-module-begin`, namely checking if `module*` is in the current stop list, but it doesn’t really have a great story for recursive calls to `local-expand`, since it will taint all the identifiers. I’ll keep thinking about it, but I think punting on it is okay.

(By the way, it appears that `local-expand` is currently perfectly okay with using a tainted identifier in `stop-ids`. Should it? I don’t immediately see a reason why that could allow someone to do something malicious, but I haven’t thought about it at length.)

lexi-lambda:
I’ve incorporated @samth’s improvements to the flattener into this pull request, and I’ve pushed a commit that makes `local-expand` use a new `#:extend-stop-ids?` keyword argument instead of requiring the list be prefixed with `'only`. Also, it turns out the existing code already stopped on `module*` inside `begin-for-syntax` due to the way `module*` expansion is deferred until after expanding the module, but I’ve adjusted a test case to ensure that behavior is maintained.

lexi-lambda:
@mflatt The tests all passed, so I think this is now possible to merge (after squashing some commits, at least). Do you have any remaining qualms or comments?

mflatt:
I’d like to look more, especially at the flattened changes.
Do you need this in v7, or can it be post-v7?

lexi-lambda:
I don’t need it in v7. If you’d like to look at the changes in more detail, and fixes for v7 are a higher priority right now, that’s fine—I can continue working on a branch. I’d like to merge this because it would let me merge my Hackett changes that depend on it, but that isn’t blocking me—I can continue to work on a branch there, too.

mflatt:
The flattener improvement isn't quite right — it effectively disables the check — but it's on a good track, and I'll push a working version separate from this PR.

mflatt:
@lexi-lambda I should have checked whether the revised flattener works with your changes, but assuming that it does, feel free merge your changes.

(Also your `module*` observations make sense in retrospect.)

lexi-lambda:
Thank you both for your help!

Issue #1084, opened 2015-10-07
raco pkg errors


lehitoskin:
I'm using Gentoo and I just installed 6.2.1 from version 6.1.1 and when downloading packages (any of them) with raco, I'm getting these errors:

```
raco setup: --- creating launchers ---
raco setup: --- installing man pages ---
raco setup: --- building documentation ---
get-doc-info error: /usr/share/doc/racket/guide/in.sxref::0: read (compiled): wrong version for compiled code
  compiled version: 6.1.1
  expected version: 6.2.1
raco setup: docs failure: delete-file: cannot delete file
  path: /usr/share/doc/racket/guide/out0.sxref
  system error: Permission denied; errno=13
raco setup: --- installing collections ---
raco setup: --- post-installing collections ---
raco pkg install: packages installed, although setup reported errors
```

I apologize if this is technically the wrong place to make the issue, there's just so many repositories.

mflatt:
It looks like something has gone wrong installing v6.2.1 over v6.1.1, with "/usr/share/doc/racket/guide/in.sxref" (and maybe other files) left over from v6.1.1.

Can you say more about how each version was installed and how v6.1.1 was installed?

Thanks!

lehitoskin:
v6.2.1 was installed through Portage. Here's the [ebuild that was used](http://gpo.zugaina.org/AJAX/Ebuild/10634650/View).

v6.1.1 was installed outside of Portage by compiling from source because an up to date ebuild didn't exist at the time (which is a real problem when you're using "bleeding-edge" Gentoo).

For both versions I enabled X, doc, futures, jit, places, and threads.

jsmaniac:
More generally, when installing a new version of Racket, if I have locally-installed packages (e.g. with `cd my-package/; raco pkg instll`), then the compiled code in their `compiled/` directories is not updated.

Instead of displaying

```
compiled/my-file_rkt.zo::1: read (compiled): wrong version for compiled code
  compiled version: 6.90.0.29
  expected version: 6.11
```

it would be nice if Racket silently re-compiled the files, and overwrote the `.zo`. The alternative is to delete all the `compiled` directories, which is error-prone (I'm always afraid `find /path/to/package -type d -name compiled -print0 | xargs -0 rm` is going to accidentally delete something that it shouldn't.

lexi-lambda:
Won’t just running `raco setup` do what you describe, @jsmaniac?

jsmaniac:
[...]

PR #2127, opened 2018-06-12
[WIP] Guide on how to upgrade from one version of Racket to another

SuzanneSoy:
Hi all!

This is a very early draft of a guide helping users upgrade from one version of Racket to another (e.g. 6.11 to 6.12)

Before I fill in the prose, translate the commands to Windows & macOS, and integrate this somewhere in the docs, I would like feedback on whether these lists of commands are the correct ones.

Comments on where this should be placed in the docs, and on how upgrade from 5.x to 6.x needs to be done are also very welcome.

Since there are a few changes from 6.x to 7.x (e.g. `(require syntax/parse)` is not needed anymore), would this be a good place to gather a short guide on how to update code to get rid of legacy stuff?

### DIFF
--- a/pkgs/racket-doc/scribblings/getting-started/upgrading.scrbl
+++ b/pkgs/racket-doc/scribblings/getting-started/upgrading.scrbl
@@ -4,33 +4,14 @@
 
 @section{Linux}
 
-@subsection{Upgrading from a release version to a newer release versions, or from a nightly to a newer nightly}
-
 @itemlist[
   @item{Download the newer release from @url{https://download.racket-lang.org/}.}
   @item{@code{chmod +x racket-X.Y-PLATFORM.sh}}
   @item{@code{./racket-X.Y-PLATFORM.sh} and answer the questions to choose where the new version should be installed.}
+  @item{@code{raco pkg migrate}
   @item{@code{raco setup}}]
 
-@subsection{Upgrading from a release version to a nightly}
-
-@itemlist[
-  @item{Download the newer release from @url{https://download.racket-lang.org/}.}
-  @item{@code{chmod +x racket-X.Y-PLATFORM.sh}}
-  @item{@code{./racket-X.Y-PLATFORM.sh} and answer the questions to choose where the new version should be installed.}
-  @item{@code{mv ~/.racket/snapshot ~/.racket/snapshot.bak} remove and make a backup of the old "snapshot" folder if one existed}
-  @item{@code{mv ~/.racket/OLD_VERSION ~/.racket/snapshot}}
-  @item{@code{raco setup}}]
-
-@subsection{Upgrading from a nightly to a release version}
-
-@itemlist[
-  @item{Download the newer release from @url{https://download.racket-lang.org/}.}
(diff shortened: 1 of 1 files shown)


gus-massa:
I'm not sure if this is the correct thread, but I use the autoinstaller in WIndows and something that is very unexpected is that after installing a new version I have to go to DrRacket and use "file">"Package Manager" and then use the "Copy from version" tab to reinstall all the additional packages.

mflatt:
@jsmaniac Overall, I think you must have in mind some situation or constraints where installation and the `raco pkg` tools don't "just work". Maybe we can better understand the problems and make tools do the right thing, instead of writing workaround instructions.

For the first case, what is the scenario where you need to run `raco setup` after installing a new release? I think you'd have to be using `PLTCOLLECTS` or something like where old collections used by the new installation — and then I'd say "don't do that". But maybe there's some other situation that doesn't currently work and should.

For the second and third cases, for most users, `raco pkg migrate` (= "Copy from version" in DrRacket) is the intended approach, instead of renaming a directory and running `raco setup`. Using `raco pkg migrate` will get the right versions of packages that were sourced from the main catalog. It should also correctly move over development links or packages that were otherwise not sourced from the main package catalog. Maybe `raco pkg migrate` isn't well enough known. Or maybe you have in mind migrating settings/configurations instead of just packages, and `raco pkg migrate` doesn't do that, but maybe it should.

jsmaniac:
@mflatt I'm not sure what PLTCOLLECTS is, but I don't think that I use it.

From my experience, when I upgrade from a nightly to another nightly, I get errors about old .zo files which are compiled for the wrong version of Racket, at least in packages which are linked from local folders. I got told a couple of times (last time by @lexi-lambda  [here](https://github.com/racket/racket/issues/1084#issuecomment-391404275)) that running `raco setup` should sort this out. Ideally, I would say that `racket`, `drracket` and `raco make` should simply ignore and/or regenerate the .zo files when they are for a different or older version.

I didn't know at all about `raco pkg migrate`. Would it be sensible for the first run of Racket / DrRacket to automatically run `migrate`? The installer could also do it in principle, but that wouldn't work for people updating using their distribution's package manager (e.g. `apt-get`).

You make a good point about the settings. I hadn't though about those, but I guess it would be nice to have them migrated too.

samth:
It's slightly different than what this guide is about, but I've helped some people who have slow Racket installs due to skewed timestamps on zo files that were fixed with `raco setup`.

mflatt:
Running `raco setup` after a snapshot upgrade does make sense. That's because all snapshots are configured with the same name, "snapshot", instead of getting a version-specific name. It has never been clear that using the same name for all snapshots was the right choice, and we should revisit that issue.

Having DrRacket offer to run `raco pkg migrate` is an interesting idea.

I worry that ignoring old-version ".zo" files is moving the wrong direction. That Racket silently ignores old-timestamp ".zo" files has caused confusion and was probably the wrong choice. Unfortunately, regenerating ".zo" files automatically is tricky, especially in `racket`. There are many issues with filesystems, permissions, file dependencies, and namespaces. There's a package that lets you opt-in to auto-compile, though, and maybe we could simplify opting in to that behavior.

PR #2130, opened 2018-06-14
"Fix" the packers.rkt test.

samth:
It appears that sometimes file timestamps can be off by 2, in addition to the
granularity-of-2-seconds issue.


mflatt:
Ok. I don't know why it's sometimes off, and I haven't found a reason to care.

PR #2131, opened 2018-06-14
Some optimizations for schemify

gus-massa:
I was looking at [startup.inc](https://github.com/racket/racket/blob/master/racket/src/racket/src/startup.inc) and I notice a few patterns. I know this file is not related, but I guess the same patterns appear in the shemified version, so I added them.

Is it possible to run also these optimizations in the creation of startup.inc? IIUC it is not possible to run this in the expander in all the linklets because the linklets may be recombined later, but the code in startup.inc is in the final form, so it would be ok to apply some reductions.

About each commit: 

**add reduction of `(if (not t) e1 e2)`**

In startup.inc, there are approximately 700 instances of `(if (not t) e1 e2)` and 300 instances of `(not #f)` (and 200 of them are in the intersection).

**add reduction of `(list)` and `null` to `'()`**

I don't imagine how this could be useful now, but it's one of the small straightforward reductions that may compose nicely with other reductions in the future.

**add reduction of `(begin e)` and `(let-values () e)`**

In startup.inc, there are approximately 5000 instances of `(let-values () e)`, so removing them will reduce the size of startup.inc in a 5%. I hope this doesn't break the continuations marks or the splicing of `begin` in the modules.

### DIFF
--- a/racket/src/schemify/optimize.rkt
+++ b/racket/src/schemify/optimize.rkt
@@ -46,6 +46,10 @@
             '#t
             v)]
        [else v])]
+    [`(list)
+     ''()]
+    [`null
+     ''()]
     [`,_
      (define u (unwrap v))
      (cond
--- a/racket/src/schemify/optimize.rkt
+++ b/racket/src/schemify/optimize.rkt
@@ -19,6 +19,12 @@
      (if (literal? t)
          (if (unwrap t) e1 e2)
          v)]
+    [`(if (not ,t) ,e1 ,e2)
+     (optimize `(if ,t ,e2 ,e1) prim-knowns knowns imports mutated)]
+    [`(not ,t)
+     (if (literal? t)
+         `,(not (unwrap t))
+         v)]
     [`(procedure? ,e)
      (define u (unwrap e))
      (cond
--- a/racket/src/schemify/optimize.rkt
+++ b/racket/src/schemify/optimize.rkt
@@ -50,6 +50,10 @@
      ''()]
     [`null
      ''()]
+    [`(begin ,e)
+     e]
+    [`(let-values () ,e)
+     e]
     [`,_
      (define u (unwrap v))
      (cond


mflatt:
The `if` and `not` optimizations might expose some other transformations at the `schemify` level, so these changes are ok with me. The only drawback could be slowing down the `schemify` pass without speeding up other passes, but a slowdown doesn't seem likely with these changes.

I don't think the `(begin e)` transformation will fire, because "schemify.rkt" already makes that simplification. Similarly, `(let () e)` shouldn't fire, because it's always `let-values` in the relevant pass, and `(let-values () e)` is similarly handled in "schemify.rkt". Maybe the function name `optimize` is misleading, because really it's a few transformations in support of various optimizations that happen in "schemify.rkt".

I don't have a clear idea of when the conversion to `'()` would help, at least at  the `schemify` level. It makes sense as part of a larger set of optimizations, but I think we want to leave those to other compiler layers.

Along those lines, I don't think applying optimizations to "startup.inc" would help. In practice, that file is compiled to bytecode or C, and the compilers to those forms already perform these kinds of optimizations. The optimize pass in `schemify` is aimed at idioms that are important to Racket and are either not recognized by Chez Scheme or that are easily obscured by schemifying. Granted, general simplications can help improve inlining by shrinking functions so that they're more likely to be inlined, but we don't want to move a lot of new complexity into this layer.

gus-massa:
If the idea is to keep this part as small as possible, perhaps it is enough to add the reduction for `(not <literal>)`, without the reduction for `(if (not t) e1 e2)`. [I think that expansion of `for` likes to create `(not #f)`. See [for.rkt#L1488](https://github.com/racket/racket/blob/master/racket/collects/racket/private/for.rkt#L1488).]

I took another look, and most of the reductions are in schemify.rkt. The file optimize.rkt has only a few reductions that are common to shemify.rkt and find-definition.rkt.

Just to be sure that I understand. About startup.inc, is the giant string included in racket.exe and parsed at run time or the giant string is compiled to bytecode and only the bytecode is included in racket.exe?

mflatt:
Ok on `not`.

Yes, "startup.inc" is compiled to bytecode (or C code) in "cstartup.inc" as part of the build process for `racket`[`cgc`] (although that does involve creating an intermediate `racketcgc` that embeds "startup.inc" directly).

samth:
The best way I've found to see the results of the bytecode optimizer for the expander is `make decompile` in the src/expander directory.

gus-massa:
Merged the `not `part in https://github.com/racket/racket/commit/9cd3798ca964ad7a1badd75c44b93686fc6511ba

PR #2136, opened 2018-06-15
Use a custom implementation of promises.

samth:
Saves 43 definitions and 997 lines of (de)compiled code.


mflatt:
I think you could drop exception handing, too.

mflatt:
LGTM

PR #2141, opened 2018-06-17
Add #:src argument to syntax:read-xml

AlexKnauth:
This PR adds an optional `#:src` argument to `syntax:read-xml` and `syntax:read-xml/element`, and puts this value into the `syntax-source` field of the result, just like `read-syntax` does with its first argument.

```racket
> (syntax:read-xml #:src "file.xml" (open-input-file "file.xml"))
#<syntax:file.xml:1:0 (tag...>
```

### DIFF
--- a/pkgs/racket-doc/xml/xml.scrbl
+++ b/pkgs/racket-doc/xml/xml.scrbl
@@ -227,12 +227,16 @@ Reads a single XML element from the port.  The next non-whitespace
 character read must start an XML element, but the input port can
 contain other data after the element.}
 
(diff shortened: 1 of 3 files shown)


jeapostrophe:
Does it make sense to add something to the test suite?

AlexKnauth:
Is this the test file I should be adding the tests to?
https://github.com/racket/racket/blob/master/pkgs/racket-test/tests/xml/test.rkt

AlexKnauth:
I have added some tests to check that the `#:src` argument goes into the `syntax-source` field of the result.

jeapostrophe:
Awesome, thanks!

PR #2146, opened 2018-06-20
optimizer: reduce (length '()) ==> 0

gus-massa:
This pattern is common in the internal code of the keyword procedures.


mflatt:
LGTM

PR #2153, opened 2018-06-27
One line revert to patch: 41fd4f3a5e0c6d4ca0ef822b168bf33a5c9068a6

LeifAndersen:
The docs had an erroneous 'only, probably from development, that should
be removed.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
@@ -232,7 +232,7 @@ identifier, the @racket[exn:fail:contract] exception is raised.
 
 @defproc[(local-expand [stx any/c]
                        [context-v (or/c 'expression 'top-level 'module 'module-begin list?)]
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -232,7 +232,7 @@ identifier, the @racket[exn:fail:contract] exception is raised.
 
 @defproc[(local-expand [stx any/c]
                        [context-v (or/c 'expression 'top-level 'module 'module-begin list?)]
-                       [stop-ids (or/c (listof identifier?) (cons/c 'only (listof identifier?)) #f)]
+                       [stop-ids (or/c (listof identifier?) empty #f)]
```
lexi-lambda: I’d just remove the `empty` special case altogether. It is subsumed by the `(listof identifier?)` contract, and with other changes, an empty list is less special than it used to be. I think documenting it separately is just confusing.

mflatt:
I think the change that this PR changes has gone away – right @lexi-lambda?

lexi-lambda:
Yes, I think you’re right.

PR #2157, opened 2018-07-02
Some decomplier improvements.

samth:


mflatt:
The first change looks good. The second one seems misleading, because `#<void>` will turn into `(void)` under a quote.

mflatt:
Ok, right – I did misunderstand.

You may want `(list 'void)` to avoid the `#0=` and have `(void)` instead of `#0#` show up in other places.

samth:
Thanks, that's definitely better.

PR #2166, opened 2018-07-09
Add option to create relative paths for 'serialize'

LeifAndersen:
Serialize would previously always create complete byte-string paths.
This adds an optional parameter to serialize (#:relative-to) to enable
relative path creation.

Now when deserialize finds a relative path, it resolves it with
respect to `current-load-relative-directory`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/serialization.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/serialization.scrbl
@@ -16,7 +16,13 @@ See @racket[serialize] for an enumeration of serializable values.}
 
 @; ----------------------------------------------------------------------
 
-@defproc[(serialize [v serializable?]) any]{
+@defproc[(serialize [v serializable?]
+                    [#:relative-directory relative-to (or/c (and/c path? complete-path?)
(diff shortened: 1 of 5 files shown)


LeifAndersen:
(As usual, I'll add tests and documentation next.)

mflatt:
Two recommendations:

 * Support the base-path/rel-to-path split that bytecode and fasl writing supports.

 * Encode relative paths as a list of path-element byte strings, so that they can be convention-independent.

And an implementation suggestion:

 * Now that `racket/fasl` (hopefullY0 handles the base-path/rel-to-path split, pull out the relevant code into a library in "private", then use it in `racket/serialize`, too.

LeifAndersen:
> Support the base-path/rel-to-path split that bytecode and fasl writing supports.

By this do you mean making them automatic? I'm happy to do that, I just added the optional argument for backwards compatibility. But I'd be happy to change that.

LeifAndersen:
Okay @mflatt I think I did your three suggestions.

LeifAndersen:
And now I've added tests+docs.

PR #2168, opened 2018-07-10
Add procedure? as an acceptable type for prop:serialize's deserialize binder.

LeifAndersen:
(Combining two features I need in the racket/serialize library.)


LeifAndersen:
Woops, the title was based on my silly branch name. Fixed.

PR #2172, opened 2018-07-16
Relative paths should still be readable.

LeifAndersen:
The resent PR to enable relative serialization resulted in serialzed
objects that weren't actually readable (containing literal path
elements). This PR converts them to bytes.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/serialize.rktl
+++ b/pkgs/racket-test-core/tests/racket/serialize.rktl
@@ -622,9 +622,14 @@
                                    "              #:" rel-mode "relative-directory"
                                    "              (find-system-path 'temp-dir))))\n"))))
       (define s (dynamic-require `(submod ,fn main) 's))
+      (define-values (in out) (make-pipe))
+      (write s out)
+      (close-output-port out)
+      (define read-s (read in))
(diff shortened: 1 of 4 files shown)


mflatt:
I think you need to use `path-element->bytes` instead of `path->bytes`.

Also, probably this conversion should be in `make-path->relative-path-elements`, instead of in both "fasl.rkt" and "serialize.rkt".

LeifAndersen:
Mmm...good call, I'll fix that.

LeifAndersen:
@mflatt Okay, fixed.

mflatt:
Unfortunately, `explode-path` can return `'up` and `'same`, and there's nothing preventing the base directory from having those elements, so you can't just map `path-element->bytes`. The code in "fasl.rkt" (which should be removed as redundant after this change), converts only elements that are paths.

LeifAndersen:
(Although I'm currently getting errors in the uni-norm tests. :/)

```
Section(uni-norm)
ERROR: regexp-match: contract violation
  expected: (or/c string? bytes? path? input-port?)
  given: #<eof>
  argument position: 2nd
  other arguments...:
   #rx"^# NormalizationTest-"
```

mflatt:
LGTM

PR #2175, opened 2018-07-19
fix random-ref error message

### LINKED ISSUE #2174: empty sequence to random-ref complains at random-sample
This contract needs to enforce the sequence passed to `random-ref` not to be empty.

https://github.com/racket/racket/blob/62f5b2c4e4cdefa18fa36275074ff9fe376ddaf3/racket/collects/racket/random.rkt#L6

At the moment:
```
> (require racket/random)
> (random-ref '())
random-sample: contract violation
;   expected: non-empty sequence for n>0
;   given: '()
```

bennn:
Fixes #2174 

For an empty sequence, print a message in terms of random-ref

Also, change all error messages to show all arguments

### DIFF
--- a/racket/collects/racket/random.rkt
+++ b/racket/collects/racket/random.rkt
@@ -20,7 +20,13 @@
                   (current-continuation-marks)))]))
 
 (define (random-ref seq [prng (current-pseudo-random-generator)])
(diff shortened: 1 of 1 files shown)


stamourv:
LGTM

PR #2177, opened 2018-07-20
cs: fix mark-table-prune

gus-massa:
* The first commit fix a small error.

* The second one is not a bug fix, and is more subtle because it has two parts. 

First, it checks if `b` is empty every iteration (this may be slower or faster, depending on how usual is that the mark-table becomes empty in the middle of that loop)

Second, it loops using the main function instead of the internal loop. IIRC this is slightly faster in the RacketVM, but I have no cue about the behavior in RacketCS.

* Also, I think that the order of the arguments is strange. I expect `(mark-table-prune a b)` to be essentially a copy of `a`  that is modified by some information that is in `b`, but the implementation uses the other order. I didn't write the commit for this change, but this function is used only once.


mflatt:
Am I misreading the code, or does your second commit call `mark-table-prune` with the arguments reversed? (If I'm reading it right, maybe best fix is to change as you suggest in the last bullet above.)

gus-massa:
Yep. Fixed, squashed and rebased.

mflatt:
LGTM, thanks!

PR #2178, opened 2018-07-20
Fix fasl bug in Racket 7.0 beta!

LeifAndersen:
The following program causes racket to error in Racket 7.

(parameterize ([current-write-relative-directory (build-path "/" "a")])
  (s-exp->fasl (build-path "/")))

This bug appears to have been introduced in Racket 7, and not in
Racket 6.x.

(You might want to consider bringing this fix into the beta for Racket 7 @stamourv !)

### DIFF
--- a/pkgs/racket-test-core/tests/racket/fasl.rktl
+++ b/pkgs/racket-test-core/tests/racket/fasl.rktl
@@ -145,17 +145,18 @@
 (test (srcloc ".../a/.." 1 2 3 4) fasl->s-exp (s-exp->fasl (srcloc (build-path "a" 'up) 1 2 3 4)))
 (test (srcloc ".../a/." 1 2 3 4) fasl->s-exp (s-exp->fasl (srcloc (build-path "a" 'same) 1 2 3 4)))
 
-(test
- (build-path "/")
- 'longer-relative
- (parameterize ([current-write-relative-directory (build-path "/" "a")])
-   (fasl->s-exp (s-exp->fasl (build-path "/")))))
+(let ([root (car (filesystem-root-list))])
+  (test
+   root
+   'longer-relative
+   (parameterize ([current-write-relative-directory (build-path root "a")])
+     (fasl->s-exp (s-exp->fasl root))))
 
-(test
- (build-path 'same)
- 'this-dir-path
- (parameterize ([current-write-relative-directory (build-path "/")]
-                [current-load-relative-directory #f])
-   (fasl->s-exp (s-exp->fasl (build-path "/" 'same)))))
+  (test
+   (build-path 'same)
+   'this-dir-path
+   (parameterize ([current-write-relative-directory root]
+                  [current-load-relative-directory #f])
+     (fasl->s-exp (s-exp->fasl (build-path root 'same))))))
 
 (report-errs)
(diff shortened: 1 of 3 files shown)


mflatt:
I agree that the `((length exploded) . >= . (length exploded-base-dir))` check is needed.

I don't think `truncating-list-tail` is needed, because the guard on `current-write-relative-directory` will make sure that `wrt-dir` extends the `base-dir`. Is there some possibility that I'm overlooking?

Then again, there's no similar check in `make-path->relative-path-elements` on its argument (which `who` is non-`#f`), so that check should be added.

We need a test that works on Windows, where "/" is not an absolute path. I recommend `(car (filesystem-root-list))` in place of `"/"`.

LeifAndersen:
WOOPS! Sorry, you are correct, that function is not needed. I put it in there when testing things out on my machine, and forgot to pull it out, I'll do it now.

LeifAndersen:
Okay @mflatt I've fixed two of the issues.

And again, @stamourv You should consider merging this into Racket 7 before release.

mflatt:
Are you also going to add the missing check, which would make sure that `serialize` doesn't fail on a bad pair of paths?

LeifAndersen:
Ah, okay. That's what you meant. (Sorry, I failed to parse that the first time around. I'll add it now.)

LeifAndersen:
And I've now pushed the new tests.

mflatt:
The test for `serialize` are good, but that's not what I was trying to get at.

I mean that
```
(current-write-relative-directory
 (cons "/x"
       "/x/y"))
```
reports an error about "/x" not being an extension of "/x/y", but
```
(serialize (string->path "/x")
           #:relative-directory (cons "/x"
                                      "/x/y"))
```
reports an error from `list-tail`.

When I added the check on `wr-dir` in `make-path->relative-path-elements`, I forgot to add a check that the first path of a pair extends the second path.

I can add the check if you prefer. I would make it an extra check after the existing one, raising an error with `raise-arguments-error` similar to the one that `current-write-relative-directory` raises.

LeifAndersen:
Oh...facepalm...okay, that makes a lot more sense. I'm adding that extra check now. Thanks.

LeifAndersen:
And it looks like all tests pass. Any other comments?

mflatt:
Wouldn't that those tests pass before the change? I think the tests will need to inspect the error message.

The change isn't quite what I had in mind, because it increases the cost unnecessarily in the common case that `base-dir` and `wrt-dir` are the same, and it checks even when `who` is `#f`. I suggest something like this:
```
(set! exploded-wrt-rel-dir
      (cond
        [(eq? base-dir wrt-dir) '()]
        [else
         (define exploded-wrt-dir (explode-path wrt-dir))
         (define base-len (length exploded-base-dir))
         (when who
           (unless (and 
                    ((length exploded-wrt-dir) . >= . base-len)
                    (for/and ([a (in-list exploded-wrt-dir)]
                              [b (in-list exploded-base-dir)])
                      (equal? a b)))
             (raise-arguments-error who
                                    "relative-directory pair's first path does not extend second path"
                                    "first path" wrt-dir
                                    "second path" base-dir)))
         (list-tail exploded-wrt-dir base-len)]))
```

LeifAndersen:
re tests: Mmm...fair. Okay, how about these tests then?

re code: Fair, that works for me.

mflatt:
LGTM

Thanks for catching the bug and making all the improvements!

LeifAndersen:
Alright then. Merging. Thanks for the feedback. :)

PR #2181, opened 2018-07-22
add stream-take + docs

odanoburu:
I believe having a `take` function for streams is very useful, and is pretty basic functionality to be included in `base`.

(this is my first non-documentation PR, so please excuse me for my ignorance and help me overcoming it :)

note: I haven't re-built and tested racket after these changes, mostly because it will take 8h+ on my computer, as per https://blog.racket-lang.org/2017/09/tutorial-contributing-to-racket.html and the fact that my computer is old.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sequences.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sequences.scrbl
@@ -1049,6 +1049,11 @@ stream, but plain lists can be used as streams, and functions such as
   the resulting stream.
 }
 
+@defproc[(stream-take [s stream?] [i exact-nonnegative-integer?])
+         stream?]{
+  Returns a list of the first @racket[i] elements of @racket[s].
+}
+
 @defproc[(stream-append [s stream?] ...)
          stream?]{
   Returns a stream that contains all elements of each stream in the
--- a/racket/collects/racket/stream.rkt
+++ b/racket/collects/racket/stream.rkt
@@ -33,6 +33,7 @@
          stream-length
          stream-ref
          stream-tail
+         stream-take
          stream-append
          stream-map
          stream-andmap
@@ -114,6 +115,14 @@
                              "stream" st)]
      [else
       (loop (sub1 n) (stream-rest s))])))
+
+(define (stream-take st i)
+  (unless (stream? st) (raise-argument-error 'stream-take "stream?" st))
+  (unless (exact-nonnegative-integer? i)
+    (raise-argument-error 'stream-take "exact-nonnegative-integer?" i))
+  (for/list ([e (in-stream st)]
+             [_i (in-range i)])
+    e))
 
 (define (stream-append . l)
   (for ([s (in-list l)])


lexi-lambda:
I like the idea of this PR (I’ve definitely wanted this function before). A few small comments:

  1. I think `stream-take` should produce a stream, not a list. In doing so, it can also be lazier, since it doesn’t have to actually realize the stream if the result is not used.

  2. If you add `stream-take`, you should probably also add `stream-drop` for symmetry.

  3. The documentation should include `@history[#:added "7.0.0.7"]` annotations.

odanoburu:
>  I think stream-take should produce a stream, not a list. In doing so, it can also be lazier, since it doesn’t have to actually realize the stream if the result is not used.

hum, that also makes sense -- in my specific use-case I needed a list, but I guess a stream makes more sense.

> If you add stream-take, you should probably also add stream-drop for symmetry.

so, there's `stream-tail`, which is guess is `stream-drop` (I would've picked that name, at least :P). is that what you mean?

> The documentation should include @history[#:added "7.0.0.7"] annotations.

indeed, thank you! but where do I find out what is the current version? (sorry, newbie here, and as I said I haven't been able to compile)

odanoburu:
now that I thought more about it, I'm more divided about if `stream-take` should return a list or a stream. it makes sense for `stream-tail` to return a stream, because its output can be infinite, and we may just want to skip the first few elements. anyone has examples of use for one or another option?

the implementation for stream output could be this one 

```racket
(define (stream-take st i)
  (unless (stream? st) (raise-argument-error 'stream-take "stream?" st))
  (unless (exact-nonnegative-integer? i)
    (raise-argument-error 'stream-take "exact-nonnegative-integer?" i))
  (if (= i 0)
      empty-stream
      (stream* (stream-first st) (stream-take (stream-rest st) (sub1 i)))))
```

it is not tail-recursive, but as it is a stream I'm not sure that is a problem -- if it is I can change it to use an accumulator or something else.

AlexKnauth:
When evaluating an element of a stream can cause a side effect or an error, it the increased laziness of `stream-take` returning a stream matters:
```racket
(define (div a b)
  (stream (not (zero? b)) (/ a b)))

;; shouldn't divide by zero, saved by laziness
(define s (stream-take (div 1 0) 2))

(cond
  ; the first element of the stream determines whether it's safe
  [(stream-first s)
   ; to look at the second element of the stream
   (stream-second s)]
  [else
    "division by zero"])

**Comment by odanoburu** [comment]
@AlexKnauth so you'd argue for returning streams? I find your example a bit artificial, but I see what you mean!

**Comment by AlexKnauth** [comment]
Yes, I'd argue for returning streams as @lexi-lambda suggested.

**Comment by odanoburu** [comment]
ok, now `stream-take` returns a stream, and added two tests too! (although I haven't tested if they succeed for the above reasons)

**Comment by AlexKnauth** [comment]
The list `take` function produces an error message when you try to take more elements from the list than exist:
```racket
(take (list 1 2 3) 89)
;take: contract violation
;  expected: a list with at least 89 elements
;  given: (list 1 2 3)
```

To be consistent the `stream-take` function should raise a (non-internal) error, but it has to be "smarter" about it; it can't just compute the length to compare it. I would recommend checking that the stream is non-empty in the else case when `(= i 0)` is false, since that's the case that requires `stream-first` and `stream-rest` to work.

AlexKnauth:
On the other hand, it might be useful to have the `stream-take` function return the stream in that case instead of erroring:
```
(stream-take (stream 1 2 3) 89)  ;=>  (stream 1 2 3)
```

But either way, the place to change is the else case when `(= i 0)` is false, by branching on `stream-empty?`.

odanoburu:
@AlexKnauth I'm sure I like the idea of throwing exceptions for this; haskell's `take` doesn't throw an exception.

but I guess this is just a different design decision, if that's what the racket community wants..

AlexKnauth:
Does anyone have any thoughts about adding `stream-drop` as well? 

What should the relationship with `stream-tail` be? Should `stream-drop` be documented as the "main" one with `stream-tail` as merely an alias?

lexi-lambda:
I think @odanoburu’s initial point that `stream-tail` already serves that purpose is correct, and I simply didn’t realize it was named that. On the other hand, it does seem unnecessarily inconsistent. I would be fine with renaming it to `stream-drop` and making `stream-tail` an alias, but I also don’t feel strongly that it’s absolutely necessary. I’ll happily defer to others’ judgement here.

A more pressing issue is that this PR appears to break some tests, as evidenced by CI failing. I have to admit that I find it startling that building Racket could take 8 hours to build on your machine! My computer is admittedly a high-end one, but it *is* still a five year old laptop, and it takes under 30 minutes to rebuild everything from scratch. Unless your computer is *really* old, some part of me wonders if something has gone wrong if it takes that long for Racket to rebuild on your machine, and if you’re planning on contributing, it seems like it might be worth figuring out.

odanoburu:
so, I finally had the time to look deeper into this. the CI build failed in only one of the 7 instances because of a timeout, so I'm considering it works. 

the build problem was being caused by a wrong test `try-basic-sequence-ops`, which is used to test both sequences in general and streams. it takes some 16 functions as arguments, and `stream-take` was blindly added to it. this made me realize that given we seem to agree that adding `stream-take` and `stream-drop` (as an alias) are good things, we should probably add `sequence-take` and `sequence-drop` too. (this way we can add those to the `try-basic-sequence-ops` test too.)
 do you agree?

@lexi-lambda I managed to build! I didn't actually try to build before, I was just guessing from the source I cited. it does take some time, as my computer is both 5-years old and on the lower-end scale, but more like 2h/3h for the first time.

bennn:
@odanoburu how did you figure 8 hours from https://blog.racket-lang.org/2017/09/tutorial-contributing-to-racket.html ?

(Maybe we should update the blog post?)

odanoburu:
@bennn 3h for make plus 4h for raco setup plus 1h bonus for old computer! :P

pkgs/racket-test-core/tests/racket/stream.rktl
```diff
@@ -67,4 +68,17 @@
...
+                     (stream-cons (/ 4 i) (loop (sub1 i))))
+                   4))
+;; stream-take preserves laziness, doesn't evaluate elements too early
+(define (guarded-second s)
+  (if (stream-ref s 0) (stream-ref s 1) #f))
+(define (div a b)
+  (stream (not (zero? b)) (/ a b)))
+(test #f guarded-second (stream-take (div 1 0) 2))
+(test 3/4 guarded-second (stream-take (div 3 4) 2))
+
```
bennn: Can you add a test for a too-large index, something like:
```
(err/rt-test (stream-take (stream 1 2) 3) exn:fail:contract? "stream-take")
```

odanoburu:
@bennn I don't know how to test the tests except by running `make`, which takes some time. how do I do it? I simply added the one you proposed and it seems to work, though.

bennn:
You can run the first test from the travisfile ([link](https://github.com/racket/racket/blob/master/.travis.yml#L33)):
```
raco test -l tests/racket/test
```

(If that takes too long, then you can comment out parts of [this file](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/all.rktl) and run the command above )

AlexKnauth:
The tricking thing about testing for an "index too large" error on `(stream-take (stream 1 2) 3)` is that the error isn't raised right away because of laziness. Maybe the expression to test to raise the error should be 
```racket
(stream->list (stream-take (stream 1 2) 3))
```
instead.

PR #2185, opened 2018-07-23
syntax-parse docs: small examples for #:with, #:attr, #:fail-when, and #:fail-unless

AlexKnauth:
This PR adds small examples for the syntax-parse directives `#:with`, `#:attr`, `#:fail-when`, and `#:fail-unless` in the documentation.

It also adds the syntax-parse-directive keywords (`#:declare`, `#:post`, `#:and`, `#:with`, `#:attr`, `#:fail-unless`, etc.) to the documentation index, so that searching for `#:with` for example takes you to the entry that describes it.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/parse/stxclasses.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/stxclasses.scrbl
@@ -325,8 +325,8 @@ The @racket[#:attr] directive is often used in syntax classes:
 
 @examples[#:eval the-eval
 (define-syntax-class ab-sum
-  [pattern (a:number b:number)
-    #:attr sum (+ (syntax-e #'a) (syntax-e #'b))])
+  (pattern (a:number b:number)
+    #:attr sum (+ (syntax-e #'a) (syntax-e #'b))))
 (syntax-parse #'(1 2)
   [x:ab-sum
    (attribute x.sum)])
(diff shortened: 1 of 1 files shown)


AlexKnauth:
At Racket Summer School I talked to a lot of people who were looking at the syntax-parse documentation and were frustrated by the lack of small examples, and also the fact that when you search for `#:with` or `#:fail-unless` on [docs.racket-lang.org](https://docs.racket-lang.org) nothing comes up. This PR is meant to solve both of those problems.

pkgs/racket-doc/syntax/scribblings/parse/stxclasses.scrbl
```diff
@@ -280,19 +288,52 @@ values such as procedures, non-prefab structures, etc---then an
...
+  [(a:number b:number)
+   #:attr sum (+ (syntax-e #'a) (syntax-e #'b))
+   (attribute sum)])
+]
+
+The @racket[#:attr] directive is often used in syntax classes:
+
+@examples[#:eval the-eval
+(define-syntax-class ab-sum
+  [pattern (a:number b:number)
```
rmculpepper: Please use a round paren before `pattern` for consistency with the rest of the docs.

AlexKnauth: Okay

PR #2188, opened 2018-07-24
Make curry properly support functions with keyword arguments

lexi-lambda:
This PR rewrites the implementation of `curry` and `curryr` from `racket/function` to properly handle functions with keyword arguments. This has come up a number of times on IRC, Slack, and the mailing list; for an example of the latter, see @dstorrs’s [recent mailing list post on the subject](https://groups.google.com/d/msg/racket-users/F8gtAia6KUY/Mdxl9TV8CwAJ).

For a while now, the documentation for `curry` has made the following claim:

> The `curry` function provides limited support for keyworded functions: only the `curry` call itself can receive keyworded arguments to be propagated eventually to *`proc`*.

As far as I can tell, this claim is simply not true. Keyword arguments do not work correctly in any capacity in the existing implementation of `curry`, even when they are provided to the initial call to `curry` itself:

```racket
Welcome to Racket v6.12.
> (define (f #:a a b) (list a b))
> ((curry f #:a 1) 2)
#<procedure:curried>
```

This PR not only fixes that immediate issue, but also adds support for keyword arguments supplied to the functions produced by `curry`. At the same time, it also makes `curry` play nice with `procedure-arity` and `procedure-keywords`, and it gives curried procedures more useful dynamic names (in the sense of `object-name`). The implementation is significantly expanded to handle these additional features, but it provides a fast path for functions without keywords to avoid any significant overhead in the most common cases.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/procedures.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/procedures.scrbl
@@ -698,25 +698,35 @@ of arguments have been accumulated, at which point the original
 (((curry list) 1 2) 3)
 (((curry list) 1) 3)
 ((((curry foldl) +) 0) '(1 2 3))
+(define foo (curry (lambda (x y z) (list x y z))))
+(foo 1 2 3)
+(((((foo) 1) 2)) 3)
 ]
 
 A function call @racket[(curry proc v ...)] is equivalent to
 @racket[((curry proc) v ...)]. In other words, @racket[curry] itself
 is curried.
 
-The @racket[curry] function provides limited support for keyworded
-functions: only the @racket[curry] call itself can receive keyworded
-arguments to be propagated eventually to @racket[proc].
-
 @mz-examples[#:eval fun-eval
   (map ((curry +) 10) '(1 2 3))
   (map (curry + 10) '(1 2 3))
   (map (compose (curry * 2) (curry + 10)) '(1 2 3))
-  (define foo (curry (lambda (x y z) (list x y z))))
-  (foo 1 2 3)
-  (((((foo) 1) 2)) 3)
-]}
+]
 
+The @racket[curry] function also supports functions with keyword arguments:
+keyword arguments will be accumulated in the same way as positional arguments
+until all required keyword arguments have been supplied.
+
+@mz-examples[#:eval fun-eval
+  (eval:no-prompt
+   (define (f #:a a #:b b #:c c)
+     (list a b c)))
+  (eval:check ((((curry f) #:a 1) #:b 2) #:c 3) (list 1 2 3))
+  (eval:check ((((curry f) #:b 1) #:c 2) #:a 3) (list 3 1 2))
+  (eval:check ((curry f #:a 1 #:c 2) #:b 3) (list 1 3 2))
+]
+
+@history[#:changed "7.0.0.7" @elem{Added support for keyword arguments.}]}
 
 @defproc*[([(curryr [proc procedure?]) procedure?]
            [(curryr [proc procedure?] [v any/c] ...+) any/c])]{
--- a/pkgs/racket-test/tests/racket/curry.rkt
+++ b/pkgs/racket-test/tests/racket/curry.rkt
@@ -30,3 +30,41 @@
 
 (check-exn exn:fail:contract? (λ () (curry 1 2)))
 (check-exn exn:fail:contract? (λ () (curry 1)))
+
+(define (kw #:a a #:b b #:c c)
+  (list a b c))
+
+(check-equal? (((curry kw #:a 1) #:b 2) #:c 3) (list 1 2 3))
+(check-equal? (((curry kw #:b 1) #:c 2) #:a 3) (list 3 1 2))
+(check-equal? (((curry kw #:c 1) #:a 2) #:b 3) (list 2 3 1))
+(check-equal? (((curry kw) #:a 1 #:b 2) #:c 3) (list 1 2 3))
+(check-equal? (((curry kw) #:b 1 #:c 2) #:a 3) (list 3 1 2))
+(check-equal? (((curry kw) #:c 1 #:a 2) #:b 3) (list 2 3 1))
+
+(check-exn exn:fail:contract? (λ () (curry kw #:d 1)))
+(check-exn exn:fail:contract? (λ () ((curry kw) #:d 1)))
+(check-exn exn:fail:contract? (λ () (curry kw 1)))
+(check-exn exn:fail:contract? (λ () ((curry kw) 1)))
+(check-exn exn:fail:contract? (λ () (((curry kw) #:a 1) #:a 2)))
+(check-exn exn:fail:contract? (λ () ((curry kw #:a 1) #:a 2)))
+
+(define (kw+pos a b #:x x #:y y)
+  (list a b x y))
+
+(check-equal? ((((curry kw+pos 1) 2) #:x 3) #:y 4) (list 1 2 3 4))
+(check-equal? ((((curry kw+pos 1) 2) #:y 3) #:x 4) (list 1 2 4 3))
+(check-equal? ((curry kw+pos 1 #:x 3) 2 #:y 4) (list 1 2 3 4))
+(check-equal? ((curry kw+pos 1 #:y 3) 2 #:x 4) (list 1 2 4 3))
+(check-equal? ((curry kw+pos 1 #:x 3 #:y 4) 2) (list 1 2 3 4))
+(check-equal? ((curry kw+pos 1 #:y 3 #:x 4) 2) (list 1 2 4 3))
+
+(check-exn exn:fail:contract? (λ () ((curry kw+pos) 1 2 3)))
+(check-exn exn:fail:contract? (λ () ((curry kw+pos) #:d 1)))
+
+(define (opt-kw+pos #:a [a #f] b)
+  (list a b))
+
+(check-pred procedure? (curry opt-kw+pos #t))
+(check-pred procedure? ((curry opt-kw+pos) #t))
+(check-equal? (curry opt-kw+pos #:a #t #t) (list #t #t))
+(check-equal? ((curry opt-kw+pos) #:a #t #t) (list #t #t))
--- a/racket/collects/racket/function.rkt
+++ b/racket/collects/racket/function.rkt
@@ -1,7 +1,7 @@
 #lang racket/base
 
 (require (for-syntax racket/base racket/list syntax/name)
-         racket/match racket/private/arity)
+         racket/list racket/match racket/private/arity)
 
 (provide identity const thunk thunk* negate curry curryr
          (all-from-out racket/private/arity)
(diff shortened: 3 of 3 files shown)


dstorrs:
lexilambda:  Here, have all the kudos.  --dstorrs

dstorrs:
Oh, more usefully:  the current implementation of curry *does* properly handle keywords as long as you are currying the 'curry' function itself.  The function that this produces can then be used to curry keywords onto an arbitrary function.  So:

Welcome to Racket v6.12.
> (define (f #:a a b) (list a b))
> ((curry f #:a 1) 2)
#<procedure:curried>

> (curry curry #:a 7)
#<procedure:curried>
> ((curry curry #:a 7) f 8)
'(7 8)

lexi-lambda:
That’s pretty bizarre, but after looking at the old implementation a little bit more, as far as I can tell, that behavior is something of an accident. This produces the same behavior:

```racket
Welcome to Racket v6.12.
> (define (f #:a a b) (list a b))
> (curry f #:a 1 2)
'(1 2)
```

…which is essentially what the program that applies `curry` to `curry` ends up doing. More generally, it looks like `curry`’s existing support for keyword arguments only works properly for functions that accept arbitrary keywords, that is, procedures created with `make-keyword-procedure`. `curry` happens to be one of these functions, so it all happens to fall out correctly when combined with the above behavior… but since the inner `curry` is given the function call fully-saturated, it isn’t really doing much currying at all.

In any case, accidental or not, the existing behavior is obviously confusing and not especially useful.

dstorrs:
> In any case, accidental or not, the existing behavior is obviously confusing and not especially useful.

No argument from me.  So, is this new functionality actually in 6.12 and available for download, or is it still waiting for merge?  I saw that the appveyor tests are failing, so I wasn't sure.

lexi-lambda:
It’s not merged yet, no. The AppVeyor tests were failing on master as well, but that’s been fixed since I opened this PR, so I’ve rebased and started a new build. Even so, I’d like some review before I merge this in, but I’m not sure exactly who is best-suited to review these changes. The most likely candidates are probably @mflatt and @elibarzilay, but there’s not much rush to merge this, since the Racket v7.0 release is imminent, and these changes won’t make it into that release.

Assuming these changes do get merged, they’ll be in Racket v7.1, which will be released in roughly three months. If you want to try them before that, though, they’ll be in the nightly snapshots as soon as this PR is merged.

mflatt:
I like these changes, but I worry about the performance of a procedure returned by `procedure-reduce-arity` and `procedure-rename`. Is there a measurable difference in speed — either for calling `curry` or for calling the function returned by `current` — for this version versus the unimproved one?

If performance turns out to matter, a typical strategy for reducing the code of `procedure-reduce-arity` is to special-case a handful of small arities. For example, `curry` might specialize 2-argument functions and 3-argument functions when given 1 or 2 initial arguments.

lexi-lambda:
Your question is a reasonable one, so I put together some benchmarks. I tested the performance of three implementations of `curry`: the current implementation, my new implementation, and my new implementation with the uses of `procedure-reduce-arity` and `procedure-rename` removed. I tested each of these implementations on the following four functions:

```racket
(define (pos a b) (void))
(define (pos-opt a b [c #f]) (void))
(define (pos-any a b . cs) (void))
(define (pos-many a b c d e f g h i j k) (void))
```

This produced the following results:



The version that uses `procedure-reduce-arity` and `procedure-rename` is obviously slower. Interestingly, the existing implementation is still faster than the new one even with those functions removed when applied to functions with small numbers of arguments but slows as the number of arguments increases. This hints at different asymptotics due to different choices about how to accumulate arguments, but given that small number of arguments are common, perhaps the new version should be tweaked to use the old strategy for the lower constant factor.

However, also look at the units: these differences are a few µs. While it’s true that there’s a relative difference—sometimes a startling relative difference—the absolute difference is small. I don’t know how much this is likely to matter in practice.

Personally, I like the niceness given by the slower `curry` that uses `procedure-reduce-arity` and `procedure-rename`, since I already avoid `curry` when performance is relevant, anyway. That said, I’d understand not wanting to eat this additional cost. If you think it’s too much, I can update this PR to keep the keyword support and drop the uses of `procedure-reduce-arity` and `procedure-rename` to get the performance on par with the existing implementation.

bennn:
Can you try the special cases mflatt suggested? (Maybe using a macro to produce the special-cased code.)

lexi-lambda:
Oops, sorry—I meant to address that point in my last comment but forgot. To be honest, I don’t understand exactly what Matthew is suggesting. Most of the situations in which I’ve seen such manual specialization performed have been simple situations in which fixed-arity cases are provided for functions of arbitrary arity, but `curry` is more complicated than, say, `map` or `foldl`. (More generally, I am a little disappointed the JIT can’t make such manual specialization unnecessary, but I understand that the Racket JIT isn’t currently particularly fancy.)

gfbee:
I think something like this is what Matthew's suggesting:
```
(case-lambda
      [(f) (do-curry f)]
      [(f a) (define-values (kws ~kws) (procedure-keywords f))
             (if (and (null? kws) (null? ~kws))
                 (match (procedure-arity f)
                   [0 (λ () (f a))]
                   [1 (f a)]
                   [2 (λ (b) (f a b))]
                   [(arity-at-least (or 0 1)) (case-lambda [() (f a)]
                                                           [(b) (f a b)]
                                                           [args (apply f a args)])]
                   [(arity-at-least 2) (letrec ([curried (case-lambda [() curried]
                                                                      [(b) (f a b)]
                                                                      [args (apply f a args)])])
                                         curried)]
                   [_ ((do-curry f) a)]) ((do-curry f) a))]
      [(f . args) (apply (do-curry f) args)])
```

lexi-lambda:
After thinking on this some more, I have updated my set of benchmarks, which I think paint an interesting picture:



The benchmarked functions are still the same, but the implementations now refer to the following:

  - `old` is the existing implementation.
  - `new` is the implementation in this PR.
  - `new/no-rename` is like `new`, but it does not use `procedure-rename`.
  - `new/no-reduce` is like `new`, but it does not use `procedure-reduce-arity`.
  - `new/neither` is like `new`, but it does not use `procedure-rename` *or* `procedure-reduce-arity`.
  - `new/custom-reduce` is like `new`, but it uses a custom implementation of `procedure-reduce-arity` written in Racket that raises error messages on invalid arities, but does not actually affect the result of `procedure-arity`.
  - `new/custom-reduce+no-rename` is like `new/custom-reduce`, but it does not use `procedure-rename`.
  - `new/optimized-reduce+no-rename` is like `new/custom-reduce+no-rename`, but it adds some special cases to the custom implementation of `procedure-rename` for small arities.

This is the custom implementation of `procedure-reduce-arity` used by `new/custom-reduce` and `new/custom-reduce+no-rename`:

```racket
(define (num-args-matches-arity? num-args arity)
  (cond
    [(exact-integer? arity)
     (= num-args arity)]
    [(arity-at-least? arity)
     (>= num-args (arity-at-least-value arity))]
    [(list? arity)
     (ormap (λ (arity) (num-args-matches-arity? num-args arity)) arity)]))

(define (procedure-reduce-arity proc arity)
  (lambda args
    (unless (num-args-matches-arity? (length args) arity)
      (error (object-name proc) "wrong number of arguments"))
    (apply proc args)))
```

Some takeaways:

  1. `procedure-rename` is pretty slow.

  2. `procedure-reduce-arity` is *really* slow, to the point of seeming absurd. I realize my Racket reimplementation of it isn’t truly replicating its behavior, since it isn’t actually changing the result of `procedure-arity`, but given I’m introducing a wrapper anyway, that doesn’t seem like it should actually change things.

  3. Without `procedure-rename` and using a custom version of `procedure-reduce-arity`, execution time is 2-4x the version without either of them.

  4. Providing some naïve special cases to `procedure-reduce-arity` had no effect whatsoever on performance.

I think my question at this point has to be to ask why `procedure-reduce-arity` is so slow. I can’t come up with a reason why a runtime primitive would induce significantly more overhead than a mostly-equivalent function defined in Racket instead of in C. @florence suggested the issue might stem different behavior in the JIT, so I reran the benchmarks with the JIT disabled, which produced the following:



Interestingly, now the version with a custom implementation of `procedure-reduce-arity` is on part with the one that uses the real implementation! Which makes me wonder if the real problem here is that the JIT isn’t handling `procedure-reduce-arity` as well as it should, but that’s outside my realm of understanding.

mflatt:
@lexi-lambda Can you post your benchmark code?

I'm wondering whether you've written things in a way that allows the optimizer to see through your `procedure-reduce-arity` by inlining; it won't see through the built-in version that way. In my experiments, the result of the built-in `procedure-reduce-arity` is about twice as fast as your Racket-implemented one, instead of the other way around.

Meanwhile, yes, the result of `procedure-rename` is slow, and it could probably be improved. (It turns out to be easy to improve in Racket-on-Chez, since I've added a relevant primitive to Chez Scheme.)

lexi-lambda:
Sure. I mostly haven’t posted it because the benchmarking code is terrible and just involves a ton of messy copy/paste, but [I’ve uploaded it here if it helps](https://gist.github.com/lexi-lambda/6235c1799e3348b3d8cbfddfe8c44bfe). I did just try adding `(set! procedure-reduce-arity procedure-reduce-arity)` to see if that would change the results, but it didn’t seem to. I wouldn’t be terribly surprised if I just didn’t understand what was happening, though.

lexi-lambda:
One more small, albeit probably not very meaningful, data point: I decided to run the same benchmarks on RacketCS just to see what would happen. I got the following results:



These seem to indicate that `procedure-reduce-arity` is significantly slower on RacketCS, but I wouldn’t be surprised if that just hadn’t been optimized much.

mflatt:
Thanks! Mostly, I misunderstood what you were measuring. I was worried about the performance of the function returned by `procedure-reduce-arity` and `procedure-rename`, but I see that you're measuring the time of calling those functions. I'll investigate more.

mflatt:
I think part of the way forward here involves switching to the arity-mask representation that Kent Dybvig suggested for Chez Scheme. An arity mask is an exact integer that has a 1 for bit N in it's two's complement representation if N arguments are allowed. An `arity-at-least` structure turns into a negative number as a mask, and `procedure-arity-includes?` is a small wrapper on `bitwise-bit-set?`.

Also, since `procedure-rename` plus `procedure-reduce-arity` is a common combination, and since they do something similar internally, then it helps a little to make `procedure-reduce-arity[-mask]` take an optional name argument.

Implementation:

https://gist.github.com/mflatt/379d032487e1951e4470263d87415b1d

Here's a preliminary performance result. To make a more clear comparison to curry implementation that uses the non-mask representation, I added a `new/merge` that passes the new name to `procedure-reduce-arity`, but it performs in between `new` and `new/no-rename`, as you'd expect.

image

Here's RacketCS after some improvements (e.g., `object-name` turned out to be disasterously slow), where the profile is similar, but everything is faster:

image

I'll finish cleaning up the implementation of arity masks and push later today, but I've pushed a preliminary version at mflatt/racket@3664c523970af03635b8227e566f97ceb6206f0b.

lexi-lambda:
Neat! Thank you for looking into this so promptly. Do you think that a version that uses both arity masks and the merging of rename and reduce-arity will be acceptably fast to replace the existing implementation of `curry`? Your benchmark results indicate it still takes 2-3x the time as the old implementation, but I feel like it’s impossible to know if that has any actual impact or not without studying real usage patterns of `curry` rather than exclusively looking at microbenchmarks.

mflatt:
Yes, I think the cost is reasonable given the benefits.

lexi-lambda:
Many thanks for your help!

PR #2192, opened 2018-07-27
Fix typo where stream-empty? should be set-empty?

pmatos:
As discussed in https://groups.google.com/d/msg/racket-users/gpbEXO6KFek/EEXKpXhlCAAJ, Alex noticed `stream-empty?` should be `set-empty?`.


pmatos:
Sorry, Alex has proposed a pull request already. Disregard mine.

PR #2197, opened 2018-07-28
fix a broken pattern-expander example

yjqww6:
The original example using syntax-rules would produce
```
~optional: cannot use identifier tainted by macro transformation in: ~optional
```


AlexKnauth:
This looks good to me. Although it might be worth raising an issue after this pull request about choosing between one of these options:
  1. Documenting the fact that pattern expanders shouldn't use `syntax-arm` or `syntax-protect` (and so shouldn't use `syntax-rules` either).
  2. Changing `syntax-parse` so that it can disarm identifiers, making `syntax-protect` okay in pattern expanders, and so make `syntax-rules` okay too.

PR #2200, opened 2018-07-29
doc: stx-trans.scrbl edits and examples

bennn:
- `local-expand` : replace ambiguous comma with an english word, to clarify
- `syntax-local-value` : replace '@emph{Unlike}` with `Unlike`, because the
   extra emphasis was overall distracting to me
- `syntax-local-value/immediate` : add example
- `syntax-local-introduce` : add example
- replace a few `and/c` with `syntax/c`

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
@@ -262,7 +262,7 @@ The @racket[stop-ids] argument controls how far @racket[local-expand] expands @r
 
  @item{If @racket[stop-ids] is any other list, then expansion proceeds recursively until the expander
        encounters any of the forms in @racket[stop-ids], and the result is the partially-expanded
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
```diff
@@ -262,7 +262,7 @@ The @racket[stop-ids] argument controls how far @racket[local-expand] expands @r
 
  @item{If @racket[stop-ids] is any other list, then expansion proceeds recursively until the expander
        encounters any of the forms in @racket[stop-ids], and the result is the partially-expanded
-       form. If @racket[extend-stop-ids?] is not @racket[#f], @racket[begin], @racket[quote],
+       form. If @racket[extend-stop-ids?] is not @racket[#f] then @racket[begin], @racket[quote],
```
lexi-lambda: I agree that an explicit use of “then” here is clearer, but it’s still grammatically necessary to include the comma. Please use both.

PR #2203, opened 2018-07-31
Fix default exercise and generate procedure for contracts

shhyou:
This pull request fixes the default values in commit ffc5720 and the default generate procedure in the internal `contract-struct-generate` function.

- In `make[-*]contract`, the default procedures for generation and exercise should not take the `ctc` argument. They should be:
    * For `#:generate`, `(λ (fuel) #f)` instead of `(λ (ctc) (λ () #f))`.
    * For `#:exercise`, `(λ (fuel) (values void '()))` instead of `(λ (ctc) (λ (fuel) (values void '())))`.
- In `build[-*]-contract-property`, the default generation procedure should be `(λ (ctc) (λ (fuel) #f))`. For `build-chaperone-contract-property`, this is further fixed by changing the default return value in `contract-struct-generate`.

The old test case
```
(check-false
 (contract-random-generate
  (make-chaperone-contract
   #:late-neg-projection
   (λ (b) (λ (f v np) v)))))
```
is fixed since `contract-random-generate` should have failed to generate a value instead of returning `#f`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -2745,11 +2745,11 @@ returns @racket[#f] but @racket[value-blame] returns @racket[#f].
           [#:generate
            generate
            (->i ([c contract?])
-                ([generator
-                  (c)
-                  (-> (and/c positive? real?)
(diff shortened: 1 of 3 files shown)


rfindler:
Thank you!

rfindler:
(what a mess I made)

PR #2204, opened 2018-08-02
improve `hash-ref` error message for non-thunk failure procedure

bennn:
Improve the `hash-ref` error message when the failure result does not
accept zero arguments. (This only changes what the error messages says.)

Example:
```(hash-ref #hash() 'a add1)```

Old message:
```
; add1: arity mismatch;
;  the expected number of arguments does not match the given number
;   expected: 1
;   given: 0
```

New message:
```
; hash-ref: contract violation
;   expected: (-> any)
;   given: #<procedure:add1>
;   argument position: 3rd
```

### DIFF
--- a/pkgs/racket-test-core/tests/racket/basic.rktl
+++ b/pkgs/racket-test-core/tests/racket/basic.rktl
@@ -2191,7 +2191,7 @@
       (test #t eq? (equal-hash-code l) (equal-hash-code (list 1 2 3)))
       (hash-set! h1 l 'ok)
       (test 'ok hash-ref h1 l)
-      (err/rt-test (hash-ref h1 'nonesuch (lambda (x) 'bad-proc)) exn:fail:contract:arity?)
+      (err/rt-test (hash-ref h1 'nonesuch (lambda (x) 'bad-proc)) exn:fail:contract? "hash-ref")
       (test #t hash-has-key? h1 l)
       (test #f hash-has-key? h1 (cdr l))
       (when hash-ref!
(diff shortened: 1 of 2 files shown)


mflatt:
@bennn I missed this change until now, noticing only because RacketCS is out-of-sync with the test.

Improving the error is a good idea, but I'm uncomfortable with reporting the error as a contract error from `hash-ref`. A function's argument contracts should be checked before anything else happens, but this error is reported only if the argument is used (after the key is not found).

I propose changing the exception back to `exn:fail:contract:arity`, but with a specialized error message that clariifies that it's via `hash-ref`. Ok if I make that change?

samth:
Could we move this check earlier? That might have backwards compatibility problems.

Also, I'm not sure why delayed contract checks seems wrong -- delayed contract checks on functions are pretty common in the contract system. :) It's certainly possible to write a contract combinator that implements this behavior.

mflatt:
Yes, moving the check earlier would change behavior.

The error should be a contract error, but it should not be characterized as the contract `(-> any)` for an argument of `hash-ref`. I'd also like it to go back to the `exn:fail:contract:arity` exception subtype.

bennn:
@mflatt yes that change is ok with me,

but I also don't understand why its bad to report this delayed check as a contract error from `hash-ref`. The list library (for example) has delayed contract errors that look similar to me: [take,drop, etc.](https://github.com/racket/racket/blob/master/racket/collects/racket/list.rkt#L145)

bennn:
Ok, take & drop were bad examples.

I'm still confused though. Here's where I think we are:
1. `hash-ref` specifies a first-order contract on its 3rd argument
2. it's bad to apply that first-order contract right away in `hash-ref` because of backwards compatibility
3. if that contract fails, its bad to report it as what it is (the 3rd argument to hash-ref didn't meet its contract) because `hash-ref` didn't check the contract right away (because step 2)

And I either don't understand or don't agree with the first "because" in step 3

(It's still ok with me if you make the change you want, I'm just trying to clarify why I'm confused.)

PR #2210, opened 2018-08-04
Exercise function application result in arrow contracts

shhyou:
Exercise function application result in `->` contracts. If the range contract is not flat, this exercise would be useful. For the `->i` contract, this pull request only exercise the result if the range contract is non-dependent because the exercise procedure demands the contract be known in advance.


rfindler:
Looks great, thanks!

PR #2217, opened 2018-08-06
doc: index the character constants

bennn:
Add things like `#\newline` to the index, to help people like me find them without having to remember they're documented on the reader's page

Other things in this file were indexed, but these constants weren't. I'm worried that means there was a reason to _not_ put `#\newline` in the index --- so I'll wait for a comment or +1 before merging this.


mflatt:
I can't recall a reason — seems like a good change to me.

Issue #2164, opened 2018-07-07
syntax-parse: "required attribute not defined" when using state in a pattern-expander


AlexKnauth:
When a pattern expander uses mutable state to generate an identifier to insert into the pattern, and this pattern expander is used in a syntax class, it gives this error:
```
pattern: required attribute is not defined: tmp0
```

It seems to be the interaction of all three of these: state, `pattern-expander`, and `define-syntax-class`. When I write the same program but taking out one of these, it works fine.

```racket
#lang racket
(require syntax/parse
         (for-syntax racket/syntax syntax/parse))

(define-syntax ~temporary
  (let ([counter 0])
    (pattern-expander
     (syntax-parser
       [(_)
        #:with tmp (begin0 (format-id #'here "tmp~a" counter)
                           (set! counter (add1 counter)))
        #'tmp]))))

(define-syntax-class my-class
  [pattern {~temporary}])
;pattern: required attribute is not defined: tmp0 in: (pattern (~temporary))
```

P.S. The original version of this used `generate-temporary` instead of a local incrementing counter. I was using a temporary to try to prevent the pattern expander from binding any *externally visible* attributes, while allowing the code it produces to have access to the syntax object it's being matched against.
```racket
(define-syntax ~temporary
  (pattern-expander
   (syntax-parser
     [(_)
      #:with tmp (generate-temporary 'tmp)
      #'tmp])))

(define-syntax-class my-class
  [pattern {~temporary}])
;pattern: required attribute is not defined: tmp2 in: (pattern (~temporary))
```

AlexKnauth:
This might be because `define-syntax-class` copies the patterns *before* expanding them, and so ends up expanding them twice. In the code below, it expands the `rhss` once on line 56, but copies them where they will be expanded a second time on line 74.

It would probably fix this if `define-syntax-class` could expand the `rhss` before copying it to two different places.
<img width="960" alt="stx-class-copies-code" src="https://user-images.githubusercontent.com/6600123/42412426-3f693626-81da-11e8-9ba4-f714d290417f.png">

rmculpepper:
Yes, that sounds right. BTW, the reason for the copying is that attributes must be collected (if there is no `#:attributes` clause, at least) during pass 1 so they can be part of the information statically available to syntax-class references; but other syntax classes might not be available until pass 2 (for forward references).

It also seems reasonable to exclude uninterned (and unreadable) symbols from the list of inferred attributes.

PR #2220, opened 2018-08-08
syntax/parse: only parse syntax-class patterns once

rmculpepper:
See #2164.

### DIFF
--- a/pkgs/racket-test/tests/stxparse/test.rkt
+++ b/pkgs/racket-test/tests/stxparse/test.rkt
@@ -932,3 +932,21 @@
 (terx #s(point 1)
       #s(point a b)
       #rx"expected more terms starting with any term")
+
+;; from Alex Knauth, issue #2164 (7/2018)
+(let ()
+  (define-syntax ~temp1
+    (pattern-expander
+     (lambda (stx)
+       (car (generate-temporaries '(tmp))))))
+  (define-syntax-class c1 (pattern (~temp1)))
+  (void))
+(let ()
+  (define-syntax ~temp2
+    (let ([counter 0])
+      (pattern-expander
+       (lambda (stx)
+         (begin0 (format-id #'here "tmp~a" counter)
+           (set! counter (add1 counter)))))))
+  (define-syntax-class c2 (pattern (~temp2)))
+  (void))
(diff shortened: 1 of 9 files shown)


AlexKnauth:
I'm getting a bunch of errors that look like this:
```racket
?: bad syntax
  in: #s(pat:fixup k:id k k id #s(arguments () () ()) #f #f)
  compilation context...:
   .../turnstile/turnstile.rkt
   .../turnstile/lang.rkt
  context...:
   do-raise-syntax-error
   apply-transformer-in-context
   apply-transformer52
   dispatch-transformer41
   loop
   for-loop
   finish-bodys
   for-loop
   finish-bodys
   do-local-expand56
   syntax-local-expand-expression33
   .../racket/private/stxparam.rkt:62:2: let-local-keys
   apply-transformer-in-context
   apply-transformer52
   dispatch-transformer41
   for-loop
   ...
```

Commenting here because this PR introduced the "fixup" patterns.

AlexKnauth:
Here's a minimal example:
```racket
#lang racket/base
(require syntax/parse)

(define-splicing-syntax-class props
  [pattern (~and (~seq stuff ...) (~seq k:id ...))])
```

AlexKnauth:
The `[(pat:fixup _ _ _ _ _ _) #t]` clause on line 355 of `rep-patterns.rkt` has only six fields, where the struct is supposed to have seven.

AlexKnauth:
The `?: bad syntax` error message looks like it's coming from a `syntax-case` expression.

The `syntax-case` expression on lines 540-717 of `parse.rkt` doesn't have clauses for `pat:fixup` or `pat:and/fixup`.
https://github.com/racket/racket/blob/master/racket/collects/syntax/parse/private/parse.rkt#L540-L717

The other `syntax-case` expressions in that file don't have clauses for that either. But since this error doesn't show up all the time, maybe the `fixup` patterns are supposed to be gone by then? Is it possible that the `(~and (~seq stuff ...) ...pat...)` is messing up the pass where it gets rid of the `fixup`s?

rmculpepper:
@AlexKnauth, should be fixed now, thanks. Yes, the fixup patterns are supposed to be gone by the time parse:S gets involved.

AlexKnauth:
Thank you!

PR #2230, opened 2018-08-16
rackunit/docs-complete: accept any `module-path?`

LiberalArtist:
The contract on `check-docs` was unnecessarily restricted to `symbol?`
module paths, which prevents using `check-docs` with submodules.


rfindler:
Thanks!

PR #2231, opened 2018-08-16
More consistently report problems with non-linear patterns.

### LINKED ISSUE #1386: `match` generates unhelpful log message about non-linear patterns
In one of my programs I get the log message "non-linear pattern used in `match` with ... at" (which seems to be logged with `log-error`).

The docs don't mention anything about non-linear patterns, so I'm not sure what I should do about it (or if I should care about it at all.)

### LINKED ISSUE #2152: match syntax error when pattern identifier appears twice with ellipses the second time
In a recent pre-release for Racket 7, and in Racket 6.12, minimal example:
```
(match 0 [`(,t ,t ...) 0])
t10: unbound identifier; also, no #%top syntax transformer is bound in: t10
```
Removing the ellipses, or using ellipses with both, or putting the ellipses usage first, are all fine.

samth:
Also document limitation, and suggest a fix.

Closes #1386 Closes #2152

### DIFF
--- a/racket/collects/racket/match/compiler.rkt
+++ b/racket/collects/racket/match/compiler.rkt
@@ -14,6 +14,8 @@
 ;; should we reorder stuff?
 (define can-reorder? (make-parameter #t))
 
+(define inside-GSeq? (make-parameter #f))
+
 ;; for non-linear patterns
 (define vars-seen (make-parameter null))
 
@@ -175,7 +177,7 @@
                      (and (bound-identifier=? v v*) (or id (list v v*)))))
                  =>
                  (lambda (id)
-                   (if (identifier? id)
+                   (if (and (identifier? id) (not (inside-GSeq?)))
                        (make-Row ps
                                  #`(if ((match-equality-test) #,x #,id)
                                        #,(Row-rhs row)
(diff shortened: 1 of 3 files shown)


lexi-lambda:
I like these documentation changes. The code changes seem to have broken a test, however.

samth:
Yes, the code changes are a bit too aggressive. I have to figure out how to make it less so, but it's not totally obvious.

racket/collects/racket/match/compiler.rkt
```diff
@@ -175,16 +177,21 @@
                      (and (bound-identifier=? v v*) (or id (list v v*)))))
                  =>
                  (lambda (id)
-                   (if (identifier? id)
+                   (if (and (identifier? id) (not (inside-GSeq?)))
```
AlexKnauth, line 180: This `(not (inside-GSeq?))` check also doesn't allow nonlinear patterns inside the "tail" of `GSeq` patterns, such as in `(list a _ ... a)` and `(cons a (list pre ... a post ...))`.

That seems too restrictive to me. In my opinion it should be restricted based on the pattern itself being under ellipses, not based on it being in the tail after something else that's under ellipses.

samth: I agree. This PR is too aggressive as I said at the time, but I didn't settle on a better fix.

PR #2234, opened 2018-08-16
Make cond track disappeared-uses for => and else

lexi-lambda:
For a long time, `cond` has been a misbehaved macro that has not tracked disappeared uses properly. This changes `cond` so that it does.


mflatt:
This looks good, although a test case (probably in "syntax.rktl") would make it better.

lexi-lambda:
@mflatt Updated with a test case.

mflatt:
LGTM

PR #2236, opened 2018-08-17
Update unixstyle-install.rkt

bdeket:
When installing from a racket snapshot installation failed (for me) because mred-temp-for-install does not exist before it is being removed. Checking if the temporary file exists solves this.


mflatt:
Thanks for the patch! I'm going to try a different repair, which more directly addresses the mistake in a change that I made recently, but I didn't understand the problem before seeing your patch.

bdeket:
Great. Thanks.
I guessed I was only treating symptoms. Good that you could cure the cause.

Issue #2212, opened 2018-08-05
questions about local-apply-transformer


yjqww6:
I found in racket `'expression` is always used as the context argument of local-apply-tranformer, but, if I understand correctly `'expression` and `'top-level` is excluded from generating use-site scopes in racket 7.
(from https://github.com/racket/racket/blob/41fd4f3a5e0c6d4ca0ef822b168bf33a5c9068a6/racket/src/expander/expand/definition-context.rkt#L245-L250 )
This program(derived from #2069 ):
```racket
#lang racket

(module a racket
  (require (for-syntax racket/syntax))
  (provide use-expander)
  (define-syntax (use-expander stx)
    (syntax-case stx ()
      [(_ id arg)
       (let* ([intdef-ctx (syntax-local-make-definition-context #f #f)]
              [transformer-id (internal-definition-context-introduce
                               intdef-ctx
                               (generate-temporary 'a))])
         (syntax-local-bind-syntaxes
          (list transformer-id)
          #`(quote #,(syntax-local-value #'id))
          intdef-ctx)
         (local-expand #`(#,transformer-id arg) 'expression '() (list intdef-ctx)))])))

(require 'a)

(define x 1)

(define-syntax (expander1 stx)
  (syntax-case stx ()
    [(_ an-x)
     (begin
       (displayln (hash-ref (syntax-debug-info #'an-x) 'context))
       #'(let ([an-x 3])
           x))]))

(use-expander expander1 x)
```
produce
```racket
(#(60778 module) #(60785 module anonymous-module) #(61088 macro))
3
```

but in 6.12, there will be a use-site scope

```racket
(#(110355 module) #(110356 module anonymous-module 0) #(111968 macro) #(111969 use-site))
1
```
I'm not clear why the behavior is changed in the new expander. Is it intended?
 If so, how to use local-apply-transformer or local-expand with internal-definition-context properly? Is `generate-expand-context` sufficient?

Issue #2069, opened 2018-05-10
strange behavior of define-match-expander


Syntacticlosure:
when i run this code below , it returns 2 as i expect
(define x sub1)
(define-match-expander succ
  (lambda (stx)
    (syntax-case stx ()
      [(_ v)  #'(app (let ([v 100]) x) v)])))

(let () (match 3
  [(succ x) x]))

but if i remove the outside `let` , an error happened : 

(define x sub1)
(define-match-expander succ
  (lambda (stx)
    (syntax-case stx ()
      [(_ v)  #'(app (let ([v 100]) x) v)])))

(match 3
  [(succ x) x])

application: not a procedure;
 expected a procedure that can be applied to arguments
  given: 100
  arguments...:

yjqww6:
I believe the scope-flipping trick is insufficient here.
how to apply syntax-local-introduce without touch use-site scope?
how to detect whether a identifier is defined in same definition context?
and even more, how to emulate use-site scopes by hand?

yjqww6:
@Syntacticlosure and I discussed a lot before creating an issue here, and it seems that #2082 does it better than we expected.

yjqww6:
After tried the latest repo, I found
```racket
#lang racket

(module a racket
  (provide use-expander)
  (require (for-syntax syntax/apply-transformer))
  
  (define-syntax (use-expander stx)
    (syntax-case stx ()
      [(_ an-x id in)
       #`(let ([x '(2)])
           #,(local-apply-transformer (syntax-local-value #'id) #'(an-x in)
                                      (syntax-local-context)))])))

(require 'a)
(define x '(1))

(define-syntax (expander1 stx)
  (syntax-case stx ()
    [(an-x id) #'(let ([an-x '(3)])
                   (id x))]))
(use-expander x expander1 car)
```
returns 1, and
```racket
#lang racket

(module a racket
  (provide use-expander)
  (require (for-syntax syntax/apply-transformer))
  
  (define-syntax (use-expander stx)
    (syntax-case stx ()
      [(_ an-x id in)
       #`(let ([x '(2)])
           #,(local-apply-transformer (syntax-local-value #'id) #'(an-x in)
                                      'expression))])))

(require 'a)
(define x '(1))

(define-syntax (expander1 stx)
  (syntax-case stx ()
    [(an-x id) #'(let ([an-x '(3)])
                   (id x))]))
(use-expander x expander1 car)
```
returns 3. 
The problem within match-expander maybe similar to the latter.

PR #2237, opened 2018-08-20
Fix for use-site scopes with local-expand ; failing test for letrec-syntaxes use-site scopes

michaelballantyne:
This PR aims to fix a bug with use-site scopes when `local-expand` is used with the `'expression` context. My fix also required relaxing the contract of `syntax-local-identifier-as-binding` to accept a syntax object, rather than just an identifier.

While working on this issue I also discovered a bug in use-site scope handling for `letrec-syntax(es+values)`. The second commit adds a failing test for that bug, but not a fix.

## Use-site scopes and local-expand 'expression

The current implementation does not apply use-site scopes for local expansions in `'expression` context. It does, however, apply them for expansions in Racket core form expression contexts. Because most macros using definition contexts eventually place the expressions in core form expression positions, the problem isn't usually noticed. It causes problems for macros that emulate core forms, like a [`my-letrec-syntax`](https://github.com/michaelballantyne/racket/blob/4b9bd26e76a99e817e58841d9f3bcf1f10aaa98f/pkgs/racket-test-core/tests/racket/macro.rktl#L2288-L2294) macro that binds a transformer and expands its body in 'expression context.

Use-site scopes created by nested local-expansions in a parent context created by `local-expand` `'expression` are also not recorded in the parent context. This causes problems for definition context macros if they are [used directly within](https://github.com/michaelballantyne/racket/blob/4b9bd26e76a99e817e58841d9f3bcf1f10aaa98f/pkgs/racket-test-core/tests/racket/macro.rktl#L2321) a `local-expand` `'expression` context, but not when used in a Racket core form expression context.

## `syntax-local-identifier-as-binding`

I found that fixing the above broke Alexis's use of `local-apply-transformer` in match expanders, syntax-parse pattern expanders, and require transformers. My fix required applying `syntax-local-identifier-as-binding` to the entire returned syntax instead of individual binders, which makes sense because patterns and require forms don't have local bindings, so I relaxed the contract to accept syntax rather than identifiers.

## `letrec-syntaxes` bug

The optimization in [`maybe-add-use-site-scope`](https://github.com/racket/racket/blob/master/racket/src/expander/expand/main.rkt#L446-L458) only adds a use-site scope if the frame id of the macro binding and the current frame id are the same. This is incorrect in the `letrec-syntaxes` case, because the definition context body is given a different frame id from the binding list, but does not have any distinguishing scope that makes the use-site scope unnecessary. I'm not sure of the right way to fix the optimization.


michaelballantyne:
@mflatt

lexi-lambda:
I’m taking a look at this now, and I’m trying to wrap my head around it (and not really succeeding; my intuition for use-site scopes is not great). I don’t think I can give substantive feedback without making a larger effort to understand the problem, but two immediate things jump out at me:

  1. It seems like all the uses of `local-apply-transformer` now need to be paired with `syntax-local-identifier-as-binding`. I don’t immediately understand why, but even if I did, it worries me slightly. The point of `local-apply-transformer` is essentially to allow for the implementation of things like match expanders, `syntax-parse` pattern expanders, and require/provide transformers, *reusing* the hygienic expansion mechanism already in place.

     Maybe it’s reasonable to expect of people who wish to do that sort of thing to understand use-site scopes and when to prune them. I could be convinced of that. But *I* don’t understand use-site scopes, at least not to the degree needed to reason about the issues in this pull request, and while perhaps it’s a little presumptuous of me, I feel like if it’s too confusing for me to immediately grasp, it’s likely too confusing to expect of most users.

     Perhaps a useful question to ask is this one: would you ever want to use `local-apply-transformer` *without* `syntax-local-identifier-as-binding`? My immediate intuition was that maybe using `syntax-local-identifier-as-binding` makes sense for binding forms like match expanders, pattern expanders, and require transformers, but provide transformers are morally use sites, not binding sites. If we still have to prune use-site scopes there, then when *wouldn’t* we prune them?

  2. Is generalizing `syntax-local-identifier-as-binding` to arbitrary syntax objects really the right thing to do? In the case of match expanders and pattern expanders, it seems like the correct thing would be to apply `syntax-local-identifier-as-binding` to the identifiers that end up in binding positions, since otherwise I fear there’s a possibility of over-pruning (though again, I only sort of understand use-site scopes, so this is based only on my vague intuition). Maybe it doesn’t actually cause problems, but could it if, say, a pattern includes an expression with internal definitions inside of it? That hammer seems too heavy to me.

     Admittedly, the case of `require` seems trickier, I agree, but it’s not actually hopeless, since I believe all uses of require transformers eventually expand into a giant set of `rename` clauses. It would be possible to apply `syntax-local-identifier-as-binding` to the `local-id` field of the produced `import` struct to strip the scopes only from the binding location and nowhere else.

     That just leaves provide transformers, but as I’ve already stated in the above point, I don’t understand why `syntax-local-identifier-is-binding` is necessary for provide transformers to begin with.

michaelballantyne:
Thanks!

An example to help consider when use-site scopes are necessary:

```
(let ([x 'outer])
  (define-syntax-rule (m arg)
    (begin
      (define arg 'middle)
      (let ([arg 'inner])
        x)))
  (m x))
```

`x` should refer to `'middle`, because the define is spliced into the surrounding context, but the `let` binding is local. Thus the use-site scope for the expansion of `m` is should be present on binding identifiers that bind locally, like that in `let`, but should be removed from binders that splice into the surrounding context. And when there are local binders, it is important that the use-site scope remain on references such that the binders can be referred to.

I don't think that any of match patterns, syntax parse patterns, require forms, or provide forms have local binding forms. (Please correct me if I'm wrong!) All bindings splice out into the surrounding context. I think this means that use-site scopes happen to be unnecessary in these cases and can be removed from all binders and all references. It's necessary to remove them at least from binders in pattern and require transformers, but doesn't matter either way for provide transformers or for references in any of the forms. Removing them from the whole syntax is much easier than digging for the binders.

When using `local-apply-transformer` to implement transformation for a language that does have local binders, we would have to be careful to only apply `identifier-as-binding` to binders that splice out via forms like `define`, and in the right definition context.

lexi-lambda:
Yes, I think your explanation mostly makes sense to me. I believe I do understand the purpose of use-site scopes, and I think I have a reasonable intuition for them for pattern-based macros. My intuition isn’t quite there yet for procedural macros that use `local-expand` and first-class definition contexts, but I’m getting there.

I was going to mention `?` and `app` from `match` and `~do` and `~parse` from `syntax-parse`, but Alex beat me to it. I am not immediately sure that those would cause problems, but they do mean a `match` or `syntax-parse` pattern can contain truly arbitrary Racket code, and everything that entails.

(`require` is more complicated since it can also contain arbitrary Racket code, but that code is usually phase 1 code like in Alex’s `filtered-in` example. It would be possible for non-phase 1 code to get lifted out, but that wouldn’t get affected by the use-site scope removal, anyway.)

michaelballantyne:
I had overlooked the embedding of Racket code in `?`, etc. Thanks for catching that. I think it is a problem.

AlexKnauth:
Your "An example to help consider when use-site scopes are necessary", modified to use `~do`:

```racket
(let ([x 'outer])
  (define-syntax ~m
    (pattern-expander
     (syntax-parser
       [(_)
        #'(~do
           (define-syntax-rule (m arg)
             (begin
               (define arg 'middle)
               (let ([arg 'inner])
                 (println x))))  ; Prints 'middle in v7.0.0.13, consistent with your example.
           (m x))])))            ; Would this break without the use-site scope, if
  (syntax-parse                  ; syntax-local-identifier-as-binding were called on it?
    [(~m)
     (void)]))
```

michaelballantyne:
An example that breaks this:

```
#lang racket

(define x #t)

(define-match-expander m
  (syntax-rules ()
    [(_ arg)
     (? (let ([arg #f]) (if x string? number?)))]))

(match "foo"
  [(m x) (void)])
; match: no matching clause for "foo"
```

That also breaks in 6.12 and 7.0, though!

lexi-lambda:
I think I would prefer that `syntax-local-identifier-as-binding` were left as it is (that is, only allowed on identifiers), since I don’t know of a situation in which the recursive-flipping behavior of `syntax-local-introduce` is necessary, and I think the difference between the two functions is helpful to guide people away from improper uses of `syntax-local-identifier-as-binding` (which I think is a confusing operation to begin with). I would also prefer that `syntax-local-identifier-as-binding` were not used where it is not necessary, so if it is irrelevant to `provide` transformers, I would prefer it not be used there.

I have one additional question, too. In this pull request, you changed the context arguments from `'expression` to `(list (gensym))`. Could you explain why? While none of the allowed contexts seem perfect here, `'expression` seems to fit—in my mind—much more closely than a list representing an internal definition context.

michaelballantyne:
@AlexKnauth your example works correctly either way, because the relevant use-site scope is created for the expansion of m, and only that created for ~m would be removed by my use of identifier-as-binding. 

@lexi-lambda I switched the context from 'expression to (list (gensym)) such that use-site scopes created during those local expansion can be removed by `syntax-local-identifier-as-binding` outside the expansion, rather that only within it. `local-expand` with `'expression` switches to a new list for tracking created use-site scopes that should be removed by `syntax-local-identifier-as-binding`, whereas for an internal definition context it does not.

I think that the problem with `?` in match reveals that you're right that `syntax-local-identifier-as-binding` should only work on identifiers.

lexi-lambda:
> I switched the context from 'expression to (list (gensym)) such that use-site scopes created during those local expansion can be removed by `syntax-local-identifier-as-binding` outside the expansion, rather that only within it. `local-expand` with `'expression` switches to a new list for tracking created use-site scopes that should be removed by `syntax-local-identifier-as-binding`, whereas for an internal definition context it does not.

I see… I didn’t realize that the context had that effect. That’s subtle.

I think that, thinking on it, that behavior makes some sense, but I don’t believe it’s documented anywhere. Since it seems relevant to anyone trying to implement that sort of thing, it would be nice if it were… though I’m not sure where it would go or how it could be usefully presented.

michaelballantyne:
Here's another example of what's broken, in all versions: 6.12, 7.0, and mine.

```
#lang racket

(module a racket
  (require racket/provide (for-syntax racket/provide-transform racket/function))

  (begin-for-syntax
    (define f (curryr string-append "*")))

  (define-syntax m
    (make-provide-pre-transformer
     (lambda (stx modes)
       (syntax-case stx ()
         [(_ arg spec)
          (pre-expand-export #'(filtered-out (let ([arg (lambda (x) x)]) f) spec) modes)]))))

  (define foo 5)

  (provide (m f foo)))

(require 'a)

foo*
; foo*: unbound identifier in: foo*
```

rmculpepper:
I don't have this fully paged in, but my recollection, reinforced by @michaelballantyne's example above, is that use-site scopes are needed to differentiate identifiers in syntax constants in a macro definition from identifiers given to a use of the macro in the same definition context. The implementation of `local-apply-transformer` creates a temporary internal-definition-context object and uses it to bind the transformer to a name so it can be called via `local-expand`. Since the transformer exists before the intdefctx, its syntax constants don't have the "danger" scope, so use-site scopes shouldn't be necessary for *that* intdefctx. (If the caller supplies other intdefctxs to `local-apply-transformer`, then perhaps their use sites are necessary, but then the burden falls on the caller to use `syntax-local-identifier-as-binding`, as usual.) So if `local-apply-transformer` had a way of requesting that use-site scopes be disabled for the temporary intdefctx that it creates, would that solve the problem? Or alternatively, extending `internal-definition-context-introduce` with a `'remove-use-site` mode and using it to remove the problematic use-site scope after the fact?

IIUC, that's why indiscriminate use of `syntax-local-identifier-as-binding` would often work. Is the example where it fails because it removes use-site scopes from intdefctxs other than the temporary `apply-local-transformer` one?

yjqww6:
@rmculpepper , I believe use-site scopes are necessary for local-apply-transformer, please see #2069 and #2212 .

michaelballantyne:
@rmculpepper I would refine your description a little and say that use-site scopes are necessary whenever use-site identifiers don't have other scopes that syntax literals within the macro lack. So in the case of let-syntax (forgetting for a moment the definition context within), use-site scopes aren't needed for a use in the body because the let-syntax scope is present in the body but not on any syntax literals within a macro. However, they're needed for letrec-syntax.

Here we're considering a transformer bound within a definition context, and then a use of local-apply-transformer within that context, which does create a nested context. However, no scopes distinguish use-site identifiers from syntax literals generated by the transformer, because the nested internal definition context scope is added to both after the fact (as an "inside-edge" scope). My (match transformer)[https://github.com/racket/racket/pull/2237#issuecomment-414761718] is an example.

Before I realized that local binding forms could be embedded within patterns, I thought I could get away with indiscriminately removing use-site scopes with syntax-local-identifier-as-binding, because all pattern variable bindings should splice into the context surrounding the match transformer expansion. As per my match transformer example, and similar examples with the other kinds of transformers using local-apply-transformer, that was wrong. In the case of syntax-parse, local bindings from racket forms like `let` within do clauses should retain the use-site scope. That means that to get the right answer, we need to selectively apply syntax-local-identifier-as-binding only to the pattern variables, attribute bindings, etc in expanded patterns.

rmculpepper:
Thanks @yjqww6 and @michaelballantyne for the links and explanation. I'm still paging in the details of the mechanics. Meanwhile, I've been thinking about the discrepancy between the behavior of `local-expand` vs CPS macros. As a starting point, here is a slight adaptation of one of the examples from #2069:

    (let ()
      (define x sub1)
      (define-match-expander succ
        (lambda (stx)
          (syntax-case stx ()
            [(_ v)  #'(app (let ([v 100]) x) v)])))
      (match 3 [(succ x) x]))

Here's a program that I think captures the spirit of the same example but uses CPS macros instead of `local-apply-transformer`:

    ;; app : match subform, interpreted by match-k
    (define-syntax app (syntax-rules ()))

    ;; (match expr extended-pattern expr) : expr
    (define-syntax-rule (match expr (pattern-id pattern-arg) rhs)
      (pattern-id pattern-arg (match-k (expr rhs))))

    ;; (match-k (expr expr) core-pattern)
    (define-syntax match-k
      (syntax-rules (app)
        [(match-k (expr rhs) (app proc var))
         (let ([var (proc expr)]) rhs)]))

    (let ()
      (define x sub1)

      (define-syntax-rule (succ v (k kenv))
        (k kenv (app (let ([v 100]) x) v)))

      (match 3 (succ x) x)))

This version of `match` supports pattern transformers using a CPS protocol rather than `local-apply-transformer`. (Briefly, a Y-consuming continuation is represented as a syntax object of the form `(M X)`, where M is a macro taking two arguments, an X and a Y. That is, it's just a closure encoded as syntax.) But unlike the original version, it works, because the use-in-same-context-as-definition is made explicit by the CPS protocol, so it gets a use-site scope. But it doesn't require any explicit identification of binders. (Why not?) I also tried a similar example with a definition form (`match-define`), and it also seemed to work, although there I had to change `succ` to split the variable-to-shadow from the variable-to-bind.

IIUC, the difference is that in the CPS version, the original arguments to `match` get a use-site scope when `succ` is expanded (because the continuation is reified as a syntax object).

Is there a way to make the mechanics of macro expansion and `local-{expand,apply-transformer}` correspond to the CPS version?

michaelballantyne:
Could you share your CPS match-define?

rmculpepper:
(define-syntax app (syntax-rules ()))
    (define-syntax-rule (match-define expr (pattern-id pattern-arg ...))
      (pattern-id pattern-arg ... (match-define-k (expr))))
    (define-syntax match-define-k
      (syntax-rules (app)
        [(match-define-k (expr) (app proc var))
         (define var (proc expr))]))
    
    (let ()
      (define x sub1)
      (define-syntax-rule (succ shadow bind (k kenv))
        (k kenv (app (let ([shadow 100]) x) bind)))
      (match-define 3 (succ x y))
      y)

michaelballantyne:
I think your understanding is correct; the difference is that the rhs of the match clause also gets the use-site scope because it's passed as the continuation. In the `match-define` case, it's that the expansion of `define` removes the use-site scope.

michaelballantyne:
I don't see a way to take advantage of the CPS parallel. The CPS version escapes the problem of binders that splice by instead passing along all the references they might be meant to bind, but I don't think that's a good solution in general.

mflatt:
@michaelballantyne I think the solution to the letrec bug is to introudce an extra scope for the body of a `letrec` (really, `letrec-values` and `letrec-values+syntaxes`). I expect to push that change after checking that it works for the main distribution and a few macro-intensive packages.

In fact, you and William Hatch observed this problem and told me the solution in email back on December 2, 2016. I recall that we had a brief discussion afterward, where I promised to add the repair... and I didn't get to it, for which I have various excuses, but really I just dropped the ball.

I wonder whether that isn't also the solution for a macro that wants to implement `letrec` itself. That is, your first test case should use `(local-expand ((make-syntax-introducer) #'ebody) 'expression null ctx)` to re-implement the corrected primitive `letrec` expansion. Or is the problem deeper?

rmculpepper:
I think the CPS parallel argues that the current criterion for *omitting* use-site scopes is wrong; they should only be omitted if the main macro *and* all macros used by `local-expand`ed terms already have intervening scopes since their definitions. What would happen if a use-site scope were added on every macro expansion step instead of the current criteria? (If I'm reading it right, that's what the redex model already does.)

Also, is there a simple, complete example that demonstrates why (and when) `syntax-local-identifier-as-binding` is needed?

mflatt:
@rmculpepper Thanks - that helps me see the deeper issue.

Here's a standalone example that fails if you comment out with `with-syntax` clause that applies `syntax-local-identifier-as-binding`:

```
#lang racket/base
(require (for-syntax racket/base
                     syntax/parse))

(define-syntax (localish stx)
  (syntax-parse stx
    [(_ defn body)
     (define m (local-expand #'defn (list (gensym)) (list #'define-values)))
     (syntax-parse m
       #:literals (define-values)
       [(define-values (id ...) rhs)
        (with-syntax ([(id ...)
                       (map syntax-local-identifier-as-binding
                            (syntax->list #'(id ...)))])
          #`(letrec-values ([(id ...) rhs])
              body))])]))

(define-syntax-rule (my-define id rhs)
  (define-values (id) rhs))

(localish (my-define x 10)
          x)
```

We could add a use-site scopes for all transformer calls (at a cost of about 10% in performance), but doing that for all expansions would include adding a use-site scope for `local-expand` in `'expression` mode, so `match` would still have to change to use `syntax-local-identifier-as-binding`. (I think the right place to do that is in the `id` case of `parse`, not in `match-expander-transform`.)

At the moment, I don't see a way around adding a use-site scope on `'expression` expansion, as  originally proposed. A secondary question is how to make that change without breaking everything; maybe the answer is to deprecate `'expression` mode and introduce a variant that works right while obliging callers to use `syntax-local-identifier-as-binding`.

michaelballantyne:
I've updated this PR with a more careful fix for require, match, and syntax-parse expanders, only using `syntax-local-identifier-as-binding` on the splicing binding positions. In `syntax-parse` I also had to apply it to identifiers bound in the declenv bound-id table, because otherwise use-site scopes could cause declaration lookup to fail for #:literals and #:datum-literals declared names.

I'm not confident I hit all the right places in syntax-parse, so a review from @rmculpepper would be appreciated! I think I addressed normal pattern variable bindings, attribute bindings due to syntax classes, `~bind` attribute bindings, and ~reflect attribute bindings. `define` bindings within a `~do` pattern will not splice outside of a pattern expander. This looks difficult to fix, because the body of the do block isn't parsed when the pattern is parsed, but instead later as part of the expansion of the result of compiling the pattern.

I've tested with the main distribution and Turnstile. I'm not sure how to evaluate whether the potential for backwards compatibility problems merits limiting the fix to a new `'expression`-like context for `local-expand` as @mflatt suggests.

PR #2243, opened 2018-08-28
Make dependent arguments in ->i return the contract directly

shhyou:
The commit inlines the dependent contract generation function in `->i`. This will simplify the implementation of a basic `->i` generator and exerciser (still WIP).


shhyou:
I think the CI failure is not related to this change.

rfindler:
Thanks!

PR #2244, opened 2018-08-28
Fix domain exerciser order in randomly generated function

shhyou:
Also, common uses like `(   (contract-random-generate (-> integer? boolean? any))   0   #f)` won't expose the problem since contracts coerced from predicates (e.g. `integer?`) have no exerciser.


rfindler:
Thanks!

PR #2250, opened 2018-09-04
Contracts: adds random generation to hash/c

DarrenN:
## Background

I noticed that `hash/c` didn't have the ability to randomly generate inputs:

```scheme
> (contract-random-generate (hash/c integer? boolean?))
; contract-random-generate: unable to construct any generator for contract:
;   (hash/c integer? boolean?) [,bt for context]
```

So at the prodding of @samth I added a simple `hash/c-generate` and `hash/c-exercise` to the contract:

```scheme
> (require racket/contract)
> (contract-random-generate (hash/c number? (listof number?)))
'#hash((0 . (0 1.308710636691283e-07)))
> (contract-random-generate (hash/c number? (hash/c number? string?)))
'#hash((-150.0-2147483647.0i
        .
        #hash((-2147483647.0-185.0i
               .
               "\uA7CBጚ䛡\U000351CD𫑫\U000B46B6\U000FAAC2\U0008DD1F\U000AA812⛕\U0010D6B6\U000D8D2B\U00048EBE㷎\U00035091𩣲Ų볈⢍읿ଓ愒妙\U0009F757㏥\U00072BE1彞辯逰能\U0005F69E뮝\U0004ACA7\U00071BC1\U00099D0F𧎝ᐾ\U00093943㜷\U0001F964髣\U00062741꧖粫\U00065E50\U00087B85\U000D696F\U000C8D02\U00046E0D\U000BE119魥\U00013BFF\U000A502D\U0009C413㍡刬ᑨ\U000CD8EC鵸\U000342E6ⷲꪌ㔱\U0003579F\U000E3620똠\U000459D9죄膅骖䎷駥\U000D72A8\U000A201E\U000BFD8E\U000FC781\U00052888᧳玠u逫煃ᨠ鷊\U0008B647遤縙髆ᇪ🞛\U00056173\U0009FDD4\U000B2F5A熑藬\U000F4E9B\U00067EF7\U0002DAA6"))))
```

Added a few simple tests as well.


rfindler:
Overall this looks good to me, but perhaps with one improvement, namely that it sometimes generates hash tables that are empty and sometimes it generates hash tables that have multiple elements in them.

shhyou:
1. `contract-random-generate/choose` can return `#f`. Perhaps there should be a check somewhere. In the case where either `gen-key` or `gen-val` is `#f`, the generation for the entire `hash/c` contract could still be successful as the empty hash is always a valid value.

2. If I understand correctly, the function returned by `multi-exercise` would need to be called with values corresponding to the supplied contract. So for example, the first result of `(multi-exercise (list (base-hash/c-dom ctc) (base-hash/c-rng ctc)))` will be a function accepting a value of `(base-hash/c-dom ctc)` and a value of `(base-hash/c-rng ctc)`.

I think `hash/c-exercise` needs to iterate over the given hash and exercise the keys and the values. It would be nice if `hash/c-exercise` can find this bug:
```
(contract
 (hash/c symbol? (-> number? boolean?))
 (hash 'key add1)
 'pos 'neg)
```

DarrenN:
@rfindler update `hash/c-generate` to periodically return empty hashes and hashes with multiple elements. Used `fuel` as a seed to do some randomization. I also look at `contract-random-generate/choose` to see if it returns `#f` but in testing never saw it do so.

@shhyou slightly hung up on the exercising of the hash, mostly due to unfamiliarity with the intricacies of contracts. I tried to put together an exercise function but run into issues when when the `rng` is a `chaperone-contract?`, which in the case of `(hash/c symbol? (-> number? boolean?))` is is. With `flat-contract?` I can check them like a predicate (ex: `(ctc value)`) but I can't figure out how to check chaperone contracts against values.

```racket
(define (hash/c-exercise ctc)
  (define env (contract-random-generate-get-current-environment))
  (define dom (base-hash/c-dom ctc))
  (define rng (base-hash/c-rng ctc))
  (λ (fuel)
    (define gen-key (contract-random-generate/choose dom fuel))
    (define gen-val (contract-random-generate/choose rng fuel))
    (cond
      [(and gen-key gen-val)
       (values
        (λ (h)
          (define key (car (hash-keys h)))
          (define val (car (hash-values h)))
          (contract-random-generate-stash
                env
                rng
                (and (dom key)
                     (rng val)))) ;; fails when rng is a chaperone-contract?
        (list dom rng))]
      [else
       (values void '())])))
```

shhyou:
`contract-struct-exercise` or `multi-exercise` is the missing piece. They extract the functions that exercise the corresponding contracts. Examples:

```
(define (hash/c-exercise ctc)
  (define env (contract-random-generate-get-current-environment))
  (define dom (base-hash/c-dom ctc))
  (define rng (base-hash/c-rng ctc))
  (λ (fuel)
    ;; passing (list dom rng) to multi-exercise will produce
    ;; a function that exercise values of form (list/c dom rng)
    ;; and a list of newly available contracts.
    (define-values (exercise-list-dom-rng available-ctcs)
      ((multi-exercise (list dom rng)) fuel))
    (values
     (λ (h)
       ;; iterate over key-value pairs
       (for ([(k v) (in-hash h)])
         (exercise-list-dom-rng (list k v))
         ;; here you can stash k and v, such as
         (contract-random-generate-stash env dom k)
         etc..))
     ;; i'm not sure if this should be (cons dom (cons rng available-ctcs))
     ;; if you stash both k and v during the iteration, including dom
     ;; and rng is better
     available-ctcs)))
```

I think you don't need to check `contract-random-generate/choose` in `hash/c-exercise`.

racket/collects/racket/contract/private/hash.rkt
```diff
@@ -198,7 +198,8 @@
   (define this-immutable (base-hash/c-immutable ctc))
   (λ (fuel)
     (define rnd (random fuel))
-    (define rnd-len (add1 (modulo (add1 rnd) fuel)))
+    ;; keep amount of random elements reasonable, no more than 4
+    (define rnd-len (add1 (modulo (add1 fuel) 4)))
```
shhyou: What is the reason to generate only 4 elements? If you worry about the size of the generated hash table, I think the approach used by `(listof _)` can be applied here to decide the size of the hash table (with a certain probability increase the size and recur):
https://github.com/racket/racket/blob/master/racket/collects/racket/contract/private/list.rkt#L30-#L34

shhyou:
Overall looks good to me!

racket/collects/racket/contract/private/hash.rkt
rfindler: The idea is that it can use the geometric distribution. The wikipedia page on it at the beginning is confusing to me, but a bit later down it says this:

> Consider a sequence of trials, where each trial has only two possible outcomes (designated 
> failure and success). The probability of success is assumed to be the same for each trial. In 
> such a sequence of trials, the geometric distribution is useful to model the number of failures
> before the first success. The distribution gives the probability that there are zero failures
> before the first success, one failure before the first success, two failures before the first 
> success, and so on.

So you can just write a loop that works like that.

DarrenN:
@shhyou / @rfindler thanks for the pro-tips on `(rand-choice)` that worked out very well!

PR #2255, opened 2018-09-06
warning and fix for openssl bad cert paths

rmculpepper:
The current linux-natipkg build has broken openssl support because the natipkg libssl.so library refers to private paths on the build machine. Unfortunately, there isn't a simple fix, because different distros put their certs at different locations, and there doesn't seem to be a single openssl build configuration that would work on both Debian and RedHat-based systems, for example.

The first commit detects when *neither* the cert file nor directory path exists and logs an error.

The second commit implements a workaround: if openssl reports a certain recognizable nonsense path for the cert file or directory, it is replaced with a better guess. **This commit would rely on changes to the natipkg openssl build.**

There's a nice summary of the ssl cert location issue here: https://www.happyassassin.net/2015/01/12/a-note-about-ssltls-trusted-certificate-stores-and-platforms/

### DIFF
--- a/racket/collects/openssl/mzssl.rkt
+++ b/racket/collects/openssl/mzssl.rkt
@@ -363,14 +363,37 @@ TO DO:
 (define-crypto X509_get_default_cert_file_env (_fun -> _string))
 
 (define (x509-root-sources)
+  ;; Workaround for natipkg openssl library: the default cert locations vary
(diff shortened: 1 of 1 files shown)


rmculpepper:
@mflatt thanks! I'll merge this tomorrow, after I've had a chance to test it against a natipkg nightly build.

PR #2257, opened 2018-09-09
Coerce the argument to a contract in contract-random-generate/choose

shhyou:
I also tried to update the documentation of contract-random-generate/choose to clear things up, but I am not sure what is the best way to describe its constraint. The generators and exercisers for any contracts are of form

```
(λ (ctc)
  (λ (fuel)
    ;; A
    (λ ()
      ;; B
      ...)))

(λ (ctc)
  (λ (fuel)
    ;; A
    (values (λ (v)
              ;; B
              ...)
            ...)))
```

The function `contract-random-generate/choose` can only be used at point A. It cannot be used at point B (or any other places).


rfindler:
How about something like this?

```
In other words,
@racket[contract-random-generate/choose] is available 
only during the dynamic extent
of the call to @racket[_generate] (or @racket[_exercise]), i.e., after it receives the
  @racket[_fuel] argument, and before it returns a thunk (or exerciser).
``

shhyou:
Sounds good. Edited.

PR #2258, opened 2018-09-09
improve expt doc description

pnwamk:
Makes the docs include another case where `expt` is undefined:

```c
    if (!SCHEME_COMPLEXP(e)) {
      neg = scheme_is_negative(e);
    } else {
      neg = !scheme_is_positive(scheme_complex_real_part(e));
    }
```


sorawee:
It looks like the [current documentation of `expt`](https://docs.racket-lang.org/reference/generic-numbers.html#%28def._%28%28quote._~23~25kernel%29._expt%29%29) is now even more detailed than the proposed change here. Is this PR obsolete? Can we close it?

PR #2262, opened 2018-09-11
add ~indexed syntax template form

rmculpepper:
**Not ready for merge, just for discussion**

This adds a `~indexed` form for syntax templates. It lets some macros that previously required `generate-temporaries` to be written as simple pattern macros.

It's intended as an alternative to the approach in the `subtemplate` package (@jsmaniac). It's less concise and less pretty than unicode subscripts, but it avoids the need for extra communication between syntax variable binding forms (`syntax-case`, `syntax-parse`, etc) and `syntax` templates.

As an example, here's a macro equivalent to `set!-values`:

    (define-syntax-rule (set!-multi (x ...) rhs)
      (let-values ([((~indexed tmp x) ...) rhs])
        (set! x (~indexed tmp x)) ...
        (void)))

    (set!-multi (a b c) (values 1 2 3))
    => (let-values ([(tmp_0 tmp_1 tmp_2) (values 1 2 3)])
         (set! a tmp_0)
         (set! b tmp_1)
         (set! c tmp_2)
         (void))

Unlike `generate-temporaries` (and, IIRC, `subtemplate`?), the `~indexed` form doesn't add any additional scopes to the identifiers it produces.


rfindler:
Nice!

Is it possible to make more than one sequence of temporaries from the same
sequence of input identifiers?

Robby

rmculpepper:
Yes, the same index variable can be used with different prefixes:

    > (with-syntax ([(x ...) '(a b c)])
        #'([(~indexed t x) (~indexed u x)] ...))
    #<syntax ((t_0 u_0) (t_1 u_1) (t_2 u_2))>

and the same prefix can be used with different index variables:

    > (with-syntax ([(x ...) '(a b c)] [(y ...) '(P Q)])
        #'(t_0 ((~indexed t x) ...) ((~indexed t y) ...)))
    #<syntax (t_0 (t_0 t_1 t_2) (t_0 t_1))>

Note that in the second example, all three `t_0` identifiers are `bound-identifier=?`.

samth:
A somewhat different way to do this would be to make `~indexed` a pattern form, so you'd write:

```
(define-simple-macro (set!-multi ((~indexed tmp x) ...) rhs)
  (let-values ([(tmp ...) rhs])
    (set! x tmp) ...
    (void)))
```

This has the advantage of being part of `syntax-parse` and fitting with the normal extension story there.

rmculpepper:
@jsmaniac no, that macro is still hygienic, because the identifiers created by `~indexed` still get the macro scope that distinguishes identifiers introduced by the macro from identifiers in the macro's arguments.

The difference is that there are no extra scopes distinguishing identifiers created by `~indexed` from other identifiers in the macro template. For example, here's a variation on `set!-multi` that returns the value assigned to the first variable instead of returning void:

    (define-syntax-rule (set!-multi (x ...) rhs)
      (let-values ([((~indexed tmp x) ...) rhs])
        (set! x (~indexed tmp x)) ...
        tmp_0))

In other words, `generate-temporaries` provides the ability to produce variable-length sequences of identifiers but adds a redundant layer of freshness to the identifiers (the extra scopes), whereas the `~indexed` form only does the variable-length sequence part and leaves freshness to the normal macro hygiene mechanism.

jackfirth:
We're trying to improve this state of affairs, correct?

```scheme
(define-syntax-rule (set!-multi (x ...) rhs)
  #:with (tmp ...) (generate-temporaries #'(x ...))
  (let-values ([(tmp ...) rhs])
    (set! x tmp) ...
    (void)))
```

The proposed `~indexed` form makes `generate-temporaries` unnecessary:

```scheme
(define-syntax-rule (set!-multi (x ...) rhs)
  (let-values ([((~indexed tmp x) ...) rhs])
    (set! x (~indexed tmp x)) ...
    (void)))
```

I'm not fond of how it requires repeating the `~indexed` bit and the source identifier multiple times. It's a little awkward. @samth's proposal solves that problem:

```scheme
(define-simple-macro (set!-multi ((~indexed tmp x) ...) rhs)
  (let-values ([(tmp ...) rhs])
    (set! x tmp) ...
    (void)))
```

...but I agree with @SuzanneSoy that putting temporary identifiers into the macro signature is confusing.

What if we made a pattern directive for this?

```scheme
(define-syntax-rule (set!-multi (x ...) rhs)
  #:temporaries tmp x
  (let-values ([(tmp ...) rhs])
    (set! x tmp) ...
    (void)))
```

PR #2266, opened 2018-09-13
Build cs variant on Travis.

samth:


gus-massa:
It is strange that [fixnum.rktl](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/fixnum.rktl) doesn't raise any error, because in the CS version fixnums have only 60+1 bits instead of 62+1 bits. I looked at the test and I think they are not strict enough to detect the two missing bits. In particular, this is important for the checks in `same-result/extremenum` ([line 104](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/fixnum.rktl#L104)) because many checks raise an error are trivially true.

And I was expecting to find some corner case like 

    (test #t eq? (greatest-fixnum) (+ (greatest-fixnum) 1 -1))
    (test #f fixnum? (+ (greatest-fixnum) 1))

but I didn't found them. I'm not sure if I've seen them before somewhere, or I only imagined that they were in some test.

I also tried to find other parts with problems, that assume that the fixnums have 62+1 bits, but there are very few. Perhaps fix [unsafe.rktl#L930](https://github.com/racket/racket/blob/247fc89f15b1fb310bf0bfd9a2c0bcd9ccf31c76/pkgs/racket-test-core/tests/racket/unsafe.rktl#L930) and add more corner cases to [contract/private/generate-base.rkt#L56](https://github.com/racket/racket/blob/a1b9fd0965887e14052a2235e5072570796eaaa8/racket/collects/racket/contract/private/generate-base.rkt#L56)

gus-massa:
I think I fixed the matrix. [Branch](https://github.com/gus-massa/racket/tree/18-10-E-samth-travis-cs) / [Commit](https://github.com/gus-massa/racket/commit/6b9ad27b92c7efad14a180c08691a4560808aadf) / [Tests ](https://travis-ci.org/gus-massa/racket/builds/448490022)

But I'm not sure if the calls to racket and raco at the end use the cs version or the classic version. Perhaps it's necessary to fix the PATH, or call racket$TARGET, or something.

samth:
I think by default `make cs` builds `racketcs` binaries. I see two ways to deal with this. One is to build as `racket`, which requires setting `RACKETCS_SUFFIX=""`. The other is to use some variable such as that in running all the commands.

gus-massa:
Some experiments using an extremely minimized version of .travis.yml (so I can try the variations faster).

* racket$TARGET: [Commit](https://github.com/gus-massa/racket/commit/af23433580fd8ae8c232fe56447e9954ac178fca) / [Tests](https://travis-ci.org/gus-massa/racket/builds/457893219)

* RACKETCS_SUFFIX: [Commit](https://github.com/gus-massa/racket/commit/8f1b744819a823a8a23df612b3fb70a2a6b6d749) / [Test](https://travis-ci.org/gus-massa/racket/builds/457904363)

The default value of "cs" for a missing RACKETCS_SUFFIX is a small problem. It can be solved with some bash magic, but it will be solved automatically some day when the default changes.

The results of the tests are a nice failure, because racket and racketcs disagree about what exception to generate when someone tries to create a string that is tooooo long. But the build is marked as a succes anyway :).

samth:
@gus-massa thanks for pushing this along. Can you either push directly to my branch (it allows edits from any committer) or make a pull request to my branch, so that we can move forward here?

samth:
This builds correctly and fails tests in seemingly plausible ways, so I'm merging it.

PR #2268, opened 2018-09-13
Track disappeared define-syntaxes in internal definition contexts

lexi-lambda:
This PR improves `'origin` tracking for internal uses of `define-syntaxes` so that Check Syntax can understand the use of `define-syntax` in the following program:

```racket
#lang racket
(let ()
  (define-syntax foo (syntax-rules ()))
  (void))
```


lexi-lambda:
Here is my reasoning for why `'origin` makes sense here: an internal-definition context is a collection of a variety of different definition forms, all of which are transformed, morally, into `letrec`. The `letrec` form clearly does “come from” a macro transformation involving a variety of disparate syntax objects, namely all the definition forms.

In the case of ordinary `define`s, there are actual binding pairs to attach the `'origin` property to, which is certainly convenient, since it feels “more specific”. The `define-syntax`es go away entirely, though, so they cannot contribute to any particular piece of syntax in the expansion. The documentation for `syntax-track-origin` says the following:

> Use the `syntax-track-origin` procedure in a macro transformer that discards syntax (corresponding to `orig-stx` with a keyword `id-stx`) leaving some other syntax in its place (corresponding to `new-stx`).

This makes sense to me. In this case, the (invisible) internal-definition transformer discards `define-syntax` forms and leaves a `letrec` in its place, so it should track origin information.

One more thing to consider: using `'disappeared-use` *wouldn’t* work here, since it would discard any additional `'origin` information on the `define-syntax` form itself, meaning macros that expand into `define-syntax` wouldn’t be properly tracked. (In practice, this actually includes `define-syntax` itself, since it is really just a macro that itself expands into `define-syntaxes`.) I think this is all consistent with my above interpretation of what an internal-definition context transformation *is*.

As for your smaller concern, using `keep-as-needed` to drop unneeded information makes sense to me.

mflatt:
That argument makes sense to me. I checked with Robby, and he agrees. But we'd be more comfortable delaying this change to just after the branch for the v7.1, so we can get more experience with the change before a release. If that sounds ok to you, it's a delay of a few days from now.

lexi-lambda:
Sure, that’s fine; this change obviously isn’t very important. Shall I plan to merge this after branch day, or should I wait for some further discussion?

mflatt:
I'm guessing that we won't end up inspiring more discussion, in which case there's no need to wait beyond branch day.

lexi-lambda:
Alright, I’ll plan on merging it then. Thank you!

rfindler:
Good plan, IMO!

PR #2270, opened 2018-09-14
Make `racket/private/serialize-struct` cross-phase

LeifAndersen:
This makes sense because a serializable struct should probably remain
serializable, even if it was inserted into a bit o' 3d syntax.


LeifAndersen:
Since no one seems to have have any objections to this, and its been open for a while, I'll go ahead and merge it.

PR #2275, opened 2018-09-17
Clean up the documentation on cross phase persistent modules

LeifAndersen:


LeifAndersen:
Attempting to also tag @michaelballantyne .

PR #2276, opened 2018-09-18
Adding build badges to the readme

chrispat:
Thought it would be nice to status on the front page of the repo.


samth:
We're no longer using Azure pipelines, having moved to GitHub Actions.

PR #2284, opened 2018-09-22
Only copy argv if not NULL

### LINKED ISSUE #2283: null pointer passed as source argument of memcpy
Null pointer is being passed in this line as source of memcpy.

https://github.com/racket/racket/blob/49d31414b768910e83b74b349a252c18da691486/racket/src/racket/src/startup-glue.inc#L155

To see it in action, compile racket with `--enable-cify --enable-ubsan` and run a `raco exe`.
You'll see a line like this:
```
../../../racket/gc2/../src/../src/startup-glue.inc:155:3: runtime error: null pointer passed as argument 2, which is declared to never be null
```

So, I have only looked at the code for about 10 mins but this function is suspicious.
```
 /* stash before allocation: */
  p = scheme_current_thread;
  p->ku.k.p1 = (void *)proc;
  p->ku.k.i1 = argc;
  p->ku.k.i2 = runstack_space;
  p->ku.k.p2 = (void *)argv;
  
  argv2 = MALLOC_N(Scheme_Object*, argc);

  p = scheme_current_thread;
  argv = (Scheme_Object **)p->ku.k.p2;
  
  memcpy(argv2, argv, sizeof(Scheme_Object *) * argc);
```

Starts by assigning `scheme_current_thread` to `p`, assigns some values to `p` and then goes on to reassign `scheme_current_thread` to `p` (why?). I think `p->ku.k.p2` which is assigned to `argv` and passed as argument 2 to `memcpy` can be traced back to `(void *) argv`, which can clearly be NULL.

pmatos:
Fixes #2283.

### DIFF
--- a/racket/src/racket/src/startup-glue.inc
+++ b/racket/src/racket/src/startup-glue.inc
@@ -138,7 +138,7 @@ static Scheme_Object *do_apply_k(void)
 static Scheme_Object *c_handle_overflow_or_space(Scheme_Object *proc, int argc, Scheme_Object **argv, int runstack_space)
 {
   Scheme_Thread *p;
-  Scheme_Object **argv2;
+  Scheme_Object **argv2 = NULL;
 
   /* stash before allocation: */
   p = scheme_current_thread;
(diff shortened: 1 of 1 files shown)


pmatos:
I feel slightly queasy by touching this file without really understanding the surrounding context but I think this should fix the issue we are seeing with ubsan enabled.

mflatt:
I'd prefer a patch that checks for `argc` being 0 instead of `argv` being NULL. I think it's dumb that C disallows NULL to operations like `memcpy` even when the size argument is 0, and it's better to make the workaround explicit insteda of potentially hiding problems with NULL and a non-0 size.

pmatos:
@mflatt I understand your pov. I have pushed a new commit to the PR.

pmatos:
@mflatt OK, I was not aware that the GC could just change `scheme_current_thread` on our backs. However, isn't it true that moving the code doesn't help much? The GC can come and change `scheme_current_thread` again between the assignment and `memcpy`.

mflatt:
The GC will only potentially move objects during a memory-allocation request. There are no additional allocation requests between the the assignment and `memcpy`.

I think the sets of changes now is probably ok, but can you squash them to make it easier to tell and to make a cleaner commit?

Thanks!

PR #2286, opened 2018-09-22
Remove test for foff != 0

### LINKED ISSUE #2285: Possibility of null or incorrect memcpy
There's an interesting comparison I cannot make sense of in `foreign.c`:

https://github.com/racket/racket/blob/49d31414b768910e83b74b349a252c18da691486/racket/src/foreign/foreign.c#L2678

This line is in the implementation of the ffi `malloc` function. The function is quite readable, if you follow the ffi docs for `malloc`. I reached this bit because Coverity says that if we are testing `from != NULL` then it's possible that `from` is `NULL` and therefore if `foff == 0` and `res != NULL`, we are calling `memcpy` with `foff` as its second argument.

Unless I am missing something this seems incorrect. From the lines:
```
      from = SCHEME_FFIANYPTR_VAL(a);
      foff = SCHEME_FFIANYPTR_OFFSET(a);
```

I expect that `from` is always non-zero, and `foff >= 0`, therefore shouldn't the condition be simply:
```
if ((from != NULL) && (res != NULL))
    memcpy(res, W_OFFSET(from, foff), size);
```

pmatos:
Coverity complains here that we are calling malloc with an offset only. Looking at the code and the docs, makes it look like that the if condition is overworked and we can remove the test for `foff != 0`.

Fixes #2285.

### DIFF
--- a/racket/src/foreign/foreign.c
+++ b/racket/src/foreign/foreign.c
@@ -2675,7 +2675,7 @@ static Scheme_Object *foreign_malloc(int argc, Scheme_Object *argv[])
   }
   res = scheme_malloc_fail_ok(mf,size);
   if (failok && (res == NULL)) scheme_signal_error("malloc: out of memory");
-  if (((from != NULL) || (foff != 0)) && (res != NULL))
+  if ((from != NULL) && (res != NULL))
     memcpy(res, W_OFFSET(from, foff), size);
   if (SAME_OBJ(mode, raw_sym))
     return scheme_make_foreign_external_cpointer(res);


abmclin:
Speaking here merely as a curious bystander. I notice that at the top of `foreign.c` there is a comment that says `foreign.c` is an auto generated file and edits need to be made to `foreign.rktc` instead.

mflatt:
I think the intent here really is to use the offset even if the pointer's base is NULL.

Maybe the right clarification is to set a separate flag when `from` and `foff` are set as a source pointer, and then test that flag to mean that `from` and `foff` are ready to be used. (But with the change, as @abmclin says, in "foreign.rkt".)

pmatos:
@abmclin You are right. I shouldn't have edited this file. Thanks for pointing it out.

pmatos:
@mflatt I will take a look. Will close this for now as it was opened with changes to the wrong file.

PR #2287, opened 2018-09-22
Avoid shift by negative number by turning condition around

pmatos:
Avoid undefined behaviour in condition which could shift by negative i. By checking first if `i >= 0` we avoid that case.


pmatos:
Felt like opening an issue again would be too much noise, therefore sending PR only. If you prefer to have an issue per PR, please let me know.

mflatt:
The `i >= 0` test here seems strange, because `sizeof(uintptr_t)` can't be zero, so `sizeof(uintptr_t) * 8- 1` should always be positive. Am I misreading the code?

PR #2291, opened 2018-09-25
Fix memory leak

pmatos:
For `case == -1`, we were not free'ing `env`, which is allocated if `envvars` was non-zero.

Allocation was done on this line:
https://github.com/racket/racket/blob/bc3777792d835b8e3ad5140159032aacf9079a90/racket/src/rktio/rktio_process.c#L1276

### DIFF
--- a/racket/src/rktio/rktio_process.c
+++ b/racket/src/rktio/rktio_process.c
@@ -1434,6 +1434,10 @@ rktio_process_result_t *rktio_process(rktio_t *rktio,
       } else {
 	RKTIO_CLOSE_SUBPROCESS_COPY(err_subprocess, 1);
       }
+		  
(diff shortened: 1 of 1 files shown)


pmatos:
Sorry for the delay on this but I have been busy with other stuff. Will try to close all my open PRs asap, starting with this one.

pmatos:
Anyway, I will follow your suggestion but generally I like to rely on free(NULL) being a no-op. In this specific case if you do `void *env = NULL;` right in the beginning and then you can remove:
```
else 
  env = NULL;
```

and replace all the `if (env) free(env);` by `free(env)`. Anyway, just my two cents. 
I will submit a PR following your suggestion.

PR #2294, opened 2018-10-01
Clarify the use of foff as a base pointer in foreign malloc

### LINKED ISSUE #2285: Possibility of null or incorrect memcpy

pmatos:
Fixes #2285.

### DIFF
--- a/racket/src/foreign/foreign.rktc
+++ b/racket/src/foreign/foreign.rktc
@@ -1914,8 +1914,19 @@ static void* SCHEME2C(const char *who,
   }
   res = scheme_malloc_fail_ok(mf,size);
   if (failok && (res == NULL)) scheme_signal_error("malloc: out of memory");
-  if (((from != NULL) || (foff != 0)) && (res != NULL))
-    memcpy(res, W_OFFSET(from, foff), size);
+
+  /* We might want to use foff as the object base address if from is NULL,
+   * therefore set a src point to use in memcpy to clarify this */
+  {
+    void *src = NULL;
+    if (from != NULL)
+      src = W_OFFSET(from, foff);
+    else if (foff != 0)
+      src = (void *)foff;
+    if (src != NULL && res != NULL)
+      memcpy(res, src, size);
+  }
+  
   if (SAME_OBJ(mode, raw_sym))
     return scheme_make_foreign_external_cpointer(res);
   else
--- a/racket/src/foreign/foreign.c
+++ b/racket/src/foreign/foreign.c
@@ -2675,8 +2675,19 @@ static Scheme_Object *foreign_malloc(int argc, Scheme_Object *argv[])
   }
   res = scheme_malloc_fail_ok(mf,size);
   if (failok && (res == NULL)) scheme_signal_error("malloc: out of memory");
-  if (((from != NULL) || (foff != 0)) && (res != NULL))
-    memcpy(res, W_OFFSET(from, foff), size);
+
+  /* We might want to use foff as the object base address if from is NULL,
+   * therefore set a src point to use in memcpy to clarify this */
+  {
+    void *src = NULL;
+    if (from != NULL)
+      src = W_OFFSET(from, foff);
+    else if (foff != 0)
+      src = (void *)foff;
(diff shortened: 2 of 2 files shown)


pmatos:
I should point out that Travis' failure is not due to the patch but due to a `tcp-connect` failure:
```
30.74s$ raco test -l tests/openssl/https
raco test: "/Users/travis/build/racket/racket/pkgs/racket-test/tests/openssl/https.rkt"
raco test: @(test-responsible 'mflatt)
tcp-connect: connection failed;
 host not found
  address: api.github.com
  port number: 443
  system error: nodename nor servname provided, or not known; gai_err=8
  context...:
   /Users/travis/build/racket/racket/racket/collects/openssl/mzssl.rkt:1700:0: ssl-connect
   /Users/travis/build/racket/racket/racket/collects/net/http-client.rkt:67:0: http-conn-open!10
   /Users/travis/build/racket/racket/racket/collects/net/http-client.rkt:272:0
   /Users/travis/build/racket/racket/racket/collects/racket/contract/private/arrow-val-first.rkt:428:3
   /Users/travis/build/racket/racket/racket/collects/net/url.rkt:195:0: http://getpost-impure-port
   /Users/travis/build/racket/racket/racket/collects/net/url.rkt:265:0: getpost-pure-port
   /Users/travis/build/racket/racket/racket/collects/racket/contract/private/arrow-val-first.rkt:428:3
   "/Users/travis/build/racket/racket/pkgs/racket-test/tests/openssl/https.rkt": [running body]
   temp37_0
   for-loop
   run-module-instance!125
   /Users/travis/build/racket/racket/pkgs/compiler-lib/compiler/commands/test.rkt:179:16
https.rkt: raco test: test raised an exception
The command "raco test -l tests/openssl/https" exited with 1.
```

samth:
I re-ran that one build and it passed.

racket/src/foreign/foreign.rktc
```diff
@@ -1914,8 +1914,17 @@ static void* SCHEME2C(const char *who,
   }
   res = scheme_malloc_fail_ok(mf,size);
   if (failok && (res == NULL)) scheme_signal_error("malloc: out of memory");
-  if (((from != NULL) || (foff != 0)) && (res != NULL))
-    memcpy(res, W_OFFSET(from, foff), size);
+
+  /* We might want to use foff as the object base address if from is NULL,
+   * therefore set a src point to use in memcpy to clarify this */
+  void *src = NULL;
```
mflatt: To accomodate the declaration and stick to C89 (which does matter for some of our supported build environments), wrap the new code with curly braces. Another possibility is to move the `src` declaration to the start of the function, but a local block seems better.

mflatt:
Thanks for updating "foreign.rktc" instead of "foreign.c". But the generated file should also be updated in the final commit, since the generated version is meant to be checked in (i.e., it's not automatically generated by the makefile).

racket/src/foreign/foreign.rktc
pmatos: Sure. Which build environments still require C89? Just out of curiosity...

pmatos:
Added the C89 definition block and regenerated the `c` file.

PR #2295, opened 2018-10-02
Fix single-character typo in docs

dalkire:
Fixing a one-character typo in the `Overview` section of the `Inside: Racket C API` docs.

### DIFF
--- a/pkgs/racket-doc/scribblings/inside/overview.scrbl
+++ b/pkgs/racket-doc/scribblings/inside/overview.scrbl
@@ -11,7 +11,7 @@ reclamation of memory used during evaluation, and the scheduling of
 concurrent threads and parallel tasks.
 
 Much of the language provided by @racketmodname[racket/base] is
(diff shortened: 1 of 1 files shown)


mflatt:
If I'm reading right, the work should be "that". It's a primitive dialect from the run-time system, and not something more primitive than the run-time system. Probably the sentence needs to be rephrased to be more clear.

dalkire:
You are probably right. I didn't understand what the "more primitive" comparator was referring to. I read it as the second paragraph in contrast to the first instead of being "more primitive" than something else (the rest of the language? another portion of `racket/base` that isn't as primitive?). It was unclear to me and I was eager to contribute something even so minor. Thanks for the quick response.

mflatt:
After looking again, I think the paragraph should just go away. Some of the future that it predicts has already come to pass in v7.0, but that's not relevant after all.

dalkire:
Thanks for the fix.

PR #2297, opened 2018-10-02
Add compilation with ubsan to test matrix.

pmatos:
We should probably add compilation with ubsan to the test matrix since it will catch potentially relevant issues during the build, like the issue fixed in #2296.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,6 +20,9 @@ matrix:
       --disable-extflonum"
   - os: linux
     compiler: gcc
+    env: PATH=./racket/bin:$PATH RACKET_CONFIGURE_ARGS="--enable-ubsan"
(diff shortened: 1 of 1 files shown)


pmatos:
Argh, travis-ci gcc is way too old for ubsan. Maybe we can get it in azure instead. Will take a look.

pmatos:
Unfortunately there's no way to enable this in travis. Maybe in azure. I will close this PR and try it in azure.

More on this, travis ubuntu environment is old... *old*! Really! Therefore the included gcc doesn't have all the goodness of ubsan libraries. Azure seems to provide a more recent environment so we might have more luck there. I will open a new PR with changes to the azure tests if I can get it to work.

PR #2302, opened 2018-10-08
Switch default _enum backing from _ufixint to _fixint

LeifAndersen:


mflatt:
Although this change would be sensible, I worry about potential compatibility issues. Since `_enum` lets a programmer pick the base type when they want to use negative values, the improvement doesn't seem significant enough to risk problems.

LeifAndersen:
I ran this through a bunch of programs and it hasn't broken anything.

I also only proposed this after it bit me (FFmpeg uses enums with negative numbers, and casting to uint caused it to break on the case of 0xffffff.) So, even if we don't fix it, I certainly think we should add a note in the docs about the different back end from C. Thoughts?

mflatt:
I think it's not actually the case that an enumerated type corresponds to `int`. See http://c0x.coding-guidelines.com/6.7.2.2.html 1147-1148. My understanding is that you can get an unsigned enumerated type in C if all the declared constants are non-negative. But, weirdly, individual names defined in the `enum` all have type `int`.

Probably it would have made sense for `_enum` to use `_fixint` if a negative number appears in the given list – but that's still a compatibility change, since the current interface sortof coerces negative values to unsigned.

At this point, I think it's best to have the documentation alert readers to the `_ufixint` choice and note that `_fixint` is needed to support negative values.

LeifAndersen:
I'm presuming you meant 1447-1448 (instead of 1147-1148, which isn't on the page you linked to).

Yes, that is stating the the implementation can be either signed or unsigned (or a char), and is implementation dependent. But the line "but shall be capable of representing the values of all the members of the enumeration", when combined with 1440 "The expression that defines the value of an enumeration constant shall be an integer constant expression that has a value representable as an int.", indicates that a signed integer will always be big enough to fit the expressible values, where as a char or signed it won't.

> since the current interface sortof coerces negative values to unsigned.

Eh, honestly not in any meaningful way (imo anyway). Consider:

```
(define _myenum
  (_enum '(unkown = -1
           left = 0
           right = 1)))
```

```
(define x (cast 'unkown _myenum _fixint))
(cast x _fixint _myenum)
```
Gives the error: 
```
enum:int->_myenum: expected a known #<ctype:ufixint>, got: 4294967295
```
(also x is bound to -1)


We get the same error when replacing `_fixint` here with `_ufixint`, except now x is bound to #xffffffff.

As such, it doesn't seem possible to cast an int to a myenum value of `'unkown`. So maybe it would make sense to have the _enum function check the range of the list and see if it _can_ be represented by the backing type, and if not throw an error?

mflatt:
Ok – signenedd in C is always confusing to me, and `_enum` currently does weird things with negative numbers, and those things lead me to be conservative. But if there's really no way to detect the change from `_ufixint` to `_fixint`, then ok. 

Note that the `history` note is currently outside the curly braces for `_enum`. Also, the note should be something like `@elem{Changed default @racket[basetype] from @racket[_ufixint] to @racket[_fixint].}` — using `@racket` around identifiers and and ending with a period.

LeifAndersen:
> Ok – signenedd in C is always confusing to me

FWIW, I think its somewhat confusing to everyone. :)

> But if there's really no way to detect the change from _ufixint to _fixint, then ok.

I don't think there is, but honestly, after thinking about it a bit more, I don't think that's the right thing to do either. And because of that, I'm inclined just to add the note and to move on for now.

> Note that the history note is currently outside the curly braces for _enum. Also, the note should be something like @elem{Changed default @racket[basetype] from @racket[_ufixint] to @racket[_fixint].} — using @racket around identifiers and and ending with a period.

Ah, good catch, thanks.

LeifAndersen:
Closing this PR and moving it to: https://github.com/racket/racket/pull/2308

PR #2306, opened 2018-10-09
Clarify match's app pattern behavior

RenaissanceBug:
If I have understood the behavior of ```app``` correctly, the documentation needs to be clarified. Currently it implies that match expects a single result, and one could reasonably infer that the matcher will require that single result to satisfy all ```_pat```s given. The error message resulting from a reasonable-but-wrong pattern like ```(app string->number (? number?) x)``` is inscrutable, so a doc fix is important here.

I have also added an example that illustrates a use case for ```app```, namely, the use case that led me to submit this PR.


samth:
I've merged this, the CI failures were fixed elsewhere by @mflatt. Thanks!

PR #2307, opened 2018-10-09
Pkg fixes

abmclin:
Various fixes for the `pkg` collection.

Fixes file:// URL handling when installing in alternative strip modes.

Added `directory-exists?` checks to `generate-stripped-directory` and `check-strip-compatible`

Test cases added for all the changes.

Documentation also updated with notes and missing information.

I have more `pkg` fixes in the pipeline but thought I'd best push these out in the hope of making the deadline before the latest Racket release is finalized.


abmclin:
Making the adjustment. Is it still OK to explicitly mention directory-exists? as the expected result when raising the argument error exception?

abmclin:
Ahh yes I do see the changes merged into the mainline so I can close this PR.

PR #2308, opened 2018-10-10
Note that _enum in Racket and enum in C use different backings.

LeifAndersen:
(It is worth noting that C does not specify the backing for enum, but
does say its range is that of a uint.)

### DIFF
--- a/pkgs/racket-doc/scribblings/foreign/types.scrbl
+++ b/pkgs/racket-doc/scribblings/foreign/types.scrbl
@@ -1738,10 +1738,10 @@ is to throw an exception.
              buffer_too_small)))
 ]
 
-Note that the default basetype is @racket[_fixint]. This
+Note that the default basetype is @racket[_ufixint]. This
 differs from C enumerations that can use any value in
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/foreign/types.scrbl
```diff
@@ -1736,7 +1736,19 @@ is to throw an exception.
     (_enum '(ok = 0
              invalid_input
              buffer_too_small)))
-]}
+]
+
+Note that the default basetype is @racket[_fixint]. This
+differs from C enumerations that can use any value in
+@racket[_ufixint]. Any @racket[_enum] using negative values
+should use @racket[_ufixint] for the base type.
```
lexi-lambda: This seems backwards: shouldn’t this say that the default base type is `_ufixint`, and users who wish to use negative values need to use `_fixint`?

LeifAndersen: Woops, yes, you are 100% correct.

PR #2312, opened 2018-10-15
Simplify copy-port

pmatos:
Simplify copy-port by removing the several `(let loop...` constructs.

I tried to read the copy and felt that the code should use the `for` constructs when possible instead of the `let loop`. I have a further patch to add a `flush?` option which will send in if you accept this or rewrite it into the old code if you don't.


racket/collects/racket/private/port.rkt
```diff
@@ -110,37 +110,36 @@
...
+    (cond
+      [(number? c)
+       (for ([dest (in-list dests)])
+         (let write-loop ([bytes-written 0])
+           (unless (= bytes-written c)
+             (define c2 (write-bytes-avail s dest bytes-written c))
+             (write-loop (+ bytes-written c2)))))
+       (loop)]
+    [(procedure? c)
+     (let ([v (let-values ([(l col p) (port-next-location src)])
```
samth: This could be rewritten as 

```racket
(define-values (l col p) (port-next-location src))
(define v (c (object-name src) l col p))
(for ...)
```

pmatos: True, I like that. I will push a new patch.

PR #2315, opened 2018-10-16
Fix fasl read fasl string

pmatos:
Add a new test for the failing case.

I noticed the bug with this example:
```
#lang racket

(require racket/fasl)

(define bn 15511210043330985984000000)
(define ser (s-exp->fasl bn))
(define newbn (fasl->s-exp ser))

(printf "~a -> ~e -> ~a~n" bn ser newbn)
```

Racket passes the test with the fix, but fails in the current HEAD.


pmatos:
@mflatt Can we still get this into 7.1?

mflatt:
How about
```
(define (read-fasl-string i [len (read-fasl-integer i)])
  (define bstr (read-bytes/exactly len i))
  (bytes->string/utf-8 bstr))
```
?

mflatt:
I'm pretty sure that change is a good idea, so I'll merge an adjusted commit.

pmatos:
@mflatt, The change is great, of course. I don't think you need to permission to change my patches. ;) Thanks for merging!

Issue #2314, opened 2018-10-16
Runtime errors during core tests


pmatos:
I found a few runtime errors, thanks to ubsan, that should be fixed. At the moment I have no time so this is a reminder for me, or a note to anyone else who would like to tackle them.

If you compile with ubsan and run the core tests you'll see:

- [x] Section *reading*: `../../../racket/gc2/../src/numstr.c:448:18: runtime error: signed integer overflow: 266666666 * 10 cannot be represented in type 'int'` - fixed by #2319
- [x] Section *unsafe*: `../../../racket/gc2/../src/number.c:5508:11: runtime error: left shift of 1 by 63 places cannot be represented in type 'long int'` - fixed by #2653 
- Section *fixnum*:
  - [ ] `../../../racket/gc2/../src/numarith.c:1271:7: runtime error: signed integer overflow: 9172034182905281024 + 1268111012645630976 cannot be represented in type 'long int'`
  - [ ]  `../../../racket/gc2/../src/numarith.c:1272:7: runtime error: signed integer overflow: 7193340147698933760 - -2058250744780168704 cannot be represented in type 'long int'`
  - [ ] `../../../racket/gc2/../src/numarith.c:1273:7: runtime error: signed integer overflow: -4508138700560545280 * -3892215211304724480 cannot be represented in type 'long int'`

This is something I will definitely return to but not in the next few days.

sorawee:
Fixed.

PR #2319, opened 2018-10-17
Avoid integer overflow

pmatos:
A runtime error was detecting and overflow on e which this fixes.
Fixes part of #2314.

### DIFF
--- a/racket/src/racket/src/numstr.c
+++ b/racket/src/racket/src/numstr.c
@@ -445,7 +445,10 @@ static double STRTOD(const char *orig_c, char **f)
 	if (!isdigit(ch))
 	  return 0; /* not a digit - bad! */
 	else {
-	  e = (e * 10) + (ch - '0');
+          /* only increment e until we know if it is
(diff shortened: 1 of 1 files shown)


pmatos:
A note on this patch. The obvious patch was to me initially to do:
```
	  if (e > CHECK_INF_EXP_THRESHOLD(0)) {
	    if (neg_exp || !is_nonzero)
	      is_zero  = 1;
	    else
	      is_infinity  = 1;
	    break;
	  }
```

The reason is that if we break the bytes in `c` are not consumed and we get an internal error. So we need to keep consuming the bytes even after detecting a zero or infinity. The check ensure that even if we know the result, we don't try to modify `e` anymore.

pmatos:
The integer overflow, by the way, can be reproduced with:
```
$ racket -e .3e2666666666
```

if your racket has been compiled with ubsan you'll see:
```
../../../racket/gc2/../src/numstr.c:448:18: runtime error: signed integer overflow: 266666666 * 10 cannot be represented in type 'int'
```

PR #2323, opened 2018-10-18
Added flush? to copy-port and documentation

pmatos:
The other day I was discussing with the @rmculpepper on slack about this and he suggested adding `#:flush?` to `copy-port`, which I did straight away on my version of `copy-port` (which was submitted as #2312). This is a further addition to that.

I add code and documentation.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/port-lib.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/port-lib.scrbl
@@ -819,7 +819,7 @@ a conversion error occurs at any point while reading from @racket[in], then
 @exnraise[exn:fail].}
 
 
-@defproc[(copy-port [in input-port?] [out output-port?] ...+) void?]{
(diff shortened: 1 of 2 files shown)


rmculpepper:
After reading the docs more closely, I think I was wrong earlier: `write-bytes-avail` is supposed to write *and flush* as much as it immediately can. So there shouldn't be any need for a separate `flush-output` call. That leaves two questions:

1. Did adding an extra `flush-output` fix your original problem? If so, is there a type of output port for which `write-bytes-avail` doesn't work as advertised?

2. Is there any advantage in doing `write-bytes-avail`in a loop over the following code?

      (begin (write-bytes s dest 0 c) (flush-output dest))

pmatos:
@rmculpepper well spotted. after reading the docs properly on `write-bytes-avail` I am now confused. Let me take a look at my example and I will get back on this.

pmatos:
@rmculpepper is correct about `write-bytes-avail` writing and flushing, so no need for this PR. Apologies, my mistake was probably something else. Something now I can't actually reproduce anymore. Will close.

Issue #2018, opened 2018-03-26
 hang building racket 6.12 on s390x


bremner:
It seems that raco setup is hanging during the build of 6.12 on
Debian/s390x [1].

The last lines look like:
``
racket/racket3m -X "/<<BUILDDIR>>/racket-6.12+dfsg1/debian/tmp/usr/share/racket/collects" -G
"/<<BUILDDIR>>/racket-6.12+dfsg1/debian/tmp/etc/racket"   --no-user-path -N "raco" -l- setup --no-user -j 2 --no-docs
--no-zo 
raco setup: bootstrapping from source...
 triggered by use of non-".zo" file
 path: /<<BUILDDIR>>/racket-6.12+dfsg1/debian/tmp/usr/share/racket/collects/compiler/cm.rkt
``

I can duplicate this interactively. Any suggestions to debug?

[1] https://buildd.debian.org/status/fetch.php?pkg=racket&arch=s390x&ver=6.12%2Bdfsg1-1&stamp=1520860407&raw=0

pmatos:
Does this also happen on HEAD? I will try to see if I can reproduce this on a s390x. First I will need to get my hands on one though.

bremner:
Paulo Matos <notifications@github.com> writes:

> Does this also happen on HEAD? I will try to see if I can reproduce
> this on a s390x. First I will need to get my hands on one though.
>

HEAD build fails at 

./racketcgc -cqu ../../racket/mksystem.rkt system.rktd "gcc -E -I. -I../../racket/include -I../../racket/src -g -O2    -DUSE_SENORA_GC    ../../racket/src/systype.c" "" "./racketcgc" "./racketcgc"
current-module-name-resolver: contract violation
  expected: (and/c (procedure-arity-includes/c 2) (procedure-arity-includes/c 4))
  given: #<procedure:standard-module-name-resolver>
Segmentation fault
make[8]: *** [Makefile:177: sysinfercgc] Error 139

I assume that is also arch specific. HEAD is currently building on my
amd64 desktop, and reached the "building documentation" stage.

pmatos:
@bremner  thanks for checking. Before I go into the pain of finding a s390x, @mflatt is s390x a supported architecture? Is this relevant?

mflatt:
Yes, s390x should be supported. There will be no JIT, but the build should work. I don't recall whether it has ever been known to work, though.

[...]

PR #2325, opened 2018-10-19
Gitlab integration for ubsan compilation

pmatos:
In order to solve:
1. compilation/testing of racket using ubsan (see #2297); 
2. compilation/testing of racket using qemu (see #2018);

I have tried several approaches through the last few weeks - from installing new versions of qemu and gcc inside travis ci to doing the same things in azure. Unfortunately both platforms are relatively old. We can of course, go the path of trying to create `apt-repos` (ubuntu package repos) with recent versions of qemu and gcc but that's just too much work. 

I have large experience with Gitlab and Gitlab CI as I use it for many years for work and pleasure and at Linki Tools I manage our own self-hosted instance, so going this way was easy for me once I remembered that recently Gitlab added the option to use Gitlab CI on Github projects.

There are many advantages of using Gitlab but for this specific user case I will mention only one: in gitlab ci, you can use whatever docker image you want. Therefore you can, in this specific case, just test on ubuntu 18.04 LTS which has all the tools we need out of the box. 

The only thing that the racket repo needs is a .gitlab-ci.yml which is similar to what other CI systems need. But for this to work someone on the racket side (@samth are you the PLT CI manager?) we need the following done once:
1. Go to https://gitlab.com/users/sign_in and create a user or sign in;
2. Create a racket group in gitlab.com  so you own `gitlab.com/racket`;
3. Create a project but when doing so choose to create 'From Github for CI` - this will create a read-only mirror that just uses the CI runners of gitlab;
4. Connect the Gitlab and Github instances so that Gitlab tests run for PRs in github - you've surely done something similar for azure and appveyour;

More information here: https://docs.gitlab.com/ee/ci/ci_cd_for_external_repos/github_integration.html

I have done this for my LinkiTools/racket repo here:
https://gitlab.com/LinkiTools/racket
 and the results of the CI performed by Gitlab are here:
https://gitlab.com/LinkiTools/racket/pipelines/33538582

As you can see each ubsan and ubsan-cs job create an artifact with the ubsan warning. If we go forward with this, we can mark the job as failing if there are runtime errors and mark it also as allowing failure so it's an amber instead of red. Hopefully it'll be green once #2314 is fixed.

I could do all the steps above except 4. because I don't have access to the racket github application token. I am happy to help, let me know if there's anything else you need.

Once this file is in place, the next step is to add qemu cross-platform testing.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,0 +1,72 @@
+image: ubuntu:18.04
+
+.prepare:
+  before_script:
+    - apt-get update && apt-get install -y git lsb-release gcc gcc-8 g++ g++-8 make libfontconfig1-dev libcairo2-dev openssl libpango1.0-dev libjpeg-turbo8-dev libncurses5-dev
+    - update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-7 700 --slave /usr/bin/g++ g++ /usr/bin/g++-7
+    - update-alternatives --install /usr/bin/gcc gcc /usr/bin/gcc-8 800 --slave /usr/bin/g++ g++ /usr/bin/g++-8
+    - export PATH=$PWD/racket/bin:$PATH
+
+envinfo:
+  extends: .prepare
+  script:
+    - cat /proc/cpuinfo
+    - lsb_release -a
+    - gcc -v
+    - export
+
+test:ubsan:
+  extends: .prepare    
+  script:
+    - mkdir logs
+    - make CPUS="2" PKGS="racket-test db-test unstable-flonum-lib net-test" CONFIGURE_ARGS_qq="--enable-ubsan CFLAGS="-fno-var-tracking-assignments"" in-place 2>&1 | tee logs/build.log
+    - raco test -l tests/racket/test 2>&1 | tee logs/test.log
+    - racket -l tests/racket/contract/all 2>&1 | tee logs/contract-test.log
+    - raco test -l tests/json/json 2>&1 | tee logs/json-test.log
+    - raco test -l tests/file/main 2>&1 | tee logs/file-test.log
+    - raco test -l tests/net/head 2>&1 | tee logs/net-head-test.log
+    - raco test -l tests/net/uri-codec 2>&1 | tee logs/net-uri-codec-test.log
+    - raco test -l tests/net/url 2>&1 | tee logs/net-url-test.log
+    - raco test -l tests/net/url-port 2>&1 | tee logs/net-url-port-test.log
+    - raco test -l tests/net/encoders 2>&1 | tee logs/net-encoders-test.log
+    - raco test -l tests/openssl/basic 2>&1 | tee logs/openssl-basic-test.log
+    - raco test -l tests/openssl/https 2>&1 | tee logs/openssl-https-test.log
+    - raco test -l tests/match/main 2>&1 | tee logs/match-main-test.log
+    - raco test -l tests/zo-path 2>&1 | tee logs/zo-path-test.log
+    - raco test -l tests/xml/test 2>&1 | tee logs/xml-test.log
+    - raco test -l tests/db/all-tests 2>&1 | tee logs/db-test.log
+    - raco test -c tests/stxparse 2>&1 | tee logs/stxparse-test.log
+  after_script:
+    - grep 'runtime error' logs/*.log > runtime-errors.log
+  artifacts:
+    paths:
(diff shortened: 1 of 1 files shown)


samth:
I would like to make this work, but I'm wary of adding yet another CI service. Using docker does sound nice, but can we use the [Travis docker support](https://docs.travis-ci.com/user/docker/)? Alternatively, would running this on the Azure Linux CI work?

pmatos:
@samth I understand however, in my opinion it's a waste of time attempting these things in travis. I looked into this a couple of months ago and failed to get it to work on both azure and travis. I looked at it again today quickly and reached the same conclusion - the services are way too constrained.

Even the travis docker support allows you to run docker images in the test but doesn't allow to run the test in a docker image which is what we need. 

I have been using Gitlab for my business for close to 5 years and I haven't seen better CI support. The fact that they added CI support for Github projects is just amazing and would allow us to do a huge amount of things that otherwise we can't. I would be happy to get this up and running if you wish.

With Gitlab would be possible to:
* test different operating systems and different linux distros;
* test different architectures through qemu;
* run static analysis with clang scan (coverity scan stopped providing analysis for free for now due to some internal failures);

I really think it would open us several doors which at the time are closed because azure and travis constrain the images we can run on those systems.

jackfirth:
For a while now I've been successfully using CircleCI to build and test racket code with and inside docker. They allow workflows with multiple jobs that each run different images, and job steps can share state with build artifacts. I'm not sure if it would be easier or harder than gitlab, but it could be an option.

jackfirth:
I don't think it has anything like that, as far as I know (not very). Gitlab CI sounds pretty great.

pmatos:
@jackfirth GitLab is pretty good. There are some features missing when compared to GitHub dashboard, projects, exploration etc. But CI-wise is light years ahead.

And that's why I will keep pestering @samth to either consider this or tell me to shut up. :)

pmatos:
I wanted to quickly point out the results obtained with the `gitlab-integration` branch. 

You can see the pipeline here: https://gitlab.com/LinkiTools/racket/pipelines/46493679
At the moment it's more of a sequence of parallel tasks but it can easily become a pipeline if needed.

One test failed: I opened #2461 
For the `test:ubsan` which passed, I store an artifact.
Check the job page: https://gitlab.com/LinkiTools/racket/-/jobs/157664139
Click on the Browse artifact button and you get the error generated by ubsan for this build:
```
 logs/test.log:../../../racket/gc2/../src/number.c:5501:11: runtime error: left shift of 1 by 63 places cannot be represented in type 'long int'
logs/test.log:../../../racket/gc2/../src/numarith.c:1246:7: runtime error: signed integer overflow: 8257794038359934464 + 3987006104043242496 cannot be represented in type 'long int'
logs/test.log:../../../racket/gc2/../src/numarith.c:1247:7: runtime error: signed integer overflow: 9084217241503894528 - -2277789645129747968 cannot be represented in type 'long int'
logs/test.log:../../../racket/gc2/../src/numarith.c:1248:7: runtime error: signed integer overflow: 3690147173609477120 * 729023046779805696 cannot be represented in type 'long int'
```

Another interesting case is the building of racket with `scan-build`, the static analyser of the LLVM toolchain. That's job `scan-build:racket`. All compiles and runs and artifacts are again stored. This time, download the artifact and store it somewhere, then open the `index.html` inside the extracted folder (you have to go up 2 folder levels).

You'll see something like this:
2019-02-07-114149_993x1052_scrot

This is just a datapoint of what is possible with a few lines of Gitlab configuration. A lot more can be done. I really think a project like racket would benefit from this flexibility.

samth:
I tried doing this on GitLab today, but I've run into a couple issues. First, GitLab just errors when I tried to connect `racket/racket`. I successfully connected `samth/racket` and tried running CI but it always fails to start.

pmatos:
Thanks for looking into this. I am sorry you have faced some problems. Let me give it a try. If I succeed, I will add you as admin. Since there's already a `racket` group I will create it as `racketlang`.

pmatos:
I have added you as an owner of `gitlab.com/racketlang`. Unfortunately I cannot connect the repo to github because i am not in the racket github group. Can you go to: https://gitlab.com/racketlang then try 'New Project', CI/CD for external repo, Github, List my github repos and press connect for `racket/racket`. Does it work?

samth:
I did that connection and it's in process importing.

pmatos:
Great. Shouldn't take long then.

PR #2327, opened 2018-10-19
fix flush-output when write-out is port indirection

rmculpepper:
On RacketCS, `flush-output` currently fails on ports where the write-out field is another output port. For example, the following program raises an error:

    #lang racket/base
    (define real-out (open-output-bytes))
    (define wrapped-out (make-output-port 'wrapper real-out real-out void))
    (write-bytes #"hello" wrapped-out)
    (flush-output wrapped-out)
    (get-output-bytes real-out)
    ;; application: not a procedure;
    ;;  expected a procedure that can be applied to arguments
    ;;   given: #<output-port:string>

### DIFF
--- a/racket/src/io/port/flush-output.rkt
+++ b/racket/src/io/port/flush-output.rkt
@@ -10,10 +10,15 @@
 
 (define/who (flush-output [p (current-output-port)])
   (check who output-port? p)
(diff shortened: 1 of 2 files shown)


mflatt:
This looks right to me - thanks!

PR #2328, opened 2018-10-21
 Avoid undefined behavior in unsafe-fxlshift test

gus-massa:
The first commit should fix the second item in https://github.com/racket/racket/issues/2314 (@pmatos: can you check that it is really fixed?). I moved the test from _unsafe.rktl_ to _fixnum.rktl_ because it's more similar to the test there. Also, I changed the test because it was checking that the unspecified result was `0`. Anyway, I think it is important to check this in case something similar appears in a dead branch that is never run but it must be compiled without problems. In particular, it's not clear for me that the undefined cases in the RacketVM, RacketCS, and C are the same, so I made a few test above and below the tricky values.

The second commit is to extend the table of values of `(fixnum-width)` to be used in the CS version. This related to https://github.com/racket/racket/pull/2266 . I also added a few sanity checks at the beginning of the file.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/fixnum.rktl
+++ b/pkgs/racket-test-core/tests/racket/fixnum.rktl
@@ -4,11 +4,30 @@
          scheme/unsafe/ops
          "for-util.rkt")
 
-(define 64-bit? (fixnum? (expt 2 33)))
+(test #t fixnum? (- (expt 2 (- 24 1))))
+(test #t fixnum? (- (expt 2 (- 24 1)) 1))
 
-(define (fixnum-width) (if 64-bit? 63 31))
-(define (least-fixnum) (if 64-bit? (- (expt 2 62)) -1073741824))
-(define (greatest-fixnum) (if 64-bit? (- (expt 2 62) 1) +1073741823))
+(define (fixnum-width)
+  (case (system-type 'word)
+    [(32) (case (system-type 'vm)
+            [(racket) 31]
+            [(chez-scheme) 30]
+            [else (error "unsuported vm value in system-type")])]
+    [(64) (case (system-type 'vm)
+            [(racket) 63]
+            [(chez-scheme) 61]
+            [else (error "unsuported vm value in system-type")])]
+    [else (error "unsuported word value in system-type")]))
+                      
+(define (least-fixnum) (- (expt 2 (- (fixnum-width) 1))))
+(define (greatest-fixnum) (- (expt 2 (- (fixnum-width) 1)) 1))
+
+(test #t fixnum? (least-fixnum))
+(test #t fixnum? (greatest-fixnum))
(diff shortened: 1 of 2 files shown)


pmatos:
Thanks @gus-massa , I will take a look at your first commit and update you on it.

pmatos:
Thanks @gus-massa , I will take a look at your first commit and update you on it.

pmatos:
Sorry @gus-massa for the delay on this. Just checked but the test still shows a runtime error.
You can see it here: https://gitlab.com/LinkiTools/racket/-/jobs/110078737

pmatos:
Oh wait... I just noticed, this is a PR and I should have checked your branch instead. Argh... sorry.. let me get back to you.

pmatos:
So, I compiled and tested your racket branch `18-10-Fixum`. I can still see the problem. Here's what I did:
```
$ git clean -xdf && make CPUS="16" PKGS="racket-test db-test unstable-flonum-lib net-test" CONFIGURE_ARGS_qq="--enable-ubsan --disable-strip CFLAGS='-O0 -ggdb'" in-place
$  racket/bin/raco test -l tests/racket/test 2>&1 | tee test-output.log
```

I attach the log I got.
[test-output.log](https://github.com/racket/racket/files/2701951/test-output.log)

You can see in it:
```
../../../racket/gc2/../src/number.c:5508:11: runtime error: left shift of negative value -4611686018427387904
../../../racket/gc2/../src/numarith.c:1272:7: runtime error: signed integer overflow: 5140707656353197568 - -4241117877892239872 cannot be represented in type 'long int'
../../../racket/gc2/../src/numarith.c:1271:7: runtime error: signed integer overflow: -6419906316588973568 + -4318998038310943744 cannot be represented in type 'long int'
../../../racket/gc2/../src/numarith.c:1273:7: runtime error: signed integer overflow: 1597552545817332736 * -328316318257648640 cannot be represented in type 'long int'
../../../racket/gc2/../src/number.c:5482:14: runtime error: left shift of negative value -1
```

gus-massa:
It's very strange. I think that I fixed the problem but the errors you see now are from the other part of fixnum.rktl. I don't understand why the other part raises an error, because when there there is an escape continuation that is called when the safe version raise an error, so the unsafe part is never called. 

(I guess that it may be a weird combination of constant folding and some optimizations that removes the error handling code when there are no error. ????? I have to take a look at the optimized version of the code in case there is something interesting/wrong there.)

pmatos:
@gus-massa with #2653 merged how can we get this one sorted as well?

gus-massa:
To make future work easier, I rebased it and fixed a merge conflict. I think that @mflatt wanted to take a look before merging.

gus-massa:
Variants of both changes have been merged, so it's better to close this.

PR #2330, opened 2018-10-22
syntax/parse: fix Check-Syntax arrows within syntax classes

### LINKED ISSUE #2320: syntax/parse: Missing Check-Cyntax arrows in syntax classes in Racket 7.1
The Check-Syntax binding arrows for pattern variables seem to have disappeared in Racket version 7.1. In versions 7.0 and 7.0.0.1--2018-06-13, the program below shows a Check-Syntax arrow on `x`, and in versions 7.0.0.11--2018-08-10 and 7.1, the same program does not show any Check-Syntax arrow on `x` or any other pattern variables.

```racket
#lang racket/base
(require syntax/parse)

(define-syntax-class example
  [pattern x  #:attr output #'x])
```

The Check-Syntax arrows are also gone for syntax classes like `expr` in:
```racket
#lang racket/base
(require syntax/parse)

(define-syntax-class example
  [pattern x:expr  #:attr output #'x])
```

I'm not sure, but this might have been introduced at the same time as the fix for issue #2164, pull request #2220.

AlexKnauth:
Fixes #2320. This puts Check-Syntax arrows back in syntax class definitions.
screenshot-issue-2320

### DIFF
--- a/racket/collects/syntax/parse/private/parse.rkt
+++ b/racket/collects/syntax/parse/private/parse.rkt
@@ -104,7 +104,8 @@
       (let ()
         (define the-rhs
           (parameterize ((current-syntax-context #'ctx))
-            (fixup-rhs (syntax-local-eval #'the-rhs-expr)
+            (fixup-rhs (syntax-local-eval
+                        (syntax-local-introduce #'the-rhs-expr))
                        (syntax-e #'splicing?)
                        (syntax->datum #'relsattrs))))
         (rhs->parser #'name #'formals #'relsattrs the-rhs (syntax-e #'splicing?) #'ctx)))]))


rmculpepper:
Thanks @AlexKnauth for tracking this down. I think your fix can be simplified to 

    (fixup-rhs (syntax-local-eval (syntax-local-introduce #'the-rhs-expr)) _ _)

AlexKnauth:
The reason I don't like that as much... does that mean that bindings in the user's program could affect how the `syntax-local-eval` evaluates? Doing it after the eval makes more sense to me because then the bindings that determine the eval remain in the "macro"'s scope.

AlexKnauth:
What if the syntax class is put in a user context where the `racket/base` bindings aren't there, or they're overridden in the user's scope?

AlexKnauth:
In my own testing, I can't figure out how to trigger it to capture something it shouldn't either way. I don't really have a reason trust that `syntax-local-eval` when there's a `syntax-local-introduce` on the input. I can't find any situations where it does anything differently from putting `syntax-local-introduce/rhs` on the output...

Why do those identifiers still get the macro-scope and not the user-scope when the input to eval is introduced?

rmculpepper:
The references in the expression being evaluated have the scope(s) corresponding to the body of the "datum-to-expr.rkt" module, which imports racket/base. If you evaluate that expression in another context where some racket/base names are shadowed, it doesn't matter, because the expression doesn't have the scope corresponding to the shadowing binding.

Compare with

    (let ([x 'outer])
      (let-syntax ([use-x1 (lambda (stx) #'x)]
                   [use-x2 (lambda (stx) (syntax-local-introduce #'x))])
        (let ([x 'inner])
          (list (use-x1)
               (use-x2)))))
    ;; => '(outer outer)

In general, if an identifier resolves to a particular binding occurrence, then adding or removing a scope S, where S does not occur in any of the binding occurrences in the environment, does not change what the identifier resolves to.

The macro scope for `parser/rhs` does not occur in any of the binding occurrences in the environment at the time when `syntax-local-eval` is called, so it does not affect the resolution of the identifiers in the given expression during its compilation and evaluation; it only affects the scopes on the syntax objects produced by `quote-syntax` subexpressions. (Note that every identifier in an expression produced by `datum->expression` is either a reference, under a `quote`, or under a `quote-syntax`; so there's no need to consider the effect that toggling the scope has on binders.)

AlexKnauth:
Okay. I've changed it to follow your suggestion.

PR #2331, opened 2018-10-23
remove plumber flush callback when custodian closes fd output port 

rmculpepper:
The main commit is the second one. See the commit for an example.

Is it a problem that the `on-close` callback is not called on custodian shutdown? So far, the only use seems to be for TCP ports to implement the "abandon" operations, which shouldn't matter if both ports are getting closed anyway.


mflatt:
This looks right to me - thanks!

PR #2337, opened 2018-10-27
fix lookup-errno on RacketCS

rmculpepper:
This adds support for `lookup-errno` on RacketCS. The contents of `errno-data.ss` were generated by running a C program on each of the three platforms and concatenating the results. The program is a pretty straightforward modification of the relevant part of racket/src/foreign/foreign.rktc; I could check it in too, but I don't know where it should go.


mflatt:
Ok, but please add "errno-data.ss" to `RUMBLE_SRCS` in "Makefile". Thanks!

PR #2339, opened 2018-10-27
Remove set, but not used variable

pmatos:
Fixes a warning by gcc


mflatt:
Unless I misremember or misread the code, `GC_set` doesn't have a side effect, so the call should be removed along with the variable.

pmatos:
@mflatt completely missed your message and forgot about this PR. `GC_set` seems to have a side-effect. Initializing GC, if it's not yet initialized. Shouldn't we keep the call to `GC_set` in that case?

pmatos:
@mflatt ping

mflatt:
That call to GC_set definitely will be after the GC is initialized. I'll commit a change that removes the call.

Thanks!

PR #2342, opened 2018-10-28
add #:result clause to for/lists and for*/lists

LiberalArtist:


pkgs/racket-doc/scribblings/reference/for.scrbl
```diff
@@ -364,7 +379,7 @@ nested.
 ]}
 
 @deftogether[(
-@defform[(for*/list (for-clause ...) body-or-break ... body)]
+@defform[(for*/list (for-clause ... maybe-result) body-or-break ... body)]
```
lexi-lambda: This looks like the wrong place for this change—I think you meant to add this on the line below, instead, which documents `for*/lists`?

LiberalArtist: Thanks for spotting this! I'll fix that now.

PR #2348, opened 2018-11-01
cmdline: Implement a help message that shows the available choices

BourgondAries:
This does not at all check if the choice you give to the program is
correct. I'd like to open a discussion regarding the desired message to
provide to the user upon:

1. Failure to provide one of the "choices" flags
2. If we should allow regexp "choices"
3. How to handle rest arguments

### DIFF
--- a/racket/collects/racket/cmdline.rkt
+++ b/racket/collects/racket/cmdline.rkt
@@ -210,10 +210,27 @@
                        [args args])
            #'(parse-command-line program-name argv (list . table) . args))))))
 
+(define (car-or-itself l)
+  (if (pair? l) (car l) l))
+
 (define (print-args port l f)
   (let loop ([l l]
              [n 1])
     (unless (null? l)
+      (define optional? (procedure-arity-includes? f n))
+      (fprintf port " ~a<~a>~a"
+               (if optional? "[" "")
+               (car-or-itself (car l))
+               (if optional? "]" ""))
+      (when (and (null? (cdr l))
+                 (procedure-arity-includes? f (+ n 2)))
+        (fprintf port " ..."))
+      (loop (cdr l) (add1 n)))))
+
+(define (print-args-choices port l f)
+  (let loop ([l l]
+             [n 1])
+    (unless (or (null? l) (= (length l) 1))
       (define optional? (procedure-arity-includes? f n))
       (fprintf port " ~a<~a>~a"
                (if optional? "[" "")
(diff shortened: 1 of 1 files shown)


BourgondAries:
@elibarzilay

BourgondAries:
@samth the change allows for an optional whitelist of flag parameters, such that running the program with `-h` or `--help` displays - in a standard manner - a list of allowed flag arguments to a flag parameter.

It's something like this (don't have the code to run and check atm).
```
--example <flag>, <(flag whitelisted-1 whitelisted-2)> : Sets an example
```

Now this patch does not check if the supplied flags are part of this set, which can be another patch. This is inspired by `argparse`'s `choices`, which is very useful for ensuring the correct flag is put in, as well as informing the user in a standard way what flags are valid.

sorawee:
Hello @BourgondAries, 

As I understand, this PR adds a capability to `racket/cmdline` to suggest what the value an id could be. That is, instead of writing:

```
```

```

```

sorawee:
Hi @BourgondAries,

As I understand, you are adding a capability to `racket/cmdline` so that instead of writing:

```
[("--mode") mode "The mode of the game: must be either fast or slow" (set! current-mode mode)]
```

we can simply write something along the line:

```
[("--mode") mode #:choices ("fast" "slow") "The mode of the game" (set! current-mode mode)]
```

The benefits are that:
- There's a uniform and standardized help message
- the library can perform a check that the option that users provide are valid (not yet implemented in this PR)

While I do think that this is a nice feature, I do not understand how your implementation works. When I download your PR and test it, I find that it doesn't work as expected on existing code. For example, consider:

```
#lang racket

(require racket/cmdline)

(command-line
 #:multi
 [("--add") first-num second-num
            "Add <first-num> and <second-num> together"
            (println (+ (string->number first-num)
                        (string->number second-num)))])
```

With the current `racket/cmdline`, when I run `racket test.rkt --help`, the following text is printed:

```
* --add <first-num> <second-num> : Add <first-num> and <second-num> together
```

With your PR, the text changes to:

```
* --add <first-num> <second-num> : <first-num> Add <first-num> and <second-num> together
```

I do not think that this is correct.

I believe you need to modify `command-line` form to support this new option. There should also be tests. The docs should be modified to reflect the change.

racket/collects/racket/cmdline.rkt
```diff
@@ -300,7 +317,7 @@
                           (cadr line))])
             (unless (and (pair? h)
                          (or (string? (car h)) (andmap string? (car h)))
-                         (andmap string? (cdr h)))
+                         (andmap (lambda (x) (or (string? x) (andmap string? x))) (cdr h)))
```
sorawee, line 320: `(andmap string? x)` should be changed to `(and (list? x) (andmap string? x))` to avoid a bad error message from the library when a value that is neither string nor list is passed in.

BourgondAries:
@sorawee this is a really old patch by now. I don't need it since I don't use racket anymore. Feel free to take the code and make your own patch based off it. I'll close this.

PR #2350, opened 2018-11-01
Documentation: eval-model.scrbl

Paul-C-Anagnostopoulos:
I completed my work on eval-model.scrbl. Please pay particular attention to the first two paragraphs in the "Phases" section and send a rewrite if necessary.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/eval-model.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/eval-model.scrbl
@@ -166,7 +166,7 @@ then an algebra student simplifies @tt{x + 1} as follows:
 @verbatim{  x + 1 = 10 + 1 = 11}
 
 Racket works much the same way, in that a set of @deftech{top-level
-variables} are available for substitutions on demand during
+variables} (see also @secref["vars-and-locs"]) are available for substitutions on demand during
 evaluation. For example, given
 
 @racketblock[
@@ -416,9 +416,10 @@ an @tech{object}, so evaluating @racket[(f 7)] starts with a
 ]
 
 Unlike in algebra, however, the @tech{value} associated with an
-procedure parameter can be changed in the body of a procedure by using
+argument parameter variable (not to be confused with parameters;
+see @secref["parameter-model"]) can be changed in the body of a procedure by using
 @racket[set!], as in the example @racket[(lambda (x) (begin (set! x 3)
-x))]. Since the @tech{value} associated with parameter @racket[x] can be
+x))]. Since the @tech{value} associated with parameter variable @racket[x] can be
 changed, an argument value cannot be substituted for @racket[x] when
 the procedure is first applied.
 
@@ -451,7 +452,8 @@ not been used before and that cannot be generated again or
 accessed directly.
 
 Generating a @tech{location} in this way means that @racket[set!]
-evaluates for @tech{local variables}, including parameters, in the same way as for
+evaluates for @tech{local variables}, including argument
+parameter variables, in the same way as for
 @tech{top-level variables}, because the @tech{local variable} is
 always replaced with a @tech{location} by the time the @racket[set!]
 form is evaluated:
@@ -495,7 +497,7 @@ Evaluation of a local-variable form, such as @racket[(let ([x (+ 1
 that replaces every instance of @racket[x] in @racket[_expr].
 
 @;------------------------------------------------------------------------
-@section{Variables and Locations}
+@section[#:tag "vars-and-locs"]{Variables and Locations}
 
 A @deftech{variable} is a placeholder for a @tech{value}, and
 expressions in an initial program refer to @tech{variables}. A
@@ -503,8 +505,8 @@ expressions in an initial program refer to @tech{variables}. A
 @tech{location}. Any other @tech{variable} is always replaced by a
 @tech{location} at run-time; thus, evaluation of expressions
 involves only @tech{locations}. A single @deftech{local variable}
-(i.e., a non-top-level, non-module-level @tech{variable}), such as a
-procedure parameter, can correspond to different @tech{locations}
+(i.e., a non-top-level, non-module-level @tech{variable}), such as an
+argument parameter variable, can correspond to different @tech{locations}
 through different applications.
 
 For example, in the program
@@ -520,7 +522,7 @@ hold the value @racket[11].
 
 The replacement of a @tech{variable} with a @tech{location} during
 evaluation implements Racket's @deftech{lexical scoping}. For example,
-when a procedure parameter @tech{variable} @racket[x] is replaced by
+when an argument parameter variable @racket[x] is replaced by
 the @tech{location} @racket[xloc], it is replaced @italic{throughout} the
 body of the procedure, including any nested @racket[lambda]
 forms. As a result, future references to the @tech{variable} always
--- a/pkgs/racket-doc/scribblings/reference/eval-model.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/eval-model.scrbl
@@ -1,4 +1,6 @@
-#lang scribble/doc
+#lang scribble/manual
+@;;;#lang scribble/doc
+
 @(require scribble/struct scribble/racket "mz.rkt" "prog-steps.rkt"
           (for-syntax racket/base))
 
@@ -9,7 +11,6 @@
                   (make-element highlighted-color (list c))))
 @(define-syntax redex
    (syntax-rules () [(_ a) (*redex (racket a))]))
-
 
 @(define hole (make-element #f (list "[]")))
 @(define (*sub c e) (make-element #f (list c "[" e "]")))
@@ -29,6 +30,13 @@ to switch to the usual langle/rangle that is used in syntax definitions.
    (syntax-rules () [(_ a b) (*state (racket a) (racket b))]))
 ;}
 
+@;{
+I worked on sections 1--9. More work to come on them and the remaining
+sections.
+
+~~Paul C. Anagnostopoulos
+}
+
 @;------------------------------------------------------------------------
 @title[#:tag "eval-model"]{Evaluation Model}
 
@@ -44,11 +52,10 @@ Racket evaluation simplifies
 (+ 1 1) @#,reduces 2
 ]
 
-The arrow @reduces above replaces the more traditional @tt{=} to
-emphasize that evaluation proceeds in a particular direction towards
-simpler expressions. In particular, a @deftech{value} is an
-expression that evaluation simplifies no further, such as the number
-@racket[2].
+The arrow @reduces replaces the more traditional @tt{=} to
+emphasize that evaluation proceeds in a particular direction toward
+simpler expressions. In particular, a @deftech{value}, such as the number @racket[2],
+is an expression that evaluation simplifies no further.
 
 @;------------------------------------------------------------------------
 @section[#:tag "cont-model"]{Sub-expression Evaluation and Continuations}
@@ -60,25 +67,26 @@ Some simplifications require more than one step. For example:
 ]
 
 An expression that is not a @tech{value} can always be partitioned
-into two parts: a @deftech{redex}, which is the part that changed in a
+into two parts: a @deftech{redex} (``reducible expression''),
+which is the part that can change in a
 single-step simplification (highlighted), and the
 @deftech{continuation}, which is the evaluation
-context surrounding an expression. In @racket[(- 4 (+ 1 1))], the redex is @racket[(+ 1 1)], and
-the continuation is @racket[(- 4 @#,hole)], where @hole takes the
-place of the redex. That is, the continuation says how to ``continue''
+context surrounding the redex. In @racket[(- 4 (+ 1 1))], the redex is @racket[(+ 1 1)], and
+the continuation is @racket[(- 4 @#,hole)], where @hole indicates the
+position of the redex. That is, the continuation says how to ``continue''
 after the @tech{redex} is reduced to a @tech{value}.
 
-Before some things can be evaluated, some sub-expressions must be
-evaluated; for example, in the application @racket[(- 4 (+ 1 1))], the
+Before some expressions can be evaluated, their sub-expressions must be
+evaluated. For example, in the application @racket[(- 4 (+ 1 1))], the
 application of @racket[-] cannot be reduced until the sub-expression
 @racket[(+ 1 1)] is reduced.
-
 Thus, the specification of each syntactic form specifies how (some of)
-its sub-expressions are evaluated, and then how the results are
+its sub-expressions are evaluated and then how the results are
 combined to reduce the form away.
 
 The @deftech{dynamic extent} of an expression is the sequence of
-evaluation steps during which the expression contains the @tech{redex}.
+evaluation steps starting when the expression is selected as a @tech{redex}
+and continuing until it is fully evaluated.
 
 @;------------------------------------------------------------------------
 @section{Tail Position}
@@ -86,12 +94,12 @@ evaluation steps during which the expression contains the @tech{redex}.
 An expression @racket[_expr1] is in @deftech{tail position} with
 respect to an enclosing expression @racket[_expr2] if, whenever
 @racket[_expr1] becomes a redex, its @tech{continuation} is the same
-as was the enclosing @racket[_expr2]'s @tech{continuation}.
+as the enclosing @racket[_expr2]'s @tech{continuation}.
 
 For example, the @racket[(+ 1 1)] expression is @italic{not} in @tech{tail
 position} with respect to @racket[(- 4 (+ 1 1))]. To illustrate, we use
 the notation @sub[_C _expr] to mean the expression that is produced by
-substituting @racket[_expr] in place of @hole in the @tech{continuation}
+substituting @racket[_expr] in place of @hole in some arbitrary @tech{continuation}
 @racket[_C]:
 
 @racketblock[
@@ -99,16 +107,17 @@ substituting @racket[_expr] in place of @hole in the @tech{continuation}
 ]
 
 In this case, the @tech{continuation} for reducing @racket[(+ 1 1)] is
-@sub[_C (- 4 @#,hole)], not just @racket[_C].
+@sub[_C (- 4 @#,hole)], not just @racket[_C]. The requirement specified in the first paragraph above is not met.
 
 In contrast, @racket[(+ 1 1)] is in @tech{tail position} with respect
-to @racket[(if (zero? 0) (+ 1 1) 3)], because, for any continuation
+to @racket[(if (zero? 0) (+ 1 1) 3)] because, for any continuation
 @racket[_C],
 
 @racketblock[
 @#,sub[_C (if (zero? 0) (+ 1 1) 3)] @#,reduces @#,sub[_C (if #t (+ 1 1) 3)] @#,reduces @#,sub[_C (+ 1 1)]
 ]
 
+The requirement specified in the first paragraph is met.
 The steps in this reduction sequence are driven by the definition of
 @racket[if], and they do not depend on the @tech{continuation}
 @racket[_C]. The ``then'' branch of an @racket[if] form is always in
@@ -118,18 +127,19 @@ branch of an @racket[if] form is also in @tech{tail position}.
 
 @tech{Tail-position} specifications provide a guarantee about the
 asymptotic space consumption of a computation. In general, the
-specification of @tech{tail positions} goes with each syntactic form,
-like @racket[if].
+specification of @tech{tail positions} accompanies the description of
+each syntactic form, such as @racket[if].
 
 @;------------------------------------------------------------------------
 @section[#:tag "values-model"]{Multiple Return Values}
 
-A Racket expression can evaluate to @deftech{multiple values}, in the
-same way that a procedure can accept multiple arguments.
+A Racket expression can evaluate to @deftech{multiple values}, similar
+to the way that a procedure can accept multiple arguments.
 
-Most @tech{continuations} expect a particular number of result
-@tech{values}.  Indeed, most @tech{continuations}, such as @racket[(+
-@#,hole 1)] expect a single @tech{value}. The @tech{continuation}
+Most @tech{continuations} expect a certain number of result
+@tech{values}, although some @tech{continuations} can accept
+an arbitrary number. Indeed, most @tech{continuations}, such as @racket[(+
+@#,hole 1)], expect a single @tech{value}. The @tech{continuation}
 @racket[(let-values ([(x y) @#,hole]) _expr)] expects two result
 @tech{values}; the first result replaces @racket[x] in the body
 @racket[_expr], and the second replaces @racket[y] in
@@ -139,7 +149,7 @@ the result(s).
 
 In general, the specification of a syntactic form indicates the
 number of @tech{values} that it produces and the number that it
-expects from each of its sub-expression. In addition, some procedures
+expects from each of its sub-expressions. In addition, some procedures
 (notably @racket[values]) produce multiple @tech{values}, and some
 procedures (notably @racket[call-with-values]) create continuations
 internally that accept a certain number of @tech{values}.
@@ -155,7 +165,7 @@ then an algebra student simplifies @tt{x + 1} as follows:
 
 @verbatim{  x + 1 = 10 + 1 = 11}
 
-Racket works much the same way, in that a set of @tech{top-level
+Racket works much the same way, in that a set of @deftech{top-level
 variables} are available for substitutions on demand during
 evaluation. For example, given
 
@@ -169,15 +179,16 @@ then
 #,(redex (+ x 1)) #,reduces #,(redex (+ 10 1)) #,reduces 11
 ]
 
-In Racket, the way definitions appear is just as important as the way
-that they are used. Racket evaluation thus keeps track of both
+In Racket, the way definitions are created is just as important as the way
+they are used. Racket evaluation thus keeps track of both
 definitions and the current expression, and it extends the set of
 definitions in response to evaluating forms such as @racket[define].
 
-Each evaluation step, then, takes the current set of definitions and
-program to a new set of definitions and program. Before a
+Each evaluation step, then, transforms the current set of definitions and
+program into a new set of definitions and program. Before a
 @racket[define] can be moved into the set of definitions, its
-right-hand expression must be reduced to a @tech{value}.
+second expression must be reduced to a @tech{value}.
+(The first expression is used without evaluation.)
 
 @prog-steps/no-obj[
 [{}
@@ -216,17 +227,18 @@ variables}, various procedures enable the modification of elements
 within a compound data structure. For example, @racket[vector-set!]
 modifies the content of a vector.
 
-To allow such modifications to data, we must distinguish between
+To explain such modifications to data, we must distinguish between
 @tech{values}, which are the results of expressions, and
 @deftech{objects}, which hold the data referenced by a value.
 
 A few kinds of @tech{objects} can serve directly as values, including
 booleans, @racket[(void)], and small exact integers. More generally,
-however, a @tech{value} is a reference to an @tech{object}. For
-example, a @tech{value} can be a reference to a particular vector that
+however, a @tech{value} is a reference to an @tech{object} stored somewhere
+else. For example, a @tech{value} can refer to a particular vector that
 currently holds the value @racket[10] in its first slot. If an
-@tech{object} is modified, then the modification is visible through
-all copies of the @tech{value} that reference the same @tech{object}.
+@tech{object} is modified via one @tech{value},
+then the modification is visible through
+all copies of the @tech{value} that reference the @tech{object}.
 
 In the evaluation model, a set of @tech{objects} must be carried along
 with each step in evaluation, just like the definition set. Operations
@@ -299,14 +311,15 @@ that create @tech{objects}, such as @racket[vector], add to the set of
 
 The distinction between a @tech{top-level variable} and an object
 reference is crucial. A @tech{top-level variable} is not a
-@tech{value}; each time a @tech{variable} expression is evaluated, the
-value is extracted from the current set of definitions. An object
-reference, in contrast is a value, and therefore needs no further
-evaluation. The model evaluation steps above use angle-bracketed
+@tech{value}, so it must be evaluated. Each time
+a @tech{variable} expression is evaluated, the
+value of the variable is extracted from the current set of definitions. An object
+reference, in contrast, is a value and therefore needs no further
+evaluation. The evaluation steps above use angle-bracketed
 @racket[<o1>] for an object reference to distinguish it from a
 @tech{variable} name.
 
-A direct object reference can never appear in a text-based source
+An object reference can never appear directly in a text-based source
 program. A program representation created with
 @racket[datum->syntax], however, can embed direct references to
 existing @tech{objects}.
@@ -319,7 +332,7 @@ The @racket[eq?] operator compares two @tech{values}, returning
 of equality is suitable for comparing objects that support imperative
 update (e.g., to determine that the effect of modifying an object
 through one reference is visible through another reference). Also, an
-@racket[eq?]  test evaluates quickly, and @racket[eq?]-based hashing
+@racket[eq?] test evaluates quickly, and @racket[eq?]-based hashing
 is more lightweight than @racket[equal?]-based hashing in hash tables.
 
 In some cases, however, @racket[eq?] is unsuitable as a comparison
@@ -351,14 +364,15 @@ In the program state
 
 evaluation cannot depend on @racket[<o2>], because it is not part of
 the program to evaluate, and it is not referenced by any definition
-that is accessible in the program. The @tech{object} @racket[<o2>] may
-therefore be removed from the evaluation by @deftech{garbage
+that is accessible by the program. The object is said to be
+not @deftech{reachable}. The @tech{object} @racket[<o2>] may
+therefore be removed from the program state by @deftech{garbage
 collection}.
 
 A few special compound datatypes hold @deftech{weak references} to
 objects. Such weak references are treated specially by the garbage
 collector in determining which @tech{objects} are reachable for the
-remainder of the computation. If an @tech{object} is reachable only
+remainder of the computation. If an @tech{object} is reachable @italic{only}
 via a @tech{weak reference}, then the object can be reclaimed, and the
 @tech{weak reference} is replaced by a different @tech{value}
 (typically @racket[#f]).
@@ -380,12 +394,12 @@ Given
 
 @verbatim{  f(x) = x + 10}
 
-then an algebra student simplifies @tt{f(7)} as follows:
+an algebra student simplifies @tt{f(7)} as follows:
 
 @verbatim{  f(7) = 7 + 10 = 17}
 
 The key step in this simplification is take the body of the defined
-function @tt{f}, and then replace each @tt{x} with the actual
+function @tt{f} and replace each @tt{x} with the actual
 @tech{value} @tt{7}.
 
 Racket procedure application works much the same way. A procedure is
@@ -402,11 +416,11 @@ an @tech{object}, so evaluating @racket[(f 7)] starts with a
 ]
 
 Unlike in algebra, however, the @tech{value} associated with an
-argument can be changed in the body of a procedure by using
+procedure parameter can be changed in the body of a procedure by using
 @racket[set!], as in the example @racket[(lambda (x) (begin (set! x 3)
-x))]. Since the @tech{value} associated with @racket[x] can be
-changed, an actual value cannot be substituted for @racket[x] when
-the procedure is applied.
+x))]. Since the @tech{value} associated with parameter @racket[x] can be
+changed, an argument value cannot be substituted for @racket[x] when
+the procedure is first applied.
 
 Instead, a new @deftech{location} is created for each @tech{variable}
 on each application. The argument @tech{value} is placed in the
@@ -437,7 +451,7 @@ not been used before and that cannot be generated again or
 accessed directly.
 
 Generating a @tech{location} in this way means that @racket[set!]
-evaluates for @tech{local variables} in the same way as for
+evaluates for @tech{local variables}, including parameters, in the same way as for
 @tech{top-level variables}, because the @tech{local variable} is
 always replaced with a @tech{location} by the time the @racket[set!]
 form is evaluated:
@@ -467,7 +481,7 @@ form is evaluated:
  3]
 ]
 
-The substitution and @tech{location}-generation step of procedure
+The @tech{location}-generation and substitution step of procedure
 application requires that the argument is a @tech{value}. Therefore,
 in @racket[((lambda (x) (+ x 10)) (+ 1 2))], the @racket[(+ 1 2)]
 sub-expression must be simplified to the @tech{value} @racket[3], and
@@ -485,13 +499,13 @@ that replaces every instance of @racket[x] in @racket[_expr].
 
 A @deftech{variable} is a placeholder for a @tech{value}, and
 expressions in an initial program refer to @tech{variables}. A
-@deftech{top-level variable} is both a @tech{variable} and a
+@tech{top-level variable} is both a @tech{variable} and a
 @tech{location}. Any other @tech{variable} is always replaced by a
-@tech{location} at run-time, so that evaluation of expressions
+@tech{location} at run-time; thus, evaluation of expressions
 involves only @tech{locations}. A single @deftech{local variable}
 (i.e., a non-top-level, non-module-level @tech{variable}), such as a
-procedure argument, can correspond to different @tech{locations}
-through different instantiations.
+procedure parameter, can correspond to different @tech{locations}
+through different applications.
 
 For example, in the program
 
@@ -506,10 +520,10 @@ hold the value @racket[11].
 
 The replacement of a @tech{variable} with a @tech{location} during
 evaluation implements Racket's @deftech{lexical scoping}. For example,
-when a procedure-argument @tech{variable} @racket[x] is replaced by
-the @tech{location} @racket[xloc], then it is replaced throughout the
+when a procedure parameter @tech{variable} @racket[x] is replaced by
+the @tech{location} @racket[xloc], it is replaced @italic{throughout} the
 body of the procedure, including any nested @racket[lambda]
-forms. As a result, future references of the @tech{variable} always
+forms. As a result, future references to the @tech{variable} always
 access the same @tech{location}.
 
 @;------------------------------------------------------------------------
--- a/pkgs/racket-doc/scribblings/reference/eval-model.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/eval-model.scrbl
@@ -499,7 +499,7 @@ expressions in an initial program refer to @tech{variables}. A
 involves only @tech{locations}. A single @deftech{local variable}
 (i.e., a non-top-level, non-module-level @tech{variable}), such as an
 argument parameter variable, can correspond to different @tech{locations}
-through different applications.
+during different applications.
 
 For example, in the program
 
@@ -527,15 +527,15 @@ access the same @tech{location}.
 
 Most definitions in Racket are in @deftech{modules}. In terms of evaluation,
 a module is essentially a prefix on a defined name, so that different
-modules can define the name. That is, a @deftech{module-level
+modules can define the same name. That is, a @deftech{module-level
 variable} is like a @tech{top-level variable} from the perspective of
 evaluation.
 
 One difference between a module and a top-level definition
 is that a module can be @deftech[#:key "declare"]{declared}
 without instantiating its module-level definitions.
-Evaluation of a @racket[require] @deftech{instantiates}
-(i.e., triggers the @deftech{instantiation} of) a declared
+Subsequently, evaluation of a @racket[require] @deftech{instantiates}
+(i.e., triggers the @deftech{instantiation} of) the declared
 module, which creates variables that correspond to its
 module-level definitions.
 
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -380,12 +386,12 @@ Given
 
 @verbatim{  f(x) = x + 10}
 
-then an algebra student simplifies @tt{f(7)} as follows:
+an algebra student simplifies @tt{f(7)} as follows:
 
 @verbatim{  f(7) = 7 + 10 = 17}
 
 The key step in this simplification is take the body of the defined
```
bennn: Change 'is take" to "is to take" or "is: take"
?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -402,11 +408,12 @@ an @tech{object}, so evaluating @racket[(f 7)] starts with a
...
-argument can be changed in the body of a procedure by using
+argument parameter variable (not to be confused with parameters;
+see @secref["parameter-model"]) can be changed in the body of a procedure by using
 @racket[set!], as in the example @racket[(lambda (x) (begin (set! x 3)
-x))]. Since the @tech{value} associated with @racket[x] can be
-changed, an actual value cannot be substituted for @racket[x] when
-the procedure is applied.
+x))]. Since the @tech{value} associated with parameter variable @racket[x] can be
+changed, an argument value cannot be substituted for @racket[x] when
+the procedure is first applied.
```
bennn: The original paragraph is clearer to me.

At least, please don't introduce "parameter variable" here ... use "argument" or "argument variable" or "formal parameter" or an existing `@tech` (if there is one).

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -481,17 +489,17 @@ Evaluation of a local-variable form, such as @racket[(let ([x (+ 1
...
+@tech{top-level variable} is both a @tech{variable} and a
 @tech{location}. Any other @tech{variable} is always replaced by a
-@tech{location} at run-time, so that evaluation of expressions
+@tech{location} at run-time; thus, evaluation of expressions
 involves only @tech{locations}. A single @deftech{local variable}
-(i.e., a non-top-level, non-module-level @tech{variable}), such as a
-procedure argument, can correspond to different @tech{locations}
-through different instantiations.
+(i.e., a non-top-level, non-module-level @tech{variable}), such as an
+argument parameter variable, can correspond to different @tech{locations}
```
bennn: Same as above, I think "procedure argument" is better.

Maybe its worth adding a `deftech` for this. For now, my choice would be "procedure-argument variable".

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -519,15 +527,15 @@ access the same @tech{location}.
...
+modules can define the same name. That is, a @deftech{module-level
 variable} is like a @tech{top-level variable} from the perspective of
 evaluation.
 
 One difference between a module and a top-level definition
 is that a module can be @deftech[#:key "declare"]{declared}
 without instantiating its module-level definitions.
-Evaluation of a @racket[require] @deftech{instantiates}
-(i.e., triggers the @deftech{instantiation} of) a declared
+Subsequently, evaluation of a @racket[require] @deftech{instantiates}
```
bennn: -1 for adding "Subsequently, "

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -540,46 +548,56 @@ For example, given the module declaration
...
+if the variable were actually named ``x in m.''
 
-@;- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
+@;------------------------------------------------------------------------
 @subsection[#:tag "module-phase"]{Phases}
 
-A module can be @tech{instantiate}d in multiple @deftech{phases}. A
-phase is an integer that, again, is effectively a prefix on the names
-of module-level definitions. A top-level @racket[require]
+The purpose of @deftech{phases} is to solve the ``eval-when'' problem that
```
bennn: What if readers don't know the "eval-when" problem? How about shortening to:

> The purpose of phases is to separate names defined at execution time versus names defined at expansion time. A module can be .... definitions. Phase 0 is the execution-time phase.

and cutting out the reference to Flatt02 b/c it appears later on

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -540,46 +548,56 @@ For example, given the module declaration
...
+
+A module can be @tech{instantiate}d in multiple @tech{phases}. A
+phase is an integer that, like a module name, is effectively a prefix on the names
+of module-level definitions. Phase 0 is the execution-time phase.
+
+A top-level @racket[require]
 @tech{instantiates} a module at @tech{phase} 0, if the module is not
 already @tech{instantiate}d at phase 0.  A top-level
-@racket[(require (for-syntax ....))] @tech{instantiates} a module at
+@racket[(require (for-syntax ...))] @tech{instantiates} a module at
```
bennn: I think the 4 dots are intentional to avoid confusion with the `...` transformer.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -540,46 +548,56 @@ For example, given the module declaration
...
+expressions are in relative @tech{phase} +2, and so on. Higher phases are 
+mainly related to program parsing instead of normal evaluation.
 
 If a module @tech{instantiate}d at @tech{phase} @math{n}
 @racket[require]s another module, then the @racket[require]d module is
 first @tech{instantiate}d at phase @math{n}, and so on
 transitively. (Module @racket[require]s cannot form cycles.) If a
 module @tech{instantiate}d at phase @math{n} @racket[require]s
-@racket[for-syntax] another module, the other module becomes
+another module @racket[for-syntax], the other module becomes
```
bennn: suggestion:

> a module `_M` via `(for-syntax _M)`, then `_M` becomes

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -540,46 +548,56 @@ For example, given the module declaration
...
 @deftech{available} at @tech{phase} @math{n+1}, and it may later be
 @tech{instantiate}d at @tech{phase} @math{n+1}.  If a module that is
-@tech{available} at phase @math{n} for @math{n>0} @racket[require]s
-@racket[for-template] another module, the other module becomes
+@tech{available} at phase @math{n} (for @math{n>0}) @racket[require]s
+another module @racket[for-template], the other module becomes
```
bennn: same suggestion as above, I think these require-specs mixed with English would be clearer as grammatically correct require-specs

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -597,62 +615,62 @@ within modules. For example, @racket[define] within @racket[begin-for-syntax] cr
...
 
 @;- - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - - -
 @subsection[#:tag "separate-compilation"]{The Separate Compilation Guarantee}
 
 When a module is compiled, its @tech{phase} 1 is instantiated. This
 can, in turn, trigger the transitive instantiation of many other
 modules at other phases, including phase 1. Racket provides a very
-strong guarantee about this instantiation called "The Separate
-Compilation Guarantee":
+strong guarantee about this instantiation called ``The Separate
```
bennn: maybe add `@index["separate compilation guarantee"]{The Separate ....}`

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -597,62 +615,62 @@ within modules. For example, @racket[define] within @racket[begin-for-syntax] cr
...
+Compilation Guarantee'':
 
-"Any @tech{effects} of the instantiation of the module's phase 1 due
-to compilation on the Racket runtime system are @tech{discarded}."
+``Any @tech{effects} of the instantiation of the module's phase 1 due
```
bennn: instead of the quotes, maybe `@nested[#:style 'inset]{....}`

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -804,7 +825,7 @@ literals (via @racket[quote-syntax]) or variable references (via
 the syntactic specification of a @tech{cross-phase persistent} module
 declaration.
 
-A documented module should be assumed non-@tech{cross-phase persistent} unless it
+A documented module should be assumed non--@tech{cross-phase persistent} unless it
```
bennn: pretty sure `-` gives the correct dash here

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -850,17 +871,17 @@ can @racket[require] the submodule. In contrast, a submodule declared
 with @racket[module*] conceptually follows its enclosing module, so
 can @racket[require] from its enclosing module, but the enclosing
 module cannot @racket[require] the submodule. Unless a submodule
-imports from its enclosing module or vice-versa, then @tech{visits} or
+imports from its enclosing module or vice versa, then @tech{visits} or
 @tech{instantiations} of the two modules are independent, and their
-implementations may even be loaded from bytecode at different times.
+implementations may even be loaded from bytecode files at different times.
```
bennn: hm, is bytecode always in a file?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -850,17 +871,17 @@ can @racket[require] the submodule. In contrast, a submodule declared
...
 
 A submodule declared with @racket[module] can import any preceding
 submodule declared with @racket[module]. A submodule declared with
 @racket[module*] can import any preceding module declared with
 @racket[module*] and any submodule declared with @racket[module].
 
 When a submodule declaration has the form @racket[(module* _name #f
-....)], then all of the bindings of the enclosing module's bodies are
+...)], then all of the bindings of the enclosing module's bodies are
```
bennn: again I think `....` is intentional

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -874,16 +895,16 @@ Every continuation @racket[_C] can be partitioned into
...
 @frame["n"]]]], and no frame @frame["i"] can be itself partitioned
-into smaller continuations. Evaluation steps add and remove frames to
+into smaller continuations. Evaluation steps add frames to and remove frames from
 the current continuation, typically one at a time.
 
 Each frame is conceptually annotated with a set of
-@deftech{continuation marks}. A mark consists of a key and its value;
-the key is an arbitrary value, and each frame includes at most one
-mark for any key. Various operations set and extract marks from
+@deftech{continuation mark}s. A mark consists of a key and its value.
```
bennn: Its fine to keep the `s` inside the `deftech`

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -946,8 +968,8 @@ In addition to the state that is shared among all threads, each thread
 has its own private state that is accessed through @deftech{thread
 cells}. A thread cell is similar to a normal mutable object, but a
 change to the value inside a thread cell is seen only when extracting
-a value from the cell from the same thread. A thread cell can be
-@deftech{preserved}; when a new thread is created, the creating
+a value from that cell in the same thread. A thread cell can be
+@deftech{preserved}: when a new thread is created, the creating
```
bennn: +1 for keeping the `;`

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -1041,24 +1063,24 @@ created with @racket[thread/suspend-to-kill] can have zero
 custodians. Extra custodians become associated with a thread through
 @racket[thread-resume] (see @secref["threadkill"]). When a thread
 has multiple custodians, it is not necessarily killed by a
-@racket[custodian-shutdown-all], but shut-down custodians are removed
-from the thread's managing set, and the thread is killed when its
+@racket[custodian-shutdown-all]. Instead, shut-down custodians are removed
```
bennn: +1 for keeping ", but" or changing to ", but rather"

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -1041,24 +1063,24 @@ created with @racket[thread/suspend-to-kill] can have zero
...
+garbage collected, at which point its subordinates become immediately
+subordinate to the collected custodian's superordinate (owner) custodian.
 
 In addition to the other entities managed by a custodian, a
 @deftech{custodian box} created with @racket[make-custodian-box]
 strongly holds onto a value placed in the box until the box's
-custodian is shut down. The custodian only weakly retains the box
-itself, however (so the box and its content can be collected if there
-are no other references to them).
+custodian is shut down. However, the custodian only weakly retains the box
```
bennn: +1 for keeping "however" mid-sentence

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -216,17 +219,18 @@ variables}, various procedures enable the modification of elements
...
-however, a @tech{value} is a reference to an @tech{object}. For
-example, a @tech{value} can be a reference to a particular vector that
+however, a @tech{value} is a reference to an @tech{object} stored somewhere
+else. For example, a @tech{value} can refer to a particular vector that
 currently holds the value @racket[10] in its first slot. If an
-@tech{object} is modified, then the modification is visible through
-all copies of the @tech{value} that reference the same @tech{object}.
+@tech{object} is modified via one @tech{value},
+then the modification is visible through
+all copies of the @tech{value} that reference the @tech{object}.
```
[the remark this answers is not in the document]
Paul-C-Anagnostopoulos: That wording, or something similar, was there before I did my first pull of this documentation. I believe the idea is that we should think of all the same values as copies of the original value. I agree it sounds a bit funky. Let's change it.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Yes, thanks!

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I discussed this with Matthew. Further up in the paragraph I introduce the term "argument parameter value" and also say not to confuse this with the (unfortunate) term "parameter." I will make sure that every instance of the term uses all three words.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: The trouble with "procedure argument variable" is that it clearly connotes a procedure call argument that is a variable, which is not what we're talking about here. I think that "procedure parameter variable" is clearly not referring to a parameter.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: You don't like that word. Is "Later," better?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Sounds good, though I'd keep the citation. I couldn't understand phases without reading his paper twice.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Aha, right. I'll fix that.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Sounds good. Why the underscore?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Okay.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Will do.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Aha, that's how to do a quote. Okay!

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I wanted an en-dash, which two hyphens produces.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I'll change it to "sources".

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I will fix this.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Okay.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Do you mean I should revert to the semicolon? I'm not sure why.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Sorry, I don't understand what you are indicating here.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Again, what?

Paul-C-Anagnostopoulos:
Ben, once I have made these changes, what is the correct way to push a new version of the file?

Also, I believe that Matthew made a change to this file, which is why the branch has conflicts. How does that get resolved?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -60,55 +59,57 @@ Some simplifications require more than one step. For example:
...
-
 Thus, the specification of each syntactic form specifies how (some of)
-its sub-expressions are evaluated, and then how the results are
+its sub-expressions are evaluated and then how the results are
 combined to reduce the form away.
 
 The @deftech{dynamic extent} of an expression is the sequence of
-evaluation steps during which the expression contains the @tech{redex}.
+evaluation steps starting when the expression is selected as a @tech{redex}
+and continuing until it is fully evaluated.
```
lexi-lambda: I think this change is incorrect. The original wording implies that the dynamic extent of an expression is the sequence of reduction steps when the expression *contains* the redex, that is, contains the hole. This means that in the evaluation of the expression `(+ 1 (+ 2 3))`, the dynamic extent of the expression includes the step of evaluation that decomposes the expression into the context `(+ 1 [])` and the redex `(+ 2 3)`. However, the new wording would exclude that step, since the expression itself is not the redex.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -60,55 +59,57 @@ Some simplifications require more than one step. For example:
...
-place of the redex. That is, the continuation says how to ``continue''
+context surrounding the redex. In @racket[(- 4 (+ 1 1))], the redex is @racket[(+ 1 1)], and
+the continuation is @racket[(- 4 @#,hole)], where @hole indicates the
+position of the redex. That is, the continuation says how to ``continue''
 after the @tech{redex} is reduced to a @tech{value}.
 
-Before some things can be evaluated, some sub-expressions must be
-evaluated; for example, in the application @racket[(- 4 (+ 1 1))], the
+Before some expressions can be evaluated, their sub-expressions must be
+evaluated. For example, in the application @racket[(- 4 (+ 1 1))], the
```
lexi-lambda: I’d be careful here—the original wording is admittedly vague, but perhaps intentionally so. When evaluating an expression like `(lambda (x) (+ 1 x))`, the sub-expression `(+ 1 x)` is *not* evaluated, and indeed, it cannot be. Therefore, not all sub-expressions are necessarily evaluated when an expression is evaluated. This is also true for forms like `if`, which evidently must not have a standard evaluation order. Perhaps there is some way to reword this paragraph to avoid the vague phrase “some sub-expressions”, but I think this is too aggressive.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -60,55 +59,57 @@ Some simplifications require more than one step. For example:
...
+and continuing until it is fully evaluated.
 
 @;------------------------------------------------------------------------
 @section{Tail Position}
 
 An expression @racket[_expr1] is in @deftech{tail position} with
 respect to an enclosing expression @racket[_expr2] if, whenever
 @racket[_expr1] becomes a redex, its @tech{continuation} is the same
-as was the enclosing @racket[_expr2]'s @tech{continuation}.
+as the enclosing @racket[_expr2]'s @tech{continuation}.
```
lexi-lambda: I think the “was” here is valuable: for a form to be in tail position, *`expr2`* must have disappeared by the time *`expr1`* becomes a redex, since otherwise *`expr2`* would be a part of *`expr1`*’s continuation. Using “was” emphasizes that *`expr2`* has been reduced away.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -60,55 +59,57 @@ Some simplifications require more than one step. For example:
...
 
 For example, the @racket[(+ 1 1)] expression is @italic{not} in @tech{tail
 position} with respect to @racket[(- 4 (+ 1 1))]. To illustrate, we use
 the notation @sub[_C _expr] to mean the expression that is produced by
-substituting @racket[_expr] in place of @hole in the @tech{continuation}
+substituting @racket[_expr] in place of @hole in some arbitrary @tech{continuation}
```
lexi-lambda: Could you explain what you’re trying to clarify with this change? I don’t understand what is unclear about using “the”, but I could be missing something.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -118,18 +119,19 @@ branch of an @racket[if] form is also in @tech{tail position}.
...
+specification of @tech{tail positions} accompanies the description of
+each syntactic form, such as @racket[if].
 
 @;------------------------------------------------------------------------
 @section[#:tag "values-model"]{Multiple Return Values}
 
-A Racket expression can evaluate to @deftech{multiple values}, in the
-same way that a procedure can accept multiple arguments.
+A Racket expression can evaluate to @deftech{multiple values}, similar
+to the way that a procedure can accept multiple arguments.
```
lexi-lambda: I think this is a good change, but I think it could be made even clearer. Maybe something like “to provide symmetry with the fact that a procedure can accept multiple arguments”?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -155,8 +157,8 @@ then an algebra student simplifies @tt{x + 1} as follows:
 
 @verbatim{  x + 1 = 10 + 1 = 11}
 
-Racket works much the same way, in that a set of @tech{top-level
-variables} are available for substitutions on demand during
+Racket works much the same way, in that a set of @deftech{top-level
+variables} (see also @secref["vars-and-locs"]) are available for substitutions on demand during
```
lexi-lambda: This doesn’t seem like the right place to define the term “top-level variable” to me, and indeed, as-written, it doesn’t define top-level variables at all. I think forward references in the evaluation model section are fine, since the evaluation model is supposed to serve as a dense, comprehensive reference by design, not a guide. Forward references seem like a bad idea in, say, The Racket Guide, but I’d keep the definition in the appropriate section here and maintain the forward reference.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -169,15 +171,16 @@ then
...
 definitions in response to evaluating forms such as @racket[define].
 
-Each evaluation step, then, takes the current set of definitions and
-program to a new set of definitions and program. Before a
+Each evaluation step, then, transforms the current set of definitions and
+program into a new set of definitions and program. Before a
 @racket[define] can be moved into the set of definitions, its
-right-hand expression must be reduced to a @tech{value}.
+second expression must be reduced to a @tech{value}.
+(The first expression is used without evaluation.)
```
lexi-lambda: This is incorrect, since definitions only have *one* expression, the right-hand side. In `(define x e)`, `x` is not an expression at all but merely an identifier, so the only expression is `e`. I personally think the original wording is alright (since this section is not intended to document the `define` form, by any means), but if you wanted to make it clearer, you could reword it to “its expression—that is, its right-hand side—must be reduced”.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
bennn: oh, and that's because "cross-phase" has a dash, ok

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -351,14 +356,15 @@ In the program state
 
 evaluation cannot depend on @racket[<o2>], because it is not part of
 the program to evaluate, and it is not referenced by any definition
-that is accessible in the program. The @tech{object} @racket[<o2>] may
-therefore be removed from the evaluation by @deftech{garbage
+that is accessible by the program. The object is said to be
+not @deftech{reachable}. The @tech{object} @racket[<o2>] may
+therefore be removed from the program state by @deftech{garbage
```
lexi-lambda: I like this `deftech` here, though I think it would be *slightly* less clumsy to write “said to not be reachable” rather than “said to be not reachable”. I wonder if a simple text search through the reference could find places to add `@tech{reachable}` where appropriate?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
bennn: yes that's what I mean

I like using colons only to indicate that a list is coming, and semicolons for joining similar sentences.

no big deal if you disagree

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
lexi-lambda: I agree that this nomenclature is confusing to me, and the old nomenclature seemed fine. Consider this a vote against this change.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
bennn: I prefer the old sentence: 
> The custodian only weakly retains the box
itself, however ....

bennn:
> Ben, once I have made these changes, what is the correct way to push a new version of the file?

Make a new commit & push it to this `documentation` branch.

```
git add <changes>
git commit
git push <origin?> documentation
```


> Also, I believe that Matthew made a change to this file, which is why the branch has conflicts. How does that get resolved?

The easiest way is with a rebase 😈 .

Checkout your master branch, pull the latest commits (which should include Matthew's changes), go back to this `documentation` branch, and rebase against master:

```
git checkout master
git pull <upstream?> master
git checkout documentation
git rebase -i master
```

First thing, `git rebase -i master` will open a text editor asking what commits you want to keep. Close the text editor immediately --- because you'll want all the commits.

Second, `git rebase -i master` will try to re-commit your changes on top of what Matthew did. This is not going to succeed because of the conflicts, but: use `git status`, follow the instructions it prints, and things should go fine. (You'll probably want to open `eval-model.scrbl`, look for `<<<<`, use `git add` on `eval-model.scrbl`, and run `git rebase --continue`)

lexi-lambda:
I think the rebase might be slightly more complicated than you suggest, @bennn, since it looks like this PR includes three commits that were squashed together when Matthew merged #2326. It’s possible that the right thing to do is instead to drop the first three commits and just keep the most recent one, then rebase that on top of master. It would also be possible to `git cherry-pick` the most recent commit onto a fresh branch on top of current master to avoid needing to rebase at all.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Colon can also mean "an explanation follows."

Paul-C-Anagnostopoulos:
Okay, folks, thanks for all the comments. I will incorporate them into the documentation and then determine how best to push them up. Full disclosure: My brain and git do not operate in the same universe.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I was trying to clarify that C is arbitrary with respect to the hole. Perhaps I ended up suggesting that the continuation might not be the one "surrounding" the hole.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I'm not sure why people would be confused by using the term "parameter" in its absolutely standard way. I understand it's overloading the Racket term "parameter." I will switch back to "argument identifier." Would a parenthetical comment explaining why we don't use the word "parameter" be okay?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Interesting. This is a standard copy edit so as not to bury the contrast word.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
lexi-lambda: It’s not to say that people won’t be familiar with the term “parameter” used in this sense, but rather that the overloading you mention makes it confusing which version of “parameter” is being used here, especially since the existing documentation takes great care to *only* use it to mean the dynamic construct rather than function arguments. I do think a margin note (using the `@margin-note` procedure) would be great if you want to write a sentence or two explaining the nomenclature.

Paul-C-Anagnostopoulos: Sounds good.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -60,55 +59,57 @@ Some simplifications require more than one step. For example:
...
+into two parts: a @deftech{redex} (``reducible expression''),
+which is the part that can change in a
 single-step simplification (highlighted), and the
 @deftech{continuation}, which is the evaluation
-context surrounding an expression. In @racket[(- 4 (+ 1 1))], the redex is @racket[(+ 1 1)], and
-the continuation is @racket[(- 4 @#,hole)], where @hole takes the
-place of the redex. That is, the continuation says how to ``continue''
+context surrounding the redex. In @racket[(- 4 (+ 1 1))], the redex is @racket[(+ 1 1)], and
+the continuation is @racket[(- 4 @#,hole)], where @hole indicates the
+position of the redex. That is, the continuation says how to ``continue''
```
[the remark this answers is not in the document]
Paul-C-Anagnostopoulos: I don't see why we want to say that the hole "takes the place of" or "replaces" the redex. It's just an indicator in the continuation for documentation purposes. How about "represents the redex"?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -540,46 +548,56 @@ For example, given the module declaration
...
+Likewise, the @racket[define-for-syntax] form is similar to @racket[define],
+but shifts the definition by +1.
 Thus, if the module is @tech{instantiate}d at phase 1,
 the variables defined with @racket[begin-for-syntax] are created at phase 2,
 and so on. Moreover, this relative phase acts as another layer of
-prefixing, so that a @racket[define] of @racket[x] and a 
-@racket[begin-for-syntax]-wrapped
-@racket[define] of @racket[x] can co-exist in a module
+prefixing, so that @racket[x] defined with @racket[define] and
+@racket[x] defined with @racket[begin-for-syntax] can co-exist in a module
```
lexi-lambda: Should this be “defined with `define-for-syntax`” or “defined with `define` wrapped with `begin-for-syntax`”?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -597,62 +615,62 @@ within modules. For example, @racket[define] within @racket[begin-for-syntax] cr
...
+space observable by other processes. Thus, effects which modify those
+bytes are not discardable, so @racket[bytes-set!], in this case, has an
 external effect.
 
 The opposite is also true: some things which appear to be external are
 actually internal. For instance, if a Racket program starts multiple
 threads and uses mutation to communicate between them, that mutation
 is purely internal, because Racket's threads are defined entirely
-internally.
+internally. They are not related to operating system threads.
```
lexi-lambda: I think this would read better as a parenthetical at the end of the preceding sentence.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Is it the case that _expr2_ must always have disappeared? If a function call is in tail position and that function always returns its first argument, isn't the first argument in tail position?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: It's odd, then, that the section is titled ``Top-Level Variables.'' But I'll make this change.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I didn't use `@tech` around all the instances in the next few paragraphs, since the reader's attention is already on the term. I did add it around an instance in the "Custodians" section.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -540,46 +548,56 @@ For example, given the module declaration
 
 the evaluation of @racket[(require 'm)] creates the variable @racket[x]
 and installs @racket[10] as its value. This @racket[x] is unrelated to
-any top-level definition of @racket[x].
+any top-level definition of @racket[x]. It's as
+if the variable were actually named ``x in m.''
```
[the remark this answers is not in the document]
Paul-C-Anagnostopoulos: Good.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I'll make it `define-for-syntax`.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: Agreed.

Paul-C-Anagnostopoulos:
I have made all the changes discussed in this thread. I'll stand by in case either of you have additional comments. Let me know when you are done.

Thanks for your help with this!

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
samth: No, in fact no argument is ever in tail position with respect to the function call.

lexi-lambda: I’m not completely sure what you’re asking. Let’s consider the following expression:

```racket
(+ 1 ((lambda (x) x) (+ 2 3)))
```

I see two interpretations of your question:

  1. You are asking if `x` is in tail position with respect to the surrounding application. In that case, the answer is **yes**. When the application becomes the redex, it is `((lambda (x) x) 3)`, and its continuation is `(+ [] 1)`. After a step of evaluation, the expression will be reduced to `(+ xloc 1)`, where `xloc` is a location with the value `3`. Therefore, the new redex is `xloc`, and its continuation is still `(+ [] 1)`, which is the same as the previous redex’s continuation, so `x` is in tail position with respect to the surrounding application.

  2. You are asking if `(+ 2 3)` is in tail position with respect to the surrounding application. In that case, the answer is **no**. When `(+ 2 3)` becomes the redex, its continuation is `(+ ((lambda (x) x) []) 1)`, which is larger than the continuation of the eventual application, `(+ [] 1)`.

In either case, what I said still holds: if an expression is in tail position, the surrounding expression will have disappeared by the time it becomes a redex.

Paul-C-Anagnostopoulos: I think my confusion has to do with when expressions are evaluated. For example, I think of the arm of an `if` as being evaluated before it is given to the continuation of the `if`. So the surrounding `if` expression is still there. But I need to think of the expression as given to the continuation and then evaluated, so that the `if` is gone when the expression is the redex.

Paul-C-Anagnostopoulos:
Is everyone done commenting on this documentation update?

lexi-lambda:
Please push your changes to your branch so that people can see them—as it stands, it doesn’t look like you’ve pushed any of the changes you’ve made based on the comments so far. (You can just push them as a new commit, if that’s easiest.)

Paul-C-Anagnostopoulos:
No, I haven't pushed the changes yet. I will do that.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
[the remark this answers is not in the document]
lexi-lambda: The idea here is that we can view Racket’s evaluation model as a series of syntactic transformations on expressions. So rather than imagining computation by visualizing an abstract machine with a program counter that executes instructions, we can literally reason about the behavior of programs via the manipulation of symbols. We can think of evaluating a program as a trace of reduction steps, so the program `(+ 1 (* 2 3))` would have the following trace:

`(+ 1 (* 2 3))` → `(+ 1 6)` → `7`

Note that this really can be a full specification of the system: we can think about the entire evaluation model as a series of purely syntactic transformations. However, in order to reason about evaluation order syntactically, we have to introduce this notion of continuations and redexes, which allow us to break expressions apart in well-defined ways. If we made the continuations and redexes explicit in the above trace, we’d have something like this instead:

`(+ 1 (* 2 3))` →<sub>decompose</sub> `(+ 1 [])`[`(* 2 3)`] →<sub>reduce</sub> `(+ 1 [])`[`6`] →<sub>recompose</sub> `(+ 1 6)` →<sub>decompose</sub> `[]`[`(+ 1 6)`] →<sub>reduce</sub> `[]`[`7`] →<sub>recompose</sub> `7`

What I’m getting at here is that the hole, `[]`, is *not* just an indication for documentation purposes! It’s a syntactic marker used in a specific, formal way to provide a syntactic reduction semantics for Racket. If you want to know more about reduction semantics, check out [PLT Redex](https://redex.racket-lang.org) and its accompanying book, [Semantics Engineering with PLT Redex](https://mitpress.mit.edu/books/semantics-engineering-plt-redex), which includes an explanation of reduction semantics and how it works.

Now, admittedly, in practice, Racket is not fully-specified this way. The evaluation model section of the Racket documentation does not include a formal treatment of Racket’s evaluation order in terms of evaluation contexts, nor does it provide a set of syntactic reduction rules. Maybe such a thing could be valuable, I don’t know, but the evaluation model section is somewhere in between: it’s not a formal specification, but it’s not entirely informal, either. In any case, I believe the takeaway is that this decomposition of a term into a redex and its continuation is a formal idea described informally, and a key idea of that decomposition is that it is syntactic; **the hole really does syntactically take the place of the redex** in the syntax we use to describe a continuation.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
lexi-lambda: +1, I like this new rewording.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -179,8 +179,8 @@ definitions in response to evaluating forms such as @racket[define].
 Each evaluation step, then, transforms the current set of definitions and
 program into a new set of definitions and program. Before a
 @racket[define] can be moved into the set of definitions, its
-second expression must be reduced to a @tech{value}.
-(The first expression is used without evaluation.)
+expression (i.e., its right-hand side) must be reduced to a @tech{value}.
+(The left-hand side is used without evaluation.)
```
lexi-lambda: This is a definite improvement, but I still have a minor quibble with the final parenthetical. It is a category error to talk about “evaluation” of the left-hand side, even in the negative, since the left hand side is not an expression. It would be like saying the `define` in `(define x 3)` is used without evaluation—it’s simply a part of the syntax of the program.

That said, it seems like you felt this needed clarifying, so I imagine there’s a reason you added the extra sentence. Could you describe what you were trying to disambiguate? There might be a better way of doing that.

Paul-C-Anagnostopoulos: I'm trying to make it clear that the name or procedure being defined is not evaluated in the context of `define`. We could, after all, require `(define 'foo ...)`. 

A minor point is that I find "left-hand side" and "right-hand side" to be odd, since the `define` form does not look like an operator.

Paul-C-Anagnostopoulos: @lexi-lambda Why can I reply to the first of your comments right above?

Anyway, what do you suggest instead of "indicates the position of". I think "takes the place of" suggests that the redex will never be recomposed into the expression.

lexi-lambda: > We could, after all, require `(define 'foo ...)`

No, we couldn’t do that, since Racket is a statically-bound, lexically-scoped language. Predictable expansion of macros necessitates knowing the binding structure of a program statically, but if we allowed something like `(define (string->symbol (read-line)) ....)`, then we wouldn’t be able to know the binding structure until the program is run.

The “left-hand side” and “right-hand side” terminology is, admittedly, a little weird in the context of Racket. If you want, you could use other phrases instead, but at that point, I think you end up with the original documentation, which just describes “its expression” (unambiguous since there is only one, as described in the documentation for `define`).

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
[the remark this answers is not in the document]
lexi-lambda: > Anyway, what do you suggest instead of "indicates the position of". I think "takes the place of" suggests that the redex will never be recomposed into the expression.

That’s a fair point. Could you just clarify that by saying “@hole replaces the redex during its simplification”? I don’t like “indicates the position of” because it seems needlessly vague compared to “replaces”. It makes me wonder *how* it indicates the redex’s position, and the answer to that is both simple and precise: it replaces it, syntactically.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -540,46 +551,54 @@ For example, given the module declaration
 
 the evaluation of @racket[(require 'm)] creates the variable @racket[x]
 and installs @racket[10] as its value. This @racket[x] is unrelated to
-any top-level definition of @racket[x].
+any top-level definition of @racket[x] (as if it were given a unique,
+model-specific prefix).
```
lexi-lambda: Minor typo: this should be “module”, not “model”.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: How about "where [] holds the position of the redex"?

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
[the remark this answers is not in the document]
Paul-C-Anagnostopoulos: Can we make it clearer by adding a few words? I'm only trying to make it clear that it does not take the place of the redex "permanently."

"... where [] takes the place of the redex as the redex is reduced."

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: [I'm confused about why this conversation is duplicated above.]

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
Paul-C-Anagnostopoulos: I have reverted to the original text, but I still do not understand.

In your second sentence above, "contains the redex" seems ambiguous, since an expression can contain multiple redexes.

lexi-lambda: No, it can’t. While I’m not sure it is explicitly explained in the current reference, there are rules that ensure each expression has a *unique decomposition* into a continuation and a redex. For example, in the expression `(+ (* 2 3) (* 4 5))`, *the* redex is `(* 2 3)`, since Racket defines a left-to-right evaluation order. The `(* 4 5)` expression can only become the redex once `(* 2 3)` has been reduced to a value, and the overall expression becomes `(+ 6 (* 4 5))`.

lexi-lambda: > So we don't say that the dynamic extent of the outer expression includes both reductions?

We do! But note that “dynamic extent” is just defined as “the sequence of **evaluation steps**” (emphasis mine) during with the expression contains the redex. Since the decomposition (into continuation+redex), reduction (of the redex), and recomposition (of the continuation+reduced redex) process comprises a *single* evaluation step, it doesn’t make sense to think of dynamic extent of applying to some of those “sub-steps” but not others.

Given the above, and given that each evaluation step reduces exactly one redex one step, by definition, the phrase “the sequence of evaluation steps during which the expression contains the redex” is precise. There is only one redex per evaluation step, and an expression contains the redex the unique decomposition selects a redex inside the expression.

Paul-C-Anagnostopoulos: I'm going to keep going even though I'm starting to wonder whether it's just me.

I'm confused about whether a dynamic extent includes only one redex. Above you said that it cannot, but then in your post just now you said that the dynamic extent of the outer expression includes both reductions [the expression `(+ (* 2 3) (* 4 5))`]. So does the dynamic extent of that expression include the reduction of only `(* 2 3)` or does it include the reduction of both arguments?

If it is only `(* 2 3)`, then wouldn't the following be clearer: "... contains the redex or its hole." After all, we are emphasizing that the hole really does take the place of the redex during reduction.

On the other hand, if it is both the subexpressions, then it doesn't seem right to refer to "_the_ redex."

lexi-lambda: > I'm confused about whether a dynamic extent includes only one redex. Above you said that it cannot, but then in your post just now you said that the dynamic extent of the outer expression includes both reductions

I never said that dynamic extent can’t have several redexes, merely that each expression has a unique decomposition into a continuation and a redex. There is one redex per evaluation step, by definition.

I think I can see now where your confusion is coming from, though: the reference plays a little loose with what “an expression” is. It’s true that `(+ (* 2 3) (* 4 5))` has exactly one redex—`(* 2 3)`—but after a single evaluation step, it becomes `(+ 6 (* 4 5))`, which has a new redex—`(* 4 5)`. The surrounding `(+ ....)` has stuck around, but the expression is fundamentally a new expression… it must be for some evaluation to have taken place!

In that sense, in order to fully specify dynamic extent, we have to think not just about the redex but also about the continuation. In both evaluation steps, the application expression `(+ ....)` is in the continuation, unchanging, and it contains the hole (and therefore the redex). We say that this is a single expression with a notion of continuity over several evaluation steps, even though its subexpressions are being reduced.

Perhaps someone else can supply a better way of describing this. The current wording matches my intuition well, so I am perhaps not the best person to suggest an improvement.

Paul-C-Anagnostopoulos: I don't see how `(+ ....)` can contain only one hole. The first redex is `(* 2 3)`, which is replaced by a hole. But then the expression is reduced to `(+ 6 (* 4 5))`, which has a second redex.

If that second expression is considered to be the original expression, then the original expression has two redexes.

If that second expression is considered to be a new expression, then the dynamic extent of the original expression doesn't cover the evaluation of `(* 4 5)`, because it is finished "containing a redex."

Thanks for your patience with this. We are making progress!

Issue #2154, opened 2018-06-28
Reconsider `#:extend-stop-ids?` argument to `local-expand`


michaelballantyne:
I started a discussion with @lexi.lambda on slack about the `#:extend-stop-ids?` addition to `local-expand`. She suggested I include others who had expressed an interest, so here it is.

I'm not convinced that uses of the feature can be reliable, so I'm proposing it be removed before the Racket 7 release.

My concern is that any macro that needs to use the feature will be fragile. The previous rule of adding all the core forms to the stop list for any partial expansion had the effect of ensuring that when a macro is expanded in expression position, it will always be a full rather than partial expansion.

`let-syntax`, `syntax-parameterize`, and anything using first-class definition contexts either rely on this guarantee, or break partial expansion of expressions by their own use of local-expand. (syntax-parameterize could be implemented to work either way). If a let-syntax expression is partially expanded such that there remain unexpanded uses of bound macro, later re-expansion will fail with an identifier out of context error.

In Alexis's use-case in Hackett, she has to be careful to not compile Hackett forms to Racket expressions that use these features, and had trouble with `match` using `syntax-parameterize`. Thus `#:extend-stop-ids?` works given you can assume a closed world of macros you might expand, and as long as the implementation of those macros isn't changed in the future to use local macro binding or syntax parameters in a way that binders and references could be separated by one of the forms expansion stops at. Other cases will break.

If a language like Hackett doesn't follow the same protocol as Racket for its macros, it would be better to define a custom expander (as Alexis did for Hackett's type language here: https://lexi-lambda.github.io/), rather than re-use Racket's forms and macros. The core forms of that language and macros that expand to them know to expect the different protocol. In the case of Hackett's partial expansion, these forms would be careful to not rely on local syntax bindings that their re-expansion doesn't re-create. For example, a typed "let" core form could use a syntax binding to attach type information to any references, but rather than expand away as in Turnstile, the form would expand to itself. Subsequent re-expansions would re-establish the syntax binding for the type information just as re-expansions of `let` re-establish the normal let binding. A toy sample of such an expander is here: https://gist.github.com/michaelballantyne/d040e4cdec8a3484fb49adaa0ce5a227

The implementation of syntax parameters should probably expose something that would allow languages like Hackett to define their own syntax-parameterize form, compatible with their expansion protocol.

I'm curious about other people's interest in `#:extend-stop-ids?`. Do you have use-cases that you think would be robust?

@florence @samth @mflatt

[...]

PR #2355, opened 2018-11-04
Reinstate the stop list after #%module-begin performs partial expansion

lexi-lambda:
This PR pulls out a minor change that I originally bundled with the [the ill-fated change to `local-expand`](https://github.com/racket/racket/issues/2154). Specifically, it changes the expansion of `#%plain-module-begin` so that it reinstates the stop list after it performs partial expansion of module-level forms. This makes it possible to meaningfully use the stop list when expanding in a `'module-begin` context. This is unrelated to the aforementioned change to `local-expand`, and I don’t think it weakens any guarantees in the same way, since it only applies if a programmer explicitly calls `local-expand` in a `'module-begin` context and supplies a stop list.

This change would be useful for me to try to get Hackett working again. While I’ve discussed with @mflatt some ways to make Hackett do what I want in the long-term, which would not depend on this change, those approaches are just too big for me to take on right now, and this would provide a path to making Hackett work in the short-term, without a massive restructuring of how it works.

(cc @michaelballantyne)


mflatt:
I don't remember any problems with this change, so it seems ok to merge.

lexi-lambda:
Thanks!

PR #2356, opened 2018-11-05
Point to body forms in missing-expression error.

samth:
Improves the error message for:

```
(define-syntax (like-lambda stx)
 (syntax-case stx  ()
   [(_ e) #'(lambda () e)]))

(like-lambda (define x 1))
```

Based on a report from @pkoronkevich.


mflatt:
I'm not clear on all the implications here, but it looks worth trying out.

PR #2367, opened 2018-11-12
Add support for collapsible contracts

dfeltey:
Not Ready to Merge!!

This pull request adds support for collapsible arrow and vector contracts in order to avoid certain pathological slowdowns in Typed Racket programs.

All the contract system tests are passing, but I still need to finish documenting the public portions of the implementation and adjust some internal names to use collapsible rather than space-efficient. Additionally there are a few changes necessary to Typed Racket's contract generation in order for it to take advantage of collapsible contracts.


samth:
If this needs matching changes to Typed Racket, can you post a PR for that as well so we can see what that looks like?

rfindler:
What changes to TR did you have in mind, @samth ?

samth:
Whatever changes @dfeltey was referring to in the original PR description.

dfeltey:
@samth There were originally plans for avoiding caching arbitrary contracts that required some changes to TR's contract generation, but I ended up removing that from this pull request. Currently there are no changes necessary to TR to support collapsible contracts.

samth:
Ok, cool.

rfindler:
There are some missing `@history` annotations and some of the ones that are there have the wrong version number in them. Otherwise, assuming tests pass, I'm happy to that this gets pushed!

dfeltey:
@rfindler Is there a way to add a history annotation to a whole section of the reference since I added one for Collapsible Contracts (racket/contract/collapsible), or is it best to just add a history annotation to all of the functions in that section individually?

rfindler:
Looks like it can go "in a section that introduces a module". Probably that's better than what I suggested. I missed that they were in a new module!

dfeltey:
@rfindler There are a few that aren't in the new module (and need fixed annotations), but maybe I should rearrange them to all come from racket/contract/collapsible so that users who require racket/contract, racket/contract/base, etc don't get any of the collapsible things in scope.

There's 1 test that keeps failing in travis, but doesn't fail locally for me:
```
FAILED recursive-contract-not-too-slow rewrite-to-double-wrap
       got #f
       expected #t
```

I'm guessing the time limit there is too low for the travis machines, but is fine for my laptop.

rfindler:
Lets change that test so that it doesn't double wrap.

I think it is a good idea that requiring racket/contract and racket/contract/base should not include any new names.

racket/collects/racket/contract/private/collapsible-common.rkt
```diff
@@ -0,0 +1,335 @@
...
+  (define new-flat-list (collapsible-leaf/c-contract-list new-collapsible))
+  (define new-blame-list (collapsible-leaf/c-blame-list new-collapsible))
+  (define new-missing-party-list (collapsible-leaf/c-missing-party-list new-collapsible))
+  (define old-proj-list (collapsible-leaf/c-proj-list old-collapsible))
+  (define old-flat-list (collapsible-leaf/c-contract-list old-collapsible))
+  (define old-blame-list (collapsible-leaf/c-blame-list old-collapsible))
+  ;; We have to traverse the list to add the new neg party where it is missing
+  (define old-missing-party-list (add-missing-parties (collapsible-leaf/c-missing-party-list old-collapsible) old-neg))
+  (define-values (not-implied-projs not-implied-flats not-implied-blames not-implied-missing-parties)
+    (for/lists (_1 _2 _3 _4) ([new-proj (in-list new-proj-list)]
```
samth: I think this could use the new `#:result` keyword for `for/lists`.

racket/collects/racket/contract/private/merge-cache.rkt
```diff
@@ -0,0 +1,69 @@
...
+
+(provide define/merge-cache)
+
+(require (for-syntax racket/base))
+
+;; How to evict things from the cache after some length of time ...
+;; Which functions are safe to put in the cache?
+;; known-good-contract?
+;; log all the predicates used in the benchmark
+;; TODO: use a thread cell here ...
```
samth: Is this comment outdated? It looks like you are using a thread cell below.

racket/collects/racket/contract/private/merge-cache.rkt
```diff
@@ -0,0 +1,69 @@
...
+
+;; weak hashtable never cleared
+;; if reach a threshold, throw away and start over
+;; limit cache size to 10, and use some LRU policy (sort ...) (priority queue)
```
samth: How did you come up with the constant 10?

dfeltey: I think this comment is outdated and was referring to various experiments I tried when we first implemented the caching strategy. We don't ever clear out the cache anymore.

dfeltey:
@samth Re: the bigger things

1. I don't know if there's a good example of this that could go in the docs. Maybe for a flat contract, but otherwise right now Racket could only support collapsing for new kinds of vector or function contracts, and I think any example like that would be artificial and far too much code to go in the docs.

2. I can add a sentence of two that says that vector contracts and certain simple function contracts are collapsed currently, is that all you had in mind?

samth:
For 2, yes that's what I meant.

For 1, I'm confused by your statement. If I write a contract with a property created with `build-collapsible-contract-property`, it will get merged with other instances of that contract, right?

dfeltey:
For contracts on higher-order values that need chaperones or impersonators, unless that contract implements the complicated shuffling of unsafe and property-only wrappers to swap out contracts I think you would end up with something that can be merged, but isn't really collapsible when applied to a value multiple times

samth:
Ok, then I'm confused about what the role of the public-facing API is here. If I can't implement my own collapsible contracts, then the documentation should say that. But maybe the docs could just explain what I would need to do, even if it's tricky? I'd certainly like to implement this for contract combinators I've defined.

dfeltey:
The docs mostly exist for mine and @rfindler's benefit to be able to maintain the implementation, I don't think things are in good enough shape for people who don't regularly work on the internals of the contract system to be able to implement their own collapsible contracts for non-flat contracts.

samth:
Wooo!

PR #2370, opened 2018-11-14
Eval model

mflatt:
Recommending some adjustments to recent "eval-model.scrbl" changes.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/eval-model.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/eval-model.scrbl
@@ -179,7 +179,7 @@ Each evaluation step, then, transforms the current set of definitions and
 program into a new set of definitions and program. Before a
 @racket[define] can be moved into the set of definitions, its
 expression (i.e., its right-hand side) must be reduced to a @tech{value}.
-(The left-hand side is not evaluated.)
+(The left-hand side is not an expression position, and so it is not evaluated.)
 
 @prog-steps/no-obj[
 [{}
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -179,7 +179,7 @@ Each evaluation step, then, transforms the current set of definitions and
 program into a new set of definitions and program. Before a
 @racket[define] can be moved into the set of definitions, its
 expression (i.e., its right-hand side) must be reduced to a @tech{value}.
-(The left-hand side is not evaluated.)
+(The left-hand side is not an expression position, and so it is not evaluated.)
```
mflatt: Trying to address @lexi-lambda's comment, since I felt similarly about the change.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -414,8 +414,8 @@ changed, the value cannot be substituted for @racket[x] when
 the procedure is first applied.
 
 @margin-note{We do not use the term ``parameter variable'' to refer to
-the parameter variable names declared with a function. This is to avoid
-confusion with Parameters (see @secref["parameter-model"]).}
+the argument variable names declared with a function. This choice avoids
+confusion with @tech{parameters}.}
```
mflatt: Using `@tech` seems better to me. Also, I'd like to avoid "this" as a noun.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -556,9 +556,11 @@ module-specific prefix).
 @;------------------------------------------------------------------------
 @subsection[#:tag "module-phase"]{Phases}
 
+@guidealso["phases"]
+
 The purpose of @deftech{phases} is to
 address the necessary separation of names defined at execution time versus
-names defined at expansion time. See Section @secref["phases"] in the Guide.
+names defined at expansion time.
```
mflatt: The current Reference style is that it never depends on the Guide to define anything. Instead, it always cross-references the Guide in a margin notes. That's not obviously the right style choice, but we reconsider it, we'd probably want to reconsider more globally. Also, I don't think `@secref["phases"]` was going to the right place, since a `#:doc` argument is needed to link to another document.

Paul-C-Anagnostopoulos:
All these changes are lovely.

PR #2371, opened 2018-11-14
fasl: find shared values in hash tables.

samth:
Also, add test to check sharing of uninterned symbols.


samth:
@mflatt Despit the comment at the top of this loop, we need this to be able to reliably round-trip the result of expansion in Pycket, since the expander uses uninterned symbols.

mflatt:
This seems like a good change. But does the expander use uninterned symbols where it should use unreadable symbols?

samth:
The use of uninterned symbols is in expand/env.rkt. I'm not sure what the reasons to choose between them is, so I can't be sure if one is better here.

mflatt:
Ah, yes, that needs to be uninterned. Thanks for the explanation.

PR #2373, opened 2018-11-15
fasl: make sure every write-bytes know where to write

cderici:
adding tests that fail without this fix

### DIFF
--- a/pkgs/racket-test-core/tests/racket/fasl.rktl
+++ b/pkgs/racket-test-core/tests/racket/fasl.rktl
@@ -73,6 +73,13 @@
 (test #f eq? #rx#"hello" (fasl->s-exp (s-exp->fasl #rx#"hello") #:datum-intern? #f))
 (test #f eq? #px#"hello" (fasl->s-exp (s-exp->fasl #px#"hello") #:datum-intern? #f))
 
+(let* ([r1 #rx"[/\u5C][. ]+ap"]
+       [r2 #px"[/\u5C][. ]+ap"]
+       [r3 #px#"[\\][. ]+ap*"])
+  (test #t equal? r1 (fasl->s-exp (s-exp->fasl r1) #:datum-intern? #f))
+  (test #t equal? r2 (fasl->s-exp (s-exp->fasl r2) #:datum-intern? #f))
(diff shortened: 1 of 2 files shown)


pkgs/racket-test-core/tests/racket/fasl.rktl
```diff
@@ -72,6 +72,9 @@
 (test #f eq? #px"hello" (fasl->s-exp (s-exp->fasl #px"hello") #:datum-intern? #f))
 (test #f eq? #rx#"hello" (fasl->s-exp (s-exp->fasl #rx#"hello") #:datum-intern? #f))
 (test #f eq? #px#"hello" (fasl->s-exp (s-exp->fasl #px#"hello") #:datum-intern? #f))
+(test #f eq? #rx"[/\u5C][. ]+ap" (fasl->s-exp (s-exp->fasl #rx"[/\u5C][. ]+ap") #:datum-intern? #f))
```
samth: Can you change these to use `equal?` and check that they produce the same thing? Also avoid duplicating the regexp.

racket/collects/racket/fasl.rkt
```diff
@@ -1,13 +1,17 @@
...
-         "private/relative-path.rkt")
+         "private/relative-path.rkt"
+         (rename-in racket/base [write-byte r:write-byte]))
 
 (provide s-exp->fasl
          fasl->s-exp)
 
 ;; ----------------------------------------
 
+(define (write-byte byte out)
```
samth: Can you do this for `write-bytes` as well?

mflatt: Also, it would be good to have a comment explaining the reason for the wrapper, since it looks pointless at first glance.

FWIW, I could have changed all the `write-byte`s to some other name instead of renaming `write-byte` on import. But if that choice doesn't seem better to you, this way seems fine to me.

samth: I suggested this approach because it makes it harder to misuse, since you'd have to write `r:write-bytes` to do the wrong thing.

mflatt: That makes sense!

PR #2375, opened 2018-11-16
untar: treat the tag \0 as a synonym for '0'

afldcr:
I had a bit of trouble extracting an archive with `file/untar`. Apparently GNU tar (and presumably others) treat the type tag `0` as a synonym of `#\0`, where `file/untar` marks it as an unknown type tag.

### DIFF
--- a/racket/collects/file/untar.rkt
+++ b/racket/collects/file/untar.rkt
@@ -62,7 +62,7 @@
   (define checksum-bytes (read-bytes* 8 in))
   (define type-byte (integer->char (read-byte in)))
   (define type (case type-byte
-                 [(#\0) 'file]
+                 [(#\0 #\nul) 'file]
                  [(#\1) 'hard-link]
                  [(#\2) 'link]
                  [(#\3) 'character-special]


racket/collects/file/untar.rkt
```diff
@@ -62,7 +62,7 @@
   (define checksum-bytes (read-bytes* 8 in))
   (define type-byte (integer->char (read-byte in)))
   (define type (case type-byte
-                 [(#\0) 'file]
+                 [(#\0 0) 'file]
```
mflatt: Thanks for the patch! But do you mean #\nul here? The variable `type-byte` is misnamed, since the byte has been converted to a character....

afldcr:
Ahh, I do indeed! I'll fix this up a bit later today.

racket/collects/file/untar.rkt
afldcr: Nice catch; fixed!

mflatt: Thanks!

PR #2384, opened 2018-11-18
clarify examples in docs of take and drop

gus-massa:
I think that 

    (take '(1 2 3 4 5) 2) ;==> '(1 2)

is more clear than 

    (take '(1 2 3 4) 2)   ;==> '(1 2)

because in the second case there are `2` items that are keep and `2` items that disappear.


mflatt:
LGTM

Issue #2388, opened 2018-11-19
broken error message in `bytes-append`


mbutterick:
First:

```racket
(apply bytes-append '(42))
```

```text
bytes-append: contract violation
  expected: bytes?
  given: 42
```

And then:

```racket
(apply bytes-append '(42 42))
```

```text
bytes-append: contract violation
  expected: bytes?
  given: 42
```

But then, unaccountably:

```racket
(apply bytes-append '(42 42 42))
```

```text
invalid memory reference.  Some debugging context lost
```

gus-massa:
Thanks. Fixed in https://github.com/racket/racket/commit/b4e3030a0d0be48f3c06bbb590da0db953e2f7dd

PR #2389, opened 2018-11-19
rumble: add type check in bytes-append with more than two arguments

gus-massa:
The type of the argument was checked when `bytes-append` was called with 1 or 2 arguments, but the check was missing for the general case with more than 2 arguments. 

This was reported by @mbutterick in https://github.com/racket/racket/issues/2388 .

### DIFF
--- a/racket/src/cs/rumble/bytes.ss
+++ b/racket/src/cs/rumble/bytes.ss
@@ -81,12 +81,14 @@
      [(a b . l)
       (check who bytes? a)
       (check who bytes? b)
-      (and (bytevector=? a b)
+      (for-each (lambda (arg)
+                  (check who byte? arg))
+                l)
+      (and (do-name a b)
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks! But I think it would be better to move the check into the existing loop that computes the size of the resulting byte string.

gus-massa:
I've added also a few minor fixes to rumble\bytes.ss . (I'll squash all of them with a more generic name.)

For example. One is for fixing

    #lang racket
    (define x (bytes 0 1 2 3 4 5 6))
    (define y (bytes-append x))
    (bytes-set! y 0 64)
    (bytes-ref x 0)  ; ==> 0, not 64

and the other for

    (bytes<? #"3" #"2" 'no)  ; ==> error, not #f

mflatt:
Great! We need to add tests to the test suite, but I'm happy to do that.

PR #2397, opened 2018-11-23
glob: add 'glob-quote'

bennn:
add a function to escape any glob wildcards in a path or string

also add a private `glob-unquote` function so that, e.g., the pattern
`a\*` matches the file named `a*` (previously, the match would fail and
I think it was impossible to match for only `a*`)

### DIFF
--- a/racket/collects/file/private/glob.rkt
+++ b/racket/collects/file/private/glob.rkt
@@ -394,30 +394,23 @@
             (string c)])))
       (string-join str* ""))))
 
+;; glob-element->filename : (and/c path-string?
+;;                                 (not/c has-**?)
+;;                                 (not/c has-glob-pattern?)) -> path-string?
+;; Convert a pattern with no (unescaped) glob wildcards into a value suitable
+;; for use in `file-exists?` queries --- by removing the escape characters
+;; from the wildcards.
+;;
+;; (If a pattern has wildcards, then it is converted to a regexp and compared
+;;  to filenames via `directory-list` and `regexp-match?`.
+;;  If not, the pattern goes through this conversion function and is compared
+;;  to filenames using `file-exists?` and `directory-exists?`.)
 (define (glob-element->filename ps)
   (if (path? ps)
     (string->path (glob-element->filename/string (path->string ps)))
     (glob-element->filename/string ps)))
 
 (define GLOB-WILDCARD-CHAR* '(#\* #\? #\[ #\] #\{ #\} #\,))
-
-(define (glob-quote/string str)
-  (define str*
-    ;; add #\\ before all wildcards
-    (for/list ([c (in-string str)])
-      (if (memq c GLOB-WILDCARD-CHAR*)
-        (string #\\ c)
-        (string c))))
-  (apply string-append str*))
-
-(module+ test
-  (test-case "glob-quote/string"
-    (check-equal? (glob-quote/string "a") "a")
-    (check-equal? (glob-quote/string "a*") "a\\*")
-    (check-equal? (glob-quote/string "*][?") "\\*\\]\\[\\?")
-    (check-equal? (glob-quote/string "racket/**/base") "racket/\\*\\*/base")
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/file/scribblings/glob.scrbl
```diff
@@ -117,6 +119,19 @@ Examples:
   This operation accesses the filesystem.
 }
 
+@defproc*[([(glob-quote [pattern string?]) string?]
+           [(glob-quote [pattern path?]) path?]
```
AlexKnauth: Why are these called `pattern`? They won't be interpreted as patterns, but literal "quoted" strings or paths, right?

pkgs/racket-doc/file/scribblings/glob.scrbl
```diff
@@ -117,6 +119,19 @@ Examples:
...
+)]{
+  Escapes all @tech{glob wildcards} in the given string or path string,
+  but not braces.
```
AlexKnauth: How would you make a glob that should match only files that have open-curly-braces in them? If it's `"\\{"`, does `(glob-quote "{")` produce that?

bennn: `(glob "{" )` should work

but probably, `glob-quote` should optionally escape `{` `,` and `}` characters too

racket/collects/file/private/glob.rkt
```diff
@@ -388,6 +399,47 @@
             (string c)])))
       (string-join str* ""))))
 
+(define GLOB-WILDCARD-CHAR* '(#\* #\? #\[ #\] #\{ #\} #\,))
```
AlexKnauth: Since backslashes have special meaning as "escape the next char", backslashes need to be escaped too, turning `(string #\\)` into `(string #\\ #\\)`, right?

racket/collects/file/private/glob.rkt
```diff
@@ -388,6 +399,48 @@
...
+        [else
+         (cons (string (car c*)) (loop (cdr c*) (- i 1)))])))
+  (apply string-append (reverse str*)))
+
+(module+ test
+  (test-case "glob-unquote/string"
+    (check-equal? (glob-unquote/string "a") "a")
+    (check-equal? (glob-unquote/string "foo\\*rkt") "foo*rkt")
+    (check-equal? (glob-unquote/string "?\\?\\]\\[\\*") "??][*")
+    (check-equal? (glob-unquote/string "\\}a\\,") "}a,")))
```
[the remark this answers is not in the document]
bennn: Then maybe `glob-unquote` should be renamed to avoid confusion, because I don't think it needs to satisfy those properties.

What I'm thinking is, `glob` needs to do something special for every wildcard in a pattern. If the wildcard is un-escaped, then it needs to match a set of files. Otherwise if the wildcard is escaped, then it needs to put the wildcard character in the string. Right now `glob-unquote` is handling that "Otherwise" case. (And unfortunately the codes for handling escaped & unescaped wildcards are in 2 different places in this module.)

Put another way:
```
glob-quote : pattern -> pattern
glob-unquote : pattern -> filename
```

racket/collects/file/private/glob.rkt
bennn: I think this was wrong, `#\\` has no meaning by itself but things like `(string #\\ #\*)` have special meaning.

racket/collects/file/private/glob.rkt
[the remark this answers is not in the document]
AlexKnauth: So why should `glob-element->filename` on `\normal` return `\normal`? (written as `"\\normal"`) Why that and not `normal` or an error?

Looking more closely at how it's used, it makes more sense that it's just transforming one element. And in particular it's only used on elements where both `has-**?` and `has-glob-pattern?` have returned false. Should its signature be updated to include that constraint?

If that's the case, would `\normal` or `?mark` be invalid inputs?

AlexKnauth:
Okay, with that comment it looks good to me.

Issue #2400, opened 2018-11-28
cs: Bad error message from `raise-argument-error`


LiberalArtist:
The error message produced by `raise-argument-error` in the many-argument case is different in the latest Racket-on-Chez snapshot (7.1.0.6) than in Racket 7.1, seemingly incorrectly. Specifically, Racket-on-Chez omits the `argument position` message field altogether and gives the `other arguments` field without the `...`.

Example:
```
Sapientia:~ philip$ racket
Welcome to Racket v7.1.
> (raise-argument-error 'example "symbol?" 0 "bad" 'ok)
; example: contract violation
;   expected: symbol?
;   given: "bad"
;   argument position: 1st
; [,bt for context]
> ,bt
; example: contract violation
;   expected: symbol?
;   given: "bad"
;   argument position: 1st
;   other arguments...:
;    'ok
;   context...:
;    eval-one-top12
;    /Applications/Racket v7.1/share/pkgs/xrepl-lib/xrepl/xrepl.rkt:1477:0
;    /Applications/Racket v7.1/collects/racket/repl.rkt:11:26
> ^D
Sapientia:~ philip$ /Applications/Racket\ v7.1.0.6/bin/racket
Welcome to Racket v7.1.0.6 [cs].
> (raise-argument-error 'example "symbol?" 0 "bad" 'ok)
; example: contract violation
;   expected: symbol?
;   given: "bad"
;   other arguments:
;    'ok
; [,bt for context]
> ,bt
; example: contract violation
;   expected: symbol?
;   given: "bad"
;   other arguments:
;    'ok
;   context...:
;    do-raise-argument-error
;    /Users/racket/build/plt/build/cs/racket/src/build/cs//compiled/expander.scm:50452:5: body of top-level
;    call-with-values
;    /Users/racket/build/plt/build/cs/bundle/racket/share/pkgs/xrepl-lib/xrepl/xrepl.rkt:1477:0
;    /Users/racket/build/plt/build/cs/racket/collects/racket/repl.rkt:11:26
;    main.sps:621:4: thunk_6
```

LiberalArtist:
It might also be considered "bad" that Racket-on-Chez seems to report paths in terms of the build machine. (There is no such directory as `/Users/racket/` on my machine.)

LiberalArtist:
[...]

PR #2401, opened 2018-11-28
cs: fix error message from `raise-argument-error`

LiberalArtist:
Closes https://github.com/racket/racket/issues/2400


LiberalArtist:
Maybe there should also be a regression test, but I'm not sure where to add one.

LiberalArtist:
This PR doesn't attempt to address my comment on https://github.com/racket/racket/issues/2400 noting that the Racket-on-Chez seems to report paths in terms of the build machine. I hadn't noticed that until after I opened the original issue.

PR #2406, opened 2018-11-30
optimizer: (cdr <list?>) is a <list?>

gus-massa:
In case it is not an error, the result is a list anyway.

I'm not sure why I didn't add this the first time (a few years ago). Perhaps I was overcautious or perhaps I had not a clear understanding of the inclusion of errors in normal types. Unless I'm missing something, `expr_implies_predicate` doesn't ensure that the expression will not generate an error.

Compiling Racket this check is triggered approximately 1700 times, but I'm not sure how many reductions this enables down the road.


mflatt:
Sounds right, but add a test to "optimize.rktl"?

PR #2407, opened 2018-11-30
doc: logging, clarify 'none level and the level ordering

bennn:
add the `'none` log level to the intro paragraph,
 declare `'none` the level of "least detail",
 and **change** prose to say "subscribes to current level and lower"

I think the change is right, but I'll wait for someone to double-check

+2 small rewordings


mflatt:
LGTM

PR #2411, opened 2018-12-02
make terminal-port? and file-stream-port? total predicates

gus-massa:
Raise an error if the argument of `terminal-port?` is not a port, so it is consistent with the [docs](https://docs.racket-lang.org/reference/port-ops.html#%28def._%28%28quote._~23~25kernel%29._terminal-port~3f%29%29) and what I expect.

I didn't add any test for this. Perhaps I can put the tests in [port.rktl](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/port.rktl), but the current tests look very different because they are more complex.


rfindler:
I think predicates are generally more useful when they are total (returning
#false instead of raising exceptions).

Robby

gus-massa:
I agree, but `file-stream-port?` also requires that the argument is a port, and from the name I guess that both are restricted to ports. I don't mind a change of the spec for both. But I think it's better to not change `port-closed?`.

I think that this is a also a design problem with the predicates in the numerical tower, for example `integer?` is total but `exact?` is not. I never can remember which are total and which are restricted to numbers.

rfindler:
I agree with all that!

mflatt:
The intent was for `terminal-port?` to require a port argument. But, mainly for compatibility reasons, I'm not so sure that we should change the implementation to match the docs, instead of the other way around.

(Before looking at the patch, I assumed that this was a CS fix to add the restriction!)

gus-massa:
OK, let's go in the opposite direction and make `terminal-port?` and `file-stream-port?` total predicates. This is a minor backward incompatibility, but I guess nobody will notice.

I finally added the tests in [file.rktl](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/file.rktl), but they could be in [port.rktl](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/port.rktl) instead.

I tried to add the changes to the clasic Racket and RacketCS. Are these changes ok?

I think it's not necessary to change the types in typed-racket [base-env.rkt#L1847-L1848](https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/base-env/base-env.rkt#L1847-L1848), but perhaps they can be made slightly more strict (with `asym-pred`  or something?).

mflatt:
I'm happy with this change.

Unfortunately, I've made the merge more difficult by changing "file-stream.rkt", but the obvious analogous change should still work. Meanwhile, it looks like I changed "fd-port.rkt" and ended up making it consistent with traditional Racket, so I think that part of the patch won't be needed.

gus-massa:
I think I rebased this correctly. fd-port.rkt doesn't change in this version.

PR #2412, opened 2018-12-02
cs: more pure primitives (draft?)

gus-massa:
This PR has a lot of small changes in the table, so it needs a lot of review. There are five deletion of repetitions and the rest are just changes `known-procedure` -> `known-procedure/pure`. I'm not sure if I need to use a more strict definition of `pure`. If there are too many errors, please mark a few and I'll try to find the others.

Most of the primitives that I marked as `pure` are predicates. They must be total (and immutable) predicates. I tried to filter the wrong cases just reading the manuals and also with some test to ensure that `(something? #t)` and `(something? 7.7)` and a few more cases don't raise an error. During this I discovered the difference between the implementation and the docs of `terminal-port?`. See also https://github.com/racket/racket/pull/2411 . As I said there, it's very difficult to guess which "predicates" are total and which have a restricted domain.

It's important to notice that there are a few primitives that almost look like a pure predicate, but are not. In particular:

*  `struct?` depends on the value of the `(current-code-inspector)` (Chez Scheme has an internal primitive `secret-struct-under-the-hood?` that may be useful, IIRC the real name is `#3%$record?`.)

* `prefab-key?` may change because the fourth item in the list is a vector that may be mutable or have a impersonator.

I tried to be careful, but some cases are tricky and I'm not sure that I removed all the wrong cases.


mflatt:
To turn that into a more concrete suggestion: I think you should remove `values` and merge, and then we should pick a better word to replace "pure".

gus-massa:
Should I write a comment next to the tricky ones, or you prefer to keep the file clean?

mflatt:
Adding short comments on the non-"pure" ones sounds like a good idea.

gus-massa:
I added also a comments to

* `make-pseudo-random-generator`: It depends on the the `current-milliseconds`. I don't expect to get complains if this is moved, but I think it's technically incorrect to mark it as pure.

mflatt:
LGTM

gus-massa:
Merged in https://github.com/racket/racket/commit/561fe7e2e79360d6e764d5a098b9e5e8ebe647d4

PR #2413, opened 2018-12-04
fix pair accessor error messages

bennn:
Change the error message for some functions like `caddr` so they
describe pair structures that they expect.

- - -

Today I tried to run some code that I hadn't recompiled and saw an error message from `caaddr` (that I didn't save). I read the error message, and it described a different kind of pair structure than `caaddr`. Deleting the `compiled/` folder fixed that issue, but it might be something else to look into.

### DIFF
--- a/pkgs/racket-test-extra/tests/racket/pair-accessor-error-message.rkt
+++ b/pkgs/racket-test-extra/tests/racket/pair-accessor-error-message.rkt
@@ -1,0 +1,88 @@
+#lang racket/base
+
+;; Check that the contract error messages for the `c*r` accessors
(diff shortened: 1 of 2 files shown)


bennn:
Oh, this pull request also has the test I used to see what accessors gave bad messages. The original output of that test is attached.
[accessors.txt](https://github.com/racket/racket/files/2642467/accessors.txt)

mflatt:
These changes look good to me.

When you saw the error with `caaddr`, were you running a build that includes the recent commit 3a198092682455638042bac961be0d6de71446fa?

bennn:
No, I don't think so (but I'm not sure).

PR #2414, opened 2018-12-05
cs: add unsafe-poll-fd, unsafe-fd->evt

rmculpepper:
This provides an alternative to unsafe-{file-descriptor,socket}->semaphore, which are not yet implemented on RacketCS.

I followed the lead of unsafe-poll-ctx-fd-wakeup in always interpreting the "file descriptor" as a socket and requiring it to be a fixnum, but is there a guarantee that a SOCKET on 32-bit Windows fits in a fixnum?

I've tested versions of the unix-socket and zeromq-r libraries modified to use unsafe-fd->evt.


pkgs/racket-doc/scribblings/foreign/port.scrbl
```diff
@@ -92,3 +109,41 @@ must be unregistered before the file descriptor or socket is closed.
...
+@item{a subsequent call occurred with the same @racket[fd] and with
+@racket['remove] for @racket[mode] (once removed, the event is
+perpetually ready).}
+
+]
+
+The synchronization result of the event is the event itself.
+
+The @racket['check-read] and @racket['check-write] modes are like
+@racket['read] and @racket['write], but the result if @racket[#f] if
```
mflatt: "if" -> "is"

mflatt:
This looks like a great addition!

After refreshing my memory, I regret making `unsafe-poll-ctx-fd-wakeup` assume a socket. In fact, the CS implementation assumes a non-socket. I think it should have an optional `socket?` argument, and I guess that argument has to default to `#t` for compatibility. The optional argument is easy to support in your implementation, of course, so I encourage that repair while I fix up `unsafe-poll-ctx-fd-wakeup`.

I don't think the fixnum requirement is a good idea. Functions like `unsafe-file-descriptor->port` and `unsafe-socket->port` accept any exact integer that fits an `intrptr_t`. So, I'd go with `exact-integer?`, and if a caller supplies an invalid file descriptor, than anything can happen, including an error naming a fictitious `handle-integer?` contract.

The fact that the table of registered file descriptors depends on the `ffi/unsafe/port` instance is different than a table that's built into the lower layer, but I can't think of any problem that would cause.

mflatt:
@rmculpepper I lost track of this one (and I never improved `unsafe-poll-ctx-fd-wakeup`). It looks like the revised commits may address my issues, in which case I think it should be merged with updated version numbers.

rmculpepper:
I didn't push it because when I was testing, I discovered that every so often one of the zeromq tests would hang, and I didn't figure out why before I had to switch contexts to something else. I saw you recently fixed something related to unsafe-poller, so maybe that was the same issue. I'll rebase and test again on Monday.

rmculpepper:
Okay, I've rebased and tested with the cs branches of the unix-socket and zeromq-r packages. The tests pass and I haven't gotten any more freezes, so I think this is ready.

mflatt:
Sounds good!

PR #2415, opened 2018-12-05
Allow explicitly providing a parent logger to define-logger

lexi-lambda:
This change makes it possible to adjust the parent of a use of `define-logger` at the module top-level without needing to mutate `current-logger`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/logging.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/logging.scrbl
@@ -132,7 +132,9 @@ Reports @racket[logger]'s default topic, if any.}
 A @tech{parameter} that determines the @tech{current logger}.}
 
 
-@defform[(define-logger id)]{
+@defform[(define-logger id maybe-parent)
+         #:grammar ([maybe-parent (code:line) (code:line #:parent parent-expr)])
+         #:contracts ([parent-expr (or/c logger? #f)])]{
 
 Defines @racketkeywordfont{log-}@racket[id]@racketkeywordfont{-fatal},
 @racketkeywordfont{log-}@racket[id]@racketkeywordfont{-error},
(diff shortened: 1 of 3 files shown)


racket/collects/racket/private/logger.rkt
```diff
@@ -27,15 +27,17 @@
   (define-syntax log-info (make-define-log 'info #'(current-logger) #'(logger-name l)))
   (define-syntax log-debug (make-define-log 'debug #'(current-logger) #'(logger-name l)))
 
-  (define (check-logger who)
-    (lambda (v)
-      (unless (logger? v)
-        (raise-argument-error who "logger?" v))
-      v))
+  (define (check-logger who v)
```
samth: Were there no uses of `check-logger` elsewhere?

lexi-lambda: Not as far as I can tell.

mflatt:
This is ok with me, but consider adding a `#:parent` keyword before the parent-logger expression, in case we want to add other options in the future.

lexi-lambda:
Good point. I’ve pushed an updated version that uses a `#:parent` keyword.

abmclin:
I am thrilled with this new change. I found it annoying having to parameterize `current-logger` before using `define-logger`

mflatt:
LGTM

pkgs/racket-doc/scribblings/reference/logging.scrbl
```diff
@@ -132,7 +132,9 @@ Reports @racket[logger]'s default topic, if any.}
 A @tech{parameter} that determines the @tech{current logger}.}
 
 
-@defform[(define-logger id)]{
+@defform[(define-logger id maybe-parent)
+         #:grammar ([maybe-parent (code:line) (code:line #:parent parent-expr)])
+         #:contracts ([parent-expr logger?])]{
```
bennn: Can the contract be `(or/c logger? #false)` instead?

lexi-lambda: Good catch. I’ve pushed an updated version that allows `#f`.

pkgs/racket-doc/scribblings/reference/logging.scrbl
```diff
@@ -142,10 +144,14 @@ Defines @racketkeywordfont{log-}@racket[id]@racketkeywordfont{-fatal},
 like @racket[log-fatal], @racket[log-error],@racket[log-warning],
 @racket[log-info], and @racket[log-debug]. The @racket[define-logger]
 form also defines @racket[id]@racketidfont{-logger}, which is a logger with
-default topic @racket['@#,racket[id]] that is a child of @racket[(current-logger)];
-the @racketkeywordfont{log-}@racket[id]@racketkeywordfont{-fatal},
-@|etc| forms use this new logger. The new logger is
-created when @racket[define-logger] is evaluated.}
+default topic @racket['@#,racket[id]] that is a child of the result of
+@racket[parent-expr], or of @racket[(current-logger)] if
+@racket[parent-expr] not provided; the
```
bennn: `parent-expr` might be not provided or `#false` now

lexi-lambda: `#f` means a different thing from specifying `(current-logger)`… do you think it needs to be reworded somehow, given that?

bennn: Oh, I misunderstood what `#f` means.

But yes I think it should be reworded. Maybe add a parenthetical like the one for `make-logger`, something like:

> .... that is a child of the result of `parent-expr` (when `parent-expr` is not `#f`), or of ....

Or maybe point to `make-logger` (this suggestion doesn't sound as good):

> The `define-logger` form also defines a new logger via `make-logger`; if `parent-expr` is not provided then `(current-logger)` is used.

lexi-lambda: I’ve updated the documentation to include your first suggestion.

PR #2416, opened 2018-12-07
A few small changes.

samth:


racket/collects/compiler/cm.rkt
```diff
@@ -2,7 +2,7 @@
 (require "private/cm-minimal.rkt"
          (submod "private/cm-minimal.rkt" cm-internal)
          racket/contract/base
-         racket/place
+         '#%place
```
mflatt: This is ok, but I can't help thinking it's better to avoid `#%` modules in all but the most primitive modules. How about adding `racket/place-channel` or `racket/private/place-channel` to just re-exports place channels from `'#%place`?

samth:
This turns out to break with places disabled. I'm trying to understand what `racket/place` does in that case.

PR #2421, opened 2018-12-10
Discuss the mutability of the result of quasiquote

shhyou:
Explicitly state that `quasiquote` allocates mutable vectors, mutable boxes and immutable hashes.


mflatt:
Looks good - thanks!

PR #2427, opened 2018-12-15
improve error-handling in uni-norm test

bennn:
+ ensure that getting the test path returns a path (and never `#f`)
+ check for `eof`
+ delete bad downloaded files
+ pause between download attempts

- - -

This test case may still fail, but if it does the error messages should be better than what's been in some recent builds, e.g.:

```
Section(uni-norm)
ERROR: with-input-from-file: contract violation
  expected: path-string?
  given: #f
The command "raco test -l tests/racket/test" exited with 2.
```


mflatt:
Thanks!

bennn:
Ok, now if the download fails the test issues a `log-warning` and "skips" the normal testing.

Is there a better way than `log-warning` to write a message that won't cause a CI failure?

PR #2430, opened 2018-12-17
reverse order of bits in mask

jbclements:
It appears to me that the mask values provided in the permissions documentation are "flipped" in the sense that world-readable is given as #o001 rather than #o004, meaning that #o444 would represent --w--w--w rather than r--r--r--.  This pull request reverses that. 

Am I missing something?

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
@@ -402,15 +402,15 @@ portable, reflecting permissions for the file or directory's owner,
 members of the file or directory's group, or other users:
 
 @itemlist[
- @item{@racketvalfont{#o100} : owner has read permission}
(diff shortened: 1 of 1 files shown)


jbclements:
Sigh... pull request is correct (I believe), but the text of my comment above is wrong, and should replace --w--w--w with --x--x--x (duh) to read:

"It appears to me that the mask values provided in the permissions documentation are "flipped" in the sense that world-readable is given as #o001 rather than #o004, meaning that #o444 would represent --x--x--x rather than r--r--r--. This pull request reverses that."

AlexKnauth:
I don’t know very much about `file-or-directory-permissions`. I haven’t used this particular function in any Racket programs yet so I think I should get a little more familiar before I should review this...

jbclements:
NP, thanks very much. BTW, did you get an e-mail when I tagged you?

John

PR #2441, opened 2018-12-31
Try to add ubsan to travis (again) (not completed)

gus-massa:
I tried to complete https://github.com/racket/racket/pull/2297 but I was not successful. I'm submitting this in case someone more familiar with travis or gcc can fix the remainder problems.

(I put this version in the allow failure section, at least for now.)

The idea is to update gcc, because it is too old in travis. More details in https://docs.travis-ci.com/user/languages/c/#gcc-on-linux

I still got this error: 

    /usr/bin/ld: unrecognized option '--push-state--no-as-needed'

This post in S.O. may be relevant. https://stackoverflow.com/questions/50024731/ld-unrecognized-option-push-state-no-as-needed I tried something like that in my last commit, but it is not working.

I'll probably not continue working on this for now, so anyone that like to complete it is welcome.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,14 +20,37 @@ matrix:
       --disable-extflonum"
   - os: linux
     compiler: gcc
-    env: PATH=./racket/bin:$PATH RACKET_CONFIGURE_ARGS="--enable-ubsan"
-  - os: linux
-    compiler: gcc
     env: PATH=./racket/bin:$PATH RACKET_CONFIGURE_ARGS="--disable-jit"
   - os: linux
     compiler: gcc
     env: PATH=./racket/bin:$PATH RACKET_CONFIGURE_ARGS="--disable-jit --disable-places
       --disable-futures --disable-extflonum"
+  - os: linux
+    #compiler: gcc
+    addons:
+      apt:
+        sources:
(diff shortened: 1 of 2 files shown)


pmatos:
This is a bottomless pit of misery. No point in spending more time on this. My suggestions is we use Gitlab for CI only. This will open many doors. I have proposed #2325 

I will make a case for this in the mailing list later on.

pmatos:
We have this now running in https://gitlab.com/racket/racket
Shall we close this PR?

gus-massa:
Since this is not progressing, let's close it until someone is interested.

(Warning for the future: It may be more difficult than what it looks.)

gus-massa:
I rebased it and moved `-fuse-ld=gold` from .travis.yml to cfg-racket, and it works now. I don't understand why the flag in .travis.yml didn't work. 

I left the last change in a separate commit, so this need some squashing.

The main problem is that it needs approximately 50 minutes in travis-ci, so sometimes it success and sometimes it timeout.

pmatos:
Is there a point to this given we are already doing it in gitlab?

pmatos:
We have this in gitlab and soon in Actions. Closing.

PR #2442, opened 2018-12-31
improve performance of wrap-expr/c

rmculpepper:
Not ready for merging yet: needs doc updates and performance testing.


rfindler:
I am not sure about the swapping (in syntax/parse/experimental/contract.rkt), but otherwise this looks like the right idea to me!

One minor thing I wasn't sure about: why did you make context limit be an optional argument instead of supplying `#f` at the new call site?

rfindler:
OH, whoops. Every application that doesn't depend on `v` needs to be pulled out of the newly-introduced lambda in racket/contract/private/base.rkt. And probably that means duplicating with the `with-contract-continuation-mark` calls, since probably all that work should count.

rmculpepper:
re swapping in expr/c: I hope that most people just used the default arguments. I'll see if I can do a scan through the package catalog to check.

re context limit being optional: No particular reason. Actually, maybe it would be better if more of those arguments were keyword args with defaults?

re pulling more work out of the lambdas: Done, just pushed a new commit (which I might rebase before merging).

I ran a simple performance test using `in-queue` and with these changes the time dropped to nearly nothing.

mfelleisen:
> On Dec 31, 2018, at 8:20 PM, Ryan Culpepper <notifications@github.com> wrote:
> 
> I ran a simple performance test using in-queue and with these changes the time dropped to nearly nothing.


You should go over all contracts :-) Happy New Year.

rfindler:
re "re context limit being optional: No particular reason. Actually, maybe it would be better if more of those arguments were keyword args with defaults?", lets leave it as not being optional. If we wanted to make an external layer for public consumption, then I agree with the keyword arguments, thing. 

re the simple performance test: great! Sounds like this is successful! Thank you.

rmculpepper:
Okay, I've pushed an update that exports `make-apply-contract` only from `racket/contract/private/base`, not from `racket/contract/base`, and it makes the context-limit argument mandatory. I've also updated the docs.

The other thing we discussed by email was eliminating the `macro-arg/c` combinator in favor of adding an argument to swap the initial blame object, but I think that can wait.

Let me know if there's anything else before this should be merged.

rfindler:
I think that wrap-expr/c needs more discussion of when/how things lift. Indeed, I think that probably the docs should say pretty strongly that if the expression doesn't lift, then the performance is substantially different than if it does (and maybe go so far as to say that if backwards compatibility weren't a concern, then there would be a syntax error if it didn't lift (or at least require an extra keyword making it clear that not lifting is desired)).

I would prefer an approach like that to one that is just saying "undefined behavior, whee!" :)

Also, can this be a single commit? (I don't have a strong opinion about that, tho.)

rfindler:
re: "The ->i version performs an order of magnitude better than the custom
contracts that seem like they should be doing the same thing. (WHY?)"

there are lots of special cases here and there. Maybe the code that ->i generates has a fast path somewhere that you need?

re: "The error messages seem appropriate; they do not mention the ->i
contract (good)... can I rely on this?"

Can you be more specific (perhaps with an example that has only the `->i` in it an a comment there?)

rmculpepper:
re `->i` fast paths: It's faster even going through `make-apply-contract`, so the speedup isn't coming from `define-module-boundary-contract`. I looked briefly at the `->i` projection and `mk-wrapper` and nothing obvious jumped out at me. Can you take a look at the commented-out code in `macro-dep-arg/c` and see if anything occurs to you?

re appropriate errors: For example, in the context of pkgs/racket-test/tests/syntax/contact/perf-nolift.rkt:

    > (foo 'bad)
    ; foo: contract violation
    ;   expected: point?
    ;   given: 'bad
    ;   in: point?
    ;   contract from:
    ;       <pkgs>/racket-test/tests/syntax/contract/perf-nolift.rkt
    ;   blaming: <pkgs>/racket-test/tests/syntax/contract/perf-nolift.rkt
    ;    (assuming the contract is correct)
    ;   at: readline-input:2.5

This is the error message that I want, but I was worried that I would see something like "in the 2nd argument of (->i ....)". Or that a future change to the contract library will add text like that to the error message. Should I rely on this aspect of `->i`'s error messages? Or would you rather leave this alone for now and revisit it if the message actually changes in the future? Or is there a way to get the performance of `->i` with explicit control over the error message?

BTW, I can squash these commits once they're ready to merge.

rmculpepper:
Whoops, I do see the message I don't want. (I think I was using `enter!` before and that threw off my experiment.) I've added a failing test case.

rfindler:
@rmculpepper looks like the usual thing of ignoring the currying. That's not a good idea. There are also some tweaks in the `->i` code for some special cases (handling, I think, flat contracts in arguments/results more efficiently that mattered in some benchmark or the other).

rfindler:
Oh, whoops, no, this is dependent. Sorry, losing my mind. The flat check might be it, tho.

rfindler:
This is the function that does the flat check:

   https://github.com/racket/racket/blob/master/racket/collects/racket/contract/private/arr-i.rkt#L1108

If you comment out the first two `cond` clauses, you'll disable the fast path.

rmculpepper:
@rfindler thanks, adding a fast path for flat contracts worked. That is, on the microbenchmarks `expr/c` is now faster than `->i` for flat contracts and only about 25% slower than `->i` for higher-order contracts.

I've changed the code so that only contract expressions that are simple variable references are lifted, to minimize the chance of changing behavior.

rfindler:
I'm unsure of the (long term) ramifications of having conditions like that in our public interfaces. Do you think it might be better to change things so that wrap-expr/c stays the same (performs badly) and then something new does lifting (and raises errors when it cannot) becomes the suggested alternative in the docs?

I'm not saying this out of any deep knowledge of the apis -- maybe there is a reason that conservatism isn't needed here.

rmculpepper:
I think the current behavior in the PR is nearly optimal for `wrap-expr/c`. I expect that most uses (directly and via `expr/c`) involve a contract expression that is a variable reference, and lifting those is usually safe. (It would be perfectly safe if I could test (at phase 1) whether an identifier that refers to an imported value binding at phase 0 is constant. I found `variable-reference-constant?`, but I don't see how to get an appropriate variable reference object at compile time.)

On the other hand, `wrap-expr/c` isn't optimal for macro contracts. The ideal situation, which would more closely mimic what `(provide (contract-out ....))` does, would be to lift non-dependent contracts at the point of the macro definition, rather than every use site. But that can't be done with the `wrap-expr/c` interface, and it can't be done with `expr/c` as an ordinary syntax class either.

rfindler:
Okay. Thanks, Ryan.

PR #2445, opened 2019-01-03
Add `udp-set-receive-buffer-size!`

greghendershott:


greghendershott:
Actually, please wait. I pushed a commit adding tests. I should have had those in the first place. Also, those will let CI help try on Windows and CS.

mflatt:
This looks great! I see a couple of very small things, like a typical hassle involving "exact integer" versus "fixnum", but I'm happy to make little adjustments after merging.

PR #2452, opened 2019-01-17
fix {DYLD,LD}_LIBRARY_PATH for starter with embedded dll-dir

### LINKED ISSUE #2450: LD_LIBRARY_PATH is not taking effect
Hello. I have an application which breaks when I try to run it like this `raco exe --gui app.rkt && ./app` with "SSL not available" error.

Sample source code:
```
~/playground λ cat libssl-load-bug.rkt
#lang racket/base
(require openssl)
(printf "ssl load error: ~a\n\n" ssl-load-fail-reason)
```

Environment(I use [NixOS](https://nixos.org)):
```
~/playground λ racket --version
Welcome to Racket v7.1.
~/playground λ echo $LD_LIBRARY_PATH

~/playground λ ls -la /usr/lib /lib
/lib: No such file or directory
/usr/lib: No such file or directory
```

Example:
```
~/playground λ raco exe --gui ./libssl-load-bug.rkt && ./libssl-load-bug
ssl load error: ffi-lib: couldn't open "libcrypto.so" (libcrypto.so: cannot open shared object file: No such file or directory)
```

`LD_LIBRARY_PATH` with correct openssl lib path should fix this, but it is not:
```
~/playground λ LD_LIBRARY_PATH=/nix/store/...-openssl-1.0.2q/lib ./libssl-load-bug
ssl load error: ffi-lib: couldn't open "libcrypto.so" (libcrypto.so: cannot open shared object file: No such file or directory)
```

It worked only with this `LD_LIBRARY_PATH` value(notice `:` after `=`):
```
~/playground λ LD_LIBRARY_PATH=:/nix/store/...-openssl-1.0.2q/lib ./libssl-load-bug
ssl load error: #f
```

It should work without colon in `LD_LIBRARY_PATH`.

rmculpepper:
The adjustment previously had the separator in the wrong position.

Closes #2450.

### DIFF
--- a/racket/src/start/ustart.c
+++ b/racket/src/start/ustart.c
@@ -490,7 +490,7 @@ int main(int argc, char **argv)
       dll_path = "";
     }
     dll_path = string_append(dll_path, ":");
-    dll_path = string_append(lib_path, dll_path);
+    dll_path = string_append(dll_path, lib_path);
     dll_path = string_append(LD_LIB_PATH "=", dll_path);
     putenv(dll_path);
   }


mflatt:
Thanks for looking into this! But should the repair be a swap on the next line, instead? I think the embedded path should probably be added before others, instead of after.

rmculpepper:
Yes, that seems better. I've pushed an update.

PR #2462, opened 2019-02-07
syntax/parse: Add prop:syntax-class

lexi-lambda:
This makes it possible for users to create bindings that function as both syntax classes and other things, in the usual way using struct type properties.


lexi-lambda:
@rmculpepper Pinging you on reviewing this.

lexi-lambda:
To answer your question: @michaelballantyne’s [syntax-generics2](https://github.com/michaelballantyne/syntax-generic2) library defines a thing called a “syntax generic”, and defining a syntax generic does the usual thing of storing some compile-time information in a binding using `define-syntax` / `syntax-local-value`. One of the things you can do with a syntax generic is test if a particular piece of syntax matches the generic, so I thought it would be nice if a syntax generic binding also functioned as a syntax class that just does the relevant test inside a `#:when`.

lexi-lambda:
Thank you for the review!

PR #2464, opened 2019-02-09
net/url: Adds patch-pure-port and patch-impure-port 

DarrenN:
## Background

[`net/url`](https://docs.racket-lang.org/net/url.html?q=net%2Furl) was missing functions to make HTTP requests with the [`PATCH`](https://en.wikipedia.org/wiki/Patch_verb) verb. API endpoints accepting `PATCH` are becoming more common.

This PR adds `patch-pure-port` and `patch-impure-port`.

### DIFF
--- a/pkgs/net-doc/net/scribblings/url.scrbl
+++ b/pkgs/net-doc/net/scribblings/url.scrbl
@@ -361,7 +361,6 @@ port} contains both the returned headers and the body. The
                         [post bytes?]
                         [header (listof string?) null])
          input-port?]
-)]
 @defproc[(patch-pure-port [URL url?]
                           [post bytes?]
                           [header (listof string?) null])
@@ -386,7 +385,6 @@ strings can be used to send header lines to the server.
                           [post bytes?]
                           [header (listof string?) null])
          input-port?]
-)]
 @defproc[(patch-impure-port [URL url?]
                             [post bytes?]
                             [header (listof string?) null])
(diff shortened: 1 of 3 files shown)


DarrenN:
Hmmm the two fails seem to be from an unrelated package? https://travis-ci.org/racket/racket/jobs/491101788#L2808

jeapostrophe:
Hi and thanks for the patch. As @greghendershott mentioned on Slack, `net/url` should really be seen as a legacy library. Rather than adding a PATCH method, I'd rather expose the generic one with a better name so that changes like this are not needed in the future.

DarrenN:
@jeapostrophe do you think there would be value in marking this library as deprecated in the documentation?

jeapostrophe:
Sure

greghendershott:
In my experience with Racket, it's very rare for things to be marked as "deprecated" in the sense of "We want to remove X and break working applications, because reasons. We hope doing something rude will seem less rude if we do it slowly." :smile:

So I think the most helpful thing would be a `@margin-note` near the top of `net/url` explaining that `http-client` is recommended for new applications, and any changes to `net/url` would be just bug fixes.

At least, that's my $0.02, but I'm not on the hook to maintain `net/url` (just some libs and pkgs that still use it).

jeapostrophe:
I was thinking of a note like the one for net/cookies

https://docs.racket-lang.org/net/cookie.html

Issue #2465, opened 2019-02-10
define-inline doesn't work on a function of 0 arity


sorawee:
```
(define-inline (f) 1)
```

results in:

```
/Applications/Racket v7.1/collects/racket/performance-hint.rkt:91:15: with-syntax*: bad syntax
  in: (with-syntax* () (syntax (let* () (let-syntax ((f (make-rename-transformer (quote-syntax f-internal)))) 1))))
  location...:
   /Applications/Racket v7.1/collects/racket/performance-hint.rkt:91:15
```

Note that the grammar at https://docs.racket-lang.org/reference/performance-hint.html#%28form._%28%28lib._racket%2Fperformance-hint..rkt%29._define-inline%29%29 seems to allow a function of 0 arity.

sorawee:
Fixed by #2466

PR #2466, opened 2019-02-11
Fix with-syntax* when it has no patterns

gus-massa:
Fix `with-syntax*` when it has no patterns. 

This also fix `define-inline` for 0-arity functions. See the bug reported in #2465.

### DIFF
--- a/racket/collects/racket/syntax.rkt
+++ b/racket/collects/racket/syntax.rkt
@@ -205,6 +205,8 @@
 
 (define-syntax (with-syntax* stx)
   (syntax-case stx ()
-    [(_ (cl) body ...) #'(with-syntax (cl) body ...)]
+    [(_ () body ...) (syntax/loc stx (let () body ...))]
+    [(_ (cl) body ...) (syntax/loc stx (with-syntax (cl) body ...))]
     [(_ (cl cls ...) body ...)
-     #'(with-syntax (cl) (with-syntax* (cls ...) body ...))]))
+     (with-syntax ([with-syntax/rest (syntax/loc stx (with-syntax* (cls ...) body ...))])
+       (syntax/loc stx (with-syntax (cl) with-syntax/rest)))]))
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -5216,6 +5216,10 @@
            (test #t equal? output
                  (bytes->string/utf-8 (get-output-bytes O #t)))))
   ;;
+  (define-inline (f0) (+ 1 0))
+  (test/output (f0)
+               1 "")
+  ;;
   (define-inline (f x) (+ x x))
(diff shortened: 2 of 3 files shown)


racket/collects/racket/syntax.rkt
```diff
@@ -205,6 +205,7 @@
 
 (define-syntax (with-syntax* stx)
   (syntax-case stx ()
+    [(_ () body ...) #'(begin body ...)]
```
samth: Should this be let instead of begin?

gus-massa: According to the docs, yes.

I though about this, but after looking at the implementation of `with-syntax` (without the star), I think that `with-syntax` (without the star) use `begin` instead of `let()`.

Perhaps we should change the implementation of `with-syntax` (without the star), or perhaps I made a mistake reading it.

gus-massa: After thinking it for another while, let's change both `with-syntax` and `with-syntax*`.

gus-massa:
As suggested by @samth, I changed the wrapper for this special case from `begin` to `let()`. To be consistent, I made the same change in `with-syntax` (without the star). This will break programs like

    #lang racket
    (with-syntax ()
      (define x 0))
    x ; ==> 0

racket/collects/racket/syntax.rkt
```diff
@@ -205,6 +205,7 @@
 
 (define-syntax (with-syntax* stx)
   (syntax-case stx ()
+    [(_ () body ...) #'(let () body ...)]
```
sorawee: Use `syntax/loc`?

gus-massa:
Added syntax location.

samth:
If we're worried about backwards compatibility for the change to `with-syntax`, we can use `(eq? (syntax-local-context) 'expression)` as a check for using the `(let () ...)` expansion.

florence:
It seems to me like its a bug to have `with-syntax` splice internal definitions at all. Especially since if `with-syntax` does bind anything, its suddenly not splicing:

```
#lang racket
(require racket/syntax)
(with-syntax ([x #'y])
  (define y 1)
  1)
y ; ==> y: unbound identifier in: y
```

So maybe we should go with the consistent if mildly less backwards compatible behavior?
(and if we do go with the inconsistent behavior we should at least document this confusing oddity).

gus-massa:
Now I think that it's better to make the change. The docs of `with-syntax` say it expands to a `let()`. And there are some strange examples with the current behaviour

    (if (with-syntax ([x #'7])
          (define temp 7)
          (eq? (syntax-e #'x) temp))
        1
        2)
    ; ==> 1
    
    (if (with-syntax ()
          (define temp 7)
          (eq? (syntax-e #'x) temp))
        1
        2)
    ; ==> error: define: not allowed in an expression context

mflatt:
I agree – this change can be merged.

PR #2468, opened 2019-02-13
Fix possible memory leak in loop

### LINKED ISSUE #2467: Possible leak for argument
I wanted to confirm my suspicions here before submitting a PR.

Is it the case that here:
https://github.com/racket/racket/blob/f8a130851f1541b1a89d4d75e79a8b36bd50d3f3/racket/src/rktio/rktio_fs.c#L920

argument `filename` is overwritten and the memory it pointed to leaks? Or is filename somehow under GC and the GC will free it during the next pass?

In general, in the rktio layer, are all arguments under GC? I assume not everything is under GC because I see calls to `free`.

pmatos:
Fixes #2467 .

### DIFF
--- a/racket/src/rktio/rktio_fs.c
+++ b/racket/src/rktio/rktio_fs.c
@@ -915,7 +915,8 @@ int rktio_make_directory(rktio_t *rktio, const char *filename)
   /* Make sure path doesn't have trailing separator: */
   len = strlen(filename);
   while (len && IS_A_SEP(filename[len - 1])) {
-    copied = MSC_IZE(strdup)(filename);
+    if (!copied)
+      copied = MSC_IZE(strdup)(filename);
     copied[--len] = 0;
     filename = copied;
   }


mflatt:
I agree that there's a potential leak currently, but I think the solution is to guard `copied = MSC_IZE(strdup)(filename);` with `if (!copied)`. When `copied` is allocated, then it needs to be freed sometime after the loop and before the function returns, so the calls to `free` look to me like they're in the right place..

Using `alloca` here is not a good choice, since there's no a priori limit on the path length.

pmatos:
I understand your suggestion but don't understand what the problem is with `alloca`. `alloca` will ensure your memory will be released and allocates the memory only once. The fact that there's no a priori limit is the same with `strdup`, except with `strdup` you're allocatting a new string each time you loop.

pmatos:
I re-read your suggestion and understand that it will solve the repeated allocation and leak problem. However, I would like to understand the issue against `alloca`. The advantage of using it here is that you don't need to `free` the allocated space explicitly.

pmatos:
You're probably aware but the checks are failing because of: https://github.com/racket/racket/commit/4ed5d7d98b8f9f901eb055ef686f8e98a5814a6a#commitcomment-32305044

mflatt:
While `strdup` allocates on the heap, `alloca` allocates on the stack — so it's more constrained in allocation size by several orders of magnitude.

pmatos:
Doh! You are right. I forgot about that point. It's quite late here. Let me sort out a new PR tomorrow morning.

pmatos:
@mflatt apologies for my brain freeze yesterday. All engines back to work. I have pushed a new commit and squashed all others. Hope it's ok now.

PR #2472, opened 2019-02-14
Mark scheme_raise_exn as noreturn - regenerate schemex

pmatos:
scheme_raise_exn raises an exception and doesn't return.
Static analysis tools find a huge amount of problems with regards
to memory leaks that are actually false positives because the tools
are not aware the function does not return. Marking it as such aids
further inspection of real problems.

### DIFF
--- a/racket/src/racket/include/scheme.h
+++ b/racket/src/racket/include/scheme.h
@@ -252,6 +252,15 @@ extern "C"
 {
 #endif
 
+/* The reason we need two preprocessor variables is that, in gcc/clang the 
+   function attribute comes after the function declaration. However,
+   in MSVC the function attribute comes before the function declaration. */
+#ifdef __GNUC__
+#define NORETURN __attribute__((__noreturn__))
+#else
+#define NORETURN
+#endif
+
 /* Allowed by all configurations, currently: */
 #define MZ_CAN_ACCESS_THREAD_LOCAL_DIRECTLY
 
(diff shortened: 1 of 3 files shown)


mflatt:
We'll need something more portable here, probably a macro that expands to `__attribute__((noreturn))` only when `__GNUC__`.

pmatos:
Can I create these defines in `schemef.h` or is there a better place to do so?

mflatt:
I think "include/scheme.h" is probably the right place.

pmatos:
I have now added a patch so it's valid only for `GNUC` implementations. It's not perfect because I wanted to add it as well to MSVC but I don't have one installed for testing and things got quickly very ugly because for MSVC, the attribute comes before the declaration as in:
```
__declspec(noreturn) void ...(...);
```
which meant the macro NORETURN would look like:
```
#define NORETURN(decl) __declspec(noreturn) decl
```
for MSVC and similarly but suffixed for GNUC. Problem is, when I do this `make exports`, stops exporting `scheme_raise_exn`. Since at the moment it's not really blocking static analysis having the `noreturn` on Windows I implemented `NORETURN` for GNUC only.

pmatos:
@mflatt let me know if there's anything missing to get this merged. I ran the static analysis on this branch and results have improved with less false positives. :) 288 issues to go!

mflatt:
Sorry that I mangled the merge. I ended up adding an `exit(0)` to the end of `scheme_raise_exn` to make the compiler that I'm using happy.

pmatos:
@mflatt sure. Which compiler are you using?

PR #2478, opened 2019-02-14
Add fxvector->cpointer.

samth:
Needs docs/tests still.


mflatt:
An `fxvector` has a different representation strategy than an `flvector`. The change here would expose a pointer whose elements are not `int`s that fit in the fixnum range, but Scheme encodings of those ints, which will vary by platform and backend (CS vs. BC). That is, the value in each `intptr_t`-sized set of bytes in the referenced memory would need to be arithmetic-shifted right by 1 to 3 bits to get an integer value, depending. Is that what you expected @Metaxal?

Metaxal:
To be clearer, perhaps, I don't particularly care of the internal representation as bytes in memory, as long as I have access to converters for `ptr-ref` and `ptr-set!`, and that I can use `memcpy` directly on the memory segments.

(Though if zero'ed bytes don't correspond to the fixnum 0, that's a little annoying as I can't use `memset` to zero out a vector.)

It's not the end of the world for me if I can't get these, I'm pretty sure I can find a workaround if needed.

mflatt:
A `_fixnum` still corresponds to `intptr_t` on the foreign side, but `_racket` would perform the conversion. If that's useful, then ok. But I had imagined the goal was to get a pointer who content can make sense on the foreign side.

Metaxal:
> But I had imagined the goal was to get a pointer who content can make sense on the foreign side.

I don't have another side. I'm allocating shared memory via `shm` and `mmap`, so as to share `flvectors` and `fxvectors` between multiple racket instances. AFAIU, I can't convert the allocated memory directly to `flvector` and `fxvector` unfortunately, so I've been rewriting `shm-flvector` and `shm-fxvector` interfaces — though perhaps I took the wrong approach?

Metaxal:
I've given this more thought and I understand why you hesitate to expose `fxvector->cpointer`. I can definitely work around this, so don't waste your time on this. Thanks for looking into it in any case.

PR #2480, opened 2019-02-15
Avoid memcpy from null pointer in do_inotify_add

### LINKED ISSUE #2479: do_notify_add memcpys from NULL
I was tracking a static analysis issue in `rktio_fs_change` but I am not sure about the correct fix so I am opening this up for discussion.

The problem is in:
https://github.com/racket/racket/blob/edeae791ab736be7bb23239188382c8e11379d85/racket/src/rktio/rktio_fs_change.c#L454

So, it's possible to reach here with `s->wds` as 0. If we start with `rktio_fs_change`, then at `do_inotify_init` if `!s` we allocate memory for `s` and set it to 0 using `memset`. Then we can `rin_inotify_start` and return but never set `s->wds`. So we return from `do_inotify_init` with `s->wds` set to 0. 

We go back to the `rktio_fs_change` and the next thing we do is call `do_inotify_add`. We just memset `s` to 0 so we take the branch `s->count == s->size` and the `new_size` is set to 32. but then we call `memcpy` with a potentially null `memcpy`. 

I will need to take another look before proposing a PR.

pmatos:
Fixes #2479

### DIFF
--- a/racket/src/rktio/rktio_fs_change.c
+++ b/racket/src/rktio/rktio_fs_change.c
@@ -450,11 +450,17 @@ static int do_inotify_add(rktio_t *rktio, const char *filename)
     int new_size = (s->size ? (2 * s->size) : 32);
     rin_wd_t *new_wds;
     int i;
+
     new_wds = (rin_wd_t *)malloc(sizeof(rin_wd_t) * new_size);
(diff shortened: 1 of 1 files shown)


pmatos:
It looks to me that it's enough to include the `memcpy` in the `if(s->wds)` because the code to initialize the `s->wds` is coming afterwards anyway.

PR #2483, opened 2019-02-15
Create "for/append-list" macro

iitalics:
If `for/list` is analogous to `map`, then this form is analogous to `append-map`.

```racket
  (check-equal? (for/append-list ([x (in-range 4)])
                  (list x (number->string x)))
                '(0 "0" 1 "1" 2 "2" 3 "3"))
```

I have found the need for this function a few times lately, especially since most of my code avoids `map` in favor of `for/list`.

Defines the following macros:
* `(for/append-list (clause ...) body-or-break ...)`
* `(for*/append-list (clause ...) body-or-break ...)`
* `(for/append-lists (id ...) (clause ...) body-or-break ...)`
* `(for*/append-lists (id ...) (clause ...) body-or-break ...)`

--- 

1. We can bikeshed the name, not sure if `append-list` is clear enough.
2. I am providing this from `racket/list`, but maybe it should just be in its own module.
3. I have some tests written but I'm not entirely sure where I should put them.

### DIFF
--- a/racket/collects/racket/list.rkt
+++ b/racket/collects/racket/list.rkt
@@ -59,10 +59,17 @@
          group-by
          cartesian-product
          remf
-         remf*)
+         remf*
+
+         ;; for loops
+         for/append-list
+         for/append-lists
+         for*/append-list
+         for*/append-lists)
 
 (require (for-syntax racket/base)
-         racket/private/list-predicates)
+         racket/private/list-predicates
+         racket/private/for-append-list)
 
 (define (first x)
   (if (and (pair? x) (list? x))
--- a/racket/collects/racket/private/for-append-list.rkt
+++ b/racket/collects/racket/private/for-append-list.rkt
@@ -1,0 +1,60 @@
+#lang racket/base
+(provide for/append-list
+         for/append-lists
+         for*/append-list
+         for*/append-lists)
+
+(require
+ (for-syntax racket/base
+             syntax/for-body))
+
+;; ---------------------------------------------------------------------------------------------------
+
+(define-syntaxes [for/append-lists for/append-list for*/append-lists for*/append-list]
+  (let ()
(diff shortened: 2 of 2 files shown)


spdegabrielle:
@iitalics This is good.  Would you consider making a package to make this available?

soegaard:
Is
```
(for/append-list ([x (in-range 4)])
                  (list x (number->string x)))
```
different from
```
(append* (for/list ([x (in-range 4)])
                         (list x (number->string x))))
```
?

sorawee:
Ideally, the former should be slightly more efficient, as it doesn't need to hold on to the intermediate sublists. 

But perhaps the following design is more generalizable to other kinds of `for`.

```
(for/list* ([x (in-range 4)])
  (values x (number->string x)))

(for/vector* ([x (in-range 4)])
  (values x (number->string x)))
```

results in:

```
'(0 "0" 1 "1" 2 "2" 3 "3")
#(0 "0" 1 "1" 2 "2" 3 "3")
```

And if you have a function that produces `list` already, you can `(apply values ...)` to it to get multiple values.

However, this conflicts with `for/hash` which uses `values` for returning key/value pair already. I don't know how `for/hash*` would look like.

sorawee:
I guess one possible design for `for/hash*` would be 

```
(for/hash* ([x (in-range 4)])
  (values x (number->string x) (- x) (number->string (- x))))
```

which is similar to the `hash` function which requires even number of arguments.

pmatos:
@iitalics do you want to push this forward?

iitalics:
Sure, I can clean this up and write docs/tests later today. I'm going to keep the API as I described originally unless there are any further objections.

sorawee:
See also an implementation of my proposals above at #4096.

iitalics:
im not currently in a position to "push this forward" any more, but i think something like this would be a neat addition to the stdlib if anyone else is interested in taking it up

PR #2489, opened 2019-02-18
Simplify conditionals after removing dead store of has_space

pmatos:
The conditional simplification looks good to me. The biggest issue
here was to understand if when `pipe_quote` is true, we can and should
go to the else clause. Actually the more I look at it the more I think
this uncovers and earlier bug where if pipe_quote is true, result and
total_length are left at NULL and 0 respectively after the block.

### DIFF
--- a/racket/src/racket/src/symbol.c
+++ b/racket/src/racket/src/symbol.c
@@ -675,9 +675,7 @@ const char *scheme_symbol_name_and_size(Scheme_Object *sym, uintptr_t *length, i
 	&& (SCHEME_TRUEP(scheme_read_number(cs, clen, 0, 0, 1, 10, 0, NULL, &dz, 1))
 	    || dz)) {
       /* Need quoting: */
-      if (pipe_quote)
-	has_space = 1; /* Use normal */
-      else {
+      if (!pipe_quote) {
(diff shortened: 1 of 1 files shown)


mflatt:
I'm not convinced that this change is correct. Does it work for `(string->symbol "1")`, which should print using vertical bars?

pmatos:
I dropped the ball on this one and will have to look at it again.

pmatos:
I just rebased this to trigger the CI tests.

pmatos:
OK, so I pushed a fix. This was a very interesting error. My initial assumption was that the initial code that looked like:
```
if (C1) {
  if (C2)
    nop
  else
    P1
}
else 
  P2
```

could be written as
```
if(C1 && !C2)
  P1
else 
  P2
```

That was my mistake. Of course, once you pointed out that there was a problem, I noticed this wouldn't work for `C1 && C2`. The solution was to change this to
```
if (C1)
  if (!C2)
    P1
else 
  P2
```

I could have alternatively done
```
if (!C1)
  P2
else if (C2)
  P1
```

but I thought that `C1` would have become less clear than it is now. 

@mflatt what do you think?

pmatos:
@mflatt ping

PR #2496, opened 2019-02-19
Mark other error functions as NORETURN

pmatos:


pmatos:
This PR cuts about 48 false positives from the static analyser. 220 errors to go.

mflatt:
The `NORETURN` annotations cause trouble with my compiler, because it can't verify that the annotation is correct. My attempt to avoid that problem in f14d7e06bba5886a0e41a801fa27a8faecde3e43 only worked for the CGC part of the build, and the code generated for 3m still confuses the compiler.

Could we make this annotation be a no-op unless is is specifically requested through a compiler flag like `-DMZ_DECLARE_NORETURN`?

pmatos:
Sorry for the delay, a cold caught me by surprise and I was off a couple of days. Have a bit to catch up to. 

I agree with you that we should streamline this `NORETURN` thing. Still, one thing is killing me: curiosity. Which compiler are you using?

Give me a few days to look into this and I will get back to you with a better PR.

samth:
This is also an issue for the 3m-generated code on drdr, as you can see here: http://drdr.racket-lang.org/51622/pkg-src/build/make#output1

That is unfortunately a very old gcc, 4.8.4 (that machine runs Ubuntu 14.0.4). If it would help, we could potentially upgrade the compiler, but @jeapostrophe would know more.

mflatt:
I use the default compiler on Mac OS, which is Clang.

@pmatos  If the `-DMZ_DECLARE_NORETURN` approach is ok with you, I can add that today so that we fix the test failures.

pmatos:
Apologies for the failures. Since it's urgent, let me work on it now. Give me a few minutes.

pmatos:
I have added two commits here. Haven't tested them yet.

pmatos:
The first one adds the `MZ_DECLARE_NORETURN` as suggested.

The second one removes the `exit(0)` you added previously as a workaround but it's really ugly so I removed it, if you don't mind. :)

mflatt:
This is exactly what I would have done - thanks!

PR #2506, opened 2019-02-26
In `case', report binding info of `else' in error

shhyou:
This pull request tries to report what `case` sees when `else` is not recognized. An example is

```racket
#lang racket/base
(match 5
  [else (case 7 [else #f])])
#|
cas.rkt:6:14: case: bad syntax (not a datum sequence)
  at: else
  in: (case 7 (else 8))
  location...:
   cas.rkt:6:14
|#
```

The proposed new error messages are as follows. The error messages correspond to the situations where

- `else` is accidentally bound as a local variable
- `else` is shadowed by some top (module) level binding
- `else` is unbound

I have some questions:

- Should `case` report errors like this?
- Is there any preferred error message style? Currently I use the same style as `raise-argument-error`.

```racket
#lang racket/base
(module M1 racket/base
  (require racket/match)
  (match 5
    [else
     (case 7 [else 8])]))
#|
cas.rkt:6:14: case: bad syntax (not a datum sequence)
  expected: a datum sequence or the binding 'else' from racket/base
  given: a locally bound identifier
  at: else
  in: (case 7 (else 8))
  location...:
   cas.rkt:6:14
|#
(module M2 racket/base
  (define else 'no)
  (case 9 [else #t]))
#|
cas.rkt:18:11: case: bad syntax (not a datum sequence)
  expected: a datum sequence or the binding 'else' from racket/base
  given: an identifier bound by the current module
  at: else
  in: (case 9 (else #t))
  location...:
   cas.rkt:18:11
|#
(module M3 racket/base
  (provide else)
  (define else 'here))
(module M4 racket/base
  (require (submod ".." M3))
  (case 9 [else #t]))
#|
cas.rkt:56:12: case: bad syntax (not a datum sequence)
  expected: a datum sequence or the binding 'else' from racket/base
  given: an unbound identifier
  at: i-am-an-identifier
  in: (case 11 (i-am-an-identifier (quote no)))
  location...:
   cas.rkt:56:12
|#
```

### DIFF
--- a/racket/collects/racket/private/case.rkt
+++ b/racket/collects/racket/private/case.rkt
@@ -67,7 +67,28 @@
                [(bad . _)
                 (raise-syntax-error 
                  #f
-                 "bad syntax (not a datum sequence)"
+                 ;; If #'bad is an identifier, report its binding in the error message.
+                 ;; This helps resolving the syntax error when `else' is shadowed somewhere
+                 (if (not (symbol? (syntax-e (syntax bad))))
+                     "bad syntax (not a datum sequence)"
+                     (string-append
+                      "bad syntax (not a datum sequence)\n"
+                      "  expected: a datum sequence or the binding 'else' from racket/base\n"
+                      "  given: "
+                      (let ([binding (identifier-binding (syntax bad))])
+                        (cond
+                          [(not binding) "an unbound identifier"]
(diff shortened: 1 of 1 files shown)


mflatt:
LGTM

PR #2507, opened 2019-02-28
Use cached llvm to run scan-build on racket

pmatos:
This change ensures that the scan-build jobs run on a different
machine dedicated to it and that the llvm build is cached so it is not
rebuilt each time it is needed.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -1,23 +1,98 @@
-image: ubuntu:18.04
+stages:
+  - prepare
+  - test
 
+# Using debian:stable-slim to perform an llvm build with Z3 support
+# The reason this is necessary is that we need Z3 support in scan-build
+# to be able to crosscheck errors with z3 which decreases the false positives
+# rate by around 5%.
+#
+# Also, this should only be performed once. Once in cache llvm won't be build anymore
+# until we force it.
+prepare-cache:
+  image: debian:stable-slim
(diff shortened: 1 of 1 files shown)


pmatos:
@samth Thank you for the quick merge. I probably should have written more about this change but when I noticed I should do so I was called by my kids who didn't take 'Wait' for an answer.

In any case, previously we were using the gitlab shared machines to do the static analysis. This had a problem, which was that I wanted to use the results of the research here:     https://dblp.org/rec/journals/corr/abs-1810-12041
Yet, no packaged LLVM, for several reasons, was built with Z3 in order to be able to do the refutation mentioned in the paper. So the solution was to build an LLVM and cache it. That's what the new job does. It first builds LLVM, caches it and then uses it to statically analyse normal racket and cs. This cache works over several jobs which means that LLVM is only built once and reused until we clear the cache and force LLVM to be rebuilt. 

Unfortunately gitlab shared runners are limited to 2 cpu, 2 hours per job which is not enough to build LLVM, so I added a dedicated runner (one of my personal machines) to doing the racket static analysis. This has 4 core and no timeout (it's set to 4 hours just in case, but in principle I can extend it as much as needed). The current system only does one job at a time with `-j 5`. If I see this needs increase I can add a dedicated runner with 16cpus. We'll see how it goes for now.

I should probably write a blog post about all of this... :)

PR #2513, opened 2019-03-05
net/url: Consult current-alist-separator-mode parameter when parsing strings and URLs

erkin:
[Docs suggest](https://docs.racket-lang.org/net/url.html) that `net/url` procedures should take `current-alist-separator-mode` into account.

This is crucial for parsing gopher links, where `;` stands for the video type.
(eg `gopher://example.org/;/foobar.ogv`)


jeapostrophe:
Can you add a few test cases? The right place is here, I believe: https://github.com/racket/racket/blob/master/pkgs/net-test/tests/net/url-string.rkt

mflatt:
The `current-alist-separator-mode` parameter is used for the *query* part of URL parsing, not the *path parameter* part.

Changing the URL parser to make the path parameter part sensitive to `current-alist-separator-mode` is probably not a good idea from the perspective of backward compatibility. It would make sense to add a new (Racket) parameter to control path parameter parsing, though—and that addition would have a side effect of improving documentation, since the treatment of `;` within a path element does not appear to be part of the standard at the level of URIs.

samth:
I believe that the [URL Standard](https://url.spec.whatwg.org/) which is the correct standard to reference here (it's what browsers/web servers use) doesn't treat `;` specially in paths. I tried the example in Chrome, Firefox, node.js, and Python and they all parsed the path as either "/;/foobar.ogv" or  "//;/foobar.ogv".

jackfirth:
@samth The docs currently state that `net/url` is meant to work with RFC 2396, which conflicts with WhatWG's spec. They're not compatible with each other, so `net/url` should probably avoid referencing both and clearly state which one it's trying to implement.

PR #2515, opened 2019-03-05
Make `sql-null` a transparent struct.

samth:
This allows it to cooperate better with Typed Racket, particularly regarding
the `Any` type. The guard and use of `#:authentic` also check that it's still
a singleton in all cases.

### DIFF
--- a/racket/collects/db/private/generic/sql-data.rkt
+++ b/racket/collects/db/private/generic/sql-data.rkt
@@ -11,15 +11,23 @@
 ;; NULL
 
 (define-values (sql-null sql-null?)
-  (let ()
+  (let ([created? #false])
     (struct sql-null ()
-            ;; must deserialize to singleton, so can't just use serializable-struct
-            #:property prop:serializable
-            (make-serialize-info (lambda _ '#())
-                                 #'deserialize-info:sql-null-v0
-                                 #f
-                                 (or (current-load-relative-directory)
-                                     (current-directory))))
+      #:transparent
+      #:authentic
+      #:guard (lambda (n)
+                (when created?
+                  (error 'sql-null "cannot create new instances of sql-null"))
+                (set! created? #true)
+                (values))
+      #:property prop:custom-write (lambda (v p w?) (write-string "#<sql-null>" p))
+      ;; must deserialize to singleton, so can't just use serializable-struct
+      #:property prop:serializable
+      (make-serialize-info (lambda _ '#())
+                           #'deserialize-info:sql-null-v0
+                           #f
+                           (or (current-load-relative-directory)
+                               (current-directory))))
     (values (sql-null) sql-null?)))
 
 (define (sql-null->false x)


lexi-lambda:
It’d be neat if this pattern were put in a module in `base` somewhere (as `make-singleton` or `define-singleton`, for example)—it seems non-trivial to get right. I don’t know where it would belong, though.

samth:
It's not obvious that this level of effort is generally useful (or even needed here).

lexi-lambda:
Sure, but it can’t really hurt, can it? And it seems like it would smooth TR interop if libraries used such a construct, which they’re pretty unlikely to do if it requires even slightly more effort than `(let () (struct s ()) (s))`.

lexi-lambda:
Let me put this another way: I’ll do the work to implement such a construct if someone tells me where they think it ought to go. (`racket/singleton`? `data/singleton`? `racket/base`?)

rmculpepper:
@samth This seems fine, but please put the `created?` variable in the currently-empty `let`. (Otherwise I think someone can use `module->namespace` and reset `created?` to make more instances.)

@lexi-lambda Can you think of two other places such an abstraction could be used in the current code base?

lexi-lambda:
Sure. All the places an opaque singleton is used seem like ripe locations to use such an abstraction:

  - The contract system does this a lot. `the-unsupplied-arg` and `contract-random-generate-fail` are the two examples that are exported, but `name-default` from `racket/contract/private/guts` also qualifies.

  - `com-omit` from `ffi/com` is defined this way.

  - `done` and `start` singletons are defined like this in `drracket/private/module-language-tools`.

It also shows up in some libraries. For example, in my own `data/maybe` library, `nothing` is a singleton value, and in my `namespaced-transformer` library, `supports-namespaced?` is a singleton value used as a language info key.

I’m sure there are other examples, but those are the ones that I found with a very naïve grep on my local machine. I’m sure some patterns weren’t caught by that search, and I also don’t intend to audit the whole package server.

jackfirth:
Naive question here, but why is it critical here that only one instance ever be created, rather than the slightly-weaker property that all created instances are `equal?` to each other and instances can be substituted freely without breaking programs? Are there places where the `eq?`-ness of these singleton values actually matters to client code? I ask because I've made singleton-like APIs before and frequently wondered how critical it is that I ensure `(and (foo-singleton? x) (foo-singleton? y))` implies `(eq? x y)`.

lexi-lambda:
I think it’s more a convention than anything else, and because there’s not really any reason *not* to provide the guarantee of `eq?`-ness. The whole point of a singleton value is that there’s only one, so you might as well use a single representative. I doubt it actually matters for efficiency most of the places it’s used, but it doesn’t hurt, either, and it avoids confusion.

(At this point, it’s so idiomatic to do this that I wouldn’t be terribly surprised if someone were tripped up by an API not guaranteeing `eq?`-ness of singletons and using `eq?` anyway out of habit. Maybe it’s mildly unfortunate that general use of `eq?` is as pervasive as it is, but that ship sailed a long time ago.)

jackfirth:
I find it more complex to guarantee `eq?` than to guarantee `equal?`, and it would be simpler for me as a library author if clients were fine with `equal?`. So I'd like to know the actual use cases for `eq?`. If it's just a conventional habit that some people used to other lisps and schemes expect, that isn't a compelling enough use case for me.

hashimmm:
I generally try to think about values being `equal?`, and not `eq?`. Usually the way I look at it is that using `eq?` is an optimization.

As an example, I recently made some functions that work on graphs of certain values. They're all checked for "equal", when doing things like merging graphs. And these values are recursive structures so checking for "equal?" is expensive. But the idea going forward is to structure the library/usage such that I'll generally just need to check for "eq?".

The documentation, too, says at some point that most libraries should be made in a way that the user need not care, and is free to do things like make impersonators to facilitate testing and whatnot.

But perhaps it is an important and common optimization. In any case, while I'm unsure of whether to encourage or discourage such patterns, what _is_ important to me is that it all plays nicely with the rest of the language. For example, this pull request is about changing sql-null so that it plays nicer with Typed Racket, which directly helps me, a user, out.

lexi-lambda:
> If it's just a conventional habit that some people used to other lisps and schemes expect, that isn't a compelling enough use case for me.

Yes, I think this is probably the main practical reason that things like `sql-null` go to an effort to preserve `eq?`, though I think there’s also a philosophical/aesthetic niceness to making a singleton truly a single value in memory. :)

rmculpepper:
@lexi-lambda That's more than I expected. Maybe?

PR #2521, opened 2019-03-07
Correct path, and add link to path

### LINKED ISSUE #2520: Path no longer up to date, and should be link
In "Units of Code" in the Racket Style guide, there is the following line:

https://github.com/racket/racket/blob/5bb837661c12a9752c6a99f952c0e1b267645b33/pkgs/racket-doc/scribblings/style/unit.scrbl#L326

It claims that in "collects/setup/scribble" all functions use latex-dest. I couldn't find this file in collects/setup/, there was no scribble file. Even if there is, it would be better to turn it into a link to the actual file. I searched for latex-dest, and maybe the file it referred to is "racket/pkgs/racket-index/setup/scribble.rkt", although I really don't understand the point being made about latex-dest, since nothing named latex-dest is exported. However that's not the point of this issue.

Edit: Let me know if I should *not* add issues for these kinds of things on GitHub, but first check on an email list or what not (I did spend 5-10 minutes searching around).

MarcKaufmann:
Fixes #2520.

### DIFF
--- a/pkgs/racket-doc/scribblings/style/unit.scrbl
+++ b/pkgs/racket-doc/scribblings/style/unit.scrbl
@@ -323,8 +323,9 @@ Pick a rule for consistently naming and ordering the parameters of your
 Finally pick the same name for all function/method arguments in a module
  that refer to the same kind of data---regardless of whether the module
  implements a common data structure. For example, in
- @filepath{collects/setup/scribble}, all functions use @racket[latex-dest]
- to refer to the same kind of data, even those that are not exported.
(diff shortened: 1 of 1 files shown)


MarcKaufmann:
I am not sure if I should have changed the white space manually or if it is done manually.

samth:
I think that just leaves the path as plain text; you probably want `@hyperlink[...]{@filepath{...}}`.

samth:
As for white space, it is not done automatically -- if you want shorter lines you have to do it yourself.

MarcKaufmann:
OK, I *think* I updated the PR correctly, let me know if that isn't the case.

PR #2522, opened 2019-03-07
Update broken URL

### LINKED ISSUE #2518: Broken link in docs
The following link in the documentation seems broken and I couldn't figure out where the tests actually are for 2htdp:

https://github.com/racket/racket/blob/5bb837661c12a9752c6a99f952c0e1b267645b33/pkgs/racket-doc/scribblings/style/testing.scrbl#L19

I am not sure that the docs are wrong on claiming that most test suites still live in /collects/tests/. That seems to be generally wrong, no?

MarcKaufmann:
Fixes #2518


MarcKaufmann:
Not sure whether it is better practice to include trailing slashes or not for urls to folders, since both work.

PR #2524, opened 2019-03-10
add racket/whereis, raco whereis command

rmculpepper:
This adds a `racket/whereis` module and a `raco whereis` command for finding local filesystem paths corresponding to Racket modules, collections, packages, raco commands, and bindings.

I wanted to add support for finding the file where a feature is documented, too, but I don't think it's possible with the current information maintained by the doc system. See the comments in `racket/whereis`.

This PR also extracts `collection-paths` from the raco test command to the new module `setup/collection-paths`.

Questions:
- Should this go in the base package?
- If so, should `racket/whereis` also be included in the `racket` language? (If not, should the module go in `setup/` instead?)
- Are `whereis-system` and `raco whereis -s` useful or should they be dropped?


samth:
Can you compare this to the raco-find-collection package by Asumu?

rmculpepper:
@samth `raco whereis` requires you to tell it what kind of thing (module/collection/pkg/etc) you're looking for, and when there are multiple paths (eg, for a collection) it shows all of them. IIUC, `raco fc` tries multiple interpretations of the given name and selects one path according to some heuristic. (Or it lets you pick one interactively.)

samth:
As someone who uses `raco fc` many times every day, I think its behavior is the right one. But more importantly, it seems like there's a lot of code that could be shared between these two packages, and that probably only one package is needed.

cc @takikawa

jackfirth:
I'm really interested in the programmatic API to this functionality, which `raco fc` does not currently provide.

samth:
Sorry if I came off as saying that this is a bad idea -- it's a good idea! I just hope we can avoid duplication.

rmculpepper:
@samth I think `raco fc` serves a different purpose from `raco whereis`, so I disagree that the behavior of `raco fc` is "the right one" in this context. Here's one example: Can you use `raco fc` to locate the `db` module? It took me three tries, and my first two resulted in a wrong answer and an error. I'm not trying to deny that `raco fc` is useful, but it has a narrower purpose (and it's optimized for that purpose).

One use I have in mind for `raco whereis` and `racket/whereis` is to help new Racket programmers understand where the different parts of Racket live in the filesystem. That's why I would prefer to make it part of the minimal distro and ideally part of `racket` language. I think it would also be helpful to update the docs where concepts like module paths, collections, and packages are introduced to tell them how to inspect the state of the Racket-feature-to-filesystem mapping. But I don't have a strategy or timeline for that.

On `racket/whereis`: this module is mostly a thin wrapper around existing code that does the interesting work. (The point of this PR is to aggregate existing functionality.) The `raco fc` command uses much of the same code, with a few exceptions. So there's not much more implementation to be shared, and the two commands are different enough that trying to unify them seems like a bad idea to me.

rmculpepper:
I've turned this into a package instead.

PR #2525, opened 2019-03-12
doc: new names for arguments to 'error'

bennn:
rename the arguments to error in the >2 arguments case, from:

```
  (error src format v ...)
```

to

```
  (error sym format-msg v ...)
```

because:
- `sym` is a symbol, not a source location
- the docs use the racket/base meaning of `format` to explain the 2nd argument

- - -

This change may confuse people because the "sym" in `(error sym)` is different from the one in `(error sym msg v ...)`, but I think people are more likely to be confused that `src` has nothing to do with srclocs.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/exns.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/exns.scrbl
@@ -106,9 +106,9 @@ exception handler obtains control, and the handler itself is
 (eval:error (raise 'failed #t))
 ]}
 
-@defproc*[([(error [sym symbol?]) any]
-           [(error [msg string?] [v any/c] ...) any]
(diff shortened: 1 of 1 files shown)


jackfirth:
Would `name` be a better alternative to `sym` and `src`?

mflatt:
How about `message-sym`, `who-sym`, `format-str`, and `message-str`?

(I could see an argument that "sym" and "str" are redundant given the contracts, but a lot of the other documentation uses the "sym" and "str" convention.)

bennn:
Ok

(I think I used those names in the intended places)

I like keeping `-sym` and `-str` on the end, but if other opinions appear I'm fine with anything that's not `src` and `format`.

mflatt:
LGTM

PR #2527, opened 2019-03-12
Add example for ellipsis use in -> contract

pmatos:
I don't find it particularly obvious how one can use the `...` in the `->` contract combinator. It seems like the reference could do with one example. I am not sure I have phrased this properly and I am happy to get some suggestions or improvement in wording or example. This was discussed on Slack between myself @samth, @jackfirth and @lexi-lambda . Maybe they want to chip in.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1184,10 +1184,32 @@ is a contract on a function that accepts a by-position argument that
 is an integer and a @racket[#:x] argument that is a boolean.
 
 As an example that uses an @racket[...], this contract:
-@racketblock[(integer? string? ... integer? . -> . any)]
+@racketblock[(integer? string? ... integer? . -> . boolean?)]
 on a function insists that the first and last arguments to
 the function must be integers (and there must be at least
-two arguments) and any other arguments must be strings.
+two arguments) and any other arguments must be strings. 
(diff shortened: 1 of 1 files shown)


rfindler:
I don't think the "particularly non intuitive" comment is helpful.

pmatos:
You're right, personal feeling have no place in the reference manual.

pmatos:
Also, let me put it out there: this PR is for the reference manual but I am unsure it's the right place for it. No other complete code examples exist on this chapter and this might make it look out of place. If anyone thinks this is better in the guide, I will close this and open a PR in the guide.

rfindler:
I think that improving the example so that it has running code (and maybe a violation as well as a call that succeeds) is a good idea, but perhaps it would be better to use `@examples`?

pmatos:
@rfindler sure, I can add that as well. You agree that this can stay in the reference or should be moved to the guide?

rfindler:
I think that a series of reference-like examples illustrating how to use the construct belongs in the guide. (It should probably start with a simpler one, tho.)

If you want, I can do it.

samth:
I think it's valuable to have examples in both the reference and the guide. The reference is what you get when you look up `->`, after all.

pmatos:
I won't be able to get to it today but if you have time, go ahead. Feel free to close this PR without merging as well in that case.

rfindler:
I've pushed some examples. Thanks!

PR #2529, opened 2019-03-13
fix docs for syntax-binding-set->syntax

yjqww6:


gus-massa:
Thanks! Merged in https://github.com/racket/racket/commit/01222bb93cafdc145a0354030a44e7a33cd609c2

PR #2537, opened 2019-03-15
doc: add an example for parsing single keyword arguments

euhmeuh:
As pointed out by this [StackOverflow question](https://stackoverflow.com/questions/55171895/how-do-i-check-for-the-presence-of-a-keyword-in-a-racket-syntax), there's no example showing how to match a keyword if it's not followed by a value.

This small example shows how to do so.

I wasn't expecting a small difficulty though: in SO question's case, we want to generate code whether the keyword is present or not, using `(if (attribute keyword?) #'(template when present) #'(template when absent))`.

But in the doc, the example needs either to generate an `else` clause, or nothing.
So I resolved to using `~@` over an empty syntax list. If you have a better way to do it, feel free to improve the PR and add commits :)


pkgs/racket-doc/syntax/scribblings/parse/ex-kw-args.scrbl
```diff
@@ -99,6 +99,27 @@ If the second subtemplate of a @racket[~?] template is
 @racket[(~@)]---that is, it produces no terms at all---the second
 subtemplate can be omitted.
 
+@section{Optional Keyword Argument as flags}
```
AlexKnauth: Make sure plurals are consistent. Either plural as plural `Optional Keyword Arguments as flags` or singular as singular `Using an Optional Keyword Argument as a flag`, but not singular as plural.

euhmeuh: Good catch!

sorawee:
Wouldn't `~bind` works better here?

rmculpepper:
Thanks! This looks good to me. For the record, here are two ways to get `~?` to handle the conditional:

    ;; Approach 1: use the maybe-keyword attr to compute a maybe-final-clause attr
    (define-syntax (mycond* stx)
      (syntax-parse stx
        [(mycond (~optional (~and error? #:error-on-fallthrough))
                 clause ...)
         #:attr fallthrough (and (attribute error?)
                                 #'[else (error 'failure "no clause matched")])
         #'(cond clause ... (~? fallthrough))]))

    ;; Approach 2: include the maybe-keyword attr in a manner that the compiler will optimize away
    (define-syntax (mycond** stx)
      (syntax-parse stx
        [(mycond (~optional (~and error? #:error-on-fallthrough))
                 clause ...)
         #'(cond clause ...
                 (~? [else (begin 'error? (error 'failure "no clause matched"))]))]))

OTOH, I think the PR's current solution is better for the docs, since it generalizes better.

AlexKnauth:
Would it be a better example for the docs if the way you used it in the template was uniform? By that I mean that you don't have to "omit" a piece the template when the flag doesn't exist. One example of this would be the `#:transparent` flag for structs, where when it exists the inspector is `#false` and when it doesn't exist the inspector is `(current-inspector)`.

Something like that might be simpler because the "inspector" piece of the template always exists, so things like `(~@ . fallthrough)` or `(~? fallthrough)` wouldn't be necessary. I want to avoid the example needing to do too many things at once.

euhmeuh:
I'll try to make something better according to the comments :)

PR #2540, opened 2019-03-18
Return a procedure from make-variable-like-transformer

AlexKnauth:
So that make-variable-like-transformer produces a value that passes both `procedure?` and `set!-transformer?`.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/transformer.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/transformer.scrbl
@@ -8,20 +8,24 @@
 
 @defmodule[syntax/transformer]
 
-@defproc[(make-variable-like-transformer [reference-stx syntax?]
-                                         [setter-stx (or/c syntax? #f) #f])
-         set!-transformer?]{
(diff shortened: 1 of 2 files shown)


euhmeuh:
Related to https://github.com/racket/racket/pull/2539 ?

AlexKnauth:
If this is merged it will make #2539 completely unnecessary, and I will close it. I will only come back to that PR if people push back on this one for adding behavior to the result an existing function.

rmculpepper:
Thanks! This looks good to me, but it needs a doc update too, both for returning a procedure and accepting a procedure for the ref-stx argument.

AlexKnauth:
Should I also update the documentation to specify that the `setter-stx` can also be a procedure? That's currently undocumented.

rmculpepper:
Yes, please. I didn't realize they were out of sync wrt `setter-stx`.

AlexKnauth:
I've added documentation for both the `reference-stx` argument and `setter-stx` argument, and I've also added `(and/c set!-transformer? (-> syntax? syntax?))` as the output contract.

PR #2544, opened 2019-03-19
[WIP] core: enhance `raise-argument-error`

abmclin:
I've created an enhanced version of the primitive `raise-argument-error` with an optional keyword argument `#:more-info` which is intended to allow users to include more information about the bad argument's role in the error output. The extra information is typeset as `<continued-message>` according to the convention defined at https://docs.racket-lang.org/reference/exns.html?q=conventions#%28part._err-msg-conventions%29

As of now this commit causes problems for various packages using Typed Racket so this commit shouldn't be merged without discussion with the TR maintainers. The problem, as far as I was able to tell, seems to be that the new `raise-argument-error`'s function signature doesn't conform to the defined types for the original primitive one at https://github.com/racket/typed-racket/blob/ade0dff65ed89c435205c5bfc17fee7f0837083c/typed-racket-lib/typed-racket/base-env/base-env-indexing-abs.rkt#L331

Here's a motivating example with a function defined with two arguments having identical contracts.

```
; src-dir : path-string?
; dest-dir : path-string?
(define (copy-directory src-dir dest-dir)
     (unless (path-string? src-dir)
          (raise-argument-error 'copy-directory "path-string?" src-dir
                                             #:more-info "the source directory to be copied"))
          (unless (path-string? dest-dir)
          (raise-argument-error 'copy-directory "path-string?" dest-dir
                                             #:more-info "the destination path to copy the source directory to"))
     ... )
```

```
(copy-directory 'bad "/usr/home/destination")
copy-directory: contract violation;
 the source directory to be copied
  expected: path-string?
  given: 'bad
```

```
(copy-directory "/usr/home/source" 'bad)
copy-directory: contract violation;
 the destination path to copy the source directory to
  expected: path-string?
  given: 'bad
```
`#:more-info` used in `copy-directory` makes it easy to tell at an immediate glance which argument with the same contract is the bad one.

I thought of this enhancement when I was doing system programming with multiple arguments having identical contracts. It was a pain identifying which argument was the faulting one in the error output simply by reading the error message, especially if the bad input is being passed through a chain of callers.

This commit also includes a simple error reporting infrastructure used to construct error messages for various ways `raise-argument-error` can be used and also displays error messages about misuses of keyword arguments in a consistent manner similar to how bad positional arguments are reported.

Reference documentation is also updated with an explanation and example of using `#:more-info` and a history annotation is included as well.

Test cases included to verify backward compatibility and new functionality.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/exns.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/exns.scrbl
@@ -163,14 +163,16 @@ default @tech{error display handler} does not show a ``stack trace'' for
 for end users.}
 
 
-@defproc*[([(raise-argument-error [name symbol?] [expected string?] [v any/c]) any]
-           [(raise-argument-error [name symbol?] [expected string?] [bad-pos exact-nonnegative-integer?] [v any/c] ...) any])]{
+@defproc*[([(raise-argument-error [name symbol?] [expected string?] [v any/c] [#:more-info more-info string? #f]) any]
+           [(raise-argument-error [name symbol?] [expected string?] [bad-pos exact-nonnegative-integer?] [v any/c] ... [#:more-info more-info string? #f]) any])]{
 
 Creates an @racket[exn:fail:contract] value and @racket[raise]s it as
 an exception.  The @racket[name] argument is used as the source
 procedure's name in the error message. The @racket[expected] argument
 is used as a description of the expected contract (i.e., as a string,
-but the string is intended to contain a contract expression).
+but the string is intended to contain a contract expression). The optional
+@racket[more-info] argument is included in the exception's error message,
+its purpose is to provide more detail about the bad argument's role.
 
 In the first form, @racket[v] is the value received by the procedure
 that does not have the expected type.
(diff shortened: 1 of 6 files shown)


abmclin:
@mflatt if you would mind reviewing this PR. Any feedback would be welcome.

abmclin:
When figuring out the TR issue, I tried using `->key` defined in https://github.com/racket/typed-racket/blob/master/typed-racket-lib/typed-racket/types/base-abbrev.rkt but it appeared it wasn't equipped to cope with the needs of the relevant section in `base-env-indexing-abs.rkt` so I experimented with extending `->*` by adding additional `syntax-parse` clauses to handle keywords the same way `->key` does but the fix didn't resolve the original Type Checker exception so I'm requesting a TR maintainer to review as well.

@samth I believe you're a TR maintainer, would you be willing to review and see what can be done about the TR conflict with the new commit?

samth:
The fix to TR is to give it a type with a keyword argument. If you have trouble with the abstraction there, just move the type annotation to `base-env.rkt`. The abstraction is just legacy code that should be deleted anyway. If you open a PR on `racket/typed-racket`, even a not-yet-working one, I'm happy to look at it.

abmclin:
Ok, I'll give it a try and open another PR against `racket/typed-racket`

mflatt:
I like this improvement, but I have hesitated to merge due to the the amount of new code that duplicates existing functionality. Unfortunately, the existing implementation can't be just dropped, because it's needed at relatively low layers in stack.

Maybe it would make sense to adjust the low-level implementations with some ugly interface (such as allowing the second argument to be a pair or list of two strings), and then have a thin layer here that provides the keyword interface and converts to the ugly primitive interface.

abmclin:
Quite understandable and I agree it is concerning. This is one of those situations when it seemed like it should have been a relatively simple addition and turned out to be a can of worms in order to preserve backward compatibility.

If I’m recalling correctly, I found the C level implementation had several entangled code paths for various supported use cases which was difficult to detangle.

If you are in agreement that we can create an ugly C interface with more complex parameters and a thin wrapper at the Racket level for keywords, I think that should make it a little easier to rework the lower level implementation.

I’d be happy to provide the necessary implementation for both the BC and CS platforms.

Also after some thought I plan to change the name of the keyword from #:more-info to #:role

mflatt:
Yes, I'm happy with an ugly C/Rumble interface, so your plan sounds great.

abmclin:
Unfortunately, I have not had any time to give this PR its proper attention and I don't anticipate having the time at least in the near future so I am closing for now.

PR #2545, opened 2019-03-19
units: Avoid use-before-initialization errors in signature contracts

lexi-lambda:
This commit fixes an issue with the fix for contracted bindings in signatures implemented in commit 5fb75e9f82. While the previous fix worked in simple cases, it introduced a problem: although signatures that define contracted bindings were able to refer to other bindings in the signature in the binding contracts, anyone doing so was at the mercy of the exporting unit’s definition order. For example, given a signature

```racket
(define-signature a^
  [(contracted
    [ctc contract?]
    [val ctc])])
```

then a unit exporting the signature would cause a use-before-initialization error if its definition for `val` appeared above its definition for `ctc`.

This limitation did not exist in the units implementation prior to the introduction of the sets-of-scopes expander in Racket v6.3 (after which contracted bindings were broken until the aforementioned fix in Racket v7.2). However, the fact that they worked at all seems semi-accidental: instead of properly indirecting references to signature bindings within binding contracts, the contract expressions were simply placed in a context in which the existing names were bound. However, this meant that any export that renamed identifiers could cause problems, which the implementation strategy taken in this commit handles just fine.

---

As an aside: this commit maintains the existing semantics of `contracted` bindings, which places the contract expressions themselves *inside* the contract boundary they introduce. There’s perhaps an argument to be made that this is flawed, in the same way that `->d` is flawed relative to `->i`, since the signature itself can violate its own contracts, but the exporting unit will be blamed. Given the historical behavior, this seems unavoidable: any contract can refer to any signature binding, so there’s no way to compute any notion of order, since `contracted` doesn’t have the explicit dependency declarations of `->i`. Speaking ideally however, it seems to me like the way `contracted` works is a little bit fishy.


lexi-lambda:
I realized after opening this PR that the units system doesn’t seem to have much in the way of maintainers at the moment, so I’m not sure who if anyone can/should review this PR (especially given that the feature in question was broken from Racket 6.3 through 7.2 without anyone touching it). I think this change is pretty safe, so I will probably just merge it unless anyone speaks up.

rfindler:
Nothing seems obviously wrong to me. But just to be sure: the set of test cases only increased, right?

lexi-lambda:
Yes, the “deleted” test cases were just test cases I had added to the wrong test module in 5fb75e9f82d8d76330cef18a57edc46b4ac4fdfa. I just moved them to the right place in this change and added some more.

rfindler:
That's what I figured. Thanks.

PR #2553, opened 2019-03-24
file/untar: allow leading as well as trailing blanks

lassik:
Some really old tar files can have both leading and trailing blanks (nul/space characters) in the same octal number field in the header.

Here are examples of such tar files:

* https://groups.csail.mit.edu/mac/ftpdir/scheme-reports/r4rs.tar.gz
* https://groups.csail.mit.edu/mac/ftpdir/scheme-reports/r5rs.tar.gz

### DIFF
--- a/racket/collects/file/untar.rkt
+++ b/racket/collects/file/untar.rkt
@@ -212,6 +212,12 @@
     #f]
    [else
     ;; traditional:
+    (define skip-head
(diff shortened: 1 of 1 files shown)


mflatt:
Coincidentally, I ran into this just yesterday, too. You patch seems slightly nicer than mine (which I had not yet pushed).

Thanks!

PR #2554, opened 2019-03-25
wrap-expr/c: Add #:phase argument to control phase of introduced syntax

### LINKED ISSUE #2455: wrap-expr/c cannot be used to add contracts on expressions used in phases other than (syntax-local-phase-level)
It’s entirely possible for a macro to receive an expression that will eventually be evaluated at phase 1 or more, and it might be useful to protect that expression with a contract. However, simply using `wrap-expr/c` or `expr/c` naïvely does not work:

```racket
(define-syntax-parser print-compile-time-integer
  [(_ {~var e (expr/c #'exact-integer?)})
   #:do [(println (syntax-local-eval #'e.c))]
   #'(void)])

(print-compile-time-integer 42)
```
```
macro-arg/c: undefined;
 cannot reference an identifier before its definition
```

This makes sense—`wrap-expr/c` has no way of guessing where the expression will be used. However, it would be nice if `wrap-expr/c` provided an optional `#:phase` argument to control where the introduced bindings come from so that the phase can be explicitly specified.

lexi-lambda:
Closes #2455.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/contract.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/contract.scrbl
@@ -27,7 +27,8 @@ contracts to macro subexpressions.
                               (or/c identifier? symbol? string? #f) #f]
                       [#:macro macro-name
                                (or/c identifier? symbol? string? #f) #f]
-                      [#:context context (or/c syntax? #f) (current-syntax-context)])
+                      [#:context context (or/c syntax? #f) (current-syntax-context)]
+                      [#:phase phase exact-integer? (syntax-local-phase-level)])
          syntax?]{
 
 Returns a syntax object representing an expression that applies the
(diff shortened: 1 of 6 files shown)


rmculpepper:
I think the phase argument should be an absolute phase level, as it is for every other function I can think of that takes a phase, instead of an offset.

Please add tests with explicit `#:positive`, `#:negative`, etc arguments too. (That's where my implementation ran into trouble.)

lexi-lambda:
Good call on the argument to `#:phase`. I wasn’t sure when I was implementing this if it should be absolute or not, but you’re right that everything else usually takes an absolute phase level (or at least a phase level relative to the module currently being expanded). I’ve made that change.

For `#:positive` and `#:negative`, this implementation handles them just fine (the key is that the `runtime` submodule re-exports `racket/base`), but I’ve improved the test suite to exercise them. I’ve also improved the test suite to try lots more kinds of phase-shifting combinations, just to make sure everything goes right in all the possible edge cases.

lexi-lambda:
@rmculpepper Pinging you on this.

rmculpepper:
I pushed a few changes. If they look okay to you, then I think this is ready. Please update the version numbers in the `@history` blocks if necessary and squash to one commit when merging. Thanks!

lexi-lambda:
Thank you for your changes! They’re definitely better than what I had written.

PR #2565, opened 2019-03-27
improve accuracy of tanh function

bdeket:
using the implementation of https://www.math.utah.edu/~beebe/software/ieee/tanh.pdf

by changing from (/ (- 1 exp2z) (+ 1 exp2z)) to (- 1 (/ 2 (+ 1 exp2z)) the accuracy after rounding is increased (I was comparing with bftanh) and removes the fluctuations around z=18.35
using the polynomial for z ϵ(1.290e-8 to 0.549) seems to increase the accuracy after rounding even further
see comparison: http://pasterack.org/pastes/48436

especially the fact that  (< (tanh 18.36)(tanh 18.37)) ;=> #t was tripping me up
the two extra conditions (z . < . 1.29e-8) and (z . < . 0.549) are optional to solve this

my knowledge about accuracy and rounding issues etc. is rather basic, so please check carefully


mflatt:
Thanks! And thanks to @pmatos for also taking a look.

PR #2566, opened 2019-03-27
ffi-lib: log success, log failures with candidates tried

rmculpepper:


mflatt:
LGTM. I'm not sure that `ffi-lib` is the right topic, because using library names has tended to work better. But `ffi/unsafe` doesn't seem good in this case, and I don't have a better suggestion.

elfprince13:
@rmculpepper - Could you give an explicit example of using this feature? It appears to be exactly what I need, but I can't seem to observe any of the logged messages.
This example code:
```racket
#lang racket

(require racket/logging
         ffi/unsafe)

(let* ([my-log (open-output-string)]
       [lib (with-logging-to-port my-log
              (lambda ()
                (ffi-lib "my-library-name"
                         #:get-lib-dirs
                         (lambda ()
                           `(,"/my/search/path"))))
              'debug)])
   
  (values lib (get-output-string my-log)))
```
returns:
```racket
#<ffi-lib>
""
```

samth:
I put your code in a file, and ran this:

```
[samth@huor:~/work/teaching/c211 (master) plt] r -W debug@ffi-lib /tmp/x.rkt 
ffi-lib: failed for (ffi-lib "my-library-name" ""), tried: 
  #<path:/my/search/path/my-library-name.so> (no such file)
  #<path:/my/search/path/my-library-name> (no such file)
  "my-library-name.so" (using OS library search path)
  "my-library-name" (using OS library search path)
  #<path:/home/samth/work/teaching/c211/my-library-name.so> (no such file)
  #<path:/home/samth/work/teaching/c211/my-library-name> (no such file)
ffi-lib: couldn't open "my-library-name.so" (my-library-name.so: cannot open shared object file: No such file or directory)
  context...:
   /home/samth/sw/plt/racket/collects/ffi/unsafe.rkt:131:0: get-ffi-lib11
   /home/samth/sw/plt/racket/collects/racket/logging.rkt:43:0
   "/tmp/x.rkt": [running body]
   temp37_0
   for-loop
   run-module-instance!125
   perform-require!78
```

As you can see, it shows some logging output.

rmculpepper:
@elfprince13 I'm not sure how to intercept messages within Racket from previously-defined loggers, but if you start racket the following command:

    PLTSTDERR="debug@ffi-lib" racket

then it will print the `ffi-lib`-related log messages to stderr. Or in DrRacket, you can use `View | Show Log` and add `debug@ffi-lib` to the top list, and then when you run your program the log messages should appear in the log window.

elfprince13:
Thanks for the quick response.

> I'm not sure how to intercept messages within Racket from previously-defined loggers, 

My impression from reading the docs was that newly created loggers would fork off of the current logger, unless told to do otherwise, which it doesn't look like the `ffi-lib` logger is, so I was hoping they would propagate upwards somehow.

> but if you start racket the following command:
> 
> ```
> PLTSTDERR="debug@ffi-lib" racket
> ```
> 
> then it will print the `ffi-lib`-related log messages to stderr. Or in DrRacket, you can use `View | Show Log` and add `debug@ffi-lib` to the top list, and then when you run your program the log messages should appear in the log window.

Unfortunately, I need to intercept these programmatically. I wanted to record the path discovered by `ffi-lib` as a bytecode dependency using [`register-external-file` from `compiler/cm-accomplice`](https://docs.racket-lang.org/raco/cm-accomplice.html#%28def._%28%28lib._compiler%2Fcm-accomplice..rkt%29._register-external-file%29%29), because I have a macro that depends on calls to an FFI function, and I need the bytecode to be recompiled when the library is updated.

Is it likely that a PR to either provide the logger, or to refactor `get-ffi-lib` to provide a helper method that just returns the filename would be accepted?

samth:
If I change your code to supply `#:logger (current-logger)` to `with-logging-to-port` I get the output I think you want.

elfprince13:
Thanks! I stopped reading after:
> Runs proc, outputting any logging that the execution of proc emits to current-logger at the specified levels and topics. 

Apparently I needed to run one line further:
> If #:logger is specified, intercepts events sent to that logger, otherwise uses a new child logger of the current logger.

Also - it turns out that there's an undocumented function named `ffi-lib-name` provided by `ffi/unsafe` which solves my problem without needing to fiddle with the logger at all.

PR #2567, opened 2019-03-27
Fix: make-complex -> make-rectangular

MichaelMMacLeod:
Fixes a simple documentation error; the function to create a complex number is `make-rectangular`, not `make-complex`.


gus-massa:
Thanks. Merged in https://github.com/racket/racket/commit/fdf13f9dfc7f179bc5d7a08a88b3449e1a6098ff

I actually like `make-complex` more. If there is ever a decision to break massively backward compatibility, I vote for  `make-complex` and `make-complex/polar`.

bennn:
Why not add those names?

Or instead, it might be helpful to add an `@index` so searches for `make-complex` point to the `make-rectangular` docs.

PR #2581, opened 2019-04-04
Preserve procedure name in single-argument make-keyword-procedure

lexi-lambda:
Currently, code that uses `make-keyword-procedure` with only one argument will end up with procedures whose names point inside `racket/private/kw.rkt`, which isn’t very helpful. This PR sticks a call to `procedure-rename` in the relevant place so that the procedure name from the first argument is used, instead.


lexi-lambda:
I realized after opening this PR that my previous implementation had a small issue, which was that procedures with no name (that is, procedures for which `object-name` produced `#f`) would cause problems. That led to me realizing some other places in the keyword procedure implementation where things could go wrong if input syntax had no source locations, so I’ve updated the PR with a slightly more general change that fixes those issues, too.

mflatt:
LGTM

PR #2582, opened 2019-04-04
class/c: Swap after, not before, adding #:important blame context

### LINKED ISSUE #2531: Blame seems wrong for `override` clauses in class/c
Here’s a small program, modified very slightly from the examples in the documentation for `class/c`:

```racket
#lang racket

(module a racket
  (provide woody2+override/c%)

  (define woody%
    (class object%
      (define/public (draw who)
        (format "reach for the sky, ~a" who))
      (super-new)))

  (define/contract woody2+override/c%
    (class/c (override [draw (->m symbol? string?)]))
    (class woody%
      (inherit draw)
      (define/public (draw2 a b)
        (string-append (draw a)
                       " and "
                       (draw b)))
      (super-new))))
 
(module b racket
  (require (submod ".." a))

  (provide woody2+broken-draw%)

  (define woody2+broken-draw%
    (class woody2+override/c%
      (define/override (draw x)
        'not-a-string)
      (super-new))))

(require 'b)
(send (new woody2+broken-draw%) draw2
      'evil-dr-porkchop
      'zurg)
```

It fails with the following error:

```
draw: broke its own contract
  promised: string?
  produced: 'not-a-string
  in: the range of
      the draw method in
      (class/c
       (override (draw (->m symbol? string?))))
  contract from: 
      (definition woody2+override/c%)
  contract on: woody2+override/c%
  blaming: (anonymous-module a)
  at: unsaved-editor:12.19
```

This is a positive violation, as indicated by the “broke its own contract” message, and it blames module `a`. But module `a` isn’t at fault, module `b` is.

I imagine this was just a small oversight in the implementation: `override` contracts are a little different from other `class/c` contracts, since instead of being a contract between a method in the contracted class and some other party adding obligations to the contracted class itself, it’s effectively a contract on overriding methods in *subclasses*, and it’s the subclasses that accept obligation to fulfill (the positive parts of) the contract. So a `blame-swap` call is probably missing someplace.

Therefore, I think blame in `override` contracts needs to be swapped. The negative party for `override` contracts represents the subclass, and the positive party represents the contracted class, but currently this seems to be backwards.

lexi-lambda:
Swapping the blame before adding #:important context associates the important party with the negative party for the purposes of picking “contract violation” versus “broke its own contract” messages in error reporting. Therefore, only swap after adding the context.

Fixes #2531.

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/test-util.rkt
+++ b/pkgs/racket-test/tests/racket/contract/test-util.rkt
@@ -351,7 +351,8 @@
 
 ;; blame : (or/c 'pos 'neg string?)
 ;;   if blame is a string, expect to find the string (format "blaming: ~a" blame) in the exn message
-(define (test/spec-failed name expression blame)
+;; important : (or/c (cons/c string? (or/c 'pos 'neg)) #f)
+(define (test/spec-failed name expression blame #:important [important #f])
   (define (has-proper-blame? msg)
     (define reg
       (cond
@@ -360,6 +361,20 @@
         [(string? blame) (string-append "blaming: " (regexp-quote blame))]
         [else #f]))
     (and reg (regexp-match? reg msg)))
+  (define (has-proper-important-blame? msg)
+    (cond
+      [(not important)]
+      [(not (or (eq? blame 'pos) (eq? blame 'neg)))
+       #f]
+      [else
+       (define important-name (car important))
+       (define important-party (cdr important))
+       (define self-or-not
+         (if (equal? (eq? blame 'pos) (eq? important-party 'pos))
+             "broke its own contract"
+             "contract violation"))
+       (define reg (string-append (regexp-quote important-name) ": " self-or-not))
+       (regexp-match? reg msg)]))
   (contract-eval
    #:test-case-name name
    `(,test-an-error
@@ -368,7 +383,8 @@
      ',expression
      (lambda (exn)
        (and (exn:fail:contract:blame? exn)
-            (,has-proper-blame? (exn-message exn))))))
+            (,has-proper-blame? (exn-message exn))
+            (,has-proper-important-blame? (exn-message exn))))))
   (define (rewrite-test wrapper wrapper-name short-wrapper-name)
     (unless (member short-wrapper-name (contract-rewrite-tests-to-skip))
       (let/ec k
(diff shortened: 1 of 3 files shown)


lexi-lambda:
@rfindler Can I ask you to take a look at this?

pkgs/racket-test/tests/racket/contract/test-util.rkt
```diff
@@ -360,6 +361,20 @@
         [(string? blame) (string-append "blaming: " (regexp-quote blame))]
         [else #f]))
     (and reg (regexp-match? reg msg)))
+  (define (has-proper-important-blame? msg)
+    (cond
+      [(not important)]
+      [(not (or (eq? blame 'pos) (eq? blame 'neg)))
```
rfindler: Is this clause necessary?

pkgs/racket-test/tests/racket/contract/test-util.rkt
```diff
@@ -360,6 +361,20 @@
...
+      [else
+       (define important-name (car important))
+       (define important-party (cdr important))
+       (define self-or-not
+         (if (equal? (eq? blame 'pos) (eq? important-party 'pos))
+             "broke its own contract"
+             "contract violation"))
+       (define reg (string-append (regexp-quote important-name) ": " self-or-not))
+       (regexp-match? reg msg)]))
   (contract-eval
```
rfindler: actually, how about a different strategy here. I think there are two things this is checking: if "broke its own" or "contract violation" occurs, and then what name appears. Can we make that be two parameters to the function, with a default value that means "don't check this part of the test"?

rfindler:
I like the overall change, but have a quibble with the way the test case infrastructure is changed.

If you want, I can do it my own way and then rebase (and probably edit slightly) your commit?

lexi-lambda:
I’m happy to try to make the changes you suggested, but if you’d like to make them yourself, by all means, feel free.

rfindler:
Thanks!

samth:
This changed an error message that TR was expecting, roughly changing "promised" to "expected" and "produced" to "given". Was this intended (in which case I'll just change the TR test)?

bennn:
@samth I think the test should change

For everyone else, here's the test and output:
```
#lang racket

;; Ensure contracts for inner work correctly

(module t typed/racket
  (provide c%)
  (define c%
    (class object%
      (super-new)
      (: m (-> Void) #:augment (-> String))
      (define/pubment (m) (inner "hi" m) (void)))))

(require (submod "." t))
(send (new (class c%
             (super-new)
             (define/augment (m) 'not-a-string)))
      m)

;; ==>
;; m: contract violation
;;   expected: String
;;   given: 'not-a-string
;;   in: the range of
;;       the m method in
;;       g28
;;   contract from: 
;;       (/Users/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-test/fail/class-contract-1.rkt
;;        t)
;;   contract on: c%
;;   blaming: <pkgs>/typed-racket-test/fail/class-contract-1.rkt
;;    (assuming the contract is correct)
;;   at: <pkgs>/typed-racket-test/fail/class-contract-1.rkt:8.11
;;   context...:
;;    /Users/ben/code/racket/fork/racket/collects/racket/contract/private/blame.rkt:347:0: raise-blame-error16
;;    /Users/ben/code/racket/fork/racket/collects/racket/contract/private/arrow-higher-order.rkt:364:33
;;    m method in c%
;;    "/Users/ben/code/racket/fork/extra-pkgs/typed-racket/typed-racket-test/fail/class-contract-1.rkt": [running body]
;;    temp37_0
;;    for-loop
;;    run-module-instance!125
;;    perform-require!78
```

rfindler:
Yes, that's intended. Are you okay (as Ben is) with the example's output?

samth:
Yes, I just hadn't expected that change to affect the wording and wanted to confirm that it was right.

rfindler:
It affects the wording (well, technically, it affects the choice of which
wording).

Robby

PR #2584, opened 2019-04-04
Make second argument of scheme_os_getcwd a size_t

pmatos:
By being an unsigned type, we ensure that when 0 is passed the compiler and static analyzer knows that there's no smaller number. 

This not only aids the static analyzer but provides semantic information for those reading the code. Lengths should in general be `size_t`. Setting them as `int` means we do not enforce the fact that the variable needs to be positive. 

I remember @mflatt mentioning sometime ago that Racket is meant to support some older C compilers. Hopefully these older C compilers have `size_t` - after all this is C99 stuff which is now a 20 year old standard. However, I think we already require `size_t` support since there are a few occurences of `size_t` in the codebase.

### DIFF
--- a/racket/src/racket/src/file.c
+++ b/racket/src/racket/src/file.c
@@ -884,7 +884,7 @@ static Scheme_Object *string_to_path_element(int argc, Scheme_Object **argv)
 /*                                                                    */
 /**********************************************************************/
 
-char *scheme_os_getcwd(char *buf, int buflen, int *actlen, int noexn)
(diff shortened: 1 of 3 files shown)


mflatt:
LGTM. I worry a little about changing the type on a public function, but it seems very unlikely to cause a problem. There are other `size_t`s in "schemef.h", so let's assume that it's ok.

pmatos:
Thanks. I tested it without any issues so will merge.

PR #2587, opened 2019-04-04
Make equal? on classes and objects see through class contracts

### LINKED ISSUE #2279: class/c contracts do not preserve equal?
This program produces `#f`:

```racket
#lang racket

(define foo%
  (class object%
    (super-new)))

(define/contract foo%-contracted (class/c) foo%)

(equal? foo% foo%-contracted)
```

lexi-lambda:
Making `equal?` do the right thing on classes turned out to be easy—it just involved adding a straightforward `prop:equal+hash` property to the `class` struct—but making it work properly for *objects* was the tricky part. The trouble is that `equal?` on objects that don’t implement the `equal<%>` interface is just ordinary structure equality, which can be relevant if objects are inspectable. Writing `(inspect #f)` in a class body is like making a struct `#:transparent`, and it has all the same ramifications for equality.

The trouble is that `class/c` creates new wrapper classes, and every class has its own struct type. Since the default behavior of `equal?` on structs is to *never* be equal to structs of different types, even subtypes, an object created from a contracted class can never be `equal?` to an object created from the same class without contracts.

The solution is to add a `prop:equal+hash` property to `object%` itself that emulates the default behavior of `equal?`, but sees through class contract wrappers. Since struct type properties are inherited by subtypes, this property will be present on all objects, and it only needs to be attached once.

Fixes #2279.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/object.rktl
+++ b/pkgs/racket-test-core/tests/racket/object.rktl
@@ -1648,6 +1648,157 @@
   (check #t #t))
 
 ;; ----------------------------------------
+;; equal? on classes
+
+(let ()
+  (define a%
+    (class object%
+      (init-field x)
+      (super-new)
+      (define/public (m) #f)))
+
+  ; subclasses are never equal? to their superclasses
+  (define b%
+    (class a%
+      (super-new)))
+  (test #f equal? a% b%)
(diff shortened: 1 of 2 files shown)


mflatt:
Ok as far as I can tell. This one might be better to save until after the impending v7.3 release branch, though.

PR #2588, opened 2019-04-05
Remove value store in ready_pos but unread

pmatos:
`ready_pos` is not read after the store of 0. Remove store.

### DIFF
--- a/racket/src/racket/src/network.c
+++ b/racket/src/racket/src/network.c
@@ -1204,7 +1204,8 @@ do_tcp_accept(int argc, Scheme_Object *argv[], Scheme_Object *cust, char **_fail
   was_closed = LISTENER_WAS_CLOSED(listener);
 
   if (!was_closed) {
-    int ready_pos = tcp_check_accept(listener, NULL);
+    int ready_pos;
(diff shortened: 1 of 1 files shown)


pmatos:
I went ahead moved the declaration to the block and removed another value that was set but not read, marking it with `(void)tcp...`. Waiting to see results of CI before merging.

pmatos:
oh... xform doesn't like the `(void)tcp_...`. Will have to look at it a bit later.

mflatt:
It's not the `(void)tcp_....` but the `int ready_pos = tcp_check_accept(listener, NULL);`. To accommodate xform, ou'll have to split that to
```
int ready_pos;
ready_pos = tcp_check_accept(listener, NULL);
```

pmatos:
ah, thanks for pointing that out - I didn't have time to look into it yet.

pmatos:
Lets hope xform is happy with https://github.com/racket/racket/pull/2588/commits/b6469e1e1bfe9e522d493b0cf3881c9bdcd43fd2

pmatos:
All good! Merging.

PR #2589, opened 2019-04-05
Remove dead assignment to ch

pmatos:


pmatos:
Good style. Will fix and merge.

Issue #2548, opened 2019-03-20
Standard library missing a proper keyword sensative apply function


LeifAndersen:
While functions like [`apply`][1] accept keywords as arguments, they don't accept a list of keywords as arguments.

[`keyword-apply`][2] comes closer, but with two major problems: 1. The keywords and their values _must_ be separated into two lists, and those lists _must_ be sorted into `keword<?` order, otherwise the function will fail.

Ideally we should also have something that parses the keyword and by index arguments, so you could call it like this: `(better-keyword-apply f '(7 #:y 5 #:x 6))`

The code for that specific function is:

```
(define (better-keyword-apply func args)
  (define-values (kws rest)
    (let loop ([args args]
               [kw '()]
               [rest '()])
      (cond
        [(null? args) (values kw rest)]
        [(keyword? (first args))
         (loop (cddr args)
               (cons (cons (first args) (second args)) kw)
               rest)]
        [else
         (loop (cdr args)
               kw
               (cons (car args) rest))])))
  (define sorted-kws (sort kws keyword<? #:key car))
  (keyword-apply func
                 (map car sorted-kws)
                 (map cdr sorted-kws)
(reverse rest)))
```

We do have something like this for macros, in the `syntax/parse/lib/function-header` module, which provides `function-header` as a syntax-class. But asking users to turn their list into a syntax object, pass it through syntax parse, order the newly separated lists, and then turn them back to raw datums, seems like a bit much to me...

[1]: https://docs.racket-lang.org/reference/procedures.html?q=apply#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._apply%29%29
[2]: https://docs.racket-lang.org/reference/procedures.html?q=keyword-apply#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._keyword-apply%29%29

AlexKnauth:
I don't think this version of `better-keyword-apply` would be a good idea, because it doesn't distinguish between keywords-for-arguments and positional-arguments-that-happen-to-be-keyword-values.

I think this version of `better-keyword-apply` would make more sense:
```
(better-keyword-apply f pos-args kw-args) -> any
  f : procedure?
  pos-args : list?
  kw-args : (dict/c keyword? any/c)
```
[...]

PR #2592, opened 2019-04-06
 add keyword-apply/dict to racket/dict 

AlexKnauth:
See #2548. The `keyword-apply/dict` function is a variation on `keyword-apply` that solves two problems:

 -  Keyword sorting
 -  Parallel-lists where keyword and value are associated only by having the "same index"

```racket
(keyword-apply/dict proc kw-dict pos-arg ... pos-args) -> any
      proc : procedure?
   kw-dict : (dict/c keyword? any/c)
   pos-arg : any/c
  pos-args : (listof any/c)
```

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/dicts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/dicts.scrbl
@@ -2,7 +2,8 @@
 @(require "mz.rkt" (for-label racket/generic))
 
 @(define dict-eval (make-base-eval))
-@examples[#:hidden #:eval dict-eval (require racket/dict racket/generic racket/contract)]
+@examples[#:hidden #:eval dict-eval
+          (require racket/dict racket/generic racket/contract racket/string)]
 
 @title[#:tag "dicts"]{Dictionaries}
 
@@ -1013,5 +1014,53 @@ See also @racket[define-custom-hash-types].
 
 
 }
+
+@section{Passing keyword arguments in dictionaries}
+
+@defproc[
+ (keyword-apply/dict [proc procedure?]
+                     [kw-dict dict?] ; (dict/c keyword? any/c)
+                     [pos-arg any/c] ...
+                     [pos-args (listof any/c)]
+                     [#:<kw> kw-arg any/c] ...)
+ any]{
+Applies the @racket[proc] using the positional arguments
+from @racket[(list* pos-arg ... pos-args)], and the keyword
+arguments from @racket[kw-dict] in addition to the directly
+supplied keyword arguments in the @racket[#:<kw> kw-arg]
+ sequence.
+
+All the keys in @racket[kw-dict] must be keywords.
+The keywords in the @racket[kw-dict] do not have to be
+sorted. However, the keywords in @racket[kw-dict] and the
+directly supplied @racket[#:<kw>] keywords must not overlap.
+The given @racket[proc] must accept all of the keywords in
+@racket[kw-dict] plus the @racket[#:<kw>]s.
+
+@examples[
+#:eval dict-eval
+(define (sundae #:ice-cream [ice-cream '("vanilla")]
+                #:toppings [toppings '("brownie-bits")]
+                #:sprinkles [sprinkles "chocolate"]
+                #:syrup [syrup "caramel"])
+  (format "A sundae with ~a ice cream, ~a, ~a sprinkles, and ~a syrup."
+          (string-join ice-cream #:before-last " and ")
+          (string-join toppings #:before-last " and ")
+          sprinkles
+          syrup))
+(keyword-apply/dict sundae '((#:ice-cream . ("chocolate"))) '())
+(keyword-apply/dict sundae
+                    (hash '#:toppings '("cookie-dough")
+                          '#:sprinkles "rainbow"
+                          '#:syrup "chocolate")
+                    '())
+(keyword-apply/dict sundae
+                    #:sprinkles "rainbow"
+                    (hash '#:toppings '("cookie-dough")
+                          '#:syrup "chocolate")
+                    '())
+]
+}
+
 
 @close-eval[dict-eval]
--- a/pkgs/racket-doc/scribblings/reference/dicts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/dicts.scrbl
@@ -1060,7 +1060,7 @@ The given @racket[proc] must accept all of the keywords in
                           '#:syrup "chocolate")
                     '())
 ]
-}
+@history[#:added "7.9"]}
 
 
 @close-eval[dict-eval]
--- a/pkgs/racket-test/tests/racket/keyword-apply-dict.rkt
+++ b/pkgs/racket-test/tests/racket/keyword-apply-dict.rkt
@@ -1,0 +1,98 @@
+#lang racket/base
+(require rackunit racket/dict racket/list racket/string racket/math)
+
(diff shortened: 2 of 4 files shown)


sorawee:
In the same spirit as `apply`, can `keyword-apply/dict` support `#:<kw> kw-arg ...` too?

jackfirth:
@sorawee That seems reasonable but also a little dangerous, since the static keywords could overlap with the dynamic ones. Example: `(keyword-apply/dict some-keyword-function (hash '#:color "green") #:color "red" '())`. I assume that would be an error?

@AlexKnauth The `gen:dict` interface allows duplicate keys (but leaves their behavior unspecified), and lists implement the dictionary interface by using `assoc`. What should this code do?

```scheme
(keyword-apply/dict
    some-keyword-function
    '((#:color . "red") (#:color . "green"))
    '())
```

I think a naive implementation of `keyword-apply/dict` would ignore the second color keyword (because that's what `dict-ref` on an association list would do), but I'd much rather the implementation throw an error. If this happened in real code it almost certainly means there's a bug hidden somewhere.

AlexKnauth:
I think that `keyword-apply/dict` should interpret all valid dicts as dict operations would interpret them. Either that's an invalid dict and `dict?` should return false, or `keyword-apply/dict` should interpret the dict as the dict operations do.

AlexKnauth:
@sorawee It does support `#:<kw> kw-arg ...`, but I didn't document that. Maybe I should.

@jackfirth `(keyword-apply/dict some-keyword-function (hash '#:color "green") #:color "red" '())` is an error just like `(keyword-apply some-keyword-function '(#:color) '("green") #:color "red" '())` is, although the error message is about `keyword-apply` instead of `keyword-apply/dict` because it just delegates to `keyword-apply` of `keyword-apply` in that situation.

AlexKnauth:
@sorawee I have now documented that `keyword-apply/dict` supports `#:<kw> kw-arg ...`. The rendered docs look like
docs

AlexKnauth:
@jackfirth I have updated the error message, to be about `keyword-apply/dict` instead of `keyword-apply`.
```
> (keyword-apply/dict void (hash '#:color "green") #:color "red" '())
keyword-apply/dict: keyword duplicated in dict and direct keyword arguments: '#:color
```

sorawee:
What is `dict/c`? I searched and only found one from `mischief` which I don't think is what you want.

pkgs/racket-doc/scribblings/reference/dicts.scrbl
```diff
@@ -1014,4 +1015,51 @@ See also @racket[define-custom-hash-types].
 
 }
 
+@section{Passing keyword arguments in dictionaries}
+
+@defproc[
+ (keyword-apply/dict [proc procedure?]
+                     [kw-dict (dict/c keyword? any/c)]
```
sorawee, line 1022: What is `dict/c`? I searched and only found one from `mischief` which I don't think is what you want.

AlexKnauth: Oh, I had assumed `(dict/c keyword? any/c)` was like `(hash/c keyword? any/c)`, but for generic dicts and not just hash tables. But I suppose if it doesn't exist, maybe just `dict?` should go there? Although I would prefer it to say the keys must be keywords... would `(and/c dict? (compose (listof keyword?) dict-keys))` work better? Kinda long though.

AlexKnauth:
Is it okay if I try to copy the implementation of `hash/c` and modify it to define `dict/c`?

samth:
I think that should be a separate PR. Maybe just use `dict?` here.

pkgs/racket-doc/scribblings/reference/dicts.scrbl
AlexKnauth: Okay, I've changed it to `dict?` for now

samth:
This needs some tests, and a history note. Also, I'm not sure that the `dict` documentation is the right place for this. In general, since this seems like a broadly-useful feature having it presented as dict-specific seems wrong.

sorawee:
Perhaps `racket/function` is a more appropriate place?

AlexKnauth:
Currently `racket/function` doesn't depend on `racket/dict` or `racket/private/dict`. Is it okay to add that dependency?

If so then `racket/function` sounds like a good place.

sorawee:
Yeah, that dependency will definitely slow down compilation time a bit.

Following is the result of running `for i in {1..10}; do time racket test.rkt; done` on various programs:

```
#lang racket/base

#|
racket test.rkt  0.12s user 0.03s system 97% cpu 0.154 total
racket test.rkt  0.13s user 0.03s system 94% cpu 0.168 total
racket test.rkt  0.11s user 0.03s system 97% cpu 0.149 total
racket test.rkt  0.12s user 0.03s system 96% cpu 0.157 total
racket test.rkt  0.12s user 0.03s system 96% cpu 0.154 total
racket test.rkt  0.12s user 0.03s system 94% cpu 0.157 total
racket test.rkt  0.12s user 0.03s system 96% cpu 0.151 total
racket test.rkt  0.12s user 0.03s system 96% cpu 0.160 total
racket test.rkt  0.12s user 0.03s system 95% cpu 0.160 total
racket test.rkt  0.12s user 0.03s system 94% cpu 0.157 total
|#
```

```
#lang racket/base

(require racket/function)

#| 
racket test.rkt  0.12s user 0.03s system 95% cpu 0.161 total
racket test.rkt  0.14s user 0.04s system 89% cpu 0.196 total
racket test.rkt  0.12s user 0.03s system 97% cpu 0.158 total
racket test.rkt  0.12s user 0.03s system 96% cpu 0.159 total
racket test.rkt  0.12s user 0.03s system 95% cpu 0.155 total
racket test.rkt  0.12s user 0.03s system 97% cpu 0.154 total
racket test.rkt  0.13s user 0.03s system 95% cpu 0.169 total
racket test.rkt  0.12s user 0.03s system 95% cpu 0.158 total
racket test.rkt  0.12s user 0.03s system 98% cpu 0.156 total
racket test.rkt  0.12s user 0.03s system 97% cpu 0.151 total
|#
```

```
#lang racket/base

(require racket/dict)

racket test.rkt  0.20s user 0.05s system 96% cpu 0.256 total
racket test.rkt  0.20s user 0.05s system 97% cpu 0.253 total
racket test.rkt  0.20s user 0.05s system 96% cpu 0.251 total
racket test.rkt  0.20s user 0.05s system 93% cpu 0.272 total
racket test.rkt  0.19s user 0.05s system 96% cpu 0.251 total
racket test.rkt  0.20s user 0.05s system 95% cpu 0.257 total
racket test.rkt  0.21s user 0.05s system 95% cpu 0.273 total
racket test.rkt  0.20s user 0.05s system 95% cpu 0.260 total
racket test.rkt  0.20s user 0.05s system 95% cpu 0.269 total
racket test.rkt  0.19s user 0.05s system 97% cpu 0.246 total
```

samth:
I'm fine with leaving it in `racket/dict` given that. The other option would be to skip generics entirely, and use hashes.

[Aside: the `hyperfine` tool is very useful for that kind of measurement.]

AlexKnauth:
I have added some tests in `pkgs/racket-test/tests/racket/keyword-apply-dict.rkt`.

sorawee:
Can this be merged by 7.9 release?

AlexKnauth:
The last review mentioned 3 things: tests, a history note, and placement. I have added tests and a history note, and we agreed on placement in `racket/dict`. If there are no objections, can I merge at the end of the day today?

PR #2595, opened 2019-04-06
Establish free-id=? equivalences in intdefs with local bindings in scope

### LINKED ISSUE #2594: `define-rename-transformer-parameter` doesn’t work inside first-class definition contexts
The following program fails with an error:

```racket
#lang racket
(require racket/stxparam)
(local [(define-rename-transformer-parameter add1*
          (make-rename-transformer #'add1))]
  (void))
```
```
rename-transformer-accessor: contract violation
  expected: rename-transformer?
  given: #f
  context...:
   rename-transformer-target
   /Applications/Racket v7.2/collects/racket/private/stxparamkey.rkt:68:2: rename-transformer-parameter-target
   maybe-install-free=id!
   syntax-local-bind-syntaxes11
   /Applications/Racket v7.2/collects/racket/private/local.rkt:20:24
```

This happens because `syntax-local-bind-syntaxes` attempts to install `free-identifier=?` equivalences in a context without the local binding environment of the definition context. To illustrate that more directly, here’s a simpler program that also triggers the issue:

```racket
#lang racket
(begin-for-syntax
  (struct indirect-rename-transformer (target-holder)
    #:property prop:rename-transformer
    (lambda (self) (syntax-local-value (indirect-rename-transformer-target-holder self)))))
(local [(define-syntax holder #'add1)
        (define-syntax add1* (indirect-rename-transformer #'holder))]
  (add1* 10))
```
```
syntax-local-value: identifier is not bound to syntax: #<syntax:defctx-rename-trans.rkt:7:60 holder>
  context...:
   maybe-install-free=id!
   syntax-local-bind-syntaxes11
   /Applications/Racket v7.2/collects/racket/private/local.rkt:20:24
```

lexi-lambda:
Fixes #2594.


lexi-lambda:
From the Travis CI output, it seems like this change broke the cify step used in the no-JIT build, and I think that is beyond my knowledge to diagnose, much less fix. Still, if you can point me in the right direction, I can give it a try.

lexi-lambda:
With 0d2dd8f578b3e5baa2e043bc0aec940366eb7182 (thanks Matthew!), the no-JIT build now completes successfully, so all seems to be well from CI’s point of view.

PR #2598, opened 2019-04-09
Fix --enable-cify option

pmatos:
I guess, a previous copy-paste AC_ARG_ENABLE meant that cify never had
its own option.

### DIFF
--- a/racket/src/cfg-racket
+++ b/racket/src/cfg-racket
@@ -829,6 +829,7 @@ enable_cgcdefault
 enable_sgc
 enable_sgcdebug
 enable_backtrace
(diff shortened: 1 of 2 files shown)


pmatos:
@mflatt is it possible that we never noticed that there was no `--enable-cify`? or this PR is non-sensical and there's a good reason for the current state of things?

mflatt:
The current state definitely looks wrong, and this change looks right.

I think `--enable-cify` has been working, anyway, because the option to complain about unrecognized flags is not enabled (or it's explicitly disabled; I forget).

PR #2599, opened 2019-04-09
cs: fix check of Chez Scheme version

gus-massa:
cs: fix check of Chez Scheme version


mflatt:
LGTM - thanks!

gus-massa:
Do I ask for inclusion in 7.3?

mflatt:
That sounds like a good idea for this one.

PR #2602, opened 2019-04-10
Remove unnecessary MZ_USE_JIT

pmatos:
There are quite a few `MZ_USE_JIT` that if I am not mistaken, are unnecessary. The outermost stays but the others go. This allows me to move a few variable declarations into the `#if` block which raise a compiler warning in `aarch64` when `! MZ_USE_JIT` is in place.

### DIFF
--- a/racket/src/racket/src/jitprep.c
+++ b/racket/src/racket/src/jitprep.c
@@ -3,16 +3,16 @@
    or unmarshaled.
 
    See "eval.c" for an overview of compilation passes and JIT
-   prepraration. */
+   preparation. */
 
 #include "schpriv.h"
 #include "schrunst.h"
 #include "schmach.h"
 
 THREAD_LOCAL_DECL(static Scheme_Object *current_linklet_native_lambdas);
-static int force_jit;
 
 #ifdef MZ_USE_JIT
+static int force_jit;
 
 void scheme_init_jitprep()
 {
@@ -391,7 +391,6 @@ static Scheme_Object *with_immed_mark_jit(Scheme_Object *o)
 
 Scheme_Object *scheme_case_lambda_jit(Scheme_Object *expr)
 {
-#ifdef MZ_USE_JIT
   Scheme_Case_Lambda *seqin = (Scheme_Case_Lambda *)expr;
 
   if (!seqin->native_code) {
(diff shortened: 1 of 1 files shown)


racket/src/racket/src/jitprep.c
```diff
@@ -391,7 +391,6 @@ static Scheme_Object *with_immed_mark_jit(Scheme_Object *o)
 
 Scheme_Object *scheme_case_lambda_jit(Scheme_Object *expr)
 {
-#ifdef MZ_USE_JIT
   Scheme_Case_Lambda *seqin = (Scheme_Case_Lambda *)expr;
 
   if (!seqin->native_code) {
```
mflatt: The `native_code` field exists only when `MZ_USE_JIT` is defined, so that's why the `#ifdef` is needed around here.

racket/src/racket/src/jitprep.c
```diff
@@ -582,9 +579,6 @@ Scheme_Object *scheme_jit_closure(Scheme_Object *code, Scheme_Object *context)
     return scheme_make_native_closure(data2->u.native_code);
```
mflatt: ditto

racket/src/racket/src/jitprep.c
```diff
@@ -9,9 +9,8 @@
 #include "schrunst.h"
 #include "schmach.h"
 
-THREAD_LOCAL_DECL(static Scheme_Object *current_linklet_native_lambdas);
-
 #ifdef MZ_USE_JIT
+THREAD_LOCAL_DECL(static Scheme_Object *current_linklet_native_lambdas);
```
mflatt: This change makes sense.

pmatos:
Well, about this. The birds eye view of the file in respect to `MZ_USE_JIT` is:
```
17 #ifdef MZ_USE_JIT

394     #ifdef MZ_USE_JIT

476     #endif 

541     #ifdef MZ_USE_JIT

585     #endif 

725 #else 

732 #endif 
```

It seems to be that the `MZ_USE_JIT` block guards starting at 394 and 541 are unnecessary. If `MZ_USE_JIT` is undefined,  only the 725 - 732 lines remain. If it's defined, then 725 - 732 is removed and all the others remain. @mflatt what am I missing?

mflatt:
Oh, I see. Yes, that makes sense.

pmatos:
ok to merge?

mflatt:
Yes - thanks!

PR #2605, opened 2019-04-12
local-require: Remove last remaining use of syntax-local-get-shadower

lexi-lambda:
I heard for a long time that `syntax-local-get-shadower` was a necessary evil that ought to be removed, but I never really looked into what it actually did. I finally did so tonight, and I found its last remaining use inside `local-require`. I puzzled for a few moments over why it was necessary there, then eventually decided it wasn’t and removed it. All the tests still passed.

Maybe there’s something I’m missing, but `syntax-local-get-shadower` seems like it isn’t necessary for `local-require` and hasn’t been necessary since the switch to the sets-of-scopes expander. So this PR just goes ahead and removes it.


mflatt:
I could be missing something, too, but this looks like an improvement to me.

PR #2606, opened 2019-04-12
In autoconf 2.70 variable runstatedir was added - ignore

pmatos:
Ignore new autoconf variable added in 2.70.
The interesting thing is that debian decided to backport this variable
to their 2.69 release so in some 2.69 autoconf this variable does not
exist but in debian ports 2.69 generates this variable. It is
nonetheless not useful for Racket, so add to ignore list.


mflatt:
LGTM

PR #2609, opened 2019-04-14
typo: reword install/clone error message

bennn:
This PR identifiers, and maybe fixes, a confusing error message.

I'm not sure how this error can occur or what it means. So, this change is basically a guess at a better wording.


mflatt:
LGTM

PR #2614, opened 2019-04-18
typo: change error messages for vectorof contract

bennn:
- change an 'an' to 'a'
- remove 'immutable' where expecting either mutable or immutable (don't
  bother to specify which, because `vector-common.rkt` doesn't bother)
- remove extra ','


rfindler:
Thanks!

Issue #2572, opened 2019-03-30
Multiple levels of contract-out missing struct bindings


camoy:
Cropped up after applying fix for #2469.

```rkt
#lang racket/base

(module foo0 racket/base
  (require racket/contract/base)
  (struct foo ())
  (provide (contract-out (struct foo ()))))

(module foo1 racket/base
  (require racket/contract/base)
  (require (submod ".." foo0))
  (provide (contract-out (struct foo ()))))

(require (submod "." foo1))
foo?
```

yields

```rkt
foo?: unbound identifier in: foo?
```

All struct bindings seem missing when requiring from the second module except the constructor.

bennn:
`provide/contract-id-foo?` is bound. Hmm...

bennn:
b40e247eddf30bf0591f8e39a146009e936cc076

PR #2618, opened 2019-04-20
contract-out: struct, keep original struct id

bennn:
Possible fix for #2572

The idea is to use the original struct id to recover the original names.

This seemed like the most reliable way to un-mangle the names.

(now lets see if the tests pass --- I've only tested on the example from the issue)

Edit: also, I replaced `map/count` with `for/list ... in-naturals ...`

### DIFF
--- a/racket/collects/racket/contract/private/provide.rkt
+++ b/racket/collects/racket/contract/private/provide.rkt
@@ -43,15 +43,44 @@
                                   stx)]
       [_ (syntax orig)])))
 
-(define-for-syntax make-applicable-struct-info
-  (letrec-values ([(struct: make- ? ref set!)
-                   (make-struct-type 'self-ctor-struct-info struct:struct-info
-                                     1 0 #f
-                                     (list (cons prop:procedure
-                                                 (lambda (v stx)
-                                                   (self-ctor-transformer ((ref v 0)) stx))))
-                                     (current-inspector) #f '(0))])
-    make-))
+;; make-contract-out-redirect-struct-info
+;; : (-> (-> (and/c struct-info? list?)) (-> identifier?) struct-info?)
+;; Create a struct-info? value from two thunks:
+;;  the 1st must be a valid argument for `make-struct-info`, and
+;;  the 2nd must return an identifier for a structure type descriptor.
+;; The 2nd thunk is used to recover the original names for a struct --- from before
+;;  `contract-out` started to mangle them.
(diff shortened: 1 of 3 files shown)


rfindler:
I don't think the change in `private/helpers.rkt` is correct.

rfindler:
Oh! I should have read the rest of the diff. I didn't think of attacking the problem like that. I like this strategy!

racket/collects/racket/contract/private/provide.rkt
```diff
@@ -983,16 +1026,6 @@
                                               field-contract-ids ...))
                        (provide (rename-out [-struct:struct-name struct:struct-name]))))))))))
 
-       (define (map/count f . ls)
-         (let loop ([ls ls]
-                    [i 0])
-           (cond
-             [(andmap null? ls) '()]
-             [(ormap null? ls) (error 'map/count "mismatched lists")]
```
bennn: Removing `map/count` removes this error check.

@rfindler let me know if I should revert & keep `map/count`

racket/collects/racket/contract/private/provide.rkt
```diff
@@ -819,18 +856,19 @@
...
-                                                                      field-contract-id)
-                                             #f)
-                                            #f))
-                                      selector-ids
-                                      field-contract-ids))]
+                           (for/list ([selector-id (in-list selector-ids)]
+                                      [orig-selector-id (in-list orig-selector-ids)]
+                                      [field-contract-id (in-list field-contract-ids)]
+                                      [index (in-naturals)])
+                             (if (is-new-id? index)
```
samth: I think this could use `#:when` and avoid the `filter`.

bennn:
@rfindler  all the id-mangling in this file seems necessary --- the new ids are eventually used in a `define-syntax`

Except for the question about `map/count` vs `for` I think this is ready

samth:
Does this need anything else to get merged? We need this for our static contract verification work, and it would be great to not have to patch Racket.

rfindler:
Sorry, I had intended to put a little time into this as I was going back and forth on the merits of the strategy but in the interest of progress, I've pushed it. I think that the changes that I wanted to try out won't affect anything except the implementation so there is no need to wait. Thanks for your patience and sorry for the delay.

samth:
No problem! Thanks for the quick response.

Issue #2611, opened 2019-04-15
segfault with malformed date-* accessors in racket repl


juhaase:
Running the following code inside a racket repl results in a segfault:

```
(define (crash) (date-day))
(crash)
```
Output:
```
Welcome to Racket v7.3.0.1.
> (define (crash) (date-day))
> (crash)
SIGSEGV MAPERR si_code 1 fault on addr 0x1003e
Aborted (core dumped)
```
The same crash happens with `date-month`, `date-year` etc.

Calling `date-day` directly (without the procedure wrapper) correctly results in an arity error:
```
Welcome to Racket v7.3.0.1.
> (date-day)
; date-day: arity mismatch;
;  the expected number of arguments does not match the given number
;   expected: 1
;   given: 0
; [,bt for context]
```

Doing the same with a self-defined struct also gives the expected arity error:
```
Welcome to Racket v7.3.0.1.
> (struct s (x))
> (define (crash) (s-x))
> (crash)
; s-x: arity mismatch;
;  the expected number of arguments does not match the given number
;   expected: 1
;   given: 0
; [,bt for context]
```

Racket Version:
```
Snapshot 20190414-e1ad8d4721
```

compiled with:
```
> gcc -v
gcc version 7.3.0 (Ubuntu 7.3.0-27ubuntu1~18.04)
```

on system:
```
> uname -a
Linux z820-server 4.15.0-39-generic #42-Ubuntu SMP Tue Oct 23 15:48:01 UTC 2018 x86_64 x86_64 x86_64 GNU/Linux
```

gus-massa:
All the structs defined in the kernel in [struct.c#L230-L235](https://github.com/racket/racket/blob/00a50ca772b555346d897d5811eeccb6807433c1/racket/src/racket/src/struct.c#L230-L235) have the same problem. The JIT compiler is not checking that they are call with the correct arity, to most of the times there is a segfault.

Thanks for the report.

PR #2619, opened 2019-04-20
Fix JIT-inlining of struct primitives

gus-massa:
Some expressions like `(date-day)` give usually an arity error, but when they are inlined by the JIT the arity check is wrong, so they produce a segfault or a nonsensical result.

[I'm not familiar with this part of the code, but the change looks very sensible.] 

This fix https://github.com/racket/racket/issues/2611


mflatt:
LGTM - thanks!

Issue #1495, opened 2016-10-25
free-identifier=? raises exception when called on identifier stored in preserved syntax property


lexi-lambda:
Apologies for the word salad title, but it pretty much summarizes the issue. I’m not totally sure if this is a bug, *per se*, or if it’s works-as-intended behavior I don’t understand, but either way, I didn’t find anything in the documentation that explains quite what I’m seeing here.

Here’s a program that reproduces the issue, made up of two files:

```racket
;;; stasher.rkt
#lang racket/base

(require (for-syntax racket/base
                     syntax/transformer)
         syntax/parse/define)

(provide foo foo-stash
         print-stash-free-id=?)

(define-syntax-parser def-id-stash
  [(_ x:id id:id)
   #:with proxy (syntax-property #'carrier 'prop (syntax-local-value #'id) #t)
   #'(define-syntax x
       (make-variable-like-transformer
        (syntax-property #'(void) 'stash
                         (syntax-property #'proxy 'prop))))])

(define-syntax-parser print-stash-free-id=?
  [(_ box-id stash)
   (define expanded-stash (local-expand #'stash 'expression '()))
   (println (free-identifier=? (syntax-local-value #'box-id)
                               (syntax-property expanded-stash 'stash)))
   #'(void)])

(define-syntax foo #'foo)
(def-id-stash foo-stash foo)

(print-stash-free-id=? foo foo-stash)
```
```racket
;;; printer.rkt
#lang racket/base

(require "stasher.rkt")

(print-stash-free-id=? foo foo-stash)
```

The first module, `stasher.rkt`, compiles fine. However, `printer.rkt` raises an error at compile-time:

```
printer.rkt:31:12: module-path-index-resolve: "self" index has no resolution
  module path index: #<module-path-index:()>
```

What’s going on here is sort of complicated, so let me break down what is happening:

  1. I define a syntax binding, `foo`, which is bound to an identifier that references itself, `#'foo`.

  2. I use a macro called `def-id-stash` to create a macro that *expands* to a piece of syntax containing a syntax property. Inside of that syntax property is the `#'foo` identifier stored in `foo`. This is done by expanding to a use of `make-variable-like-transformer`. *However*, importantly, the way the property’s value is passed is not direct—it is itself passed as a preserved syntax property.

[...]

PR #2624, opened 2019-04-26
Add `syntax/format`, which contains `~id`, `~id/1`, and `~symbol`

lexi-lambda:
This PR adds a new library, `syntax/format`, which provides a `racket/format`-like interface to the functionality of `format-id` and `format-symbol` from `racket/syntax`. The chief difference between the two is that `~id` automatically adds `'sub-range-binders` properties as appropriate by default, which helps macro authors cooperate with Check Syntax.

Examples:

```racket
> (~id #'hyphenated "-" #'id)
#<syntax hyphenated-id>
> (~id #'made #\- 'from #\- '#:many #\- "pieces")
#<syntax made-from-many-pieces>
> (syntax-property (~id #'foo "-" #'bar) 'sub-range-binders)
'(#(#<syntax foo-bar> 0 3 0.5 0.5 #<syntax:eval:1:0 foo> 0 3 0.5 0.5)
  #(#<syntax foo-bar> 4 3 0.5 0.5 #<syntax:eval:1:0 bar> 0 3 0.5 0.5))
```

The main reason this PR creates a new module instead of adding to `racket/syntax` is to avoid making `racket/syntax` depend on the contract system. If people dislike that decision, it would be possible to either adjust the implementation to do the checks manually, without using the contract system, or to move most of `racket/syntax` into a new `racket/private/syntax` module to break any circular dependencies in `collects/`. Any of those is fine with me; I don’t have a strong opinion on the matter. (In general I think it’s nice to have syntax helpers go in the `syntax/` collection, but the existing `format-id` and `format-symbol` bindings are already in `racket/syntax`, so it could make sense for them to go together.)

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/format.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/format.scrbl
@@ -15,8 +15,9 @@
 @defproc[(~id [v (or/c identifier? string? symbol? keyword? char? number?)] ...
               [#:context context (or/c syntax? #f) #f]
               [#:source source source-location? context]
-              [#:props props (or/c syntax? #f 'infer) 'infer]
-              [#:track track (or/c #t (-> identifier? identifier?) #f) #t])
+              [#:props props (or/c syntax? #f) (if track-introduce #f context)]
+              [#:track track-introduce (or/c (-> identifier? identifier?) #f)
+               (if (syntax-original?) syntax-local-introduce identity)])
          identifier?]{
 
 Builds an @tech[#:doc refman]{identifier} with a name constructed from the @racket[v]s, using
@@ -31,16 +32,14 @@ preceeding @litchar{#:} characters are not included).
   (~id #'hyphenated "-" #'id)
   (~id #'made #\- 'from #\- '#:many #\- "pieces"))
 
-If @racket[props] is a @tech[#:doc refman]{syntax object} or @racket[#f], then
+If @racket[props] is a @tech[#:doc refman]{syntax object}, then
 @tech[#:doc refman #:key "syntax property"]{syntax properties} from @racket[props] are copied to the
-new identifier in the same way as for @racket[datum->syntax]. If @racket[props] is @racket['infer],
-then the properties of the new identifier are copied from @racket[context] if @racket[track] is
-@racket[#f]; if @racket[track] is not @racket[#f], then no properties are copied.
+new identifier in the same way as for @racket[datum->syntax].
 
-If @racket[track] is not @racket[#f], then a @racket['sub-range-binders] @tech[#:doc refman]{syntax
-property} is added to the new identifier. The property value is a list containing an element for each
-@tech[#:doc refman]{identifier} @racket[_v-id] provided for @racket[v], and each element is an
-immutable @tech[#:doc refman]{vector} with the shape
+If @racket[track-introduce] is not @racket[#f], then a @racket['sub-range-binders]
+@tech[#:doc refman]{syntax property} is added to the new identifier. The property value is a list
+containing an element for each @tech[#:doc refman]{identifier} @racket[_v-id] provided for @racket[v],
+and each element is an immutable @tech[#:doc refman]{vector} with the shape
 
 @(racketblock
   (vector-immutable _result-id _v-id-position _v-id-length 0.5 0.5
@@ -57,24 +56,25 @@ the identifier should be treated as uses of the @racket[_v-id]s; see
   #:eval (make-format-eval) #:once
   (syntax-property (~id #'foo "-" #'bar) 'sub-range-binders))
 
-If @racket[track] is a procedure, then it is applied to @racket[_result-id] and to each @racket[_v-id]
-inside the @racket['sub-range-binders] property. If @racket[track] is @racket[#t], then
-@racket[syntax-local-introduce] is used if @racket[(syntax-transforming?)] is @racket[#t], otherwise
-@racket[identity] is used. (This default corresponds to the correct behavior when each @racket[_v-id]
-comes from the input to a @tech[#:doc refman]{syntax transformer}.)
+When @racket[track-introduce] is a procedure, then it is applied to @racket[_result-id] and to each
+@racket[_v-id] inside the @racket['sub-range-binders] property. The default value uses
+@racket[syntax-local-introduce] if @racket[(syntax-transforming?)] is @racket[#t], and otherwise uses
+@racket[identity]; this accommodates the common case when each @racket[_v-id] comes from the input to
+a @tech[#:doc refman]{syntax transformer}.
 
 @history[#:added "7.3.0.3"]}
 
 @defproc[(~id/1 [v (or/c identifier? string? symbol? keyword? char? number?)] ...
                 [#:context context (or/c syntax? #f 'infer) 'infer]
                 [#:source source (or/c source-location? 'infer) 'infer]
-                [#:props props (or/c syntax? #f 'infer) 'infer]
-                [#:track track (or/c #t (-> identifier? identifier?) #f) #t])
+                [#:props props (or/c syntax? #f 'infer) (if track-introduce #f context)]
+                [#:track track-introduce (or/c #t (-> identifier? identifier?) #f)
+                 (if (syntax-original?) syntax-local-introduce identity)])
          identifier?]{
 
 Like @racket[~id], but exactly one @racket[v] must be an @tech[#:doc refman]{identifier}, which is
-automatically used in place of @racket[context] and/or @racket[source] if their value is
-@racket['infer]. All other arguments are handled the same way as for @racket[~id].
+automatically used in place of @racket[context], @racket[source], and/or @racket[props] if their value
+is @racket['infer]. All other arguments are handled the same way as for @racket[~id].
 
 @(examples
   #:eval (make-format-eval) #:once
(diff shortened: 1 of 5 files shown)


pkgs/racket-doc/syntax/scribblings/format.scrbl
```diff
@@ -0,0 +1,106 @@
...
+                     racket/format
+                     syntax/format
+                     syntax/srcloc))
+
+@(define make-format-eval (make-eval-factory '[syntax/format]))
+
+@title[#:tag "format"]{Creating Formatted Identifiers}
+@defmodule[syntax/format]
+
+@defproc[(~id [v (or/c identifier? string? symbol? keyword? char? number?)] ...
```
bennn: A `v` can be an `identifier?` too

pkgs/racket-doc/syntax/scribblings/format.scrbl
```diff
@@ -0,0 +1,106 @@
...
+preceeding @litchar{#:} characters are not included).
+
+@(examples
+  #:eval (make-format-eval) #:once
+  (~id #'hyphenated "-" #'id)
+  (~id #'made #\- 'from #\- '#:many #\- "pieces"))
+
+If @racket[props] is a @tech[#:doc refman]{syntax object} or @racket[#f], then
+@tech[#:doc refman #:key "syntax property"]{syntax properties} from @racket[props] are copied to the
+new identifier in the same way as for @racket[datum->syntax]. If @racket[props] is @racket['infer],
```
bennn: Should `props` be `prop`, to match the argument name in the docs for `datum->syntax`?

pkgs/racket-doc/syntax/scribblings/format.scrbl
```diff
@@ -0,0 +1,106 @@
...
+
+@(examples
+  #:eval (make-format-eval) #:once
+  (syntax-property (~id #'foo "-" #'bar) 'sub-range-binders))
+
+If @racket[track] is a procedure, then it is applied to @racket[_result-id] and to each @racket[_v-id]
+inside the @racket['sub-range-binders] property. If @racket[track] is @racket[#t], then
+@racket[syntax-local-introduce] is used if @racket[(syntax-transforming?)] is @racket[#t], otherwise
+@racket[identity] is used. (This default corresponds to the correct behavior when each @racket[_v-id]
+comes from the input to a @tech[#:doc refman]{syntax transformer}.)
```
bennn: Add a `defproc` for the default?

If you do, then it might be helpful to name the new function as the default for `#:track`, instead of `#t`

racket/collects/syntax/private/format.rkt
```diff
@@ -0,0 +1,163 @@
...
+
+(define (jumble->list v leaf?)
+  (let recur ([v v])
+    (cond
+      [(leaf? v) (list v)]
+      [(list? v) (append-map recur v)]
+      [(pair? v) (append (recur (car v)) (recur (cdr v)))]
+      [else '()])))
+
+(define (sub-range-binder-leaf? v)
```
bennn: Can this be shared with Check Syntax somehow?

pkgs/racket-doc/syntax/scribblings/format.scrbl
lexi-lambda: I don’t understand your comment… `identifier?` is already the first predicate in the `or/c`.

pkgs/racket-doc/syntax/scribblings/format.scrbl
lexi-lambda: I think `#:props` is a better name, and it’s what `format-id` already uses, so I’d prefer to keep it. I don’t think there will be any confusion due to the missing “s”, but I could clarify the wording somewhat if you feel it’s important.

rmculpepper:
I'd much prefer `id-append` (or maybe `id+` if you really value brevity) over `~id`. The rationale behind `~a` was that it converted values like "~a" in a format string (and in particular, it produced a string).

samth:
Is there a reason not to just change `format-id` to add the appropriate properties? That seems like it would be (a) simpler to learn and (b) would improve a bunch of existing programs.

lexi-lambda:
@bennn I have pushed a change that simplifies the handling of both `#:props` and `#:track`. The original design was intended to make it easier for users to create wrappers around `~id` and `~id/1` and just pass the default value along (which is why `~id/1` uses `'infer` as the default argument, for example), but after reading your comments and thinking things over, the enormous API complexity they added was not worth it. Now they just have default values that do the obvious thing.

@rmculpepper Good point. Maybe `id-append` is a good name? I’m not sure it’s quite right, but I can’t think of anything better.

@samth I considered it, and I started there. The tricky thing is that getting everything to line up just right automatically requires the sort of strange defaults that `~id` uses: you want to copy the source location so that jumping to a definition in a different file works right (since that doesn’t use the `'sub-range-binders` information), but you *don’t* want to copy properties because that makes the new id `syntax-original?` and Check Syntax will treat it like an ordinary identifier for other purposes. I couldn’t figure out how to make a good API that was backwards compatible, so I gave up. If you think there’s a change in a good “sweet spot” that could be made to `format-id`, though, I’m happy to implement it.

pkgs/racket-doc/syntax/scribblings/format.scrbl
bennn: Oops

jackfirth:
Should `format-id` just be deprecated then? Is there any good reason to use it instead of `~id`?

samth:
I guess I don't see why the backwards compatible solution is bad. It seems like just adding `'sub-range-binders` would be backwards compatible, and you could drop them if they would be added by the `props` argument. Obviously that wouldn't automatically add the source location because `format-id` does that differently (maybe the `src` argument should have defaulted to `lctx`) but that seems less bad than creating a new API.

lexi-lambda:
@samth It really isn’t about dropping an old `'sub-range-binders` property, it’s about originalness. If properties are copied over, then the new id will be `syntax-original?`, and this makes DrRacket do nonsense:

Screen Shot 2019-04-29 at 13 52 42

I don’t think there is a way to magically infer what a user of `format-id` hoped would happen well enough to make it do this automatically without causing nonsense like the above.

samth:
I'm still confused. What I suggested was that `format-id` *not* copy properties (unless you supply the argument that does that). Instead, you just change `format-id` to do the `'sub-range-binders` thing which is the crucial innovation here.

rmculpepper:
@samth, @lexi-lambda I've made a branch (rmculpepper/racket@b-format-id-subrange) that adds a `#:subs?` argument to `format-id` that computes and adds a `'sub-range-binders` property. It's currently off by default, but it might make sense to default to on. It seems reasonable to make it override any existing `'sub-range-binders` property transferred by a `#:props` argument. I haven't made it try to adjust and transfer sub-range-binders from component identifiers; I'm not sure about the rationale for that behavior in this PR.

I ran into an issue when testing it, however, that I think either means the protocol with Check Syntax needs to change or we might want to rethink the idea of fully automatic support for sub-range binders. Here's the issue. The first test I wrote was similar to the `define/hyphen` example in the docs, but instead of a definition macro, I made a let-like macro. And it didn't work, and as far as I can tell it can't possibly work, because the sub-range binder records are produced before the `let` adds the scope for the binding of the synthesized identifier, so the records can't possibly have a "key" that is `free-identifier=?` to the binder in the expanded program. (This is yet another instance of the problem that syntax property values are invisible to the hygiene algorithm.)

I've thought briefly about changing Check Syntax's interpretation of sub-range-binder records. My current half-baked idea is to allow a symbol instead of an identifier in the first element of the vector, and interpret that as applying to the identifier the property is attached to, as long as the identifier's symbolic name matches the symbol in the property. (Requiring that the symbol matches would protect somewhat against bad syntax property transfers.) Thoughts? (@rfindler too)

lexi-lambda:
@samth Ah, I see. My bet had been that wouldn’t help very much, since I figured most people who use `format-id` specified `#:props` to at least let DrRacket draw *something* (since it will be `syntax-original?`). However, a quick grep through my installed packages indicates that you’re totally right: people use `format-id` all the time, but not *one* of my installed packages (or anything in collects) pass `#:props`. So you’re probably right: `format-id` could be changed to do something useful even for existing idiomatic uses of `format-id`.

lexi-lambda:
@rmculpepper Oops, I submitted my comment just a couple moments after you wrote yours, and I hadn’t seen it. I’ve bumped into the issue with the interaction between `'sub-range-binders` and local binding forms, too, but I don’t know what the right fix is.

It’s really a bigger problem than this PR is really about, but I don’t know how to fix it. It is really a part of a much broader issue, which is that we have a bunch of protocols that stuff syntax objects in syntax properties, but the expander doesn’t recur into syntax property values when it does various recursive operations (scope manipulation and MPI shifting are the obvious ones that come to my mind). I think a nicer fix would be to have properties that the expander *does* recur into, which would fix #1495 as well as various Check Syntax bugs, but that’s a pretty heavy hammer, so I’m reluctant to suggest that change without a lot of thought.

lexi-lambda:
I’m going to close this for now, since it sounds like some changes need to be made, but I’m not 100% sure what they are yet, and I don’t think I’ll get to figuring it out terribly soon. I hereby release all the code in this PR into the public domain, so if anyone wishes to improve it and re-open it with any changes at all, they may do so.

PR #2628, opened 2019-04-27
file/tar: add #:timestamp parameter

bennn:
Add a keyword argument, same as the one for zip

Fixes an "unbound" id in the docs

- - - 

I think this was supposed to be in: https://github.com/racket/racket/commit/ce6b8b8d5c8

But, instead, we could change the docs to match the current impl. of `tar`


mflatt:
Looks good to me.

PR #2629, opened 2019-04-27
Remove contract for define/generic

sorawee:
`local-id` and `method-id` don't have a syntax identifier as a value, so
a contract is inappropriate here.


sorawee:
@mflatt is this good to merge?

mflatt:
Yes

PR #2633, opened 2019-04-30
fix documentation for custom-write-accessor mode

AlexKnauth:
The `write-proc` mode argument can be `#t`, `#f`, `0`, or `1`. The 0 and 1 represent print mode and quote depth.

The documentation for [`gen:custom-write`](https://docs.racket-lang.org/reference/Printer_Extension.html#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._gen~3acustom-write%29%29) specifies that the `write-proc`

> takes three arguments: the structure to be printed, the target port, and an argument that is `#t` for write mode, `#f` for display mode, or `0` or `1` indicating the current quoting depth for print mode.

The `0` and `1` are missing in the documentation for `custom-write-accessor`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/custom-write.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/custom-write.scrbl
@@ -89,7 +89,7 @@ This function is often used in conjunction with @racket[make-constructor-style-p
        (make-constructor-style-printer
         (lambda (obj) 'point)
         (lambda (obj) (list (point-x obj) (point-y obj)))))]))
- 
+
   (print (point 1 2))
 
   (write (point 1 2))]
@@ -108,7 +108,7 @@ property, @racket[#f] otherwise.}
 
 
 @defproc[(custom-write-accessor [v custom-write?])
-         (custom-write? output-port? (or/c boolean? (integer-in 0 1)) . -> . any)]{
+         (custom-write? output-port? (or/c #t #f 0 1) . -> . any)]{
 
 Returns the custom-write procedure associated with @racket[v].}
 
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/custom-write.scrbl
```diff
@@ -108,7 +108,7 @@ property, @racket[#f] otherwise.}
 
 
 @defproc[(custom-write-accessor [v custom-write?])
-         (custom-write? output-port? boolean? . -> . any)]{
+         (custom-write? output-port? (or/c boolean? (integer-in 0 1)) . -> . any)]{
```
bennn: I'd rather `0 1` than `(integer-in ....)`. It's shorter, and then there's no need to think about what kind of interval `integer-in` gives.

I'd also prefer `#f #t` over `boolean?` but that's less likely to be confusing.

AlexKnauth: `(or/c boolean? 0 1)` includes `0.0` and `1.0`, while `(or/c boolean? (integer-in 0 1))` requires `0` and `1` to be exact.

bennn: I still prefer `0` and `1` for the docs, whether or not the implementation allows `0.0` and `1.1`

AlexKnauth: I also tried `(one-of/c #t #f 0 1)`, which I expected to require exact `0` and `1` since the documentation says it uses `eqv?` as the comparison predicate. But that contract also let inexact `0.0` and `1.0` through. That might be a mistake in the implementation of `one-of/c`.

AlexKnauth: Okay, I've changed it to use `(or/c #t #f 0 1)` for the documentation, even though it's less accurate.

AlexKnauth:
Using `(or/c #t #f 0 1)` in the documentation for `custom-write` should be just as fine as using `(or/c 0 1)` in the documentation and error messages for `print`.

The `or/c` might not be strictly accurate, but it gets the idea across well enough and is consistent with existing code and existing documentation.

Should I merge this?

PR #2635, opened 2019-04-30
rktio: Make sure not to close subprocess standard in/out/err

### LINKED ISSUE #2634: Cannot write to subprocess stdin after closing Racket’s stdin
At least on macOS 10.13.6, closing the Racket process’s stdin seems to break sending data to subprocesses on stdin. This program reproduces the behavior:

```racket
#lang racket
(close-input-port (current-input-port))
(match-define (list out in _ err _)
  (process* (find-executable-path "cat")))
(displayln "hello" in)
(close-output-port in)
(port->string out #:close? #t)
(port->string err #:close? #t)
```
```
$ racket close-stdin.rkt
""
"cat: stdin: Bad file descriptor\n"
```

If I remove the `(close-input-port (current-input-port))` line, the program produces this output, instead (which is the output I expect):

```
"hello\n"
""
```

Some information about my environment:

```
$ sw_vers
ProductName:	Mac OS X
ProductVersion:	10.13.6
BuildVersion:	17G65
$ racket --version
Welcome to Racket v7.3.0.3.
```

lexi-lambda:
Fixes #2634.

I originally wrote a slightly more conservative change than the one in this PR, which just added the following guards:

```patch
diff --git a/racket/src/rktio/rktio_process.c b/racket/src/rktio/rktio_process.c
index f5c6fb616d..3007959863 100644
--- a/racket/src/rktio/rktio_process.c
+++ b/racket/src/rktio/rktio_process.c
@@ -1458,17 +1458,23 @@ rktio_process_result_t *rktio_process(rktio_t *rktio,
 
 	/* Close unwanted descriptors */
 	if (!stdin_fd) {
-	  rktio_reliably_close(to_subprocess[0]);
+	  if (to_subprocess[0] != 0) {
+	    rktio_reliably_close(to_subprocess[0]);
+	  }
 	  rktio_reliably_close(to_subprocess[1]);
 	}
 	if (!stdout_fd) {
 	  rktio_reliably_close(from_subprocess[0]);
-	  rktio_reliably_close(from_subprocess[1]);
+	  if (from_subprocess[1] != 1) {
+	    rktio_reliably_close(from_subprocess[1]);
+	  }
 	}
 	if (!stderr_fd) {
           if (!stderr_is_stdout) {
             rktio_reliably_close(err_subprocess[0]);
-            rktio_reliably_close(err_subprocess[1]);
+	    if (err_subprocess[1] != 2) {
+	      rktio_reliably_close(err_subprocess[1]);
+	    }
           }
 	}
```

While that worked, I realized afterward that the whole business seemed unnecessary given the call to `rktio_close_fds_after_fork` immediately after. So I tried this patch, instead, which seems to work, but maybe there’s something important about the existing way of doing things that would make the more conservative change a better one.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/subprocess.rktl
+++ b/pkgs/racket-test-core/tests/racket/subprocess.rktl
@@ -569,6 +569,37 @@
     (delete-directory/files dir)))
 
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Make sure that closing file descriptors in the parent process doesn't
+;; cause problems
+
+(let ()
+  (define out (open-output-string))
+  (define err (open-output-string))
+  (test #t 'subprocess-closed-stdin
+        (parameterize ([current-output-port out]
+                       [current-error-port err])
+          (system* self "-e"
+                   (format "~s" '(let ()
+                                   (define self (vector-ref (current-command-line-arguments) 0))
+                                   (close-input-port (current-input-port))
+                                   (define-values (subp out in err)
+                                     (subprocess #f #f #f self "-e"
+                                                 (format "~s" '(begin
+                                                                 (display (read-line))
+                                                                 (display (read-line)
+                                                                          (current-error-port))))))
+                                   (displayln "hello" in)
+                                   (displayln "goodbye" in)
+                                   (close-output-port in)
+                                   (copy-port out (current-output-port))
+                                   (copy-port err (current-error-port))
+                                   (subprocess-wait subp)
+                                   (exit (subprocess-status subp))))
+                   "--" self)))
+  (test "hello" get-output-string out)
+  (test "goodbye" get-output-string err))
+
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks for investigating! I like the more conservative approach, because `rktio_close_fds_after_fork` seems like a hack, and I don't entirely trust the system calls that are supposed to report an fd maximum (just because they have gone wrong for me in the past, although I'm not aware of any issues right now).

lexi-lambda:
Makes sense to me. I’ll push the more conservative change.

lexi-lambda:
@mflatt Unfortunately, just before merging this, I realized the change was wrong. It made the assumption that if any of the new file descriptors were file descriptors 0, 1, or 2, then those numbers would be used for standard in, standard out, and standard error, respectively, which is completely bogus. It also didn’t accommodate the case where multiple streams share the same file descriptor in the parent process.

I’ve pushed a significantly larger change that handles all those things, as well as a slightly modified test case that exposed at least some of them. (Certainty of testing everything is tricky, especially from the Racket layer, since the behavior depends on exactly which file descriptors the OS happens to decide to use, but the test cases that exist seemed to catch a couple of mistakes I made while making this new implementation pretty well.)

mflatt:
Thanks! If there's still any problem here, it's also beyond what I can see. :)

PR #2638, opened 2019-05-01
assf also consumes a list of pairs

sorawee:


AlexKnauth:
Should this change the result contract as well?

sorawee:
Thanks. Addressed the issue in the latest commit.

PR #2639, opened 2019-05-02
contract: two small changes

bennn:
- move a comment
- simplify a helper function


rfindler:
Should `fast-append` just check `l2` and, if it isn't null, simply call `append`?

rfindler:
Also, if I'm not mistaken, the change you made is one that the optimizer will undo, turning the code back into the version that takes two arguments.

rfindler:
(well, maybe not chez)

bennn:
Calling `append` would definitely be easier to read

PR #2642, opened 2019-05-05
racket/contract: Fix (box/c #:immutable #f) failing on mutable boxes

LLazarek:
Example failure:
```
(define/contract _
  (box/c any/c
         #:immutable #f)
  (box 1))
;; =>
; _: broke its own contract;
; promised a mutable box
;  produced: '#&1
;  in: (box/c any/c #:immutable #f)
;  contract from: (definition _)
;  blaming: (definition _)
```

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/box.rkt
+++ b/pkgs/racket-test/tests/racket/contract/box.rkt
@@ -173,4 +173,16 @@
                      (box (list values))
                      'pos 
                      'neg)])
-      ((car (unbox f)) 3))))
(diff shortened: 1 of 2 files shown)


lexi-lambda:
Maybe add a test that ensures that `(box/c any/c #:immutable #f)` actually rejects immutable boxes? Otherwise, this LGTM.

rfindler:
I apologize for being slow here (I was having a email conversation with Lukas, however).

But please do not merge things in the contract system without asking me, ok?

PR #2648, opened 2019-05-09
Enable MAXLOAD option in root Makefile

pmatos:
MAXLOAD translates to the --load-average (-l) option in make.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -49,14 +49,15 @@ PLT_SETUP_OPTIONS =
 SRC_CATALOG = $(DEFAULT_SRC_CATALOG)
 
 CPUS = 
+MAXLOAD = "-1.0"
 
 in-place:
 	if [ "$(CPUS)" = "" ] ; \
          then $(MAKE) plain-in-place PKGS="$(PKGS)" ; \
-         else $(MAKE) cpus-in-place CPUS="$(CPUS)" PKGS="$(PKGS)" ; fi
+         else $(MAKE) cpus-in-place CPUS="$(CPUS)" MAXLOAD="$(MAXLOAD)" PKGS="$(PKGS)" ; fi
 
 cpus-in-place:
-	$(MAKE) -j $(CPUS) plain-in-place JOB_OPTIONS="-j $(CPUS)" PKGS="$(PKGS)"
+	$(MAKE) -j $(CPUS) --load-average=$(MAXLOAD) plain-in-place JOB_OPTIONS="-j $(CPUS) --load-average=$(MAXLOAD)" PKGS="$(PKGS)"
 
 # Explicitly propagate variables for non-GNU `make's:
 LIBSETUP = -N raco -l- raco setup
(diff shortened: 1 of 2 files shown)


pmatos:
I am slightly concerned about support in other platforms as this is definitely supposed to work with GNU make on unix systems. If you could test with your favourite platform, that would be great.

pmatos:
The motivation to implement this is related to CI. Although we have `CPUS`, that's not enough when you might have many other jobs but don't know how many. In a CI setting you have a fixed number of CPUS but many jobs and you want to be using all of the CPUS without getting your load too high. So you might want to say, all jobs should use all CPUS but the load average should be higher than the number of CPUS. You can now do that with `make CPUS=16 MAXLOAD=16`. If you have 10 jobs, without `MAXLOAD` you can potentially get 160 processes, but with MAXLOAD, `make` will only spawn a process if the load average is lower than `MAXLOAD`.

Makefile
```diff
@@ -49,14 +49,15 @@ PLT_SETUP_OPTIONS =
 SRC_CATALOG = $(DEFAULT_SRC_CATALOG)
 
 CPUS = 
+MAXLOAD = "-1.0"
 
 in-place:
 	if [ "$(CPUS)" = "" ] ; \
```
samth: Should this test for the `MAXLOAD` variable? Or if not, the INSTALL.txt documentation should say that `MAXLOAD` does nothing without `CPUS`.

mflatt:
I think it's not going to work to have a gmake-specific flag in the makefile. Does it not work to just use `--load-average` directly, like this?
```
make --load-average=12 CPUS=8
```

Makefile
pmatos: `MAXLOAD=-1.0` is the default i.e. no maximum - I set it like this so I don't need to test it here.

pmatos:
But with `nmake` (this is the microsoft make, right?) it should be a problem because I don't touch the windows rules - or I tried not to.

Makefile
samth: My point is that `make MAXLOAD=12` will not take the true branch of that conditional, and thus will not pass `--load-average` at all.

pmatos:
Oh no, the CI is failing because the `--load-average` is being passed to `raco`, somehow I missed that. Will have to revisit the issue then.

pmatos:
> I think it's not going to work to have a gmake-specific flag in the makefile. Does it not work to just use `--load-average` directly, like this?
> 
> ```
> make --load-average=12 CPUS=8
> ```

The problem here is that the `--load-average` won't make it down the sub-make calls - untested assertion.

Makefile
pmatos: Ah, right. Well spotted. OK, thanks for the comments. Will need to revisit the patch.

mflatt:
I'm pretty sure that recursive calls to `make` will see the `--load-average` argument. (I think the mechanism is that `make` sets and uses a `MAKEFLAGS` environment variable.) The propagation of a `-j` argument similarly relies on that effect.

pmatos:
> I'm pretty sure that recursive calls to `make` will see the `--load-average` argument. (I think the mechanism is that `make` sets and uses a `MAKEFLAGS` environment variable.) The propagation of a `-j` argument similarly relies on that effect.

I am testing that now. I though the whole `CPUS` was because that effect did not occur. Or is it just so yu can use it for `raco` calls?

pmatos:
I think you are right:
https://www.gnu.org/software/make/manual/html_node/Options_002fRecursion.html

mflatt:
Right – `CPUS` is just so the same number can be passed to `raco`.

pmatos:
I was hopeful my experiment would show this works but I am not 100% convinced. 
Looking at the code I am actually concerned about these lines in several of the configure files: `MAKEFLAGS=`, which removes I think the options initially passed to `make`. I will need to investigate this for a little longer.

pmatos:
Closing for now. If a patch is needed I will open a new PR. Thank you for your time.

PR #2650, opened 2019-05-10
Add configure option --enable-generations (enabled by default).

pmatos:
This option allows the user to enable or disable (with
--disable-generations or --enable-generations=no) generations in
gc2. This is, in most cases, a bad idea.

### DIFF
--- a/racket/src/cfg-racket
+++ b/racket/src/cfg-racket
@@ -828,6 +828,7 @@ enable_csdefault
 enable_cgcdefault
 enable_sgc
 enable_sgcdebug
+enable_generations
 enable_backtrace
 enable_cify
 enable_pthread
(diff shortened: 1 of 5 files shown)


mflatt:
This looks like the right idea. Before merging, though, I'd like to see how much difference it makes to further adjust the GC for single-generation mode, and I should be able to get to that today.

pmatos:
This just failed on AppVeyor... 
```
c:\projects\racket\racket\src\racket\gc2\newgc.c(772) : error C2065: 'USE_GC_GENS' : undeclared identifier
failed compile
```

Doesn't fail on Linux. Interesting...

mflatt:
There's a "mzconfig.h" in "racket/src/worksp" that needs to be manually updated (since we don't try to use `configure` on Windows).

pmatos:
ah, thanks.

pmatos:
OK, updated. This is an untested commit as I don't have a windows machine at hand, hopefully CI will check all is ok.

pmatos:
Argh, of course. Because we don't have configure or preprocessing of this header we need to hardcode the correct value for `USE_GC_GENS`.

mflatt:
Thanks! I'll merge this (squashed) with a few small changes:

 * Adjust `configure` help to fit in 80 columns

 * Add a `show_explicitly_disabled`, which compensates some for the lack of argument checking by confirming when the flag is spelled correctly

 * Drop signal-handling code in the GC when it's not going to be used

 * Adjust the commit message to explain when disabling generational GC might be a good idea

Finally, I'll add another commit that adjusts the GC heuristic to better fit no-generations mode. DrRacket is more usable that way, for example.

pmatos:
Thanks. I will close the PR.

PR #2651, opened 2019-05-13
Fix instructions to run benchmarks

pmatos:
According to https://github.com/racket/racket/blob/842769c501c3d904b0c258ee4f177d4d9d837b3a/pkgs/racket-benchmarks/tests/racket/benchmarks/shootout/run.rkt#L140 the option might now be `typed-racket-non-optimizing`. Can you confirm and merge / delete branch if ok?


pmatos:
Thanks

PR #2653, opened 2019-05-14
Remove test causing left shift overflow

pmatos:
Doing 1 << 31 and 1 << 63 on 32 and 64 bit archs respectively causes lshift overflow reported by ubsan.

I would like to open a discussion on this specific test. I am convinced, but not certain, that this is actually undefined behaviour and therefore the test should not rely on a zero result.

The respective code in `number.c` triggers ubsan:
```
src/number.c:5501:11: runtime error: left shift of 1 by 63 places cannot be represented in type 'long int'
```

I looked at `fold_fixnum_bitwise_shift` and this triggers the signal `scheme_signal_error("unsafe-fxlshift: shift is too large");`. However, we still go ahead and do the shift inside: `v = v op SCHEME_INT_VAL(argv[i]);` which is generated by the macro `UNSAFE_FX`. I understand from the comments this was added by @mflatt a year ago to test folding but this specific test shouldn't rely on an undef. behaviour since on other platforms the result might be non-zero. Maybe there's a better test for the constant folding that can replace this one instead?

### DIFF
--- a/racket/src/racket/src/number.c
+++ b/racket/src/racket/src/number.c
@@ -5484,7 +5484,7 @@ static Scheme_Object *fold_fixnum_bitwise_shift(int argc, Scheme_Object *argv[])
 #define UNSAFE_FX(name, op, fold, type, no_args)             \
  static Scheme_Object *name(int argc, Scheme_Object *argv[]) \
  {                                                           \
-   intptr_t v;                                               \
+   type v;                                                   \
    int i;                                                    \
    if (!argc) return no_args;                                \
    if (scheme_current_thread->constant_folding) return fold(argc, argv);     \


mflatt:
I think the problem here is that `intptr_t v` in the `UNSAFE_FX` is should be `type v`. Then, the calculation uses an unsigned shift, which would be specified.

The test is still questionable in that it's making claims that are not guaranteed by the documentation. But I don't have a better idea for performing the intended check.

gus-massa:
I made a similar proposal in https://github.com/racket/racket/pull/2328 that was never merged.

I think this should be changed, but I propose more changes:

* `(unsafe-fxlshift 1 62)` is safe from the C point of view, but it is not safe from the Racket point of view. It is weird, because it is a negative fixnum.

* It is important to test that if the code never runs, the constant-folding doesn't raise an error.

* The fixnum in RacketCS has less bits. Perhaps `fixnum-width` should be a Racket primitive instead of a big `cond`.

mflatt:
I had forgotten about  #2328 and probably never looked at it closely. Changing the test to just making sure that an `unsafe-flshift` call compiles ok as in #2328 is probably good enough. Meanwhile, it's probably a good idea to also fix `UNSAFE_FX` to match the intended use of `type`.

pmatos:
I have reinstated the test and used `type` in `UNSAFE_FX` for `v`. In any case, I should look at #2328 changes by @gus-massa 
Apologies @gus-massa but I completely forgot that work you did.

gus-massa:
I think it's better to reorganize the commits in one that only fix the problem in numbers.c and one that fix the whitespace/indentation errors in unsafe.rktl (IIUC the net effect after squashing is only in the superfluous spaces, there are no changes in the tests.). This makes more easy to understand the history an use blame a few months/years later.

About https://github.com/racket/racket/pull/2328 : sometimes PR get lost in the pile, and it's good to refresh it. Let's hope that  https://github.com/racket/racket/pull/2328 or a variant is merged.

pmatos:
@gus-massa yes, my editor automatically removes the spaces and I didn't notice it in the commit. I will filter that out before merging.

pmatos:
This seems to fix the ubsan problem we are seeing. I removed the spaces issue mentioned by @gus-massa from PR. @mflatt ok to merge?

pmatos:
By the way, @gus-massa I noticed you were adding ubsan checks to travis but we have it running in gitlab https://gitlab.com/racket/racket

pmatos:
@mflatt ping

Issue #1721, opened 2017-06-14
Quadratic work done by `(for/list ... (in-producer ... ) ... )`


kalbr:
I've stumbled on a performance bug while running code from an old mailing list [thread](https://groups.google.com/d/msg/racket-users/8955N0wivE8/OTxbYY5CV9wJ) comparing alternatives to using generators. It was introduced in commit 5e94a906cd5935.

I've attached a more thorough benchmark script with results, but the short of it is that the function produced by
```racket
(define (make-real-generator N)
  (generator
   ()
   (for ([i (in-range N)])
     (yield i))))
```
used in the form
```racket
(length (for/list ([x (in-producer (make-real-generator 40000) void?)])
               x))
```
takes around 20 seconds from commit 5e94a906cd5935 onwards but took around 0.4s beforehand. 
The `for/vector` form is unaffected and had similar runtime to `for/list` beforehand. The slowdown looks (vaguely) quadratic:
```
#<procedure:make-real-generator> 
with for/vector
N=40000 cpu time: 359 real time: 363 gc time: 21
N=20000 cpu time: 170 real time: 170 gc time: 8
N=10000 cpu time: 85 real time: 85 gc time: 3
with for/list
N=40000 cpu time: 19545 real time: 19610 gc time: 4538
N=20000 cpu time: 4168 real time: 4178 gc time: 628
N=10000 cpu time: 1001 real time: 1003 gc time: 68
10000
```

The code that produced the above snippet is attached: [quadratic-for-list.test.rkt.txt](https://github.com/racket/racket/files/1073888/quadratic-for-list.test.rkt.txt)

The results:
[after.txt](https://github.com/racket/racket/files/1073870/after.txt)
[before.txt](https://github.com/racket/racket/files/1073871/before.txt)

Apologies for the ugly formatting!

mflatt:
Thanks for the report!

The root problem here is Racket's implementation of continuations, where continuation capture or restore can take O(N) time for a continuation of size N. Each access of the generator is working in a longer continuation.

The short-term solution is probably to change the `for[*]/list` expansion back. The long-term solution is probably a better implementation of continuations.

PR #2656, opened 2019-05-16
Add `for[*]/foldr[/derived]` and reimplement `for[*]/stream` using it

lexi-lambda:
`for/fold` is a left fold, which is normally what you want in a call-by-value language such as Racket, but it makes efficient lazy iteration difficult. This commit adds a new `for/foldr` iteration form (along with `for*/` and `/derived` variants) that provides a right fold operation that offers complete control over precisely how lazy the iteration ought to be.

In simple microbenchmarks, reimplementing `for/stream` using `for/foldr` instead of `for` plus a generator can be almost 40x faster on large streams, and it can even be slightly faster to build ordinary lists. Running [this tiny benchmark](https://gist.github.com/lexi-lambda/f34b3f75ab62fb0cd145a5b7cd909b3b) on my machine yields the following results:

bench-for-foldr-full

As you can see, the old `for/stream` implementation is horribly slow on large streams, so let’s zoom in a little to get a better comparison:

bench-for-foldr-clamped

Overall, the performance of loops using `for/foldr` seem about on par with handwritten loops using named `let`. Even `for/list` is slower than `for/foldr`… but I think that’s just because `for/list` builds up the list using tail recursion and `reverse`s it at the end, and that needs to stay the way it is due to the issue Matthew mentioned in #1721. In any case, it’s a small difference, and I only bring it up to make clear that `for/foldr` is fast!


lexi-lambda:
@mflatt Thanks for taking a look. Your comment about the flattener change makes sense… I wasn’t really sure how it worked, but I have taken some time today to try to understand better what’s going on. I’ve pushed a different change now, which is hopefully closer to the right one.

mflatt:
This looks great – thanks!

PR #2657, opened 2019-05-16
Trim spaces

pmatos:
It's a non-functional commit.


pmatos:
Follow-up to #2653

pmatos:
@mflatt ping

PR #2665, opened 2019-05-23
Remove Boehm GC

pmatos:
I am trying to create a branch that makes life easier for @mflatt to remove Boehm GC. It's not ready for merging, which is why I am creating it as a draft PR. There is quite a bit of testing to do and some other things to fix because they still expect Boehm to be there.


jeapostrophe:
Doesn't 3m require cgc to build the tool that modifies the C code with the precision annotations? And even with Chez, I believe that we require a Racket implementation of Chez to do the bootstrap to get the Racket compiler. So in each path, it seems like we need some sort of Racket. I suppose we could have a GC-less Racket, but I'm skeptical that it wouldn't use an absurd amount of memory. Maybe I'm just wrong and confused?

pmatos:
That's right. we can use sgc, I am not removing that. I am removing the first gc : boehm gc.

jeapostrophe:
Duh, thank you for fixing me.

jackfirth:
My *goodness* that's an impressive delta

image

pmatos:
@mflatt see what you think when you get the time. It looks ok on my side on Linux. CI says it's fine on Windows but unsure on MacOS. I might have missed removing some windows/macos specific files.

pmatos:
Thanks for the pointers. I will do some cleaning and come back to you so we can do some further testing.

pmatos:
Should've tried `grep -Ri boehm` at project root. There's still quite a bit of cleanup/work to do.

mflatt:
Not sure how much you've looked at MSVC projects and solutions before. In case you decide to look at that, I think it's a matter of removing configurations form some of the ".vcproj" and ".vcxproj" files and also removing references to those configurations in ".sln" files.

pmatos:
I will be looking at the project files today. In the meantime I saw the comment:
```
/* Racket creates Windows threads for various purposes, including
   non-blocking FILE reads. Unfortunately, these threads can confuse
   the Boehm GC if they move virtual pages around while its marking. So we
   remember each created thread and suspend it during GC.

   This work is not necessary if GC_use_registered_statics is set. */
```
in `port.c`, I assume that this affects SGC as well, correct? Therefore the appropriate thing to do here is to replace the reference of `Boehm GC` by `SGC`.

mflatt:
That comment is specific to the Boehm GC. SCG always runs in the mode analogous to having `GC_use_registered_statics` set.

pmatos:
We are slowly getting there... :)

pmatos:
Specifically you might want to take a look at my change: 
https://github.com/racket/racket/pull/2665/commits/1ac7a9112131471bd2a743828b447044e21944ee
There's no occurrence in the code of `GC_use_registered_statics`. If SenoraGC and 3m don't have the same problem as Boehm with regards to the Windows threads mentioned in the comment, we could remove this whole remember threads stuff.

pmatos:
It seems Azure found some problems with the Windows build. I will take a look.

pmatos:
This is my last open PR, so lets get this sorted... :)

pmatos:
> I plan to merge this as part of the other changes tomorrow (and I have a set of commits staged, so please don't merge before).

Sure - thanks.

pmatos:
Finally merged - thanks @mflatt for seeing this through. Closing.

PR #2666, opened 2019-05-27
cs: use print-[m]pair-curly-braces

gus-massa:
The first commit adds the use of `print-mpair-curly-braces` to the cs version. This should fix most of the errors in racket-r7rs onver racketcs. (cc: @lexi-lambda) More details in [travis/jobs/537575250](https://travis-ci.org/gus-massa/racket-r7rs/jobs/537575250#L662) I think that only [travis/jobs/537575250#L641](https://travis-ci.org/gus-massa/racket-r7rs/jobs/537575250#L641) will remain.

The second commit is for `print-pair-curly-braces`. I'm not 100% happy with it because it uses a  complicated way to select the curly braces for structs. Also, I'm using struct in the test of other values to force the use of the constructor, for example in `(list #0={s} #0#)`. I'd be nice to use other way to force the constructor because there are too many parameters that control how structs are shown, for example `print-vector-length`.

Also, it's necessary to add support for `print-vector-length` but I think it's better to separate the code for `vector`s/`fxvector`s/`flvector`s in it's own file instead of using the version for `list`s. So, I left it for another time.


racket/src/io/print/mlist.rkt
```diff
@@ -1,22 +1,26 @@
...
+         "config.rkt")
 
 (provide print-mlist)
 
 (define (print-mlist p who v mode o max-length graph config)
   (define unquoted? (eq? mode PRINT-MODE/UNQUOTED))
-  (let ([max-length
+  (define curly? (and (not unquoted?)
+                      (config-get config print-mpair-curly-braces)))
+  (let ([max-lenght
```
mflatt: "th" got transposed here

mflatt:
Except for the transposition typo, LGTM.

gus-massa:
Before merging, I changed the way `curly?` was calculated in [print/list.rkt#L16-L24](https://github.com/racket/racket/blob/master/racket/src/io/print/list.rkt#L16-L24) because the previous version was breaking some test.

PR #2667, opened 2019-05-28
Say "GRacketCS" rather than "GracketCS"

jessealama:
On case-sensitive filesystems on macOS, these are
distinct (leads to a file not found error). On
case-insensitive systems, the change should not matter.


mflatt:
Thanks!

Issue #2344, opened 2018-10-29
[racketcs] `make cs` does not work on FreeBSD


abmclin:
Firstly, is it planned for Racket-on-Chez to be supported on BSD family operating systems? 

If so here's a summary of all the issues I've found. I took `make cs` as far as I was able with my level of experience. It appears that the issues I ran into seem to be only arising from ChezScheme.

I'm running GhostBSD which is based on 64-bit FreeBSD kernel 11.1.

The first issue encountered is missing boot files for `ta6fb` architecture. As explained in https://github.com/cisco/ChezScheme/issues/7 BSD boot files are not built by default. Following the recommended method in the issue comments, I cross-compiled and copied the missing boot files to my FreeBSD machine which allows make to continue.

Second issue is ChezScheme's makefiles are only compatible with GNU make, to work around that I aliased `make` with `gmake`

The third issue and this is as far as I got, is a compilation error occurring when compiling `io.sls` an attempt is made to load `libc.so` which fails with invalid file format error. In FreeBSD variants, `libc.so` is a linker script which resolves to `libc.so.7` Apparently the linker doesn't know how to handle the FreeBSD case.

In case this is relevant, I see some code in `ChezScheme/mats/foreign.ms` which attempts to deal with loading `libc.so` on FreeBSD machines.

If RacketCS won't be supported on BSD family in the foreseeable future - which I hope won't be the case, then feel free to close this issue. Though I will continue to advocate for BSD support :-)

mflatt:
Does b9c042329bfa45bf0c5d08fc75c401b43cac5284 let you use `gmake cs` (as opposed to having to globally redirect `make` to `gmake`), or is there still some other issue with `gmake cs`?

abmclin:
with the latest commits `gmake cs` continues some more but there's a new issue with ld not finding libiconv during a compilation step.

I've attached a file of the last 50 lines of `gmake cs` output. The entire output is a 183 MB file, let me know if you want all of it.
[truncated_out.txt](https://github.com/racket/racket/files/2530232/truncated_out.txt)

abmclin:
As far as I can tell libiconv is installed on my system.

`ldconfig -r | grep libiconv` returns `liconv.2 => /usr/local/lib/libiconv.so.2`

samth:
From that output, it looks like Chez (or maybe Racket) is looking for libiconv without a version. Is there a package you can install that has the unversioned .so file?

abmclin:
[...]

PR #2669, opened 2019-05-30
Restructure makefile variable definition

jessealama:
The existing definition of FWVERSION has the toplevel
structure of an if-then-else, each terminal of which is an
assignment. With BSD makes, this is problematic (tested with
OpenBSD 6.5). To make such makes happe, the definition of
FWVERSION should have a traditional

  VAR = VAL

shape. In this commit, we achieve that by "shelling out",
putting the if-then-else logic into a $(shell ...)
expression.

Tested on OpenBSD 6.5 and macOS Mojave, both with its
built-in make and GNU Make.

### DIFF
--- a/racket/src/racket/version.mak
+++ b/racket/src/racket/version.mak
@@ -4,10 +4,9 @@ FWVERSION_Y = $(shell grep 'define MZSCHEME_VERSION_Y ' $(mainsrcdir)/racket/src
 FWVERSION_Z = $(shell grep 'define MZSCHEME_VERSION_Z ' $(mainsrcdir)/racket/src/schvers.h | cut -d ' ' -f 3)
 FWVERSION_W = $(shell grep 'define MZSCHEME_VERSION_W ' $(mainsrcdir)/racket/src/schvers.h | cut -d ' ' -f 3)
 
-ifneq ($(FWVERSION_W),0)
-FWVERSION = $(FWVERSION_X).$(FWVERSION_Y).$(FWVERSION_Z).$(FWVERSION_W)
-else ifneq ($(FWVERSION_Z),0)
-FWVERSION = $(FWVERSION_X).$(FWVERSION_Y).$(FWVERSION_Z)
-else
-FWVERSION = $(FWVERSION_X).$(FWVERSION_Y)
-endif
+FWVERSION = $(shell \
+if [ $(FWVERSION_W) != 0 ]; \
+then echo $(FWVERSION_X).$(FWVERSION_Y).$(FWVERSION_Z).$(FWVERSION_W); \
+elif [ $(FWVERSION_Z) != 0 ]; then echo $(FWVERSION_X).$(FWVERSION_Y).$(FWVERSION_Z); \
+else echo $(FWVERSION_X).$(FWVERSION_Y); \
+fi)


mflatt:
I'm happy to merge this change, but including "version.mak" is supposed to be enabled by the `configure` script only when using GNU make. Can you tell what goes wrong with the test that's in `configure.ac`?

```
is_gmake=`make -v no-such-target-we-hope 2>&1 | grep "GNU Make"`
if test "$is_gmake" = "" ; then
  ....
else
  ....
  INCLUDEDEP="-include"
fi
```

jessealama:
I was building Racket-on-Chez, so I did a straight `make cs`, without the usual configure step. Perhaps that's my mistake?

mflatt:
Ah, I see that the test is backwards in "cs/c/configure.ac". Let's try fixing that first, and I've pushed a repair.

jessealama:
a4545594f62f62bcbdd84e09b6ecf2299eb07bc4 does the trick. A different error occurs during build, later in the process, but the error that prompted me to open the pull request is gone. Tested on fresh OpenBSD 6.5 and FreeBSD 12.0 installs, with their built-in makes.

jessealama:
After taking a closer look at the status of Chez on FreeBSD and OpenBSD, I think we can close this pull request. It appears that some work is required to get Chez to build on those two platforms. In any case, thanks for identifying the bug with make detection. That, certainly, needed to be fixed.

mflatt:
I was able to build Chez Scheme on OpenBSD 6.3 with the patch at
https://github.com/mflatt/ChezScheme/tree/openbsd
Does that get closer for you?

jessealama:
Which libuuid are you using? I take it the UUID library that comes out-of-the-box in OpenBSD (at least, on 6.5) isn't right. Are you using [this one](http://www.ossp.org/pkg/lib/uuid/)?

mflatt:
No, the e2fsprogs one: http://e2fsprogs.sourceforge.net/

mflatt:
I've updated the racket/ChezScheme repo to build on FreeBSD, OpenBSD, and (because I'm feeling especially reckless today) NetBSD. The build now uses the out-of-the-box OSSP UUID on OpenBSD and FreeBSD and the native UUID on NetBSD.

jessealama:
Awesome, thanks a lot! I'm having trouble building on FreeBSD 12.0. I've installed gmake, libiconv, and ossp-uuid. The build goes quite far (much further than before), but dies at a linker step:

```
[snip tons and stuff]
gmake[10]: Entering directory '/root/racket/racket/src/build/cs/c/rktio'
gmake[10]: 'librktio.a' is up to date.
gmake[10]: Leaving directory '/root/racket/racket/src/build/cs/c/rktio'
gmake[9]: Leaving directory '/root/racket/racket/src/build/cs/c/rktio'
gmake[8]: Leaving directory '/root/racket/racket/src/build/cs/c/rktio'
gmake racketcs
gmake[8]: Entering directory '/root/racket/racket/src/build/cs/c'
cc -g -O2   -pthread -I/root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb -I../../../cs/c/../../rktio -Irktio -I. -L/usr/local/lib -o raw_racketcs main.o boot.o /root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb/kernel.o rktio/librktio.a -pthread -ldl -lm
/usr/bin/ld: error: undefined symbol: uuid_make
>>> referenced by xxhash.c
>>>               /root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb/kernel.o:(S_unique_id)

/usr/bin/ld: error: undefined symbol: uuid_export
>>> referenced by xxhash.c
>>>               /root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb/kernel.o:(S_unique_id)

/usr/bin/ld: error: undefined symbol: uuid_destroy
>>> referenced by xxhash.c
>>>               /root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb/kernel.o:(S_unique_id)

/usr/bin/ld: error: undefined symbol: libiconv_open
>>> referenced by xxhash.c
>>>               /root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb/kernel.o:(s_iconv_open)

/usr/bin/ld: error: undefined symbol: libiconv_close
>>> referenced by xxhash.c
>>>               /root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb/kernel.o:(s_iconv_close)

/usr/bin/ld: error: undefined symbol: libiconv
>>> referenced by xxhash.c
>>>               /root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb/kernel.o:(s_iconv_from_string)

/usr/bin/ld: error: undefined symbol: libiconv
>>> referenced by xxhash.c
>>>               /root/racket/racket/src/build/ChezScheme/ta6fb/boot/ta6fb/kernel.o:(s_iconv_to_string)
cc: error: linker command failed with exit code 1 (use -v to see invocation)
gmake[8]: *** [Makefile:211: raw_racketcs] Error 1
gmake[8]: Leaving directory '/root/racket/racket/src/build/cs/c'
gmake[7]: *** [Makefile:59: cs] Error 2
gmake[7]: Leaving directory '/root/racket/racket/src/build/cs/c'
gmake[6]: *** [Makefile:349: cs-after-racket-with-abs-paths] Error 2
gmake[6]: Leaving directory '/root/racket'
gmake[5]: *** [Makefile:345: cs-after-racket-with-racket-and-scheme-src] Error 2
gmake[5]: Leaving directory '/root/racket'
gmake[4]: *** [Makefile:339: cs-after-racket-with-racket] Error 2
gmake[4]: Leaving directory '/root/racket'
gmake[3]: *** [Makefile:324: racket-configured-then-cs] Error 2
gmake[3]: Leaving directory '/root/racket'
gmake[2]: *** [Makefile:318: racket-then-cs] Error 2
gmake[2]: Leaving directory '/root/racket'
gmake[1]: *** [Makefile:291: plain-cs] Error 2
gmake[1]: Leaving directory '/root/racket'
gmake: *** [Makefile:285: cs] Error 2
```

The solution is probably very simple; I'm surely missing something easy here. I've tried adding `-L/usr/local/lib` (which is where the iconv and uuid libraries are setting), to no avail.

abmclin:
Try adding -liconv option too.

It’s not enough to specify a library search path, need to also specify
which libraries to link against.

https://github.com/racket/racket/issues/2344 has some notes which may be
useful. I’ll take a look later today to see if the issues I noticed in the
above issue ticket has been resolved.

mflatt:
Ah, the latest problem is that I built Chez Scheme but forgot to try linking `racketcs`, which involves repeating some linking information that is embedded in the Chez Scheme makefiles. I'll fix that.

mflatt:
I've pushed an update to the configure script that should fix the linking problem.

jessealama:
That works -- thanks! Just got a build on FreeBSD 12.0 to work without a hitch.

OpenBSD 6.5 is another issue, though. I get this error:

```
[snip tons of stuff]
gcc -g -O2 -Wall -pthread -m64 -fPIC -m64 -DHAVE_HIDDEN  -DPIC -c -o objs/gzread.o gzread.c
gcc -g -O2 -Wall -pthread -m64 -fPIC -m64 -DHAVE_HIDDEN  -DPIC -c -o objs/gzwrite.o gzwrite.c
gcc -shared -Wl,-soname,libz.so.1,--version-script,zlib.map -g -O2 -Wall -pthread -m64 -fPIC -m64 -DHAVE_HIDDEN -o libz.so.1.2.11 adler32.lo crc32.lo deflate.lo infback.lo inffast.lo inflate.lo inftrees.lo trees.lo zutil.lo compress.lo uncompr.lo gzclose.lo gzlib.lo gzread.lo gzwrite.lo  -lc
rm -f libz.so libz.so.1
ln -s libz.so.1.2.11 libz.so
ln -s libz.so.1.2.11 libz.so.1
gcc -g -O2 -Wall -pthread -m64 -m64 -DHAVE_HIDDEN -o examplesh example.o -L. libz.so.1.2.11
gcc -g -O2 -Wall -pthread -m64 -m64 -DHAVE_HIDDEN -o minigzipsh minigzip.o -L. libz.so.1.2.11
(cd ../lz4/lib; CFLAGS="-g -O2 -Wall -pthread -m64" gmake liblz4.a)
compiling static library
ld -r -X -o ../boot/ta6ob/kernel.o statics.o segment.o alloc.o symbol.o intern.o gcwrapper.o gc-ocd.o gc-oce.o number.o schsig.o io.o new-io.o print.o fasl.o vfasl.o stats.o foreign.o prim.o prim5.o flushcache.o schlib.o thread.o expeditor.o scheme.o compress-io.o i3le.o ../zlib/libz.a ../lz4/lib/liblz4.a
ld: error: section type mismatch for .eh_frame
>>> ../lz4/lib/liblz4.a(lz4.o):(.eh_frame): SHT_X86_64_UNWIND
>>> output section .eh_frame: SHT_PROGBITS

ld: error: section type mismatch for .eh_frame
>>> ../lz4/lib/liblz4.a(lz4frame.o):(.eh_frame): SHT_X86_64_UNWIND
>>> output section .eh_frame: SHT_PROGBITS

ld: error: section type mismatch for .eh_frame
>>> ../lz4/lib/liblz4.a(lz4hc.o):(.eh_frame): SHT_X86_64_UNWIND
>>> output section .eh_frame: SHT_PROGBITS

ld: error: section type mismatch for .eh_frame
>>> ../lz4/lib/liblz4.a(xxhash.o):(.eh_frame): SHT_X86_64_UNWIND
>>> output section .eh_frame: SHT_PROGBITS
gmake[13]: *** [Makefile:35: ../boot/ta6ob/kernel.o] Error 1
gmake[12]: *** [Makefile:20: build] Error 2
gmake[11]: *** [Makefile:19: build] Error 2
gmake[11]: Leaving directory '/root/racket/racket/src/build/ChezScheme'
gmake[10]: *** [Makefile:152: scheme-make-finish] Error 2
gmake[10]: Leaving directory '/root/racket/racket/src/build/cs/c'
gmake[9]: *** [Makefile:138: scheme-make-checkout] Error 2
gmake[9]: Leaving directory '/root/racket/racket/src/build/cs/c'
gmake[8]: *** [Makefile:131: scheme] Error 2
gmake[8]: Leaving directory '/root/racket/racket/src/build/cs/c'
gmake[7]: *** [Makefile:56: cs] Error 2
gmake[7]: Leaving directory '/root/racket/racket/src/build/cs/c'
gmake[6]: *** [Makefile:349: cs-after-racket-with-abs-paths] Error 2
gmake[6]: Leaving directory '/root/racket'
gmake[5]: *** [Makefile:345: cs-after-racket-with-racket-and-scheme-src] Error 2
gmake[5]: Leaving directory '/root/racket'
gmake[4]: *** [Makefile:339: cs-after-racket-with-racket] Error 2
gmake[4]: Leaving directory '/root/racket'
gmake[3]: *** [Makefile:324: racket-configured-then-cs] Error 2
gmake[3]: Leaving directory '/root/racket'
gmake[2]: *** [Makefile:318: racket-then-cs] Error 2
gmake[2]: Leaving directory '/root/racket'
gmake[1]: *** [Makefile:291: plain-cs] Error 2
gmake[1]: Leaving directory '/root/racket'
gmake: *** [Makefile:285: cs] Error 2
```

Is this a W^X problem?

abmclin:
I have seen this same error on variants of FreeBSDs which uses Clang as their default system tool chain including the LLVM ld. The problem on those FreeBSDs and I’m guessing the same for your OpenBSD is that when after compiling gcc invokes `ld` it actually invokes the LLVM ld which was installed at `/usr/bin/ld` not the GNU ld which was installed at `/usr/local/bin/ld`

Apparently the LLVM ld doesn’t know how to handle the .eh_frame section type. The GNU ld does handle it fine so if you can make sure the correct `ld` is invoked your build hopefully will succeed this time.

mflatt:
Running `ld` directly often creates problems, and I long ago learned not to do that in Racket makefiles — mostly by not doing unusual things like linking object files into a new object file.

Would it be better to go through the C compiler with `-Wl,-r` and `-nostdlib` as suggested here?

https://stackoverflow.com/questions/993406/is-it-possible-to-create-an-object-file-from-other-object-files-in-gcc?lq=1

As long as the C compiler is GCC or Clang, maybe this would work more consistently.

pmatos:
Interesting. I was today working on a PR to use GCC instead of ld in Chez to do the linking. My reason to do this was because if you use ld directly you cannot compile with LTO which is something I was interested in. If you go ahead with using Gcc instead of ld, great. Otherwise I propose a patch in due time. I will be away from Saturday to Thursday so if not tomorrow, I will get back to this end of next week.

pmatos:
Not sure if it's related to any problems you are seeing but I noticed on Chez (another thing I was preparing a PR for) that the CC flag is not propagated to the zlib compilation so if you get let's say Ubuntu lts which by default has gcc7 and install gcc8 whose binary is gcc-8 and then do a build with `CC=gcc-8` everything is compiled with Gcc 8 except zlib which will use Gcc 7 which might cause problems. (for example if you use LTO, which is how I noticed this)

jessealama:
@pmatos thanks for taking a look into this!

pmatos:
@jessealama I was more speaking from the top of my head. Haven't really looked at this specific issue although it was not for lack of trying. I did try to run freebsd using vagrant but it just won't connect [1]. How are you using freebsd, do you actually have an installation of this OS?

[1] - https://github.com/lavabit/robox/issues/55

jessealama:
@pmatos I've got a FreeBSD 12.0 install on a VPS service. As far as I can tell, it's a standard install. I notice that the Vagrant box you link to is using FreeBSD 11.0. (I'm not sure if that makes any difference, but I thought I'd note it.) If you need GCC/LLVM version information from my instance, let me know.

jessealama:
I'm closing this pull request. The problem I initially pointed out has been fixed. I'll create a follow-on bug report for the OpenBSD build problems.

PR #2679, opened 2019-06-04
Remove time limit on memory test

pmatos:
I have seen intermittent failures in CI because the servers are too busy and running racket on emulation with ubsan makes things worse therefore a test that should be about memory ends up failing because it goes over the time limit. 

My alternative is to patch this during CI before running the test but I would prefer not having to patch racket before running tests so I suggest removing the time constraint. This is not a blocker so if you do not agree, I will patch racket during CI.


pmatos:
An example from last night: https://gitlab.com/racket/racket/-/jobs/224595765

pmatos:
Fair enough. Let me start by attempting a factor of 10 and see if it works.

PR #2685, opened 2019-06-06
More tests for (inexact->exact +inf.0)

gus-massa:
The first commit adds tests where the expression is not in tail position, to check that the optimizer doesn't remove it.

This detect a bug in the signature of `inexact->exact` in the patched version of Chez Scheme, because it is marked as `safeonggodargs`, so when the argument is a `number?` it is replaced by the unsafe version that is `discardable`. So the expression is discarded and the error is omitted when it is not in tail position.

This only get the [three errors in HEADCS as expected](https://travis-ci.org/gus-massa/racket/builds/542264743), and is easy to fix https://github.com/gus-massa/ChezScheme/commit/32c33eb8231a0a9729e2e0e6689af1fb18ed55b8

---

The second commit adds the same trick to `err/rt-test` so this behaviour in tested in many more expressions. This creates about 60 errors in the test of all version of Racket, so there are a lot of corner cases to review and fix.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/number.rktl
+++ b/pkgs/racket-test-core/tests/racket/number.rktl
@@ -739,9 +739,9 @@
 (err/rt-test (inexact->exact +inf.0))
 (err/rt-test (inexact->exact -inf.0))
 (err/rt-test (inexact->exact +nan.0))
-(err/rt-test (begin (inexact->exact +inf.0) 'not-an-error))
-(err/rt-test (begin (inexact->exact -inf.0) 'not-an-error))
-(err/rt-test (begin (inexact->exact +nan.0) 'not-an-error))
+#;(err/rt-test (begin (inexact->exact +inf.0) 'not-an-error))
+#;(err/rt-test (begin (inexact->exact -inf.0) 'not-an-error))
+#;(err/rt-test (begin (inexact->exact +nan.0) 'not-an-error))
 
 #reader "maybe-single.rkt"
 (when has-single-flonum?
--- a/pkgs/racket-test-core/tests/racket/number.rktl
+++ b/pkgs/racket-test-core/tests/racket/number.rktl
@@ -739,6 +739,9 @@
 (err/rt-test (inexact->exact +inf.0))
 (err/rt-test (inexact->exact -inf.0))
 (err/rt-test (inexact->exact +nan.0))
+(err/rt-test (begin (inexact->exact +inf.0) 'not-an-error))
+(err/rt-test (begin (inexact->exact -inf.0) 'not-an-error))
+(err/rt-test (begin (inexact->exact +nan.0) 'not-an-error))
 
 #reader "maybe-single.rkt"
 (when has-single-flonum?
--- a/pkgs/racket-test-core/tests/racket/testing.rktl
+++ b/pkgs/racket-test-core/tests/racket/testing.rktl
@@ -220,12 +220,15 @@ transcript.
     [(expr exn-type?) (thunk-error-test (lambda () (eval expr)) expr exn-type?)]))
 
 (require (only-in racket/base [lambda err:mz:lambda])) ; so err/rt-test works with beginner.rktl
-(define-syntax err/rt-test
+(define-syntax-rule (err/rt-test e . rest)
+  (begin
+    (do-err/rt-test e . rest)
+    (do-err/rt-test (let () e 'not-an-error) . rest)))
+(define-syntax do-err/rt-test
   (lambda (stx)
     (syntax-case stx ()
       [(_ e exn?)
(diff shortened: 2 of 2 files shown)


mflatt:
This is great!

So far, I see:

  * `SCHEME_PRIM_IS_OMITABLE_ALLOCATION` is incorrectly set for `make-hash` and similar functions. Maybe the solution is to have a flag that constrains that spec to 0 arguments.

  * The optimizer is likely too optimistic about some `make-struct-type` and `make-struct-type-property` calls.

 * The annotation on `abs` should say that it wants and produces a real.

 * The `unsafe-weak-hash-iterate-key` annotations probably have to be more conservative, since there are some valid uses that produce errors.

 * The failures in `prompt` are probably because a test assumes that an errorring expression is evaluated only once.

 * The chaperone tests get stuck because an interposition raises an exception and leaves the hash table in a locked state. Not immediately sure what to do about that.

gus-massa:
I pushed the fix of the signature to [racket/ChezScheme](https://github.com/racket/ChezScheme) and to the PR for cptypes https://github.com/cisco/ChezScheme/pull/192 so I don't forget it in case it is merged.

I think it's better to not merge the change of this PR to racket now, but try to fix the errors one by one and then merge it.

I think it's better to rename the old version of `err/rt-test` to `err/rt-test/once` so it can be used when the expression must be evaluated only once, and make the new version of `err/rt-test` evaluate the expression as many times as necessary. (In this PR 2 times, but perhaps later we can think another pattern that is interesting.)

mflatt:
I've pushed repairs for the problems that this change exposed.

Using `err/rt-test/once` seems like the right idea for the rest. I didn't push a `err/rt-test/once` change, but here's a commit where I tried it out:

https://github.com/mflatt/racket/commit/ec7e93961a2e507363f59727cafb7f17f52c5692

The test driver in "htdp-test" will need to change to use `err/rt-test/once`, too. I don't know if there are other places that would need to change.

gus-massa:
Nice. I tried to fix some parts yesterday, but I got stuck and went to sleep. I put them in a branch a few minutes ago just for reference [gus-massa/19-6-ExactInf-Z](https://github.com/gus-massa/racket/tree/19-6-ExactInf-Z)

I'll try to take a look to see the differences. For example, I tried to add the case `(make-hash '())`, but I don't remember to have seen it too much in the wild.

About `err/rt-test/once`, perhaps we can send an email to the dev mailing list about the internment change. For example it is interesting to note that there are about 2500 uses of `err/rt-test` in the test suit, and most of them are fine. Only 40 discovered an error that was fixed and only 15 have to be replaced with `err/rt-test/once`.

mflatt:
I agree that it's a good idea to change `err/rt-test` and fix up with `err/rt-test/once` as needed. I don't think that "testing.rktl" is so widely used that we need to poll the dev list; it's not documented, advertised, or even in module form. So, let's merge the change and see what needs to be fixed.

mflatt:
When I try Racket CS, then besides `inexact->exact` on infinities, I see failures for `abs`, `truncate`, `floor`, `ceiling`, `round`, `make-rectangular`, and `make-polar` — all functions that require reals, but are marked as `safeongoodargs` for `number`. Also, `(angle 0)`, which should be a divide-by-zero.

Is the right change to update those functions with `real` instead of `number`, and also remove `safeongoodargs` for `angle`?

gus-massa:
I'm fine with merging. I was thinking about about an announcement in the dev mailing list, not a pool, perhaps after the merge. When this change makes a test fails, it will be weird because it will be an error in code that looks correct.

I think that the changes of the signatures are correct. (It's strange because the type `number` in the signature makes Chez Scheme try the function in `1+2i`, so I expected that it should be raising an error in their tests. But looking at the test routine again, IIUC it never test the functions with all the correct parameters to ensure it is not failing.)

I took a looks at htdp-test and I agree that it's important to fix. In particular because the double evaluation will leak to students and beginners, and it's too difficult to explain all the magic in macros. It also makes some trick with modules, that I don't know if they can be evaluated twice.) I also made a global search in github https://github.com/search?q=err%2Frt-test&type=Code and there are more hits than I expected. Most look some some kind of variation of the implementation of Racket, but I didn't take a very detailed look.

samth:
@gus-massa I'm pretty confident that all of those hits are copies of the Racket test suite itself, in some form, and thus probably include the infrastructure code.

gus-massa:
Merged. https://github.com/racket/racket/commit/589bd920a9e910073b5a174ceea8d2dae689edae https://github.com/racket/racket/commit/845dada2575cf05639454093568200e5479b3604 https://github.com/racket/racket/commit/d75b5251705845cdaaecf6f62fc03c202a7792f5

PR #2695, opened 2019-06-13
Add vector-empty?

cwfoo:
Add `vector-empty?`.

Returns `#t` if the given vector is empty (i.e. its length is 0), `#f` otherwise.


pmatos:
Lets not let this bit-rot. I am merging this since @bennn approved.

PR #2700, opened 2019-06-16
Fix #1399: fix struct-copy bugs by exploiting struct-info

### LINKED ISSUE #1399: struct-copy fails if the structure type transformer binding is renamed
The following example fails with a syntax error:

``` racket
#lang racket/base

(module struct racket/base
  (provide (struct-out point))
  (struct point (x y) #:transparent))

(require (rename-in 'struct [point point2d]))

(struct-copy point2d (point2d 1 2) [x 3])
```

```
struct-copy: accessor name not associated with the given structure type
  in: point2d-x
```

This happens because `struct-copy` synthesizes accessor names by concatenating the symbol provided as the structure type with each field symbol, separated by a hyphen. This seems incorrect, given that it assigns significance to the identifier’s _name_ when really its _binding_ should be all that matters.

sorawee:
This PR fixes three bugs:

1. Accessors are required at use-site in order to use struct-copy.
This PR removes that requirement since the information is already available in
struct-info. The following program used to fail prior the PR but will now pass.

    ```
    (module a racket
      (provide a)
      (struct a (b)))

    (require 'a)
    (struct-copy a (a 1) [b 2])
    ```

2. struct-copy fails if the structure type transformer binding is renamed (#1399).
The following program used to fail prior the PR but will now pass.

    ```
    (module struct racket/base
      (provide (struct-out point))
      (struct point (x y) #:transparent))

    (require (rename-in 'struct [point point2d]))

    (struct-copy point2d (point2d 1 2) [x 3])
    ```

3. With supertype, it's possible to construct colliding accessors,
causing struct-copy to update an incorrect field. The following program produced
incorrect outputs prior this PR but will now be correct.

    ```
    (module a racket
      (provide a)
      (struct a (b-c) #:transparent))

    (require 'a)
    (struct a-b a (c) #:transparent)

    (struct-copy a-b (a-b 1 2) [b-c #:parent a 10])
    ;; before the PR: (a-b 1 10), after the PR: (a-b 10 2)
    (struct-copy a-b (a-b 1 2) [c 10])
    ;; before the PR: (a-b 1 10), after the PR: (a-b 1 10)
    ```

    Another buggy variant of 3) is also fixed.

    ```
    (module a racket/base
      (provide (all-defined-out))
      (struct a (b-c) #:transparent))

    (require 'a)
    (struct a-b a (d) #:transparent)
    (struct-copy a-b (a-b 1 2) [c 10]) ; used to run before the PR, is a syntax error after the PR
    ```

The key idea is that the actual struct name (if the struct is created via `struct`
or `define-struct`) can be extracted from the name of struct predicate.
The actual field names then can be precisely extracted from accessors.

Note that struct-infos that are created manually by `make-struct-info` didn't work
with struct-copy. This PR didn't attempt to fix that because it requires a significant
change that would not be backward compatible with the current struct info.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/struct.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/struct.scrbl
@@ -426,8 +426,9 @@ Returns @racket[#t] if @racket[v] is a predicate procedure produced by
               ((fld-id [field-id expr]
                        [field-id #:parent parent-id expr]))]{
 
-Creates a new instance of the structure type @racket[id] with the same
-field values as the structure produced by @racket[struct-expr], except
+Creates a new instance of the structure type @racket[id] (which is defined via a
+@seclink["define-struct"]{structure type defining form} such as @racket[struct])
+with the same field values as the structure produced by @racket[struct-expr], except
 that the value of each supplied @racket[field-id] is instead
 determined by the corresponding @racket[expr]. If @racket[#:parent]
 is specified, the @racket[parent-id] must be bound to a parent
@@ -438,12 +439,11 @@ encapsulates information about a structure type (i.e., like the
 initial identifier bound by @racket[struct]), and the binding
 must supply a constructor, a predicate, and all field accessors.
 
-Each @racket[field-id] is combined with @racket[id] 
-(or @racket[parent-id], if present) to form
-@racket[id]@racketidfont{-}@racket[field-id] (using the lexical
-context of @racket[field-id]), which must be one of the accessor
-bindings in @racket[id]. The accessor bindings determined by different
-@racket[field-id]s must be distinct. The order of the
+Each @racket[field-id] must correspond to a @racket[field-id] in
+the @seclink["define-struct"]{structure type defining forms} of @racket[id]
+(or @racket[parent-id], if present). The accessor bindings determined by different
+@racket[field-id]s under the same @racket[id] (or @racket[parent-id], if present)
+must be distinct. The order of the
 @racket[field-id]s need not match the order of the corresponding
 fields in the structure type.
 
@@ -829,6 +829,9 @@ specified through a transformer binding to such a value.}
 
 Encapsulates a thunk that returns structure-type information in list
 form. Note that accessors are listed in reverse order, as mentioned in @secref{structinfo}.}
+Note that the field names are not well-defined for struct-type informations
+that are created with this method, so it is likely not going to work well
+with forms like @racket[struct-copy] and @racket[struct*].
       
 @(struct-eval '(require (for-syntax racket/base)))
 @(struct-eval '(require racket/match))
--- a/pkgs/racket-test-core/tests/racket/struct.rktl
+++ b/pkgs/racket-test-core/tests/racket/struct.rktl
@@ -1117,6 +1117,30 @@
   
   (syntax-test #'(struct-copy t (t 1 2 3) [a #:parent p 11])))
 
+(module test-struct-rename racket/base
+  (provide (rename-out [point point2d]))
+  (struct point (x y) #:transparent))
+
+(let ()
+  (local-require 'test-struct-rename)
+  (test (point2d 3 2) 'struct-copy1 (struct-copy point2d (point2d 1 2) [x 3])))
+
+(module test-struct-parent racket/base
+  (provide a)
+  (struct a (b-c) #:transparent))
+
+(let ()
+  (local-require 'test-struct-parent)
+  (struct a-b a (c) #:transparent)
+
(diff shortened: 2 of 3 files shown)


sorawee:
The tests now passed.

CC: 
- @mflatt who implemented the original `struct-copy`.
- @lexi-lambda, @LeifAndersen, @AlexKnauth, @samth from #1399.
- @bennn from #2618

bennn:
I tried for a few minutes, and haven't found an example where the predicate name (`(caddr the-struct-info)`) doesn't look like the accessor names.

lexi-lambda:
This is a curious protocol, in my opinion. This works today, for structure-type transformer bindings created with `struct`, but it means if we ever add a `#:predicate-name` option to `struct`, or if a user defines a macro that provides such a capability, then it will break again.

Why not add the new information explicitly, in a separate channel? We already have `prop:struct-auto-info`, and I don’t think anything’s stopping us from adding a `prop:struct-field-info`. That seems less hacky and less likely to paint ourselves into a corner than making assumptions about the predicate name.

AlexKnauth:
> Why not add the new information explicitly, in a separate channel? We already have `prop:struct-auto-info`, and I don’t think anything’s stopping us from adding a `prop:struct-field-info`. That seems less hacky and less likely to paint ourselves into a corner than making assumptions about the predicate name.

There was a pull request (still open, but inactive and with merge conflicts) that added `make-extended-struct-info` and `extract-struct-field-info` for that: https://github.com/racket/racket/pull/732.

Though adding `prop:struct-field-info` seems like it would be more flexible and more extensible than a new "extended-struct-info" structure type.

sorawee:
Will try `prop:struct-field-info`. Might take some time though since I'm not familiar with struct props, and I spent non-research work quota for this week already.

If anyone wants to take this over, however, feel free to do so.

sorawee:
I propose that we either abandon this PR or merge it for now as it is a strict improvement (though hacky). I'm still not familiar with struct props after one year, and I'm not sure if I will next year, so a proper fix from me is unlikely to happen soon.

rmculpepper:
I'm in favor of merging this, but it needs updated docs for struct-copy describing the new way that "field" names are mapped to accessors. Specifically, the third paragraph of the current docs needs to be rewritten.

rfindler:
I'm not sure that bullet point 1 falls under the "strict improvement" rubric. The code was doing an explicit check before, if I'm not confused? Do we know the rationale of that check?

rfindler:
Ah, whoops: get my struct code confused. It is the match pattern `struct*` that is doing an explicit check for something like bullet point 1: https://github.com/racket/racket/blob/master/racket/collects/racket/match/struct.rkt#L67 . Do we know the rationale for that explicit check and should that happen here too?

sorawee:
@rfindler Well, fixing `struct*` is the next goal!

The main purpose of code that you linked is to detect bad field names, I think.

```
#lang racket 

(struct foo (bar))
(match-define (struct* foo ([BAD-FIELD x])) (foo 1))
; struct*: field name not associated with given structure type in: foo-BAD-FIELD
```

Detecting bad field names is obviously important, and there are several ways to implement it. The strategy that `struct-copy` did before this PR and the way the `struct*` currently does is that it lookups accessors from the environment. That is, _it makes an extra assumption that accessors are available in the environment_. The three problems described above are unintentional consequences from this assumption. And what this PR is trying to do is to remove this assumption, since there's a way to get the bad field name detection working without this assumption. 

And yes, `struct*` has the same problem that `struct-copy` had before this PR. E.g., here's the problem from the second point.

```
#lang racket

(module struct racket/base
  (provide (struct-out point))
  (struct point (x y) #:transparent))

(require (rename-in 'struct [point point2d]))

(match-define (struct* point2d ([x 3])) (point2d 1 2))
; struct*: field name not associated with given structure type in: point2d-x
```

rfindler:
Ah, got it. Thanks.

bennn:
what do you know: `contract-out` uses the same "predicate -> struct name" protocol [link](https://github.com/racket/racket/blob/master/racket/collects/racket/contract/private/provide.rkt#L1093) --- to match parent-struct selectors against the field names in a contract

+1 for merging

sorawee:
@rmculpepper let me know if the updated docs look good to you. And if that's the case, it would be ready to merge I think.

rfindler:
I'm not clear on which parts of the syntax information in the struct makes sense to use in what context so in that sense I'm not very confident about the change. I do believe that the discussion is headed in a good direction, however, so I'm in favor. As for for the code itself, I took a look and don't see anything obviously wrong.

samth:
Can you squash this down into one commit?

samth:
Can you rebase again so that the tests re-run?

sorawee:
Done. The tests are all passed.

PR #2705, opened 2019-06-19
Fix cosmetic error in code example

tfidfwastaken:
For the second example in thread mailboxes (Section 18.2) in the Racket Guide the formatted output always shows `+` even if the operation is something else.
This patch fixes the output to show the right operation.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/concurrency.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/concurrency.scrbl
@@ -101,8 +101,9 @@ items then send the results to the main thread.
               (match (thread-receive)
                 [(list oper1 oper2 result-thread)
                  (thread-send result-thread
-                              (format "~a + ~a = ~a"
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/guide/concurrency.scrbl
```diff
@@ -101,8 +101,9 @@ items then send the results to the main thread.
               (match (thread-receive)
                 [(list oper1 oper2 result-thread)
                  (thread-send result-thread
-                              (format "~a + ~a = ~a"
+                              (format "~a ~a ~a = ~a"
                                       oper1
+                                      (if (equal? operation +) '+ '-)
```
bennn: What about using `(object-name operation)` ?

tfidfwastaken: That is a better solution! I'll update accordingly

PR #2706, opened 2019-06-19
contract: collapsible, prune leaves if eq? and trusted

bennn:
Change the condition for filtering leaf contracts via `eq?`.
Before, we looked for flat or chaperone contracts.
After, we look for flat or trusted contracts.

This seems to be better in two ways:

1. Untrusted chaperones could have side effects, and now those will no
   longer be dropped
2. Trusted impersonator contracts can be removed

But maybe I missed something, and it's not safe to remove `eq?` and
trusted impersonators (i.e. `2` is bad).

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/collapsible-arrow.rkt
+++ b/pkgs/racket-test/tests/racket/contract/collapsible-arrow.rkt
@@ -902,14 +902,16 @@
 
   (test/spec-passed/result
    'calculate-drops:untrusted-chaperone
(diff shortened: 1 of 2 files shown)


bennn:
After talking with Dan & Robby offline, it **is** safe to remove trusted impersonators.

The trouble with impersonators is that they might change their behavior depending on context. If that can happen, it's unsafe to drop. A trusted impersonator never does such things.

I think this is ready to merge now.

rfindler:
If the tests are passing, please do squash the commits and merge. Thanks!

PR #2709, opened 2019-06-24
Move NORETURN to the correct place

pmatos:
This follows change to https://github.com/racket/ChezScheme/commit/2e3a618b0072d547b6c5abe6dd8dbac36a98c10e. 
It removes `MZ_DECLARE_NORETURN`, and implements `NORETURN` for all supported compilers.

In the prototypes, the `NORETURN` keyword has to be added to the head of the prototype in order to work with MSVC. For this to work, it was necessary to fix `makex.rkt` script as well.

### DIFF
--- a/racket/collects/compiler/private/xform.rkt
+++ b/racket/collects/compiler/private/xform.rkt
@@ -734,7 +734,7 @@
           (printf "#define RET_NOTHING { SET_GC_VARIABLE_STACK((void **)__gc_var_stack__[0]); return; }\n")
           ;; A non-value return inserted at the end of a void-returning function:
           (printf "#define RET_NOTHING_AT_END { SET_GC_VARIABLE_STACK((void **)__gc_var_stack__[0]); }\n")
-          
+
           ;; Declare a temp variable to hold the return value of the indicated type:
           (printf (if callee-restore?
                       "#define DECL_RET_SAVE(type) type __ret__val__;\n"
--- a/racket/src/racket/include/scheme.h
+++ b/racket/src/racket/include/scheme.h
@@ -252,11 +252,25 @@ extern "C"
 {
 #endif
 
-#if defined(MZ_DECLARE_NORETURN) && defined(__GNUC__)
-#define NORETURN __attribute__((__noreturn__))
+#if !defined(NORETURN)
+#if defined(__GNUC__) || defined(__clang__)
+#define NORETURN __attribute__((noreturn))
+#elif defined(_MSC_VER)
+#define NORETURN __declspec(noreturn)
 #else
 #define NORETURN
-#endif
+#endif /* defined(__GNUC__) || defined(__clang__) */
+#endif /* !defined(NORETURN) */
+
+#if !defined(UNREACHABLE)
+#if defined(__GNUC__) || defined(__clang__)
+#define UNREACHABLE __builtin_unreachable()
+#elif defined(_MSC_VER)
+#define UNREACHABLE __assume(0)
+#else
+#define UNREACHABLE
+#endif /* defined(__GNUC__) || defined(__clang__) */
+#endif /* !defined(NORETURN) */
 
 /* Allowed by all configurations, currently: */
 #define MZ_CAN_ACCESS_THREAD_LOCAL_DIRECTLY
@@ -1670,7 +1684,7 @@ MZ_EXTERN Scheme_Object *scheme_eval_waiting;
 #endif
 
 #ifdef MZ_USE_JIT
-MZ_EXTERN void scheme_jit_longjmp(mz_jit_jmp_buf b, int v);
+MZ_EXTERN NORETURN void scheme_jit_longjmp(mz_jit_jmp_buf b, int v);
 MZ_EXTERN void scheme_jit_setjmp_prepare(mz_jit_jmp_buf b);
 # define scheme_jit_setjmp(b) (scheme_jit_setjmp_prepare(b), scheme_call_mz_setjmp((b)->jb))
 #else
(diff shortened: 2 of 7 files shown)


pmatos:
@mflatt I want to fix that before merging. Can you please tell me the specific clang version you are using?

mflatt:
Below is full info on the main one I use, LLVM 10.0.1 (clang-1001.0.46.4). I also checked an older version, LLVM 8.0.0 (clang-800.0.42.1) in an El Capitan installation, and the warnings looks the same.

To avoid the warning, I had in mind adding
```
#ifdef __clang__
# pragma clang diagnostic ignored "-Winvalid-noreturn"
#endif
```

```
laptop% gcc -v
Configured with: --prefix=/Library/Developer/CommandLineTools/usr --with-gxx-include-dir=/Library/Developer/CommandLineTools/SDKs/MacOSX10.14.sdk/usr/include/c++/4.2.1
Apple LLVM version 10.0.1 (clang-1001.0.46.4)
Target: x86_64-apple-darwin18.6.0
Thread model: posix
InstalledDir: /Library/Developer/CommandLineTools/usr/bin
```

pmatos:
Thanks. I had an hour to look into this. I haven't reached a solution that's satisfying but I have a few notes and a question:
* I see some `NO_SCHEME_EXNS` guards (`error.c` for example) and looking a the code it seems that when this is not defined we use `scheme_longjmp`, and friends. When it is defined, I didn't yet manage to understand what's going on. However, my big question is: How is this even set? It doesn't seem to be a configure flag and there are only 7 occurrences of it. Is this something that is there for historical reason and can maybe be refactored/removed?
* One of the functions that clang complains about seems to be indeed my mistake. `scheme_raise_exn` seems to return to the caller because `do_raise` returns a `Scheme_Object`.
* The other functions are in `xsrc`. The Xform compiler is adding at the end of the functions `RET_NOTHING_AT_END`, which ends in a `return`. Is there a reason why this macro has a `return` after the `SET_GC_VARIABLE_STACK`? 

I need to look into it deeper as I haven't had much time but I would like to get this fixed properly avoiding workarounds, if possible. Also, it's interesting for me to dig around the source code so I slowly understand how it works.

mflatt:
`NO_SCHEME_EXNS` is definitely obsolete and could be cleaned out.

The `do_raise` function probably returns `Scheme_Object*` as an early-ish experiment in telling the compiler that raising an exception doesn't continue the rest of the function (i.e., meant to be called as `return do_raise(....)`).

I don't know whether the `return` in `RET_NOTHING_AT_END` is needed, but probably not.

pmatos:
Thanks for the feedback. Let me clean things up and come up with an improved PR.

pmatos:
> The `do_raise` function probably returns `Scheme_Object*` as an early-ish experiment in telling the compiler that raising an exception doesn't continue the rest of the function (i.e., meant to be called as `return do_raise(....)`).
> 

I am confused about something here. In `error.c` if `do_raise` returns nothing and should only be called as `do_raise(...)` without a `return` because it never does, then `sch_raise` doesn't return either. However, this causes a problem with the code in `scheme_init_exn`:
```
  scheme_addto_prim_instance("raise",
			     scheme_make_noncm_prim(sch_raise,
                                                    "raise",
                                                    1, 2),
			     env);
```

`scheme_make_noncm_prim` expects a `Scheme_Prim*` as first argument which is a function pointer that returns a `Scheme_Object`. Do we need to create a different typo for these non-return functions? I am a bit lost in this part of the code.

mflatt:
I think you probably need to cast the revised `sch_raise` to `Scheme_Prim*` as an argument to `scheme_make_noncm_prim`.

pmatos:
> I think you probably need to cast the revised `sch_raise` to `Scheme_Prim*` as an argument to `scheme_make_noncm_prim`.

OK, tested and can't see any problems. Opened #2717

pmatos:
@mflatt with all the side fixes put in until now I don't expect you would see any issues with `clang`. OK to merge this one as well?

mflatt:
With this patch, I still get
```
racket/src/error.c:4623:1: warning: function declared 'noreturn'
      should not return [-Winvalid-noreturn]
```
because `scheme_top_level_do` and `do_raise_inside_barrier` are not declared as `NORETURN`, and the former can't be.

I think this may be the point where I gave up before.

pmatos:
Interesting. I didn't see that straight away for some reason (possibly non clean build or messed up the order in which I merged the patches). But with a clean build, rebase from master etc, I can see it. I have an idea but it might have to wait to tomorrow. RN is next.

pmatos:
By the way, thank you for your time merging and reviewing all of my patches. It's really appreciated.

pmatos:
I am confused why there are so many commits in this PR. My guess is that it has to do with all the merges from master and rebases I have done. Let me try to simplify this with a squash and force-push.

pmatos:
Ah, it worked. 

@mflatt `noreturn` cascades down. So, a lot of the functions were required to be marked no return. On the issue you mentioned with `scheme_do_top_level`, the trick was to implement `unreachable` annotation and use it for this case to tell the compiler that although `scheme_do_top_level` does return (not marked with `NORETURN`) the line that follows it is unreachable.

I have tested with both `clang 8.0.0` and `gcc 8.3.0` on linux and got no errors. Could you please test on your side as well? The `__assume` trick on windows to mark unreachable comes from https://docs.microsoft.com/en-us/cpp/intrinsics/assume?view=vs-2017 but it was not tested on my side yet.

If all is well, I will start doing CI with `-Werror`. Will also be adding CI on Windows as I just got a Windows machine for this specific purpose.

pmatos:
@mflatt That's perfect news! Merging...

mflatt:
Not sure I'm reading things right, but it looks like this change may have caused a problem for OpenBSD: https://github.com/mflatt/ChezScheme/commit/4fdd7b04aba7fc48d4ef6baf4c64b58a22cec3ea#commitcomment-34166930

mflatt:
Looks like `__builtin_unreachable` appeared in GCC 4.5. On OpenBSD 6.3, GCC is version 4.2.

samth:
It appears (from online docs) that the `clang` version (5.0.1) on OpenBSD 6.3 should support `__builtin_unreachable`.

PR #2710, opened 2019-06-26
Update general-function.scrbl

llNoahll:
In order to break the contract, there should be more than one argument.


rfindler:
Thanks!

PR #2712, opened 2019-06-26
Remove reference to NO_SCHEME_EXNS

pmatos:
This is a followup to the discussion in #2709


pmatos:
Record time reply! Thanks.

PR #2713, opened 2019-06-30
Add `property/c` contract combinator for checking arbitrary first-order properties

lexi-lambda:
This PR adds a new contract combinator, `property/c`, which constructs flat contracts that check first-order properties of values given an accessor for the property and a contract that the property should conform to. For example, rather than writing

```racket
(define list-of-three-elements/c (and/c list? (lambda (lst) (= (length lst) 3))))
```

one can write

```racket
(define list-of-three-elements/c (and/c list? (property/c length (=/c 3))))
```

This does not save very much typing, but it provides dramatically better error reporting for contract violations. When `list-of-three-elements/c` is applied to the value `'(1 2)`, the first contract produces the error

```
lst: broke its own contract
  promised: ???
  produced: '(1 2)
  in: an and/c case of
      (and/c (listof any/c) ???)
```

while the second contract using `property/c` produces

```
lst: broke its own contract
  promised: (=/c 3)
  produced: 2
  in: the length of
      an and/c case of
      (and/c
       (listof any/c)
       (property/c length (=/c 3)))
```

which is much more informative.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1091,6 +1091,33 @@ This function is a holdover from before @tech{flat contracts} could be used
 directly as predicates. It exists today for backwards compatibility.
 }
 
+@defproc[(property/c [accessor (-> any/c any/c)]
+                     [ctc flat-contract?]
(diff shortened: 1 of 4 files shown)


rfindler:
This looks good to me. Thanks!

PR #2714, opened 2019-06-30
Do not add explicit return statement when it doesn't exist

pmatos:
Remove whitespace from empty lines as well.
Adding return statement where it doesn't exist, causes problems with
functions marked no return so it should be avoided.

Related to #2709 - with this PR, clang won't complain any longer about issues with xform generated sources about functions marked no return that do indeed return.


pmatos:
Happy to do the whitespace removal as a separate PR.

pmatos:
Not sure I understand why it's seeing two commit now. Maybe I have done something wierd in git with the forced pushes into different branches. Let me restart this process. Will close this.

PR #2715, opened 2019-07-01
Do not add explicit return statement when it doesn't exist

pmatos:
Adding return statement where it doesn't exist, causes problems with
functions marked no return so it should be avoided.

Related to #2709 - with this PR, clang won't complain any longer
about issues with xform generated sources about functions marked
no return that do indeed return.


pmatos:
Followup to #2714

PR #2719, opened 2019-07-01
racket/private/serialize: eliminate unnecessary vector->list

LiberalArtist:


mflatt:
Out of curiosity, does this make a noticeable difference in one of your programs?

LiberalArtist:
No, or at least I hadn't noticed it if so. I was looking at the source to confirm that I understood how and when sharing is detected (i.e. that it's based on `eq?`). I'm working with some large prefab structs, so I happened to look at that case particularly closely, and then I noticed that all of the uses of `vector->list` were constructing superfluous intermediate values.

I haven't tried to measure the performance impact, though I'd guess it would usually be dwarfed by the cost of IO in practice.

mflatt:
What do you think about using `for .... in-vector` and `for/list .... in-vector` directly, instead of wrapping them as `vector-map` and `vector-for-each`? After looking at this a little more, I think that would be better.

LiberalArtist:
That makes sense. I'll change it now.

mflatt:
Merged with a small correction: `(x v)` -> `(loop x)`. Thanks!

PR #2720, opened 2019-07-02
don't expect Racket CS failures on Travis

mflatt:
Racket CS should pass the test suite, now, and it has been passing on Travis. I'm not sure of the ".travis.yml" syntax, so let's see if this is right.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -6,6 +6,9 @@ matrix:
     compiler: clang
     env: PATH=./racket/bin:$PATH
   - os: linux
+    compiler: clang
+    env: PATH=./racket/bin:$PATH TARGET="cs"
+  - os: linux
     compiler: gcc
     env: PATH=./racket/bin:$PATH
   - os: linux
@@ -22,13 +25,6 @@ matrix:
     compiler: gcc
     env: PATH=./racket/bin:$PATH RACKET_CONFIGURE_ARGS="--disable-jit --disable-places
       --disable-futures --disable-extflonum"
-  - os: linux
-    compiler: clang
-    env: PATH=./racket/bin:$PATH TARGET="cs"
-  allow_failures:
-  - os: linux
-    compiler: clang
-    env: PATH=./racket/bin:$PATH TARGET="cs"
 before_script:
 - git config --global user.email "travis-test@racket-lang.org"
 - git config --global user.name "Travis Tester"


sorawee:
@mflatt can we reorder so that Racket CS is the first item? Most builds take 30 mins, but Racket CS takes 50 mins. The max jobs that Travis can run is 5, so Racket CS should be among the first batch to get the total running time of max(50, 30 + 30) rather than the current configuration, which is 30 + max(30, 50)

.travis.yml
```diff
@@ -1,7 +1,6 @@
 sudo: false
 language: c
 matrix:
-  include:
```
sorawee: Did you intentionally delete this line? It doesn't look correct to delete it...

pmatos:
Great job - thanks for the heads up. Have removed `allow_failure` on the Gitlab CI side as well. 31e7226e459cb5df885871c91ee598f9a8c699de

gus-massa:
IIRC the `include` line is not necessary if there is not `allow_failures`. It's probably a parsing quirk for backward compatibility or to enable some bad formed files that are almost correct. I think it is better to keep the `include` line.

I like the idea of launching RacketCS first, so the wall time of the test is smaller.

mflatt:
Thanks! I added back `include` (removed only because it seemed to have been added with `allow_failures`) and reordered.

sorawee:
Hmm. It fails actually.

```
raco setup: error: during making for <pkgs>/scribble-lib/scribble
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/scribble-lib/scribble/blueboxes.rkt
raco setup: error: during making for <pkgs>/typed-racket-lib/typed/file
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/typed-racket-lib/typed/file/md5.rkt
raco setup: error: during making for <pkgs>/typed-racket-lib/typed/file
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/typed-racket-lib/typed/file/sha1.rkt
raco setup: error: during making for <pkgs>/typed-racket-lib/typed/file
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/typed-racket-lib/typed/file/tar.rkt
raco setup: error: during making for <pkgs>/typed-racket-lib/typed/racket
raco setup:   real->single-flonum: unsupported
raco setup:     compiling: <pkgs>/typed-racket-lib/typed/racket/date.rkt
Makefile:95: recipe for target 'in-place-setup' failed
```

mflatt:
Not sure it would have helped in this case, but I guess some help detecting that kind of failure (caused by an external change) is a good reason to push this change.

PR #2722, opened 2019-07-02
openssl: add key and cert-chain args to ssl-make-{client,server}-context, and other fixes

rmculpepper:
This PR makes it easier to create a server-verifying ("secure") client context with client credentials. It is rare for https clients to need client credentials, but some database servers use TLS client credentials for authentication.


mflatt:
LGTM

Issue #2724, opened 2019-07-04
Would be beautiful to have `combine-output-ports` procedure.


EligiusSantori:
With specs:
```
(combine-output-ports what . with) → output-port?
    what : output-port?
    with : list of output-port?
```

As example it very useful in this case:
```
(with-logging-to-port ((combine-output-ports
                        (current-error-port)
                        (open-output-file "my.log"))
    MY-SYNTAX
)
```
Procedure `make-output-port` is too unfriendly for this purposes.
I wonder why such procedure isn't already exists in `racket/port` package.

lexi-lambda:
I agree that such a function would be useful to have in `racket/port`, but in the meantime, it might help you to look at `make-pipe` and `copy-port`, which together make an implementation of this function pretty easy since `copy-port` accepts multiple output ports.

lkh01:
Unless @EligiusSantori wants to, I could implement this as my first contribution to racket :)
Seems straightforward.

sorawee:
Closed by #2726

PR #2726, opened 2019-07-05
Add combine-output-ports

hgluka:
This PR adds a `combine-output-ports` procedure to `racket/port`, as mentioned in #2724.

Usage (in REPL):
```
> (define port-a (open-output-string))
> (define port-b (open-output-string))
> (define port-c (open-output-string))
> (define port-d (open-output-string))
> (define port-abcd (combine-output-ports port-a port-b port-c port-d))
> (display "hello world" port-ab)
> (println (get-output-string port-a))
"hello world"
> (println (get-output-string port-b))
"hello world"
> (println (get-output-string port-c))
"hello world"
> (println (get-output-string port-d))
"hello world"
```

It is implemented using `make-pipe-specials` and `copy-port`, in a similar manner to `merge-input`. It works with a variable number of arguments (all of which need to be satisfy `output-port?`).

There is a bug, though, that I could use some help with.
The output of the above program is different depending on where it's run.
In REPL, it works as expected. In racket, when run as a file, all of the ports are empty. And in DrRacket, some of the ports are empty when their string is `println`ed, but they are all ok in the corresponding REPL.

I suspect it's a synchronization problem with:
```
(thread
  (lambda ()
    (apply copy-port (cons pin (cons what with)))))
```
But I didn't find a fix yet.

Any help would be appreciated.

### DIFF
--- a/racket/collects/racket/port.rkt
+++ b/racket/collects/racket/port.rkt
@@ -169,70 +169,70 @@
 (define (combine-output a-out b-out)
   (struct buffer (bstr out start end) #:authentic #:mutable)
   (define pending #f)
-
-  (define (write-pending/blocking!)
+  (define lock (make-semaphore 1))
+  (define ready-evt (replace-evt lock (λ (_) (replace-evt out1 (λ (_) out2)))))
+  (define retry-evt (handle-evt
+                     (replace-evt
+                      (semaphore-peek-evt lock)
+                      (λ (_) (replace-evt out1 (λ (_) out2))))
+                     (λ (_) #f)))
+  (define (write-pending!)
     (when pending
-      (write-bytes
-       (buffer-bstr pending)
-       (buffer-out pending)
-       (buffer-start pending)
-       (buffer-end pending))
-      (set! pending #f)))
-
-  (define (write-pending/non-blocking!)
-    (when pending
-      (define n
-        (write-bytes-avail*
-         (buffer-bstr pending)
-         (buffer-out pending)
-         (buffer-start pending)
-         (buffer-end pending)))
+      (define bstr  (buffer-bstr pending))
+      (define out   (buffer-out pending))
+      (define start (buffer-start pending))
+      (define end   (buffer-end pending))
+      (define n (write-bytes-avail* bstr out start end))
       (when n
         (cond
-          [(= n (- (buffer-end pending) (buffer-start pending)))
+          [(= n (- end start))
            (set! pending #f)]
           [(> n 0)
-           (set-buffer-start! pending (+ (buffer-start pending) n))]))))
-
-  (define (write-out bstr start end non-blocking? breakable?)
-    (cond
-      [non-blocking?
-       (write-pending/non-blocking!)
-       (and
-        (not pending)
-        (write-out* write-bytes-avail* bstr start end))]
-      [else
-       (write-pending/blocking!)
-       (cond
-         [(= start end)
-          (flush-output a-out)
-          (flush-output b-out)
-          0]
-         [breakable?
-          (write-out* write-bytes-avail/enable-break bstr start end)]
-         [else
-          (write-bytes bstr a-out start end)
-          (write-bytes bstr b-out start end)])]))
-
+           (set-buffer-start! pending (+ start n))]))))
+  (define (write-out bstr start end non-blocking? enable-break?)
+    (define result
+      (call-with-semaphore
+       lock
+       (λ ()
+         (write-pending!)
+         
+         (cond
+           [pending       retry-evt]
+           [(= start end) 0]
+           [enable-break? (write-out* write-bytes-avail/enable-break bstr start end)]
+           [else          (write-out* write-bytes-avail* bstr start end)]))
+       (λ () retry-evt)))
+    (when (eqv? result 0)
+      (flush-output out1)
+      (flush-output out2))
+    result)
   (define (write-out* write-initial bstr start end)
-    (define m (write-initial bstr a-out start end))
+    (define m (write-initial bstr out1 start end))
     (cond
       [(or (not m) (= m 0))
-       #f]
+       retry-evt]
       [else
        (define n
-         (or (write-bytes-avail* bstr b-out start (+ start m))
+         (or (write-bytes-avail* bstr out2 start (+ start m))
              0))
        (when (< n m)
-         (set! pending (buffer bstr b-out (+ start n) (+ start m))))
+         (set! pending (buffer bstr out2 (+ start n) (+ start m))))
        m]))
-
   (define (close)
-    (write-pending/blocking!))
-
+    (call-with-semaphore
+     lock
+     (λ ()
+       (when pending
+         (write-bytes (buffer-bstr pending)
+                      (buffer-out pending)
+                      (buffer-start pending)
+                      (buffer-end pending))
+         (set! pending #f))))
+    (flush-output out1)
+    (flush-output out2)) 
   (make-output-port
-   'tee-port
-   (replace-evt a-out (λ (_) b-out))
+   'tee
+   ready-evt
    write-out
    close))
 
--- a/racket/collects/racket/port.rkt
+++ b/racket/collects/racket/port.rkt
@@ -170,7 +170,7 @@
   (struct buffer (bstr out start end) #:authentic #:mutable)
   (define pending #f)
   (define lock (make-semaphore 1))
-  (define ready-evt (replace-evt lock (λ (_) (replace-evt a-out (λ (_) b-out)))))
+  (define ready-evt (replace-evt a-out (λ (_) b-out)))
   (define retry-evt (handle-evt
                      (replace-evt
                       (semaphore-peek-evt lock)
--- a/racket/collects/racket/port.rkt
+++ b/racket/collects/racket/port.rkt
@@ -170,11 +170,11 @@
   (struct buffer (bstr out start end) #:authentic #:mutable)
   (define pending #f)
   (define lock (make-semaphore 1))
-  (define ready-evt (replace-evt lock (λ (_) (replace-evt out1 (λ (_) out2)))))
+  (define ready-evt (replace-evt lock (λ (_) (replace-evt a-out (λ (_) b-out)))))
   (define retry-evt (handle-evt
                      (replace-evt
                       (semaphore-peek-evt lock)
-                      (λ (_) (replace-evt out1 (λ (_) out2))))
+                      (λ (_) (replace-evt a-out (λ (_) b-out))))
                      (λ (_) #f)))
   (define (write-pending!)
     (when pending
@@ -195,7 +195,6 @@
        lock
        (λ ()
          (write-pending!)
-         
          (cond
            [pending       retry-evt]
            [(= start end) 0]
@@ -203,20 +202,20 @@
            [else          (write-out* write-bytes-avail* bstr start end)]))
        (λ () retry-evt)))
     (when (eqv? result 0)
-      (flush-output out1)
-      (flush-output out2))
+      (flush-output a-out)
+      (flush-output b-out))
     result)
   (define (write-out* write-initial bstr start end)
-    (define m (write-initial bstr out1 start end))
+    (define m (write-initial bstr a-out start end))
     (cond
       [(or (not m) (= m 0))
        retry-evt]
       [else
        (define n
-         (or (write-bytes-avail* bstr out2 start (+ start m))
+         (or (write-bytes-avail* bstr b-out start (+ start m))
              0))
        (when (< n m)
-         (set! pending (buffer bstr out2 (+ start n) (+ start m))))
+         (set! pending (buffer bstr b-out (+ start n) (+ start m))))
        m]))
   (define (close)
     (call-with-semaphore
@@ -228,8 +227,8 @@
                       (buffer-start pending)
                       (buffer-end pending))
          (set! pending #f))))
-    (flush-output out1)
-    (flush-output out2)) 
+    (flush-output a-out)
+    (flush-output b-out))
   (make-output-port
    'tee
    ready-evt
--- a/racket/collects/racket/port.rkt
+++ b/racket/collects/racket/port.rkt
@@ -166,59 +166,35 @@
 (define (call-with-input-bytes str proc)
   (with-input-from-x 'call-with-input-bytes 1 #t str proc))
 
-(define (combine-output-ports what . with)
-  (define port-list (cons what with))
-  (map (lambda (p)
-         (or (output-port? p)
-             (raise-argument-error 'combine-output-ports "output-port?" p)))
-       port-list)
-  (define (evts-and evts)
-    (if (null? (cdr evts))
-        (car evts)
-        (wrap-evt (car evts)
-                  (let ([wrapped-evts (evts-and (cdr evts))])
-                    (lambda (v)
-                      (sync wrapped-evts))))))
+(define (combine-output-ports port-a port-b)
+  (define-values (b-pin b-pout) (make-pipe 4096))
+  (define (do-write-ab write-fun-a write-fun-b s start end)
+    (let ([written-to-a
+           (subbytes s
+                     start
+                     (+ start (write-fun-a (subbytes s start end) port-a)))])
+      (if (equal? (write-fun-b written-to-a port-b) (bytes-length written-to-a))
+          (bytes-length written-to-a)
+          b-pout)))
+  (define (flush-b write-fun-b)
+    (let ([b-buffer-content (read-bytes (pipe-content-length b-pin) b-pin)])
+      (write-fun-b b-buffer-content port-b)))
   (make-output-port
    'combined-output-port
-   ; combined port is ready when all inner ports are ready
-   (evts-and port-list)
-   ; write-out procedure
-   (lambda (s start end non-block? breakable?)
-     (let ([s (subbytes s start end)])
-       (andmap (lambda (p)
-                 (cond [non-block? (write-bytes-avail* s p)]
-                       [breakable? (parameterize-break #t (write-bytes s p))]
-                       [else (write-bytes s p)]))
-               port-list)))
-   ; closing combined port doesn't close nested ports
-   void
-   ; write-out-special only if all ports
-   ; can do it
-   (and (andmap port-writes-special? port-list)
-        (lambda (v no-buffer&block? breakable?)
-          (andmap (lambda (p)
-                    (cond [no-buffer&block? (write-special-avail* v p)]
-                          [breakable? (parameterize-break #t (write-special v p))]
-                          [else (write-special v p)]))
-                  port-list)))
-   ; get-write-evt when all ports can write atomic
-   (and (andmap port-writes-atomic? port-list)
-        (lambda (s start end)
-          (let ([s (subbytes s start end)])
-            (evts-and (map (lambda (p)
-                             (write-bytes-avail-evt s p))
-                           port-list)))))
-   ; get-write-special-evt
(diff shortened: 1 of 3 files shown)


lkh01:
Update:
I found that adding `(sleep 0)` after every write to the combined port fixed the problem.
I'll try making some kind of wrapper port with `make-output-port` that does just that.

mflatt:
A `(sleep 0)` never fixes a concurrency problem. It only makes some bugs easier to find and some other bugs harder to find. :)

I think the approach of using a thread is probably not a good fit for the behavior you want. Creating a custom port whose write function copies ito other ports will likely work out better.

lkh01:
That makes sense. I'll rework it. 
Thanks for your help 👍

lkh01:
There's still some more tests to write, but I'll mark it ready for review since the `combine-output-ports` function seems to be done.

racket/collects/racket/port.rkt
```diff
@@ -164,6 +166,66 @@
...
+                            (sync wrapped-evts)))))))
+  (make-output-port
+   'combined-output-port
+   ; evt
+   (apply wrap-evt-with-evts port-list)
+   ; write-out 
+   (lambda (s start end non-block? breakable?)
+     (let ([s (subbytes s start end)])
+       (if non-block?
+           (map (lambda (p)
```
samth: This indentation isn't what we'd normally use here -- if there are multiple lines, then the last argument should also go on its own line.

racket/collects/racket/port.rkt
```diff
@@ -164,6 +166,66 @@
...
   (with-input-from-x 'call-with-input-bytes 1 #t str proc))
 
+(define (combine-output-ports what . with)
+  (define port-list (cons what with))
+  (map (lambda (p)
+         (or (output-port? p)
+             (raise-argument-error 'combine-output-ports "output-port?" p))) port-list)
+  (define wrap-evt-with-evts
+    (lambda (evt . evts)
+      (if (equal? (length evts) 0)
```
samth: This can be `null?`.

racket/collects/racket/port.rkt
```diff
@@ -164,6 +166,66 @@
...
+                        (parameterize-break #t (write-bytes s p))) port-list)
+                 (map (lambda (p)
+                        (write-bytes s p)) port-list))
+             (bytes-length s)))))
+   ; close
+   void
+   ; write-out-special
+   (and (andmap port-writes-special? port-list)
+        (lambda (v no-buffer&block? breakable?)
+          (if no-buffer&block?
```
samth: I think this might be clearer by moving the conditional inside the `map` and using `cond`.

racket/collects/racket/port.rkt
```diff
@@ -164,6 +166,66 @@
...
+              (if breakable?
+                  (map (lambda (p)
+                         (parameterize-break #t (write-special v p))) port-list)
+                  (map (lambda (p)
+                         (write-special v p)) port-list)))))
+   ; get-write-evt
+   (and (andmap port-writes-atomic? port-list)
+        (lambda (s start end)
+          (let ([s (subbytes s start end)])
+          (apply wrap-evt-with-evts
```
samth: This should be indented.

racket/collects/racket/port.rkt
```diff
@@ -164,6 +166,66 @@
 (define (call-with-input-bytes str proc)
   (with-input-from-x 'call-with-input-bytes 1 #t str proc))
 
+(define (combine-output-ports what . with)
+  (define port-list (cons what with))
+  (map (lambda (p)
+         (or (output-port? p)
+             (raise-argument-error 'combine-output-ports "output-port?" p))) port-list)
+  (define wrap-evt-with-evts
```
samth: It looks like this function is always called using `apply`, so it might make sense to just have it take a list as an argument.

lkh01: Thanks, these comments are very useful.
I've also been thinking about defining all the functions passed to `make-output-port` explicitly, instead of "on the spot". I feel like that would make the intent of the `combine-output-ports` function much clearer. What do you think?

lkh01:
This is weird. Everything works on my machine.
Also, tests pass on travis-ci, but all the builds still produce the same error:
```
raco setup: --- checking package dependencies ---                  [7:48:19]
raco setup: found undeclared dependency:
raco setup:   mode: run
raco setup:   for package: "typed-racket-lib"
raco setup:   on package: "pconvert-lib"
raco setup:   dependent source: /home/travis/build/racket/racket/racket/share/pkgs/typed-racket-lib/typed-racket/base-env/compiled/base-env_rkt.zo
raco setup:   used module: (lib "mzlib/pconvert-prop.rkt")
raco setup: --- summary of package problems ---                    [7:48:21]
raco setup: undeclared dependency detected
raco setup:   for package: "typed-racket-lib"
raco setup:   on package:
raco setup:    "pconvert-lib"
Makefile:95: recipe for target 'in-place-setup' failed
make[3]: *** [in-place-setup] Error 1
make[3]: Leaving directory '/home/travis/build/racket/racket'
Makefile:77: recipe for target 'plain-in-place' failed
make[2]: *** [plain-in-place] Error 2
make[2]: Leaving directory '/home/travis/build/racket/racket'
Makefile:59: recipe for target 'cpus-in-place' failed
make[1]: *** [cpus-in-place] Error 2
make[1]: Leaving directory '/home/travis/build/racket/racket'
Makefile:54: recipe for target 'in-place' failed
make: *** [in-place] Error 2
The command "make $TARGET  \ CPUS="2"  \ PKGS="racket-test db-test unstable-flonum-lib net-test"  \ CONFIGURE_ARGS_qq="$RACKET_CONFIGURE_ARGS"  \ RACKETCS_SUFFIX="$RACKETCS_SUFFIX"" exited with 2.
```

samth:
That error is not your fault; it's caused by a change elsewhere that I'll
fix.

mflatt:
Thanks for the implementation! Unfortunately, there's a mismatch between what output ports normally guarantee in terms of atomicity and what this implementation of `combine-output-ports` guarantees. Really, the problem is that there's no way to provide the guarantees that would be most natural. Using Racket's implement system, you can't in general implement an "and" on events like writing to multiple output ports. But there are approximations that would make combined output ports useful, as long as the limitations are documented.

Consider the first argument to `make-output-port`. It's supposed to be an event that becomes ready when a write will definitely succeed, at least for one byte. The implementation here causes the event to become ready as soon as the first output port is ready in that sense; it then *blocks* until all of the other ports are ready. That doesn't work right if the caller combines the output port with some other event in a `sync` where the other event becomes ready before some of the output ports. Using `replace-evt` instead of `wrap-evt` would be more useful; the combined output port as an event would become ready as soon all all ports have reported being ready since sync start. Although one of the output ports might stop being ready meanwhile, and although there may never have been a point when all ports were ready at once, as long as the limited meaning of "ready" is documented, `combine-output-port` would probably be useful.

A more significant problem is the implementation of the `write-bytes` function. It current uses `write-bytes-avail*` on each the output ports individually, but there's no guarantee that bytes will be written by `write-bytes-avail*`, so written bytes can just get lost for one or more ports (or all of them). I think some amount of buffering is unavoidable; write as many bytes as possible to the first output port, and then don't try to write any more bytes to the first port until those same bytes can be successfully written to all the other output ports. For example, if 10 bytes go out to the first port and none go to the second port, then keep all 10 bytes for the next time that the `write-bytes` handler is called, and continue trying to write those 10 bytes instead of any newly supplied bytes. Flush mode should probably succeed only if there are no bytes pending for any ports.

The current implementation of `breakable?` mode is the wrong approach. The key constraint on `write-bytes/enable-break` is that a write succeeds or a break exception is raised, but not both. That can't be implemented using `parameterize-break`, because the break exception might be raised just after a recursive call to `write-bytes` returns. The `breakable?` mode will need to be implemented with `write-bytes-avail/enable-break` for the first port, so that bytes are either written to that first port or not written to any port. For the rest of the ports, continue like non-breaking mode: try to write the bytes immediately, but save them for the future if they can't be written. After bytes are written to the first port, there should be no break exceptions allowed while trying to write to the remaining ports.

It only gets harder with the `write-bytes-avail-evt` callback. You'd have to combine success of a recursive `write-bytes-avail-evt` on the initial port with a commitment to sending the bytes to other ports before any further writes, and that combination has to be atomic. It can be done with a helper thread, but I suggest just not trying to support atomic writes.

lkh01:
Hey, that's a lot of useful info.
I realized that the `evts-and` should use `replace-evt` instead of `wrap-evt`, and when I fixed that, all those issues you mentioned came up. I tried fixing `get-write-evt` by adding a check that makes sure that all ports are ready (again) with the fixed version of `evts-and`, before sending the event off to individual ports. I'm not sure if that would accomplish what I intended now that you pointed out that a port can stop being ready after the check on it is done. 
But then I tried to fix `write-bytes` and got stuck. Your idea and explanation kicked me back into gear and I think I finally understood the the problem.
I'll work on correcting `write-bytes` first and then I'll tinker with `get-write-evt` for a bit, but as you say, atomic writes might not be necessary.
I also think that `combine-output-ports` would still be useful even with the limitations.
Thanks for your help and sorry for a late reply.

lkh01:
I decided to rewrite the function and just combine two ports instead of a variable number of them. I found it easier to think about the problem that way.
The current implementation is also not completely right - the `no-buffer&block?` clause in `write-out` is throwing me off. I'm not sure how to not lose bytes without buffering _and_ without blocking. I think the other `cond` clauses are correct (but I'm not really sure).

samth:
I wanted this today, so I'd like to figure out how to get this added to `racket/port`, or move it to a separate package. 

Two things I note -- first, the docs still suggest multiple argument. Second, it would be great to mention `tee` (the unix utility) in the documentation. But really it needs a re-review from @mflatt.

97jaz:
@mflatt I was reading your comment above about how the `write-byes` function should operate, and it wasn't clear to me whether you think it should return a successful result (I mean a positive number of bytes written) only when all of the ports have been written to or when just the first has. It seems like there are downsides to both.

hgluka:
@97jaz I haven't looked at this code for a while, but just reading his comment, I think that, when working with two ports, it makes sense to write to both, and if the number of bytes written mismatch, buffer what wasn't written into their separate buffers, and return the lower of the two numbers. 

That should even be easy to extend to multiple (variable number of) ports.

97jaz:
@hgluka After thinking about this more, I don't think that could work, and I think the answer to my earlier question for Matthew must be that we consider the write successful, because we've already sent the bytes to one port. If we tell the caller that we sent fewer bytes than we did, the caller would generally end up trying to send the same bytes again. 

So, in your scenario where there are two ports, you'd need to return the greater of the two results but then refuse [with a `#f` result?] any more writes until the original write could be completed on both ports.

97jaz:
Here's a [draft of an implementation](https://gist.github.com/97jaz/ac2e446d70f80dcffc34f35ad294a7d9). 

Caveats:
- I'm not sure what, if anything, should be done about thread-safety. Specifically, should access to the pending write data be guarded by a mutex?
- Some implementations of UNIX `tee` write warnings to stderr and then plug on ahead if a write fails. Others allow the user the option of warnings or hard errors. It would be nice to do something similar here, if a write raises an exception (because I/O failed, for example).
- This implementation uses `racket/match` which would probably disqualify it from going in `racket/port`. (However, the use of `match` is minimal and can easily be removed.)
- It tees exactly two output ports. It's easy enough to support an arbitrary number of tee'd ports by building a tree of pairs, but it would be more efficient to have a flatter representation, which isn't too big a lift.
- It doesn't support atomic writes or specials or much of anything beyond the bare minimum.
- It's barely tested at all.

hgluka:
See, I got you to the right answer. That was on purpose!

Jokes aside, your draft implementation looks good to me. I'll go ahead and adapt it to `racket/port`, and then it will be ready to review.

97jaz:
I just noticed something interesting. I saw in the documentation that the `write-out` parameter to `make-output-port` is supposed to receive an _immutable_ byte string. That led me to realize that the `subbytes` call in my draft implementation should be unnecessary; we could instead store the given byte string, along with the correct offsets for the pending data. (The `subbytes` call is only there to ensure that we copy the input, so that it doesn't get mutated underneath us.)

But it turns out that the byte string given to `write-out` is not immutable. (And, in fact, it seems to be mutated and reused.) I'm running on Racket CS, so this may be peculiar to the Chez runtime.

97jaz:
To be a bit more specific: the contract given in the docs does not claim that the byte string is immutable (it just says `bytes?`), but the prose description does:
```
write-out — either an output port, which indicates that writes should be redirected to the given port, or a procedure of five arguments:

an immutable byte string containing bytes to write;
[...]
```

97jaz:
Updated the gist with a bugfix. It now also assumes that the byte string given to `write-out` will _not_ be mutated and reused. Not sure if that's a valid assumption or not.

mflatt:
I think this is very close to an implementation we can merge, but ports normally should be thread-safe. So, either the implementation should be thread-safe, or the documentation should make clear that it isn't.

Making the port thread-safe is probably mostly a matter of creating a semaphore and using `call-with-semaphore` around sequences of non-blocking operations. Avoid non-blocking operations, unless they can be done outside the lock (like in the case of flushing with no pending bytes, probably). Where writing to a port would block, return the port wrapped as an event that returns `#f`, so the operation will be retired when the port is ready.

Meanwhile, I'll push a repair to Racket CS so that the given byte string is immutable.

97jaz:
> Avoid non-blocking operations, unless they can be done outside the lock (like in the case of flushing with no pending bytes, probably).

@mflatt Was that supposed to read "avoid blocking operations"?

97jaz:
Also:

> Where writing to a port would block, return the port wrapped as an event that returns `#f`, so the operation will be retired when the port is ready.

Like `(wrap-evt output-port (lambda (_) #f))`? 

It wasn't obvious to me, from the documentation, that that would be legal, since returning an event is described in terms of `write-bytes-avail-evt`, where the event is only ready when bytes have been committed to the port.

I take it that, in this case, the event becomes ready when it's possible to write to the port, and the code sees the `#f` as an indication that nothing has actually been written, and the write should be attempted again immediately?

97jaz:
I'm a bit unclear on how blocking writes should be implemented in terms of non-blocking ones w.r.t. multiple threads (assuming that's what you're suggesting). My understanding of `write-bytes` is that it's supposed to block until `(- end start)` bytes have been written. Allowing a second thread to start a write after the first has requested a blocking write would violate that expectation, right? Or is that concern handled by a different layer of code, so that the custom port doesn't need to worry about it?

mflatt:
@97jaz Yes, avoid *blocking* operations, and an event that returns `#f` is good for provoking a retry. An output port is not obliged to block, even when it is allowed, and returning an event is always ok.

97jaz:
> An output port is not obliged to block, even when it is allowed [...]

Okay, but in that case, I think the documentation for `write-bytes` isn't right. It says:
> Like `write-string`, but writes bytes instead of characters.

And the documentation for `write-string` says:
> The result is the number of characters written to `out`, which is always `(- end-pos start-pos)`.

It sounds like there is no such guarantee (about the return value) for `write-bytes`, if I understand you correctly. It might write fewer bytes.

mflatt:
Maybe I'm answering the wrong question. The `write-bytes` and `write-string` functions keep waiting and retrying as long as needed for the underlying output port to accept the written bytes. That is, as you said, any needed blocking and retrying happens at a different layer than a port's writing method.

97jaz:
Great, thanks @mflatt!

97jaz:
@mflatt Is there a way to attempt a non-blocking flush of the underlying output ports?

97jaz:
Here's an updated gist: https://gist.github.com/97jaz/ac2e446d70f80dcffc34f35ad294a7d9

In this version, the code flushes the underlying ports while holding the lock. I'm not actually sure if this is a problem, since, if the flushes block, that will allow other threads to execute; they just won't be able to use this port, which doesn't seem like such a bad thing. At any rate, it's easy to avoid holding the lock for these flushes if there's value in doing so.

I've tried running this on the test that @hgluka provided in his PR, but there are a couple of problems:
- This code runs successfully on 7.6 BC. In 7.7 CS it fails with `internal error: choice event discovered after interrupt/retry callbacks`.
- There's a commented-out definition of `ready-evt`, which is only ready for synchronization when waiting on the semaphore won't block. This seems important, at least for the non-blocking case. However, with this definition, 7.6 BC hangs. I've determined that when the test calls `(flush-output port-ab)`, the call to `write-out` waits forever on the semaphore, which I do not understand and probably means I've made some boneheaded mistake. On 7.7 CS, it again fails with the internal error mentioned in the previous item.

mflatt:
@97jaz There's no API for a non-blocking flush attempt. I haven't looked at the new implementation, but I expect it's ok to be outside the lock when flushing, because flushing probably just means flushing the other ports (i.e., no state changes specific to the combining port). Yes, flushing the combining port should flush the target port.s Holding the lock while blocking (for a flush) is bad, because it makes concurrent operations that should be non-blocking get stuck if they try to take the lock.

I'll look into the CS internal error when I can.

hgluka:
@97jaz I think the reason `flush-output` hangs is that `retry-evt` is syncing `lock` inside of `call-with-semaphore`. I'm not sure if `(semaphore-peek-evt lock)` would be correct for `ready-evt`, but this might work:
```racket
  (define ready-evt (replace-evt lock (λ (_) (replace-evt out1 (λ (_) out2)))))
  (define retry-evt (handle-evt
                     (replace-evt
                      (semaphore-peek-evt lock)
                      (λ (_) (replace-evt out1 (λ (_) out2))))
                     (λ (_) #f)))
```

That way it doesn't hang and the current test passes. I'll have to make a thread-safety test as well, to check if this behaves correctly.

hgluka:
@97jaz The docs for [events](https://docs.racket-lang.org/reference/sync.html) say: "Synchronizing an event may affect the state of the event. For example, when synchronizing a semaphore, then the semaphore’s internal count is decremented, just as with semaphore-wait." 

Also, do we have to use `call-with-semaphore/enable-break` for `write-bytes-avail/enable-break` to work properly?

97jaz:
@hgluka Okay, I suppose the code that drives the custom output port probably `sync`s on the event. That makes sense. (I guess there's not much point in returning an event if no one is going to `sync` on it.)

As for `call-with-semaphore/enable-break`, I wondered about that too, but I don't think so. First off, we should never be blocking on the semaphore to start with (due to the use of the non-`#f` `try-fail-think` argument to `call-with-semaphore`). So a break would never be long delayed while waiting on lock acquisition. Second, the guarantee that Matthew mentioned in his earlier post (that the write either succeeds or a break exception is raised but not both) doesn't seem to require enabling breaks while polling the semaphore.

mflatt:
I've lost track of the status here. Is the current code a candidate for merging, or is work still in progress?

hgluka:
@mflatt Not quite ready yet, I haven't actually implemented the last batch of changes we talked about. I'll get to it soon<sup>TM</sup>.

hgluka:
@mflatt I think it's ready for review now.

mflatt:
This looks like it would work, except that I don't think `ready-evt` should involve `lock` at all. Including lock means that waiting on the output port would take the lock and never release it.

To make tests for cases where one or the other port blocks, you could use `make-pipe` with a limit. For example, when combining two pipes each with a limit of 10, then the combined port should not be ready after writing 10 bytes to it, and it should become ready only after reading one or more bytes from the read ends of both pipes. I'm willing to add that test while merging, if you're ready to be done with this.

Finally, the documentation needs `@history[#:added "7.7.0.9"]`. I'd be happy to add when merging, too.

PR #2728, opened 2019-07-06
json: fixing scientific notation exponent sign bug

corpix:
if exponent contains leading zeroes sign will always be `+`. example:

```console
$ racket                                                                                                                                                                                                                                                      (6s 38ms)
Welcome to Racket v7.3.
> (require json)
> (jsexpr->string 0.000001)
"1e-06"
> (string->jsexpr (jsexpr->string 0.000001))
1000000.0
```

cons: we will miss leading zeroes from error message [here]( https://github.com/racket/racket/compare/master...corpix:fix/json-scientific-notation?expand=1#diff-8e8b6de3ce46d90c0bdf1c838b642aebR421), could we do better?

### DIFF
--- a/pkgs/racket-test/tests/json/json.rkt
+++ b/pkgs/racket-test/tests/json/json.rkt
@@ -1,6 +1,6 @@
 #lang at-exp racket/base
 
-;; Mathias, added test for contracts on read-json 
+;; Mathias, added test for contracts on read-json
 
 (require json racket/string tests/eli-tester)
 (require racket/port)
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks very much for the repair!

I pushed a further change to avoid the question of error reporting. I don't like the behavior that parsing "1e-0x" treats the "x" as terminating the number instead of an error, but doing so preserves the old behavior (and allowing ill-formed JSON seems much less worse than the bug you found and fixed).

PR #2739, opened 2019-07-14
Small cleanup of hash-ref code

97jaz:
Instead of having different kinds of hash implement the
default-value-or-thunk-on-failure protocol, let's move
that to hash.ss and make the internal functions just
accept a default value (with no special handling for thunks).

I think this makes the code a bit simpler, but it also should
make it easier to abstract over the hash-ref code, which,
in turn, should make it easier to add hash-ref-pair and
hash-ref-key. (Or so I hope.)

### DIFF
--- a/racket/src/cs/rumble/hash.ss
+++ b/racket/src/cs/rumble/hash.ss
@@ -257,31 +257,36 @@
 (define hash-ref
   (case-lambda
     [(ht k)
-     (let ([v (hash-ref ht k none)])
+     (let ([v (hash-ref/none ht k)])
        (if (eq? v none)
            (raise-arguments-error
             'hash-ref
             "no value found for key"
             "key" k)
            v))]
-    [(ht k fail)
-     (cond
-      [(mutable-hash? ht)
-       (lock-acquire (mutable-hash-lock ht))
-       (let ([v (hashtable-ref (mutable-hash-ht ht) k none)])
-         (lock-release (mutable-hash-lock ht))
-         (if (eq? v none)
-             ($fail fail)
-             v))]
-      [(intmap? ht) (intmap-ref ht k fail)]
-      [(weak-equal-hash? ht) (weak-hash-ref ht k fail)]
-      [(and (impersonator? ht)
-            (authentic-hash? (impersonator-val ht)))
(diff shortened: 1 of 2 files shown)


racket/src/cs/rumble/hash.ss
```diff
@@ -257,31 +257,36 @@
...
-             ($fail fail)
-             v))]
-      [else (raise-argument-error 'hash-ref "hash?" ht)])]))
+    [(ht k default)
+     (let ([v (hash-ref/none ht k)])
+       (if (eq? v none)
+           (fail-hash-ref default)
+           v))]))
+
+(define (hash-ref/none ht k)
```
97jaz: i really wanted to make this `hash-ref/default` and have `hash-ref` pass `none` into it, but I ran into the fact that `impersonate-hash-ref` doesn't have any parameter for a default, and I was a bit scared to add one.

mflatt:
I may be misreading the code, but I don't think this does the right thing for hash-table locks and exceptions or other escapes. Wouldn't a hash table stay locked if a no-such-key exception is raised or if a failure thunk escaped?

97jaz:
> Wouldn't a hash table stay locked if a no-such-key exception is raised or if a failure thunk escaped?

I don't think so. With this change, the failure thunk is never used until after the lookup is complete. In other words, for all hash types, the behavior is:
- look up the value using `none` as the default
- if we get `none` back (at which point, all locks should be released), apply the `fail-hash-ref` function to the user-supplied default/thunk.

97jaz:
Maybe it would be useful for me to change the names of arguments to reflect the fact that while `hash-ref` accepts a `default-value-or-thunk`, with special handling in the thunk case, all of the `-ref` functions that are specific to a particular kind of hash do not. They just accept a default value, which is only ever returned, never applied.

mflatt:
Thanks, I think I see how I was misreading before.

Minor suggestion: I recommend using `let` instead of `begin0` at this level, just to make sure the compiler doesn't include unnecessary handling for multiple values — even though I would expect the optimizer to be effective as written currently.

97jaz:
I'm going to close this, since it will be part of a larger PR soon.

PR #2744, opened 2019-07-17
Ran tests/file/main as a program

pmatos:
Somehow nobody noticed before file tests have not been running.
This enables the tests/file/main tests by running it as a program as suggested.

You can see from: 
https://travis-ci.org/racket/racket/jobs/559220863

```
$ raco test -l tests/file/main
raco test: (submod "/Users/travis/build/racket/racket/pkgs/racket-test/tests/file/main.rkt" test)
raco test: @(test-responsible '(eli mflatt))
run as program for tests
The command "raco test -l tests/file/main" exited with 0.
```
This should probably have failed with a non-zero exit so we could have noticed it earlier.

The reason I have created a PR for this change is that I want to give early warning that merging this will break CI because the file tests are currently failing with:
```
# racket -l tests/file/main
test: 1/134 test failures:
/root/.racket/7.4.0.1/pkgs/racket-test/tests/file/gzip.rkt:98:8: test failure in (sync/timeout 10 ch)
  expected: 'success
       got: 'didnt-fail-in-tar-gzip
  context...:
   "/root/.racket/7.4.0.1/pkgs/racket-test/tests/file/main.rkt": [running body]
   temp37_0
   for-loop
   run-module-instance!125
   perform-require!78
```

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -113,7 +113,7 @@ test:ubsan:
     - raco test -l tests/racket/test 2>&1 | tee logs/test.log
     - racket -l tests/racket/contract/all 2>&1 | tee logs/contract-test.log
     - raco test -l tests/json/json 2>&1 | tee logs/json-test.log
-    - raco test -l tests/file/main 2>&1 | tee logs/file-test.log
+    - racket -l tests/file/main 2>&1 | tee logs/file-test.log
     - raco test -l tests/net/head 2>&1 | tee logs/net-head-test.log
     - raco test -l tests/net/uri-codec 2>&1 | tee logs/net-uri-codec-test.log
     - raco test -l tests/net/url 2>&1 | tee logs/net-url-test.log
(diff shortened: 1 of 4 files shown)


pmatos:
Here's an example of this happening: https://gitlab.com/LinkiTools/racket/-/jobs/253482722

pmatos:
I am worried about azure above (and appveyor saying the tests succeed), because if you look at the logs, the job is definitely failing.

mflatt:
The intent of the "run as a program" success was meant to make `raco test` run the tests in the directory just once, but the setup was mangled. In the fixup, I'm not sure I picked the right approach, but it should be better than the old behavior, and I think it means you don't have to make this change. 

In any case, I fixed the test.

mflatt:
For Azure, at least, it looks like a "script" doesn't fail if a single command fails.

Obligatory Stack Overflow example: https://stackoverflow.com/questions/55194446/how-do-i-get-my-azure-devops-pipeline-build-to-fail-when-my-linting-script-retur

pmatos:
Thanks for the fix. Not sure what to do for Azure. @samth, what was the motivation for adding azure pipelines to the CI system?

PR #2745, opened 2019-07-17
Repairs for syntax/template on Racket CS

jamesbornholt:
Consider this program, which is just a simplified version of the test in [syntaxlibs.rktl](https://github.com/racket/racket/blob/c397dba1450d5955e2347bdfe43e19500701dfa8/pkgs/racket-test-core/tests/racket/syntaxlibs.rktl):

```racket
#lang racket

(require (for-syntax syntax/template))

(define-syntax (a-template-test stx)
  (syntax-case stx ()
    [(_ tmpl)
     (transform-template #'tmpl
                #:save (lambda (stx) stx)
                #:restore-stx (lambda (v stx datum)
                                (datum->syntax stx datum stx stx stx)))]))

(equal? '(1) (syntax->datum (a-template-test (1))))
```

There are two problems that prevent compiling this program to bytecode (i.e., `raco make` fails) on both 3m and CS:

1. The argument to `#:restore-stx` should be a syntax object, not a literal procedure. Bytecode compilation fails on Racket 3m:
    ```
    $ raco make thing.rkt
    write: cannot marshal value that is embedded in compiled code
      value: #<procedure:...racket/thing.rkt:11:30>
    ```
    In other words, the test isn't right (although we never actually test bytecode marshalling of that test, just its run-time behavior).

2. The default argument to `#:cons-stx` in `transform-template` should also be a syntax object, not a literal procedure. This seems to still work on 3m, but on CS, compilation fails trying to serialize `cons`:
    ```
    $ racocs make thing.rkt
    fasl-write: invalid fasl object #<procedure:cons>
      compilation context...:
       /Users/james/code/racket/thing.rkt
    ```

The second issue prevents Rosette installing correctly on Racket CS. After the fix, Rosette's tests that involve `transform-template` work correctly even after bytecode compilation.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/template.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/template.scrbl
@@ -13,7 +13,7 @@
                              [#:leaf-datum-stx leaf-datum-proc-stx syntax? #'(lambda (v) v)]
                              [#:pvar-save pvar-save-proc (identifier? . -> . any/c) (lambda (x) #f)]
                              [#:pvar-restore-stx pvar-restore-stx syntax? #'(lambda (d stx) stx)]
-                             [#:cons-stx cons-proc-stx syntax? cons]
+                             [#:cons-stx cons-proc-stx syntax? #'cons]
                              [#:ellipses-end-stx ellipses-end-stx syntax? #'values]
                              [#:constant-as-leaf? constant-as-leaf? boolean? #f])
          syntax?]{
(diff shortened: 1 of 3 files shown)


mflatt:
@rmculpepper Does this look right to you?

I blame the difference between traditional Racket and Racket CS on the traditional Racket bytecode marshaler being too liberal, an accident of its implementation.

rmculpepper:
I'm not familiar with this module, but that seems right. 

I didn't update this module to work with the new template forms, but I don't know how much of a problem that is. The only use I found was in the R6RS implementation.

mflatt:
Ah, ok. I didn't remember this module, and I just assumed that it had more of a connection to `syntax/parse`. I'll merge the repair.

@jamesbornholt Thanks!

Issue #2607, opened 2019-04-13
Feature request: hash-ref-key operation for accessing representative key of `equal?`-based hash tables


lexi-lambda:
For `equal?`-based hash tables, it can be useful to be able to access the “representative” key stored in the table that is `equal?` to some given key (but not necessarily `eq?`). For example, a key type used to record bindings in a binding form might use `bound-identifier=?` in its implementation of `equal?`, and obtaining the original key would be useful for signaling duplicate binding errors (since the original key and the new key could have different source locations).

Possible names for the operation: `hash-key`, `hash-key-ref`, `hash-ref-key`, `hash-ref/key`, or `hash-assoc` (which could return both key and value, a la `assoc`).

mflatt:
This seems like a good idea. As a specific example, `datum-intern-literal` uses an internal version of this operation.

97jaz:
If this operation were added, I think we'd need to specify which key is retained after multiple `hash-set[!]`s with keys that are equal under `equal?` but not under `eq?`. E.g.,
```racket
(require racket/match)

(define (test hash setter)
  (define key1 "hello")
  (define key2 (substring key1 0))

(hash-set! tbl key1 1)
(hash-set! tbl key2 2)

(match-define (cons retained-key _) (hash-assoc tbl key1))

;; which is it?
(eq? retained-key key1)
(eq? retained-key key2)
```

That behavior isn't currently consistent across the several kinds of hashes in racket/c and racket/cs. Should it be?

97jaz:
There is a kind of consistency to the current behavior, now that I look at it more closely. Mutable hashtables consistently retain the older key; immutable ones consistently use the newer one. That does make a kind of sense. But I still think this behavior (or _some_ behavior) would need to be documented to make this feature useful.

97jaz:
I put together a [patch](https://github.com/97jaz/racket/commit/7014f90a3bce7645bbcf2914a0407d6873934024) that adds `hash-assoc` to the C-based racket implementation. I've taken the approach of adding an `interned_key` out-parameter to a number of existing hash-table/hash-tree functions.

@mflatt Is this a reasonable approach? It does add a bit of overhead to existing code.

As for racket/cs, it will be simple to add this to immutable hashes. As for mutable ones, there's a simple, but inefficient, approach using a combination of `hashtable-contains?` and `hashtable-cell` (mutatis mutandis for `eq?`-based hashtables). An efficient approach would require an addition to Chez tables (though probably a simple one).

[...]

PR #2756, opened 2019-07-23
Adds hash-ref-key primitive

97jaz:
For now, the operation is implemented on mutable tables in Chez
using a combination of hashtable-contains? and hashtable-cell.
A more efficient version will require modifying Chez.

See #2607


pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -273,6 +273,75 @@ result:
...
+
+(define table (make-hash))
+(hash-set! table original-key 'one)
+(hash-set! table key-copy 'two)
+
+(eq? (hash-ref-key table "hello") original-key)
+(eq? (hash-ref-key table "hello") key-copy)
+]
+
+Conversely, an immutable hash retains the key that was most-recently
```
97jaz: This behavior could be easily changed to match that of immutable hashes, if we want to do that.

racket/src/cs/rumble/hash.ss
```diff
@@ -1065,6 +1116,17 @@
                             hash-procs-ref
                             ht k none))
 
+(define (impersonate-hash-ref-key ht k)
+  (impersonate-hash-ref/set 'hash-ref-key #f
+                            (lambda (ht k v) (hash-ref-key ht k none))
+                            (lambda (procs ht k none-v)
+                              (let-values ([(new-k _) (|#%app| (hash-procs-ref procs) ht k)])
+                                (values new-k
+                                        (lambda (ht k none-v)
```
97jaz: This lambda requires a closure  (as `procs` is bound by the enclosing lambda). Not sure if that's okay.

Also, if `ht` is a hash chaperone and the resulting key isn't `chaperone-of?` the original, the error message will describe the key as a "value":
```racket
                   [else
                    (let ([new-r (new-v-or-wrap next-ht new-k r)])
                      (when chaperone?
                        (unless (chaperone-of? new-r r)
                          (raise-chaperone-error who "value" new-r r)))
                      new-r)]
```

To correct that, maybe we should add a `what-r` parameter to `impersonate-hash-ref/set`?

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -273,6 +273,75 @@ result:
...
+
+(eq? (hash-ref-key table2 "hello") original-key)
+(eq? (hash-ref-key table2 "hello") key-copy)
+]
+
+If @racket[hash] is an @tech{impersonator}, then the returned key
+will be determined as described in the documentation to
+@racket[impersonate-hash].
+
+@see-also-caveats[]}
```
mflatt: Need a `history` annotation here.

racket/src/cs/rumble/hash.ss
mflatt: Adding a `what-r` to `impersonate-hash-ref/set` sounds like a good idea.

racket/src/cs/rumble/intmap.ss
```diff
@@ -63,35 +61,42 @@
...
+          (loop (Br-left t))
+          (loop (Br-right t)))]
 
-   [(Lf? t)
-    (if (key=? et key (Lf-key t))
-        (Lf-value t)
-        ($fail def))]
+     [(Lf? t)
+      (if (key=? et key (Lf-key t))
+          (with-leaf t)
```
mflatt: Does replacing the known selector used with a call to an unknown function affect performance? I would not expect `do-$intmap-ref` to be inlined, but I may guess wrong.

racket/src/racket/src/schvers.h
```diff
@@ -16,7 +16,7 @@
 #define MZSCHEME_VERSION_X 7
 #define MZSCHEME_VERSION_Y 4
 #define MZSCHEME_VERSION_Z 0
-#define MZSCHEME_VERSION_W 1
+#define MZSCHEME_VERSION_W 3
```
mflatt: Also need to update "info.rkt" in the "base" package.

97jaz: Fixed in 2faf3b6cfc5d7d14fa2b3b26b39289c7e6b1a59a

pkgs/racket-doc/scribblings/reference/hashes.scrbl
97jaz: Fixed in ea8574a786565cf0221862c05578aeb9ab5349ec

racket/src/cs/rumble/intmap.ss
97jaz: Performance seems minimally effected by this change. 

To give you some more context, originally I implemented `do-$intmap-ref` as syntax, after the model of [`lookup-keyval` in Chez's implementation of mutable eq-hashtables](https://github.com/cisco/ChezScheme/blob/58663d92a353969e3b27de2e1a44df571036e1b2/s/library.ss#L1380). However, this had a marked negative effect on the "large table" tests from `hash-demo`:

Original implementation:
```
large tables
cpu time: 123 real time: 123 gc time: 2   MB: 292
cpu time: 162 real time: 162 gc time: 2   MB: 292
```

With `do-$intmap-ref` as syntax:
```
large tables
cpu time: 162 real time: 162 gc time: 2   MB: 292
cpu time: 336 real time: 336 gc time: 2   MB: 292
```

With `do-$intmap-ref` as function:
```
large tables
cpu time: 125 real time: 125 gc time: 2   MB: 292
cpu time: 164 real time: 164 gc time: 3   MB: 292
```

The other tests in that demo, however, are unaffected. I have no idea what to make of this. If Chez does loop unrolling, then I could see how the increased code size might have some negative cache effects, but I don't think it does any unrolling.

racket/src/cs/rumble/hash.ss
97jaz: Fixed in 42425bb70de068932e0aaced5d133fe7583f8b76

racket/src/cs/rumble/intmap.ss
97jaz: Well, I guess the [unrolling story](https://github.com/cisco/ChezScheme/blob/58663d92a353969e3b27de2e1a44df571036e1b2/s/cp0.ss#L79) is a bit more complicated than that.

mflatt:
Thanks! The version number at master is now already 7.4.0.2, so if you could rebase, update to 7.4.0.3, and squash, that would make the cleanest merge. But I can also work with these commits if you prefer.

97jaz:
- version (and history annotation) updated to 7.4.0.3
- squashed
- rebased against upstream master

PR #2769, opened 2019-07-28
Fix printing when print-vector-length is enabled

gus-massa:
**Fix `pretty-print` with `print-vector-length`**
In some cases, `(vector x 2 3 3 3)` was pretty-printed as `"(vector x 2 3)"` when
`print-vector-length` was enabled.

Also print `"(fxvector)"` instead of `"(fxvector )"`.

The test fail because `write` and `pretty-write` have different behavior:

    #lang racket

    (define-struct b (x y) #:transparent)

    (parameterize([print-vector-length #t])
      (pretty-write (b 1 1))  ; ==> #(struct:b 1 1)
      (writeln (b 1 1)))      ; ==> #3(struct:b 1)

I'm not sure if it's better to change `write` or `pretty-write`. (Or just add some flag to ignore this specific test.)

**CS: Fix `display` and `print` with `print-vector-length`**
Don't use `print-vector-length` in `display`.
Use `print-vector-length` in `print` for `fxvector`s and `flvector`s.

In this part I'm not sure if it's good to call `print-list` for the output or just try to copy the relevant part of `print-list` to vector.rkt.


mflatt:
Thanks for these repairs!I learn towards making `pretty-write` behave like `write`.

gus-massa:
I noticed that in pretty.rkt I fixed `wr*` but I forget to fix `pr*`. I'll fix that too, but it looks difficult to test. (I tried to add a test, but I used a port with `width='infinity` that is a change in the wrong direction. I'll revert most of the new commit in this branch.)

gus-massa:
I think I fixed all the problems. There is:

* a tiny fix to my previous commit for the `cs` version.

* more tests with more corner cases and corner^2 cases.

* a "copy" of my previous commit that fixed the single line version `pretty-print` to the multi line version.

* a fix to `pretty-write`, to get the same output that `write`

My only concern is that the multi-line version of `pretty-print` and friends don't have enough tests. But I prefer to merge this as is (after some rebasing and squashing).I have an idea about how to add the tests for the multi-line version but I must think the details, so I expect to send the PR in a few weeks.

mflatt:
Looks good as far as I can tell!

gus-massa:
Merged

PR #2770, opened 2019-07-28
macro in example said let, not let-values

gknauth:


bennn:
Hm. `let-values` looks like the wrong thing based on earlier parts of the file ([link](https://docs.racket-lang.org/reference/syntax-model.html#(part._fully-expanded)) and the output of `raco expand`, but later examples "fully expand" `let` to `let`.

So, LGTM

PR #2771, opened 2019-07-28
Fixes bug in intmap equality

97jaz:
Collision nodes were previously only testing for the equality
of keys and ignoring values. 🤦‍♂

### DIFF
--- a/racket/src/cs/rumble/intmap.ss
+++ b/racket/src/cs/rumble/intmap.ss
@@ -206,12 +206,6 @@
           [(key=? et key (caar xs)) (loop (cdr xs))]
           [else (cons (car xs) (loop (cdr xs)))])))
 
-(define ($collision-has-key? et t key)
-  (let loop ([xs (Co-pairs t)])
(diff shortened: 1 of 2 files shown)


racket/src/cs/rumble/intmap.ss
```diff
@@ -390,9 +384,12 @@
...
                  (let loop ([xs xs])
-                   (cond [(null? xs) #t]
-                         [($collision-has-key? et b (caar xs)) (loop (cdr xs))]
-                         [else #f])))))]
+                   (cond [(null? xs)
+                          #t]
+                         [else
+                          (let ([pair ($collision-ref et b (caar xs) values #f)])
+                            (and pair
+                                 (eql? (cdar xs) (cdr pair))))])))))]
```
97jaz: Well, this isn't right either. In my haste, I forgot to recur... Patch forthcoming.

97jaz: Updated, with a better test.

mflatt:
Thanks!

PR #2772, opened 2019-07-29
Mark struct-type properties as discouraged in favor of generic interfaces, not deprecated

AlexKnauth:
This pull request changes the documentation for the struct-type-properties `prop:custom-write`, `prop:equal+hash`, `prop:stream`, and `prop:dict`, to mark them as "discouraged" in favor of their respective generic interfaces, but not "deprecated".

The word "deprecated" has implications that they are intended to be removed in the future and you should absolutely avoid using them and migrate existing code away. However this is not the case with these struct-type properties that have corresponding generic interfaces. The generic interfaces are a convenience layer on top, so the properties are fully compatible and no behavior is going away. We want to leave the possibility that in situations where using the convenient generic-interfaces isn't an option, falling back to the properties is valid and isn't dangerous for future racket versions.

A transcript of some discussion on Slack for context on why "discouraged" is a better word than "deprecated":

> AlexKnauth 7:46 PM
> In the documentation for prop:custom-write and prop:equal+hash, it says that they are “deprecated”. In some languages, this means that they are intended to be removed in the future and you should absolutely avoid using them and migrate existing code away. I have gotten the impression that that isn’t what “deprecated” means here for those two struct-type-properties.
So what does it mean in this context?

> AlexKnauth 8:20 PM
> On both of those it looks like @stamourv was the one to introduce that word in these two commits:
> - custom-write https://github.com/racket/racket/commit/cee88f05dd29c99f9edf7051401bfb6434780b64
> - equal+hash https://github.com/racket/racket/commit/390cd02b52648637802b3a72af95355e13bfde2c

> stamourv 8:18 AM
> @AlexKnauth: Your understanding of the word "deprecated" is indeed in line with the Racket meaning. My understanding at the time was not. I don't think there was ever a plan to remove them, but rather a desire to steer people away from the older way of doing things and towards the new one. There needs to be a word for that, but deprecated isn't it.

> sorawee 8:29 AM
> “discouraged”?

> Sancho 10:28 AM
> Especially if we say what is preferred

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/booleans.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/booleans.scrbl
@@ -97,9 +97,9 @@ in the case of @tech{complex numbers}. Two characters are
 @defproc[(eq? [v1 any/c] [v2 any/c]) boolean?]{
 
 Return @racket[#t] if @racket[v1] and @racket[v2] refer to the same
-object, @racket[#f] otherwise. As a special case among @tech{numbers}, 
-two @tech{fixnums} that are @racket[=] are also the same according 
-to @racket[eq?]. See also @secref["model-eq"]. 
+object, @racket[#f] otherwise. As a special case among @tech{numbers},
+two @tech{fixnums} that are @racket[=] are also the same according
+to @racket[eq?]. See also @secref["model-eq"].
 
 @examples[
 (eq? 'yes 'yes)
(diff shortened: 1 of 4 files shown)


stamourv:
Yes, that's a good idea.
Nit: could you split off the whitespace changes to a separate commit, or skip them altogether? It makes it harder to see what the commit is actually doing.

AlexKnauth:
They are in a separate commit, so do you mean a different pull request?

Or I can easily take the whitespace commit out

stamourv:
Oh, my bad. I was looking at the overall diff. That's fine by me then, if you're ok showing up in git blame. ;)

AlexKnauth:
I'm okay with that

AlexKnauth:
Am I okay to merge?

stamourv:
Yes.

Issue #2711, opened 2019-06-26
Racket Guide (7.3.7 Notation) results of the sample code.


llNoahll:
I have tried the sample code, but _ and res seem to have the same effect.
```
> (module test1 racket
    (define x '())
    (define (get-x) x)
    (define (f) (set! x (cons 'f x)))
    (provide
     (contract-out
      [f (->i () [res (begin (set! x (cons 'ctc x)) any/c)])]
      [get-x (-> (listof symbol?))])))
> (require 'test1)
> (get-x)
'(ctc)
> (f)
> (get-x)
'(f ctc)
> (f)
> (get-x)
'(f f ctc)
> (f)
> (get-x)
'(f f f ctc)
```
```
> (module test2 racket
    (define x '())
    (define (get-x) x)
    (define (f) (set! x (cons 'f x)))
    (provide
     (contract-out
      [f (->i () [_ (begin (set! x (cons 'ctc x)) any/c)])]
      [get-x (-> (listof symbol?))])))
> (require 'test2)
> (get-x)
'(ctc)
> (f)
> (get-x)
'(f ctc)
> (f)
> (get-x)
'(f f ctc)
> (f)
> (get-x)
'(f f f ctc)
```

I have tried the following code, it seems to work.
```
> (module test3 racket
    (define x '())
    (define (get-x) x)
    (define (f) (set! x (cons 'f x)))
    (provide
     (contract-out
      [f (->i () [_ () (begin (set! x (cons 'ctc x)) any/c)])]
      [get-x (-> (listof symbol?))])))
> (require 'test3)
> (get-x)
'()
> (f)
> (get-x)
'(f ctc)
> (f)
> (get-x)
'(f ctc f ctc)
> (f)
> (get-x)
'(f ctc f ctc f ctc)
```
```
> (module test4 racket
    (define x '())
    (define (get-x) x)
    (define (f) (set! x (cons 'f x)))
    (provide
     (contract-out
      [f (->i () [res () (begin (set! x (cons 'ctc x)) any/c)])]
      [get-x (-> (listof symbol?))])))
> (require 'test4)
> (get-x)
'()
> (f)
> (get-x)
'(ctc f)
> (f)
> (get-x)
'(ctc f ctc f)
> (f)
> (get-x)
'(ctc f ctc f ctc f)
```

I thought I didn't need an empty list if res or _ doesn't depend on anything else, but it didn't seem like this.

samth:
Fixed by #3186

PR #2774, opened 2019-07-30
Display all strings in the result channel.

llNoahll:
There may be still a string in the result-channel, because it is possible that the main thread exits before the result-thread calls `(channel-get result-channel)' and display its result.


mflatt:
Thanks for the repair!

Without your PR explanation, I think this will seem subtle in the guide. Would it make sense to add a comment next to the last line, maybe like this?
```
(channel-put result-channel "") (code:comment "waits until result-thread has printed all other output")
```

llNoahll:
Yes, it looks better. By the way, I think there may be some bugs in `guide/contracts/general-function.scrbl', lines 466 and 472. 
https://github.com/racket/racket/issues/2711

PR #2775, opened 2019-07-30
Ensure interned_key is initialized

97jaz:
I'm actually a bit puzzled here, because I'm not sure why the tests for absent keys at the end of `chaperone.rktl` pass currently.

The `interned_key` pointer is only set when a key is found (or, alternatively, when the returned _value_ is non-NULL). So I'd expect this to be an uninitialized pointer when a key isn't found. But the test passes consistently, so maybe I'm missing something.


mflatt:
Racket3m initializes any uninitialized pointer to NULL, so that's probably why tests passed before.

PR #2791, opened 2019-08-18
[DOC]: simplify documentation for keyword arguments

MalloZup:
I personally find that with this is more clear the documentation.

It could be that the scribblings syntax isn't correct but It should afaik


sorawee:
OK to merge?

PR #2793, opened 2019-08-21
revert error message for '#lang  racket'

bennn:
If '#lang ' is followed by another whitespace, say that we expected
a single space (like Racket 6.12 would say).

EDIT: wow, github autocorrects `#lang<space><space>racket` to use one space. The commit message on my machine shows 2 spaces

### DIFF
--- a/racket/src/expander/read/extension.rkt
+++ b/racket/src/expander/read/extension.rkt
@@ -81,33 +81,36 @@
   (define accum-str (accum-string-init! config))
   (when init-c
     (accum-string-add! accum-str init-c))
-  (let loop ()
-    (define c (peek-char/special in config))
-    (cond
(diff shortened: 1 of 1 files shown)


mflatt:
I don't think this is what you had in mind. It changes the error message only when `#lang` is followed by a single space and then an end-of-file, as opposed to the case when there's one extra space. Also, there's a problem using the "expected a single space" message at this particular point in the reader, since it's also for `#!`.

Unless it sounds like I'm confused, I'll try a different repair.

mflatt:
This commit is meant to fix the error message: d14a4f75a158d8011ffa1da5a27212d8c34b3524

bennn:
That commit is much better, thanks!

Issue #2797, opened 2019-08-27
cannot build on alpinelinux


zaoqi:
```
$ docker run -it --rm alpine
# wget -O - https://mirror.racket-lang.org/installers/7.4/racket-minimal-7.4-src-builtpkgs.tgz | tar -xvz
# apk add gcc libc-dev binutils make
# cd racket-7.4/src
# ./configure
# make -j4
......
gcc -Wall  -I./.. -I./../include  -I./../../rktio -I../../rktio -g -O2  -pthread  -DUSE_SENORA_GC  -c ./syntax.c -o syntax.o
gcc -Wall  -I./.. -I./../include  -I./../../rktio -I../../rktio -g -O2  -pthread  -DUSE_SENORA_GC  -c ./thread.c -o thread.o
gcc -Wall  -I./.. -I./../include  -I./../../rktio -I../../rktio -g -O2  -pthread  -DUSE_SENORA_GC  -c ./type.c -o type.o
gcc -Wall  -I./.. -I./../include  -I./../../rktio -I../../rktio -g -O2  -pthread  -DUSE_SENORA_GC  -c ./unwind/libunwind.c -o unwind.o
In file included from ./unwind/libunwind_i.h:45,
                 from ./unwind/libunwind.c:31:
./unwind/libunwind.h:60: warning: "_XOPEN_SOURCE" redefined
 # define _XOPEN_SOURCE /* needed for Mac OS X */

In file included from /usr/include/inttypes.h:8,
                 from ./unwind/../../mzconfig.h:33,
                 from ./unwind/../../sconfig.h:20,
                 from ./unwind/libunwind.c:26:
/usr/include/features.h:16: note: this is the location of the previous definition
 #define _XOPEN_SOURCE 700

In file included from ./unwind/libunwind.c:31:
./unwind/libunwind_i.h:54: warning: "__BYTE_ORDER" redefined
 # define __BYTE_ORDER __LITTLE_ENDIAN

In file included from /usr/include/sys/types.h:70,
                 from ./unwind/libunwind_i.h:42,
                 from ./unwind/libunwind.c:31:
/usr/include/endian.h:11: note: this is the location of the previous definition
 #define __BYTE_ORDER __BYTE_ORDER__

./unwind/libunwind.c: In function '_Ux86_64_uc_addr':
./unwind/libunwind.c:3085:55: error: 'REG_R8' undeclared (first use in this function); did you mean 'EM_RX'?
[...]

PR #2807, opened 2019-09-02
Make __BYTE_ORDER definition conditional

pmatos:
Also remove duplicate definition.
Related to #2797


mflatt:
Does it work to just delete the definitions of `__LITTLE_ENDIAN`, `__BIG_ENDIAN`, and `__BYTE_ORDER`?

It looks like those are all defined by "/usr/include/endian.h", so we want to avoid defining them. So, we could change the names, but I don't see any use of those names to change.

pmatos:
I tried the local configs I had and removing the definitions worked. However I might have broken less used (by myself) configs - and we both know how good I am at breaking configs I don't use. :-)

Force pushed new commit.

mflatt:
Seems likely to work, so let's merge this.

PR #2809, opened 2019-09-03
Implements negated unicode categories in pregexps

97jaz:
The grammar for pregexps includes:
```
   | \p{‹property›} Match (UTF-8 encoded) in ‹property›
   | \P{‹property›} Match (UTF-8 encoded) not in ‹property›
```
and ‹property› is defined as:
```
   ‹property› ::= ‹category› Includes all characters in ‹category›
               |  ^‹category› Includes all characters not in ‹category›
```
That is to say, there are two independent ways to negate one of
these character classes. The Racket implementation of regexps
(as opposed to the C implementation) does not recognize negated
categories. This PR fixes that.

### DIFF
--- a/racket/src/regexp/parse/unicode.rkt
+++ b/racket/src/regexp/parse/unicode.rkt
@@ -1,5 +1,6 @@
 #lang racket/base
-(require "chyte.rkt"
+(require (only-in racket/bool xor)
+         "chyte.rkt"
          "chyte-case.rkt"
          "ast.rkt"
          "config.rkt"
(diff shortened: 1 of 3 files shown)


97jaz:
I noticed this when trying to build @jeapostrophe's [line-expression code](https://github.com/racket/racket2-rfcs/pull/114), which [uses this feature](https://github.com/racket/racket2-rfcs/pull/114/files#diff-f60a827ecc318716be8cdd6f77e74da0R28). The code wouldn't build in racketcs.

racket/src/regexp/parse/unicode.rkt
```diff
@@ -67,7 +73,8 @@
                            "unrecognized property name in `\\~a{}`: `~a`"
                            (integer->char p-c)
                            (list->string (map integer->char l)))]))
-    (values (rx:unicode-categories categories (= p-c (char->integer #\p)))
+    (define prop-negated? (= p-c (char->integer #\p)))
```
97jaz: Er, this is clearly misnamed, since it's true when the property is _not_ negated. And if it were correctly named, `xor` would be the wrong truth function. (I'd want the complement instead.)
So the current code works but is confusing. I'll change that...

97jaz: Better now.

mflatt:
Thanks!

PR #2811, opened 2019-09-05
Define _GNU_SOURCE instead of __USE_GNU

pmatos:
__USE_GNU should only be used internally by features.h and never user-defined. _GNU_SOURCE is the user visible macro.

Related to #2797


pmatos:
There's a problem here...

pmatos:
There are some subtleties due to the order in which headers show up. I will hold on this one until we discuss the future of our slice within #2797

pmatos:
This is actually not required - so closing without merging.

PR #2817, opened 2019-09-12
Fix headers so they don't break compilation with modern C++.

elfprince13:
Add space between string literals.
Necessary so that modern C++ compilers don't misinterpret `SPLS_LINUX` as a user-defined literal type.


pmatos:
@elfprince13 Thanks for the report. Have you experienced a problem with any modern compiler?

elfprince13:
@pmatos - Yes, when compiling a project with Clang 8, and `--std=c++17` 

n.b. I'm not trying to compile racket itself as C++17; however, I have a file that needs to `#include <scheme.h>`, which has `sconfig.h` as a transitive include.

pmatos:
@mflatt thanks for getting here before I did. :)

PR #2819, opened 2019-09-14
Add macro-introduction scopes when expanding rename transformers

### LINKED ISSUE #2818: Macro-introduced uses of rename transformer references lose their macro-introduction scope and are considered syntax-original?
When a rename transformer is expanded, it is replaced with its target. This can affect `syntax-original?`-ness, since macro-introduction scopes may be lost. For example, see how Check Syntax interprets this program:

Screen Shot 2019-09-14 at 13 07 57

I think the solution to this problem is to treat rename transformer expansions the same way as other macro applications, adding a fresh macro-introduction scope each time. Uses that actually are original should still be treated as references to the original variable by Check Syntax because the expansion will leave behind an `'origin` property, and `free-identifier=?` equivalences installed by rename transformers will cause Check Syntax to treat that as a reference to the original variable, too.

A side-effect of that change would be that rename transformers marked `'not-free-identifier=?` would *not* be treated as references to their targets by Check Syntax any longer. However, I think that’s actually probably the right behavior, so I think that’s okay.

lexi-lambda:
Fixes #2818.


mflatt:
Merged (manually to resolve a conflict)

lexi-lambda:
Thanks for adding this `history` note for me; I just hadn’t gotten around to it yet.

Issue #2563, opened 2019-03-26
for + malformed in-value gives internal error message


florence:
`(for ([x (in-value 1 2)]) x)` gives `racket/collects/racket/private/for.rkt:228:13: ?: bad syntax in: ((x) (in-value 1 2)) `

whereas `(in-value 1 2)` gives `in-value: arity mismatch;`

This doesn't happen for `in-list` so I suspect this is specific to `in-value`

gus-massa:
I think the difference is that `in-list` has at the bottom a clause `[_ #f]`. When the special syntax gets a wrong number of arguments, it calls the fallback that is a function, that generates the error. [private/for.rkt#L2050](https://github.com/racket/racket/blob/master/racket/collects/racket/private/for.rkt#L2050)

I didn't test it, but I think the fix is to add a symilar clause at the bottom of `in-value` [private/for.rkt#L2118](https://github.com/racket/racket/blob/master/racket/collects/racket/private/for.rkt#L2118).

The same modification should be made to most `in-whatever` forms, because they have the same problem.

gus-massa:
Fixed in https://github.com/racket/racket/commit/3907f35d1d9e3ef77302a98f2b25ec5b2884e797

PR #2821, opened 2019-09-15
Fix error of malformed in-values in a for clause

gus-massa:
Fix error of malformed `in-values` in a `for` clause.

For example

    (for ([x (in-value 1 2)]) x)

should raises a run time error, not a syntax error.

Fix similar error in other `in-something` macros, and add test for all of them.

Closes https://github.com/racket/racket/issues/2563

---

Also, fix name the of `in-directory`, when used as a function outside a `for`.

---

There is a second commit I'll post shortly. I tried to move the repeated pattern of test for the bad sequences generators to a macro in for-utlil.rkt . The problem is that I get an error 

    ERROR: err/rt-test: undefined;
     cannot reference an identifier before its definition
      in module: top-level

For the successful cases there is similar code with `test`. I think the problem is that `test` is a function but `err/rt-test` is a macro. I don't know how to solve this, but the easy solution is to ignore the second commit.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/file.rktl
+++ b/pkgs/racket-test-core/tests/racket/file.rktl
@@ -1979,6 +1979,8 @@
           (parameterize ([current-directory tmp-dir])
             (for/hash ([f (mk)])
               (values f #t)))))
+  (err/rt-test (for/list ([f (in-directory tmp-dir #f '?)]) f))
+  (err/rt-test (in-directory tmp-dir #f '?))
   (delete-directory/files tmp-dir))
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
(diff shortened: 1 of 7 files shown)


mflatt:
LGTM

mflatt:
Oh, I see what you mean about the second commit. Is it practical to use `thunk-error-test` directly? Or maybe that's the wrong direction, since it seems like a hack that "test-util.rkt" generates references to top-level variables like `test` by referring to identifiers that are otherwise unbound.

gus-massa:
Merged the first commit in  https://github.com/racket/racket/commit/3907f35d1d9e3ef77302a98f2b25ec5b2884e797

I think that to use `thunk-error-test` I need to duplicate all the cases in `err/rt-test`. (It has a two cases, when the expression is in a value position and when it is in an ignored position, perhaps some day I'll add a case for the test position...)

One possibility is to rename "test-util.rkt" to "test-util.rkt**L**" and use `#lang load` there too, but I have to try it because I don't understand the subtle interactions.

Issue #2808, opened 2019-09-03
MZ_NORETURN causing problems in obscure configurations


pmatos:
There are some more obscure - less used - configurations where the `MZ_NORETURN` definition is causing problems.
1. 32-bit Windows build fails with Visual Studio 2017;
2. backtraces are more often disabled on 64-bit Linux;

@mflatt thinks 1 might be due to a compiler bug and 2 due to a problem in our slice of libunwind. I will investigate and propose a PR. Due to 2, http://drdr.racket-lang.org/52498/racket/share/pkgs/drracket-test/tests/drracket/repl-test-raw.rkt started failing.

pmatos:
2. might not be the case. I have run the test `repl-test-raw.rkt` on checkout b7654d9a84 which predates any `NORETURN` work and the test was already failing then.

pmatos:
Doing a bisect to confirm.

pmatos:
I was confused. The problem is indeed `NORETURN` specifically in the `scheme_raise_exn` procedure:
```
3e301d22b50b210f6b3f36b01e037e07a20a3e74 is the first bad commit
commit 3e301d22b50b210f6b3f36b01e037e07a20a3e74
Author: Paulo Matos <pmatos@linki.tools>
Date:   Wed Feb 13 15:20:04 2019 +0100

    Mark scheme_raise_exn as noreturn - regenerate schemex
    
    scheme_raise_exn raises an exception and doesn't return.
    Static analysis tools find a huge amount of problems with regards
    to memory leaks that are actually false positives because the tools
    are not aware the function does not return. Marking it as such aids
    further inspection of real problems.

 racket/src/racket/src/schemef.h | 2 +-
 racket/src/racket/src/schemex.h | 2 +-
 2 files changed, 2 insertions(+), 2 deletions(-)
```

mflatt:
Would it be ok to go back to a preprocessor variable to enable annotations, maybe something like `MZ_PRECISE_RETURN_SPEC`?

pmatos:
Yes, let's do that for now. My time has been limited to look at this. I am sorry. If you don't do this in the meantime, I add a PR first thing tomorrow. 

On 10 September 2019 16:59:18 CEST, Matthew Flatt <notifications@github.com> wrote:
>Would it be ok to go back to a preprocessor variable to enable
>annotations, maybe something like `MZ_PRECISE_RETURN_SPEC`?
>
>-- 
>You are receiving this because you were assigned.
[...]

PR #2823, opened 2019-09-16
Enable MZ_NORETURN annotation only if MZ_PRECISE_RETURN_SPEC is defined

pmatos:
Unfortunately, MZ_NORETURN spec is causing a few problems - see #2808
It would be great to fix these but due to lack of time, this is a
workaround that should keep things working until all supported
configurations accept MZ_NORETURN properly.

### DIFF
--- a/racket/src/racket/include/scheme.h
+++ b/racket/src/racket/include/scheme.h
@@ -253,7 +253,9 @@ extern "C"
 #endif
 
 #if !defined(MZ_NORETURN)
-# if defined(__GNUC__) || defined(__clang__)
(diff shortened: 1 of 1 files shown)


pmatos:
Looks like this should work. I was a bit concerned about some changes to the return values of the functions, but VS seems to compile it just fine.

I would have really liked to improve this further but I am already being spread so thin that I will focus on my future Racket contribution efforts on RacketCS only since that'll be in the short-term future anyway.

pmatos:
I am sorry this took so long. Sorted now!

PR #2824, opened 2019-09-18
Remove me from .mailmap

gcr:
hey everyone! my name changed recently, so i'm trying to remove my old name from the Internet.

I'm not sure how or where this file is used and I'm honored to be listed among the likes of Matthew and Eli and Sam and etc, but my last contribution was a few years back...

thanks!

(i, the author of this patch, hereby assign copyright of this change to the racket developers)


samth:
Hi @gcr,

We're happy to remove you from the mailmap, but note that your old name is still in various commits. If you would like a different name in the mailmap so that your old name is no longer shown in the log, or for us to use a different name in other places where your name appears, we can do that too. 

Additionally, if you would prefer to have this discussion by email instead of on GitHub for whatever reason, my email address is samth@indiana.edu.

PR #2839, opened 2019-09-27
Specify that inbound contributions are licensed under MIT/Apache/LGPL.

samth:
This is a step towards re-licensing, as described in #1570 but not an actual change in the license yet (we still need permission from some people). This simply states that contributions from here on are licensed by their contributors with _all of_ MIT, Apache 2, and LGPL. That way, further contributions after today don't require asking new people for permission.

This also adds a CONTRIBUTING.md file with minimal information other than licensing. It's not intended to stand in the way of other improvements to that file in the future.

The LGPL license stays in its current location to avoid having to modify the build right now, but the new licenses are at the top of the repository.

Thanks to our counsel from the SFC for help preparing this, and to @otherjoel and @zyrolasting for their assistance with this project.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -1,0 +1,10 @@
+Contribute to Racket by submitting a pull request, joining the
+[development mailing list](https://lists.racket-lang.org), or visiting
+the IRC or Slack channels.
+
+By making a contribution, you are agreeing that your contribution is
+licensed under the LGPLv3, Apache 2.0, and MIT licenses. Those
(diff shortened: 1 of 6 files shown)


gus-massa:
Will this be added to all the repos in https://github.com/racket ?

zyrolasting:
:100: Thanks so much for getting everyone on the same page.

samth:
@gus-massa Yes, either this or racket/wxme#3 will be added to all the repositories.

PR #2842, opened 2019-10-01
Simplify some code and avoid string conversion.

samth:


racket/src/expander/eval/collection.rkt
```diff
@@ -422,13 +416,8 @@
...
                                                              check-compiled?)
                                 (let ([alt-file-name
-                                       (let* ([file-name (if (path? file-name)
-                                                             (path->string file-name)
-                                                             file-name)]
-                                              [len (string-length file-name)])
-                                         (and (len . >= . 4)
-                                              (string=? ".rkt" (substring file-name (- len 4)))
-                                              (string-append (substring file-name 0 (- len 4)) ".ss")))])
+                                       (and (regexp-match '#rx"[.]rkt$" (path->bytes file-name))
```
mflatt: If `file-name` can be a string, then `path->bytes` would fail. Is it definitely a path?

Also, using a regexp is probably slower. Do we care? (Converting to bytes and back is probably fastest.)

samth: I first wrote it with `path-get-suffix` but then switched to this because that pattern was used elsewhere. I'll change it to use `subbytes` there and elsewhere.

PR #2843, opened 2019-10-01
expander: improve error reporting of conflicts during require

Bogdanp:
Adds an additional line to the error message that is raised when a
required module provides a binding that is already provided by another
required module.  The additional line displays the name of the first
module that provides the binding.

The error before this change:

    tmp/c.rkt:4:9: module: identifier already required
      at: x
      in: "b.rkt"
      location...:
       tmp/c.rkt:4:9

and after:

    tmp/c.rkt:4:9: module: identifier already required
      at: x
      in: "b.rkt"
      also provided by: "a.rkt"
      location...:
       tmp/c.rkt:4:9


racket/src/expander/demo.rkt
```diff
@@ -736,10 +736,24 @@
 
 (check-error
  (eval-module-declaration '(module m '#%kernel
-                            (#%require '#%kernel 
+                            (#%require '#%kernel
                                        'definition-shadows-initial-require)))
  #rx"already required")
 
+(eval-module-declaration '(module require-conflict-is-sourced-a '#%kernel
```
samth: These tests should maybe be in `pkgs/racket-test-core/tests/module.rktl` since this file is mostly a demo.

Bogdanp: Thanks! I've made the change.

mflatt:
@Bogdanp  This is a great improvement! But would it be ok to move the message addition to the message-suffix argument of `raise-syntax-error`, so that it appears after the "at" and "in" part? Those two parts seem more useful to start than the extra info. (If you're ok with the change but would prefer that I do that as a second step, I'd be happy to merge your change so far as-is.)

mflatt:
@Bogdanp One more thought: is `~v` the format for the error? I'd expect `~.s` to work better for symbolic module paths, `lib` and `file` paths, etc., but I haven't tried it.

Bogdanp:
> move the message addition to the message-suffix argument of raise-syntax-error, so that it appears after the "at" and "in" part?

I wasn't aware that `raise-syntax-error` took a message suffix. That's definitely a lot better.

> I'd expect `~.s` to work better for symbolic module paths, lib and file paths, etc., but I haven't tried it.

You're right. That does seem more appropriate.

I've updated the commit. The error now looks like this:

```
tmp/c.rkt:4:9: module: identifier already required
  at: x
  in: "b.rkt"
  also provided by: "a.rkt"
  location...:
   tmp/c.rkt:4:9
```

and

```
tmp/submod.rkt:12:9: module: identifier already required
  at: x
  in: (quote b)
  also provided by: (quote a)
```

Thanks!

mflatt:
Thanks!

Issue #2803, opened 2019-08-30
gen:dict methods break when alists have duplicate keys


LiberalArtist:
Association lists are [dictionaries](https://docs.racket-lang.org/reference/dicts.html) in the sense of `gen:dict` and `racket/dict`. They are specified to compare keys using `equal?`. However, unlike other types of dictionaries, association lists cannot guarantee key uniqueness by construction.

While `dict-ref` and `dict-set` handle duplicate keys consistently (they use the first mapping), `dict-remove`, the iteration methods, and many of the derived dictionary methods fail to cope with association lists that contain duplicate keys.

Examples:
```racket
#lang racket

(define problem
  '([a . 1]
    [a . 2]
    [a . 3]))

(dict-ref problem 'a)
;; -> 1
;; OK

(dict-set problem 'a 4)
;; -> '([a . 4] [a . 2] [a . 3])
;; expected: either this or '([a . 4])

(dict-remove problem 'a)
;; -> '([a . 2] [a . 3])
;; expected: '()

(dict-count problem)
;; -> 3
;; expected: 1

(dict-keys problem)
;; -> '(a a a)
;; expected: '(a)

(dict-values problem)
;; -> '(1 2 3)
;; expected: '(1)

(dict->list problem)
;; -> problem
;; expected: '([a . 1])

(for ([{k v} (in-dict problem)])
  (cons k v))
;; -> problem
;; expected: '([a . 1])
```

muscar:
Hi,

I'm having a look at this, and will submit a PR shortly.

Alex

Issue #1710, opened 2017-06-02
racket/dict blamed for buggy method implementations


LiberalArtist:
As discussed in the [mailing list](https://groups.google.com/d/msg/racket-users/3fTKqqcOnxs/hU9XqZpsAgAJ), implementing a `gen:dict` method in a way that breaks its contract results in blame for `racket/dict`. For example:
```
(struct bad-dict ()
  #:methods gen:dict
  [(define (dict-set this k v)
     'not-a-new-dict)])

(dict-set (bad-dict) 'a #t)

dict-set: broke its own contract
  promised: dict?
  produced: 'not-a-new-dict
  in: the _r result of
      (->i
       ((d (dict-implements/c dict-set))
        (k (d) (dict-key-contract d))
        (value (d) (dict-value-contract d)))
       (_r dict?))
  contract from: <collects>/racket/dict.rkt
  blaming: <collects>/racket/dict.rkt
   (assuming the contract is correct)
  at: <collects>/racket/dict.rkt:190.2
```
Should `racket/dict` protect itself from method implementers?

PR #2846, opened 2019-10-03
Fix gen:dict methods for alists with duplicate keys

muscar:
This fixes https://github.com/racket/racket/issues/2803.

Note that while this works, the implementation is not particularly fast. Especially for iterators.

### DIFF
--- a/racket/collects/racket/private/dict.rkt
+++ b/racket/collects/racket/private/dict.rkt
@@ -2,6 +2,7 @@
 
 (require racket/private/generic ; to avoid circular dependencies
          racket/private/generic-methods
+	 racket/list
          racket/vector
          (only-in racket/private/hash paired-fold)
          (for-syntax racket/base))
@@ -131,7 +132,7 @@
      [else
       (let ([a (car xd)])
         (if (equal? (car a) key)
-            (cdr xd)
+            (loop (cdr xd))
             (cons a (loop (cdr xd)))))])))
 
 (define (vector-iterate-first d)
@@ -159,23 +160,23 @@
 (define (assoc-count d)
   (unless (assoc? d)
     (raise-argument-error 'dict-count "dict?" d))
-  (length d))
+  (length (assoc-keys d)))
 
 (struct assoc-iter (head pos))
 
 (define (assoc-iterate-first d)
   (unless (assoc? d)
     (raise-argument-error 'dict-iterate-first "dict?" d))
-  (if (null? d) #f (assoc-iter d d)))
+  (if (null? d) #f (assoc-iter d (assoc->list d))))
 
 (define (assoc-iterate-next d i)
   (cond
     [(and (assoc-iter? i)
-          (eq? d (assoc-iter-head i)))
+          (equal? d (assoc-iter-head i)))
      (let ([pos (cdr (assoc-iter-pos i))])
        (if (null? pos)
            #f
-           (assoc-iter d pos)))]
+           (assoc-iter d (assoc->list pos))))]
     [(assoc? d)
      (raise-mismatch-error
       'dict-iterate-next
@@ -251,16 +252,28 @@
     (proc (car x) (cdr x))))
 
 (define (assoc-keys d)
-  (for/list ([x (in-list d)])
-    (unless (pair? x)
-      (raise-argument-error 'dict-keys "dict?" d))
-    (car x)))
+  (remove-duplicates
+    (for/list ([x (in-list d)])
+      (unless (pair? x)
+        (raise-argument-error 'dict-keys "dict?" d))
+      (car x))))
+
+(define (assoc-map-unique f d)
+  (let loop ([xd d]
+	     [keys (assoc-keys d)])
+    (cond
+      [(or (null? xd) (null? keys)) null]
+      [else
+       (let ([a (car xd)])
+         (if (equal? (car a) (car keys))
+             (cons (f a) (loop (cdr xd) (cdr keys)))
+             (loop (cdr xd) (cdr keys))))])))
 
 (define (assoc-values d)
-  (for/list ([x (in-list d)])
-    (unless (pair? x)
-      (raise-argument-error 'dict-values "dict?" d))
-    (cdr x)))
+  (assoc-map-unique cdr d))
+
+(define (assoc->list d)
+  (assoc-map-unique (lambda (x) x) d))
 
 (define (fallback-copy d)
   (unless (dict-implements? d 'dict-clear dict-set!)
@@ -406,7 +419,7 @@
     (define dict-for-each assoc-for-each)
     (define dict-keys assoc-keys)
     (define dict-values assoc-values)
-    (define dict->list values)
+    (define dict->list assoc->list)
     (define dict-empty? null?)
     (define dict-clear assoc-clear)])
   #:defaults ()
--- a/pkgs/racket-test/tests/generic/alist.rkt
+++ b/pkgs/racket-test/tests/generic/alist.rkt
@@ -16,15 +16,30 @@
    (define (dict-count dict)
      (length (remove-duplicates (alist-v dict) #:key car)))])
 
-
 (module+ test
   (require rackunit)
 
   (define d1 '((1 . a) (2 . b)))
+  (define d2
+    '([a . 1]
+      [a . 2]
+      [a . 3]))
 
   (check-true (dict? d1))
   (check-eq? (dict-ref d1 1) 'a)
   (check-equal? (dict-count (dict-remove d1 2)) 1)
   (check-false (dict-mutable? d1))
   (check-true (dict-can-remove-keys? d1))
-  (check-true (dict-can-functional-set? d1)))
+  (check-true (dict-can-functional-set? d1))
+
+  (check-eq? (dict-ref d2 'a) 1)
+  (check-equal? (dict-set d2 'a 4) '([a . 4] [a . 2] [a . 3]))
+  (check-equal? (dict-remove d2 'a) '())
+  (check-eq? (dict-count d2) 1)
+  (check-equal? (dict-keys d2) '(a))
+  (check-equal? (dict-values d2) '(1))
+  (check-equal? (dict->list d2) '([a . 1]))
+  (check-equal?
+    (for/list ([{k v} (in-dict d2)])
+      (cons k v))
+    '([a . 1])))


muscar:
Thanks, @bennn. I'll try to see what's up with the Travis failure.

Thanks for pointing out the formatting issues, @lnicola.

muscar:
The `unsafe` section passes for me locally :/

racket/collects/racket/private/dict.rkt
```diff
@@ -251,16 +252,28 @@
...
-    (car x)))
+  (remove-duplicates
+    (for/list ([x (in-list d)])
+      (unless (pair? x)
+        (raise-argument-error 'dict-keys "dict?" d))
+      (car x))))
+
+(define (assoc-map-unique f d)
+  (let loop ([xd d]
+             [keys (assoc-keys d)])
```
LiberalArtist: In line with @bennn's comment, I think it might turn out to be expensive to implement `assoc-map-unique` in terms of `assoc-keys` and `remove-duplicates`: that seems like a lot of extra `equal?` tests and extra intermediate lists. I'm still thinking about it, but I can imagine an `assoc-fold-unique` that could be used to implement `assoc-count`, `assoc-keys`, `assoc-values`, and `assoc->list`. I think the key insight `remove-duplicates` is to use `member` for lists of less than length 40 and a hash-table otherwise, whereas most of the complexity is from needing to handle general cases:
https://github.com/racket/racket/blob/f8bc4e8fa1d37d90a66e0e975dcbc244cd3c9796/racket/collects/racket/list.rkt#L403-L406

Alternatively, we presumably expect the common case to be lists with no duplicate keys, so maybe it would pay off to avoid allocation in those cases.

That said, I think making these functions correct is good and valuable, even if more could potentially be done to make them efficient.

racket/collects/racket/private/dict.rkt
```diff
@@ -159,23 +160,23 @@
...
   (unless (assoc? d)
     (raise-argument-error 'dict-iterate-first "dict?" d))
-  (if (null? d) #f (assoc-iter d d)))
+  (if (null? d) #f (assoc-iter d (assoc->list d))))
 
 (define (assoc-iterate-next d i)
   (cond
     [(and (assoc-iter? i)
-          (eq? d (assoc-iter-head i)))
+          (equal? d (assoc-iter-head i)))
```
LiberalArtist: I don't think I understand the reason for changing `eq?` to `equal?` here. I think `(dict-iterate-next d i)` is only allowed with an `i` from a previous call to `(dict-iterate-first d)` for an `eq?` `d`, so I'm not sure why the potentially-very-expensive structural equality check would be needed.

muscar:
@LiberalArtist thanks for the feedback. I'll have a look at your suggestions tomorrow.

racket/collects/racket/private/dict.rkt
muscar: You're right. I've implemented `assoc-fold-unique` which uses an immutable `set` to keep track of the keys seen. The set could be made mutable, but it's unlikely that this would make a significant difference for small alists.

I've re-implemented the `assoc-count`, `assoc-keys`, `assoc-values` and `assoc->list` functions in terms of `assoc-fold-unique`.

Note that this has `foldl` semantics, and will build the resulting lists in reverse, but that should be OK since `dict`s don't guarantee the order anyway.

racket/collects/racket/private/dict.rkt
muscar: This was a misunderstanding on my side of how the code works. I've reverted the change. Thanks.

racket/collects/racket/private/dict.rkt
```diff
@@ -244,23 +196,39 @@
       (raise-argument-error 'dict-map "dict?" d))
     (proc (car x) (cdr x))))
 
+(define (assoc-fold-unique f init d)
+  (unless (assoc? d)
+    (raise-argument-error 'assoc-fold-unique "dict?" d))
+  (let loop ([xd d]
+             [acc init]
+             [seen (set)])
```
Bogdanp: Feel free to ignore this, but small lists (< ~100 elements) tend to perform a lot better than sets. It might be worthwhile to use one or the other based on the size of `d`.

Here's a microbenchmark that shows the difference: https://gist.github.com/Bogdanp/e67c9d1c8c33d6b0df2fbbb24b8c2e29

bennn: I added hashes with `#true` keys to the microbenchmark:
https://gist.github.com/bennn/28cf78e1c832603b2fc8ba1c8f245c22

And it looks like both of those perform better than sets. So, I'd go with immutable hashes and change `(set-add ....)` to `(hash-set ... #true)`

muscar: Hi @bennn and @Bogdanp. I've changed the code to use immtable hash sets.

muscar:
Hi @LiberalArtist, @bennn. The PR is green. I can't seem to merge it myself. Caa one of you please merge it?

bennn:
Yes, but I was first going to wait a day or two for other comments

racket/collects/racket/private/dict.rkt
```diff
@@ -2,6 +2,7 @@
 
 (require racket/private/generic ; to avoid circular dependencies
          racket/private/generic-methods
+         racket/set
```
97jaz: I don't think this is used, is it?

racket/collects/racket/private/dict.rkt
```diff
@@ -244,23 +196,39 @@
       (raise-argument-error 'dict-map "dict?" d))
     (proc (car x) (cdr x))))
 
+(define (assoc-fold-unique f init d)
+  (unless (assoc? d)
+    (raise-argument-error 'assoc-fold-unique "dict?" d))
```
LiberalArtist: The `assoc?` check is repeated both here and in `assoc-count`. I think it should either always be done in the public function, or always done here, but using a `who` argument to report errors in terms of the public name.

racket/collects/racket/private/dict.rkt
```diff
@@ -244,23 +196,39 @@
       (raise-argument-error 'dict-map "dict?" d))
     (proc (car x) (cdr x))))
```
LiberalArtist: I could see a counter-argument, but I think `accoc-map` and `assoc-for-each` should probably also use `assoc-fold-unique`.

racket/collects/racket/private/dict.rkt
```diff
@@ -244,23 +196,39 @@
...
 
+(define (assoc-fold-unique f init d)
+  (unless (assoc? d)
+    (raise-argument-error 'assoc-fold-unique "dict?" d))
+  (let loop ([xd d]
+             [acc init]
+             [seen (make-immutable-hash)])
```
LiberalArtist: It might be nice to implement the optimization from `remove-duplicates` to use a plain list for seen when the given list is less than length 40 (with `member` and `cons` instead of `hash-has-key?` and `hash-set`), and perhaps also to special-case lists of length ≤1.

bennn: Can you provide benchmarks (or larger programs) to check whether this would pay off?

(I'm not even sure that the previous change, removing `remove-duplicates`, is going to save much time on the programs that end up using alist dicts.)

LiberalArtist: As I think about @samth's comment, one way of considering it would be that, if you are using alist dicts, you've made a choice for some reason to not use hash tables, and you've done that knowing the performance trade-offs (e.g. potentially faster for small dicts, likely much slower for very big ones). That's pushing me further toward thinking this code should use `member` and `cons` rather than hash-tables (as `remove-duplicates` does for small-ish lists, and as @Bogdanp also suggested). I might even skip the `(if (< (length d) 40) ... ...)` check, on the grounds that there is an explicit choice here not to use hash-tables.

LiberalArtist: But overall, I think having the dict methods behave correctly is a much bigger deal than too much performance tuning. Especially for alists-as-dicts, you've already signed up to pay for generic dispatch and likely to get less inlining, which many of the list-specific functions go to great lengths to promote.

bennn: Is it ok to merge this, then, and leave the tuning for another patch?

LiberalArtist: Sounds good to me! (I don't have commit access.)

samth:
Unfortunately this broke some existing code, see https://github.com/greghendershott/aws/issues/64

greghendershott:
Things like `form-urlencoded->alist` -- which you might use for HTML forms or for HTTP query parameters -- can produce association lists with "duplicate keys". If someone then uses `in-dict` as a convenient way to iterate the alist, it should return all the mappings. At least, it has done so in Racket for many years, going back as early as 6.2.

I haven't thought about this very deeply, but _maybe_ there should be a _new_ interface for dictionaries that add stricter constraints than association lists can implement --- `gen:strict-dict` or some such??

Also, the slower performance seems quite noticeable in [this graph from DrDr](http://drdr.racket-lang.org/53149/racket/share/pkgs/aws/aws/sigv4.rkt) assuming I'm understanding the axes correctly:

canvas

samth:
The performance changes in that graph seem to be in November 2018 and April 2019.

greghendershott:
Ah OK. So disregard the performance part. :smile:

bennn:
Ok, it sounds like we need to revert this PR and add an `in-assoc`, so new programs don't need to rely on the old, strange behavior of `in-dict` (strange = all dicts except assoc lists ignore duplicate keys).

I guess we could revert only the changes to `in-dict` and hope there are no other issues, but I'd rather revert all.

jackfirth:
I like the suggestion @greghendershott gave of adding a new, stricter dictionary interface. The fact that the current one allows and encourages this kind of thing makes it pretty busted IMO.

AlexKnauth:
I don't think we should add a stricter dictionary interface, I think we should add a less-strict multimap interface, which explicitly allows multiple entries for the same key.

```
dict-ref : [Dictof K V] K -> V
multimap-ref : [Multimapof K V] K -> [Setof V]
```

jackfirth:
Do both. A less-strict multimap interface [would be great](https://docs.racket-lang.org/rebellion/Multidicts.html), but on its own it wouldn't fix the problems of the current dictionary interface.

LiberalArtist:
This is unfortunate.

Re @greghendershott and @jackfirth's comments about a new dictionary interface, I just remembered https://github.com/racket/racket/issues/1710 (`gen:dict` doesn't fully protect itself from buggy method implementations). I haven't tried it yet, but I wonder if it's actually true that "all dicts except assoc lists ignore duplicate keys"—I suspect you may be able to implement `gen:dict` in a way that mimics the current behavior for assoc lists.

I'm not completely sure what the right thing to do is. At a minimum, I guess the docs should be changed to reflect what being a the dictionary interface actually does or doesn't guarantee. I do think it would be confusing for `in-dict` to be a special case. On the other hand, I wonder if it really makes sense for `dict-remove` to remove only the first mapping for a key, which is the current behavior for alists.

I've actually written `in-alist`, `in-alist-keys`, and `in-alist-values`, though they would need some refactoring to go in `#lang racket`.

mflatt:
One month has passed, and it's now only one month until the v7.6 branch, so this needs to get resolved.

mflatt:
Looks like the current plan is to revert: #2901 

I'll add a test and documentation improvements if no one else does. Otherwise, this is a last call for alternatives.

PR #2848, opened 2019-10-06
Avoid incorrect use of `cadr`.

### LINKED ISSUE #2847: Crashes when attempting to construct (unquote [value])
It is impossible to construct (unquote 4) without causing Dr. Racket to crash.  Running `(list 'unquote 4)` causes the program to hang.

samth:
Fixes #2847.

### DIFF
--- a/racket/collects/racket/pretty.rkt
+++ b/racket/collects/racket/pretty.rkt
@@ -779,10 +779,13 @@
             (cond
               [(and (or (eq? (car expr) 'unquote)
                         (eq? (car expr) 'unsyntax))
-                    (symbol? (cadr expr)))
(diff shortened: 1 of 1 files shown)


gus-massa:
I didn't find other similar error. I hope this can be merged for 7.5.

jbclements:
LGTM. Interesting shadowing issue here; in some ways, it's another roundabout argument against shadowing; if the arguments had been named something other than car & cdr, it might have been less tempting to write cadr.

rfindler:
Or maybe the right thing is to use drr (or add stuff to the emacs mode) so
you add the binder and then just rename!

Robby

PR #2849, opened 2019-10-06
Track all thread in CPU time reported by `time`.

samth:
Not sure if this is the right fix, but it works correctly for the program I had.


mflatt:
Thanks! But I have a patch — not yet pushed — that does something similar and avoids `statistics` (to avoid its allocation overhead).

samth:
I wasn't sure if you had gotten to it yet, and also wanted to check if this was the right solution.

PR #2851, opened 2019-10-07
Ensure errkind_str is not used uninitialized

pmatos:

PR #2852, opened 2019-10-08
Enable MZ_ASSERT on Linux

pmatos:
Related to #2851


mflatt:
I'm in favor of this change, but let's wait until after the v7.5 branch (which should happen about now), just in case.

pmatos:
No problem - merging right after 7.5 branch happens.

pmatos:
Release branches created - now merging.

PR #2859, opened 2019-10-16
Fix libcrypto native-lib install on macOS 10.15

bubba:
Catalina now aborts when dlopen'ing the versionless libcrypto/libssl dylibs.
This removes the versionless libs on macOS from the list of versions.

https://forums.developer.apple.com/thread/119429
https://github.com/racket/racket/issues/2018#issuecomment-542835900


mflatt:
Thank you!

PR #2862, opened 2019-10-18
Add set-filter to racket/set

whichxjy:
This PR adds `set-filter` to `racket/set`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sets.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sets.scrbl
@@ -25,7 +25,7 @@ datatypes are all sets:
 
 A @deftech{hash set} is a set whose elements are compared via @racket[equal?],
 @racket[eqv?], or @racket[eq?] and partitioned via @racket[equal-hash-code],
-@racket[eqv-hash-code], or @racket[eq-hash-code].  A hash set is either
+@racket[eqv-hash-code], or @racket[eq-hash-code]. A hash set is either
 immutable or mutable; mutable hash sets retain their elements either strongly
 or weakly.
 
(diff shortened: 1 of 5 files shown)


jackfirth:
Sets are [sequences](https://docs.racket-lang.org/reference/sequences.html), so why not use [`sequence-filter`](https://docs.racket-lang.org/reference/sequences.html#%28def._%28%28lib._racket%2Fsequence..rkt%29._sequence-filter%29%29)?

whichxjy:
> Sets are [sequences](https://docs.racket-lang.org/reference/sequences.html), so why not use [`sequence-filter`](https://docs.racket-lang.org/reference/sequences.html#%28def._%28%28lib._racket%2Fsequence..rkt%29._sequence-filter%29%29)?

1. Since there is a [map function for set](https://docs.racket-lang.org/reference/sets.html?q=set-map#%28def._%28%28lib._racket%2Fset..rkt%29._set-map%29%29) even if the `sequence-map` existed, I was wondering if using a custom filter for set would improve the performance.
2. The result of `sequence-filter` is also a sequence, but using `set-filter` will get a list without using `sequence->list`.

jackfirth:
Generally we'd like to encourage using sequences over concrete lists, so ideally you would only rarely need to use `sequence->list`. Is there a list function you wanted to use that didn't have a generic alternative in `racket/sequence`, such as `apply`?

As for `set-filter`, it's probably not a good precedent. I don't think it has any performance advantages. It's docs ought to be updated to ask clients to consider using `sequence-filter` instead.

whichxjy:
Actually, I want to perform `filter` and `symmetric-difference` operations on some sets. It would be great if there's a `set-filter` function that returns a set (but why does `set-map` return a list?). If I use sequences instead of sets, there's no operation about `symmetric-difference` in `racket/sequence`. It seems like adding something like `sequence-symmetric-difference` to `racket/sequence` is not a good idea.

jackfirth:
I think a good way to go about that would be to add a `sequence->set` operation. Then you'd have this:

```
(define filtered (sequence->set (sequence-filter some-set))
(symmetric-difference filtered other-set)
```

And you'd be able to do the same thing for any of the other sequence-processing functions, without needing to make a set-specific version of each function.

samth:
Unfortunately, going from a set to a sequence and back will be much less efficient than creating a set in the first place.

jackfirth:
@samth Agreed. In my experience though, it's pretty rare in practice to have only _one_ operation you want to perform on a collection _and_ you want to end up with the same type of collection you started with. Most collection-manipulating code has multiple steps to it like mapping and filtering, and often you're building a different kind of collection than you started with. In those circumstances, the laziness of the sequence operators will be more useful than the efficiency gains of avoiding conversions to and from sequences.

sorawee:
I might be pointing out the obvious, but there's `for/set` which is capable of mapping, filtering, producing a set, and iterating over a set. I do think that the function variants in `racket/set` is a mess (why are argument positions for `set-map` and `map` different?) and probably should be cleaned up for Rhombus.

jackfirth:
@whichxjy Would using `for/set` work for you? Something like this:

```racket
(define filtered
  (for/set ([v (in-immutable-set some-set)]
             #:when (should-include-in-filtered? v))
    v))
```

The `in-immutable-set` part is optional, but using it makes `for/set` faster.

whichxjy:
@jackfirth It works. Thank you!

PR #2868, opened 2019-10-23
fix for (setup) when (current-library-collection-links) contains hash 

djholtby:
setup/setup does not treat `current-library-collection-links` correctly when it contains a hash table.

`setup-core` expects hash table's values to be `path?`, but they are actually `(listof path?)`

Additionally, it uses `string-split` with incorrect parameter order, and calls `cc!` with a `(listof string?)` when `cc!` is expecting `(listof path?)`


mflatt:
Thanks!

PR #2878, opened 2019-10-29
added udp-ttl and udp-set-ttl!

dstorrs:
Adds the udp-ttl and udp-set-ttl! functions.  These are essential for the port-prediction NAT-traversal algorithm described here: http://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.455.3700&rep=rep1&type=pdf


samth:
This is quite complete, but we'll probably need to add both tests and documentation. Do you want to do that? Otherwise I can probably do it.

dstorrs:
On Tue, Oct 29, 2019 at 3:43 PM Sam Tobin-Hochstadt
<notifications@github.com> wrote:
>
> This is quite complete, but we'll probably need to add both tests and documentation. Do you want to do that? Otherwise I can probably do it.

I'd rather not, if that's okay.  I need this for work and it's been a
fair significant diversion already, so I'd prefer not to spend more
time on it.   Also, see my previously mentioned position on the
challenges of modifying the Racket documentation.

mflatt:
It looks like the `IP_TTL` option should use `int` instead of `u_char` (unlike `IP_MULTICAST_TTL`). I'll make that change while adding tests and docs.

dstorrs:
You're brilliant. Thank you, Matthew.

PR #2879, opened 2019-10-30
Update complexity discussion in hashes.scrbl

elfprince13:
The current documentation is several years out of date (as of 6ac8bcd441ee23ecd604e431ce959cc2c33a6c3a).

Note that it's not uncommon for HAMTs to actually have bounded depth, and end in buckets, similar to a traditional hashtable, but my cursory read through didn't suggest that that was the case here. @mflatt can hopefully confirm one way or the other.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -29,8 +29,9 @@ certain algorithms. Use @racket[immutable?] to check whether a hash
 table is immutable.
 
 @margin-note{Immutable hash tables actually provide @math{O(log N)}
-access and update. Since @math{N} is limited by the address space so
-that @math{log N} is limited to less than 30 or 62 (depending on the
+access and update (where the base of the @math{log} is 32.
+Since @math{N} is limited by the address space so
+that @math{log N} is limited to less than 7 or 13 (depending on the
 platform), @math{log N} can be treated reasonably as a constant.}
 
 For @racket[equal?]-based hashing, the built-in hash functions on


samth:
Note that RacketCS uses a different implementation with probably different constants, you can see it here: https://github.com/racket/racket/blob/master/racket/src/cs/rumble/intmap.ss

mflatt:
@elfprince13 Thanks for the suggestions! Partly because of the different implementation @samth mentions, I think it's probably not helpful to write the note in terms of base 32. But the numbers currently there are also a bit off, both because the `log` base should have been specified as 2 and because "30 or 62" is an overestimate for Racket CS (which has 8-byte alignment on 64-bit platforms). Maybe just note that `log_2 N` is limited by the address space?

mflatt:
Specifically, performance for a given hash table depends on the number of elements in the table, not the number of bits in the hash used for tables. (The number of bits in the hash matters, but it's always the same, so it's a constant factor.)

mflatt:
It's true that the hash-code and bits-per-level sizes also provide an upper bound on the tree depth. But the bound corresponding to the number of items N in a hash table is more interesting, because that bound applies to a specific tree: Assuming that the hash function works right, the performance of hash-table operations is O(log N) for a hash table of with N keys, even when N varies in a range that is much less than the maximum number of possible objects. The bound implied by the hash-code size is weaker in the sense that it's the same for all hash tables, no matter how many items the hash table contains. If we're already going to mention the O(log N), it doesn't seem necessary to go into internal details about hash codes.

To put it another way, the number of items in a hash table does bound the average trie depth.  With the current trie implementations, the hash-code-based depth could be at a maximum even for just two items in the hash table if their hash code collide for all but the last few bits. But as long as the hash function is working right, the trie will tend to branch out and have an average depth O(log N).

elfprince13:
I mostly just feel like as long as you're arguing that your implementation provides "effectively constant" access times, you might as well go for the smallest constant you can.

PR #2886, opened 2019-11-02
Fix #2885: convert number entity to character

### LINKED ISSUE #2885: read-xml discards data
Some XML documents (in my case these are XML files from [MediaWiki API](https://www.mediawiki.org/wiki/API:Main_page)) encode newline with `&#13;&#10;`. For instance:

```
<?xml version="1.0"?>
<blah foo="1 &amp;&#13;&#10;2" />
```

should contain an encoding of the string:

```
1 &
2
```

The problem is that `read-xml` completely discards `&#13;&#10;`, so it's impossible to recover the newline back. 

Here's a demonstrating Racket program:

```
#lang racket

(require xml)

(define x #<<EOF
<?xml version="1.0"?>
<blah foo="1 &amp;&#13;&#10;2" />
EOF
  )

(read-xml (open-input-string x))
```

This outputs:

```
(document (prolog (list (p-i (location 1 0 1) (location 1 21 22) 'xml "version=\"1.0\"")) #f '()) (element (location 2 0 23) (location 2 33 56) 'blah (list (attribute (location 2 6 29) (location 2 30 53) 'foo "1 &2")) '()) '())
```

Note that the `sxml` package retains the data and in fact even convert them to newline automatically:

```
#lang racket

(require sxml)

(define x #<<EOF
<?xml version="1.0"?>
<blah foo="1 &amp;&#13;&#10;2" />
EOF
  )

(ssax:xml->sxml (open-input-string x) '())
```

outputs:

```
'(*TOP* (*PI* xml "version=\"1.0\"") (blah (@ (foo "1 &\r\n2"))))
```

sorawee:
CC: @jeapostrophe

### DIFF
--- a/pkgs/racket-test/tests/xml/num-entity.rkt
+++ b/pkgs/racket-test/tests/xml/num-entity.rkt
@@ -1,0 +1,15 @@
+#lang racket/base
+(require xml)
+
+(module+ test
+  (require rackunit)
+
+  (define ex #<<EOF
+<?xml version="1.0"?>
+<blah foo="1 &amp;&#13;&#10;2" />
+EOF
+    )
+
+  (check-equal?
+   (xml->xexpr (document-element (read-xml (open-input-string ex))))
+   '(blah ((foo "1 &\r\n2")))))
--- a/racket/collects/xml/private/reader.rkt
+++ b/racket/collects/xml/private/reader.rkt
@@ -180,7 +180,7 @@
     [(gt) ">"]
     [(quot) "\""]
     [(apos) "'"]
-    [else (and (number? name) (string (integer->char name)))]))
+    [else #f]))
 
 (define-struct (EOF source) ())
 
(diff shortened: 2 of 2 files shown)


sorawee:
I see the error in this PR, but I want to disagree with these errors.

The grammar of xexpr is that each attribute must be a string. This dictates that I must convert `10` to `\n` in attributes to make it representable in the grammar.

But if I should/could do that, then what's the justification of keeping `10` in the body of xml? Since we need to convert `10` to `\n` at some places already anyway, why don't we do it on all of them? What's the point of keeping `10` at only some places and not the other?

jeapostrophe:
Thank you! Good change.

sorawee:
@jeapostrophe this PR is actually bad. There's a bug when I copied and pasted the code.`(string ...)`  should be changed to `(list ...)`. Otherwise, there will be an error:

```
append: contract violation
  expected: list?
  given: "("
```

But this also clearly means that the new test that I added didn't get run. Do you know what's going on?

jeapostrophe:
I fixed it.

sorawee:
Thanks! So the other question is, do you know how to make the test working properly?

jeapostrophe:
Are you talking about the CI build? I have no idea about that. DrDr will run it though.

sorawee:
Yep, I am talking about the CI build. I'll try to see if there's a way to make it work.

pmatos:
@sorawee sorry for the delay here. Just returned from a week holidays. Are you talking about the gitlab CI system? If you need some tests to run there, I can make it happen.

sorawee:
@pmatos Yes, I attempted to fix it at #2887, but it didn't work. Obviously, I don't understand the testing infrastructure well enough. If you could make it work, that would be really great.

In short: every file in the directory `tests/xml` that has the `test` submodule should be tested, but currently only `tests/xml/test.rkt` is run.

samth:
The problem is that you kept -l as the flag to raco test. You might want
-c, but probably better to just make that file require the others.

pmatos:
> The problem is that you kept -l as the flag to raco test. You might want -c, but probably better to just make that file require the others.

@sorawee let me know if you want to go ahead and do this or if you want me to. Lets not deadlock on this issue.

sorawee:
@pmatos please do it!

pmatos:
> @pmatos please do it!

Will do.

pmatos:
@sorawee done under https://github.com/racket/racket/commit/1eb64f3961c879d7a6dd161b7a4a37829340d7d4

sorawee:
Thanks! :)

PR #2887, opened 2019-11-02
test all xml tests

sorawee:


sorawee:
Hmm. I don't know how to make this work, and I spent too much time on it already. Closing... It should really be fixed though.

PR #2892, opened 2019-11-06
Docs: Show correct signatures for `bound-id-set-map` and `bound-id-set-for-each`

### LINKED ISSUE #2881: Documented signatures for `bound-id-set-map` and `bound-id-set-for-each` incorrect
The documented signatures for `bound-id-set-map` and `bound-id-set-for-each` seem wrong:

```
@defproc[(bound-id-set-map [s bound-id-set?]) list?]
@defproc[(bound-id-set-for-each [s bound-id-set?]) void?]
```

The documentation suggests they're comparable to `free-id-set-map` and `free-id-set-for-each`, which each take a function to call on each element, but that argument is missing here.

Calling `procedure-arity` confirms that these procedures have arities of 2, so I think this is simply a documentation issue.

rocketnia:
This fixes #2881.


gus-massa:
I forgot about this. Merged in https://github.com/racket/racket/commit/73dd32cd51a3b008a5ed0b471f859830287a4a2f . Thanks!

PR #2893, opened 2019-11-06
add a built-in function to determine if a given function is a struct property predicate

capfredf:
https://github.com/racket/typed-racket/pull/863 depends on `struct-type-property-predicate-procedure?`. For example, when generating contracts for `prop:custom-write` of type `(StructProperty (-> Self Output-Port (U 0 1 Boolean) Void) pred?)`, we need to know if pred? is a predicate for `prop:custom-write`. Otherwise the system would be unsound.

The pull request is a draft. I have added the function to the chez side and to the c side.


capfredf:
@mflatt should I revert any changes that removed trailing spaces in my pull request?

racket/src/racket/src/struct.c
```diff
@@ -3370,7 +3370,21 @@ struct_prop_getter_p(int argc, Scheme_Object *argv[])
...
+        && SAME_TYPE(SCHEME_TYPE(SCHEME_PRIM_CLOSURE_ELS(v)[0]), scheme_struct_property_type)))
+      return scheme_false;
+
+  if (argc == 1) {
+      return scheme_true;
+
+  } else {
+      prop = SCHEME_PRIM_CLOSURE_ELS(v)[0];
+      prop_t = argv[1];
+      return (scheme_equal(prop, prop_t) ? scheme_true : scheme_false);
```
mflatt: You can use `SAME_OBJ` (pointer equality) instead of `scheme_equal` here.

racket/src/cs/rumble/struct.ss
```diff
@@ -18,94 +18,113 @@
...
 (define (struct-type-property-accessor-procedure? v)
   (and (procedure? v)
        (let ([v (strip-impersonator v)])
          (with-global-lock* (hashtable-ref property-accessors v #f)))
        #t))
 
+(define/who struct-type-property-predicate-procedure?
+  (case-lambda
+   [(v) (struct-type-property-predicate-procedure? v #f)]
+   [(v spt) (and (procedure? v)
```
mflatt: Do you mean to allow the second argument to be `#f` as part of the public interface of `struct-type-property-predicate-procedure?`? If not, you could move the implementation to a `do-struct-type-property-predicate-procedure?` helper.

racket/src/cs/rumble/struct.ss
```diff
@@ -18,94 +18,113 @@
...
+                (let* ([v (strip-impersonator v)]
+                       [spt-c (with-global-lock* (hashtable-ref property-predicates v #f))])
+                  (cond
+                   [(not spt-c) #f]
+                   [(not spt) #t]
+                   [else (impersonator-of? spt spt-c)])))]))
```
mflatt: I'd expect `eq?` to work here. Do structure properties support impersonators in some way (that I've forgotten) to makes `impersonator-of?` needed?

racket/src/racket/src/struct.c
```diff
@@ -5999,8 +6024,9 @@ static Scheme_Object *do_chaperone_struct(const char *name, int is_impersonator,
       ("(or/c " extra "struct-accessor-procedure?\n"      \
        "      struct-mutator-procedure?\n"                 \
        "      struct-type-property-accessor-procedure?\n"  \
+       "      struct-type-property-predicate-procedure?\n"  \
```
mflatt: Is this intended? I don't think property predicates are allowed as arguments to `chaperone-struct` or `impersonate-struct`, but I could be lost here.

racket/src/cs/rumble/struct.ss
capfredf: I used eq? at first, but ran into a situation where typed-racket creates add contracts to struct properties. I thought those contracted values were impersonators of the original. Turns out they aren't... 

I will revert back to eq?

mflatt:
@capfredf - should I look at this more now, or are you still working on this?

capfredf:
@mflatt Oh, you can take another look at this, thank you

mflatt:
I'll make some adjustments and merge:

 * change `struct-type-property-predicate-procedure?` to require #f or a structure type property as its second argument (instead of allowing just anything), except

 * change `struct-type-property-predicate-procedure?` to treat as #f second argument like having only one argument; it's not always clear whether that's the right thing but experience suggests that it's usually the right choice to have some value mean the same as not supplying the argument

 * add docs

mflatt:
Merged as 4405ed669f900439703c855faa947f16d44a87cd

Issue #2890, opened 2019-11-05
Version detection logic broken in 7.4


winny-:
In `cfg-racket` there are two related bugs that in conjunction break version detection (when `--enable-shared` is used):

1. When [GNU Make is not being used][gmake], the [`grep .. | cut ...` pipeline][pipeline] will output three dots separated by newlines. This pipeline appears to be a carryover from 7.3.

```
$ pwd
/var/tmp/racket-7.4/src
$ grep " MZSCHEME_VERSION " racket/src/schvers.h | cut -d '"' -f 2
.
.
.

```

2. [GNU Make detection is guarded by GCC detection][gmake]. The GCC detection will  not detect a GCC installed with a [target triplet][tt] in its program name. For example on Gentoo my system compiler is `x86_64-pc-linux-gnu-gcc`.

I used the following patch to move the GNU make detection logic outside of the GCC detection guard. It does not address the problems of GCC target triplets or the non-GNU Make fallback version detection.

```diff
diff --git a/cfg-racket b/cfg-racket
index 61487cc..d1a4795 100755
--- a/cfg-racket
+++ b/cfg-racket
@@ -4601,23 +4601,28 @@ if test "$ARFLAGS" = '' ; then
 fi
 SUB_CONFIGURE_EXTRAS="${SUB_CONFIGURE_EXTRAS} AR_FLAGS="'"'"${ARFLAGS}"'"'
 
+is_gmake=`make -v no-such-target-we-hope 2>&1 | grep "GNU Make"`
+
 # If using gcc, we want all warnings:
 if test "$CC" = "gcc" ; then
   COMPFLAGS="$COMPFLAGS -Wall"
 
   # Use -MMD when we have gcc and gnumake:
-  is_gmake=`make -v no-such-target-we-hope 2>&1 | grep "GNU Make"`
   if test "$is_gmake" = "" ; then
     WX_MMD_FLAG=""
   else
     WX_MMD_FLAG="-MMD"
-    INCLUDEDEP="-include"
   fi
 
   # Compile mzdyn.o with -fPIC
   DYN_CFLAGS="-fPIC"
 fi
 
+# Use GNU Make feature if present
+if test "$is_gmake" != ""; then
+    INCLUDEDEP="-include"
+fi
+
 if test "$REZ" = '' ; then
  REZ=/Developer/Tools/Rez
 fi
```

I hope this helps :) I could open a PR, but these decisions likely need a lot of testing, and somebody more familiar with Racket's configure scripts.

[...]

PR #2897, opened 2019-11-08
Improve gcc detection in configure script

pmatos:
We have been detecting gcc by the CC variable but this fails under
ubuntu for example where you might specify CC='gcc-8' to ./configure.

Related to #2890


pmatos:
Oh, I forgot about that one. Really should find a way to either get mac for testing or emulate it.

pmatos:
Will update PR.

PR #2900, opened 2019-11-09
net-lib/net/sendurl.rkt: simplification and cross-platform feature parity

rrthomas:


pkgs/net-lib/net/sendurl.rkt
```diff
@@ -93,24 +78,23 @@
...
     (error 'send-url "expected a string, got ~e" url-str))
   (let ([url-str (if escape? (escape-url url-str) url-str)])
     (if (procedure? (external-browser))
       ((external-browser) url-str)
-      (case stype
-        [(macosx)  (send-url/mac url-str)]
-        [(windows) (send-url/win url-str)]
-        [(unix)    (send-url/unix url-str (force separate-window?))]
+      (case (system-type)
+        ['macosx  (send-url/mac url-str)]
```
samth: This happens to work because of how the reader abbreviation for quote works, but these should be parenthesized, not quoted.

rrthomas: Thanks, updated.

rrthomas:
I've tested this on GNU/Linux and macOS (including `#:browser` for `send-url/mac`). So far I've been unable to build Racket on Windows, so haven't been able to test it there; I've got my fingers crossed since I didn't actually change `send-url/win`.

rmculpepper:
I believe the reason `xdg-open` wasn't already on the list is because it handles file URLs with queries incorrectly. I'm not sure about `open` on Mac OS.

rrthomas:
@rmculpepper I just tested `xdg-open` and `open` with queries and they work.

rmculpepper:
@rrthomas It didn't work for me. Specifically, when I build with your patch, delete `~/.racket/racket-prefs.rktd` (to unset the browser preference), and then run `raco docs lambda`, it fails with the following error:

browser-run: process execute failed: '(#<path:/usr/bin/xdg-open> "file:///home/ryan/projects/racket/racket/doc/search/index.html?q=lambda")
gio: file:///home/ryan/projects/racket/racket/doc/search/index.html?q=lambda: Error when getting information for file “/home/ryan/projects/racket/racket/doc/search/index.html?q=lambda”: No such file or directory

This is on Ubuntu 18.04.3, with xdg-open 1.1.2.

rrthomas:
@rmculpepper I'm sorry, I misread your previous comment, and only tested http URLs with parameters. I'll work out a fix.

rrthomas:
Update: I checked again on macOS, and found that using `open` doesn't work there either, in the same way. Going back to the previous code, only the formulation where you specify the browser works, i.e. `osascript -e "tell application X to open location Y"`. Rather than find the default browser for this case (which is tricky and error-prone), it seems sensible to fall back to the trampoline method in this case too (worst case is a some tiny files littering a temp directory), so I'll update the PR to do that.

rrthomas:
I have also tried to understand the preferences controlling whether a new browser window is desired by default. I cannot find any mention of either `'new-browser-for-urls` or `'default-browser` in either racket or drracket except inn the definition of `separate-by-default?`, which suggests that, absent a manually-entered preference (and undocumented since the start of git history), this will never be true.

rmculpepper:
The new changes look good to me. Thanks!

It might be better to use the trampoline unconditionally for `open` and `xdg-open`. Previously, calling send-url on non-html files (eg, "notes.txt", "report.pdf", "song.mp3") would open those files in a browser, but `{xdg-,}open` would open them in the preferred application for those file extensions. On the one hand, that might be considered a feature, but on the other, the docs strongly imply that the url is handled by a browser. (I'm not sure if anyone relies on the difference, but I'd err towards keeping the same behavior.)

rrthomas:
@rmculpepper Thanks for the further review. Great minds think alike: I've already added a change to use the trampoline unconditionally.

rmculpepper:
It looks like the trampoline is only used if the url has a `#` or `?` in it.

rrthomas:
Sorry, @rmculpepper, I didn't read your last message carefully enough.

I agree with your reading of the docs: they do imply that the URL will be opened in a browser. However, the precise wording is: "Opens @racket[str], which represents a URL, in a platform-specific
manner."

The current behavior on both Windows and macOS is to open files in the preferred application. That's the behavior I've replicated with xdg-open: this is by design, and for example it allows Scribble's DrRacket button code to be greatly simplified. The functionality of opening a file in its preferred application is useful and not otherwise available.

I see two options here:

1. Make `send-url` open the URL strictly in a browser, being aware that this is not how the system commands on any of our platforms work (so in the long run this may be more confusing for users). Add a new function that launches a file in its preferred application.
2. Keep the PR the way it is.

rrthomas:
I have now also fully updated the documentation; it pretty much consists of removing documentation of deprecated exports.

mflatt:
I haven't looked at this PR (thanks for working on it!), but please don't remove documentation. As a rule, we rarely "deprecate" in the sense of removing bindings, although documentation may change to discourage the use of a binding.

rrthomas:
@mflatt: thanks, I'll restore the documentation. Is it reasonable to discourage the use of the preference? The general idea of the PR is that the situation for Unix should match that on Windows and macOS, where there's no preference.

Relatedly, I was planning to remove the preference itself from DrRacket. Again, that wouldn't remove the bindings. Is that OK?

mflatt:
I'm ok with removing the preference, but I don't feel well informed about the state of things. From what I know, it could make sense to discourage the use of the preference in the documentation.

rfindler:
I also don't feel informed about the state of things, but I don't think  that it doesn't make sense to get rid of the preference for all possible linux installations as there are (at least I think there are?) people who aren't using fancy OS's on linux machines and appreciate being able to set this preference.

It does make sense to me to drop the preference when we are in well-known good environments, tho (or just to not consult it).

rrthomas:
Thanks both @mflatt  and @rfindler for your feedback. I'll get this PR finished first and address your feedback on my PR for drracket!

rrthomas:
I have now tested this PR on Ubuntu, macOS and Windows, launching HTML files and PDF files from Scribble, and testing fragment/query support by running `raco docs lambda`. All of these work.

I've restored the documentation previously removed.

Is there anything else I need to do to land this PR? Many thanks to everyone who's helped with it!

rrthomas:
@rmculpepper  @robby @mflatt ping

mflatt:
This does not work for me on Windows, I think because it's looking for "start" as an executable.

While we could change things to run "start" as a shell command, using the Windows shell is in the "you had one problem, and now you have five problems" category. Probably `shell-execute` is a better way to go on Windows, given the trampoline that fixes up problems with fragments and queries.

rrthomas:
@mflatt, I'm baffled. An earlier version did what you describe, but the current version uses `shell-execute`. I've checked the commit hashes on the PR, and they seem fine.

mflatt:
@rrthomas I must have merged incorrectly in my checkout, because I see `shell-execute` in your branch. I'll try again.

mflatt:
worked this time 🤷‍♂

LiberalArtist:
I was curious about the comment in <https://github.com/racket/racket/commit/e4d43fa> that "XonX is defunct." It does seem like XQuartz development hasn't been especially active recently, but I still use it semi-regularly for some (non-Racket) applications, and Racket still appears to support building with `--enable-xonx`.

rrthomas:
@LiberalArtist I don't use macOS much, so I looked into XonX, and found that its Wikipedia page mentions it in the past tense. I can't find mentions of releases post-2004. After a bit more investigation, am I correct in concluding that the current name for this is "XQuartz"?

I reviewed the code change to `sendurl.rkt`, and I still think it's correct: on macOS, the user-configured browser should be used. If it is desired to pretend that racket is really of `(system-type)` `'unix`, that should be done at a deeper level, not just in `sendurl.rkt`.

LiberalArtist:
Some context: macOS is a Unix ([certified UNIX 03 compliant](https://www.opengroup.org/openbrand/register/brand3653.htm)), but a fairly strange one, and obviously it has a proprietary desktop API, Cocoa, which is unrelated to the XDG/[freedesktop.org](https://freedesktop.org) interfaces used by other modern Unix-like desktops. In particular, it uses the proprietary Quartz display server rather than X11. Over the history of Mac OS X, both Apple and independent developers have implemented X11 servers to facilitate running software written for normal Unix-like systems: XQuartz is the current one (though it looks like there is or was some effort to develop a successor). IIUC, XonX was the first one, but it also makes sense as a generic abbreviation for "**X**11 **on** Mac OS **X**."

I don't think I've ever compiled Racket with `--enable-xonx`, but it is documented as configuring Racket to use GTK and X11 rather than Cocoa, and it seems to also make other things more Unix-ey, e.g.: https://github.com/racket/racket/blob/b1b8beece58c0345c4d9daf95ebcb53ed9a63d40/racket/src/rktio/rktio_fs.c#L1906-L1913

In particular, `--enable-xonx` causes `(system-type)` to return `'unix`. Before <https://github.com/racket/racket/commit/e4d43fa>, `net/sendurl` was using `(system-library-subpath)` to detect when Racket was actually running on a Mac. I would guess the rationale is that a Mac is unlikely to have the browser executables used in the `'unix` case available: even if someone uses e.g. Firefox, which does come with the `firefox` command on Mac, the way macOS installs software and sets up the `PATH`, especially for graphical applications, would usually defeat `find-executable-path`.

I don't have a strong opinion about what behavior would be correct. The old behavior does seem more likely to actually succeed, but, on the other hand, arguably `--enable-xonx` could mean you don't want Mac-specific behavior. I'd love to hear why people are actually using `--enable-xonx`, if indeed anyone is.

I do agree that parsing `(system-library-subpath)` in `net/sendurl` feels like the wrong level of abstraction, though [it is used](https://docs.racket-lang.org/raco/matching-platform.html) for platform specifications in the package system and for FFI purposes. It also seems like `(system-type)` doesn't actually say much about what its result means.

PR #2901, opened 2019-11-10
Revert "Fix gen:dict methods for alists with duplicate keys (#2846)"

samth:
This breaks existing packages; see https://github.com/greghendershott/aws/issues/64
for an example.

Unfortunately, we probably have to live with this limitation of the
interface because of existing code.

This reverts commit 966399dec6782ab22641162abf6a528597e352a7.


pmatos:
Can we add with this revert a test to check that we don't in the future break this again?

PR #2902, opened 2019-11-10
Separate GCC and GNU Make detection

winny-:
Related to #2890.

This ensures when GNU Make is `make`, the `version.mak` file is used
regardless of `CC`.


winny-:
> The job exceeded the maximum time limit for jobs, and has been terminated.

pmatos:
It looks good to me, thanks. It will probably conflict with #2897 but it won't be a pain to merge.

pmatos:
Thanks @winny- for the PR and @mflatt for the review. I am merging.

Issue #1627, opened 2017-03-05
syntax->string documentation and implementation don't line up


endobson:
The implementation has the contract: `(and/c syntax? stx-list?)`
https://github.com/racket/racket/blob/master/racket/collects/syntax/to-string.rkt#L82
The documentation has the contract `stx-list?`
https://github.com/racket/racket/blob/master/pkgs/racket-doc/syntax/scribblings/to-string.scrbl#L10

Paul-C-Anagnostopoulos:
This inconsistency is still present. I'm happy to fix it if someone tells me the correct contract. From looking at `syntax->string`, I believe that it has to be the first contract above.

bennn:
Yes I think the first contract is the right one

Paul-C-Anagnostopoulos:
Shall I correct the documentation?

bennn:
No that's ok

Paul-C-Anagnostopoulos:
The documentation shouldn't be consistent with the function?

If it doesn't matter, then this issue can be closed, I think.

bennn:
I mean: no that's ok because when I saw your last comment I went ahead & opened #2351

Paul-C-Anagnostopoulos:
Aha, got it. I didn't see that until after I made my comment above.

Issue #1629, opened 2017-03-05
syntax->string doesn't work on syntax pairs


endobson:
```
(syntax->string #'((a . b)))
; map: contract violation
;   expected: list?
;   given: #f
;   argument position: 2nd
; [,bt for context]
```

Bug is here where it assumes that if the syntax object is a syntax pair then it is a syntax list.
https://github.com/racket/racket/blob/master/racket/collects/syntax/to-string.rkt#L61

lkh01:
Hey, I came across this issue [here](https://github.com/racket/racket/wiki/Easy-bugs-to-fix) and I want to fix it and https://github.com/racket/racket/issues/1628.

It looks like it's still broken in the current build (v7.5.0.7), so I'll take a shot at it.

samth:
Fixed by #2911

Issue #1628, opened 2017-03-05
syntax->string removes underscores from its output


endobson:
I'm trying to use syntax->string but it removes _ from its output. This behavior is not documented at all.

https://docs.racket-lang.org/syntax/syntax-helpers.html?q=syntax-%3Estring#%28def._%28%28lib._syntax%2Fto-string..rkt%29._syntax-~3estring%29%29
```
Builds a string with newlines and indenting according to the source locations in stx-list; the outer pair of parens are not rendered from stx-list.
```

https://github.com/racket/racket/blob/master/racket/collects/syntax/to-string.rkt#L30

PR #2911, opened 2019-11-15
fix syntax->string

lkh01:
This pull request solves #1629.

```
> (syntax->string #'((a . b)))
"(a . b)"
```
It also fixes an undocumented issue that happens when called on a list with dot notation. This was the output before:
```
> (syntax->string #'((a . (b c))))
"(a    b c)"
``` 
And it now works like this:
```
> (syntax->string #'((a . (b c))))
"(a . (b c))"
```
There are a couple more issues, I think. In order to preserve the formatting in the resulting string, sometimes it is necessary to find the location of the `.` in a pair, but I haven't found a way to do that yet. 
For example:
```
> (syntax->string #'((a        . b))
"(a . b)"
> (syntax->string #'((a   .      b))
"(a .      b)"
```
With that in mind, this is still a work in progress.

### DIFF
--- a/racket/collects/syntax/to-string.rkt
+++ b/racket/collects/syntax/to-string.rkt
@@ -27,7 +27,10 @@
             (cond
               [(eq? 'code:blank (syntax-e c))
                (advance c init-line!)]
-              [(eq? '_ (syntax-e c)) (void)]
+              [(eq? '_ (syntax-e c)) 
+               (advance c init-line!)
+               (printf "_")
+               (set! col (+ col 1))]
               [(eq? '... (syntax-e c))
                (void)]
               [(and (pair? (syntax-e c))
@@ -58,7 +61,19 @@
                (define c-paren-shape (syntax-property c 'paren-shape))
                (printf "~a" (or c-paren-shape #\())
                (set! col (+ col 1))
-               (map (loop init-line!) (syntax->list c))
+               (define se (syntax-e c))
+               (define (build-string-from-pair sp)
+                 (cond
+                   [(syntax? sp)
+                    (printf " . ")
+                    (set! col (+ col 3))
+                    ((loop init-line!) sp)]
+                   [else
+                    ((loop init-line!) (car sp))
+                    (build-string-from-pair (cdr sp))]))
+               (if (list? se)
+                   (map (loop init-line!) se)
+                   (build-string-from-pair se))
                (printf (case c-paren-shape
                          [(#\[) "]"]
                          [(#\{) "}"]
--- a/pkgs/racket-test/tests/syntax/to-string.rkt
+++ b/pkgs/racket-test/tests/syntax/to-string.rkt
@@ -1,0 +1,9 @@
+#lang racket
+(require syntax/to-string
+         rackunit)
+
+(check-equal? (syntax->string #'((a . 
+                                      b))) "(a . \n     b)")
+(check-equal? (syntax->string #'((a b c d))) "(a b c d)")
+(check-equal? (syntax->string #'(a 'b #(a b c) c)) "a 'b #(a b c) c")
+(check-equal? (syntax->string #'((a b  _   d))) "(a b  _   d)")


bennn:
I don't think it's possible to find the position of the `.` in `#'((a        . b)`

racket/collects/syntax/to-string.rkt
```diff
@@ -58,7 +58,23 @@
                (define c-paren-shape (syntax-property c 'paren-shape))
                (printf "~a" (or c-paren-shape #\())
                (set! col (+ col 1))
-               (map (loop init-line!) (syntax->list c))
+               (letrec ([sl (syntax->list c)]
+                     [se (syntax-e c)]
+                     [build-string (lambda (s)
```
bennn: style: if this code was an inner-loop in the else-branch below --- instead of a `letrec` --- then I think I would've had an easier time understanding what input this function expects

racket/collects/syntax/to-string.rkt
```diff
@@ -58,7 +58,23 @@
...
+                                       (begin
+                                         (printf " . ")
+                                         (set! col (+ col 3))
+                                         ((loop init-line!) s))
+                                       (begin
+                                         ((loop init-line!) (car s))
+                                         (build-string (cdr s)))))])
+                 (cond
+                   [(and sl 
+                         (list? se))
```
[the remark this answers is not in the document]
lkh01: Oh, you're right, `syntax->list` and `syntax-e` seem to be `equal?` for lists. Thanks.

lkh01:
I'm not quite sure either, but `pkgs/racket-test-core/tests/racket/stx.rktl` seems a bit more appropriate than `pkgs/racket-test-core/tests/racket/syntax.rktl` judging by their contents. There's also the `pkgs/racket-test/tests/syntax/` folder. That's where `pkgs/racket-test/tests/syntax/free-vars.rkt` is, and `free-vars` is just below `syntax->string` in the docs.

racket/collects/syntax/to-string.rkt
lkh01: I haven't found an elegant solution to this.
I've changed the name to `build-string-from-pair`, hoping that the intent is clearer now.

bennn:
Ah, that folder sounds best. I'd make a new file `pkgs/racket-test/tests/syntax/to-string.rkt`

racket/collects/syntax/to-string.rkt
```diff
@@ -58,7 +58,19 @@
                (define c-paren-shape (syntax-property c 'paren-shape))
                (printf "~a" (or c-paren-shape #\())
                (set! col (+ col 1))
-               (map (loop init-line!) (syntax->list c))
+               (letrec ([se (syntax-e c)]
+                        [build-string-from-pair (lambda (sp)
+                                                  (if (syntax? sp)
```
capfredf: style: use `cond` in place of `if + begin`?

racket/collects/syntax/to-string.rkt
```diff
@@ -58,7 +58,19 @@
                (define c-paren-shape (syntax-property c 'paren-shape))
                (printf "~a" (or c-paren-shape #\())
                (set! col (+ col 1))
-               (map (loop init-line!) (syntax->list c))
+               (letrec ([se (syntax-e c)]
+                        [build-string-from-pair (lambda (sp)
```
sorawee: I personally find that the rightward drift here is really too much. Can't we use internal definitions instead of `letrec`? Something like:

```
(define se (syntax-e c))
(define (build-string-from-pair sp)
  (cond
    [...]
    ...))
(if (list? se)
    ...
    ...)
```

racket/collects/syntax/to-string.rkt
```diff
@@ -58,7 +58,19 @@
...
+                                                  (cond
+                                                    [(syntax? sp)
+                                                      (printf " . ")
```
sorawee: Nit: wrong indentation. DrRacket's auto indentation would align the left paren with the left paren on the above line.

racket/collects/syntax/to-string.rkt
```diff
@@ -58,7 +58,19 @@
...
+                                                      (set! col (+ col 3))
+                                                      ((loop init-line!) sp)]
+                                                    [else
+                                                      ((loop init-line!) (car sp))
```
sorawee: Ditto re: indentation.

racket/collects/syntax/to-string.rkt
```diff
@@ -58,7 +58,19 @@
...
+                                                      (build-string-from-pair (cdr sp))]))])
+                 (if (list? se)
+                    (map (loop init-line!) se)
```
sorawee: Ditto re: indentation.

racket/collects/syntax/to-string.rkt
lkh01: Thanks for the review. You're right, that was a case of fullscreen tunnel vision.

lkh01:
Not sure how high this is on the priority list, and I don't want to ping anyone, but I think it's ready to merge.

samth:
Before I merge this, does this also fix #1627 and #1628?

lkh01:
I believe that #1627 was fixed by #2351, but as far as #1628 goes, I was going to leave it for later, but it seems like an easy fix, so I'll do it now.

lkh01:
The tests that failed seem like an unrelated bug. 
But looking through the logs I realized that the tests I added in `pkgs/racket-test/tests/syntax/to-string.rkt` aren't being run. There are a couple of such folders in `pkgs/racket-test/tests` that are ignored, `pkgs/racket-test/tests/syntax` being one of them, while `pkgs/racket-test/tests/stxparse`, for example, is tested.

I'm not really sure how that part of the test suite works. There's this function in `pkgs/racket-test/tests/run-automated-tests.rkt` https://github.com/racket/racket/blob/b1b8beece58c0345c4d9daf95ebcb53ed9a63d40/pkgs/racket-test/tests/run-automated-tests.rkt#L29
but some of the tests that _are_ being run like `pkgs/racket-test/tests/stxparse/*` aren't in that definition.

samth:
I think this looks good, and leaving that test there is fine. Right now it will be run by drdr.racket-lang.org; maybe in the future we want to include it in some other places that run tests.

Can you push a single squashed commit?

lkh01:
@samth 
Oh, looks like I forgot to notify you. Single squashed commit is there.

PR #2915, opened 2019-11-18
In-place build: cope with default setting of prefix

rrthomas:
configure scripts look for and read a local configuration file given by the 
environment variable CONFIG_SITE. This can set variables such as prefix.

Racket’s build system was assuming that prefix would be set to NONE unless
a
--prefix command line argument was given. But it could be set by a 
CONFIG_SITE configuration file instead.

Hence, for in-place build add an explicit --disable-useprefix option, to 
cause any prefix setting to be ignored, and use that in the top-level 
Makefile.

Regenerate the configure scripts to get the updated code.


mflatt:
Thanks!

PR #2916, opened 2019-11-20
Initial commit to new CI system based on GitHub Actions

pmatos:
Jobs to build CGC, 3m and CS separately on Linux.
macOS is failing during CGC install - need to investigate.

It also runs tests/racket/test but there are some flaky places and
wills tests that require some fixing before merging.

### DIFF
--- a/.github/workflows/ci-push.yml
+++ b/.github/workflows/ci-push.yml
@@ -1,0 +1,186 @@
+name: CI
+
+on: [push]
+
+jobs:
+  
+# Build jobs
+# These jobs build each Racket component separately and tests on the component start as soon as each
+# component finishes building.
+
+  build-racketcgc:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-18.04, macos-latest]
+        cify: [cify, nocify]
+          
+    runs-on: ${{ matrix.os }}
+    env:
+      RACKET_EXTRA_CONFIGURE_ARGS: ""
+    
+    steps:
+    - uses: actions/checkout@v1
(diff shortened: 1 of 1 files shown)


samth:
Do we really want --enable-xonx? That seems unlikely to be the most important configuration.

pmatos:
> Do we really want --enable-xonx? That seems unlikely to be the most important configuration.

Honestly, I don't know how to build on mac. If I build without `--enable-xonx` and all the other flags, I get a message saying I need `--enable-xonx` or `--enable-macprefix`. I just chose `--enable-xonx` and it seemed to work. What's your suggestion for build parameters on mac?

pmatos:
It seems that `--enable-macprefix` works, so using that instead for now. In any case, there are a couple of flaky tests (sometimes they pass, other times they don't) in `racket/test` that I would really like to ensure they work consistently before getting them merged.

1. Section places: https://github.com/LinkiTools/racket/runs/311655656#step:8:30
2. Section wills: https://github.com/LinkiTools/racket/runs/311655656#step:8:44
3. Section synchronization https://github.com/LinkiTools/racket/runs/311655704#step:8:33

pmatos:
Since the builds are working - would it make sense to push the builds for now to have something in there and add the tests slowly as we make sure they are not flaky? What do others think?

samth:
I'm in favor of that.

pmatos:
@samth thanks - I will split and squash the commits into two. The build part will be merged and the test part with flaky tests will be opened as a new PR.

pmatos:
Merging the build part - opening new PR for testing part.

PR #2917, opened 2019-11-20
Add workflow for PR continuous integration

pmatos:
Here's a workflow for PRs - my point of view here is that this should be quick while covering most of the features we regularly use. Once commit happens, we run a normal workflow taking more variations into consideration.

If the normal workflow fails but passed this workflow, we can always revert the commit. Having a PR workflow taking a multiple of hours makes no sense. I think we should help contributors here by allowing them to use our CI by creating draft PRs (which automatically run our CI) and get a development turnaround that's as quick as possible.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -1,4 +1,4 @@
-name: CI: Pull Request
+name: CI Pull Request
 
 on: [pull_request]
 
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -32,7 +32,7 @@ jobs:
         raco test -l tests/match/main
         raco test -l tests/zo-path
         raco test -c tests/xml
-        raco test -c tests/db/all-tests
+        raco test -l tests/db/all-tests
         raco test -c tests/stxparse
 
   buildtest-win:
@@ -52,7 +52,7 @@ jobs:
         call msvcprep.bat x86_amd64
         call build.bat
         call ..\..\racket.exe csbuild.rkt
-        ..\..\raco.exe pkg install racket-lib
+        ..\..\raco.exe pkg install --auto racket-lib
     - name: Test
       shell: cmd
       run: |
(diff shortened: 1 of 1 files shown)


samth:
Could we make this complete faster by installing Racket from the ppa/the latest release and using that to build Racket3m instead of first building RacketCGC?

pmatos:
It's slightly annoying that this workflow is not running on itself. No idea if it's all green so don't want to merge it. I will attempt to run this in my branch by changing it to be a pull workflow instead.

pmatos:
> Could we make this complete faster by installing Racket from the ppa/the latest release and using that to build Racket3m instead of first building RacketCGC?

Good idea. I have to drop this for a couple of hours but I will try to see how much faster we can get by grabbing racket from ppa.

samth:
I think we should run RacketCS tests as well on PRs, at least on one platform.

pmatos:
> Could we make this complete faster by installing Racket from the ppa/the latest release and using that to build Racket3m instead of first building RacketCGC?

Oh - before I go... isn't it true that RacketCS sometimes need a pretty recent racket to compile and grabbing the latest stable from ppa might not work? I remember this used to be the case - maybe not anymore.

samth:
Yes, RacketCS needs the latest Racket, so we would want to build it after building Racket3m (in parallel with testing Racket3m maybe for fastest turnaround).

mflatt:
Racket3m may need the latest development Racket instead of the latest release, too.

pmatos:
> Racket3m may need the latest development Racket instead of the latest release, too.

In that case, it doesn't make sense to use a pre-built racket binary if I am thinking this through. I will just get the Windows build fixed for now.

samth:
@mflatt, when was the last time that happened?

mflatt:
@samth The last time that "collects/compiler/private/xform.rkt" changed, probably, which was September 30.

samth:
I guess we could distinguish two situations -- when `xform` changes and we need to use the up-to-date version (happens pretty often), and when there's a bug in `racketcgc` that causes xform not to work and thus we need the up-to-date `racketcgc`. In the latter case, we really need the latest Racket from source, in the former case we could just use the old Racket to run the new xform.

However, it's probably not worth it to figure out how to use new-xform with existing-racket just for this use case.

pmatos:
@samth This workflow has passed: https://github.com/LinkiTools/racket/runs/314844108
Shall we get this upstream and think about optimizations later? I am worried about flaky tests but if we find them, we remove them from the workflow and open a bug to fix them.

pmatos:
I can see that the `CI Pull Request` workflow from this PR, is being triggered on itself: Good.

PR #2918, opened 2019-11-20
Add job to test racket CGC on Linux and macOS

pmatos:


pmatos:
Lets get this working - in case we notice some tests are flaky, we should disable them until fixed.

.github/workflows/ci-push.yml
```diff
@@ -184,3 +184,36 @@ jobs:
         name: racketcs-${{ matrix.os }}-x64_git${{ github.sha }}
         path: ../racketcs-${{ matrix.os }}-x64_git${{ github.sha }}.tar.bz2
 
+# Tests
+  test-cgc:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [ubuntu-18.04, macos-latest]
```
samth: Why not windows?

pmatos: The build was not working when I wrote this. Will add it next.

PR #2922, opened 2019-11-21
change `datum` to work with syntax variables too

rmculpepper:
This PR changes `datum` from `syntax/datum` to cooperate with syntax variables and syntax/parse attributes in addition to datum variables bound with datum-case. This has two nice consequences:

1. If x is a depth-1 syntax variable, then `(datum (x ...))` is equivalent to `(syntax->list #'(x ...))`. Moreover, if y is also a depth-1 syntax variable, then `(datum ((x y) ...))` is equivalent to `(map syntax->list (syntax->list #'((x y) ...)))`, and so on. That is, `datum` retains existing syntax objects in syntax variable bindings, but it never creates additional syntax.

2. If a is a depth-1 attribute with non-syntax values---say, an `ast` attribute that contains structures parsed from the syntax---then the list of values can be obtained with `(datum (a ...))`, which makes the list-ness apparent, unlike `(attribute a)`.

See the new test file for more examples.

### DIFF
--- a/racket/collects/racket/private/template.rkt
+++ b/racket/collects/racket/private/template.rkt
@@ -364,7 +364,7 @@
     (define (lookup id depth0)
       (define (make-pvar var check pvar-depth)
         (define (make-ref var)
-          (cond [check `(t-check-var (,check ,var 0 #t (quote-syntax ,id)))]
+          (cond [check `(t-check-var (,check ,var 0 ,stx? (quote-syntax ,id)))]
                 [else `(t-var ,var)]))
         (define (make-src-ref var id)
           (cond [check `(#%expression (,check ,var 1 #f (quote-syntax ,id)))]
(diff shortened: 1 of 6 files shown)


rmculpepper:
I have added docs and made one more change: a datum variable used within a syntax template now raises an error, rather than being treated as a constant. That reduces the asymmetry between syntax and datum variables and templates.

racket/collects/racket/private/template.rkt
```diff
@@ -386,18 +386,24 @@
...
                (define check (and attr (attribute-mapping-check attr)))
                (make-pvar var check pvar-depth)]
-              [(and (not stx?) (s-exp-pattern-variable? v))
-               (define pvar-depth (s-exp-mapping-depth v))
-               (define var (s-exp-mapping-valvar v))
-               (make-pvar var #f pvar-depth)]
+              [(s-exp-pattern-variable? v)
+               (cond [stx?
+                      ;; datum variable in syntax template is error
+                      (wrong-syntax id "datum variable not allowed in syntax template")]
```
bennn: Should the docs talk about / show this error? (I didn't see any discussion)

Issue #2906, opened 2019-11-12
Can't build on msys64


rrthomas:
I'm trying to build racket on mingw64, and get this error that I don't understand:

    make[8]: Entering directory '/home/rrt/repo/racket/racket/src/build/racket/gc2'
    env XFORM_PRECOMP=yes ../racketcgc -G /home/rrt/repo/racket/build/config -O "info@compiler/cm error" -l- setup --boot /home/rrt/repo/racket/racket/src/build/../racket/gc2/../../setup-go.rkt ../../compiled --tag ++out /home/rrt/repo/racket/racket/src/build/../racket/gc2/xform-mod.rkt --depends --cpp "gcc -E -I./.. -I/home/rrt/repo/racket/racket/src/build/../racket/gc2/../include -I/home/rrt/repo/racket/racket/src/build/../racket/gc2/../../rktio -I../../rktio -g -O2  -IC:/msys64/mingw64/lib/libffi-3.2.1/include  -DUSE_SENORA_GC   -fno-omit-frame-pointer "  --keep-lines -o ++out xsrc/precomp.h /home/rrt/repo/racket/racket/src/build/../racket/gc2/precomp.c
    In file included from C:/msys64/home/rrt/repo/racket/racket/src/build/../racket/gc2/precomp.c:2:
    C:/msys64/home/rrt/repo/racket/racket/src/racket/src/schpriv.h:10:10: fatal error: scheme.h: No such file or directory
       10 | #include "scheme.h"
          |          ^~~~~~~~~~
    compilation terminated.
    xform: cpp failed
    make[8]: *** [Makefile:205: xsrc/precomp.h] Error 1

The key thing here is that the file that can't be found, `scheme.h`, is indeed in a directory for which there is a `-I` flag, so I don't understand why it's not found.

I can't find any specific instructions in the docs or mailing list archives that seem to pertain to this, so I assume I'm doing something wrong, but I'm unsure what: I just git cloned racket and ran "make".

rrthomas:
I get the same error if I build from a racket-minimal tarball.

pmatos:
@rrthomas Thanks for the report, I will be looking at this.

pmatos:
First attempt to build `master` branch fails in a different place: `libffi`. This is using packaged `libffi` - will try installing `libffi` from `mingw64`.
```
[...]

PR #2923, opened 2019-11-22
Use AC_CHECK_MEMBERS to check for dirent name length flds

pmatos:
This allows us to remote DIRENT_NO_NAMLEN, and simply slightly
rktio_platform.h.

Fix for problem seen while investigating #2906


pmatos:
Sorry - I should just stop assuming I know how autotools works. :-)
Can you pls test it on your setup now? I tested it on a FreeBSD machine which defines `namlen` and it looks ok. You mention `sys/stat` and `sys/dir` but as far as I understand it the struct is defined in `dirent.h`.

PR #2931, opened 2019-11-27
Fix spelling typos

bwignall:
Should be non-semantic.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos.


bwignall:
@sorawee , thank you, I have added your suggested fix.

sorawee:
oh... I have actually deleted that suggestion in GitHub because I am not sure if I parsed that sentence correctly...

bwignall:
I read and re-read it, and it can go either way in my mind. Happy to have whichever you prefer.

PR #2932, opened 2019-11-27
Load core-tests instead of mz-tests

pmatos:
`mz-tests.rktl` became `core-tests.rktl` in 29a0c44c98d5c8d3660e4bd10eb566c7ec6e46e2.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/expand.rktl
+++ b/pkgs/racket-test-core/tests/racket/expand.rktl
@@ -122,7 +122,7 @@
  'expand-load
  #f
  (lambda ()
-   (namespace-set-variable-value! 'expand-load "mz-tests.rktl")))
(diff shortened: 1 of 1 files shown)


pmatos:
> This change makes sense to me.
> 
> The test hasn't been run in a very long time, obviously, and I forgot that it exists. Given that we haven't found the test useful enough to maintain and run, the easy choice would be to just discard it. Still, the first few ways that it currently fails worries me, and those problems may merit further investigation. Since the test driver encodes information about how other parts of the test suite were set up to support it (e.g., `mz-test-syntax-errors-allowed?`), it also makes sense to keep the test around.

Merged.
Shall I open an issue so this is not forgotten?

mflatt:
Good idea. I opened an issue.

PR #2934, opened 2019-11-27
Add per-section coretests

pmatos:
Due to a few failures and flaky tests, I am adding per-section testing for `racket-core-tests`. For now just to CGC but hoping to add 3M and CS soon.

Removing draft status as soon as I see a full pass.


pmatos:
Unfortunately this also revealed a flaky test on macos in the synchronization section: https://github.com/racket/racket/pull/2934/checks?check_run_id=322900862#step:4:49

I will open a bug for it. We might have to do a per-section testing script on CI PR as well.

pmatos:
Removing from draft as tests are passing.

pmatos:
If it looks good to you, let me know since I should squash before merging.

pmatos:
> I'm uncomfortable with this approach, but it's possible that I don't understand.
> 
> I don't see why this needs to be specified external to the test files, and why it needs to be in YAML instead of Racket. If a test doesn't make sense in Racket CGC (and some of them don't), then we should fix the test to declare that it doesn't work in Racket CGC; that way, the test can be skipped reliably in all ways of running the test suite.

Running the `raco test -l tests/racket-test...` way doesn't work because some tests are flaky, others fail for certain configurations etc, so in the short term the idea was to do a per-section testing until we either decide to skip the test in racket, remove it from testing or fix it so it's not failing. In the long term we want to go back to running the tests the way they should be run. I just don't have the time to look into all of them in detail and wanted to create a testing infrastructure that's all green before looking at each test individually. 

I could, of course, skip the tests in racket itself but I thought it made more sense to do it here.

pmatos:
@mflatt Not sure if my earlier message was clear but I meant to finish off by saying that I will take whatever path you feel is better. If skipping sections in racket for CGC at the moment is the way to go even in the short term, I will do that.

pmatos:
@mflatt on the slack #CI channel, @samth clarified the point you were trying to make, which I didn't get. That changing the things straight in Racket and not in Yaml is better among other things because one can more easily reproduce the results locally.

I will be re-factoring this PR taking into consideration your fixes the in the meantime and the comments you made earlier.

.github/workflows/ci-push.yml
```diff
@@ -190,3 +190,36 @@ jobs:
...
+        working-directory: ../
+        run: tar -xvjf racketcgc-${{ matrix.os }}-${{ matrix.cify }}-x64_git${{ github.sha }}.tar.bz2
+      - name: Extend PATH with Racket executable
+        working-directory: ../
+        run: echo "::set-env name=PATH::$PWD/racketcgc/bin:$PATH"
+      - name: Check for Racket
+        run: racket --version
+      - name: Install catalog and required packages
+        run: |
+          racket -l- pkg/dirs-catalog --immediate $PWD/rktcat $PWD/pkgs/
```
samth: Can't you do this with `make PKGS="racket-test"`?

pkgs/racket-test-core/tests/racket/core-tests.rktl
```diff
@@ -25,7 +25,11 @@
 (load-relative "deep.rktl")
 (load-relative "contmark.rktl")
 (load-relative "prompt.rktl")
-(load-relative "will.rktl")
+(unless (eq? (system-type 'gc) 'cgc)
```
samth: Should we just do this for the specific test that fails?
Also, we could just mark this whole test as allowed-to-fail, right?

.github/workflows/ci-push.yml
pmatos: Good question - isn't that incantation to build the whole racket? Once we have racket built, we can just add those dirs to the catalog, no?

pkgs/racket-test-core/tests/racket/core-tests.rktl
pmatos: We could do that just for the test that fails, if it's simple to find out which test that is. I have honestly not looked at the code for `will.rktl`.

Regarding marking the test as allowed-to-fail, I am not sure in GitHub yet but when you do that in GitLab, the whole CI is marked as yellow and I found that really makes the whole system pointless because if a job is marked as yellow, the whole pipeline is marked as yellow it makes it much harder to understand if the root cause for the CI yellowness changed, or not. However, this is speaking from GitLab experience, not GitHub. Unless you are really against this, I would prefer to mark for now the test as skipped on `cgc`.

pmatos: Just pushed a change to move the skip check into `will.rktl`. Unfortunately finding the `sync` tests to skip is proving to be harder. Is there any good way to know in which line the failing tests are in?

samth: Basically you look for calls that seem similar to the failures. Otherwise you run the file directly and look at where the failure appears in the large amount of output.

pmatos: Ah - thanks. I think wtih `sync` with need to go the generate large amount of output and explore. :-)

pmatos:
Of course - locally I cannot reproduce the sync errors to understand which tests are failing. I will temporarily make the tests verbose in the PR, so I can find out what's happening and then skip them.

mflatt:
@pmatos I think I probably fixed the `sync` test, but maybe you were seeing a different problem than the one I saw.

pmatos:
> @pmatos I think I probably fixed the `sync` test, but maybe you were seeing a different problem than the one I saw.

Oh, let me try rebasing then.

pmatos:
I have rebased and pushed/forced. Lets see!

pmatos:
It has passed the tests: https://github.com/LinkiTools/racket/commit/2873b1521343e8a4f4c872e369c47b6df11ea345/checks?check_suite_id=333405985

Are we happy with this? If so, I will squash and merge. Requesting review.

samth:
Looks good to me

pmatos:
Thanks @samth

PR #2944, opened 2019-11-30
More complete set of tests in Actions

pmatos:
This PR adds a more complete set of tests for Linux (Ubuntu 18.04) and MacOS.

It runs the same tests as travis and the others.

### DIFF
--- a/.github/workflows/ci-push.yml
+++ b/.github/workflows/ci-push.yml
@@ -220,28 +220,42 @@ jobs:
         run: echo "::set-env name=PATH::$PWD/racketcgc/bin:$PATH"
       - name: Check for Racket
         run: racket --version
-      - name: Install catalog and required packages
+      - name: Install catalog
         run: |
           racket -l- pkg/dirs-catalog --immediate $PWD/rktcat $PWD/pkgs/
           raco pkg config --set catalogs $PWD/rktcat/ https://pkgs.racket-lang.org https://planet-compats.racket-lang.org
-          raco pkg install --auto racket-test
-      - name: Run tests
-        run: |
-          raco test -l tests/racket/test
-          racket -l tests/racket/contract/all
-          raco test -l tests/json/json
-          raco test -l tests/file/main
-          raco test -l tests/net/head
-          raco test -l tests/net/uri-codec
-          raco test -l tests/net/url
-          raco test -l tests/net/url-port
-          raco test -l tests/net/encoders
-          raco test -l tests/openssl/basic
-          raco test -l tests/openssl/https
-          raco test -l tests/match/main
-          raco test -l tests/zo-path
-          raco test -c tests/xml
-          raco test -c tests/stxparse
+      - name: Install racket-test dependency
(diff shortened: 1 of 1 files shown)


samth:
@mflatt there's no `tests/main.rkt` for that to run (but maybe there should be)

.github/workflows/ci-push.yml
```diff
@@ -226,7 +226,26 @@ jobs:
           raco pkg config --set catalogs $PWD/rktcat/ https://pkgs.racket-lang.org https://planet-compats.racket-lang.org
           raco pkg install --auto racket-test
       - name: Run tests
-        run: raco test -l tests/racket/test
+        run: |
```
samth: One drawback of writing it this way is that it gets run as a single shell command, so it's harder to tell what failed.

.github/workflows/ci-push.yml
```diff
@@ -226,7 +226,26 @@ jobs:
...
+          raco test -l tests/racket/test
+          racket -l tests/racket/contract/all
+          raco test -l tests/json/json
+          raco test -l tests/file/main
+          raco test -l tests/net/head
+          raco test -l tests/net/uri-codec
+          raco test -l tests/net/url
```
samth: I thought we needed to install "net-test" as well.

mflatt:
Sorry — I meant `raco test -c tests`.

samth:
That runs substantially more tests, including all the pkg tests, all the futures tests.

The other drawback is that a single command makes it harder to tell which test failed from the output. It's unfortunate that there's a tradeoff between scripting things in Racket instead of YAML and making them easy to understand in the output, but I don't know a way around it at the moment.

pmatos:
> That runs substantially more tests, including all the pkg tests, all the futures tests.
> 
> The other drawback is that a single command makes it harder to tell which test failed from the output. It's unfortunate that there's a tradeoff between scripting things in Racket instead of YAML and making them easy to understand in the output, but I don't know a way around it at the moment.

Maybe we can run those nightly? as in `raco test -c tests`?

.github/workflows/ci-push.yml
pmatos: Hummm, I assume not since the test run did not fail but I will check.

.github/workflows/ci-push.yml
pmatos: Right - I will make it one step per test before merging.

.github/workflows/ci-push.yml
pmatos: It seems net-test is not needed to install as a package because it's installed as a dependency of `racket-test` installed earlier:
```
Downloading repository git://github.com/racket/string-constants/?path=string-constants-lib
The following uninstalled packages were listed as dependencies
and they were automatically installed:
 dependencies of racket-test:
   compiler-lib
   sandbox-lib
   compatibility-lib
   eli-tester
   planet-lib
   net-lib
   net-test
...
```

PR #2945, opened 2019-11-30
Actions for Windows CI

pmatos:
Addings CI on Windows using Actions.

### DIFF
--- a/.github/workflows/ci-push.yml
+++ b/.github/workflows/ci-push.yml
@@ -190,11 +190,29 @@ jobs:
         name: racketcs-${{ matrix.os }}-x64_git${{ github.sha }}
         path: ../racketcs-${{ matrix.os }}-x64_git${{ github.sha }}.tar.bz2
 
-# Tests
-# Unfortunately Actions does not support atm yaml anchors
-# otherwise all the following jobs could be simplified
-# Note: the reason we cannot transform this into a matrix
-# build is because we cannot use variables in the needs keyword.
+  build-win:
+    runs-on: windows-latest
+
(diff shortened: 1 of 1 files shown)


pmatos:
Need to rebase this one before merging... Thanks for the `review+`.

pmatos:
I am having some problems doing the `csbuild` on windows.

It seems that on Windows the ChezScheme checkout doesn't not do a `--recurse-submodules`.
https://github.com/LinkiTools/racket/commit/12da59fbd4358e3b8b0ea3e31d6e864e50d23997/checks?check_suite_id=336770366#step:3:6482

I attempted to pass those options to `csbuild.rkt` but was not successful. What's the correct way to build CS on Windows?

pmatos:
To solve the above issue I used `csbuild.rkt -- --recurse-submodules --depth=1`. These were in turn passed to `git` during `clone`. @mflatt shouldn't we mention this in the `README` or automatically clone submodules?

pmatos:
> To solve the above issue I used `csbuild.rkt -- --recurse-submodules --depth=1`. These were in turn passed to `git` during `clone`. @mflatt shouldn't we mention this in the `README` or automatically clone submodules?

I am actually a bit confused as to why a simple call of `csbuild.rkt` works on CI PR but not on CI Push.

mflatt:
The `csbuild.rkt` script sets up submodules when `--pull` is provided, but `git clone` is run if the directory doesn't exist, even if `--pull` is not provided. It would be sensible to always include `--recurse-submodules` in that automatic `git pull`.

PR #2948, opened 2019-12-02
Remove support for azure-pipelines

pmatos:
At this point Actions PR CI and Push CI cover all tests in Azure Pipelines so they
are not needed.


samth:
I also disabled this in the Azure Pipelines settings.

samth:
Next question -- should we remove the AppVeyor tests as well? I believe they're all covered, although I don't know about versions of Windows/Visual Studio.

pmatos:
> Next question -- should we remove the AppVeyor tests as well? I believe they're all covered, although I don't know about versions of Windows/Visual Studio.

I will take a look next.

PR #2956, opened 2019-12-08
doc: pidefterm -> deftech

bennn:
Use `deftech` for "structure subtype" so that Typed Racket can more
easily reference it

- - -

After this change, a `raco setup` finished cleanly. (Though IIUC, all `pidefterm` should be safe to replace with `deftech`)


mflatt:
LGTM

PR #2959, opened 2019-12-10
On Linux build with clang as well

pmatos:
On Push CI build with `clang` besides gcc.
This is one step towards replacing the current testing on travis - next up is the configure flags matrix.

### DIFF
--- a/.github/workflows/ci-push.yml
+++ b/.github/workflows/ci-push.yml
@@ -127,9 +127,11 @@ jobs:
       working-directory: ./racket/src
       run: make -j  $((cpus+1)) install
     - name: Tarballing
+      if: matrix.cc == 'gcc'
       working-directory: ../
       run: tar -cvjf racket3m-${{ matrix.os }}-${{ matrix.cify }}-x64_git${{ github.sha}}.tar.bz2 racket3m
     - uses: actions/upload-artifact@master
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-push.yml
```diff
@@ -121,9 +127,11 @@ jobs:
       working-directory: ./racket/src
       run: make -j  $((cpus+1)) install
     - name: Tarballing
+      if: matrix.cc == 'gcc' || matrix.os == 'macos-latest'
```
samth: Why do we explicitly need to test this here?

pmatos: We do not need to test racket build with `gcc` and `clang`. We build with `clang` but do not test the build. We test the `gcc` one. So we only tarball the `gcc` version. At the same time, `macos` only builds with `clang`, so we tarball in that case as well.

Happy to change this if you think it's worth it.

samth: That seems fine (the odds of finding a miscompilation in clang that causes test failures is fine). Can you add a comment to that line about why?

pmatos:
@samth requested comment clarification added at https://github.com/racket/racket/commit/b14b2400fd3ebf693fee53b144063f4fd6473b47

Issue #2963, opened 2019-12-12
racket3m nojit, noefp fails memory test in will section


pmatos:
With Racket3m on both Linux and Mac with cify or no cify, if I compile racket with no jit and no extflonums, no futures and no places, the test:
```
Errors were [1]:
(Section (got expected (call)))
((wills) (#f #t (many-vectors-in-reasonable-space? #f)))
```
fails predictably.

Here's all the occurrences:
https://github.com/LinkiTools/racket/runs/345208378#step:9:41
https://github.com/LinkiTools/racket/runs/345208430#step:9:40
https://github.com/LinkiTools/racket/runs/345208517#step:9:40
https://github.com/LinkiTools/racket/runs/345208599#step:9:42

I expected this to be dependent on gc behaviour, however there might be a relation between no jit and gc behaviour. efp might in itself not be the problem.

@mflatt should this test be disabled without a jit?

pmatos:
Unfortunately this still seems to be failing after 58f51f1: https://github.com/LinkiTools/racket/runs/347193643#step:10:43

mflatt:
That's a different test. It checks whether a contract applications runs in under 5 seconds, so it's easy to see why that causes problems. I don't yet have a good idea on how to handle this one.

samth:
That problem is #2964.

pmatos:
> That problem is #2964.

Right - mixed this up. Sorry.

pmatos:
Fixed wtih https://github.com/racket/racket/commit/58f51f15eddfe41f4a6c17630091441230fa93ac
Thanks.

PR #2965, opened 2019-12-12
Disable will section test with jit disabled

pmatos:
Related to #2963

### DIFF
--- a/pkgs/racket-test-core/tests/racket/will.rktl
+++ b/pkgs/racket-test-core/tests/racket/will.rktl
@@ -388,7 +388,8 @@
                       'done
                       (unbox (loop (sub1 n)))))))))
 
-(unless (eq? 'cgc (system-type 'gc))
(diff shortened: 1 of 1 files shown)


pmatos:
Would it be ok to disable this to fix #2963 or do you have another suggestion?

mflatt:
I'm looking into this, because I don't see a reason that it should fail with 3m. But I also haven't been able to trigger a failure on my machine, yet.

pmatos:
You might need a slower machine. I have one around. Will try it tomorrow if you don't sort it out in the meantime.

mflatt:
I think I understand now. It's a question of the cost of non-tail recursion, which is higher without the JIT. I'll adjust the test.

pmatos:
Closing without merging as 58f51f15eddfe41f4a6c17630091441230fa93ac is a better fix by @mflatt

PR #2968, opened 2019-12-12
update expander logging

rmculpepper:
This simplifies the expander logging in some places and adds logging for arming/disarming and scope changes (eg syntax-local-introduce) so the macro stepper can better track term identity.

This relies on corresponding changes to the racket/macro-debugger repo.

This needs a version bump when it gets merged.


mflatt:
LGTM

rmculpepper:
Thanks @mflatt. Should I squash this or keep the commits separate?

mflatt:
I would squash, but I have no problem with keeping them separate if you prefer that.

PR #2969, opened 2019-12-12
Fix typos

bwignall:
Should be non-semantic.

Uses https://en.wikipedia.org/wiki/Wikipedia:Lists_of_common_misspellings/For_machines to find likely typos, with https://github.com/bwignall/typochecker to help automate the checking.


samth:
I think those changes will be fine.

bwignall:
@bennn , @samth : if there are any concerns, I would be happy to revert those changes, if you want. It's not my intention to make semantic changes.

mflatt:
LGTM. I'll merge.

PR #2975, opened 2019-12-16
Update fixnums documentation to suggest require'ing only fixnum-relat…

dzoep:
…ed parts of racket/unsafe/ops and not the rest of unsafe ops.

It should prevent users from require'ing unsafe versions of for example vector-* operations which may lead to crashes in different parts of the code which do not need to use fx* operations at all.


dzoep:
That is another option. That way all of racket/unsafe/ops gets in but only unsafe-fx* gets renamed to fx*. I would prefer pulling only unsafe-fx* as fx* and keeping the namespace otherwise untouched (as it is in my PR right now) - but I do not know how others might be using it and when. Maybe we should get some input from more users?

samth:
Given the lack of additional input, I merged this as-is.

PR #2976, opened 2019-12-16
Modules with tests for unsafe mode can be cross-phase persistent.

samth:
Also, mark several more modules as cross-phase persistent.


mflatt:
This needs a doc update, but it seems ok otherwise.

Issue #1699, opened 2017-05-23
Suggestion: struct/contract & struct/derived


winny-:
I really like contract programming, but was surprised when I noticed there is a define-struct/contract, but no struct/contract. Maybe racket should ship with a struct/contract syntax?

LiberalArtist:
+1

LiberalArtist:
And relatedly, a `struct/derived` similar to `define-struct/derived` would be very useful.

hashimmm:
Additionally, since `define-struct/contract` supports `#:property`, it should be able to support `#:methods`, yes? Maybe `struct/contract` version could do that, too. Does that suggesteion merit a separate issue?

LiberalArtist:
`struct` is actually implemented via `define-struct/derived`, so that part would be a particularly easy pull request, I think. I've never looked at the implementation of `define-struct/contract`, but I imagine it would be similar. The only difference between `struct` and `define-struct` is that `struct` adds a `#:constructor-name` option unless there already is one.

https://github.com/racket/racket/blob/5c775fa04c11697a25655b1738aa96f14368745a/racket/collects/racket/private/struct.rkt#L11-L28

spdegabrielle:
Potential "Good first bug"?
If so - please add label.

spdegabrielle:
Potential "Good first bug"?
If so - please add label.
Was suggested at https://github.com/racket/racket/wiki/Racketeer-Office-Hours-2017-Task-Ideas

PR #2979, opened 2019-12-17
Adds struct/contract

tommymchugh:
Implementation for one of the features requested in #1699

### DIFF
--- a/racket/collects/racket/contract/region.rkt
+++ b/racket/collects/racket/contract/region.rkt
@@ -1,6 +1,7 @@
 #lang racket/base
 
 (provide define-struct/contract
+         struct/contract
          define/contract
          with-contract
          current-contract-region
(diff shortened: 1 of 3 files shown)


rfindler:
Thanks for your interest and work!

I am sorry, however, it is hard to see what you have changed here, due to
the very large number of changes to white space. Can you please separate
those out?

Robby

rfindler:
Oh, sorry! I have found the "ignore whitespace changes" feature in the github GUI now. I'll take a look.

rfindler:
I think the change has to be a little more nuanced because `struct` doesn't bind the constructor as `make-...`, the "make-" prefix isn't there.

Also, it looks like you've moved ever so slightly away from the style guide guidelines by moving a helper function into a `let*` out of a `define`.

tommymchugh:
Gotcha! Thanks for the direction on this, I'll work on this

tommymchugh:
@rfindler Sorry for the delay getting back on this, was on holiday past few days. This now takes into account all the differences between struct and define-struct.

tommymchugh:
43e4128 cleans code to remove whitespaces that were very annoying to review code with.

racket/collects/racket/contract/region.rkt
```diff
@@ -123,332 +124,392 @@
...
+                                                            (list
+                                                             #'(define-syntax struct-name
+                                                                 (make-struct-info
+                                                                  (λ ()
+                                                                    (list #f #f #f
+                                                                          (list #f) (list #f)
+                                                                          #,(quoter #'super))))))]))]
+                                                      [define-returned-value-temps (if (equal? type 'define-struct/contract)
+                                                                                       #`(define maker-value ... temp-maker-name ...)
+                                                                                       (void))])
```
rfindler: This should be #`(void), not (void).

rfindler:
It looks like this part is basically ready. The docs update is missing. I can do that if you'd like.

Also, I see that the program below produces the unfortunate output below it (there is an extra argument showing in the contract in the error message), but this doesn't seem to be a problem with the changes you're doing here, as it also happens for `define-struct/contract`.

```
#lang racket
(struct/contract s ([x integer?][y boolean?]))
(s 1 2)
```

produces this output:

```
$  racket ~/tmp.rkt
s: contract violation
  expected: boolean?
  given: 2
  in: the 2nd argument of
      (-> integer? boolean? symbol? any)
  contract from: (struct s)
  blaming: /Users/robby/tmp.rkt
```

tommymchugh:
@rfindler I took care of the docs and the syntax quasiquote addition, and updated the define-struct/contract to better match the styling of struct and define-struct relationship in the docs. 

Yeah, in regards to the weird error statement, without looking at it deeply, I think it has to be because of the structure of the `#:guard` procedure that was used to enforce the contract, which isn't changed with these additions. I'll file an issue and take a little deeper look over next few days to see what's going on with that and see if we can change anything to better reflect the contract violation

rfindler:
Okay, can you squash and rebase and then I'll merge?

tommymchugh:
Done. Once tests pass again, should be good to go.

rfindler:
Thanks!

PR #2980, opened 2019-12-17
Trim requires.

samth:
Eliminates another ~300 instantiations from `racket -cl racket/base`.

small-scheme.rkt is no longer used in `racket/private`.

c.rkt was an accidentally-committe file.


mflatt:
Ok!

samth:
Thanks to @rmculpepper for `raco check-requires`.

PR #2982, opened 2019-12-19
Fix typos, via a Levenshtein-style corrector

bwignall:
Should be non-semantic.

Uses an improved version of the tool that found typos in https://github.com/racket/racket/pull/2969 .


pmatos:
Thanks!

PR #2987, opened 2019-12-19
Pass ubsan flags to rktio build

### LINKED ISSUE #2984: ubsan flags not passed to rktio
When compiling to `--enable-ubsan`, the ubsan flags are not passed to the rktio build.

```
$ ./configure --prefix=/home/pmatos/installs/racket-7.5_ubsan --enable-ubsan --enable-futures --enable-jit --enable-places --enable-extflonum --disable-docs --enable-threads
...
$ make
...
gcc -g -O2 -Wall   -pthread -I. -I. -o rktio_fs.o -c ./rktio_fs.c
...
```

pmatos:
We have also removed ubsan flags from CPPFLAGS - since these are
compile flags, not preprocessing flags.

To pass the flags to rktio, we split the use of CFLAGS and CPPFLAGS
in rktio and ensure that flags in rktio do not update flags if make
is called in sub-make form (in a standard racket build, the rktio make
is called as a sub-make from the main racket make).

Fixes #2984


pmatos:
I just looked at the changes and noticed that for some reason this `runstatedir` thing is being added. I wonder why. For a long time racket had these and my arch linux kept removing them so I was always careful not to commit those blocks. Recently @mflatt removed them and I have just created this change in gentoo linux, which is trying to add those again. 

Will add a commit to remove these again.

mflatt:
I think it's a question of autoconf version. It doesn't matter to me if `runstatedir` shows up again.

pmatos:
> I think it's a question of autoconf version. It doesn't matter to me if `runstatedir` shows up again.

Think it is the autoconf version. Arch has a patched up version of autoconf which includes this runstatedir thing. I have removed from the PR anyway. Let me know what you think about the changes.

pmatos:
> I'm not immediately convinced that these flags only matter for compilation. Macros sometimes reflect on compiler options. For example, including `-O2` in `CPPFLAGS` tends to matter for builsing Racket3m. Does including ubsan flags in `CPPFLAGS` cause trouble, or was it just meant to clean up?

It should only matter for compilation. `CPPFLAGS` should be preprocessor only options - used for `gcc -E` and friends. If `-O2` causes differences when not passed in `CPPFLAGS`, there's a bug somewhere. It would be interested to understand why.

In any case, it was just clean-up. I am happy to add them back until I have time to investigate.

> 
> Meanwhile, I am somehow missing the part where `CPPFLAGS` is propoagate to the `rktio` configure. Can you help me see exactly where that happens?

These are propagated due to the `export CPPFLAGS`, similarly with `export CFLAGS`. These are exported to sub-make calls.
Line 33: https://github.com/LinkiTools/racket/blob/f1eb284fbb73037572e6a6c09b4e154a3e2fb6fa/racket/src/racket/Makefile.in#L33

This is described in https://www.gnu.org/software/make/manual/html_node/Variables_002fRecursion.html#Variables_002fRecursion

mflatt:
It looks like `export` may be a GNU make extension, although I'm not sure. How about instead passing the variables in the `$(make)` call for rktio, which is the way that's usually done in the makefiles currently?

For an example of how flags that you normally think of as compiler flags can affect preprocessing, see the use of `__SSE_MATH__` in "scheme.h". I don't know that any Racket header files check the optimization mode (although I may forget); mostly I've seen it happen with system header files. I understand the idea that compiler flags *shouldn't* affect preoprocessing, but that's not how it works out.

pmatos:
> For an example of how flags that you normally think of as compiler flags can affect preprocessing, see the use of `__SSE_MATH__` in "scheme.h". 

You mean `__SSE2_MATH__`?
```
/* gcc defines __SSE2_MATH__ when SSE2 floating point is enabled: */
#ifdef __SSE2_MATH__
# define C_COMPILER_USES_SSE 1
#endif

#ifdef C_COMPILER_USES_SSE
# if defined(MZ_TRY_EXTFLONUMS) && !defined(MZ_NO_EXTFLONUMS)
#  define MZ_LONG_DOUBLE
#  ifdef ASM_DBLPREC_CONTROL_87
#   define ASM_EXTPREC_CONTROL_87
#  endif
# endif
# ifdef ASM_DBLPREC_CONTROL_87
#  undef ASM_DBLPREC_CONTROL_87
# endif
# if defined(MZ_USE_JIT_I386) && !defined(MZ_NO_JIT_SSE)
#  define MZ_USE_JIT_SSE
# endif
#endif
```

mflatt:
Yes, that's the one I meant to write.

pmatos:
> It looks like `export` may be a GNU make extension, although I'm not sure. How about instead passing the variables in the `$(make)` call for rktio, which is the way that's usually done in the makefiles currently?
> 

Lets do that instead. As I understand it, the FreeBSD make might not support this - need to get FreeBSD onto CI...

pmatos:
> Yes, that's the one I meant to write.

Still don't quite understand the use of compiler flags in this example. It just defines a bunch of preprocessing defines. What I am interested in knowing is, how can a compilation flag like -O2 passed to the preprocessor can change the preprocessor output.
 
```
$ gcc -E output.c -o output1.i
$ gcc -O2 -E output.c -o output2.i
```

If that's the case the result of these (`output1.i` and `output2.i`) would be different. I understand that the issue might be in how racket build system works, in which case we would be better off fixing the build system, right?

mflatt:
Ok, let me dig up the specifics... Try compiling this program with and without `-O2`:
```
#include <stdio.h>

int main()
{
#ifdef __OPTIMIZE__
  printf("optimizing\n");
#else
  printf("not optimizing\n");
#endif
  return 0;
}
```
Then, grep for `__OPTIMIZE__` in "/usr/include".

On an Ubuntu 16.04 installation with gcc 5.4.0, I get different compilation results and lots of grep hits.

pmatos:
For the record, actually BSD make does support `export variable=value`. 

>      .export variable ...
> 	     Export the	specified global variable.  If no variable list	is
> 	     provided, all globals are exported	except for internal variables
> 	     (those that start with `.').  This	is not affected	by the -X
> 	     flag, so should be	used with caution.  For	compatibility with
> 	     other make	programs `export variable=value' is also accepted.

https://www.freebsd.org/cgi/man.cgi?make(1)

But lets not introduce a new feature into the build system - instead I will do the change to pass them explicitly to make.

pmatos:
> Ok, let me dig up the specifics... Try compiling this program with and without `-O2`:
> 
> ```
> #include <stdio.h>
> 
> int main()
> {
> #ifdef __OPTIMIZE__
>   printf("optimizing\n");
> #else
>   printf("not optimizing\n");
> #endif
>   return 0;
> }
> ```
> 
> Then, grep for `__OPTIMIZE__` in "/usr/include".
> 
> On an Ubuntu 16.04 installation with gcc 5.4.0, I get different compilation results and lots of grep hits.

Yes - I know what you mean:
```
$ gcc -E -dM - < /dev/null | grep OPTIMIZE    
$ gcc -O1 -E -dM - < /dev/null | grep OPTIMIZE
#define __OPTIMIZE__ 1
$ gcc -Os -E -dM - < /dev/null | grep OPTIMIZE
#define __OPTIMIZE__ 1
#define __OPTIMIZE_SIZE__ 1
```

But GCC also says (https://gcc.gnu.org/onlinedocs/cpp/Common-Predefined-Macros.html):
**__OPTIMIZE__
__OPTIMIZE_SIZE__
__NO_INLINE__

    These macros describe the compilation mode. __OPTIMIZE__ is defined in all optimizing compilations. __OPTIMIZE_SIZE__ is defined if the compiler is optimizing for size, not speed. __NO_INLINE__ is defined if no functions will be inlined into their callers (when not optimizing, or when inlining has been specifically disabled by -fno-inline).

    These macros cause certain GNU header files to provide optimized definitions, using macros or inline functions, of system library functions. **You should not use these macros in any way unless you make sure that programs will execute with the same effect whether or not they are defined.** If they are defined, their value is 1.**

So, while I understand the sentiment the racket system shouldn't depend on these unless we are providing specific variants which depend on how racket was compiled, which afaik we aren't.

My attempt here was to simplify the build and not put all the flags into the same basket. If we don't make use of `__OPTIMIZE__` and friends (we probably shouldn't), I think we should, at least going forward, attempt to separate concerns. The build system is complex enough and mixing both types of flags doesn't help. Are you worried that separating `CFLAGS` and `CPPFLAGS`  is going to create an unoptimized rktio library?

mflatt:
The original question was whether the ubsan flags belong in `CFLAGS` or `CPPFLAGS`. It's not a question of "mixing" the sets of flags but of deciding where to put some specific flags.

Putting flags in `CPPFLAGS` *does* help. Time and time again, I've run into trouble putting flags only in `CFLAGS` and not in `CPPFLAGS`, because it turns out that system headers depend on those flags.

This is because Racket3m uses `gcc -E` as part of its build process. Even though the Racket sources don't use `__OPTIMIZE__` dierectly, they do use system headers, and system headers use preprocessor definitions to expand differently depending on compile flags—even ones that you might not think affect preprocessing.

The problem may well start with the idea of using `gcc -E`—and that's never going to happen with `rktio`, I think. But I assume the ubsan flags are meant to apply to the Racket3m build, and that's why I suggest putting them in `CPPFLAGS`.

pmatos:
> The original question was whether the ubsan flags belong in `CFLAGS` or `CPPFLAGS`. It's not a question of "mixing" the sets of flags but of deciding where to put some specific flags.
> 

https://gcc.gnu.org/onlinedocs/gcc-9.2.0/gcc/Instrumentation-Options.html#Instrumentation-Options
These `ubsan` options are to control runtime instrumentation. They shouldn't go into `CPPFLAGS`. 

> Putting flags in `CPPFLAGS` _does_ help. Time and time again, I've run into trouble putting flags only in `CFLAGS` and not in `CPPFLAGS`, because it turns out that system headers depend on those flags.
> 

OK, lets say that you are right and there are flags passed in `CPPFLAGS` that change runtime behaviour. If your comment a general comment, or specific to `ubsan` options. If it's unrelated to `ubsan` options, then I can we can pass them only in `CFLAGS`, right?

> This is because Racket3m uses `gcc -E` as part of its build process. Even though the Racket sources don't use `__OPTIMIZE__` dierectly, they do use system headers, and system headers use preprocessor definitions to expand differently depending on compile flags—even ones that you might not think affect preprocessing.
> 

OK - I am still confused about which flags could affect the build like that, even assuming we abuse `gcc -E`. However, I also recognize you have been working on racket for eons and seen all sorts of trouble but if we have no specific case against passing `ubsan` in `CFLAGS` only, would you be ok with it as it is?

> The problem may well start with the idea of using `gcc -E`—and that's never going to happen with `rktio`, I think. But I assume the ubsan flags are meant to apply to the Racket3m build, and that's why I suggest putting them in `CPPFLAGS`.

If we put `ubsan` options in `CPPFLAGS` I assume along the build we'll see `gcc -E -fsanitize=undefined ...`, which doesn't make sense. Unless of course, racket picks out the options from the `gcc -E`  from the command line to go ahead and use it to compile something in the background but then we have a different problem.

I know, by the way, I am being annoying with this. I apologize, but I would really like to slowly clean this up. Assume `ubsan` works in `CFLAGS` only - if it doesn't we'll notice in CI (like I just did with `rktio` and then we'll revisit. I won't take any more of your time. You have the last word. If you say - I want it in `CPPFLAGS`, I will put them in `CPPFLAGS`.

mflatt:
Here's the kind of thing I'm concerned about: https://gcc.gnu.org/ml/gcc-patches/2019-09/msg01286.html

mflatt:
I thin the emoji reaction means that I'm still not clear, so I'll try again.

Suppose that

 * A patch is merged to GCC so that `-fsanitize=undefined` defines a preprocessor macro.

 * A system header file starts using the preprocessor macro to expand differently when `-fsanitize=undefined` is enabled.

 * Racket source file `eval.c` includes that system header with the consequence that the expansion of `eval.c` is different with and without `-fsanitize=undefined`.

Then, if `-fsanitize=undefined` is in `CFLAGS` and not `CPPFLAGS`, building Racket3m goes like this:

 * `gcc -E eval.c` + xform -> `xsrc/eval.c`

 * `gcc -fsanitize=undefined xsrc/eval.c` -> `eval.o`, maybe

At this point, `eval.o` has been compiled in an inconsistent way. The expansion of system-header macros assumed no `-fsanitize=undefined`, but the expansion result was compiled with `-fsanitize=undefined`. Best case scenario is a compilation error. Just as likely, if past experience is any guide, `eval.o` is generated but Racket crashes at runtime due to the compile-time mismatch.

It may never happen that compilers acquire a preprocessor macro as a result of `-fsanitize=undefined` or that system headers start using it. But, again, past experience suggests that it's possible, perhaps even likely. I was surprised to find a specific proposal to add a preprocessor flag that I could point to, but only a little surprised.

pmatos:
Actually I completely understood the message with the link to the gcc ml. The emoji was meant to express discontentment. 

Although I just noticed the emoji label is confused - which was not the intention. My apologies. Your further explanation confirmed what I was thinking from your ml reference.

pmatos:
> Here's the kind of thing I'm concerned about: https://gcc.gnu.org/ml/gcc-patches/2019-09/msg01286.html

I am with Jakub here that this shouldn't be merged into GCC though, and it wasn't so far. However, it made into llvm: https://clang.llvm.org/docs/LanguageExtensions.html#extensions-for-dynamic-analysis
This was the reason for my emoji.

From where I stand, I still think this is a mistake - as I think this shouldn't go into gcc. However, in pragmatic terms, what you say is correct. If include headers start using `__has_feature(address_sanitizer)` and friends then we have a problem. It still shouldn't be a problem is the semantics of the preprocessed code remains the same but...  semantics! 

I am not happy but I might have to capitulate on this one. At least until we have more CI coverage - one day I would like to backtrack, do the change and see what breaks. I recognize now, without better testing, it might be too risky.

I will update the patch and go have a beer.

pmatos:
Added a new patch https://github.com/racket/racket/pull/2987/commits/cfe34422139eec3bb5e20b5e2dd2b5b45f7598f1
It shows up incorrectly on my github conversation timeline because my clock is set 1hr in the past, which messed up with the git commit time, I think.

mflatt:
Beer well deserved.

Last little question: is the `?` in `CFLAGS ?= @CFLAGS@` needed? I think it was needed when using environment variables, but not when using makefile arguments.

I've avoided `?=` in general, but only because I haven't tried to use it much and get confused about what is allowed by different `make`s. (Is `nmake` relevant here? I don't think so.)

pmatos:
> Beer well deserved.
> 

:+1: 
but not yet ... lets merge this first. 

> Last little question: is the `?` in `CFLAGS ?= @CFLAGS@` needed? I think it was needed when using environment variables, but not when using makefile arguments.
> 

Right - this is only needed with environment variables. It will define `CFLAGS` and `CPPFLAGS` if not defined in the environment. New patch added.

> I've avoided `?=` in general, but only because I haven't tried to use it much and get confused about what is allowed by different `make`s. (Is `nmake` relevant here? I don't think so.)

Microsoft nmake or Fowlers nmake?
Can we agree here that for *NIX systems we support only 2 makes: `GNU make`, and `BSD make`?
If you want to support more makes, which ones do you have in mind? I would like to setup CI build jobs with these makes to guarantee continued support.

pmatos:
Interestingly Fowlers nmake seems to be no longer maintained. I assume then, you refer to Microsoft name, which is not relevant here.

samth:
Definitely `nmake` in the racket build system context means the one for Windows.

pmatos:
@mflatt waiting for review before merging.

pmatos:
All good, merging! I will be away for a bit. Have a great holiday season. Merry Christmas and a happy new year.

PR #2991, opened 2019-12-28
Fix #2986: error-print-width 3 shouldn't error

### LINKED ISSUE #2986: (error-print-width 3) incorrectly errors in 3m
This line https://github.com/racket/racket/blob/master/racket/src/racket/src/error.c#L3066 should be `>=`, not `>`.

sorawee:


mflatt:
Thanks! But this needs a test case, probably near other `error-print-width` tests in `pkgs/racket-test-core/tests/racket/print.rktl`.

sorawee:
@mflatt hmm, this looks correct to me but the test is failing. Do you do what I did wrong?

bennn:
Does it work if you run locally?

Maybe each `test` needs its own `(parameterize ....)`

gus-massa:
The old test above use `unquoted-printing-string` https://github.com/racket/racket/blob/996c91bc79a6a571bf087083d14ed7c61bc3b9a9/pkgs/racket-test-core/tests/racket/print.rktl#L253-L254

But you are not using it https://github.com/racket/racket/blob/996c91bc79a6a571bf087083d14ed7c61bc3b9a9/pkgs/racket-test-core/tests/racket/print.rktl#L309-L310

I still don't understand why that is relevant and why `error-print-width` is changing the output if neither tests use the error port. My guess is that there is some hidden redirection, but I can't find it.

sorawee:
Right. I'm not using it because the tests in that section are really testing `unquoted-printing-string`, and I don't mean to test it.

Re @bennn: 

1. No, it doesn't work when I run locally.
2. It should work, because there are similar multiple tests that are in the same body of a `parameterize` in the same file. If it works there, it should work here too.

gus-massa:
[Perhaps I'm missing an easier way.] The problem is that `error-print-width` is only used when an exception is build by `raise-???-error`. So the easiest way to test it that I found was

    #lang racket

    (define (format/errmsg x)
      (with-handlers ([exn:fail?
                       (lambda (e)
                         (substring (exn-message e) 52))])
        (raise-argument-error 'Hello "Word" x)))

    (parameterize ([error-print-width 10])
      (format/errmsg "123456789012345678901234567890"))
    ; ==> "\"123456..."

    (parameterize ([error-print-width 20])
      (format/errmsg "123456789012345678901234567890"))
    ; ==> "\"1234567890123456..."
      
    (parameterize ([error-print-width 40])
      (format/errmsg "123456789012345678901234567890"))
    ; ==> "\"123456789012345678901234567890\""

    (newline)

    (parameterize ([error-print-width 10])
      (format/errmsg 123456789012345678901234567890))
    ; ==> "1234567..."

    (parameterize ([error-print-width 20])
      (format/errmsg 123456789012345678901234567890))
    ; ==> "12345678901234567..."
      
    (parameterize ([error-print-width 40])
      (format/errmsg 123456789012345678901234567890))
    ; ==> "123456789012345678901234567890"

sorawee:
Ouch. Thanks for the investigation. It's weird because the `~.a` is supposed to use `error-print-width`. In any case, I'd probably avoid doing `with-handlers`/`raise-argument-error` and just use `unquoted-printing-string` for simplicity.

gus-massa:
I didn't know `"~.a"`. That's really much simpler.

    (parameterize ([error-print-width 4]) 
      (format "~.a" "1234567890")) ; ==> "1..."

My guess is that there is a bug in `"~.a"`.

sorawee:
Yep, there is indeed another bug in `print.c`. Pushing a fix to the PR soon.

sorawee:
Great. I'm glad I added these corner case tests. RacketCS is also buggy... Consider:

```
(error-print-width 6)
(format "~.a" "12345")
```

In 3m, this returns  `"12345"`. In RacketCS, this returns `"123..."`.

sorawee:
@mflatt I don't see an easy way to fix RacketCS. In particular, the function [`p`](https://github.com/racket/racket/blob/master/racket/src/io/print/main.rkt#L167) writes to the object `o` rather than a temporary buffer, so it's difficult to do something like "write maxlength characters, but if there are more than maxlength characters, fix that by changing what already happened: replacing the last 3 characters to `...`". One possible way, I think, is to replace `o` by a temporary buffer in [places like this](https://github.com/racket/racket/blob/master/racket/src/io/print/main.rkt#L82), but I don't know if that's a good solution (e.g., in terms of performance)

sorawee:
Temporary buffer has its own difficulty which is that it needs to be a growable array (we can't just allocate maxlen cells because users might `(error-print-width 10000000)`). In 3m, this [block of code](https://github.com/racket/racket/blob/master/racket/src/racket/src/print.c#L1206) implements the growable array, which really doesn't look pretty.

mflatt:
Merged, including a fix for `"~.a"` in Racket CS.

PR #3005, opened 2020-01-14
Fix typos

bwignall:
Should be non-semantic.

Uses a further-improved version of the tool that found typos in #2982.


pmatos:
@bwignall Thanks for this - it's really useful.

PR #3012, opened 2020-01-20
Attempt to fix contract-not-too-slow macos failures

### LINKED ISSUE #2964: Contract test fails in macos without jit
A contract test is failing without jit on macos:
https://github.com/LinkiTools/racket/runs/346077520#step:10:43

```
FAILED recursive-contract-not-too-slow
       got #f
  expected #t
FAILED recursive-contract-not-too-slow rewrite-to-add-opt/c
       got #f
  expected #t
```

I haven't looked at it but the `slow` in the name might mean this is dependent on the cpu speed. I will take a look at it but I am starting to think we should not have timing in tests and if we do use timing, we should attempt to use cpu time, instead of elapsed time. Again, I haven't looked at this test yet, so it might actually not be time related.

pmatos:
Fixes #2964

Initially this is just adding some debug information. The strange
thing is that this only fails on macos. I don't have one to reproduce
and cannot reproduce it on linux.

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/recursive-contract.rkt
+++ b/pkgs/racket-test/tests/racket/contract/recursive-contract.rkt
@@ -135,12 +135,15 @@
         (define l (build-list n (λ (x) (λ (x) x))))
         (define-values (_ cpu real gc)
           (time-apply (λ () (contract c l 'pos 'neg)) '()))
+        (printf "cpu: ~a, real: ~a, gc: ~a~n" cpu)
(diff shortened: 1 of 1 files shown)


pmatos:
Not ready for review yet - just added some debug information and created PR to be able to run this on macos VMs provided by GitHub.

pmatos:
Fixed with fb63f399ef3c543409f7e88b07ea153650d00615, closing.

PR #3024, opened 2020-01-27
Docs update: bug fix in embedding examples

KDr2:
As the title.


gus-massa:
Thanks! Merged in https://github.com/racket/racket/commit/a01f9ada999188f7f4bb7e5206887ac515268e07

It's too late to include this in the 7.6 version, but the fix will be included in the 7.7 version.

PR #3025, opened 2020-01-28
Delete docs for four obsolete expanded module stx props

greghendershott:
Rewrite docs for a fifth that used to refer to them.

This is just me trying to be helpful based on my (mis?)understanding
of a Slack conversation with @mflatt -- in other words someone should
definitely review this before merging.


mflatt:
Thanks!

PR #3026, opened 2020-01-29
net/http-client: add support of deflate content encoding

szclsya:
This commit allows net/http-client to request `deflate` encoded content and decode the response.

### DIFF
--- a/pkgs/net-doc/net/scribblings/http-client.scrbl
+++ b/pkgs/net-doc/net/scribblings/http-client.scrbl
@@ -98,7 +98,7 @@ configured to @tech{auto-reconnect}.
                           [#:method method (or/c bytes? string? symbol?) #"GET"]
                           [#:close? close? boolean? #f]
                           [#:headers headers (listof (or/c bytes? string?)) empty]
-                          [#:content-decode decodes (listof symbol?) '(gzip)]
+                          [#:content-decode decodes (listof symbol?) '(gzip deflate)]
                           [#:data data (or/c false/c bytes? string? data-procedure/c) #f])
          void?]{
 
@@ -120,7 +120,7 @@ are accepted is automatically added.
 
 If @racket[close?] is @racket[#t] and @racket[headers] does not
 contain a @litchar{Connection} header, then a @litchar{Connection:
-close} header will be added.
+close} header will be added (currently, @racket['gzip] and @racket['deflate] are supported).
 
 This function does not support requests that expect
 @litchar{100 (Continue)} responses.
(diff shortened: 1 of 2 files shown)


mflatt:
This direction seems like a good idea, so thanks! But the patch needs some improvement before it can be merged:

 * `gzip` and `deflate` need to be added to the same `Accept-Encoding` header.

 * The handling of `deflate` should not be a copy-and-paste version of the handling of `gzip`. Those should share the implementation. The copies appear to differ only in `gunzip-through-ports` versus `inflate`.

 * Probably `'deflate` should be added to the default `#:content-decode` argument everywhere. Technically, that's backward-incompatible, but it's probably more helpful than not.

 * The documentation needs to be updated to describe `'deflate`.

szclsya:
Thanks for the advice!
I've pushed some changes, including:
- The decoding process is now in a local function (not sure if this is  the best practice, but it seems to work and looks clean to me).
- The process of adding `Accept-Encoding` has been rewritten so that it can now handle multiple encoding methods. (Not sure if we should add a step to check if the list only contains supported encoding methods)
- `deflate` is now in all default `#:content-decode`.
- The doc has been updated to reflect the new defaults and the supported encoding methods.

mflatt:
Sorry for the delay here. This looks very good, but I think two tweaks are worthwhile:

* Instead of using `local`, we'd normally just use `(let () ....)` to wrap a group of definitions and expressions (and not depend on `racket/local`).

* In the documentation, there should be a history annotation like `@history[#:version "7.6.0.9" @elem{Added support for @racket['deflate] decoding.}]` for each of the functions that now support `'deflate`.

If you're interested in making these changes, that would be great. But I appreciate the work you've done already, and if you'd prefer that I merge and make these small changes, I'd be happy to do that.

Thanks again!

szclsya:
Both sorted! Take a look.

mflatt:
I tried running "pkgs/net-test/tests/net/http-client.rkt", and it looks like extra quotes are added because `~s` is used where `~a` is needed in a format string for `Accept-Encoding`.

Meanwhile, I gave you the wrong `history` declaration. It should be `@history[#:changed "7.6.0.9" @elem{Added support for @racket['deflate] decoding.}]`, with `#:changed` instead of `#:version`. For `http-conn-recv!`, instead of two `history` notes, they should be combined as

```
@history[#:changed "6.1.1.6" @elem{Added the @racket[#:method] argument.}
         #:changed "7.6.0.9" @elem{Added support for @racket['deflate] decoding.}]
```

szclsya:
Ah sorry about the Scribble mistake. I'm not very familiar with this format (yet).
Other than that, it should work fine now.

mflatt:
Merged as a squashed commit 08d33a773dc3a776a13165b71b67ff012493f7fa.

I adjusted the commit to update tests in "pkgs/net-test/tests/net/http-client.rkt", which in turn suggested adding `'deflate` to `http-sendrecv/url` from `net/url` (the older API) and its documentation.

PR #3031, opened 2020-02-02
new bug template

spdegabrielle:
part of interim fix for https://github.com/racket/drracket/issues/114


spdegabrielle:
Please see https://github.com/racket/drracket/issues/114 for motivation for the PR

sorawee:
Provided that `.github/ISSUE_TEMPLATE_DRAFT/bug_report.md` exist already, I don't think that we should create a new one. Instead, we should rename `ISSUE_TEMPLATE_DRAFT` to `ISSUE_TEMPLATE`. Should we close this PR then? You are working on a new PR already anyway.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -0,0 +1,31 @@
...
+3. step 3
+4. etc.
+
+**Expected behavior**
+A clear and concise description of what you expected to happen.
+
+**Screenshots**
+If applicable, add screenshots to help explain your problem.
+
+**Desktop (please complete the following information):**
```
sorawee, line 26: What does "Desktop" mean?

spdegabrielle: @sorawee it asks the submitter to provide the OS& version and the racket version.


--
Desktop (please complete the following information):

OS: [Please include version, etc. e.g. Windows 7, 32 bit, MacOS Catalina 10.15.2, Debian x86-64, Raspbian etc.]
Racket Version [e.g. 7.5]
Additional context Add any other context about the problem here.

sorawee: I understand the below two lines (OS and Racket version), but I don't understand what "Desktop" means. If I use a laptop computer, not a desktop computer, then I don't need to fill in this section?

In any case, this is off-topic. Do you agree that this PR should be closed, and #3130 should be used instead?

spdegabrielle: yes

PR #3034, opened 2020-02-03
Haiku: preliminary patches

extrowerk:


extrowerk:
Build currently fails here:
```
make[9]: Entering directory '/Ports/Testring/racket/racket/src/build/racket'
make cstartup_`./racketcgc -cu /Ports/Testring/racket/racket/src/build/../racket/src/startup-select.rkt`
make[10]: Entering directory '/Ports/Testring/racket/racket/src/build/racket'
echo "Bad startup choice, probably an error running startup-select.rkt"
Bad startup choice, probably an error running startup-select.rkt
exit 1
Makefile:343: recipe for target 'cstartup_' failed
```
Any hint would be greatly appreciated. Thanks!

mflatt:
What output do you get if you run just `./racketcgc` in "<build-dir>/racket"?

Meanwhile, since pthreads are always available, `enable_pthread_by_default=yes` is a good idea.

extrowerk:
>What output do you get if you run just ./racketcgc in "/racket"?
justasec, other computer

>Meanwhile, since pthreads are always available, enable_pthread_by_default=yes is a good idea.
Yep, but somehow something implicitly added -lpthread to  the $LIBS, but pthread functions lives in liroot on Haiku.

mflatt:
I think `use_flag_pthread="no"` will avoid `-pthread`.

extrowerk:
Still the same error, running racketcgc in strace results: http://0x0.st/izl8.txt

extrowerk:
Running the comman manually results:

```
/Ports/Testring/racket/racket/src/build/racket> ./racketcgc -cu /Ports/Testring/racket/racket/src/build/../racket/src/startup-select.rkt
?: free identifier found in linklet
  in: call-with-break-parameterization
Kill Thread
```

extrowerk:
However GC supports Haiku, but the in-tree version is way too old. Would it be possible to update it? Could it be a culprit?

mflatt:
The GC in "racket/src/racket/gc" isn't actually used. (We plan to remove it from the tree, but just haven't gotten to that, yet.) The GC uses of `racketcgc` is "racket/src/racket/sgc". So, I don't yet have a good next suggestion, but it shouldn't be the out-of-date GC.

pmatos:
If haiku is going to be supported, we need a plan for testing to avoid bit rot.

pmatos:
I will prioritise finishing the removal of boehm gc pr, if it helps this.

pmatos:
On testing, what's the usual process to cross-test haiku? Vagrant? Are there any up to date boxes?

extrowerk:
Vagrant should be ok, but i have no experience with it, sorry: https://www.haiku-os.org/blog/kallisti5/2018-11-27_haiku_r1beta1_in_vagrant/

pmatos:
@extrowerk  Sorry for the delay on this. I got haiku on vagrant working. I will take a look at compiling racket with your patch.

racket/src/racket/sconfig.h
```diff
@@ -378,6 +378,33 @@
...
+#  define REGISTER_POOR_MACHINE
+#  define MZ_USE_JIT_I386
+#  define MZ_TRY_EXTFLONUMS
+# elif defined(__amd64__)
+#  define SCHEME_PLATFORM_LIBRARY_SUBPATH "amd64-haiku"
+#  define REGISTER_POOR_MACHINE
+#  define MZ_USE_JIT_X86_64
+#  define MZ_TRY_EXTFLONUMS
+# else
+#  error Unported platform.
```
pmatos: `Unported` sounds quite unusual to me. Did you mean `Unsupported` instead?

pmatos: Take this comment with a pinch of salt - I am not a native speaker so it might be perfectly fine.

pmatos:
@extrowerk I am surprised you don't see this in your build:
```
gcc -Wall  -I./.. -I./../include  -I./../../rktio -I../../rktio -g -O2   -DUSE_SENORA_GC  -c ./string.c -I. -o string.o
./string.c:13:11: fatal error: schsys.h: No such file or directory
 # include "schsys.h"
           ^~~~~~~~~~
compilation terminated.
Makefile:284: recipe for target 'string.o' failed
make[6]: *** [string.o] Error 1
```

I am getting this on a vagrant haiku box. Will investigate.

pmatos:
> @extrowerk I am surprised you don't see this in your build:
> 
> ```
> gcc -Wall  -I./.. -I./../include  -I./../../rktio -I../../rktio -g -O2   -DUSE_SENORA_GC  -c ./string.c -I. -o string.o
> ./string.c:13:11: fatal error: schsys.h: No such file or directory
>  # include "schsys.h"
>            ^~~~~~~~~~
> compilation terminated.
> Makefile:284: recipe for target 'string.o' failed
> make[6]: *** [string.o] Error 1
> ```
> 
> I am getting this on a vagrant haiku box. Will investigate.

Disregard this - I made a rebase of extrowerk:haiku_supporting_patches over master but didn't notice when things didn't go as planned and not all changes were in place.

pmatos:
I can see the segfault in `racketcgc` but can't get `gdb` to work. Something I have explained elsewhere: https://dev.haiku-os.org/ticket/15703

racket/src/racket/sconfig.h
extrowerk: Probably a copy-paste typo... Will fix.

extrowerk:
> I can see the segfault in `racketcgc` but can't get `gdb` to work. Something I have explained elsewhere: https://dev.haiku-os.org/ticket/15703

You probably don't want to use gdb at all on Haiku, as it is pretty old, the Debugger in the Haiku menu is the preferred way to debug programs on Haiku, it have a cli interface too.

pmatos:
> > I can see the segfault in `racketcgc` but can't get `gdb` to work. Something I have explained elsewhere: https://dev.haiku-os.org/ticket/15703
> 
> You probably don't want to use gdb at all on Haiku, as it is pretty old, the Debugger in the Haiku menu is the preferred way to debug programs on Haiku, it have a cli interface too.

I can see that - gdb 6, had to compile with `-gdwarf-2`. Can you point me to the haiku debugger cli interface manual? I can't find it the manual or the binary so I am literally lost here.

extrowerk:
The main GUI debugger called Debugger, which is in /system/apps
User guide and nanual: https://www.haiku-os.org/docs/userguide/en/applications/debugger.html

I think the same binary used for cli debug sessions, but i never used that yet, will check afternoon.

pmatos:
To be honest here I tried to use `Debugger -c` but I couldn't properly understand how to do things like `handle SIGSEGV noprint`, which you need to do so you can properly pass on the signals back to racket and find the exact cause of the problem.

Without extra support on this front, I am not sure I will be able to help debug this issue.

extrowerk:
Sadly i can't really help with that, but i try my best to escalate the problem and try to get an answer for you.
Am i correct that trough Vagrant you have only CLI acess? AFAIK the CLI Debugger is somewhat limited in functionality. Is vagrant absolutely required in this case? Maybe it would be much easier to use virtualbox in this case, and return to vagrant in a later stage.

pmatos:
> Sadly i can't really help with that, but i try my best to escalate the problem and try to get an answer for you.
> Am i correct that trough Vagrant you have only CLI acess? AFAIK the CLI Debugger is somewhat limited in functionality. Is vagrant absolutely required in this case? Maybe it would be much easier to use virtualbox in this case, and return to vagrant in a later stage.

Yes, vagrant is cli only. I can try installing Haiku in virtualbox but I have limited time at the moment and have never installed Haiku before so was trying to avoid that to be honest. However, if that's the only way, let me know and I will have a go.

extrowerk:
Sadly i cant help with Debugger, sorry.

PR #3035, opened 2020-02-03
[CS] regexp-replace: handle #f submatches

### LINKED ISSUE #3032: regexp-replace Racket 7.5 CS
There is a bug in Racket 7.5 CS:

```
> (regexp-replace "(x)?" "" (lambda a ""))
  context...:
   condition->exn
   body of top-level
   dynamic-wind
   eval-one-top12
   /home/pltbuild/build/plt-cs-release/bundle/racket/collects/racket/repl.rkt:11:26

```

However, it works fine in Racket 7.5 Regular:

```
> (regexp-replace "(x)?" "" (lambda a ""))
""
```

97jaz:
A regexp can match while having sub-patterns that are not used in
the match. In regexp-replace, the "insert" argument can refer to
these unused sub-matches. If a function is given for the "insert"
argument, it will receive these unused submatches as #f values.
If a string or byte-string is given, then any reference to an
unused sub-match (by means of a back-reference) is treated as the
empty string or byte-string.

Fixes #3032.


mflatt:
Thanks!

PR #3038, opened 2020-02-07
Add option --enable-werror

pmatos:
This adds an option `--enable-werror` to configure so that `-Werror` is added to `CFLAGS` without overwriting any default flags in place. This `--enable-werror` follows the practice by other free software.

I have also changed the CI workflow to make use of this.

### DIFF
--- a/.github/workflows/ci-push.yml
+++ b/.github/workflows/ci-push.yml
@@ -37,6 +37,7 @@ jobs:
       run: >
         ./configure 
         --prefix=$GITHUB_WORKSPACE/../racketcgc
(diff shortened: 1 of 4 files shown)


pmatos:
> The `configure` and CI change looks good.
> 
> For "jitcommon.c", I wonder whether the right change is to add ` USED_ONLY_FOR_FUTURES` to the end of the declaration (like on line 264).

Thanks for the review. I will take a look.

pmatos:
Didn't know about `USED_ONLY_FOR_FUTURES`. What about now?

PR #3041, opened 2020-02-11
pkg-stage: ~return a list, not a set~ change contract

### LINKED ISSUE #3029: `pkg-stage` broke its own contract
in:

```
#lang racket/base
(require pkg/lib)

(define (table->desc t)
  (pkg-desc (hash-ref t 'source)
            #f 
            (hash-ref t 'name)
            (hash-ref t 'checksum)
            #f))

(pkg-stage (table->desc (get-pkg-details-from-catalogs "cover")))
```

I get the contract violation:

```
pkg-stage: broke its own contract
  promised: list?
  produced: (set '(lib "cover/tests/submods/prog.rkt") '(lib "cover/tests/basic/prog.rkt") '(lib "cover/main.rkt") '(lib "cover/tests/simple-multi/2.rkt") '(lib "cover/tests/exit.rkt") '(lib "cover/tests/basic/raise.rkt") '(lib "cover/tests/error-file.rkt") '...
  in: the range of
      (->*
       (pkg-desc?)
       (#:force-strip?
        boolean?
        #:in-place?
        boolean?
        #:namespace
        namespace?
        #:quiet?
        boolean?
        #:strip
        (or/c #f 'source 'binary 'binary-lib)
        #:use-cache?
        boolean?)
       (values
        string?
        path?
        (or/c #f string?)
        boolean?
        (listof module-path?)))
  contract from: <collects>/pkg/lib.rkt
  blaming: <collects>/pkg/lib.rkt
  at: <collects>/pkg/lib.rkt:224.3
```

bennn:
Change `pkg-stage` to call a different helper function that returns a list. Fixes #3029.

This fix puts lists inside all `install-info` structs.

Another simpler fix is to call `list->set` at the end of the `pkg-stage` function.

### DIFF
--- a/pkgs/racket-doc/pkg/scribblings/lib.scrbl
+++ b/pkgs/racket-doc/pkg/scribblings/lib.scrbl
@@ -186,7 +186,7 @@ is a directory within @racket[path]).
                     [#:in-place? in-place? boolean? #f]
                     [#:namespace namespace namespace? (make-base-namespace)]
                     [#:strip strip (or/c #f 'source 'binary 'binary-lib) #f]
-                    [#:force-strip? force-string? boolean? #f]
+                    [#:force-strip? force-strip? boolean? #f]
                     [#:use-cache? use-cache? boolean? #f]
                     [#:quiet? quiet? boolean? #t])
          (values string? path? (or/c #f string?) boolean? (listof module-path?))]{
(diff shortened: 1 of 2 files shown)


mflatt:
If I'm reading things right, `install-info-module-paths` is used in "install.rkt", in which case the `set->list` approach is probably the way to go.

mflatt:
For that matter, `pkg-stage` is also used internally (e.g., in "archive.rkt"), and those call sites currently expect a set.

Is changing the contract a possibility here?

bennn:
Changing the contract sounds fine --- the current one always errors.

mflatt:
Changing the contract sounds right to me.

PR #3048, opened 2020-02-17
Remove obsolete LGPL license files (can I do this yet?)

kryptine:


samth:
I agree that these need changing, but instead I think they should adopt the same text as in https://github.com/racket/typed-racket/blob/master/LICENSE

PR #3054, opened 2020-02-24
Stake out a more descriptive definition of equality

jackfirth:
Doesn't tackle `eq?` yet, but I'm coming for it next.

Follow up to #3019. Questions, comments, suggestions, torches, and pitchforks are all welcome.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/equality.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/equality.scrbl
@@ -11,16 +11,46 @@ preferred for most use cases.
 
 @defproc[(equal? [v1 any/c] [v2 any/c]) boolean?]{
 
- Two values are @racket[equal?] if and only if they are @racket[eqv?],
- unless otherwise specified for a particular datatype.
+ Determines whether @racket[v1] is ``equal to'' @racket[v2], in a manner that
+ depends on the datatype of @racket[v1] and @racket[v2].
 
- Datatypes with further specification of @racket[equal?] include
- strings, byte strings, pairs, mutable pairs, vectors, boxes, hash
- tables, and inspectable structures. In the last six cases, equality
- is recursively defined; if both @racket[v1] and @racket[v2] contain
- reference cycles, they are equal when the infinite unfoldings of the
- values would be equal. See also @racket[gen:equal+hash] and
- @racket[prop:impersonator-of].
+ Different datatypes have different implementations of @racket[equals?], but the
+ golden rule uniting them is that @deftech{equal means interchangeable} --- if
+ two values are @racket[equal?] then it should be possible to substitute one for
+ the other in a program without breaking the program. Such substitution happens
+ all the time in real world programs, as many Racket functions use the
+ @tech{contract} system to substitute their inputs and outputs with wrappers
+ that enforce invariants. This substitution is usually implemented using
+ @tech{chaperones} and @tech{impersonators}; see @racket[prop:impersonator-of]
+ for details on how this interacts with @racket[equal?].
+
+ In addition to the @tech{equal means interchangeable} rule, all datatypes are
+ expected to abide by the following principles when it comes to equality:
+
+ @itemlist[
+           
+ @item{Equality is @emph{reflexive}, meaning values are always equal to
+   themselves. For any value @racket[_x], @racket[(equal? _x _x)] returns true.}
+  
+ @item{Equality is @emph{symmetric}, meaning the order of the arguments to
+   @racket[equal?] does not matter. For any values @racket[_x] and @racket[_y],
+   @racket[(equal? _x _y)] returns the same answer as @racket[(equal? _y _x)].}
+
+ @item{Equality is @emph{transitive}, meaning that for any three values
+   @racket[_x], @racket[_y], and @racket[_z], if @racket[(equal? _x _y)] returns
+   true and @racket[(equal? _y _z)] returns true then @racket[(equal? _x _z)]
+   returns true.}]
+
+ Most Racket datatypes implement equality by recursively comparing their
+ contents: strings compare each of their characters, lists compare each of their
+ elements, and transparent structures compare each of their fields. This type of
+ behavior is called @deftech{structural equality}, because two values are equal
+ if they have the same structure. Two values containing cyclic data are
+ structurally equal when the infinite unfoldings of the cycles in both values
+ would be equal. When creating new @tech{structure types} with the
+ @racket[struct] form, this behavior is the default if the
+ @racket[#:transparent] option is specified. To implement other forms of
+ equality, see the @racket[gen:equal+hash] @tech{generic interface}.
 
  @(examples
    (equal? 'yes 'yes)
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -11,16 +11,46 @@ preferred for most use cases.
...
- unless otherwise specified for a particular datatype.
-
- Datatypes with further specification of @racket[equal?] include
- strings, byte strings, pairs, mutable pairs, vectors, boxes, hash
- tables, and inspectable structures. In the last six cases, equality
- is recursively defined; if both @racket[v1] and @racket[v2] contain
- reference cycles, they are equal when the infinite unfoldings of the
- values would be equal. See also @racket[gen:equal+hash] and
- @racket[prop:impersonator-of].
+ Determines whether @racket[v1] is ``equal to'' @racket[v2], in a manner that
```
rfindler: I think we can improve upon the use of scare quotes. How about "Determines whether v1 and v2 are the same values, in a specific meaning of the word ``same'', captured by the following general properties" and then give the properties. Then go on to discuss how this is actually determined in a per-datatype way (and have a link to the relevant prop for new ones).

pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -11,16 +11,46 @@ preferred for most use cases.
...
+ depends on the datatype of @racket[v1] and @racket[v2].
+
+ Different datatypes have different implementations of @racket[equal?], but the
+ golden rule uniting them is that @deftech{equal means interchangeable} --- if
```
rfindler: This use of the word "interchangable" is fraught for people with a PL background. Equal? values aren't interchangable in this sense when you get technical.

pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -11,16 +11,46 @@ preferred for most use cases.
...
+ two values are @racket[equal?] then it should be possible to substitute one for
+ the other in a program without breaking the program. Such substitution happens
+ all the time in real world programs, as many Racket functions use the
```
rfindler: The use of substitution and the use of wrappers here is confusing to me.

pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -11,16 +11,46 @@ preferred for most use cases.
...
+ @item{Equality is @emph{symmetric}, meaning the order of the arguments to
+   @racket[equal?] does not matter. For any values @racket[_x] and @racket[_y],
+   @racket[(equal? _x _y)] returns the same answer as @racket[(equal? _y _x)].}
+
+ @item{Equality is @emph{transitive}, meaning that for any three values
+   @racket[_x], @racket[_y], and @racket[_z], if @racket[(equal? _x _y)] returns
+   true and @racket[(equal? _y _z)] returns true then @racket[(equal? _x _z)]
+   returns true.}]
+
+ Most Racket datatypes implement equality by recursively comparing their
```
rfindler: I like this paragraph the best. Fundamentally, I think that equal? should be known as an equality relation (three rules just above), it should be known as "open" (in the sense that you can add definitions to it for your own operations) and it should be thought of as roughly "traversing the two objects recursively, looking for subparts that differ". Those seem to me to be the key points of this relation.

pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -11,16 +11,46 @@ preferred for most use cases.
...
+ contents: strings compare each of their characters, lists compare each of their
+ elements, and transparent structures compare each of their fields. This type of
+ behavior is called @deftech{structural equality}, because two values are equal
```
bennn: This `deftech` is confusing without other kinds of equality to contrast to.

pkgs/racket-doc/scribblings/reference/equality.scrbl
mbutterick: How can this be a “golden rule” if `gen:equal+hash` doesn’t enforce any of these attributes?

sorawee: The problem is that this is a higher-order property, so it's not really enforce-able at the definition site (you would need a verifier for that, and yet it still won't work for all situations. These problems are in general undecidable). We can imagine dynamically checking some properties like symmetry by, for every call of `(equal? a b)`, calling `(equal? b a)` to see if they result in the same thing, but that's also prohibitively expensive.

sorawee: But I think it's accurate to say that "equality must be equivalence relation" is a golden rule. Everything breaks otherwise.

rfindler: I think of the golden rule as one that can be broken, but it isn't right Right Thing to break it. :) 

That said, it does usually refer to a specific rule so maybe another phrase is better?

mbutterick: “Equal means interchangeable” is a curious phrase, because a) it doesn’t illuminate the meaning of “equal” — it just invokes a different word (“interchangeable”) that is no more concrete, and b) if the essential notion of “equal” is “equivalence relation”, fair enough, but there are plenty of equivalence relations that do not produce intuitively “interchangeable” values (for instance `abs`)

sorawee: I think another thing that @rfindler was trying to point out regarding "Equal means interchangeable" is that it's technically false. `(+ 1 1)` is `equal?` to `2`, but they are not interchangeable in some contexts. E.g., `(quote (+ 1 1))` and `(quote 2)`.

rfindler: I was thinking of something slightly different. I think that the quote thing isn't a big issue as you can just rule out contexts inside quotes. But if you have somewhere `(define x (list 1 2))` even if that exact `x` is in scope, you cannot replace it with `(list 1 2)` because these can be distinguished (using `eq?`). But more importantly, I don't think that this kind of thing is really what is important about `equal?` anyway (I put what I think is important about it in a different comment around here somewhere...)

samth:
What's the current status of this PR?

jackfirth:
I still want to pursue this, but lately I haven't been spending much time
on Racket. I'm not sure when that trend will change but once it does I plan
to pick this back up.

pmatos:
There's an approved review from @mflatt but no merge. What are the actions needed to merge this or close the PR? Maybe if there's still a lot of work to be done, we can close the PR and @jackfirth can reopen upon his return? 

This is to avoid many open PRs that are essentially dead-ends.

rfindler:
It looks to me like there are several small comments that aren't responded to but that this is close to being ready to merge? (I'm not sure I'm reading things correctly, however.)

jackfirth:
@rfindler is correct. There's several small adjustments to make, but not any significant content additions or deletions. I think the adjustments are important and this PR shouldn't be merged without them, so I'm going to close this PR until I come back to it. I've been catching up on racket today so that might be soon, but I'm far too inconsistent of a person to make any promises.

PR #3055, opened 2020-02-25
contract: use impersonator prop. for impersonator-unconstrained-domain->/c

bennn:
rename the contract property from a chaperone prop to an impersonator
prop (looks like this was an old copy/paste error)


rfindler:
This looks good to me, but needs a test case. Thanks!

rfindler:
(specifically, we need something that checks that `chaperone-contract?` returns `#f` on one of those.)

PR #3057, opened 2020-02-25
Make get-pure-port redirection case-insensitive

jamesbornholt:
GitHub's CDN seems to have recently started returning the `Location` header for a redirect with a lowercase `l`, which breaks the redirect logic. The HTTP spec says that header names are case-insensitive, so we need to look for either version.

### DIFF
--- a/racket/collects/net/url.rkt
+++ b/racket/collects/net/url.rkt
@@ -320,7 +320,7 @@
 
     (define new-url
       (ormap (λ (h)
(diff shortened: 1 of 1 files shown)


sorawee:
But it would fail `LOCATION`, right? Why don't we just use the case-insensitive flag?

jamesbornholt:
Yes, good point, done. (If I ever write a HTTP server it will now be sending `lOcAtIoN:`)

PR #3060, opened 2020-02-27
style: consistent figure style

bennn:
Use a consistent style for the "good / bad" examples:

- start the code at the far left of each box
- keep a little space between the top of the box & the label

- - -

### Before, at the top of the page:
parenthesis-before

### After
parenthesis-after

### Before, at the bottom:
space-before

### After
space-after

### DIFF
--- a/pkgs/racket-doc/scribblings/style/textual.scrbl
+++ b/pkgs/racket-doc/scribblings/style/textual.scrbl
@@ -30,21 +30,24 @@ your code.
  (define (conversion f)
    (* 5/9 (- f 32)))
 ]
- @filebox[@tt{really bad}
- @codeblock{#lang racket
+ @racketmod[#:file
+ @tt{really bad}
+ racket
+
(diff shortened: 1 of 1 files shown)


jbclements:
LGTM. I was really unsure of which example to copy in adding my text.

sorawee:
@MichaelMMacLeod raised a very good point in Slack that `racketmod` seems to fix the style, making the "really bad" example looks just like the "good" example.

bennn:
I reverted the changes to that first example
b804936db01efb5e34165a9295a8201fca137b24

Couldn't find a nice way to get the right example & the right alignment.

MichaelMMacLeod:
Also, shouldn't this be the correct formatting of parentheses in the second example?
image
Putting them on the same line defeats the purpose of that section.

bennn:
yes indeed
567fd9e86613b84297cdbf277a738a1d278436fb

PR #3061, opened 2020-02-29
doc: contract, clarifications

bennn:
- reword the first-page explanation of chaperone contracts;
  try to give a positive description by talking about wrappers first
  (rather than starting with what chaperones maybe don't do)

- name the recognizer functions for chaperone & impersonator contracts
  (on the first page)

- clarify that `contract-projection` and `contract-val-first-projection`
  are bad --- that there's a preferred alternative

- describe the outputs of `contract-projection` and
  `contract-val-first-projection`; their docs were identical before, but
  now re-use prose from sec 8.7 (Building New Contract Combinators)


bennn:
Approved by Robby & Matthias over email.

(We all agree that the first page could be even better, but don't have ideas at the moment)

rfindler:
Thanks!

PR #3066, opened 2020-03-04
Doc: more illustrative example for the filter-map function

sorawee:
The current example could as well be written using `filter`. This patch makes the example complicated a little bit to illustrate how `filter-map` could be used (differently than `filter`).


gus-massa:
I like your example better, but I'd change the `-2` to `-4` too. Also, the title of the commit should include `filter-map`, something like "_add more illustrative example to filter-map docs_".

PR #2678, opened 2019-06-04
Syntax parse improvement (doc and syntax class library)

sorawee:
This PR:

- Change the `function-header` syntax class to support the attribute `name`. Add accompanying tests.
- Switch the internal implementation from experimental forms to non-experimental forms.
- Add syntax class library examples.
- Define a form for documenting syntax class attributes.
- Fix various typos.

PR #3067, opened 2020-03-04
Preserve undocumented `args` syntax class attribute.

samth:
PR #2678 unintentionally removed this attribute, but it was used
at least by "collections-lib" and "static-rename-lib".

cc @sorawee @lexi-lambda @jackfirth @rmculpepper

### DIFF
--- a/racket/collects/syntax/parse/lib/function-header.rkt
+++ b/racket/collects/syntax/parse/lib/function-header.rkt
@@ -6,7 +6,7 @@
 (provide function-header formal formals)
 
 (define-syntax-class function-header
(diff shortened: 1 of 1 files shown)


mflatt:
This does resolve the problems I had building those packages, and all packages in my installation now build.

sorawee:
Sorry!

lexi-lambda:
Thank you!

Those are both my packages, and arguably I am at fault for depending on something undocumented. I am also willing to update my packages to not depend on the undocumented attribute, or perhaps the attribute ought to be documented after all.

PR #3068, opened 2020-03-05
Add generalized common-prefix procedures

trouvant:
Adds `split-common-prefix*`, `take-common-prefix*`, and `drop-common-prefix*`, which behave like the unstarred procedures but operate on an arbitrarily-sized list of lists, rather than only two lists. Where `split-common-prefix` and `drop-common-prefix` return the tails of the two argument lists, the corresponding new procedures return a list containing the tails of the lists in the passed list.

This update generalizes the behaviours of the original common-prefix procedures but leaves them untouched, retaining backwards-compatibility.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/pairs.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/pairs.scrbl
@@ -1113,6 +1113,24 @@ combined with the from-right functionality of @racket[take-right],
 @history[#:added "6.3"]{}
 }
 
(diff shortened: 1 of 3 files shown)


bennn:
What about changing the original functions to take >=2 arguments?

samth:
I mostly agree with @bennn, although it might be inconvenient to deal with dynamic numbers of values for the results. At a minimum, though, it seems like the existing implementations could re-use this code.

PR #3075, opened 2020-03-20
Fix #3071: quote and whitespace bugs in `syntax->string`

### LINKED ISSUE #3071: Quote-related issues in syntax->string
1. `(syntax->string #'(quote . 0))` errors because the code assumes that if the `car` of a pair is `quote`, then the pair must be a list.
2. `(syntax->string #'(,hello))` results in `"(unquotehello)"`. Same problem occurs for quasiquote, unquote-splicing,  syntax, quasisyntax, unsyntax, unsyntax-splicing.

lkh01:
This PR solves the issues raised in #3071. 

Mainly, `syntax->string` does not work properly with `quote` (instead of `'`), and does not work at all with unquote, quasiquote, etc.
There is also an issue with preserving whitespace that comes after the last term in a syntax object.

So far, I have resolved only the `quote` and `'` issue. Unquote and others should be quite similar. Whitespace seems to be a bit trickier at first glance.


racket/collects/syntax/to-string.rkt
```diff
@@ -50,11 +50,12 @@
                                    (printf "; ")))
                            l))]
               [(and (pair? (syntax-e c))
-                    (eq? (syntax-e (car (syntax-e c))) 'quote))
+                    (eq? (syntax-e (car (syntax-e c))) 'quote)
+                    (eq? (syntax-span (car (syntax-e c))) 1))
```
sorawee: Could you add a comment that when the above three conditions are met, it must be the case that it's the shorthand form, which always constructs a syntax list, justifying the use of `syntax->list` below.

samth:
I think this needs some tests to show the new behavior and distinguish from the old behavior.

sorawee:
I think it might make more sense to create the following hash:

```
#hasheq((quote . "'")
        (quasiquote . "`")
        (unquote . ",")
        (unquote-splicing . ",@")
        (syntax . "#'")
        (quasisyntax . "#'")
        (unsyntax . "#,")
        (unsyntax-splicing . "#,@"))
```

And look up this hash instead of adding more `cond` clauses.

lkh01:
I guess this is a style choice, but I wasn't sure if I should write out all the tests explicitly (as I did) or shorten them as well by using a hash table.

samth:
Shortening the tests with a hash table is fine, but what you wrote is also fine.

PR #3082, opened 2020-03-25
Explicit zero assign to avoid warning

pmatos:
GCC 9.3 things that `h2` might reach line 385 uninitialized.
```
./hash.c:385:14: warning: 'h2' may be used uninitialized in this function [-Wmaybe-uninitialized]
  385 |       h = (h + h2) & mask;
      |           ~~~^~~~~
```


pmatos:
The code is too dense for me to follow atm, however it's certainly safe to zero-assign here. There are some zero-assignments in the first few if branches that I was tempted to remove but will keep in order not to create possible problems given I have not read the code carefully enough.

mflatt:
LGTM

Issue #3050, opened 2020-02-19
Illegal Instruction on freescale i.mx53 (armv7l)


pmatos:
Upstream issue reference for: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=948789
Mailing list information: https://groups.google.com/d/msg/racket-dev/o631-azv-5g/r9Ct1H_KDAAJ

Cc: @bremner

bremner:
Paulo Matos <notifications@github.com> writes:

> Upstream issue reference for: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=948789
> Mailing list information: https://groups.google.com/d/msg/racket-dev/o631-azv-5g/r9Ct1H_KDAAJ

I'm not sure if this is a surprise or not, but it looks like the 7.6 is
breaking in the same place:

https://buildd.debian.org/status/fetch.php?pkg=racket&arch=armhf&ver=7.6%2Bdfsg1-1&stamp=1584149113&raw=0

bremner:
David Bremner <david@tethera.net> writes:

> Paulo Matos <notifications@github.com> writes:
>
>> Upstream issue reference for: https://bugs.debian.org/cgi-bin/bugreport.cgi?bug=948789
>> Mailing list information: https://groups.google.com/d/msg/racket-dev/o631-azv-5g/r9Ct1H_KDAAJ
>
> I'm not sure if this is a surprise or not, but it looks like the 7.6 is
> breaking in the same place:
>
> https://buildd.debian.org/status/fetch.php?pkg=racket&arch=armhf&ver=7.6%2Bdfsg1-1&stamp=1584149113&raw=0

Paulo earlier asked if this was still getting an illegal
instruction. It's a bit surprising (to me), but running in GDB I get
SIGSEGV, I would have thought SIGILL. In any case dissassembly shows an
"stmdb" instruction, as before.

d

pmatos:
I can't repro locally. Would you be able to restore my access to debian dev machines?

bremner:
Paulo Matos <notifications@github.com> writes:

> I can't repro locally. Would you be able to restore my access to debian dev machines? 
>

I sent a message about this earlier today to you personally, but I guess
something ate it:

              As for machine access, I suspect we need to restart the process here,
              which involves you sending me some GPG signed text, and me signing and
              forwarding;

                      https://dsa.debian.org/doc/guest-account/

pmatos:
[...]

PR #3083, opened 2020-03-25
Change return type of rktio_get_milliseconds to be unsigned

pmatos:
This fixes a major issue in arm 32bits, detected by ubsan, where
conversion of type to milliseconds results in overflow.

```
rktio_time.c:92:21: runtime error: signed integer overflow: 1584975753 * 1000 cannot be represented in type 'long int'
```

### DIFF
--- a/racket/src/racket/include/mzwin.def
+++ b/racket/src/racket/include/mzwin.def
@@ -105,6 +105,7 @@ EXPORTS
  scheme_warning
  scheme_raise
  scheme_log_level_p
+ scheme_log_level_topic_p
  scheme_log
  scheme_log_w_data
  scheme_log_message
--- a/racket/src/racket/include/mzwin3m.def
+++ b/racket/src/racket/include/mzwin3m.def
@@ -105,6 +105,7 @@ EXPORTS
  scheme_warning
  scheme_raise
  scheme_log_level_p
+ scheme_log_level_topic_p
  scheme_log
  scheme_log_w_data
  scheme_log_message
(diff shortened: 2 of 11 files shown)


pmatos:
On arm32bits we have an integer overflow. I have done a very small change but we could if we wanted improve the state of the code in general. I think there are other undetected cases going on since `rktio_timestamp_t` is an `intptr_t` and it would make sense to make it a `uintptr_t`. If this is accepted, I will look into it after looking at #3050, which is how I found this.

Unfortunately this does not help in fixing #3050.

pmatos:
> It would have been better to use unsigned in the first place, but I'm not clear how much trouble it will cause to change interfaces, so that will take more time to investigate. For now, I recommend not changing the return type, but changing the inside of the function to use unsigned arithmetic. Similarly, "fun.c" line 9902 could temporarily switch to unsigned for the arithmetic.
> 

Even if I use unsigned arithmetic to perform the calculation, the result is still too large to be expressed as a 32bit signed integer, at least that's my understanding and why we don't see the problem on a 64bit machine.

My tests showed no problems in changing the return type to unsigned though.

mflatt:
It looks like changing to `uinptr_t` won't cause a problem, since no C declarations are derived from the "rktio.h" declarations (e.g., nothing exposed by "schemex.h" or similar generated headers). Anything expecting an `intptr_t` will just get a coercion, of course. So, I'm not convinced that changing the interface is ok.

But I'll take the opportunity to clean up things that received the result into an `intptr_t` and performed further arithmetic. I'll merge your commit along with those changes.

mflatt:
Merged as 9c2b84d84965bc8a3f519f2cd4104df26e855e0f

pmatos:
@mflatt thanks for sorting this one out.

PR #3084, opened 2020-03-28
Use new URL for GitHub tarballs.

samth:
GitHub switched URLs for tarballs, redirecting to the new one with
a 302. However, old versions of Racket don't follow redirects in
`raco pkg install`, so they broke (before 6.3). Using the new URL
should work for everyone.

Reported by @greghendershott.

### DIFF
--- a/racket/collects/pkg/private/stage.rkt
+++ b/racket/collects/pkg/private/stage.rkt
@@ -340,7 +340,7 @@
           (define new-url
             (url "https" #f "github.com" #f #t
                  (map (λ (x) (path/param x empty))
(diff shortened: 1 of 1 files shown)


mflatt:
This seems ok — but it doesn't help for old Racket versions, since it only updates the latest client, right?

Also, I think this is a download mode that isn't typically used by the current Racket client. If I'm not confused, it's only used when the `PLT_USE_GITHUB_API` environment variable is set.

samth:
Yes, that's right, and after more investigation I don't think the problem for pre-6.3 can be fixed.

PR #3090, opened 2020-04-02
Static Analysis CI workflow using LLVM scan-build

pmatos:
This brings static analysis using LLVM scan-build to racket github Actions. Another step towards replacing the jobs we have running in Gitlab.


pmatos:
> This looks good. A few thoughts:
> 
>     1. could this be a separate file?
>

I wanted that as well but scabuild-3m and scanbuild-cs rely on build-racketcgc. Since the `needs` keyword requires jobs in the same workflow, this needs to be in the same file. However, I have a way to split the file which I will open a PR for tomorrow.
 
>     2. can you point to some sample output?

Yes.
https://github.com/pmatos/racket/actions/runs/68966656
For example:
https://github.com/pmatos/racket/suites/566969371/artifacts/3770953

PR #3091, opened 2020-04-02
Free buffer in error case.

samth:
Found by scan-build.
cc @pmatos.

### DIFF
--- a/racket/src/rktio/rktio_fs.c
+++ b/racket/src/rktio/rktio_fs.c
@@ -1949,8 +1949,10 @@ char *rktio_system_path(rktio_t *rktio, int which)
 #else      
       return append_paths(home, "racket-prefs.rktd", 1, 0);
 #endif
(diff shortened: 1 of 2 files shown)


racket/src/rktio/rktio_fs.c
```diff
@@ -633,6 +633,7 @@ char *rktio_get_current_directory(rktio_t *rktio)
    free(s);
    get_posix_error();
  }
+ free(s);
```
mflatt, line 636: Did you mean for this `free` to be here?

samth: Aha, that explains why this doesn't work. Thanks!

That free was about another coverity report, but that one is a false positive because getcwd is terrible.

samth:
This is finally what I meant to do.

mflatt:
LGTM

PR #3101, opened 2020-04-14
Perf integrate

samth:
This enables the Linux `perf` tool to know function names generated by the Racket BC jit.


pmatos:
This is really interesting - I recently read about this from @wingo here:
https://www.mail-archive.com/guile-devel@gnu.org/msg15374.html

Let me know if I can offer a hand and thanks for working on this.

samth:
It needs three things. (1) To fix whatever is causing it to segfault currently. (2) To enable it with a build-time or run-time option instead of `if (1)`. (3) To make it work on Racket CS.

samth:
See also https://chromium.googlesource.com/v8/v8/+/refs/heads/master/src/diagnostics/perf-jit.cc

pmatos:
I was quickly looking at the patch as I find this interesting and increasingly important for Racket but I would definitely start with Racket CS. It seems CS does not use `codetab.inc` so a completely different approach is necessary, although I haven't seen where the changes are necessary.

samth:
Implementing this for Racket CS shouldn't be that hard -- the key thing is to figure out where the machine code gets written, and add something to write to the relevant file. You also need to be able to get the name of the function being compiled, but that's probably pretty easy.

samth:
More information about formats and integration with perf:
https://bugzilla.mozilla.org/show_bug.cgi?id=1759173
https://github.com/torvalds/linux/blob/816e51dfb5ba47e4f989af656c956a8c0cc686c0/tools/perf/Documentation/jitdump-specification.txt

PR #3103, opened 2020-04-15
Switch /replacement and /our-replacement names

MarcKaufmann:
The call to `random-sample/replacement` gets called when `(not
replacement?)` holds, while `random-sample/out-replacement` gets called
otherwise.

This is probably because the algorithm for the without replacement case
actually replaces parts of the sample, but that is different from the
meaning of a sample with and without replacement, which means with
possible duplicates or without.

Important note: I don't know how to run tests on my local fork, so I didn't test this.

### DIFF
--- a/racket/collects/racket/random.rkt
+++ b/racket/collects/racket/random.rkt
@@ -21,7 +21,7 @@
 
 (define (random-ref seq [prng (current-pseudo-random-generator)])
   (define samples
-    (random-sample/out-replacement seq 1 prng))
+    (random-sample/replacement seq 1 prng))
   (unless samples
     (raise-argument-error 'random-ref "non-empty sequence" seq))
   (vector-ref samples 0))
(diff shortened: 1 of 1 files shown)


samth:
Just checking -- these are just internal names, so this shouldn't have any user-visible impact, right?

MarcKaufmann:
Yes, but I am in the process of reporting a bug, in the process of which I noticed this as I thought this was the bug.

MarcKaufmann:
I thought that I had synchronized my fork before creating the PR but apparently didn't, hence all those commits. I then rebased my branch that I used for the PR, and now it's a mess. How do I fix this (and obviously sorry for the mess)?

samth:
You probably need to rebase on top of `racket/master` (or whatever the remote is called for the racket/racket repository) and then force-push to the appropriate branch of your fork.

MarcKaufmann:
OK, I *think* I fixed it. I used `git rebase fix-random-sample-naming` while on branch master, having gotten myself confused between the branch-to-rebase and the branch-rebasing-onto...

pmatos:
ping! Can we look at this again so it is not forgotten? There's an approved review by @bennn but no merge.

bennn:
I was planning to wait until after the 7.7 release

samth:
The branch has already happened.

PR #3106, opened 2020-04-17
Constrain the values of a and b in function not to be both NULL

### LINKED ISSUE #2686: Possible issue with splice_ptr_array
I might be just misunderstanding the function, but please bear with me. I looked at it because scan-build shows a possible null dereference.
2019-06-06-211802_785x873_scrot

However, one of the things that scan-build is not aware of is that `al = length of a` and `bl = length of b`. So I think that the complaint is bogus. However, what might not be bogus is that if the assumptions are correct, the loop in the else branch is unnecessary because we will only enter it if `!b` in which case `bl == 0`.  What am I missing here?

pmatos:
Fixes #2686

### DIFF
--- a/racket/src/racket/src/thread.c
+++ b/racket/src/racket/src/thread.c
@@ -26,6 +26,8 @@
 # include <malloc.h>
 #endif
 
+#include <assert.h>
+
 #ifndef SIGNMZTHREAD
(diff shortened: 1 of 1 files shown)


pmatos:
How did I miss MZ_ASSERT?  Sorry. Will fix and merge.

PR #3107, opened 2020-04-17
Minor documentation fix to build.md

khardy:
Minor documentation fix.


mflatt:
The source for this document is "pkgs/racket-build-guide/build.scrbl", so we should update there.

PR #3114, opened 2020-04-21
Escape "~" in printed record names.

### LINKED ISSUE #3113: [cs] `dump-memory-stats` crashes with contracted keyword-argument functions
Summary: `dump-memory-stats` crashes when there's a function `define`d in the same module that has:
1. a mandatory keyword argument
2. a contract attached to it (e.g. by `contract` or `define/contract`)

Specifically, it crashes with a format error: `fprintf: incomplete format directive "~-"`.

It does not happen when the function is defined in another module and provided with `contract-out`. 
It also does not happen without the contract.

Example to reproduce (tested on `v7.7.0.2 [cs]`):
```racket
#lang racket

(define foo
  (contract (-> #:x any/c any/c)
            (λ (#:x x) 0)
            'pos
            'neg))

(dump-memory-stats)
```
Output:
```
Begin Dump                                                                                        
Current memory use: 92757968                                                                      
Begin RacketCS                                                                                    
 locked                         0            0  |          1          352                         
 fxvector                       0            0  |          3        8,240
 ... many more lines ...
 #<will-executor>               1           48  |          9          432
 #<procedure-chaperone>         1           48  |          1           48
fprintf: incomplete format directive "~-"
  context...:
   condition->exn
   do-raise
   dynamic-wind
   go
   dump-memory-stats
   call-with-values
   call-in-empty-metacontinuation-frame
   body of "/tmp/test.rkt"
   for-loop
   run-module-instance!
   perform-require!
   namespace-require+
   #%for-each
   call-in-empty-metacontinuation-frame
   thunk_10
   call-in-empty-metacontinuation-frame

### PR DESCRIPTION
**Pull request by samth**
`~-` is a Chez printf format specifier, and having these names makes
it more likely that bugs like #3113 will appear.

Closes #3113.

### DIFF
--- a/pkgs/racket-test/tests/racket/dump-tilde-name.rkt
+++ b/pkgs/racket-test/tests/racket/dump-tilde-name.rkt
@@ -1,0 +1,20 @@
+#lang racket/base
+
+(require racket/contract)
+
+;; This test checks that names of records using ~ can be dumped successfully
+;; The contracted procedure is here because the rumble implementation has a
+;; ~ in the name.
+
+(define foo
+  (contract (-> #:x any/c any/c)
+            (λ (#:x x) 0)
+            'pos
+            'neg))
+(struct ~s ( x))
+
+(define v (~s 1))
+
+(dump-memory-stats)
+
+v
--- a/racket/src/cs/chezpart.sls
+++ b/racket/src/cs/chezpart.sls
@@ -22,7 +22,7 @@
                            with-input-from-file with-output-to-file
                            call-with-output-file
                            file-position
-                           write display newline port-name port-closed? write-char
+                           write newline port-name port-closed? write-char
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

**Comment by samth** [comment]
This is not sufficient -- just creating a struct with a name like `~s` has this problem too. I think somehow `fprintf` needs to be avoided entirely for the type names, but maybe @mflatt has a different suggestion. I'll see what I can make work.

**Comment by samth** [comment]
This now works, and there's a test.

**Comment by mflatt** [comment]
Classic bug. But isn't the solution to either switch to `display` or add the missing `"~a"` in each
```
                    (chez:fprintf (current-error-port)
                                  (layout-line ....))
```
?

samth:
Ah yes, that's much better.

samth:
Now ready.

Issue #3112, opened 2020-04-21
Commit 734491 possibly caused CI breakages


pmatos:
It looks like that https://github.com/racket/racket/commit/734491b511e2fa94a5e12074a804ba4dab8230ef might have caused some breakages in `scan-build` and test jobs on `ubuntu-18.04`.

* `scan-build` jobs for 3m and cs fail with 
```
make[3]: Entering directory '/__w/racket/racket/racket/src/racket'
/__w/racket/racket/../racketcgc/bin/racket -cu ./src/cify-check.rkt cstartup.inc
standard-module-name-resolver: collection not found
make cstartup.inc
  for module path: (lib "setup")
make[4]: Entering directory '/__w/racket/racket/racket/src/racket'
  collection: "setup"
  in collection directories:
   /github/home/.racket/7.7.0.2/collects
```

  * https://github.com/racket/racket/runs/597631077#step:8:205
  * https://github.com/racket/racket/runs/597631127#step:9:24

* `test` jobs on ubuntu-18.04 are now failing with:
```
Section(sandbox)
ERROR: resolve-path: `exists' access denied for /home/runner/work/racket/racket
```

For example: https://github.com/racket/racket/runs/597631099#step:9:79

mflatt:
I haven't investigated much, but here's an idea of what might be happening.

The sandbox may be trying to resolve a path that starts "/home/runner/work/racket/racket/../racketcgc". It may have permission to look in "/home/runner/work/racket/racketcgc", but not in "/home/runner/work/racket", so it can't simplify the path with ".." to the path without "..". And this problem wouldn't have happened before the repair commit, because the installation was incorrectly in-place instead of unix-style, so the original path with ".." didn't need to be recorded anywhere.

If that's right, some ideas for a repair:
 * change the CI build to be in-place instead of unix-style by supplying `--enable-origtree`
 * change the CI build to avoid ".." in the supplied prefix path
 * change the sandbox to simplify some path earlier, while it still has permission
 * change the build or installation to simplify paths (but that seems likely to run afoul of intended use and update of soft links)

pmatos:
@mflatt thanks for the investigation. I haven't really looked much into it yet besides filing the bug but at the same time I notice that there's a PR from @samth to fix this so I am happy! :) I will take a look at Sam's PR.

pmatos:
[...]

PR #3115, opened 2020-04-21
Avoid .. in --prefix.

samth:
cc @pmatos. Relevant to #3112.

### DIFF
--- a/.github/workflows/ci-push.yml
+++ b/.github/workflows/ci-push.yml
@@ -18,6 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       RACKET_EXTRA_CONFIGURE_ARGS: ""
+      PREFIX_DIR: "$HOME/work/racket"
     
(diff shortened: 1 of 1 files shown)


.github/workflows/ci-push.yml
```diff
@@ -18,6 +18,7 @@ jobs:
     runs-on: ${{ matrix.os }}
     env:
       RACKET_EXTRA_CONFIGURE_ARGS: ""
+      PREFIX_DIR: "$HOME/work/racket"
```
pmatos, line 21: This might break if/when github changes `$GITHUB_WORKSPACE` to be something else. We should use something based on that, I think.

pmatos: I see this is the change you did throughout the file. Maybe we can use `` `realpath $GITHUB_WORKSPACE/../racketcgc` `` so that it's `sh` instead to simplify the path.

PR #3117, opened 2020-04-22
Fix CI yml syntax

pmatos:
Do not use working directory on uses.
Also - update actions download/upload artifact to use v1 instead of master.


pmatos:
Required to fix current YML CI file - merging.

PR #3118, opened 2020-04-22
Check that file specified by --enable-racket exists

pmatos:


pmatos:
> Is using `!` before `test` portable? Ok if so, otherwise I think it needs to go between `test` and `-f`.

Good point! It works but probably not as portable as `test ! ...`. Changed. Thanks for the comment! :)
Fixing and merging.

PR #3119, opened 2020-04-23
Fix RacketCGC file location for scanbuild jobs

pmatos:


pmatos:
At the moment, with this PR, I am forcing a run to show me the files in `runner.temp` in order to understand why scanbuild jobs are failing.

pmatos:
In a moment of desperation I listed all the files under root `/`. It was illuminating after I downloaded the raw logs:
https://pipelines.actions.githubusercontent.com/e1WEYesiYcyYj2hbddKO3MdIECXrwDUp0z57oQhIrpYt5uUjZZ/_apis/pipelines/1/runs/529/signedlogcontent/82?urlExpires=2020-04-24T07%3A02%3A18.9304129Z&urlSigningMethod=HMACV1&urlSignature=2QBC3rgNS6zlLlRV1U4%2FuRubA1rthxxpV6aEjkVCDEU%3D

Turns out it seems that the files are still there, just under a different location. The reason this causes no problems elsewhere is because the `scanbuild` jobs are the only ones that currently run inside a docker container. `scanbuild-cgc` works because it doesn't use the `download-artifact`. I have a feeling this action doesn't play well with containers.

If you look at the raw log all of the files are under `__w` and the home folder `/home` is empty. Docker container is created with:
```
2020-04-24T06:56:34.3614006Z ##[command]/usr/bin/docker create --name 04819be4070c47b2ac36e3332032673a_pmatosscanbuildlatest_15531c --label c27d31 --workdir /__w/racket/racket --network github_network_85bcdc42dafd4b71bfd4c739ba176ef0  -e "HOME=/github/home" -e GITHUB_ACTIONS=true -e CI=true -v "/var/run/docker.sock":"/var/run/docker.sock" -v "/home/runner/work":"/__w" -v "/home/runner/runners/2.169.1/externals":"/__e":ro -v "/home/runner/work/_temp":"/__w/_temp" -v "/home/runner/work/_actions":"/__w/_actions" -v "/opt/hostedtoolcache":"/__t" -v "/home/runner/work/_temp/_github_home":"/github/home" -v "/home/runner/work/_temp/_github_workflow":"/github/workflow" --entrypoint "tail" pmatos/scan-build:latest "-f" "/dev/null"
```

I will need to decipher that and see if there's anything that could be conflicting with our image. Also need to check how things actually start in the right place and apparently after the `download-artifact` they go somewhere else.

pmatos:
OK - confirmed by a GitHub Partner the problem is on their side: https://github.community/t5/GitHub-Actions/runner-temp-being-emptied-between-jobs/m-p/55433/highlight/true#M9555

Given his explanation I will attempt a workaround until I am contacted by their engineers with a fix.

pmatos:
The reproduced issue here: https://github.com/pmatos/actions-test/runs/614871639?check_suite_focus=true

pmatos:
After hardcoding runner temp, we have now another issue:
```
/__w/_temp/racketcgc/bin/racket  -O "info@compiler/cm" -l- setup --chain ./../setup-go.rkt ../compiled cstartup.inc cstartup.d ./src/cify-startup.rkt ./src/startup.inc ./src/schvers.h
standard-module-name-resolver: collection not found
  for module path: (lib "setup")
  collection: "setup"
  in collection directories:
   /github/home/.racket/7.7.0.3/collects
   /home/runner/work/_temp/racketcgc/share/racket/collects
```

It's likely this is related to possibly `$HOME` not having the correct value. Unsure here what could be going on, but it's certainly a blocker to get this working again.

PR #3120, opened 2020-04-23
Minor typo, use #t for consistency

sorawee:


samth:
I'm not sure about removing `#true` -- is it inconsistent with the rest of that page in particular?

sorawee:
Yes, the rest of that page uses `#t`.

PR #3124, opened 2020-04-23
Remove scanbuild jobs for 3m and cs

pmatos:
This is temporary until I can fix them. Unfortunately, I have no idea
what's going on on the Actions side. In the meantime I would like to
see green ticks for master commits.

Working on a fix under #3119


pmatos:
This just removes the jobs that are broken so we work on a clean sheet. Merging without review.

PR #3125, opened 2020-04-23
Closes #2464: deprecate net/url

sorawee:
This PR deprecates `net/url` as discussed in #2464.

CC: @DarrenN, @jeapostrophe, @greghendershott

### DIFF
--- a/pkgs/net-doc/net/scribblings/url.scrbl
+++ b/pkgs/net-doc/net/scribblings/url.scrbl
@@ -8,7 +8,12 @@
                      (only-in net/url-connect current-https-protocol)
                      openssl))
 
-@title[#:tag "url"]{URLs and HTTP}
+@title[#:tag "url"]{URLs and HTTP: Legacy HTTP Client}
+
+@deprecated[@racketmodname[net/http-client]]{
+  The @racketmodname[net/http-client] module supports a modern HTTP protocol
+  and provides functionalities such as the support of the @tt{PATCH} request method.
+}
 
 @defmodule[net/url]{The @racketmodname[net/url] library provides
 utilities to parse and manipulate URIs, as specified in RFC 2396


mflatt:
I think a good change will require more than this deprecation label. For one thing, this section of the documentation also houses `net/url-structs` (in a subsection), which is not covered by `net/http-client`, so the deprecation claim is misleading. For another, `net/http-client` doesn't cover all of the functionality of `net/url`, such as handling redirections.

I don't know how much an improvement will involve reorganizing documentation versus adding more functionality to `net/http-client` versus introducing a new module that has just the good parts of `net/url` not covered by `net/http-client` and `net/url-structs` — but it will be some combination of that.

sorawee:
Agreed. Let me close this PR then since the actual change needs to be a lot more involved than this.

samth:
My feeling is similar to @mflatt. Just fetching data from a url is still easier with `net/url`. My sense is the right thing is a library that's _easier_ than `net/url`, supports all the features from `net/http-client`, and resembles high-quality HTTP request libraries in other languages, such as Python's `request`, Rust's `reqwest`, or Clojure's `clj-http`.

sorawee:
@Bogdanp does `http-easy` support everything that `net/url` / `net/http-client` does? If so, I think we should revive this PR, but point toward `http-easy` instead.

Bogdanp:
@sorawee While `http-easy` implements a superset of the HTTP-related functionality in both modules, it doesn't reprovide any of the URL-specific functionality from `net/url-{structs,string,unit,sig}` (which are useful even outside of an HTTP context (eg. `file`, `git`, `ftp`, and so on, URLs)). So, I don't think it would be right to call all of `net/url` deprecated in favor of `http-easy`, but maybe it would make sense to add pointers to `net/http-client` (as a lower-level interface to HTTP) and `net/http-easy` (as a higher-level interface) under the `URL Functions` section of the `net/url` docs, which is where most of the HTTP-related procedures are documented afaict.

PR #3130, opened 2020-04-27
Update bug_report.md

spdegabrielle:


samth:
Note that this isn't actually used -- we should figure out of this is what we want and then use it or remove it.

spdegabrielle:
closed due to lack of interest.

Issue #2974, opened 2019-12-16
Move Actions CI to a container based solution on Linux


pmatos:
We need to move our Linux jobs to a container based solution on Linux because the Linux VMs keep having intermittent issues like:
https://github.com/racket/racket/runs/348273932#step:6:102

Issues like these are well-known in the community:
https://github.community/t5/GitHub-Actions/ubuntu-latest-Apt-repository-list-issues/td-p/41122

The solution is to select a base container for the work in a job. I started working on this but we are facing another problem - also known in the community. The Actions CI syntax does not allow to conditionally select a container which means that to do so we need to separate the Linux/MacOS matrix further duplicating code (because it also does not support yaml).
More information is here:
https://github.community/t5/GitHub-Actions/Run-matrix-job-on-macos-and-on-ubuntu-in-container/m-p/32796#M1249

Current solution is possibly remove the matrix for `ubuntu-latest` and `macos-latest` and duplicate the jobs for these platforms.

pmatos:
The problem with this is that there are some tests that do not like to be ran inside a container. Take for example:
https://github.com/LinkiTools/racket/runs/350415970#step:10:86

These could be tests that for example, require access to special devices not available inside a container unless you have privileged access. @jackfirth might know more about these issues.

jackfirth:
Based on [this comment](https://github.com/racket/racket/blob/eda5f7a817f1588d783f2467c1d3fe0faf2ca529/pkgs/racket-test-core/tests/racket/subprocess.rktl#L450) in the failing test code, I think the problem is that some of the subprocess tests assume that the Racket process is *not* the init process (PID 1). But docker containers run the container's entrypoint as PID 1 and you don't get an init process unless you explicitly set one up when building the container. I think the right way to fix the test is to try to remove its assumption that there's an init process.

samth:
Couldn't we just use sh as the init process to work around this issue?

pmatos:
I am going back to this issue since it would be useful to get the work containerized. It will help reproduction later on when we have a script to trigger the actions locally (hopefully one day).

Given the comments from @jackfirth and @samth I will try to see what's going on with tests either assuming an init process or explicitly setting up one in the docker container. Maybe the later is better so we don't go around changing potentially many tests. Open to suggestions though.

jackfirth:
Either approach sounds reasonable to me.

[...]

PR #3134, opened 2020-04-30
Run Linux CI jobs inside docker container 

pmatos:


pmatos:
Not ready for review yet - marked Draft.

pmatos:
To fix #2974

pmatos:
> Have we seen more repetitions of the failures from #2974? If not, we maybe don't need to do this.

The reason to do this is not due to #2974. 
#2974 was a consequence of trying to do this.

There are two big reasons from the top of my head to do the ci inside a container.
* reproducibility: If something fails in an ubuntu-18.04 vm (in actions), we might not easily reproduce it. Even if you have ubuntu-18.04 or an image of it (using docker) you're not running the same environment as github since github pre-installs a bunch of stuff on the vm. Also, github can change the vm under your feet and things start failing without any changes to the repo. You avoid all of this by using a container tag, which only changes when you want to.
* self-hosting: for speed and cross-testing we might self-host in the future. With self-hosting you're stuck to running things in a container for security reasons. If you cannot run in a container then you cannot safely self-host the actions runner.

pmatos:
@samth not merging this yet since we still have test failures due to running inside a container.

pmatos:
We are close to getting this one closed but not yet... Strange thing is that there's a failure I cannot locally reproduce: https://github.com/pmatos/racket/runs/703260943?check_suite_focus=true#step:10:162

This `procedure-running?` keeps returning `#t` but `#f` is expected.

mflatt:
Looks like the test is bad, because it fails if the OS doesn't reap a process within 250ms. The repair is probably to keep waiting indefinitely — so a test failure means getting stuck, but that happens for some tests.

pmatos:
Even after @mflatt fix, we still have a problem with the test timing out. This, I think, it's racket unrelated. Probably related to how things work inside a docker container and this: https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/

I tried this locally and cannot reproduce which is quite strange but I will keep investigating.

pmatos:
> Even after @mflatt fix, we still have a problem with the test timing out. This, I think, it's racket unrelated. Probably related to how things work inside a docker container and this: https://blog.phusion.nl/2015/01/20/docker-and-the-pid-1-zombie-reaping-problem/
> 
> I tried this locally and cannot reproduce which is quite strange but I will keep investigating.

OK, so the reason I couldn't reproduce this locally is because I ran the call to docker differently than GitHub does. I ran it like this: `docker run -ti racket/racket-ci:testdeps /bin/bash` while GitHub instead of bash calls `tail -f /dev/null`. Bash is not an init process but will reap zombie processes, while `tail` will not. Using `--init` (current solution) works because docker will install `tini` as the init process (pid 1).

I was thinking of installing `dumb-init` in the container, but @mflatt pointed me to the `--init` option which I was not aware of. Thanks for that.

pmatos:
I think we are ready - this patch basically splits Linux and MacOS workflows since they cannot use the same jobs anyway because Linux now runs inside a container so splitting the files makes it easier to understand results. Mainly it runs everything inside a container which opens the way for better reproducibility, self-hosted runners (which are important for cross-compilation and testing).

This has been on my mind for quite a few months so I am happy to see this through. @samth you approved this a couple of weeks ago. Would you care to reapprove if you are still ok with it and have no questions?

PR #3135, opened 2020-04-30
Move snapshot generation to different workflow

pmatos:
Only run snapshot generation on push events to master.


pmatos:
@samth I think I captured here what we really want, which is snapshot generation for master, not for any other branch created within the repo.

.github/workflows/ci-snapshot.yml
```diff
@@ -0,0 +1,45 @@
...
+      - master
+
+jobs:
+  create-installer:
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v1
+    - run: |
+        git checkout -b github_actions_build_branch___ # have to be on a branch
```
samth, line 15: Is there an actions variable for the branch that was pushed?

pmatos:
I will wait for further comments to merge this since it seems you might have some improvement in mind.

.github/workflows/ci-snapshot.yml
samth: Ah, good to know. Then I can fix that to checkout the correct branch.

The problem is that snapshot building clones and then pulls from the repository that you start with, which doesn't work in "detached head" mode. So we need to be on some branch.

samth:
My goal is to (eventually) have all the release building done automatically via CI, but that's a long way off. For the moment, this is good.

PR #3144, opened 2020-05-01
pattern-expanders should not arm output patterns

### LINKED ISSUE #2199: pattern expanders can't use syntax-protect or syntax-rules
Right now, pattern expanders can't use `syntax-protect`, which means they can't use `syntax-rules` either. However, this is not documented anywhere. We should fix this by either:

 1. Documenting the fact that pattern expanders shouldn't use `syntax-arm` or `syntax-protect` on their results (and so shouldn't use `syntax-rules` either).
 2. Changing syntax-parse so that it can disarm the "core" pattern forms, making `syntax-protect` okay in pattern expanders, and so making `syntax-rules` okay too.

See also #2197.

AlexKnauth:
Closes #2199 by documenting the fact that pattern expanders shouldn't use `syntax-arm` or `syntax-protect` on their results (and so shouldn't use `syntax-rules` either). Does the same for sequence syntax / for-clause transformers.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/for.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/for.scrbl
@@ -152,7 +152,7 @@ element.
               #:contracts ([length-expr exact-nonnegative-integer?])]{
 
 Iterates like @racket[for/list], but results are accumulated into
-a vector instead of a list. 
+a vector instead of a list.
 
 If the optional @racket[#:length] clause is specified, the result of
 @racket[length-expr] determines the length of the result vector.  In
@@ -520,7 +520,7 @@ nested.
 
 @deftogether[(
 @defform[(for*/list (for-clause ...) body-or-break ... body)]
-@defform[(for*/lists (id ... maybe-result) (for-clause ...) 
+@defform[(for*/lists (id ... maybe-result) (for-clause ...)
            body-or-break ... body)]
 @defform[(for*/vector maybe-length (for-clause ...) body-or-break ... body)]
 @defform[(for*/hash (for-clause ...) body-or-break ... body)]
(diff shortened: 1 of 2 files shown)


AlexKnauth:
After fixing the `@tech{dye packs}` reference here are screenshots of the rendered docs,
For `pattern-expander`:
pattern-expander
For `define-sequence-synatx`:
define-sequence-syntax

pkgs/racket-doc/scribblings/reference/for.scrbl
```diff
@@ -716,7 +716,16 @@ instead of @racket[syntax-protect].
 (for/list ([d (in-digits 1138)]) d)
 
 (map in-digits (list 137 216))
-]}
+]
+
+Sequence syntax transformer procedures should not arm their
+output clauses with syntax @tech{dye packs}, though they
+may arm expressions embedded within the clauses. Because of
+this, avoid using @racket[syntax-rules] for
```
sorawee, line 724: The Racket Guide for `syntax-protect` mentions `syntax-rules`, `syntax-id-rules`, and `define-syntax-rule` as forms that automatically protect their expansion results (in the stdlib). Should we include all of them here too for consistency? Alternatively, use the phrase "avoid using forms like @racket[syntax-rules]"?

AlexKnauth: That's a good idea. Does

> avoid using forms like `syntax-rules` and `syntax-id-rules`

Sound good enough? I don't think `define-syntax-rule` applies as much because it doesn't return a transformer value like `syntax-rules` and `syntax-id-rules` do.

sorawee: Oh yeah, good point!

AlexKnauth: I have updated it so the pattern-expander docs say "forms like `syntax-rules` and `syntax-id-rules`, and the sequence-syntax docs say "forms like `syntax-rules`".

sorawee:
This is obsolete since syntax arming is no longer used since Racket 8.3. Closing.

PR #3147, opened 2020-05-03
docs: for/stream: mention multiple values

yurkobb:
Resulting from [a discussion on racket-users mailing list](https://groups.google.com/forum/#!topic/racket-users/Cavti-N5u6o), a small improvement to the documentation of the `for/stream` form.

Before I am confident with this, I need a little advice:
1. Is there a general approach to (not?) recommending specific third-party packages in the main distribution documentation? I included a mention of the [stream-values](https://pkgs.racket-lang.org/package/stream-values) package directly, but what if someone comes up with a different package - the docs would then give preference to one package of many. Is that desirable?
2. I'm not sure the link to the package is the best way to do it - maybe I should link to the manual instead?
3. I would also add a link to the racket guide under the mention of "multiple values" - would [`@other-doc`](https://docs.racket-lang.org/scribble/base.html#%28def._%28%28lib._scribble%2Fbase..rkt%29._other-doc%29%29) be the correct way to do it?

Also, I'm building racket locally now to inspect the rendered docs and make sure there are no parsing errors.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sequences.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sequences.scrbl
@@ -1192,10 +1192,12 @@ stream, but plain lists can be used as streams, and functions such as
   @racket[body] will not be evaluated until the resulting stream is forced. This
   allows @racket[for/stream] and @racket[for*/stream] to iterate over infinite
   sequences, unlike their finite counterparts.
-  
+
   Please note that these forms don't support returning multiple values on each
-  iteration, but third party packages exist that provide this functionality.
-  
+  iteration, but third-party packages exist that provide this functionality
+  (see @hyperlink[https://pkgs.racket-lang.org/package/stream-values]{stream-values}
+  for example).
+
   @examples[#:eval sequence-evaluator
     (for/stream ([i '(1 2 3)]) (* i i))
     (stream->list (for/stream ([i '(1 2 3)]) (* i i)))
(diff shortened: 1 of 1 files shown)


sorawee:
Oh... let's not reference external packages. I thought that you will only mention the fact that `for/stream` doesn't support multiple values.

yurkobb:
Okay, I have removed the mention of external packages and linked the guide section on multiple values. Also tested building locally. Submitting.

SimonLSchlee:
Maybe mentioning a package isn't so bad? 
I mean it is only natural that there will be things which aren't solved in the main libraries, but because they have solutions that exist as packages, they also won't be added to the main libraries.
In the racket/date documentation there is precedent in the form of mentioning gregor, which I might have found way later without it. 
https://docs.racket-lang.org/reference/time.html?q=date#%28part._date-string%29

I can understand not wanting to promote one out of many "competing" packages, but to me this seems to be a useful special case, where a user would expect to have such a tool.
For the case of multiple packages, why not mention multiple? Or maybe link to a shared tag on the package site?
Personally I think greater interconnected-ness in the form of more mentions is a great thing.

bennn:
In this case, searching for `for/stream` on docs.racket-lang points to [these docs and to stream-values](https://docs.racket-lang.org/search/index.html?q=for%2Fstream).

I know that's not much, but I think we should merge this PR without the link and think about something more systematic for pointing caveats to packages.

@yurkobb can you change `don't` to `do not`

sorawee:
One more nit in addition to @bennn's: I think you should use `@tech{multiple values}` instead of `@seclink[...]{multiple values}`. It will link to the definition of multiple values in the Racket reference.

yurkobb:
Updated accordingly, rebased on current master.

PR #3151, opened 2020-05-04
Use CPUS instead of -j on top level makefile

pmatos:
AFAIU -j will only work on a make file level but is not propagated to
raco.

### DIFF
--- a/.github/workflows/ci-snapshot.yml
+++ b/.github/workflows/ci-snapshot.yml
@@ -32,7 +32,7 @@ jobs:
         DISTRO_BUILD_SITE_DEST: "${{ runner.temp }}/site-dest/"
       run: |
         export cpus=$(grep -c ^processor /proc/cpuinfo)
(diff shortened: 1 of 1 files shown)


pmatos:
Pretty sure I discussed this with @mflatt awhile ago and this was the whole point of `CPUS` vs `-j`.

pmatos:
> I think installers doesn't handle CPUS, though.

I am confused - the installers target uses other targets to build bc and cs so I would expect it to work. I see for example:
```
cpus-cs:
	$(MAKE) -j $(CPUS) plain-cs JOB_OPTIONS="-j $(CPUS)"
```

samth:
I should somehow figure out how to check this. But you're probably right.

mflatt:
I think `CPUS` probably does not work for `snapshot-site` and similar, even for the server-side build, because those targets go through targets like `plain-base` instead of `base`.

pmatos:
Closing pull request since `CPUS` don't work on `snapshot-site` (see @mflatt comment).

PR #3152, opened 2020-05-04
Only create snapshots when action is ran from main repo

pmatos:


pmatos:
If you know the secrets, you can remove the line locally on your own fork. This just avoids starting the jobs for each racket fork that's committed and end up failing when it gets to using the secrets.

PR #3157, opened 2020-05-05
Move static analysis jobs to their own CI file to split workflows

pmatos:


pmatos:
The motivation here is that by splitting jobs into more workflows when possible we get a better overview of things on the Actions tab, while ensuring that we can rerun jobs more selectively (The 're-run jobs' button always reruns a whole workflow.

The reason I didn't do this earlier is because I needed cgc to start static analysis of 3m and cs earlier and forgot about the use of the snapshot from the setup-racket action. Hopefully it will work 99.9% of the time. It might not work if between a build that we are analysing and a snapshot some internal bytecode or something of the sort is modified/added/deleted, I presume. However, that's probably a quite rare these days.

@mflatt @samth what do you think?

samth:
While failure is unlikely based on a snapshot, why worry about the extra 1 minute of compile time for the static analysis builds that building cgc causes?

pmatos:
> While failure is unlikely based on a snapshot, why worry about the extra 1 minute of compile time for the static analysis builds that building cgc causes?

hummm, well it's still about 6mins, plus the extraction/download of the artifact on the subsequent jobs.
https://github.com/racket/racket/runs/645515166?check_suite_focus=true

Of course, we can rebuild on this workflow. Tbh, for reliability reasons, the right thing might be to rebuild it here in this workflow. Maybe even inside each job so we don't pay the compression and artifact upload price.

pmatos:
I modified the jobs, lets see how it goes.

pmatos:
Also, I added the static analysis to pull_request just to test in the PR. I will remove it, so it runs on push only once we agree this is fine. Couldn't get `act` to build this locally.

pmatos:
It looks great now - and the speed build of racketcgc takes 3mins!

pmatos:
And the racketCS static analysis only reports 21 errors. :open_mouth:

samth:
When I last looked, I think they were all false positives.

PR #3159, opened 2020-05-06
Add workflow to test racket with sanitizers

pmatos:


pmatos:
> Maybe there should be a flag to `raco test` that does the `tee` for you.

I will open an enhancement request for that.

pmatos:
Not merging this yet because we are once again having problems with temp directories inside containers.

PR #3160, opened 2020-05-06
Remove gitlab CI

pmatos:
After #3159 is merged, gitlab CI files can be removed.


pmatos:
I am wondering if moving the docker images for `racket/racket-ci` is going to cause problems with DockerHub builds...

samth:
Yes but it should be easy to change.

samth:
I wonder if the Docker images should go in the racket-ci repo. They don't need to rebuild on every racket commit.

pmatos:
> I wonder if the Docker images should go in the racket-ci repo. They don't need to rebuild on every racket commit.

Unsure we can do that from docker but from Actions we could have a workflow to rebuild and push the images only when the dockerfiles change. Would you like me to add a PR to do that?

PR #3170, opened 2020-05-09
Add precise annotations to improve scanbuild

pmatos:


.github/workflows/scanbuild_static-analysis.yml
```diff
@@ -25,6 +25,7 @@ jobs:
       run: >
         ./configure
         CFLAGS="-O0 -g"
+        CPPFLAGS="-DMZ_NORETURN -DMZ_PRECISE_RETURN_SPEC -DMZ_UNREACHABLE"
```
mflatt, line 28: DO you want `-DMZ_NORETURN`? I thought that would disable the effect of `-DMZ_PRECISE_RETURN_SPEC`?

pmatos: You're right! I misread the defines. Sorry about that. Updated. 
Unreachable attributes are also defined by default so no need for `MZ_UNREACHABLE`. 

Feel like they need better names but ok for this PR I guess.

PR #3175, opened 2020-05-12
Add future tests to CI

pmatos:
After @mflatt work on futures for CGC (https://github.com/racket/racket/commit/761c577479852938086b7a83946fc03ef5cd1f84),  here's a second attempt at enabling tests.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -32,6 +32,7 @@ jobs:
         raco test -l tests/match/main
         raco test -l tests/zo-path
         raco test -c tests/xml
+        raco test -c tests/future
         raco test -l tests/db/all-tests
         raco test -c tests/stxparse
         raco test -c tests/syntax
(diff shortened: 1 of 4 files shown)


pmatos:
Lets see how this goes before merging...
https://github.com/pmatos/racket/actions/runs/102220122

pmatos:
So, this is not yet ready: https://github.com/pmatos/racket/runs/666292138?check_suite_focus=true#step:23:20281

It says the `custodian-limit-memory` is unsupported on the platform (read CGC).
Which is strange to me because the docs don't say this is unsupported in CGC, just that memory accounting needs to be available. A quick glance at the code didn't allow me to decipher what's going on here but initially I thought it was lack of threads but threads are enabled for the build.

Anyway, we are one step further. If nobody looks at this in the meantime I will investigate more after work.

pmatos:
@mflatt is `custodian-limit-memory` not supposed to work on CGC?

mflatt:
That's correct.

mflatt:
There are two tests that won't work for CGC, and probably it's best to just make them do nothing when `(system-type 'gc)` is `'cgc`

pmatos:
CI looks good now - merging.

PR #3177, opened 2020-05-12
Reference to the Friendly Environment Policy should be put into the right section

kryptine:


mfelleisen:
Why do we have to impose Friendly on contributors?

samth:
I don't think the Friendly Environment policy should be thought of as part of copyright issues, or applying only to people who contribute code. Is there a different phrasing that would accomplish what you want here?

mfelleisen:
I do not think that there should be a phrase/link at all. The very fact that a Friendly exists is enough.

mfelleisen:
Thank you.

kryptine:
What are we supposed to do now? Remove the reference to the FEP, move it to a separate file and rephrase it, or anything else?

mfelleisen:
Remove it. 

A Friendly is like a Style guide. If someone violates it, we point it out if it's important. Otherwise it stands for itself.

samth:
I'll discuss this with Matthias off-line and we'll come to an agreement.

kryptine:
Have you come to an agreement yet?

samth:
We discussed this, and we came to two conclusions.

1. We want to keep all the information that's currently in `README.md` there.
2. We already have a file in `.github/CONTRIBUTING.md` with most of the information there, but which is missing links. Changing this PR to add those, or to add additional helpful text there would be great.

kryptine:
I have added the links.

samth:
Thanks for noticing the issue in #3219. I think my previous comment was unclear. We don't want to add a top-level `CONTRIBUTING.md` file -- we've tried to minimize the number of files in the top-level of the repository, and the `.github/CONTRIBUTING.md` file provides the same experience on GitHub. Since we're not adding that file, I also want to keep information in the `README.md` file that's there now. Adding more info to `.github/CONTRIBUTING.md` (and fixing the discrepancy between them) is great, as is adding more info to `README.md` as needed.

PR #3178, opened 2020-05-12
Ensure that the GMP ARM extra file has the correct extension

### LINKED ISSUE #3176: Racket 7.7 fails to build on armhf
Hi, I am currently working on the [Alpine Linux](https://alpinelinux.org/) racket package. While upgrading it to 7.7 I noticed that the build failed on armhf. The error message is:

```
make[7]: *** No rule to make target 'gmp_arm_gcc.o', needed by 'mzobjects'.  Stop.
```

I haven't investigated this further yet but may this be a parallel build issue? The full build log: [racket-7.7-r0.log](https://github.com/racket/racket/files/4614386/racket-7.7-r0.log)

pmatos:
We hardcoded in some places the extension of the GMP ARM extra file
with .o, however the Makefile rule to build the file will use .lo is
--enable-shared is enabled. This commit fixes the discrepancy by not
hardcoding the extension anywhere.

Fixes #3176


pmatos:
@mflatt I used `addsuffix` in make which definitely works with gnu make I tried on linux. I couldn't find any information on portability of `addsuffix`. I remember at certain you were worried about certain other makes but can't exactly remember if they are relevant for this particular PR.

mflatt:
Yes, this is a place where we need to avoid GNU make extensions. A possible alternative is to add `LTO=@LTO@` to the makefile and use `EXTRA_GMP_OBJ='gmp_arm_gcc.$(LTO)`.

mflatt:
Was there a problem witth using `EXTRA_GMP_OBJ='gmp_arm_gcc.$(LTO)'` in `configure.ac`? That would avoid the need for a conditional in the makefile.

pmatos:
> Was there a problem witth using `EXTRA_GMP_OBJ='gmp_arm_gcc.$(LTO)'` in `configure.ac`? That would avoid the need for a conditional in the makefile.

Initially my first attempt was to do that but I had forgotten to `LTO=@LTO@`.
Then I posted my initial review which had the makefile problem. After you mentioned that we can do `LTO=@LTO@`, I applied that in the Makefile. I think now it's the cleanest when we do this in the configure file. I will see the CI results and run it as well on the arm to confirm.

pmatos:
@mflatt its working on the arm, ok to merge is CI is green?

mflatt:
This still doesn't look right to me, because it looks like `$(LTO)` is substituted too early, and `EXTRA_GMP_OBJ` got lost from the makefile. The commit above shows what I had in mind.

pmatos:
> This still doesn't look right to me, because it looks like `$(LTO)` is substituted too early, and `EXTRA_GMP_OBJ` got lost from the makefile. The commit above shows what I had in mind.

Took me a while to understand that I was testing the wrong commit. :-1: 
It looks pretty ugly since it looks like you have a dangling assignment in `Makefile.in` but it does seem to work.

pmatos:
But the arm is failing now - something is off but I have the kids running circles around me. I will take a look again once I have some peace and quiet.

mflatt:
I think you need to change `$(EXTRA_GMP_OBJ)` in the makefile back to `@EXTRA_GMP_OBJ@`.

pmatos:
@mflatt sorry for the back and forths. what do you think?

PR #3183, opened 2020-05-13
Change `tanh` to produce inexact values when approximating near 0.

samth:
Since the approximation is that (tanh x) = x for small x, previously
(tanh 1/10000000000000000000) produced an `exact?` answer, even though
this is an approximation.

Found by Typed Racket random testing:
http://drdr.racket-lang.org/54493/racket/share/pkgs/typed-racket-test/tr-random-testing.rkt


mflatt:
LGTM

PR #3184, opened 2020-05-13
Mark a few more simple expressions as `immediate-default?`.

samth:
This reduces the number of `unsafe-undefined` uses in the flattened
expander by about 10% (682 -> 611).


mflatt:
Why `car` and `cdr`? (I didn't see any specific cases covered by those in the expander, and they seem more arbitrary than other things.)

Probably there should be an `immutable?` test on hash tables, just to be safe.

samth:
The uses of `car` are all in `racket/private/string` and while loading the expander compiles them, they'll be dropped from the flatted linklet, so I've removed them.

I omitted `immutable?` because the string and bytes cases don't have that check. Should I add it there too?

mflatt:
Fair point on mutable strings. I'm ok with continuing the assumption that the value is immutable or ok to reference multiple times.

samth:
Would it make sense to allow all identifiers whose binding is from `'#%kernel`?

bennn:
How could a mutable hash show up here ... is 3d syntax the only way?

If so, maybe it's worth adding vector and struct literals too.

PR #3186, opened 2020-05-14
Fix bug in Racket Guide (7.3.7 Notation).

NoahStoryM:
```(begin (set! x (cons 'ctc x)) any/c)``` should be evaluated whenever ```(f)``` is called. If there isn't an arg list ```()```, it would only be evaluated 1 time.

https://github.com/racket/racket/issues/2711


jbclements:
@NoahStoryM How would you like your name to be listed in the list of Racket 7.8 contributors? NoahStoryM?

(Sorry for the off-topic question.)

PR #3191, opened 2020-05-15
Add i386 to CI

pmatos:


pmatos:
This is ongoing work and not ready to merge due to the current OOM error here:
https://github.com/pmatos/racket/runs/674110574?check_suite_focus=true#step:9:442

samth:
This all looks sensible to me. Can you squash before merging, though?

samth:
Can you try disabling test steps on everything that isn't the most common versions, (3m with jit and no-cify, CS)?

pmatos:
> Is this ready?

Sorry - I have been busy with other things. Let me look into this next.

pmatos:
Can't believe this has been opened since May... *sigh*

pmatos:
To test i386, there's a really good solution - use a i386 docker image running on a x86_64 machine. This works really well because the host provides backwards compatibility and docker allows us to specify the platform with `docker create --platform linux/386 ...`.

Unfortunately, `--platform` requires us to enable experimental features in the daemon, which is not supported on github hosted runners, so I will add an x86_64 runner for this.

PR #3196, opened 2020-05-20
implement dict/c in racket/dict/contract

AlexKnauth:
This pull request adds a `dict/c` contract combinator to a new module `racket/dict/contract`. `(dict/c key/c value/c)` is similar to a generic dictionary version of `(hash/c key/c value/c)`.

### DIFF
--- a/racket/collects/racket/dict/contract.rkt
+++ b/racket/collects/racket/dict/contract.rkt
@@ -1,0 +1,501 @@
+#lang racket/base
+
+(provide dict/c)
+
+(require racket/contract
+         racket/dict
+         racket/match
+         racket/list
+         (for-syntax racket/base
+                     racket/list
+                     racket/syntax
+                     syntax/stx))
+
+;; ---------------------------------------------------------
+
+;; dict/c :
+;;   ChaperoneContract
+;;   Contract
+;;   [Contract]
+;;   [#:immutable (U Bool 'dont-care)]
+;;   [#:flat? Bool]
+;;   ->
+;;   Contract
+;; The iter contract is only checked if it's not flat
+(define (dict/c dom rng
+                [iter any/c]
+                #:immutable [immutable 'dont-care]
+                #:flat? [flat? #f])
+  (unless (member immutable '(#t #f dont-care))
+    (raise-argument-error 'dict/c
+                          "(or/c #t #f 'dont-care) for the #:immutable argument"
+                          immutable))
+  (define dom-ctc (if flat?
+                      (coerce-flat-contract 'dict/c dom)
+                      (coerce-chaperone-contract 'dict/c dom)))
+  (define rng-ctc (if flat?
+                      (coerce-flat-contract 'dict/c rng)
+                      (coerce-contract 'dict/c rng)))
+  (define iter-ctc (if flat?
+                       any/c
+                       (coerce-contract 'dict/c iter)))
+  (cond
+    [(or flat?
+         (and (eq? immutable #t)
+              (flat-contract? dom-ctc)
+              (flat-contract? rng-ctc)
+              (flat-contract? iter-ctc)))
+     (flat-dict/c dom-ctc rng-ctc any/c immutable)]
+    [else
+     (impersonator-dict/c dom-ctc rng-ctc iter-ctc immutable)]))
+
+;; immutable-dict? : Any -> Bool
+(define (immutable-dict? d)
+  (and (dict? d)
+       (dict-implements? d 'dict-set 'dict-remove)
+       (not (or (dict-implements? d 'dict-set!)
+                (dict-implements? d 'dict-remove!)))))
+
+;; mutable-dict? : Any -> Bool
+(define (mutable-dict? d)
+  (and (dict? d)
+       (dict-implements? d 'dict-set! 'dict-remove!)
+       (not (or (dict-implements? d 'dict-set)
+                (dict-implements? d 'dict-remove)))))
+
+;; dict-key-chaperone-allowed? : Any -> Bool
+;; Returns true if `d` can work with chaperones on keys.
+;; Hashes based on `equal?` can, but not on `eq?` or `eqv?`
+(define (dict-key-chaperone-allowed? d)
+  (cond [(hash? d) (hash-equal? d)]
+        [else #t]))
+
+;; ---------------------------------------------------------
+
+;; ... --> boolean
+;;  returns #t when it called raise-blame-error, #f otherwise
+(define (check-dict/c dom-ctc immutable flat? val blame neg-party) 
+  (cond
+    [(dict? val)
+     (cond
+       [(and (not flat?)
+             (not (flat-contract? dom-ctc))
+             (not (dict-key-chaperone-allowed? val)))
+        (raise-blame-error
+         blame val #:missing-party neg-party
+         '(expected
+           "equal?-based dict due to higher-order domain contract"
+           given:
+           "~e")
+         val)
+        #t]
+       [else
+        (case immutable
+          [(#t) 
+           (cond
+             [(immutable-dict? val) 
+              #f]
+             [else
+              (raise-blame-error 
+               blame val #:missing-party neg-party
+               '(expected "an immutable dict" given: "~e") val)
+              #t])]
+          [(#f)
+           (cond
+             [(mutable-dict? val)
+              #f]
+             [else
+              (raise-blame-error 
+               blame val #:missing-party neg-party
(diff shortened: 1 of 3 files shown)


rfindler:
Is there a reason not to put this into `racket/contract`? I guess the dependencies would have to be trimmed (but it would be fine if it didn't go into `racket/contract/base`.)

Is it possibly to spell "dictionaries" correctly in the docs?

The unsoundness around flat contracts applied to mutable data structures came about in other combinators because of backwards compatibility concerns. I am torn here -- it would be nice to avoid introducing that into new combinators but consistency with the other parts of the library is a competing (and good) concern too. 

And can you explain why we can't have chaperone contracts and have to go directly to impersonator ones? (Or am I misunderstanding that part of the docs?) Maybe this is also related to the issues wrt to flat contracts that I asked in the previous paragraph?

rmculpepper:
It seems like this should be able to use `{chaperone,impersonate}-struct` with the struct type property accessor to redirect the dict method table. See the general approach of `redirect-generics` from `racket/generic`.

But maybe we have different behaviors in mind. In general, if a value implements two interfaces X and Y, then if we apply a contract to the X interface, does the result still implement the Y interface? (The interfaces can be related, like "sequence" and "is a list", or "dict" and "is a hash".) I haven't looked recently, but if the Racket documentation doesn't establish terminology for this, it would be helpful to add. And unless one mode is clearly wrong or impractical, it would be good to support both modes whenever possible. I have a hard time reconciling the version that removes the Y interface with the idea that chaperones cooperate with `equal?`, though.

Re placement: I think this is better in `racket/dict`. My impression is that contracts related to things present in `racket/base` go in `racket/contract`; contracts for things defined in other libraries go in those libraries (eg, `object/c` in `racket/class`).

rfindler:
@rmculpepper re placement: totally. That makes more sense than my suggestion. (Of course, `racket/contract` may or may not totally follow this rule (didn't check...) but I agree that that is a nice rule!)

mflatt:
There's at least one precedent for removing Y when chaperoning X: `chaperone-evt` turns various things (ports, channels, etc.) into just evts.

But I think mostly keeping Y is more common. For example, chaperoning an applicable struct as a procedure does not remove any of the struct properties.

AlexKnauth:
(Re: @rfindler And can you explain why we can't have chaperone contracts and have to go directly to impersonator ones?)

Because I don't know how to chaperone all dicts inscluding all the possible structs that could possibly implement `gen:dict` or `prop:dict`. In particular I tried using `chaperone-generics gen:dict`, but even basic things like `(chaperone-generics gen:dict '())` and `(chaperone-generics gen:dict (hash))` fail.

AlexKnauth:
Although `chaperone-generics` does seem to work for structs, even opaque structs, that implement `gen:dict` or `prop:dict`, so perhaps I can rewrite this to dispatch on whether it's a non-struct dict like list-hash-vector, and otherwise uses `chaperone-generics`.

rfindler:
Not sure how much sense it would make but perhaps worth considering if the contract itself could list the generics that it was meant to work for and then that listing could somehow convey the information needed to set up the chaperones properly? (And then we could default to some reasonable set of them somehow?)

jackfirth:
Does the `sequence/c` combinator face similar issues surrounding subtype perseverance? Do we want `dict/c`'s handling of those issues to be consistent with `sequence/c`?

AlexKnauth:
`sequence/c` create a new sequence with `make-do-sequence`. The current draft of `dict/c` creates a new struct that implements `gen:dict`, but I'm currently trying to rewrite it to a version that dispatches on whether it's a "primitive" dict or a "struct-generic" dict and uses different chaperones for those so it doesn't have to do that.

jackfirth:
Should `sequence/c` do that too? (Which would be a totally separate pull request.)

AlexKnauth:
I have rewritten it to use `redirect-generics gen:dict` when it's a struct-dict and not a built-in dict type such as hash/assoc/vector. However, I still don't know how to make it a chaperone for all methods.

I'm specifically having trouble with `dict-ref`, with the optional `failure-result` argument. When the `failure-result` is used, I don't want the "value" contract to apply to that result. But when the failure-result isn't used, and the "value" result happens to be exactly equal to the failure-result, I do want the "value" contract to apply to it. I don't know how to distinguish those. How do I distinguish those though?

I tried to take inspiration from Carl Eastlund's `dict/c` implemented in `mischief/contract`, but it doesn't use chaperones so it is free to both add an escape continuation, and change the `failure-result` input to always be a new synthesized procedure. That way the new synthesized failure-result procedure can call the escape continuation. If I'm constrained to chaperones, I don't know how to do either of those things.

How can I detect whether the `failure-result` is used-or-not by the dict's implementation of `dict-ref`, while keeping the `failure-result` a chaperone of the original failure-result?

I have also asked on the Racket Users list: https://groups.google.com/d/msg/racket-users/kHQmJh63-mU/8SXYlmZzAAAJ

AlexKnauth:
Okay, I have done "something" to try to make it do chaperones. However before I update the documentation, I want feedback on whether the weird things I did in the [manual implementation of `redirect-ref`](https://github.com/racket/racket/pull/3196/files#diff-72fd559ac6fbcabde3075b1dbcea57bfR638-R688) look right or not. In my opinion it is very convoluted and feels dirty.

The "easy" case is if the "default" argument supplied is already a procedure. In that case it chaperones that procedure and uses mutation to tell whether it was called or not, and if so what the result was.

The "hard" case is where the "default" argument is not a procedure. Since the chaperone can't make it a procedure in the actual official call, I have it do an extra call to the function it's supposed to be chaperoning during the chaperone itself. There's something about this that feels evil... it means running the users implementation of `dict-ref` twice for a single call.

Am I wrong and this stuff is actually okay? Or is my feeling justified and this is something to be avoided enough that it's not worth making it a chaperone contract? Or, is there a way around this that allows it to distinguish whether the default was used or not in this case?

racket/collects/racket/dict/contract.rkt
```diff
@@ -0,0 +1,738 @@
...
+                       (λ (ans)
+                         (set! default-result ans)
+                         ans))))]
+           [else
+            (define default-called?
+              (let/ec default-called!
+                ;; TODO:
+                ;; Is calling the function being chaperoned here evil?
+                ;; It kinda feels evil but I'm not sure.
+                (op self key (λ () (default-called! #t)))
```
AlexKnauth, line 680: This line specifically is what feels evil to me. Is it wrong in a procedure chaperone, to call the function I'm supposed to be chaperoning during the chaperone's wrapper-proc? This is what would make it run the user's implementation of `dict-ref` twice for a single call.

AlexKnauth:
I have updated it to use `redirect-generics` to create an impersonator. I can't make it a chaperone due to the problems above with `dict-ref`'s `failure-result` argument, so is an impersonator enough?

jackfirth:
Not supporting chaperone contracts is too bad, but I think it's better than the alternative of adding a whole new interface that implementations need to extend just to support contracts.

rfindler:
For the record, and perhaps unsurprisingly, I don't agree with this
particular weighting of these two options. :)

I am not sure I have followed all of the technical details but it certainly
is the case that each new data type we add needs it have its own
implementation of chaperones (and that has worked well for us, both better
than not having contracts and better than doing things like trying to use
function wrappers for functions, etc.)

Having hooks in the generics system to expose each data types
"chaperoneness" seems right in line with that.

Robby

AlexKnauth:
Hm. I just thought of a different angle. The worst problem with the chaperone version only comes because of the non-procedure version of failure-result. So it can be side-stepped if the `dict-ref` method from the `gen:dict` interface always receives a procedure for failure-result.

I think this would require separating the user-callable `dict-ref` function from the generic-interface hook version. The user-callable `dict-ref` would pass it through to the hook version if it's a procedure, but create a thunk if it's not a procedure. That way the hook version always receives a procedure whether the user passed it as a procedure or not.

Does that sound like a good plan? Are there any pitfalls I'm not aware of?

AlexKnauth:
There is a pitfall with separating the user-callable `dict-ref` and the hook `dict-ref`: with `define/generic` users can call the hook version directly with a non-procedure failure-result argument:
```
(define/generic gen-dict-ref dict-ref)
(gen-dict-ref dict key #f)
(gen-dict-ref dict key 0)
(gen-dict-ref dict key '())
```
This would break if the `dict` had a contract that relied on this separation because the contract's version of `dict-ref` needs the failure-result to be a procedure in order to chaperone it, at least the way I'm approaching it.

Is there a way around this?

PR #3197, opened 2020-05-20
Check for 0.0 imaginary part in `exp`.

### LINKED ISSUE #3194: racket and racketcs disagree on exp on large imaginary number
```
[samth@homer:~/sw/plt/extra-pkgs/gui (master) plt] racket -e "(exp 1.7e+308+0.0i)"
+inf.0+nan.0i
[samth@homer:~/sw/plt/extra-pkgs/gui (master) plt] racketcs -e "(exp 1.7e+308+0.0i)"
+inf.0+0.0i
```

This currently makes it difficult to fix a soundness bug in the TR optimize on Racket CS, where the optimizer causes the Racket BC result on Racket CS. The options there are:
1. Declare this not a problem, even though the optimizer makes the result worse on Racket CS from the untyped program.
2. Fix this bug.
3. Remove optimization of `exp` on imaginary numbers.

samth:
Follows Chez Scheme and Guile. Turns `(exp 10000.+0.0i)` into
`+inf.0+0.0i` instead of `+inf.0+nan.0i`, which is analagous to
the behavior for exact 0 in the complex part.

Fixes #3194.


mflatt:
Ok as far as I can tell. It would be good to add tests with `-0.0` as the imaginary part to make sure the sign is preserved.

PR #3200, opened 2020-05-21
Add initial LGTM configuration for Racket

pmatos:


pmatos:
First time looking into LGTM config - tries to fix extraction failing as seen here:
https://lgtm.com/projects/g/racket/racket/rev/pr-96db9356619a8d5e8c54115f532e7bf52165f8cd

samth:
This used to work without configuration (as in, yesterday). Do you know what happened?

pmatos:
> This used to work without configuration (as in, yesterday). Do you know what happened?

Last time it worked seems to have been on Feb 1st. Not sure what changed back then and I didn't notice the failure because it was not clear. I cannot access the setting for the racket project so am unsure how the setup but surely you can see if anything changed and if the setup is for both PRs and Pull Requests. 

If you go to https://lgtm.com/projects/g/racket/racket/?mode=list
You'll see that in the hover window that shows up when you go to the red `!` triangle, the results you see there are for `commit c81228f31eec2f023409e3c523b49ce5c1ca911f`, which was probably the last one that worked. To be honest, I find the interface slightly confusing but I am sure that LGTM is definitely not extracting C/C++ properly due to the lack of libcairo.

samth:
Apparently there are no settings, and that's the only supported location for the config file, so this is fine. We should probably change it to run `make base`, though, since it's a static analysis and thus running `raco setup` for lots of packages isn't so helpful.

pmatos:
I think now by setting a step for `configure` and `index`, this is doing the right think and *not* running `raco setup` for all those packages.

PR #3202, opened 2020-05-21
Add more examples to I/O Patterns in the Guide

alomew:
My thoughts here were to make it as easy as possible to use a port in your code, especially if you've never used them before.

I tried to introduce examples that

1. more obviously used an external input (in this case a file) rather than using a mock string
2. have examples that only use input -- "how do I deal with this input functionally?"

It might be bold to have an almost anti-port example first, but I'm thinking it's good to have something really copy-pasteable for someone who hasn't gotten much headway with the rest of the section, and just wants access to the contents of a text file in their program.


mflatt:
Thanks for this improvement, apologies for the delay. I'll merge this so it's in the upcoming release. I hope it's ok that I tweak a few style details afterward, like using `filebox` for the example files.

PR #3203, opened 2020-05-21
Strip package-content-state from info files when stripping to source

florence:
Currently `raco pkg install --source` leaves `'built` packages marked as `'built` packages.


mflatt:
Running tests with this commit shows a failure in "test-binary.rkt". It's in the "check content of source package" case, which probably reflects an intended change in this commit. Can you take a look?

mflatt:
I'll change `fixup-info` to not write "info.rkt" if the new content is equal to the old content. That seems like a good idea in general, and it avoids the test failure.

PR #3206, opened 2020-05-22
Build Docker Images from GitHub Actions

pmatos:


pmatos:
This will only work once we add the password for `racketci` to secrets. Contacted @samth privately.

pmatos:
Once this is merged, we remove the Autobuild on the docker side.

PR #3215, opened 2020-05-25
Reduce racket-doc dependencies.

samth:
Substantially reduces the dependencies of "racket-doc" (and thus many other things). This makes installing many packages that have documentation faster and download much less.

Notable removals include "plot", "math", "drracket", and "web-server".

This is accomplished in a few ways:
- using `#:indirect` in many references to module names
- a new `scribblings/private/docname` library that includes the titles of commonly-referenced documents so they can be referred to with `seclink` and `#:indirect`.
- Rewriting a few bits to not refer to things (like specific identifiers) that cannot be `#:indirect`
- Similar changes to a bunch of other documentation packages.

There is one potentially-controversial change here, which is the creation of the "scheme-doc" package from some of the documents in "racket-doc", without making "racket-doc" depend on "scheme-doc". That requires adding new dependencies in some packages. I've already written the necessary changes for packages in the main distribution, and I plan to do the same incrementally for other packages. I don't think this is a big problem, though, since (a) it will still be in the main-distribution, and (b) empirically no one cares about having dependency errors.

There is also a change to improve dependency error reporting; you now get an indication of what part of the other package (such as a particular document tag) is depended on.

There is also a change to `setup-printf` to make it somewhat more robust, otherwise using it with a string involving `struct~s/c` goes poorly.

~The major remaining todo, besides review, is reorganizing the commits to be more logical.~


pkgs/racket-doc/scribblings/getting-started/getting-started.scrbl
```diff
@@ -1,5 +1,5 @@
 #lang scribble/doc
-@(require scribble/manual)
+@(require scribble/manual "../guide/guide-utils.rkt")
```
mflatt, line 2: This worries me, because I don't think of "getting-started" as building on the implementation of "guide", and someone modifying "guide-utils.rkt" probably would not expect changes to affect "getting-started" or other documents. Is there a better way to organization whatever needs to be shared?

samth: I pushed a change duplicating the definition of `Quick` from `guide-utils.rkt`.

There's another duplication of the `keep` function from `scribblings/quick/keep` in this commit; sharing it seemed like more trouble than needed here.

pkgs/racket-doc/scribblings/reference/numbers.scrbl
```diff
@@ -6,8 +6,7 @@
                      racket/unsafe/ops
                      racket/require
                      racket/random
-                     racket/list
-                     math/flonum))
+                     racket/list))
```
samth, line 9: This `require` of `math/flonum` seemed to have no uses, or at least I couldn't find them.

samth: Ah, it's for `fllogb`

samth:
This is now significantly improved, and reduces the dependencies of "racket-doc" by a lot -- with one problem. The Guide section of futures still relies fundamentally on the "futures-visualizer" package. I don't know how to fix that, yet.

rmculpepper:
Can you use `make-log-based-eval` for the futures examples?

samth:
That works for evaluation, but not for linking identifiers to their documentation.

rmculpepper:
Github needs an emoji for "oh yeah, good point..." :)

Maybe fake the effect of require `for-label` using `make-element-id-transformer` and a link-element with an indirect link?

samth:
Unfortunately, indirect links work for sections and module names but not identifiers.

sorawee:
Is it worth doing the "best effort" uncontroversial reduction? That is, it's OK to keep some dependencies that are use `for-label` for now. Just focus on reducing dependencies needed by `racketmodname` and `other-manual`.

Personally, I think any reduction at all is better than a fully featured PR that will take years to get merged.

samth:
I'm skeptical that just doing that will actually fix the problem. But if you try it out and it works then I'm happy to help finish it.

samth:
This is now ready for further review. I've added a more informative top comment. cc @rmculpepper @mflatt @rfindler

samth:
One potential issue: If we don't do https://github.com/racket/db/pull/30 because we want to fix the `db` docs have the correct links to the `bit-vector?` documentation, then we need a different approach there. Probably it will require splitting the `data/enumerate` docs from the rest of the `data/*` docs, but we can discuss this on that PR.

samth:
Since @sorawee asked about this:

To make `racket-doc` not depend on `typed-racket-lib`, there are two needed changes. One is removing types for `scribble/blueboxes`. This is easy, see https://github.com/samth/scribble/commit/ad74d018e978990d3903e688d9bf80575d1c4299. 

The other is dealing with the following chain:
```
racket-doc -> xrepl-doc -> macro-debugger -> images-lib -> typed-racket-lib
```

The Guide (in `racket-doc`) references identifiers from `xrepl` (thus `xrepl-doc`). The XRepl manual references identifiers from `macro-debugger-text-lib` but those are documented with everything else in `macro-debugger`. The macro debugger tool uses an icon from `images` (in `images-lib`). The images collection is entirely written in Typed Racket.

The "simplest" way to break this is to split the macro debugger documentation in half, but that's kinda ugly.

rfindler:
Can you explain more about the introduced dependency errors?

samth:
@rfindler Which ones? There should be some caused by the new "scheme-doc" package, as discussed above and fixed by the various linked PRs. There shouldn't be any others, although I might have failed to submit an important PR -- it's hard to check where all the changes are on my file system.

rfindler:
I'm asking because of the top comment is all. I didn't get what you meant by (b).

samth:
Ah, ok. Here's some more explanation. The change to add "scheme-doc" and not make it a dependency of "racket-doc" means that anyone who has some documentation that uses `#lang scheme`  (for example) in an example in their documentation but currently depends only on "racket-doc" will have a new missing dependency on "scheme-doc". 

The necessary fixes for packages in the main distribution are already done in the form of pull requests that you can see linked from this one. For other packages, I think there won't be that many (most people use `#lang racket`). Furthermore, my experience has been that people tend not to care about dependency errors as reported by https://pkg-build.racket-lang.org, even when they can be fixed by just running `raco setup --fix-pkg-deps`. The current status of most recently-updated packages having few dependency errors is because I've automatically submitted PRs to fix those errors for many many packages. I will eventually do that for these packages too.

Plus the missing dependency will (a) not affect docs.racket-lang.org, (b) not affect anyone who starts from a regular distribution (c) in the worst case, result in some red underlines in documentation, not a broken build or application.

rfindler:
Thank you! Makes a lot of sense to me.

greghendershott:
:tada:

PR #3220, opened 2020-05-28
Improve Windows CI Testing

pmatos:


pmatos:
This also fixes a problem with the previous testing whereby we were building CS but testing 3m, so we never actually tested both 3m and CS with Actions. This fixes that and does the build/test for both pull requests and pushes.

PR #3221, opened 2020-05-28
Updated sgn documentation to reveal behaviour for signed zeros.

sanchom:
This behaviour surprised me! But, it's how it is intended to work, based on the tests for the `sgn` function.


sorawee:
Can `+inf.0` and `-inf.0` also be added to the example? It could be considered "not a number", so it would be clarify this in the example.

sanchom:
Done!

PR #3228, opened 2020-06-02
set path guidance

spdegabrielle:
resolve https://github.com/racket/racket-lang-org/issues/101 ?

### DIFF
--- a/pkgs/racket-doc/scribblings/more/more.scrbl
+++ b/pkgs/racket-doc/scribblings/more/more.scrbl
@@ -95,8 +95,7 @@ start @exec{racket} with no command-line arguments:
   > 
 }
 
-@margin-note*{Set your @tt{PATH} environment variable so you can use raco and other Racket command line functions. On macOS:@tt{sudo sh -c 'echo "/Applications/Racket v7.7/bin" >> /etc/paths.d/racket'} and 
- @tt{sudo xattr -r -d com.apple.quarantine /Applications/Racket\ v7.7/} to use command line tools (@tt{raco}). On Windows one of the following add the racket @tt{bin} path to @tt{Path} in 'Environment Variables' (under System Properties, Advanced tab)}
+@margin-note*{Set your @tt{PATH} environment variable so you can use raco and other Racket command line functions. On macOS:@tt{sudo sh -c 'echo "/Applications/Racket v7.7/bin" >> /etc/paths.d/racket'}. On Windows add the racket @tt{bin} path to @tt{Path} in 'Environment Variables' (under System Properties, Advanced tab)}
 
 Assuming that you have Editline installed,@margin-note*{To use GNU
 Readline instead of Editline, set the @envvar{PLT_READLINE_LIB}
--- a/pkgs/racket-doc/scribblings/more/more.scrbl
+++ b/pkgs/racket-doc/scribblings/more/more.scrbl
@@ -96,12 +96,12 @@ start @exec{racket} with no command-line arguments:
 }
 
 @margin-note{Set your @tt{PATH} environment variable so you can use
- @exec{raco} and other Racket command line functions. On MacOS:
+ @exec{raco} and other Racket command line functions. On Mac OS:
  @tt{sudo sh -c @literal{'}echo "/Applications/Racket v@version{}/bin" 
  >> /etc/paths.d/racket@literal{'}} (This assumes you have installed
  Racket in the @tt{Applications} folder). On Windows: add the racket
- @tt{bin} path to @tt{Path} in `Environment Variables' (under System
- Properties, Advanced tab)}
+ @tt{bin} path to @onscreen{Path} in `@onscreen{Environment Variables}'
+ (under @onscreen{System Properties}, @onscreen{Advanced} tab)}
  
 Assuming that you have Editline installed,@margin-note*{To use GNU
 Readline instead of Editline, set the @envvar{PLT_READLINE_LIB}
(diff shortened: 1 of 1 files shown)


spdegabrielle:
Please squash commits

pkgs/racket-doc/scribblings/more/more.scrbl
```diff
@@ -95,6 +95,8 @@ start @exec{racket} with no command-line arguments:
   > 
 }
 
+@margin-note*{Set your @tt{PATH} environment variable so you can use raco and other Racket command line functions. On macOS:@tt{sudo sh -c 'echo "/Applications/Racket v7.7/bin" >> /etc/paths.d/racket'}. On Windows add the racket @tt{bin} path to @tt{Path} in 'Environment Variables' (under System Properties, Advanced tab)}
```
sorawee, line 98: 1. I think it's better to use `margin-note` rather than `margin-note*`
2. Use `@exec{raco}`.
3. I think Racket people prefer MacOS over macOS
4. Add a space after `macOS:`
5. Add a `:` after `Windows`.
6. Users might not know that they need to change `7.7` to the current version of Racket. We also shouldn't hard code a version number since we need to change it every release. I'm not sure what placeholder we should put there.

samth: You can produce the current version number with the `version` function.

sorawee: I was hoping for something like `echo "/Applications/Racket v<racket version number>/bin"` with a different font for `<racket version number>` to make it clear that it's a placeholder that users need to fill in. But I guess the `version` function is a simpler solution.

rfindler: Perhaps also worth noting: the program isn't guaranteed to be in /Applications, even if people installed one of the standard releases. They are encouraged to drag it there, but they might drag it to the Desktop or they might just run it on the mounted dmg (I know the latter is a real thing as there was a bug that affected only people like that and we got reports....took a while to figure out that the dmg thing was relevant!)

spdegabrielle: Made all suggested changes. Please review and let me know if you see any issues.

spdegabrielle:
Please *Squash* if you are happy.

pkgs/racket-doc/scribblings/more/more.scrbl
```diff
@@ -95,6 +95,8 @@ start @exec{racket} with no command-line arguments:
   > 
 }
 
+@margin-note{Set your @tt{PATH} environment variable so you can use @exec{raco} and other Racket command line functions. On MacOS: @tt{sudo sh -c 'echo "/Applications/Racket v@version{}/bin" >> /etc/paths.d/racket'} (This assumes you have installed Racket in the @tt{Applications} folder). On Windows: add the racket @tt{bin} path to @tt{Path} in 'Environment Variables' (under System Properties, Advanced tab)}
```
pmatos, line 98: Should we wrap this?

spdegabrielle: what is the etiquette here? Can I click 'Resolve conversation' or does it need to be a repo owner?

spdegabrielle: image

spdegabrielle:
Hi

I think I have addressed comments by @sorawee @samth and @rfindler 

Please let me know if there are any issues blocking this PR.

image

sorawee:
The single quote is supposed to be straight, but it's not. If anyone copies the command and uses it without correction, it won't work. I think you can use [`literal`](https://docs.racket-lang.org/scribble/base.html?q=literal#%28def._%28%28lib._scribble%2Fbase..rkt%29._literal%29%29) to fix the problem.

pkgs/racket-doc/scribblings/more/more.scrbl
sorawee: I think what @pmatos meant is that the Scribble source should be broken up to several lines, not one long line.

spdegabrielle:
Thanks @sorawee that fixed the problem; 
image

OK now?

sorawee:
Sorry, I just saw that there's also another wrong shape of quote at `Environment Variables`. The opening one should be backtick.

pkgs/racket-doc/scribblings/more/more.scrbl
spdegabrielle: wrap fixed
image

still renders OK
image

spdegabrielle:
wrap in code fixed.
image

spdegabrielle:
> Sorry, I just saw that there's also another wrong shape of quote at `Environment Variables`. The opening one should be backtick.

I have revitalised with a cuppa and a jammy dodger 
IMG_0752
and changed the quote to `
image

mflatt:
A few more tweaks? "MacOS" should be "Mac OS" (at least by the Racket documentation rules). For the Windows part, I'd use `@onscreen` for "Path", "Environment Variables", "System Properties", and "Advanced". (But if you're tired of all these little changes, I'd be happy to make adjustments after merging.)

spdegabrielle:
No problem. How does this look?

image

mflatt:
The quotes around "Environment Variables" now seem redundant to me.

Somewhat more substantively: Do we want to recommend `>>` or `>` for `/etc/paths.d/racket`? It seems like `>` has the advantage that it will replace a path for an older installation, but maybe you had something specific in mind with `>>`.

sorawee:
I think `>>` is better. It doesn't delete any data.

spdegabrielle:
> The quotes around "Environment Variables" now seem redundant to me.
fixed 

I would use `>` but I followed the example in Beautiful Racket.

image

spdegabrielle:
https://beautifulracket.com/setting-the-mac-os-path.html

mflatt:
Ok - unless there are further comments, I'll merge this soon.

PR #3230, opened 2020-06-03
Consolidate current racket-ci images into a single one

pmatos:
This simplifies jobs (like PR jobs that have) to build and test in a
single container and cannot choose between build or test images.


pmatos:
Expecting tests to fail because image doesn't exist yet. 
@samth if you are happy with the change, I will create the image for a first time manually so that we can see the tests pass once.

pmatos:
Thanks @samth . Initial image pushed manually and jobs restarted.

pmatos:
All green - merging.

PR #3233, opened 2020-06-04
Improve docs on the return value of git-checkout

pmatos:


pkgs/net-doc/net/scribblings/git-checkout.scrbl
```diff
@@ -47,7 +47,8 @@ to download the repository whose name on the server is
 @racket[repository] (normally ending in @filepath{.git}).  The tree
 within the repository that is identified by @racket[ref] (which can be
 a branch, tag, commit ID, or tree ID) is extracted to
-@racket[dest-dir], and the result id an ID corresponding to @racket[ref].
+@racket[dest-dir], and it returns a string containing a commit id corresponding
```
mflatt, line 50: Nit: "ID" instead of "id" for consistency.

mflatt, line 50: Nit: "ID" instead of "id" for consistency.

pmatos: Sure.

PR #3240, opened 2020-06-07
Fixes #3236: preserve disappeared-use on local, block, and class

### LINKED ISSUE #3236: Missing disappeared-uses for block?
```
#lang racket

(require racket/block)

(define-syntax-rule (define2 x v) (define-values (x) v))

(block
 (define2 x 1)
 x)
```

The above code doesn't have any arrow for `define2`.

sorawee:


rfindler:
Does this improve the check syntax arrows for class? If so, shall I add some tests to drracket, perhaps like the ones in [this directory](https://github.com/racket/drracket/tree/master/drracket-tool-test/tests/check-syntax)?

sorawee:
Yes, it improves the check syntax arrows for forms that expand to `define-syntaxes` inside a `class`.

```
(class object%
  (super-new)

  (define-syntax (mac stx)
    #'1))
```

currently doesn't show an arrow from `define-syntax` to `#lang racket`, but with this PR, it does.

rfindler:
Okay, I'm happy to add the test cases there if there isn't a better place.

sorawee:
That would be really great, thanks!

rfindler:
Okay. I'm not savvy enough to coordinate pull requests across repos, but here's a diff of what would currently be a failing test and will probably pass after this is merged.

```
diff --git a/drracket-tool-test/tests/check-syntax/syncheck-direct.rkt b/drracket-tool-test/tests/check-syntax/syncheck-direct.rkt
index 3a1aeeb9..e95015fb 100644
--- a/drracket-tool-test/tests/check-syntax/syncheck-direct.rkt
+++ b/drracket-tool-test/tests/check-syntax/syncheck-direct.rkt
@@ -223,7 +223,17 @@
                    '((64 68 0.5 0.5) (125 130 0.5 0.5))
                    '((99 110 0.5 0.5) (131 134 0.5 0.5))))
 
-
+(check-equal? (get-binding-arrows
+               (string-append
+                "#lang racket\n"
+                "(class object%\n"
+                "  (super-new)\n"
+                "  (define-syntax (mac stx)\n"
+                "    #'1))"))
+              (set '((6 12) (14 19))
+                   '((6 12) (20 27))
+                   '((6 12) (31 40))
+                   '((6 12) (45 58))))
 
 ;                                                       
 ;                                                       
```

sorawee:
I think we need to let the `drracket` repo fails momentarily and then fix it with your patch.

mflatt:
This is ok as far as I can tell, so I'll merge it.

PR #3242, opened 2020-06-08
fix #3239

### LINKED ISSUE #3239: [CS] result of struct-type-make-predicate claims to be constructor
On Racket CS, the result of `struct-type-make-predicate` claims to be a constructor rather than a predicate:
```racket
#lang racket
(struct posn (x y)
  #:transparent)
(define pred (struct-type-make-predicate struct:posn))
(struct-predicate-procedure? pred) ; -> #f, should be #t
(struct-constructor-procedure? pred) ; -> #t, should be #f
```
The problem seems to be here: https://github.com/racket/racket/blob/05dfd85853a12c90cb7ff8af41687ad8c660a466/racket/src/cs/rumble/struct.ss#L939-L949

capfredf:
[CS] struct-type-make-predicate creates a value that passes
struct-predicate-procedure?


mflatt:
Thanks!

PR #3250, opened 2020-06-14
add ssl-channel-binding

rmculpepper:
This adds an `ssl-channel-binding` operation on ssl ports that gets the channel binding data described by https://tools.ietf.org/html/rfc5929. This is needed for SCRAM-SHA-256-PLUS authentication, which is available in PostgreSQL 11 and later. (I have changes ready for the sasl and db repos.)

### DIFF
--- a/pkgs/racket-doc/openssl/openssl.scrbl
+++ b/pkgs/racket-doc/openssl/openssl.scrbl
@@ -850,7 +850,8 @@ channel binding for TLS is described in
 
 If the channel binding cannot be retrieved (for example, if the
 connection is closed), an exception is raised.
-}
+
+@history[#:added "7.7.0.9"]}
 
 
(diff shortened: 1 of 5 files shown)


mflatt:
The documentation needs a `history` note.

Are the any OpenSSL version issues here, or will the newly referenced functions be available in any OpenSSL version that was viable before?

rmculpepper:
Thanks, I'll add the history note.

This code works on OpenSSL 1.0.2 and later. On older OpenSSL versions, the new function fails with an "implementation not found" error, which seems reasonable given that 1.0.1 hit EOL at the end of 2016.

mflatt:
Ok — the fact that it's only an error if the missing function is needed (I forgot about `make-not-available`) sounds fine to me.

PR #3252, opened 2020-06-16
Add more kw source locations.

samth:


mflatt:
This is ok as far as I can tell.

PR #3260, opened 2020-06-21
Force expression context for `local` and `block`

michaelballantyne:
Also adds a helper `make-expression-transformer` to `syntax/transformer` and refactors `block` to use `racket/base` and `syntax/` dependencies.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/context.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/context.scrbl
@@ -13,6 +13,11 @@ Returns a list suitable for use as a context argument to
 expansion. The context list builds on @racket[(syntax-local-context)]
 if it is a list.}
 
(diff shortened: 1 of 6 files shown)


mflatt:
This seems good, but needs tests for `block` and `local` to make sure they now work in non-`'expression` positions. The current `block` tests are in "macro.rktl". I don't find any `racket/local` tests, but "macro.rktl" would be ok (or maybe "syntax.rktl" or some other place would be better).

mfelleisen:
There are some very basic tests of `local` in 

  ` htdp/htdp-test/tests/htdp-lang/`

Let’s make sure they work, too.

michaelballantyne:
Tests added, and `htdp-test/tests/htdp-lang/test-htdp.rktl` passed.

mflatt:
Thanks!

PR #3267, opened 2020-06-27
Add prop:struct-field-info

sorawee:
This PR adds `prop:struct-field-info` which is implemented to provide
static information about field names. The property is attached to all
struct types generated by `define-struct`.

The PR also modifies kernstructs to have the property.
Finally, the PR switches `struct-copy` to use the static field name
information when it's available.

It remains to change `contract-out` and `match`'s `struct*` to
recognize/attach this new property, but this could be done
separately in the future.

This PR is largely based on #732, though the approach is slightly
different.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/struct.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/struct.scrbl
@@ -939,7 +939,7 @@ The list should contain every immediate field name
 (that is, not including fields from its super struct type)
 in the reverse order.
 
-@examples[
+@examples[#:escape no-escape
 #:eval struct-eval
 (struct foo (x))
 (struct bar foo (y z))
--- a/pkgs/racket-doc/scribblings/reference/struct.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/struct.scrbl
@@ -789,7 +789,8 @@ derived from @racket[struct:struct-info] or with the
 @racket[prop:struct-info] property that also implements
 @racket[prop:procedure], and where the instance is further is wrapped
 by @racket[make-set!-transformer]. In addition, the representation may
-implement the @racket[prop:struct-auto-info] property.
+implement the @racket[prop:struct-auto-info] and
+@racket[prop:struct-field-info] properties.
 
 Use @racket[struct-info?] to recognize all allowed forms of the
 information, and use @racket[extract-struct-info] to obtain a list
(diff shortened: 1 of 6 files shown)


mflatt:
Thanks! I will look more soon.

Meanwhile, do you know why #732 was never merged? It looks like we may have simply lost track, but any perspective you have (or an explanation of why your version is slightly different) would be welcome.

sorawee:
I didn't know why it didn't get merged, but right now the PR has a lot of conflicts and Dionna said that she's not "very active anymore with racket [to resolve the conflict]". She said that I'm free to base my PR on hers, however.

The differences between my PR and Dionna's are:
- I'm trying submit a PR with minimal changes. That means I'm not going to add new features like a new struct matcher for `match`. Only fixing bugs.
- The static information in Dionna's PR is in `extended-struct-info` which must be constructed by providing the existing `struct-info`. The static information in my PR is in a struct property `struct-field-info` which can be attached to existing `struct-info`. In this sense, my approach (or rather, the approach that @lexi-lambda and @AlexKnauth suggested in #2700) is less invasive.
- The field names in Dionna's PR includes those of super struct. Mine only has field names of the immediate struct. I believe mine makes more sense because field name is a concept for a level of a struct, where field names are all distinct in each level. So I don't think including field names of parent structs is a good idea. (Plus, extracting field names or number of fields in each level would become much easier. Here's how it's currently done: https://github.com/racket/racket/blob/4ad93bcca0a1b9399f029bec969011f2245548a5/racket/collects/racket/contract/private/provide.rkt#L846)

mflatt:
Thanks for the background, and thanks to @deeglaze for the original work.

These changes look good to me, and I detect no problems when I try building. So, I'm happy to merge, and I'll add a `@history[#:added "7.7.0.10"]` in the new documentation.

I think the change to `struct-copy` is meant to change the behavior of some uses of `struct-copy` — to make certain uses work at all. Some `struct-copy` tests to demonstrate would be helpful in another update.

sorawee:
> I think the change to struct-copy is meant to change the behavior of some uses of struct-copy — to make certain uses work at all. Some struct-copy tests to demonstrate would be helpful in another update.

Not really. The change to `struct-copy` is meant to be a _proper_ fix for #1399, but this was fixed already (improperly) in #2700, with tests already added.

rfindler:
Could there be a test case that would now pass but wouldn't before if the
predicate name were munged in some way? Or maybe just a test case that
demonstrates struct-copy working directly with some code that makes a
struct definition without using "struct" but using the new apis directly?

Robby

sorawee:
It might be possible. I will try.

Issue #2740, opened 2019-07-15
XDG Base Directory Support


billop:
Could support of [XDG Base Directories](https://wiki.archlinux.org/index.php/XDG_Base_Directory) be considered? It would be nice to be freedesktop.org compliant and use XDG base directories by default. The sooner this is fixed the better, since the more the incorrect behavior is assumed the more difficult it will be to fix, and racket would continue to pollute home directories.

willghatch:
For what it's worth, I've created the [basedir](https://github.com/willghatch/racket-basedir) package for Racket for package authors to easily use the XDG basedir standard, though it doesn't help with the Racket core.

I would be in favor of moving the `~/.racket` directory contents and the `~/.racketrc` into appropriate XDG directories (with fallback behavior).

lhanson:
This would be fantastic, there's no downside to respecting users' preferences.

roket1428:
Any news?

a-vrma:
workaround for now: `export PLTUSERHOME="$XDG_DATA_HOME"/racket`. You can use `~/.local/share` instead of `XDG_DATA_HOME`.

Stolen from https://wiki.archlinux.org/index.php/XDG_Base_Directory.

sorawee:
Fixed. Closing.

PR #3268, opened 2020-06-27
XDG Base Directory support

rmculpepper:
PR for discussion about changing Racket to follow the XDG Base Directory specification. See also #2740. 

Here's a summary of the changes to `find-system-path`:

    'pref-dir  = $XDG_CONFIG_HOME/racket or ~/.config/racket (was ~/.racket)
    'addon-dir = $XDG_DATA_HOME/racket or ~/.local/share/racket (was ~/.racket)
    'init-dir  = same as pref-dir (was ~/)
    'init-file = $XDG_CONFIG_HOME/racket/racketrc.rktl or ~/.config/racket/racketrc.rktl (was ~/.racketrc)

A new key is added:

    'cache-dir = $XDG_CACHE_HOME/racket or ~/.cache/racket

This directory is used as the default location for the "download-cache" directory. On Windows and Mac OS, `'cache-dir` currently returns the same path as `'addon-dir`.

These changes do not include any attempts at fallbacks or compatibility with the old `.racket` paths. One way of supporting both conventions would be symlink `~/.local/share/racket` and `~/.config/racket` (and `~/.cache/racket`, if you like) to an existing `~/.racket` directory.

Note: changes to rktio don't seem to automatically cause gracket to get recompiled, so I had to delete the racket/src/build directory and rebuilt before drracket saw the changes, for example.

### DIFF
--- a/racket/src/io/host/rktio.rkt
+++ b/racket/src/io/host/rktio.rkt
@@ -89,3 +89,5 @@
 
 ;; Only in the main place:
 (void (rktio_do_install_os_signal_handler rktio))
+
+(void 1) ;; force recompile due to ../../rktio/rktio.rktl change
--- a/pkgs/racket-doc/scribblings/raco/config.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/config.scrbl
@@ -161,8 +161,8 @@ directory}:
  @item{@indexed-racket['download-cache-dir] --- a path string used as
        the location for storing downloaded package archives. When not
        specified, packages are cached in a @filepath{download-cache}
-       directory in the user's add-on directory as reported by
-       @racket[(find-system-path 'addon-dir)].}
+       directory in the user's cache directory as reported by
+       @racket[(find-system-path 'cache-dir)].}
 
  @item{@indexed-racket['download-cache-max-files] and
        @indexed-racket['download-cache-max-bytes] --- real numbers that
--- a/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
@@ -82,8 +82,9 @@ by @racket[kind], which must be one of the following:
  if it is defined, otherwise it is the current directory.}
 
  @item{@indexed-racket['init-dir] --- the directory containing the
- initialization file used by the Racket executable.
- It is the same as the @tech{user's home directory}.}
+ initialization file used by the Racket executable.  On Unix, it is
+ the same as the result returned for @racket['pref-dir]; on Mac OS and
+ Windows, it is the same as the @tech{user's home directory}.}
 
  @item{@indexed-racket['init-file] --- the file loaded at start-up by
  the Racket executable. The directory part of the
@@ -92,9 +93,9 @@ by @racket[kind], which must be one of the following:
 
   @itemize[
 
-  @item{@|AllUnix|: @indexed-file{.racketrc}}
+  @item{Unix and Windows: @indexed-file{racketrc.rktl}}
 
-  @item{Windows: @indexed-file{racketrc.rktl}}
+  @item{Mac OS: @indexed-file{.racketrc}}
 
   ]}
 
--- a/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
@@ -52,8 +52,11 @@ by @racket[kind], which must be one of the following:
  the current executable is used as the home directory.}
 
  @item{@indexed-racket['pref-dir] --- the standard directory for
- storing the current user's preferences. On Unix, the directory is
- @filepath{.racket} in the @tech{user's home directory}.  On Windows,
+ storing the current user's preferences. On Unix, the directory is the
+ @filepath{racket} subdirectory of the path specified by
+ @indexed-envvar{XDG_CONFIG_HOME}, or @filepath{.config/racket} in the
+ @tech{user's home directory} if @envvar{XDG_CONFIG_HOME} is not set to
+ an absolute path. On Windows,
  it is @filepath{Racket} in the @tech{user's home directory} if
  determined by @envvar{PLTUSERHOME}, otherwise in the user's
  application-data folder as specified by the Windows registry; the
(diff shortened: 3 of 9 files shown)


samth:
I'm a fan of this idea generally, but I think backwards compatibility here is necessary. Unfortunately that involves search paths, but I don't see a better option.

This discussion for Rust is also interesting (they haven't ended up doing it): https://github.com/rust-lang/rfcs/pull/1615 and includes similar discussion for Windows and Mac.

mflatt:
I'm also in favor.

For backward compatibility, instead of search paths, maybe it would make more sense for DrRacket and `racket/init` and `racket/gui/init` to create the suggested filesystem links automatically.

For the MacOS `.racketrc`, then I think it makes sense to continue to match Unix (i.e., change it for MacOS, too).

For `'cache-dir` on MacOS, probably the path should be in `~/Library/Caches`.

samth:
Would the "creating links" process move the existing content?

racket/src/io/host/rktio.rkt
```diff
@@ -89,3 +89,5 @@
 
 ;; Only in the main place:
 (void (rktio_do_install_os_signal_handler rktio))
+
+(void 1) ;; force recompile due to ../../rktio/rktio.rktl change
```
samth, line 93: Could this be fixed with a Makefile change?

mflatt:
Unless I misunderstand what you mean, I don't think the old content should be moved, in case someone is also running old versions. So, "creating links" would really mean creating links.

samth:
Oh, I see that the suggestion is to create the links in the other direction.

rmculpepper:
I prefer the idea of migration rather than search path, but I don't think racket et al should do this automatically. For example, consider the case when ~/.racket exists and the user runs racket with PLTUSERHOME set to some fresh directory with the intention of starting with running a program in a clean environment and protecting their default configuration from changes. If racket automatically creates links back to ~/.racket, it defeats both goals.

What would the consequence of failing to maintain backwards compatibility? The addon directory mainly contains a cache and version-specific subdirectories, so there's no need to worry about them. The main thing that's not version-specific is the preference file (racket-prefs.rktd). So the main effect would be that the new version of DrRacket would lose the user's previous configuration. It would make sense for DrRacket to detect that case (the old pref file exists and the new one does not) and offer to copy or link preferences.

Another potential frustration would be people mistaking ~/.racket for the current location of Racket preferences, etc. Maybe DrRacket (and raco, maybe) could create ~/.racket/racket-has-moved.txt if ~/.racket exists, as a reminder. (The existence of that file could also signal that migration has been taken care of and not to ask again, even for eg a fresh PLTUSERHOME.)

mflatt:
I don't understand the `PLTUSERHOME` scenario. In that case, wouldn't the check be for `${PLTUSERHOME}/.racket` and not `~/.racket`? That is, if `PLTUSERHOME` is set and `${PLTUSERHOME}/.racket`, then links might be created to `${PLTUSERHOME}/.racket`. Or do I misunderstand?

I can see how there would be a problem if `PLTUSERHOME` is set and `XDG_CONFIG_HOME` is set to something not in `PLTUSERHOME`. Probably if `PLTUSERHOME` is set, then `XDG_CONFIG_HOME`, `XDG_DATA_HOME`, and `XDG_CACHE_HOME` should not be used.

I'm not sure of all the implications of backward incompatibility, but here's one to add to the list: `raco pkg migrate` would not find packages to migrate.

samth:
I use `PLTADDONDIR` for all of my installs and that is maintained across versions because it's built from git. So changing that behavior would be a big problem for me.

rmculpepper:
@mflatt You're right, for some reason I was thinking of .racket as not also being relative to `PLTUSERHOME`. But I think the problem does occur if `PLTUSERHOME` is not set, ~/.racket exists, and `XDG_{CONFIG,DATA}_HOME` are set to fresh directories. If the intention is isolation from existing directories, automatically creating a link defeats it.

I didn't understand the problem in the situation you described in your second paragraph, but it seems reasonable to me for a user to set all of those variables and expect them them to take precedence as they currently do in the PR. It would be surprising to ignore one variable if a more general variable was set. (Note that the XDG variables are already ignored if they do not contain absolute paths.)

I agree about `raco pkg migrate`.

@samth This PR doesn't change the behavior of the `PLTADDONDIR` environment variable and the `-A` command-line option; they still override the default addon directory.

mflatt:
After looking at this again, I think my reflex to avoid search paths may be unhelpful here.

How about this?

* If `PLTUSERHOME` is set, then `XDG...` environment variables are treated as unset.
* If `XDG...` environment variables are set, then paths based on them are used.
* If `XDG...` enviornment variables are not set and ".racket" exists, then ".racket"-based paths are used; an extant ".racketrc" similarly redirects `init-dir` and `init-file` if `XGD_CONFIG_HOME` is not set (but ".racketrc" doesn't by itself trigger uses of ".racket" paths, and ".racket" doesn't trigger the use of ".racketrc").

The idea here is that `PLTUSERHOME` continues to be the one-stop, redirect-everything environment variable for Racket. Otherwise, if you have `XDG...` variables set, then we assume that you mean them. (These don't seem to be set by default in any of the OS installations I checked.) Finally, if ".racket" or ".racketrc" exists, then we assume that you haven't yet converted — but you get modern conventions in a fresh environment.

rmculpepper:
It feels wrong to behave differently depending on whether an `XDG...` variable is set or not; I think that unset and set to default should produce the same behavior. I don't have any sources for that, though.

Checking for ".racket" is unstable; consider the following situation: A user installs a new (XDG-aware) version of Racket, sets preferences. Then they install an old version and use it enough to create "~/.racket". Then the new version suddenly "loses" its previous state.

mflatt:
Ok, I agree on both points. But I still think that an automatic fallback to the old directory structure will save trouble, so here's another attempt:

 * If `PLTUSERHOME` is set, then `XDG...` environment variables are treated as unset. The XDG defaults don't actually use `~`, but are based on a home directory that can be redirected by `PLTUSERHOME`.

 * The preferences directory is

     - `$XDG_CONFIG_HOME/racket` if it exists or if `.racket` in the home directory does not exist,
     - `.racket` in the home directory if it exists and if `$XDG_CONFIG_HOME/racket` does not exist

   and similar for each file or directory, independently

The idea is still that `PLTUSERHOME` continues to be the one-stop, redirect-everything environment variable for Racket. Meanwhile, XDG paths are unless only old-style paths exist; if only old-style paths exist, we assume that you have not converted and want to keep the old structure.

mflatt:
Just to avoid duplicating effort: I have a draft implementation of the above fallback compatibility rules. So far, so good, but I need to check more.

mflatt:
The branch https://github.com/mflatt/racket/tree/xdg has an implementation of the suggested fallback. The relevant tests are at https://github.com/mflatt/racket/blob/xdg/pkgs/racket-test-core/tests/racket/file.rktl#L1847

rmculpepper:
Okay, that seems reasonable to me.

mflatt:
Ok, good — I'll merge and we'll see how it goes.

mflatt:
Just for the record: The change to separate `'cache-dir` from `'addon-dir` broke a `raco pkg` test, because tests that were intended to be separated by `PLTADDONDIR` ended up sharing a cache. I think that's ok, since `raco pkg` changed to use `'cache-dir`. Adjusting the test suite to use `PLTUSERHOME` is easy.

sorawee:
This is already merged. Closing.

PR #3274, opened 2020-06-28
sxml doc comment

jbclements:
proposed margin note re: sxml

### DIFF
--- a/pkgs/racket-doc/xml/xml.scrbl
+++ b/pkgs/racket-doc/xml/xml.scrbl
@@ -30,6 +30,11 @@ Declaration (DTD) processing, including preservation of DTDs in read documents, 
 It also does not expand user-defined entities or read user-defined entities in attributes.
 It does not interpret namespaces either.
 
+@margin-note{In addition to the library described by this document, there is another
+ @racket[sxml] library based on work by Oleg Kiselyov
+ which users may find faster in processing large documents. It is
+ available through racket's package server.}
+
 @; ----------------------------------------------------------------------
 
 @section{Datatypes}


samth:
I think a note about performance is a mistake. 

Also, this should use `@racketmodname` with the indirect option.

jbclements:
My programmatic intuition is that repairing the performance of the built-in xml reading is not going to be easy. Do you think I'm mistaken?

samth:
My intuition is that if sxml can perform well, then the xml library can perform equally well -- they do basically the same thing.

samth:
@alex-hhh pointed out on the mailing list that the `xml-count-bytes` option can significantly improve parsing speed, and mentioning that in the doc comment seems sensible. Maybe a parameter to not count anything at all would be even more helpful.

alex-hhh:
Reading the `xml-count-bytes` documentation, it is misleading at least to me:

> A parameter that determines whether read-xml counts characters or bytes in its location tracking. The default is #f, which counts characters.

This is not what it does.  When `#f`, the xml reader simply calls `port-count-lines!` on the input port.  Note that someone else might have already called `port-count-lines!`  previously, or the port might have been created with `port-count-lines-enabled` set to `#t`.

Regardless of what the `xml-count-bytes` setting is, location is always recorded for all elements by calling `port-next-location` which always returns the line, column and position of the port.  Presumably the line and column will be zero when  line counting is disabled, and it is unclear if position counts bytes or characters, but it is not affected by the `port-count-lines!` as far as I understand.

So, to my best understanding from reading the code and associated documentation, location is always recorded, as follows:

* line and column numbers will be 0, positions counts characters when `xml-count-bytes` is set to `#t`
* line, column and positions (which still counts characters) are recorded when `xml-count-bytes` is set to `#f` (default)

samth:
I agree that this is misleading as-is. How about something like:

> A parameter which determines whether line, column, and character positions are recorded in `@racket[location]` values. If the value is `#f`, the default, then all 3 kinds of positions are recorded (enabled by initially calling `port-count-lines!` on the port passed to `read-xml` and other reading functions). If the value is a true value, then line and column positions are not tracked, and the  offset field of `location` elements records the position of the element in bytes. 
>
> Note that setting `xml-count-bytes` to `#t` can significantly increase parsing speed.

rfindler:
If setting the parameter to #t causes the library to call the
(side-effecting operation) port-count-lines!, then that should probably
also be mentioned.

Robby

samth:
Setting it to `#f` (the default) causes `port-count-lines!` to be called; if the value is `#t` then `port-count-lines!` is not called. I tried to make that clear in what I wrote -- is there something you'd prefer in the suggested text?

rfindler:
Oops, no. I just need to learn to read, apparently.

alex-hhh:
Given that the port no longer counts bytes for any value of `xml-count-bytes`, would it not be possible to make it do nothing, document it as deprecated and introduce a new parameter, `xml-track-line-locations` which is documented what it does and has a "fast option" (i.e. not track lines) by default.  I am not sure how many people need the location information in the XML documents they parse.

samth:
I believe that when `xml-count-bytes` is true, there will still be numbers in the `offset` field of the `location` structs, and they will correspond to byte positions in the input file (as opposed to UTF-8 character positions, for example). That information is available even when `port-count-lines!` is not called. 

I agree that the name and default value are confusing, and it would be better to have better defaults. One other possibility would be to have a function `read-xexpr` which simply skipped the intermediate XML data structures and produced xexprs directly. I expect this would be more competitive with the `sxml` operations as well.

jbclements:
Closing this pull request, it appears unpopular.

PR #3278, opened 2020-06-29
Fix contract-out for struct

### LINKED ISSUE #3266: Failure on multi-layer contracted struct with parent
Problem occurs when attaching contracts to a struct with a parent that goes through multiple `contract-out` layers.
```rkt
#lang racket

(module first racket
  (provide (contract-out (struct foo ([x any/c])))
           (contract-out (struct (bar foo) ([x any/c]))))
  (struct foo (x))
  (struct bar foo ()))

(module second racket
  (require (submod ".." first))
  (provide (contract-out (struct foo ([x any/c])))
           (contract-out (struct (bar foo) ([x any/c])))))

(require 'second)
(provide (contract-out (struct (bar foo) ([x any/c]))))
```

Error (on 7.7 BC):
```
substring: starting index is out of range
  starting index: 24
  valid range: [0, 5]
  string: "foo-x"
```

### LINKED ISSUE #3269: Can't use contract-out on struct type whose super type has the same field name
```
#lang racket 

(struct foo (x))
(struct bar foo (x))
```

is a valid code. Therefore, we should be able to `contract-out` these structs. However:

```
#lang racket 

(struct foo (x))
(struct bar foo (x))
(provide (contract-out (struct (bar foo) ([x any/c] [x any/c]))))
```

results in:

```
racket/collects/racket/contract/private/provide.rkt:974:45: module: identifier already defined
  at: provide/contract-field-contract-x-bar
  in: (define-values (provide/contract-field-contract-x-bar) (coerce-contract (quote provide/contract) any/c))
```

### LINKED ISSUE #3271: #:omit-constructor can be repeated
```
#lang racket

(provide (contract-out (struct test () #:omit-constructor #:omit-constructor)))

(struct test ())
```

probably should result in a syntax error.

### LINKED ISSUE #3272: super type in struct's contract-out is unchecked
```
#lang racket

(module first racket
  (provide (contract-out (struct (bar baz) ([x any/c]))))
  (struct baz (x))
  (struct foo (x))
  (struct bar foo ()))

(require 'first)
(bar 1)
```

should result in an error because `baz` is not a super type of `bar`.

sorawee:
- A part of contract-out's code generation for struct assumes that
there's no parent struct and uses the provided struct name for
everything. This causes duplicate definitions when there are duplicate
field names where one is in a child struct and another is
in a parent struct. This PR fixes the problem.
- Disallow multiple #:omit-constructor
- Deprecate super-id. This information is unnecessary since we can
extract it from static struct information already. Attempting to
check that super-id is well-formed is error-prone due to how
the super struct type could be contracted which shields us from
detecting that they are indeed the super type.
- Utilize static struct field name information, and provide
the information when exporting a struct.

This PR is largely based on #732.

Fixes #3266
Fixes #3269 
Fixes #3271 
Fixes #3272

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1838,7 +1838,7 @@ earlier fields.}}
   (code:line)
   (code:line #:unprotected-submodule submodule-name)]
  [contract-out-item
-  (struct id/super ((id contract-expr) ...)
+  (struct id/ignored ((id contract-expr) ...)
     struct-option)
   (rename orig-id id contract-expr)
   (id contract-expr)
@@ -1847,8 +1847,8 @@ earlier fields.}}
   (code:line #:∀ poly-variables)
   (code:line #:forall poly-variables)]
  [poly-variables id (id ...)]
- [id/super id
-           (id super-id)]
+ [id/ignored id
+             (id ignored-id)]
  [struct-option (code:line)
                 #:omit-constructor])]{
 
@@ -1876,13 +1876,8 @@ first variable (the internal name) with the name specified by the
 second variable (the external name).
 
 The @racket[struct] form of @racket[contract-out]
-provides a structure-type definition, and each field has a contract
-that dictates the contents of the fields. The structure-type
-definition must appear before the @racket[provide] clause within the
-enclosing module. If the structure type has a parent, the second
-@racket[struct] form (above) must be used, with the first name
-referring to the structure type to export and the second name
-referring to the parent structure type. Unlike a @racket[struct]
+provides a structure-type definition @racket[id], and each field has a contract
+that dictates the contents of the fields. Unlike a @racket[struct]
 definition, however, all of the fields (and their contracts) must be
 listed. The contract on the fields that the sub-struct shares with its
 parent are only used in the contract for the sub-struct's constructor, and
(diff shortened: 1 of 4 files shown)


pkgs/racket-doc/scribblings/reference/contracts.scrbl
```diff
@@ -1918,7 +1916,8 @@ is bound to vectors of two elements, the exported identifier and a
 syntax object for the expression that produces the contract controlling
 the export.
 
-@history[#:changed "7.3.0.3" @list{Added @racket[#:unprotected-submodule].}]
+@history[#:changed "7.3.0.3" @list{Added @racket[#:unprotected-submodule].}
+         #:changed "7.7.0.9" @list{Deprecated @racket[ignored-id].}]
```
rfindler, line 1920: Maybe "started ignoring @racket[ignored-id]" instead of "deprecated"

racket/collects/racket/contract/private/provide.rkt
```diff
@@ -731,7 +762,7 @@
                 (and (identifier? (syntax name))
                      (identifier? (syntax super)))
                 #t]
-               [else #f])))
+               [_ #f])))
```
rfindler, line 765: oh, thanks!

rfindler:
Assuming the tests all pass, I'm good to merge this. It would be good to turn all of the nice examples in the various other (linked) examples into the test suite. I can do that separately, however.

sorawee:
I include tests from various issues here: https://github.com/racket/racket/pull/3278/files#diff-efd3bb96dffdd927e6f82c30c27ea2dd. Is this what you want?

pkgs/racket-doc/scribblings/reference/contracts.scrbl
sorawee: Should I change the prose as well? I use the word "deprecate" there too -- I mean, I'm not clear what do you want to do.

rfindler: Deprecate is a fine word for the current (post-PR) state, but thing that changed is that ignored-id is now ignored. So I suggest changing only the history annotation.

It is a minor point for sure :)

rfindler:
> I include tests from various issues here: https://github.com/racket/racket/pull/3278/files#diff-efd3bb96dffdd927e6f82c30c27ea2dd. Is this what you want?

Oh, sorry! I missed that. Looks great to me.

rfindler:
Okay, the tests all pass. Thanks!

PR #3281, opened 2020-07-02
rktio: add rktio_sendfile and tcp-sendfile primitives

Bogdanp:
There's quite a bit left to be done here, but I wanted to get some feedback before finishing this. Does this seem like a reasonable approach? 

I'm limiting this to TCP sockets because that's what most platforms support. With this change in place, the web server's (and other similar software) file serving performance could be improved.


racket/src/io/network/tcp-port.rkt
```diff
@@ -92,3 +93,22 @@
...
      (close-port p)]))
+
+(define/who (tcp-sendfile dst-p src-p [offset 0] [nbytes 0])
+  (define op (->core-output-port dst-p #:default #f))
+  (unless (tcp-output-port? op)
+    (raise-argument-error who "(and/c tcp-port? output-port?)" dst-p))
+  (define ip (->core-input-port src-p))
+  (unless (file-stream-port? ip)
+    (raise-argument-error who "(and/c file-stream-port? input-port?)" src-p))
+  (unless (>= offset 0)
```
mflatt, line 104: This doesn't check that `offset` is an exact integer. Probably it's best to use `exact-nonnegative-integer?` here. Even better, convert to use the `check` form. Also, you'll need to do something about bignums before trying to send them to `rktio_sendfile` — maybe looping over fixnum-sized chunks or only writing a fixnum-sized check and letting the caller loop.

racket/src/racket/src/network.c
```diff
@@ -1457,6 +1459,81 @@ void scheme_tcp_abandon_port(Scheme_Object *port) {
...
+
+      if (SCHEME_INPUT_PORTP(argv[1])) {
+        ip = scheme_input_port_record(argv[1]);
+        if (SCHEME_FALSEP(scheme_file_stream_port_p(1, &ip))) {
+          scheme_wrong_contract("tcp-sendfile", "(and/c file-stream-port? input-port?)", 1, argc, argv);
+          return NULL;
+        }
+
+        if (argc > 2) {
+          if (SCHEME_INTP(argv[2])) {
```
mflatt, line 1486: `SCHEME_INTP` recognizes only fixnums, but `exact-nonnegative-integer?` includes bignums. Figuring out what to do with bignums is always a pain, but that's something to work out here, too.

mflatt:
Less important than other things you're still working on here, but FWIW: I'm ambivalent about the name `tcp-sendfile`. It's helpful to programmers who know about the `sendfile` system call, but it's not a great description of of the function (i.e., not the name you'd pick from scratch).

Bogdanp:
Thanks for the feedback! 

Regarding the contracts, I think I'll go with a non-negative fixnum for now and side-step the bignum issue. Most (all?) file systems have file size limits below the max fixnum for the systems they run on so this is a reasonable constraint.

Regarding the name, I think it probably makes sense to keep `rktio_sendfile` at the rktio level, but I agree that it should be called something else at the Racket level. How about `tcp-copy-file` or `tcp-write-file`?

mflatt:
Dealing with non-fixnums is a pain, but functions constrained to fixnums are a pain for clients. It's not difficult to imagine a program that tries to send 1GB, for example, which will work until it's moved to a Racket CS on a 32-bit platform. An interface that lets you truncate a bignum to a fixnum (by returning the number of sent bytes) is probably a better way to go.

Things with "file" in the name usually take path arguments. Maybe `tcp-copy-port`?

Bogdanp:
I've changed Racket-level API so that the function, now named `tcp-copy-file`, takes a TCP port, a file path and start and end offsets. I think this slightly simplifies things for the user compared to the original approach and it has the added benefit that we don't have to worry about preserving offsets after issuing the system calls.

Regarding the offsets, I agree that we need to allow values larger than a fixnum. I hadn't realized when I wrote my last comment that fixnums were so constrained on 32bit CS. However, we are limited by the platforms here. On Windows, the `OVERLAPPED` structures permit us to use unsigned 64bit offsets on all systems, but the other platforms use `off_t` values for the offsets, which are signed word-sized integers so on 32bit systems we're still limited to files smaller than 2^31. I believe the approach I've taken in the CS implementation of `tcp-copy-file` does the right thing in that it'll allow values larger than a fixnum but only up to the max signed integer for the platform and chez's FFI takes care of converting the bignum values to `intptr_t`s, but please let me know if I have that wrong.

Assuming this still seems like a good idea, I think what's left to do is:

* rktio: implement sendfile emulation 
* rktio: add ac checks for the system calls and fall back to emulation when not available
* BC: implement `tcp-copy-file`
* CS & BC: add tests
* document the new function

racket/src/io/network/tcp-port.rkt
```diff
@@ -92,3 +95,52 @@
...
+       (close-input-port ip)
+       (send tcp-output-port op raise-write-error sent)]
+
+      [(zero? (- nbytes sent))
+       (rktio_sendfile_status_free rktio status)
+       (close-input-port ip)]
+
+      [else
+       ;; TODO(bogdan): Need to allow breaks?
+       (sync op)
```
Bogdanp, line 145: Is it OK to do something like `((if (break-enabled) sync/enable-break sync) op)` at this level? Can I then wrap that in an exception handler to free resources when a break is raised? I think the answer to both of those is "yes", but I figured it'd be best to ask because I haven't noticed any other code in `src/io` do that.

racket/src/io/network/tcp-port.rkt
```diff
@@ -92,3 +95,52 @@
...
      (close-port p)]))
+
+(define max-file-offset-for-platform
+  (case (system-type 'word)
+    [(32) (sub1 (expt 2 31))]
+    [(64) (sub1 (expt 2 63))]))
+
+(define/who (tcp-copy-file dst-p path
+                           [start 0]
+                           [end (file-size path)])
```
yjqww6, line 106: Using file-size seems to have race conditions, would it be better to use #f to indicate sending full file?

Bogdanp: Yes, that does seem like a better approach. I'll do that.

racket/src/io/network/tcp-port.rkt
mflatt: It's not ok to allocate a primitive resource, like `rktio_make_sendfile_status`, and then allow breaking without freeing the resource. Probably a sendfile-status obbject will need to be registered with the current custodian.

More generally, having a `sync` in the middle of a function suggests that it's not the right core function, and that function producing an evt is the right idea instead.

mflatt:
I'm skeptical of the change to take a fliename instead of a port, because tangles one complicated and fraught process (mapping names to files) with another complicated process (sending bytes through TCP). As a concrete example of the difference, having a port means that a file could meanwhile be safely deleted or renamed while the send is in progress, which is often a useful pattern (on competent filesystems). At the same time, I see your point about not having to update the original port's offset, though.

The right API may be to more directly imitate `sendfile` on non-blocking output: only write bytes that can be immediately written, and return the number of written bytes. Then, there are no blocking operations, which makes resource management easier. And if the fast version only applies to ports that don't have line counting enabled, then that makes the offset-update problem easy.

A very `sendfile` API is not a good fit for Windows overlapped I/O (and perhaps considering Windows is part of what sent you down the current path). A more general interface that works well with overlapped I/O would involve making the sendfile-status object explicit, making the API evt-oriented, and perhaps making the API `dup` the given port's file descriptor to avoid issues with concurrent use of the port and to make ownership of the file resource more clear. I can image runnng `sendfile` on a blocking fd in a separate OS thread to make it more like overlapped I/O. But all of that is too much work if the goal is just to expose `sendfile` functionality.

Bogdanp:
Point taken about ports vs. paths. You're spot on that trying to fit Windows into a sendfile-like API is what led me to the current design. 

My main motivation here is to learn more about Racket's internals and I don't immediately need this functionality for anything so I'll slowly work towards the more general interface that you described since it seems like that'll produce an overall better/more useful result in the end and I'll have learned more by doing it.

Thanks!

Bogdanp:
I'm closing this until I have time to pick it back up.

PR #3282, opened 2020-07-02
Create pull_request_template.md

spdegabrielle:
intended to guide contributors to make PRs that are easier to review and accept.

It uses the guidance in https://docs.racket-lang.org/racket-build-guide/contribute.html and https://docs.racket-lang.org/style 

A screenshot is below  - to try it yourself you should make a PR against https://github.com/spdegabrielle/clash ; 
- the PR description field should be pre-populated with the text in pull_request_template.md as per the image below.

image

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,0 +1,40 @@
+**IMPORTANT: Please do not create a Pull Request without creating an issue first.**
+
+*Any change needs to be discussed before proceeding. Failure to do so may result in the rejection of the pull request.*
+
+## General Contribution Guidelines
+
+* Try to follow the style guide. https://docs.racket-lang.org/style
+* When you fix a bug or create a new feature, include a test case for it.
+* Include new or updated documentation as appropriate.
+* Build with your changes.
+
+More information: https://docs.racket-lang.org/racket-build-guide/contribute.html
+
+----
+
+## Make the PR easy for reviewers
+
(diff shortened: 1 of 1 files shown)


samth:
I'm not sure that we need a template for this. I don't think we have a problem with Pull Requests not having the right information right now -- at least, I haven't encountered such and I think I look at racket PRs more than anyone else. Do you have some examples where this was a problem? I think it would need to be a substantial problem to be worth making the PR-creation process more work.

spdegabrielle:
Many PR’s entail a lot of discussion - even PRs between the core team https://github.com/racket/gui/pull/12 
and I’m guilty myself of not following the style guide. 


And the text is just guidance - it can be ignored / deleted when submitting the PR.

The goal is to reduce the workload on the repo owners.

samth:
Right, lots of discussion is to be expected. But my point is that I don't think a template will help address problems we have. With bug reports, you see lots of instances of people being asked things like what version of Racket or what OS they have, or how to reproduce the bug, so it's good to have a template that asks those things specifically. But that doesn't seem like it's the case for pull requests. (racket/gui#12 is an issue because no one looked at it for many years, so thus all the context was lost -- again, not fixable with a template.)

spdegabrielle:
I’m happy for the PR to be closed/rejected. I’m not committed  - it is just a suggestion.

PR #3285, opened 2020-07-04
apparent typo in collection-search looks like error should be raise-a…

jbclements:
…rgument-error


jbclements:
Sorry, should have added comment:

It looks like this was meant to be a use of "raise-argument-error", not a use of "error". The existing code signals a meta-error, because these isn't a legal use of "error". 

Test added as well.

mflatt:
I agree — this repair looks right.

samth:
@jbclements is this ready to merge?

PR #3289, opened 2020-07-05
win32: add Power Shell script to set env vars

Bogdanp:
I noticed yesterday that running `msvcprep.bat` from within Power Shell doesn't modify any environment variables. This doesn't seem to be because of any sort of problem with `msvcprep.bat` itself. The added PS script runs `msvcprep.bat` in a sub shell then reads the environment variables line by line and sets them in the current shell. 

Cherry-picked from #3281.

### DIFF
--- a/racket/src/worksp/README.txt
+++ b/racket/src/worksp/README.txt
@@ -21,6 +21,10 @@ tries to find Visual Studio and run its "vcvarsall.bat". Provide an
 argument such as "x86" (32-bit mode) or "x86_amd64" (64-bit mode) to
 select the build mode.
 
+When using Power Shell, use the "msvcprep.ps1" script instead of
+"msvcprep.bat".  Running the latter from PS will appear to work, but
+will not actually change any environment variables.
+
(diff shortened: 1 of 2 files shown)


racket/src/worksp/README.txt
```diff
@@ -21,6 +21,10 @@ tries to find Visual Studio and run its "vcvarsall.bat". Provide an
 argument such as "x86" (32-bit mode) or "x86_amd64" (64-bit mode) to
 select the build mode.
 
+When using Power Shell, use the "msvcprep.ps1" script instead of
+"msvcprep.bat".  Running the latter from PS will appear to work, but
+will not actually change any environment variables.
```
mflatt, line 26: Nit: "Power Shell" should be "PowerShell". Also, I'd use "PowerShell" instead of the "PS" abbreviation just once.

Bogdanp:
@mflatt I've addresses the change you requested and added an additional change to allow variables with `=` in their values.

PR #3292, opened 2020-07-08
Clarify that synchronizing on a semaphore decrements its counter

97jaz:
This behavior is mentioned in the `Events` section of the manual but not in the section that deals specifically with semaphores.


jackfirth:
Thanks! This has bugged me before.

97jaz:
Thanks Jack!

PR #3294, opened 2020-07-13
CI support on self-hosted arm runners

pmatos:


pmatos:
We have some foreign function failures for arm64: https://github.com/racket/racket/pull/3294/checks?check_run_id=865244797#step:5:116

Holding off until this is clarified.

cc: @mflatt

mflatt:
Support for unions and arrays in BC is a hack, since libffi doesn't support them, so this failure is not surprising in retrospect (although it means I forgot to try BC on the test suite during my builds).

mflatt:
I've refined the union/array hack for AArch{32,64} in commit 740f345937a669581c3c1b3e75a018b3742ea7b6, and now "foreign-test.rktl" passes in my builds.

pmatos:
> I've refined the union/array hack for AArch{32,64} in commit [740f345](https://github.com/racket/racket/commit/740f345937a669581c3c1b3e75a018b3742ea7b6), and now "foreign-test.rktl" passes in my builds.

Thanks  - rerunning the tests.

.github/workflows/ci-pr.yml
```diff
@@ -37,6 +37,39 @@ jobs:
...
+    container:
+      image: racket/racket-ci:latest
+      options: --init
+
+    runs-on: [self-hosted, linux, ARM64]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Build
+      run: make CPUS=$(nproc) PKGS="racket-test db-test unstable-flonum-lib net-test"
```
samth, line 50: Should this also run Racket CS?

pmatos:
> > I've refined the union/array hack for AArch{32,64} in commit [740f345](https://github.com/racket/racket/commit/740f345937a669581c3c1b3e75a018b3742ea7b6), and now "foreign-test.rktl" passes in my builds.
> 
> Thanks - rerunning the tests.

@mflatt your changes fixed the tests. Thanks.

.github/workflows/ci-pr.yml
pmatos: I didn't want PRs to take to long to return results which is why I thought just building 3m for now (the default) was enough. I would say that for PRs, the default is enough and we run CS when CS becomes the default, by building from a snapshot of cgc or something. If there's a problem in CS, it will show when pushed and then we can revert if the problem occurs.

I think for `racket` at the moment, it doesn't really matter too much if the PR CI is super quick, but generally it's a good idea to have a quick turnaround for PRs and in-depth testing for Pushes.

pmatos:
For the same reason I mentioned @samth, PR CI takes about 20mins, but arm64 takes 1h30m so maybe we can test non-x86 archs only on push. I will refactor this to be for push only.

.github/workflows/ci-pr.yml
samth: That makes sense to me for PRs, but I think we do want regular testing for pushes. Maybe the performance issues will get better when "Apple Silicon" is available, if GHA provides some.

mflatt: It may take so long because it uses BC to simulate Chez Scheme to build Chez Scheme, and BC doesn't have a JIT on AArch64. For example, on a Pi 4, it takes 15 minutes just to compile "cpnanopass.ss" using the simulated Chez Scheme. (The process is much faster for me on AArch32, since there's JIT support.)

I have been thinking about ways to improve the build time — including ways that avoid adding an AArch64 JIT to BC.

samth: One possibility is to use GitHub Actions caching support to cache boot files, which could then be used to avoid the long bootstrap process in most cases.

pmatos: > That makes sense to me for PRs, but I think we do want regular testing for pushes. Maybe the performance issues will get better when "Apple Silicon" is available, if GHA provides some.

Certainly, for pushes it makes sense to run even if it's slow.
I will get a Mac Mini once they are out for testing and CI, so hopefully we can use that once it's out.

pmatos: > It may take so long because it uses BC to simulate Chez Scheme to build Chez Scheme, and BC doesn't have a JIT on AArch64. For example, on a Pi 4, it takes 15 minutes just to compile "cpnanopass.ss" using the simulated Chez Scheme. (The process is much faster for me on AArch32, since there's JIT support.)
> 
> I have been thinking about ways to improve the build time — including ways that avoid adding an AArch64 JIT to BC.

Well - at the moment those 1h30m is not even building CS. That's just CGC and 3m. If we tried CS would definitely be a bit more. And we'll see how much more once we get it running in push. I will get this added tomorrow.

pmatos: > One possibility is to use GitHub Actions caching support to cache boot files, which could then be used to avoid the long bootstrap process in most cases.

That might be useful for the push CI - can we really use old boot files, to build a recent CS?

samth: Note that in this case building CS directly from CGC might make things faster, instead of the intermediate 3m step

jbclements:
does this mean that we shouldn't claim that 7.8 suports aarch64 ? Or not for BC, at least?

samth:
I believe that whether or not @mflatt merges the fix for that bug, there will be plenty of bugs in the AArch64 support, but most things will work.

.github/workflows/ci-pr.yml
mflatt: Ah I misunderstood the 1h30m. It looks like a CGC-them-3m build takes 50 minutes, and then testing is 40 minutes. During the build, roughly 20 minutes is `raco setup`. I bet a lot of the build time is compiling the cified "cstartup.inc".

pmatos:
After some investigation there's a problem with 32bits docker and networking. I reported the issue upstream: https://github.com/moby/moby/issues/41217

For now, I will disable ARM testing.

PR #3296, opened 2020-07-15
Minor change to the issue template

sorawee:
See how it looks at https://github.com/sorawee/racket/issues/new/choose


sorawee:
Scratch this as well... What's going on...

Issue #3300, opened 2020-07-17
package regressions in 7.8


samth:
Currently, the following packages either fail to build, or have test failures, regressing from 7.7. 

7.7 build: https://pkg-build.racket-lang.org
7.8 pre-release build: https://plt.eecs.northwestern.edu/release-pkg-build/

[EDIT] using check-boxes to indicate "IMO not a release blocker (anymore)"

### Genuine regressions

- [x] clang, digimon, graphics, lambda-sh, psd, pydrnlp, ricoeur-tei-utils, schema, w3s
    * Failure is here: https://plt.eecs.northwestern.edu/release-pkg-build/server/built/fail/digimon.txt
    * Problem is likely related to one of the struct-info changes (cc @sorawee)
- [x] planning
    * Failure: https://plt.eecs.northwestern.edu/release-pkg-build/server/built/fail/planning.txt
    * xexpr error caused by a bug in racket/scribble#232. (cc @jackfirth)
- [x] doc-coverage
    * Failure is here: https://plt.eecs.northwestern.edu/release-pkg-build/server/built/test-fail/doc-coverage.txt
    * Missing documentation for `for-clause-syntax-protect` `syntax-pattern-variable?`. Are these supposed to be exported?
- [x] lwc2016
    * Failure: https://plt.eecs.northwestern.edu/release-pkg-build/server/built/fail/lwc2016.txt
    * New `translate` identifier provided by `slideshow` causes a conflict (introduced in racket/pict#57 by @florence)
- [x] measures-with-dimensions
    * Failure: https://plt.eecs.northwestern.edu/release-pkg-build/server/built/fail/measures-with-dimensions.txt
    * Problem with submodule path access in Typed Racket. Possible that racket/typed-racket#907 is related (cc @bennn)
- [x] timable
    * Failure: https://plt.eecs.northwestern.edu/release-pkg-build/server/built/test-fail/timable.txt
    * a date test that fails when run between 23:00 and 0:00. (cc @yanyingwang)

### Possible build system bugs

- [x] openweather, tjson
    * Failure: https://plt.eecs.northwestern.edu/release-pkg-build/server/built/fail/openweather.txt
    * Seems to be a permission error inside documentation building

### pkg-build issues
- [x] egg-herbie-linux, egg-herbie-osx, egg-herbie-windows
    * Likely a transient failure (something changed between fetching the sha and fetching the package)
[...]

PR #3301, opened 2020-07-18
Fix unchecked struct with custom constructor

sorawee:
Fix a bug introduced in #3267

Related to: #3300

### DIFF
--- a/pkgs/racket-test-core/tests/racket/struct.rktl
+++ b/pkgs/racket-test-core/tests/racket/struct.rktl
@@ -1477,4 +1477,10 @@
 
 ;; ----------------------------------------
 
+(let ()
+  (struct exn:foo exn () #:constructor-name make-exn:foo)
+  (test "foo" exn-message (make-exn:foo "foo" (current-continuation-marks))))
+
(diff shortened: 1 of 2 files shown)


jbclements:
@sorawee can you confirm that this was a simple typo? That is, the changed "make-struct-info" would never be correct, and this was just a case of a small piece of untested code? I'm super-excited about getting this merged.

sorawee:
It's not a typo per se. Before #3267, the `make-struct-info` call is correct. But in #3267, I added one more argument to the call. This simultaneously affects the functions:
- `make-self-ctor-checked-struct-info`
- `make-checked-struct-info`
- `make-self-ctor-struct-info`
- `make-struct-auto-info`
- and `make-struct-info`

However, I also changed every function mentioned above except the last one (which I forgot) so that they accept one more argument. This can be done because these functions except the last one are not exported.

This PR fixes the mistake in #3267 by changing `make-struct-info` to `make-field-struct-info`, which accepts one more argument. Note that we can't directly change `make-struct-info` like we did before because `make-struct-info` is public, so changing it would break backward compatibility.

PR #3303, opened 2020-07-20
If CI is running on release branch, use Chez release

pmatos:


samth:
Assuming this works we need to do the same on other platforms, I think.

samth:
@jbclements Is it ok to merge this to the release branch? It should fix some of the failures you mentioned.

mflatt:
JFYI, the `release` branch wasn't updated at `racket/ChezScheme`, but I just synced it.

jbclements:
Quick rewind. At the moment, unless I’m mistaken, the release’s use of racket/ChezScheme is fairly primitive; we start the build process with the master branch, and re-synchronize it if necessary (e.g. earlier in this cycle it was necessary). Using a release branch makes sense, but it definitely raises some questions (for me, at least):

1) Does the chez release need to be a part of the whole package-catalog setup? I imagine not, because it’s not a package. Just checking, here.
2) If we start using a release branch, should the build scripts check for a release branch and automatically update it? (This is currently a more-than-one-step script because of the need to maintain a ChezScheme-x and a ChezScheme repository (etc etc), but I see no reason why that script couldn’t be run automatically.
3) Am I deeply confused about something? Always a possibility.

John

samth:
1. ChezScheme is not a package, so it can't be in the package catalog.
2. I don't understand by "re-synchronize it if necessary" means. When the build scripts run currently, do they check out `racket/ChezScheme@master`? Or do they check out a specific commit? If the latter, where is the identity of that commit stored? 
3. What is the "ChezScheme-x" referred to here?

mflatt:
@jbclements It's probably a good idea to make the scripts pull from the `release` branch at `racket/ChezScheme`, but it's not a pressing improvement.

@samth `ChezScheme-x` is an intermediate repo check that has to do with how the build process sets up the `ChezScheme` repo and submodules so they can be pulled  by clients. It's a peculiar dance that accommodates old `git` clients that deal less gracefully with submodules. That setup is currently outside of what `distro-build` does, but probably it should be sorted out better.

PR #3304, opened 2020-07-20
Point to the release branch for other workflows

sorawee:
Similar to 5daba1fbb1c260990


sorawee:
Like 5daba1f, it should be merged to the release branch as well.

cc: @jbclements

pmatos:
Looks good to me as well. Thanks for the PR @sorawee

PR #3310, opened 2020-07-26
hash.rkt: Add hash-intersect.

scolobb:
I wrote myself a `hash-interesect` and decided to share it with the community.  I imitated `hash-union` as much as it made sense.

This is my first pull request to this repository, so I'm taking all kinds of minor stylistic remarks, in addition to normal suggestions of corrections/improvements.

I "tested" these changes by following `racket/src/README.txt` and building a minimal Racket, but I didn't find a way to actually do something like `make test` on the minimal build.  I'm running NixOS.

### DIFF
--- a/racket/collects/racket/hash.rkt
+++ b/racket/collects/racket/hash.rkt
@@ -28,6 +28,31 @@
                          (combine/key k (hash-ref one k) v)
                          v))))
 
+(define (hash-intersect
+         #:combine [combine #f]
+         #:combine/key [combine/key
+                        (if combine
+                            (λ (_ x y) (combine x y))
+                            (hash-duplicate-error 'hash-intersect))]
+         . hashes)
+  (define one (car hashes))
+  (define rest (cdr hashes))
+  (define empty-h (hash-clear one)) ;; empty hash of same type as one
+  (define (argmin f lst) ;; avoid racket/list to improve loading time
+    (for/fold ([best (car lst)] [fbest (f (car lst))]
+               #:result best)
+              ([x lst])
+      (define fx (f x))
+      (if (< fx fbest) (values x fx) (values best fbest))))
+  (for/fold ([res empty-h])
+            ([k (in-hash-keys (argmin hash-count hashes))])
+    (if (for/and ([h (in-list hashes)]) (hash-has-key? h k))
+        (hash-set res k
+                  (for/fold ([v (hash-ref one k)])
+                            ([hm (in-list rest)])
+                    (combine/key k v (hash-ref hm k))))
+        res)))
+
 (provide/contract
  [hash-union (->* [(and/c hash? immutable?)]
                   [#:combine
@@ -42,4 +67,11 @@
                    #:combine/key
                    (-> any/c any/c any/c any/c)]
                    #:rest (listof hash?)
-                   void?)])
+                   void?)]
+ [hash-intersect  (->* [(and/c hash? immutable?)]
+                       [#:combine
+                        (-> any/c any/c any/c)
+                        #:combine/key
+                        (-> any/c any/c any/c any/c)]
+                       #:rest (listof hash?)
+                       (and/c hash? immutable?))])
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -758,4 +758,41 @@ h
 
 }
 
+@defproc[(hash-intersect [h0 (and/c hash? (not/c immutable?))]
+			 [h hash?] ...
+                         [#:combine combine
+                                    (-> any/c any/c any/c)
+                                    (lambda _ (error 'hash-intersect ...))]
+                         [#:combine/key combine/key
+                                     	(-> any/c any/c any/c any/c)
+                                     	(lambda (k a b) (combine a b))])
+	 (and/c hash? immutable?)] {
+
+Constructs the hash table which is the intersection of @racket[h0]
+with every hash table @racket[h].  In the resulting hash table, a key
+@racket[k] is mapped to a combination of the values to which
+@racket[k] is mapped in each of the hash tables.  The final values are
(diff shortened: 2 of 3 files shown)


scolobb:
OK, I figured out I was calling `test` wrong in `hash.rktl`.  I also tried to figure out how to run the tests by looking at CI failure logs, but it didn't work.  I'll come back tomorrow :slightly_smiling_face:

racket/collects/racket/hash.rkt
```diff
@@ -1,5 +1,6 @@
 #lang racket/base
-(require racket/contract/base)
+(require racket/contract/base
+         (only-in racket/set set-intersect))
```
sorawee, line 3: I don't know how matter this is gonna be, but this change increases the startup time of the program:

```
#lang racket/base
(require racket/hash)
```

from 0.18s to 0.22s. If it matters, you could use a (mutable) hash mapping to `#t` to simulate a set to avoid requiring `racket/set`.

scolobb: Thank you @sorawee for your suggestion.  I ended up doing the intersection using a `for/fold` with a list accumulator, so no `set-intersect` any more.

rmculpepper:
Hi @scolobb, thanks for the PR. Here are some comments:

There are different kinds of hashes (eq? vs equal? key comparison, normal vs weak, etc). It would be best to return the same kind of hash as the first argument, like `hash-union`.

I think the combine/key argument should probably be mandatory, since unlike `hash-union` duplicates are usually expected.

Your PR inspired me to try writing a few different versions and benchmarking them. I've attached a self-contained module (with a .txt file extension to make github happy). In the examples I tried, the fastest is usually v3 if fewer than half of the keys are kept, otherwise v2.

[hash-intersect.rkt.txt](https://github.com/racket/racket/files/4997806/hash-intersect.rkt.txt)

scolobb:
Thanks a lot for your feedback @rmculpepper !  I didn't realize that keeping the hash type was important, but it clearly is.  

You are right: making `combine` mandatory makes a lot of sense.  However, since there is also `combine/key`, I would ideally need to make specifying either `combine` or `combine/key` mandatory, but not necessarily make any one of them mandatory on its own.  (Commit a2b8c11 actually made `combine` mandatory, but then the use cases in which only `combine/key` was given didn't work.)

Reading your implementations was very educational.  Reducing the number of `hash-set` is a very good idea, which considerably reduces the running time.  After some thinking, I chose your v2 to replace my original implementation because it is a tiny bit simpler than v3, and since we are talking about a general hash table library, I don't think we can tell in advance whether most of the keys will be discarded or kept.

I am quite surprised that, in some cases (sporadic?) my original implementation is considerably faster.  I guess I have to ponder more :slightly_smiling_face:

scolobb:
I'll be on vacation until August 27 and I'll be able to handle all the changes/comments (if needed) when I'm back.

scolobb:
Could this PR be merged, what do you think?  I don't have any further ideas for improvement personally, but I will be happy to get more advice or suggestions.

sorawee:
- One possible, minor optimization is to find a hash table with the least number of elements and use it as the main hash table to loop through. 
- @rmculpepper's code contains some tests that makes sure that the type of hash table is preserved.

I don't know if any of these is worth adding to the PR.

scolobb:
Thank you very much @sorawee for your feedback!  I actually thought I included @rmculpepper tests in my PR, but I am obviously wrong.

My schedule is somewhat tight at the moment, but I will try to update my PR according to your suggestions within the next couple of days.

scolobb:
> One possible, minor optimization is to find a hash table with the least number of elements and use it as the main hash table to loop through.

I realized that, in fact, this variant was already in @rmculpepper code, so I essentially copied and pasted that one, with one significant difference: @rmculpepper 's code uses `argmin`, defined in `racket/list` and not loaded in `hash.rkt`.  I added a hand-made implementation to `hash-intersect` to avoid loading `racket/list`, but the implementation in `racket/list` is more verbose than mine and probably better optimized.

Is my solution fine or should I rather opt for requiring `racket/list`?

> @rmculpepper's code contains some tests that makes sure that the type of hash table is preserved.

I included those tests into the updated commit.

scolobb:
Do you think there are any further improvements to be done to this PR?  I don't see any, but maybe I am overlooking something.

mflatt:
It looks like this is just about ready to be merged (thanks!), but I see a few small things to fix:

* The documentation for `hash-intersect` should end with `@history[#:added "7.8.0.11"]` to record when it became available.

* The documentation should mention that that comparison predicate (`eq?`, `eqv?`, or `equal?`) of the first argument determines the one for the result.

* Is it still the case that performance depends on the size of the first argument? I don't think so.

* The documentation's contract says that `combine` can be `#f`, but the contract in the code doesn't allow that. Also, it doesn't say that the default is to report an error. I think you probably meant this documentation to be more like `hash-union`.

* The documentation says `(combine v0 vi)` in one place that should be `(combine v vi)`, and it refers to `hi` instead of `h`.

mflatt:
P.S. - Sorry that this has moved so slowly, but I'll pay attention and move things along when the small doc corrections are ready.

scolobb:
Thank you @mflatt for your detailed feedback!  I think I implemented all of your suggestions.

I had some trouble understanding the reference to `hi` vs. `h`, but I think I got it right after all.

samth:
I think the contract for `h0` should require `immutable?`, instead of `(not/c immutable?)`, right?

scolobb:
@samth Ah yes, you are right!  I dumbly copied the contract on `h0` from the docs of `hash-union!`…

The actual contract in `hash.rkt` is good (`(and/c hash/c immutable?)`), but it should indeed be corrected in the Scribble docs.

samth:
Do you want to submit a fix, or should I do it?

scolobb:
Oh, thank you for the opportunity, here's the PR: #3648.

Issue #3308, opened 2020-07-23
hash-ref error from show-import-tree with #:path-to


sorawee:
**What version of Racket are you using?**
v7.8.0.5

**What program did you run?**

```
#lang racket/base

(require syntax/moddep)

(show-import-tree 'racket/base
                  #:path-to 'racket/private/define-et-al)
```

**What should have happened?**

It shows a path successfully.

**If you got an error message, please include it here.**

```
hash-ref: contract violation
  expected: hash?
  given: #f
  argument position: 1st
  other arguments...:
   '(("    " #<path:/Users/sorawee/git/racket/racket/collects/racket/private/define-et-al.rkt> "" 0) ("   " #<path:/Users/so...
   #f
  context...:
   /Users/sorawee/git/racket/racket/collects/syntax/moddep.rkt:20:4: loop
   /Users/sorawee/git/racket/racket/collects/racket/private/map.rkt:68:13: for-each
   /Users/sorawee/git/racket/racket/collects/syntax/moddep.rkt:20:4: loop
   /Users/sorawee/git/racket/racket/collects/racket/private/map.rkt:68:13: for-each
   /Users/sorawee/git/racket/racket/collects/syntax/moddep.rkt:20:4: loop
   /Users/sorawee/git/racket/racket/collects/racket/private/map.rkt:68:13: for-each
   /Users/sorawee/git/racket/racket/collects/syntax/moddep.rkt:20:4: loop
   /Users/sorawee/git/racket/racket/collects/racket/private/map.rkt:68:13: for-each
   /Users/sorawee/git/racket/racket/collects/syntax/moddep.rkt:20:4: loop
   "/Users/sorawee/playground/test.rkt": [running body]
   temp35_0
   for-loop
   run-module-instance!
   perform-require!
```

**Please include any other relevant details**
MacOS

maueroats:
I think the root of this problem is in commit 32b274dee15e5c6068655ac21e429fda82b3de4d from 2018 (sorry, need to learn how to link the diff for a file).

There is a [loop in moddep.rkt](https://github.com/racket/racket/blob/b533a13d2032950c4939fd99b6c498b823d08914/racket/collects/syntax/moddep.rkt#L21-L28) that only runs when `#:path-to` is given, but it looks like it assumes that `seen` is nonzero (which only happens when `#:dag` is also given). 

[...]

PR #3312, opened 2020-07-27
Fix show-import-tree hash-ref contract violation

maueroats:
See note in issue. Contains no test code. Danger: produced massive output from #3308.


pmatos:
Thanks for the PR - I have changed the title to make to more clear and added 3308 to the linked issues.

PR #3314, opened 2020-07-28
s/comparing with with/comparing it with/

deeglaze:
Duplicated word in documentation.


pmatos:
Thanks for the PR. Merging.

PR #3316, opened 2020-07-29
Solo Chez build/test workflow

pmatos:


pmatos:
To test solo Chez we are doing:
* fetch pb boot files
* build using pb boot files
* build again but using the architecture boot files
* test

@mflatt does this look ok to you?

pmatos:
Just noticed the tests are not being run in parallel - which is a wasted opportunity. Certainly on the arm64 boards, it feels like it's a waste given they are being used just for the CI but only a single core is being used out of the 4.

samth:
How long does step 3 take there?

For parallel tests, probably the easiest thing to do is use make.

pmatos:
> How long does step 3 take there?
> 
> For parallel tests, probably the easiest thing to do is use make.

The building is very fast. Much faster than step 2. 

build with pb:
* arm64 14m50s
* x86_64 3m09s

build boot files:
* arm64 2m21s
* x86_64 40s

test:
* arm64 1h37m
* x86_64 30m37s

The biggest issue is testing. I will try running the testing in parallel and see what happens. It should create any problems but you never know. :+1: 

Although I did notice that even the Chez bootstrap when compiling the scheme files is done serially (even with `make -j`) so it could be improved.

pmatos:
All green - I will merge this and create a separate PR for when I have this running in parallel.

PR #3322, opened 2020-07-30
Speed up solo chez testing

pmatos:
This allows multiple test configurations to run in parallel, therefore
taking advantages of multiple processors.

### DIFF
--- a/.github/scripts/test.sh
+++ b/.github/scripts/test.sh
@@ -1,22 +1,80 @@
 #!/bin/bash
+
+# Catch INT to kill subprocesses
+trap end_make INT
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-runmats() {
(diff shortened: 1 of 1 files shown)


pmatos:
This provides a really good speed up on my machine, going from 1273 secs to 116 secs for testing racket chez, at the cost of a slightly more complex bash script.

I was tempted to rewrite this in racket and was going to do so but decided against it in order to avoid adding a dependency on racket for testing chez. However, at some point it might pay off as we don't want to keep growing this bash script, I guess.

pmatos:
We don't actually run these tests on Mac, but I just noticed that `nproc` doesn't actually exist there, so we would have to check if we are on mac and go through `sysctl` instead. Let me know if you think we should change this.

mflatt:
I'm ok with deferring the `sysctl` improvement to later.

samth:
My one thought about this is that it would be nice to have these scripts somewhere more general than the .github folder, since they're generally useful.

PR #3323, opened 2020-07-30
Improve (sinh x) : x~0

bdeket:
```
(require plot math/bigfloat math/flonum)
(define a 1e-3)(define b .15)
  (parameterize ([plot-x-transform log-transform])
    (plot (list (function (λ (x) (flulp-error (sinh x) (bigfloat->flonum (bfsinh (bf x)))))   a b #:color 'blue)
                (function (λ (x) (flulp-error (mysinh x) (bigfloat->flonum (bfsinh (bf x))))) a b))))
```
sinh
sinh-det

Adding more terms doesn't seem worth it.


racket/collects/racket/math.rkt
```diff
@@ -58,6 +58,17 @@
           [(real? z)
            (let loop ([z z])
              (cond [(z . < . 0) (- (loop (- z)))]
+                   [(< z .13)
+                    ;tailor expansion near 0 to avoid cancellation
```
samth, line 62:
```suggestion
                    ;; Taylor expansion near 0 to avoid cancellation
```

racket/collects/racket/math.rkt
```diff
@@ -58,6 +58,17 @@
           [(real? z)
            (let loop ([z z])
              (cond [(z . < . 0) (- (loop (- z)))]
+                   [(< z .13)
+                    ;; Taylor expansion near 0 to avoid cancellation
+                    ;~ z+z^3/3!+z^5/5!+...
+                    (define z^2 (* z z))
+                    (define-values (s z^i)
+                      (for/fold ([s z]
+                                 [z^i (* z z^2)])
```
sorawee, line 67: Instead of using `define-values`, you can use:

`(for/fold ([s ....] [z^i ....] #:result s) (....) ....)`

bdeket: ah thanks,
now I wish I knew this before...

bdeket:
I now noticed that math/base already has flsinh and friends for better accuracy, so maybe this is not really worth it.

samth:
I think it's always worth it to make these functions better

bdeket:
I deleted this PR by accident. I recreated my branch in [my fork](https://github.com/bdeket/racket/tree/patch-2). What is the best way (and what is still needed) to get this merged?

samth:
I think you need to open a new PR since the branch was deleted. Beyond that, I would make the change @sorawee suggested to use `#:result`, and I'd also want an opinion from someone who knows more floating-point than me, such as @mflatt @pavpanchecka or @soegaard.

PR #3326, opened 2020-08-03
Fix codeql scanning by installing uuid-dev

pmatos:


pmatos:
Good - now passing code scanning!

mflatt:
For the uuid-dev part, Chez Scheme is not supposed to depend on libuuid anymore — but I see that I forgot to update some makefiles. I'll push a makefile repair, and then the install of uuid-dev can be skipped.

pmatos:
> For the uuid-dev part, Chez Scheme is not supposed to depend on libuuid anymore — but I see that I forgot to update some makefiles. I'll push a makefile repair, and then the install of uuid-dev can be skipped.

If you fix this and codeql starts passing I will close this without merging.

pmatos:
Fixed in fcd7295457ba199972089645bb9f698ae6e2bff2 by removing remaining uuid deps. Closing without merging.

PR #3327, opened 2020-08-03
Fix target names post-CS becoming the default

pmatos:


mflatt:
I think either `RACKETBC_SUFFIX=""` needs to be supplied with `bc-in-place` or `racketbc` needs to be run instead of `racket`.

pmatos:
> I think either `RACKETBC_SUFFIX=""` needs to be supplied with `bc-in-place` or `racketbc` needs to be run instead of `racket`.

Right - thanks.

PR #3328, opened 2020-08-03
Fix tool name in SARIF files depending on racket build mode

pmatos:
Attempt to debug something in the CI system. Disregard for now... :)


pmatos:
As I initially thought the `adjust.sh` file is not changing the tool name appropriately.
https://github.com/racket/racket/pull/3328/checks?check_run_id=939679381#step:9:16

pmatos:
Fixing this under this PR - changing title and adding commits for fix.

PR #3330, opened 2020-08-03
Speed up solo chez testing (#3322)

pmatos:
This allows multiple test configurations to run in parallel, therefore
taking advantages of multiple processors.

### DIFF
--- a/.github/scripts/test.sh
+++ b/.github/scripts/test.sh
@@ -1,22 +1,80 @@
 #!/bin/bash
+
+# Catch INT to kill subprocesses
+trap end_make INT
+
 DIR="$( cd "$( dirname "${BASH_SOURCE[0]}" )" >/dev/null 2>&1 && pwd )"
 
-runmats() {
-  echo make allxhelp "$@"
-  make -C ${MACH}/mats allxhelp "$@" 2>&1 | tee -a Make.out | grep '^matting '
+if [ -z ${MACH} ]; then
+    echo "Please define MACH" 1>&2
+    exit 1
+fi
+
+ERR=0
+MATS=${MACH}/mats
+SUMMARYFILE=${MACH}/summary
+declare -a PIDS
+
+# Clean and create summary file
+rm -f "${SUMMARYFILE}"
+touch "${SUMMARYFILE}"
+
+end_make() {
+    echo "Caught Ctrl-C" 1>&2
+    for i in "${!PIDS[@]}"; do
+        local pid="${PID[$i]}"
+        echo "Killing pid ${pid}" 1>&2
+        kill ${pid}
+        rm -r "${MATS}"."${i}"
+    done
+    exit 1
 }
 
-# Split these out so that we get output every 10 minutes on Windows builds.
-runmats o=0
-runmats o=0 cp0=t
-runmats o=3
-runmats o=3 cp0=t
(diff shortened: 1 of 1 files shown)


pmatos:
This was merged in #3322 but reverted due to breakages during testing.
There are some possible race conditions which need to be understood and avoided.

pmatos:
> I have little insight here, but it looks good to me.

Thanks, I will not upload just yet because there are a few race conditions which I need to fix here. If I change some of the tests I will ask you to look into it before merging.

gus-massa:
There are a lot of temporal files that are created before and during the test. Can this be the problem?

pmatos:
> There are a lot of temporal files that are created before and during the test. Can this be the problem?

Yes - that's the problem. And I haven't had the time to fix the issue to make the tests parallelizable.

gus-massa:
This part of Mf-base looks suspicious:
https://github.com/racket/racket/blob/5c455885734dddd4400f11af1c0ba1e7348e9189/racket/src/ChezScheme/mats/Mf-base#L403-L409

pmatos:
> This part of Mf-base looks suspicious:
> 
> https://github.com/racket/racket/blob/5c455885734dddd4400f11af1c0ba1e7348e9189/racket/src/ChezScheme/mats/Mf-base#L403-L409

Interesting - it looks like we could generate a temp dir for these instead, right?

gus-massa:
I'm just guessing from the names, but I guess that instead of `upupsrcdir/mats` it should use `upsrcdir`, but I guess that `upsrcdir` is not defined (??????????).

mflatt:
I'm not following the original context, but it's meant to be "upup" in "Mf-base". The first "up" is for the machine directory (like "ta6le") and the second one is for "mats" within that directory. That is, the paths are about getting from a workarea "mats" directory to the source "src" directory.

pmatos:
> I'm not following the original context, but it's meant to be "upup" in "Mf-base". The first "up" is for the machine directory (like "ta6le") and the second one is for "mats" within that directory. That is, the paths are about getting from a workarea "mats" directory to the source "src" directory.

So, The reason I haven't gotten to merge this patch is that there are failures when the tests run concurrently. This is likely due to artifacts built during the testing. I haven't however had the time to investigate this. I was hoping that what @gus-massa pointed to would be the place where these artifacts are built in  and we could add some sort of `mktemp` there somewhere so that each thread of testing would not conflict with each other. However, from your comment this is not the case and therefore this will remain open until I have time to look into it. Unfortunately it's also low priority because the Chez tests are per se not a bottleneck.

gus-massa:
Let´s try again. 

The test in [examples.ms](https://github.com/racket/racket/blob/master/racket/src/ChezScheme/mats/examples.ms) use a shared directory, and each one compiles a few .ss files to .so files. My guess is that the .so files are incompatible.

The directory is setup in https://github.com/racket/racket/blob/5c455885734dddd4400f11af1c0ba1e7348e9189/racket/src/ChezScheme/mats/Mf-base#L148 but perhaps it must be changed also in other parts. 

[I agree that this is low priority, the Chez Scheme tests are no so slow.]

gus-massa:
It will be interesting to see if this is simplified after this PR is merged related https://github.com/cisco/ChezScheme/pull/561#

PR #3332, opened 2020-08-03
makefile: Add an UPDATE_PKGS_OPTIONS config variable

lexi-lambda:
This is analogous to `PLT_SETUP_OPTIONS`, but for the `raco pkg update` step.

I find this useful because I sometimes want to be able to add `--pull try` to ignore failures to update local clones. I used to do this by (hackily) overriding the value of `UPDATE_PKGS_ARGS` directly, but that is no longer possible under the new `Makefile` since `UPDATE_PKGS_ARGS` is eagerly expanded by the preprocessor.


mflatt:
This is fine by me, but maybe `PKG_UPDATE_OPTIONS` would be better externally, since it more closely matches `pkg update`?

Less important: Spaces would be better instead of tabs in  a `\` continuation for a variable definition.

lexi-lambda:
> This is fine by me, but maybe `PKG_UPDATE_OPTIONS` would be better externally, since it more closely matches `pkg update`?

Sure. I’ve renamed the variable, and I also took the opportunity to rename the internal `UPDATE_PKGS_ARGS` and `INSTALL_PKGS_ARGS` for consistency with `ALL_PLT_SETUP_OPTIONS`.

> Less important: Spaces would be better instead of tabs in a `\` continuation for a variable definition.

Good catch—I’ve replaced the tabs with spaces.

PR #3333, opened 2020-08-04
Preserve multivalued-ness of stream in stream-take

sorawee:


sorawee:
Note that there's nothing in the doc to change _regarding multiple values_, because it's already phrased as if it should work (though in fact it didn't before this PR).

That doesn't mean the doc has no issue. There are in fact multiple issues in the doc, but those will be addressed in a separate PR.

Issue #3336, opened 2020-08-05
Dead store in notdisplay


pmatos:
Clang static analyzer shows up that `notdisplay` is a dead store. This is a work around for those trying to fix the featured [*Call to Racket* issue](https://racket-news.com/2020/08/racket-news-issue-36.html#calltoracket). The security vulnerability issue is not public, and therefore here's a screenshot.

![2020-08-05-094650_1457x847_scrot](https://user-images.githubusercontent.com/7911/89386050-d44cdb80-d700-11ea-838e-eb0a4c92e793.png)

You can reproduce this by compiling Racket CGC using `scan-build`. Detailed steps can be seen in the [GitHub Actions workflow for static analysis](https://github.com/racket/racket/blob/dcf034280efcdac798d9c8a6e876067f862b8af0/.github/workflows/scanbuild_static-analysis.yml#L10).

PR #3341, opened 2020-08-06
Remove dead store identified by a static analyzer

jessealama:
Fix for security [issue 248](https://github.com/racket/racket/security/code-scanning/248) (public analogue is issue [3336](https://github.com/racket/racket/issues/3336)).


pmatos:
@jessealama there's a failure at: https://github.com/racket/racket/pull/3341/checks?check_run_id=952244161#step:4:71
Could this be caused by the patch? Are you certain that `to_quote()` doesn't have side effects?

pmatos:
Ah yes - I checked the source - it definitely has side-effects.
LGTM now.

I will leave it to @mflatt to review.

PR #3358, opened 2020-08-21
docs: regexp size limit only applies to Racket BC

LiberalArtist:
Racket CS is happy to evaluate e.g. `(regexp (make-string 100000 #\x))`.


sorawee:
Does CS have a limit at all? If so, should we document that?

PR #3365, opened 2020-08-25
gracket/gc2 makefile

bdeket:
Without this change, the below build comands with the latest snapshot from the Utah server fail for me.
This reflects the GRACKETLDFLAGS in the makefile one folder above. However I have no idea if this has wider implications...
```
mkdir build
cd build
..\configure --enable-bcdefault
make
```

### DIFF
--- a/racket/src/bc/gracket/gc2/Makefile.in
+++ b/racket/src/bc/gracket/gc2/Makefile.in
@@ -64,7 +64,7 @@ XFORM_CPP_ARGS = -I$(srcdir)/../../gc2 $(NOGCINC) $(OPTIONS) @PREFLAGS@ $(XFORM_
 XFORM = $(XFORM_CMD) --cpp "$(CPP) $(XFORM_CPP_ARGS)" @XFORMFLAGS@ -o ++out
 XFORMDEP = $(srcdir)/../../gc2/xform-mod.rkt $(srcdir)/../../gc2/gc2.h
 
-GRACKETLDFLAGS = $(LDFLAGS) -L../../racket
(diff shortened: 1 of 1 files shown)


pmatos:
Interesting, if I run the above from `racket/racket/src/` I cannot repro the issue.
Is this on a Linux system, I presume? Any specific distro/arch?

Do you have the same issue building in-tree? So running the configure/make straight from `racket/racket/src/` instead of using a  build subdir?

mflatt:
I'm not sure why I'm not seeing a build failure, but the change does look right.

bdeket:
sorry, I had another flag set:
../configure --enable-shared --enable-bcdefault
without the second flag this other flag, the build indeed passes, otherwise libtool complains

mflatt:
I didn't see a failure even with `--enable-shared`, but I'm still happy to merge this.

bdeket:
Just for the record since it's already merged (thanks!)
I tried again (in src, not in a build directory)
with
./configure --enable-bcdefault && make
everything works
with
./configure --enable-shared --enable-bcdefault && make
the make script stops with:
```
libtool --mode=link --tag=CC gcc -rpath /home/bert/builds/racket-snap/src/racket-7.8.0.9/lib -pthread -L../../racket -o ../gracket3m grmain.lo ../../libracket3m.la  
/usr/bin/libtool: line 7706: cd: ../../racket: No such file or directory
libtool:   error: cannot determine absolute directory name of '../../racket'
make[6]: *** [Makefile:106: ../gracket3m] Error 1
```
Running on archlinux

PR #3368, opened 2020-08-26
Fix #3363: track disappeared-use for case's else

### LINKED ISSUE #3363: case doesn't track disappeared-use for else
7.8 [cs]

**What program did you run?**

```
#lang racket
(case 1
  [else 2])
```

There should be an arrow from `else` to `#lang racket`

N/A

N/A

sorawee:


racket/collects/racket/private/case.rkt
```diff
@@ -8,18 +8,27 @@
                          "stxcase-scheme.rkt"
                          "qqstx.rkt" "define.rkt" "sort.rkt" "fixnum.rkt"))
   (#%provide case)
+
+  ;; identifier? :: any/c -> boolean?
+  (define-for-syntax (identifier? stx)
```
mflatt, line 13: `identifier?` is available from `"stx.rkt"`, so it would be better to import that.

sorawee: Thanks, fixed.

PR #3369, opened 2020-08-26
rktio: get CPU count from the OS on FreeBSD and OpenBSD

Bogdanp:


xxyzz:
Sorry to bother. I don't know why it uses `sysctlbyname` instead of `sysconf` on macOS and BSD. They should get the same result.

Bogdanp:
@xxyzz thanks! That does seem better and it works on all three platforms. This also made me realize the two branches were getting different values: on Linux the number of "online" CPUs was being requested, whereas on OS X, it was the number of configured CPUs. Should we keep getting the configured procs (`_SC_NPROCESSORS_CONF`) on macOS for backwards-compatibility?

pmatos:
I don't think we should be changing the current behaviour to be honest.

pmatos:
Ah, ok - just noticed you updated your comment. The number of online cpus might be interesting but could be a pkg or something of the sort.

PR #3386, opened 2020-09-05
Add arrows from definitions to all-defined-out

sorawee:
Currently, `(all-defined-out)` expands to all identifiers defined in a
module. However, this won't cause arrows to be drawn from definitions
to `all-defined-out` because the source location of expanded identifiers
are of definitions rather than `(all-defined-out)`.

This PR fixes the source location of the exported identifiers.
However, since these identifiers are potentially armed, modifying it
will cause it to be tainted and unusable. The workaround is to
use the syntax property `disappeared-use` instead, and since we are
modifying the syntax, we also need to mark `original-for-check-syntax`.

Other provide transformers whose `export-orig-stx` is faithful could also
benefit from this PR. Provide transformers like `struct-out` however is
not faithful (its `export-orig-stx` is struct definitions rather than
`struct-out` itself), so no arrow will be drawn.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/module.rktl
+++ b/pkgs/racket-test-core/tests/racket/module.rktl
@@ -1652,7 +1652,7 @@ case of module-leve bindings; it doesn't cover local bindings.
 (require (rename-in racket/base [lib racket-base:lib]))
 
 (let ()
-  (define (find-disappeared stx id)
+  (define (find-disappeared stx pred)
     (let loop ([s stx])
       (cond
        [(syntax? s)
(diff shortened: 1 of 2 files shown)


sorawee:
CC: @mflatt and @rfindler

rfindler:
I like very much what you're trying to do here! The diff isn't very easy to read, however, because of changes that are code style changes. That said, it looks okay to me (assuming it works and tests pass and drracket builds and runs, etc.)

sorawee:
I think I'm going to abandon this PR. We cannot really guarantee what `export-orig-stx` will be like in general, so using a uniform approach for all provide transformers looks like a bad idea. Moreover, for provide transformers like `struct-out`, _had_ its `export-orig-stx` been faithful, then there would have been two arrows for the following program:

```
#lang racket
(struct a ())
(provide (struct-out a))
```

I believe a better approach is that each provide transformer should explicitly attach `disappeared-use` manually, so that it can precisely control how arrow will be drawn.

Issue #3378, opened 2020-08-31
assignment to identifiers bound by for-like forms can break invariants


LiberalArtist:
Identifiers bound by `for`- and `for/fold`-like forms—for both accumulators and sequence elements—can be `set!`. Assigning to these variables can break invariants or expose implementation details of a `for`-like form or for-clause transformer (as in `define-sequence-syntax`).

Some examples:

1.  ```racket
    #lang racket
    
    (for/lists (acc)
               ([v (in-range 2)])
       (unless (zero? v)
         (set! acc 'bad))
       v)
    ```
    ->
    ```
    racket/private/for.rkt:1963:23: reverse: contract violation
      expected: list?
      given: '(1 . bad)
    ``` 
2.  ```racket
    #lang racket
    
    (require (for-syntax syntax/parse))
    
    (define-sequence-syntax in-my-naturals
      (λ () #'in-naturals)
      (syntax-parser
        [[(n) (_)]
         #'[(n) (:do-in ([(n) 0])
                        #t
                        ([n n])
                        #t
                        ([(n) n])
                        #t
                        #t
                        ((add1 n)))]]))
    
    (for/list ([i (in-my-naturals)]
               #:when (< i 3))
      (set! i 'bad))
    ```
    ->
    ```
    add1: contract violation
      expected: number?
      given: 'bad
    ```

mflatt:
Fixing `for/lists` is easy enough.

For `:do-in`, the answer may be to add "don't do that" to the documentation. That is, alert users of `:do-in` that arbitrary code runs between the bindings and the last two clauses of `:do-in`. Among the uses of `:do-in` in the `collects/racket/private/for.rkt`, it looks like only `*in-directory` goes wrong that way, and I didn't find other broken uses, so maybe the problem is not common.

mflatt:
Well, fixing `for/lists` maybe isn't as easy as I thought, but I'm still looking at it.

mflatt:
I don't see a good way to fix `for/lists`, so I'm falling back to "don't do that" documentation there, too  — at least for now.

The problem is related the strange scoping of identifiers in `for/fold`. More fundamentally, I think it's a mismatch between the intuition that a `for/fold` accumulator corresponds to a single variable versus the functional implementation with nested recursions where a new variable is needed each time around and at each layer of nesting.

LiberalArtist:
[...]

PR #3388, opened 2020-09-05
Prevent assignment to `for` bindings

LiberalArtist:
This is an attempt at resolving https://github.com/racket/racket/issues/3378 by making it a syntax error to attempt to `set!` any of the identifiers bound by `for`-like forms. All of the existing tests pass (except those added in 8ca4977 that test the behavior I'm trying to change), but this needs more tests and, if it works out, a corresponding change to the documentation from 8ca4977. It may also need some more `syntax-protect`s.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/file.rktl
+++ b/pkgs/racket-test-core/tests/racket/file.rktl
@@ -2279,7 +2279,8 @@
           (parameterize ([current-directory tmp-dir])
             (for/hash ([f (in-directory)])
               (let ([real-f f])
-                (set! f 'trying-to-break-in-directory)
+                ;; FIXME: add test that this would be an error:
+                ;; (set! f 'trying-to-break-in-directory)
                 (values real-f #t)))))
     (define (mk) (in-directory))
     (test ht
--- a/pkgs/racket-test-core/tests/racket/for.rktl
+++ b/pkgs/racket-test-core/tests/racket/for.rktl
@@ -642,7 +642,8 @@
       (for*/lists (lst) ([x '()])
         (define-syntax (m stx) #'0)
         m))
-
+#|
+;; FIXME: test that this is an error
 (test '(bad 1)
       'for/lists-weird-set!
       (for/lists (acc)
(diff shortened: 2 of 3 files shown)


mflatt:
Sorry for the delay here.

While I appreciate this experiment, it's a kind of backward-incompatible change to the semantics that  we normally refrain from. Sometimes, these kinds of changes are necessary to keep the language safe or to close a security hole. I don't think that's the case here, though. For example, it's reasonable for `for/lists` to check that it has a list to reverse. (Also, shadowing can cause the same kinds of problems as  assignment — unless your changes also cover shadowing, but then it's a further change to the semantics.) Assigning to a `for` variable is a bad idea, but I think it's well defined and safe, so a programmer who misuses assignment that way just makes their own code bad.

If I'm not misunderstanding the safety issues, maybe this change is better to take up in Rhombus?

LiberalArtist:
I don't have a very firm opinion on this. I have never seen this be an issue in practice—certainly I have never wanted to `set!` a `for` variable—which probably suggests leaving well enough alone.

The aspect that seems most concerning is that users of `define-sequence-syntax`, in particular, are sort of encouraged to use unsafe operations for performance after an initial check. If a sequence implementer hadn't considered the possibility of `set!` (which I hadn't, until #3378), the breakage could be worse than a bad error message. On the other hand, it's encouraging that only `in-directory` actually seemed vulnerable to assignment, and it used safe functions.

> (Also, shadowing can cause the same kinds of problems as assignment — unless your changes also cover shadowing, but then it's a further change to the semantics.)

I'm not sure I'm thinking of what you have in mind: do you mean something like this?
```racket
#lang racket
(for/fold ([x 'a])
          ([x '(b)])
  (define x 'c)
  #:break #f
  (define x 'd)
  x)
```
Compatibility aside, what I was aiming for was roughly to be able to imagine that `for/fold` desugared to an application of `sequence-fold` to a `lambda` expression. ("Roughly" because that doesn't account for things like `#:break`, `#:final`, and nested iteration.) The `for/fold` implementation itself, each sequence transformer, and the user code would have their own identifiers, as though they were functions accepting arguments and returning results, rather than bits of code sliced up and inlined around each other.

PR #3391, opened 2020-09-06
reqprov: fix disappeared-use

sorawee:
This PR fixes several issues regarding disappeared-use in reqprov.rkt

<s>1. Add `disappeared-use` for each exported identifier expanded from
`all-defined-out` with a syntax location of `all-defined-out` so that
Check Syntax can draw arrows from definitions to `all-defined-out`.</s>

2. Copy/preserve disappeared-use for all provide subforms instead
of only the first one. The following program doesn't have an arrow from
`struct-out` to `#lang racket` prior this PR, but it does after this PR.

        #lang racket
        (provide a? (struct-out a))
        (struct a ())

3. Fix incorrect `disappeared-use` for `struct-out`: it is inappropriate
to check if an identifier is bound to a struct info in provide
pre-transformer because it would not be able to handle backlinks.
This PR moves the attachment from provide pre-transformer to provide
transformer, allowing Check Syntax to draw an arrow for the identifier
`abc` in:

       #lang racket
       (provide (struct-out abc))
       (struct abc ())

### DIFF
--- a/pkgs/racket-test-core/tests/racket/module.rktl
+++ b/pkgs/racket-test-core/tests/racket/module.rktl
@@ -1678,10 +1678,6 @@ case of module-leve bindings; it doesn't cover local bindings.
                           (define a 1)))])
     (test #t find-disappeared form (id=? #'struct-out))
     (test #t find-disappeared form (λ (id) (eq? (syntax-e id) 'abc))))
-  (let ([form (expand #'(module m racket/base
-                          (provide (all-defined-out))
-                          (define abc 1)))])
-    (test #t find-disappeared form (λ (id) (eq? (syntax-e id) 'abc))))
   (let ([form (expand `(module m racket/base
                         (require (only-in racket/base car))))])
     (test #t find-disappeared form (id=? #'only-in)))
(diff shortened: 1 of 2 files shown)


sorawee:
CC: @mflatt and @rfindler. Hopefully the diff will be easier to read this time.

rfindler:
Probably you want to write some test cases (for the new behavior). Sadly they wouldn't go in this repo but there is some [code here](https://github.com/racket/drracket/blob/master/drracket-tool-test/tests/check-syntax/syncheck-direct.rkt) that you can add to.

sorawee:
I did include some tests (see https://github.com/racket/racket/pull/3391/files#diff-20b0e3ab5f7a3c0c96abc2ed491d0207), but I can add more tests in DrRacket too.

rfindler:
Oh! Sorry. It did seem like a lot of changes for what was in the description. :)

sorawee:
I updated the PR so that the PR is strictly bug fixing and could be readily merged. Particularly, I remove changes to `all-defined-out` since the changes will break the renaming functionality in Racket Mode, and fixing it requires https://github.com/racket/drracket/pull/417 to be integrated first.

rfindler:
This diff is backwards, right? I'm not going crazy?

Screen Shot 2020-09-09 at 6 31 24 PM

sorawee:
Well, I reverted the changes, so it's supposed to be backwards! If you look at the overall diff, there should be no change in those areas (well, there's an unfortunate whitespace change on one line...)

sorawee:
Ping, any thoughts on this? It's been a month.

PR #3396, opened 2020-09-10
for: optimize literals

### LINKED ISSUE #883: The for clause [i (#%datum <integer>)]
Currently (for/list ([i 100]) i) and (for/list ([i (in-range 100)]) i) doesn't produce the same code.
I suggest special casing the clause [i (#%datum <integer>)].

```
(require racket/pretty)
(pretty-print (syntax->datum (expand #'(for ([j 100]) j))))
(pretty-print (syntax->datum (expand #'(for ([j (in-range 100)]) j))))

(equal? (syntax->datum (expand #'(for ([j 100]) j)))
         (syntax->datum (expand #'(for ([j (in-range 100)]) j))))
```

sorawee:
Closes #883

### DIFF
--- a/pkgs/racket-test-core/tests/racket/for.rktl
+++ b/pkgs/racket-test-core/tests/racket/for.rktl
@@ -1166,6 +1166,9 @@
   (define-syntax-rule (#%datum . x) (#%old-datum . 3))
   (test-sequence [(0 1 2)] 5))
 
+(err/rt-test (for/list ([x -1]) x))
+(err/rt-test (for/list ([x 1.5]) x))
+
 ;; ----------------------------------------
 
 (report-errs)
--- a/pkgs/racket-test-core/tests/racket/for.rktl
+++ b/pkgs/racket-test-core/tests/racket/for.rktl
@@ -1158,7 +1158,27 @@
 (let ()
   (test #t 'same-expansion-for-integer-clause
         (equal? (syntax->datum (expand #'(for ([j 100]) j)))
-                (syntax->datum (expand #'(for ([j (in-range 100)]) j))))))
+                (syntax->datum (expand #'(for ([j (in-range 100)]) j)))))
+
+  (test #t 'same-expansion-for-list-clause
+        (equal? (syntax->datum (expand #'(for ([j '(1 2 3)]) j)))
+                (syntax->datum (expand #'(for ([j (in-list '(1 2 3))]) j)))))
+
+  (test #t 'same-expansion-for-vector-clause
+        (equal? (syntax->datum (expand #'(for ([j #(1 2 3)]) j)))
+                (syntax->datum (expand #'(for ([j (in-vector #(1 2 3))]) j)))))
+
+  (test #t 'same-expansion-for-hash-clause
+        (equal? (syntax->datum (expand #'(for ([(i j) #hash((1 . 2) (3 . 4))]) j)))
+                (syntax->datum (expand #'(for ([(i j) (in-immutable-hash #hash((1 . 2) (3 . 4)))]) j)))))
+
+  (test #t 'same-expansion-for-string-clause
+        (equal? (syntax->datum (expand #'(for ([j "abc"]) j)))
+                (syntax->datum (expand #'(for ([j (in-string "abc")]) j)))))
+
+  (test #t 'same-expansion-for-bytes-clause
+        (equal? (syntax->datum (expand #'(for ([j #"abc"]) j)))
+                (syntax->datum (expand #'(for ([j (in-bytes #"abc")]) j))))))
 
 ;; #%datum is picked up (from @gus-massa)
 (let ()
--- a/pkgs/racket-test-core/tests/racket/for.rktl
+++ b/pkgs/racket-test-core/tests/racket/for.rktl
@@ -1167,17 +1167,22 @@
   (test-sequence [(0 1 2)] 5))
 
 ;; for expanded in expression context
-(let ()
+(module test-for-expansion racket
+  (provide bar%)
   (define foo%
     (class object%
       (super-new)
       (define/public (abc) '(1 2 3))))
 
-  (test #t class?
-        (class foo%
-          (super-new)
-          (inherit abc)
-          (for ([x (abc)]) #t))))
+  (define bar%
+    (class foo%
+      (super-new)
+      (inherit abc)
+      (for ([x (abc)]) #t))))
+
+(let ()
+  (local-require 'test-for-expansion)
+  (test #t object? (new bar%)))
 
 (err/rt-test (for/list ([x -1]) x))
 (err/rt-test (for/list ([x 1.5]) x))
--- a/pkgs/racket-test-core/tests/racket/for.rktl
+++ b/pkgs/racket-test-core/tests/racket/for.rktl
@@ -1154,6 +1154,18 @@
   (test (void) 'for/foldr-result-delay-2 (force result))
   (test #t 'for/foldr-result-delay-3 evaluated?))
 
+;; same expansion (from @soegaard)
+(let ()
+  (test #t values
+        (equal? (syntax->datum (expand #'(for ([j 100]) j)))
+                (syntax->datum (expand #'(for ([j (in-range 100)]) j))))))
+
+;; #%datum is picked up (from @gus-massa)
+(let ()
+  (local-require (only-in racket (#%datum #%old-datum)))
+  (define-syntax-rule (#%datum . x) (#%old-datum . 3))
+  (test-sequence [(0 1 2)] 5))
+
 ;; ----------------------------------------
 
 (report-errs)
--- a/pkgs/racket-test-core/tests/racket/for.rktl
+++ b/pkgs/racket-test-core/tests/racket/for.rktl
@@ -1156,7 +1156,7 @@
 
 ;; same expansion (from @soegaard)
 (let ()
-  (test #t values
+  (test #t 'same-expansion-for-integer-clause
         (equal? (syntax->datum (expand #'(for ([j 100]) j)))
                 (syntax->datum (expand #'(for ([j (in-range 100)]) j))))))
 
@@ -1165,6 +1165,19 @@
   (local-require (only-in racket (#%datum #%old-datum)))
   (define-syntax-rule (#%datum . x) (#%old-datum . 3))
   (test-sequence [(0 1 2)] 5))
+
+;; for expanded in expression context
+(let ()
+  (define foo%
+    (class object%
+      (super-new)
+      (define/public (abc) '(1 2 3))))
+
+  (test #t class?
+        (class foo%
+          (super-new)
+          (inherit abc)
+          (for ([x (abc)]) #t))))
 
 (err/rt-test (for/list ([x -1]) x))
 (err/rt-test (for/list ([x 1.5]) x))
--- a/pkgs/racket-test-core/tests/racket/for.rktl
+++ b/pkgs/racket-test-core/tests/racket/for.rktl
@@ -1168,21 +1168,16 @@
 
 ;; for expanded in expression context
 (module test-for-expansion racket
-  (provide bar%)
+  (provide foo%)
   (define foo%
     (class object%
       (super-new)
-      (define/public (abc) '(1 2 3))))
-
-  (define bar%
-    (class foo%
-      (super-new)
-      (inherit abc)
-      (for ([x (abc)]) #t))))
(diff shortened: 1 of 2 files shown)


sorawee:
Welp, turns out that full expansion via `local-expand` might not be the right thing to do.

https://github.com/racket/macro-debugger/blob/master/macro-debugger/macro-debugger/view/frame.rkt#L274

fails with `get-menu-bar: unbound identifier`

Partial expansion suffices here, so I will switch to that. That being said, I'm still surprised that the full expansion has this caveat.

pmatos:
> Welp, turns out that full expansion via `local-expand` might not be the right thing to do.
> 
> https://github.com/racket/macro-debugger/blob/master/macro-debugger/macro-debugger/view/frame.rkt#L274
> 
> fails with `get-menu-bar: unbound identifier`
> 
> Partial expansion suffices here, so I will switch to that. That being said, I'm still surprised that the full expansion has this caveat.

This is really interesting and something I would like to understand better. Why is full expansion creating problems here? Do you understand the detials?

sorawee:
I don't know. I think a part of the puzzle is that `for` probably should wrap itself in `#%expression`.

sorawee:
One possibility is that the macro is expanded in a wrong syntax context, during period where `get-menu-bar` is not "registered". Perhaps `for` should wrap itself in `#%expression` and that might help? Will try it next.

rmculpepper:
It's usually bad to discard the results of `local-expand`; in this case, it can cause right-hand sides to be expanded twice (incurring unnecessary costs, duplicating compile-time side effects, etc).

I think you're right that `for` (et al) should delay expansion until they're in expression context. Probably any expression-producing macro that calls `local-expand` (or `syntax-local-value`, etc) should do that, for the same reason. But you should still use partial expansion instead of full expansion here.

pmatos:
Thank you both @sorawee and @rmculpepper  for expanding on the issue.

yjqww6:
If literal numbers could be optimized, I hope list and others too.

mflatt:
@samth @sorawee This change causes a TR test failure: http://drdr.racket-lang.org/55424/cs/racket/share/pkgs/typed-racket-test/succeed/for.rkt

Since it looks like the optimization specifically triggers the failure (and not the `load-expand` or `#%expression`), I imagine that it's best to adjust TR, but I'm not sure.

mflatt:
@jbclements This change causes a test failure in the stepper: http://drdr.racket-lang.org/55423/cs/racket/share/pkgs/htdp-lib/stepper/examples/external-interface-example.rkt

Is that something easy/reasonable to fix in the stepper?

sorawee:
For Typed Racket, it's the same error as:

```
(for/sum: : Integer ((i : Byte (in-range 4))) i)
```

Perhaps `in-range` could be smarter?

Relatedly: Typed Racket in fact tries to add optimizations like this too.

- https://github.com/racket/typed-racket/blob/master/typed-racket-test/optimizer/tests/in-bytes.rkt
- https://github.com/racket/typed-racket/blob/master/typed-racket-test/optimizer/tests/in-list.rkt
- https://github.com/racket/typed-racket/blob/master/typed-racket-test/optimizer/tests/in-range.rkt
- https://github.com/racket/typed-racket/blob/master/typed-racket-test/optimizer/tests/in-string.rkt
- https://github.com/racket/typed-racket/blob/master/typed-racket-test/optimizer/tests/in-vector.rkt
- https://github.com/racket/typed-racket/blob/master/typed-racket-test/optimizer/tests/in-vector2.rkt

and I think these optimizations would now be dead code, because optimizations from this PR will take precedence over Typed Racket's.

For the stepper, it looks like several case splitting in `htdp-lib/stepper/private/annotate.rkt` needs to handle `#%variable-reference`.

samth:
Making those TR tests useful again should be easy, since you could just use a local variable. 

The most useful thing for the actual TR bug would be to come up with what the new expanded code is that doesn't typecheck.

sorawee:
@samth I went through what the type checker computes and found:

```
'(form #<syntax:/Users/sorawee/git/racket/racket/collects/racket/private/for.rkt:2120:52 pos>)
'(expected #f)
(list 'result (tc-results (list (tc-result Byte (Top | Bot) pos)) #f))
'(form #<syntax:/Users/sorawee/git/racket/racket/collects/racket/private/for.rkt:2120:56 inc>)
'(expected #f)
(list 'result (tc-results (list (tc-result One (Top | Bot) inc)) #f))
'(form #<syntax:/Users/sorawee/git/racket/racket/collects/racket/private/for.rkt:2120:19 (#%app unsafe-fx+ pos inc)>)
(list 'expected (tc-results (list (tc-result Byte (Top | Top) -)) #f))
(list 'result (tc-results (list (tc-result Positive-Index (Top | Top) -)) #f))
```

So the issue is that the expanded code contains `(#%app unsafe-fx+ pos inc)`. The type checker only knows that `pos` is between 0 and 255, and it can't guarantee that `(#%app unsafe-fx+ pos inc)` will still be less than 255.

Here's the full expanded code:

```
(#%plain-module-begin
 (#%expression
  (#%expression
   (#%expression
    🔓(let-values ([(🔒start) 🔒(quote 0)] [(🔒end) 🔒(quote 4)] [(🔒inc) 🔒(quote 1)])
        🔒(if (#%app variable-reference-from-unsafe? (#%variable-reference))
              (#%app void)
              (let-values () (#%app check-range start end inc)))
        (#%app
         (letrec-values (((for-loop)
                          (lambda (acc pos)
                            (if (#%expression 🔒(#%app unsafe-fx< pos end))
                                (let-values ([(🔒i) pos])
                                  (if (#%expression 🔒(quote #t))
                                      (let-values ([(acc)
                                                    🔒(let-values (((acc) acc))
                                                        (let-values ([(acc)
                                                                      (let-values ()
                                                                        (let-values (((new)
                                                                                      (let-values ()
                                                                                        i)))
                                                                          (#%app
                                                                           +
                                                                           acc
                                                                           new)))])
                                                          (#%app values acc)))])
                                        (if (if:18 🔒(quote #t)
                                                   (#%expression:19 (#%app not (quote #f)))
                                                   (quote #f))
                                            (#%app
                                             for-loop
                                             acc
                                             🔒(#%app unsafe-fx+ pos inc))
                                            🔒acc))
                                      🔒acc))
                                🔒acc))))
           for-loop)
         (quote 0)
         🔒start))))))
```

One interesting question is, why did it use to work? The answer is that prior this PR, `for` could not specialize the clause, so it generates the generic `(#%app make-sequence '(i) (quote 4))`, and Typed Racket "picks"

```
(-> Univ -Byte         (seq-vals (list -Byte)))
```

as the type of `make-sequence`, so things work out. Here's the fully expanded code in `#lang racket`.

```
(let-values ([(pos->vals
               pos-pre-inc
               pos-next
               init
               pos-cont?
               val-cont?
               all-cont?)
              (#%app make-sequence '(i) (quote 4))])
  (#%app void)
  (#%app
   (letrec-values (((for-loop)
                    (lambda (pos)
                      (if (if pos-cont? (#%app pos-cont? pos) (quote #t))
                          (let-values ([(i all-cont?/pos)
                                        (let-values ([(i)
                                                      (#%app pos->vals pos)])
                                          (#%app
                                           values
                                           i
                                           (if all-cont?
                                               (lambda (pos)
                                                 (#%app all-cont? pos i))
                                               (quote #f))))]
                                       [(pos)
                                        (if pos-pre-inc
                                            (#%app pos-pre-inc pos)
                                            pos)])
                            (if (if val-cont? (#%app val-cont? i) (quote #t))
                                (let-values ([()
                                              🔒(let-values ()
                                                  (let-values ([()
                                                                (let-values ()
                                                                  (let-values ()
                                                                    i)
                                                                  (#%app
                                                                   values))])
                                                    (#%app values)))])
                                  (if (if (if all-cont?/pos
                                              (#%app all-cont?/pos pos)
                                              (quote #t))
                                          (#%app not (quote #f))
                                          (quote #f))
                                      (#%app for-loop (#%app pos-next pos))
                                      🔒(#%app values)))
                                🔒(#%app values)))
                          🔒(#%app values)))))
     for-loop)
   init))
```

sorawee:
One possible hack is to make Typed Racket's `for` wrap useless `identity` around every literal for-clause. Something like:

```
(for/sum ([x (identity 4)]) i)
```

This will inhibit this PR's optimization. Typed Racket's optimizer then can re-optimize it while being able to successfully type check.

samth:
I think the new expanded code should be made to type check. The test should be enough, that's why in-range works at all in Typed Racket.

sorawee:
What test are you talking about?

Also, note that `in-range` doesn't work in this particular case either.

```
#lang typed/racket
(for/sum: : Integer ((i : Byte (in-range 4))) i)
```

results in the same type error, even before this PR.

samth:
So, after looking at this more, the problem is the same as:
```
(let loop ([x : Byte 0])
  (if (< x 5)
      (loop (+ x 1))
      (void)))
```
which also doesn't typecheck. The issue is that there's no type "smaller" than `Byte`, and Byte + 1 isn't always a Byte, of course.

We could add a `ByteMinus1` type, and put that in everywhere, and then this would work, but that's somewhat annoying.

samth:
The other alternatives are somewhat unpalatable, though -- rewrite `for` clauses generally to avoid this, or accept this breaking, or revert this optimization in general.

sorawee:
What about this?

> One possible hack is to make Typed Racket's `for` wrap useless `identity` around every literal for-clause. Something like:
> 
> ```
> (for/sum ([x (identity 4)]) i)
> ```
> 
> This will inhibit this PR's optimization. Typed Racket's optimizer then can re-optimize it while being able to successfully type check.

Might be considered "rewrite for clauses generally", but at least it only affects `#lang typed/racket` and not `#lang racket`.

yjqww6:
Note that the following program type checks
```
#lang typed/racket
(require racket/unsafe/ops)
(let-values ([(start) (quote 0)] [(end) (quote 4)] [(inc) (quote 1)])
  (#%app
   (letrec-values (((for-loop)
                    (lambda (acc [pos : Nonnegative-Fixnum])
                      (if (#%app unsafe-fx< pos end)
                          (let-values ([(i) pos])
                            (if (quote #t)
                                (let-values ([(acc)
                                              (let-values (((acc) acc))
                                                (let-values ([(acc)
                                                              (let-values ()
                                                                (let-values (((new)
                                                                              (let-values ()
                                                                                i)))
                                                                  (#%app
                                                                   +
                                                                   acc
                                                                   new)))])
                                                  (#%app values acc)))])
                                  (if (if (quote #t)
                                          (#%app not (quote #f))
                                          (quote #f))
                                      (#%app
                                       for-loop
                                       acc
                                       (#%app unsafe-fx+ pos inc))
                                      acc))
                                acc))
                          acc))))
     for-loop)
   (quote 0)
   start))
```
By declaring pos is nonnegative, unsafe-fx< would constrain it to Byte type through occurrence typing.
So does attaching a type-label to pos in `in-range` help?

samth:
@sorawee I think that implementation will be painful, and I worry that it would be a pessimization for existing TR programs.

@yjqww6 I'm surprised that works, but it does suggest that if we change the heuristics for guessing the type annotation for `pos` to guess that correctly, we can fix this. That seems promising.

sorawee:
Should we revert this PR while waiting for the fixes? I feel bad that drdr is failing.

samth:
We came up with an idea for a solution, which is a struct property to disable this optimization. I plan to implement it soon.

sorawee:
Typed Racket error is fixed. The stepper's error is still not fixed, IIUC. @jbclements, is it possible to fix the stepper by the 7.9 release? Or should we temporarily revert this PR instead?

sorawee:
The stepper error is fixed. Thanks, @jbclements!

jbclements:
No problem... it's not entirely clear to me that this is the *right* fix; the issue comes up (but only in language levels that the stepper doesn't currently completely support) because of the presence of (%variable-reference) in expanded code. I don't know exactly when (%variable-reference) was added to Racket, but I've never heard of it. After reading the docs, it looks like something I can treat like other opaque values, and that's what I've done. Yes, I'm as old as dirt. :)

jbclements:
Wait, I can't write hash-percent? Ah well.

samth:
Just for historical accuracy, `#%variable-reference` was added in version 300 in 2005, and the version of it seen here was added in 4.1.3 in 2008.

mfelleisen:
Just for historical accuracy, dirt is older than 15 years, indeed 2005 is nothing compared to dirt.

PR #3408, opened 2020-09-21
use gcc build-in functions to perform unsafe fixnum arithmetic

xxyzz:
and eliminate overflow errors from fixnum test when racket bc compiled with
--enable-ubsan(issue #2314)

gcc document: https://gcc.gnu.org/onlinedocs/gcc/Integer-Overflow-Builtins.html

### DIFF
--- a/racket/src/bc/src/numarith.c
+++ b/racket/src/bc/src/numarith.c
@@ -1246,13 +1246,13 @@ static Scheme_Object *fx_abs(int argc, Scheme_Object *argv[])
 #define UNSAFE_DIV(r, x, y) r = x / y
 #define UNSAFE_REM(r, x, y) r = x % y
 #if (__GNUC__ >= 5) || defined(__clang__)
-  #define UNSAFE_PLUS(r, x, y) __builtin_add_overflow(x, y, &r)
+  #define UNSAFE_PLUS(r, x, y)  __builtin_add_overflow(x, y, &r)
   #define UNSAFE_MINUS(r, x, y) __builtin_sub_overflow(x, y, &r)
-  #define UNSAFE_MULT(r, x, y) __builtin_mul_overflow(x, y, &r)
+  #define UNSAFE_MULT(r, x, y)  __builtin_mul_overflow(x, y, &r)
 #else
-  #define UNSAFE_PLUS(x, y, r) r = x + y
-  #define UNSAFE_MINUS(x, y, r) r = x - y
-  #define UNSAFE_MULT(x, y, r) r = x * y
+  #define UNSAFE_PLUS(r, x, y)  r = x + y
+  #define UNSAFE_MINUS(r, x, y) r = x - y
+  #define UNSAFE_MULT(r, x, y)  r = x * y
 #endif
 
 UNSAFE_FX(unsafe_fx_plus, UNSAFE_PLUS, fx_plus, scheme_make_integer(0), )
--- a/racket/src/bc/src/numarith.c
+++ b/racket/src/bc/src/numarith.c
@@ -1228,11 +1228,14 @@ static Scheme_Object *fx_abs(int argc, Scheme_Object *argv[])
   return o;
 }
 
-#if !defined(__GNUC__) && !defined(__clang__)
-#  define __attribute__(x)
+#if __GNUC__ >= 8 || __clang_major__ >= 4
+# define NO_SANITIZE_SIGNED_INTEGER_OVERFLOW \
+  __attribute__((no_sanitize("signed-integer-overflow")))
+#else
+# define NO_SANITIZE_SIGNED_INTEGER_OVERFLOW
 #endif
 #define UNSAFE_FX(name, op, fold, zero_args, PRE_CHECK)                 \
- __attribute__((no_sanitize("signed-integer-overflow")))                \
+ NO_SANITIZE_SIGNED_INTEGER_OVERFLOW                                    \
  static Scheme_Object *name(int argc, Scheme_Object *argv[]) \
  {                                                           \
    intptr_t v;                                                          \
--- a/racket/src/bc/src/numarith.c
+++ b/racket/src/bc/src/numarith.c
@@ -1243,11 +1243,45 @@ static Scheme_Object *fx_abs(int argc, Scheme_Object *argv[])
    return scheme_make_integer(v);                                       \
  }
 
-UNSAFE_FX(unsafe_fx_plus, +, fx_plus, scheme_make_integer(0), )
-UNSAFE_FX(unsafe_fx_minus, -, fx_minus, scheme_false, if (argc == 1) v = -v;)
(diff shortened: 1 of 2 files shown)


xxyzz:
Er, I though Windows also uses gcc.

mflatt:
This is an interesting idea to avoid unspecified behavior.

Since the intent is to avoid errors from ubsan, it would make sense to enable the use of `__builtin` only when it will work:
```
#if (__GNUC__ >= 5) || defined(__clang__)
...
#else
...
#endif
```

Meanwhile, instead of copying and pasting the code from the `UNSAFE_FX` macro, it would be better to change the macro to take an function-shaped arguments instead of an operator-shaped argument, and then that argument (as another macro) can be made to use `__builtin...` or just use an operator.

pmatos:
I haven't checked yet, but won't these `__builtin...` functions create unnecessary slowdown? Shouldn't we attempt instead to avoid the integer overflow in the first place?

racket/src/bc/src/numarith.c
```diff
@@ -1243,12 +1243,46 @@ static Scheme_Object *fx_abs(int argc, Scheme_Object *argv[])
...
 UNSAFE_FX(unsafe_fx_rem, %, fx_rem, scheme_false, )
 
+static Scheme_Object *unsafe_fx_plus(int argc, Scheme_Object *argv[]) {
+  intptr_t v;
+  int i;
+  if (scheme_current_thread->constant_folding) return fx_plus(argc, argv);
+  if (!argc) return scheme_make_integer(0);
+  v = SCHEME_INT_VAL(argv[0]);
+  for (i = 1; i < argc; i++) {
+    __builtin_add_overflow(v, SCHEME_INT_VAL(argv[i]), &v);
```
pmatos, line 1256: Here the function `__builtin_add_overflow` returns a boolean value, which is `true` if there's an overflow. Shouldn't that be checked, if we were to use this? I might be missing how this fixes the issue though.

pmatos:
...and thanks @xxyzz for attempting to solve this long-standing issue.

mflatt:
I doubt that there's any slowdown, since the `__builtin`s will expand to maybe an extra machine instruction or two compared to an operator. Anyway, performance has already gone wrong if control reaches here, because the unsafe operation was presumably meant to be inlined for performance (and wasn't).

The point of the unsafe operations is that they don't try to avoid overflow. If there's a way to tell ubsan "we know — this unsafe function is explicitly meant to have potentially unspecified behavior", then I think that would be just as good as using `__builtin`s.

xxyzz:
I didn't figure out how to modify the macro because `__builtin` functions must be called directly. I'm also kind confused about this issue since they are unsafe procedures and I think it's reasonable to get a runtime error.

pmatos (approved):
> I doubt that there's any slowdown, since the `__builtin`s will expand to maybe an extra machine instruction or two compared to an operator. Anyway, performance has already gone wrong if control reaches here, because the unsafe operation was presumably meant to be inlined for performance (and wasn't).
> 
> The point of the unsafe operations is that they don't try to avoid overflow. If there's a way to tell ubsan "we know — this unsafe function is explicitly meant to have potentially unspecified behavior", then I think that would be just as good as using `__builtin`s.

Thanks for the clarification regarding performance.

I confirmed this and I am actually quite surprised by it.
If you don't check the overflow return result, no extra instructions are generated. Yay!
If you compile with ubsan, no ubsan checks are done because you could check for overflow and didn't so overall it seems to work just fine.

I am happy with this solution. I guess only @mflatt suggestions of merging this into a macro are left.

mflatt:
To modify the macro, I'd do something like this:

```
#define UNSAFE_FX(name, do_op, fold, zero_args, PRE_CHECK)                \
....
     do_op(v, v, SCHEME_INT_VAL(argv[i]));                                  \
....

#if (__GNUC__ >= 5) || defined(__clang__)
# define UNSAFE_ADD(r, x, y) __builtin_add_overflow(x, y, &r)
....
#else
# define UNSAFE_ADD(r, x, y) r = x + y
....
#endif 

UNSAFE_FX(unsafe_fx_plus, UNSAFE_ADD, fx_plus, scheme_make_integer(0), )
....
```

mflatt:
Thanks! One small repair is needed before we can merge: Please make sure that `#` is at the start of the line (since some compilers require that) and uses spaces after `#` and before `define` to indent. I also suggest a comment to record that the point of using `__builtin` is to cooperate with ubsan.

pmatos:
@xxyzz May I ask something else, if you don't mind? The failure in CI is because we expected a failure of ubsan and now we don't fail. Can you change line https://github.com/racket/racket/blob/54367de430b3e6df8958903ed0f81be481725244/.github/workflows/ci-ubsan.yml#L87 to match what we do for Chez?

So the step it should look something like this:
```
run: |
  grep 'runtime error' logs/*.log > runtime-errors_git${{ github.sha }}.log || true
  test ! -s runtime-errors_git${{ github.sha }}.log
```

This would be much appreciated. Thanks.

pmatos:
LGTM - Will wait for @mflatt merge.

Thanks @xxyzz - guess you're the next Call To Racket Champion! :trophy: 
Keep your eyes peeled on the first issue of 2021 as I will have some surprises to ship to the Champions for 2020.

xxyzz:
This is a way to disable sanitizer with function attribute(https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-undefined) but I don't know how to apply it only to Linux and Mac.

pmatos:
> This is a way to disable sanitizer with function attribute(https://clang.llvm.org/docs/UndefinedBehaviorSanitizer.html#disabling-instrumentation-with-attribute-no-sanitize-undefined). It also works on GCC!

That's clang only afaiu. Your solution is gcc/clang compatible so I would go with that.

xxyzz:
It actually works on GCC... I just passed the test on my laptop.

xxyzz:
I guess GCC(https://github.com/gcc-mirror/gcc/tree/master/libsanitizer) uses some code from clang(https://github.com/llvm/llvm-project/tree/master/compiler-rt/lib/ubsan)...

mflatt:
I like the `__attribute__` approach, but it seems that this attribute is recognized only by GCC 8 and up. For Clang, support is at least as old as 3.8; unless you can pin down an earlier version, using Clang 4 as a boundary would be safe and probably good enough.

Instead of defining `__attribute__` away on other platforms, it's better to define some new identifier like `NO_SANITIZE_SIGNED_INTEGER_OVERFLOW` as `__attribute__((no_sanitize("signed-integer-overflow")))` for GCC 8 and up and Clang 4 and up, and define it as empty everywhere else. That way, the choice doesn't affect other potential uses of `__attribute__`.

mflatt:
Thanks!

PR #3418, opened 2020-09-30
Small fixes to enable LTO compilation

pmatos:
To do this in racket/src:
```
  ./configure CFLAGS="-O3 -march=native -flto"
              LDFLAGS="-O3 -march=native -flto"
              --prefix=...
```

One fix deals with variable `errnum` that might be used uninitialized.
The other is to ensure that `boot_file_data` is marked as used
otherwise, it is removed by LTO and ends up crashing build
in `embed-boot.rkt`.

### DIFF
--- a/racket/src/ChezScheme/c/new-io.c
+++ b/racket/src/ChezScheme/c/new-io.c
@@ -78,7 +78,7 @@ static int is_valid_lz4_length(iptr count);
   do { body;                                            \
     if (ok) { flag = 0; }                               \
     else {                                              \
-      INT errnum;                                       \
+      INT errnum = 0;                                   \
       S_glzerror((fd),&errnum);                         \
       S_glzclearerr((fd));                              \
       if (errnum == Z_ERRNO) {                          \
@@ -117,7 +117,7 @@ static int is_valid_lz4_length(iptr count);
   do { body;                                            \
     if (ok) { flag = 0; }                               \
     else {                                              \
-      INT errnum;                                       \
+      INT errnum = 0;                                   \
       S_glzerror((fd),&errnum);                         \
       S_glzclearerr((fd));                              \
       if (errnum == Z_ERRNO) { flag = 1; }              \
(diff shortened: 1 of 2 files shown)


racket/src/ChezScheme/c/new-io.c
```diff
@@ -99,7 +99,7 @@ static int is_valid_lz4_length(iptr count);
   do { body;                                            \
     if (ok) { flag = 0; break; }                        \
     else {                                              \
-      INT errnum;                                       \
+      INT errnum = 0;                                   \
```
mflatt, line 102: Should do the same around lines 81 and 120?

racket/src/cs/c/main.c
```diff
@@ -53,6 +53,7 @@ static void scheme_set_dll_procs(scheme_dll_open_proc open,
 # define embedded_dll_close NULL
 #endif
 
+__attribute__((used))
```
mflatt, line 56: I think we'll need the usual macro approach to include this `__attribute__` only for compilers that support it.

mflatt (changes requested):
Thanks!

racket/src/cs/c/main.c
pmatos: Yes - we need to do that. I haven't checked yet - should the headers where we have `MZ_` stuff like this be shared with Chez code? Or do you want to have a new macro on the Chez side for this?

racket/src/ChezScheme/c/new-io.c
pmatos: We can - although `gcc` didn't complain about those. I will check it anyway.

racket/src/cs/c/main.c
pmatos: We already have this definition under `PRESERVE_IN_EXECUTABLE` in `start/config.inc` so I will just use that.

pmatos:
OK - it should be good now.

racket/src/cs/c/main.c
mflatt: It will have to be a new macro. It's not even the Chez side, exactly, since "src/cs/c/main.c" is outside of Chez Scheme proper (so, it doesn't see the misnamed "version.h", for example, where some macros like this live).

mflatt: Great!

mflatt (approved):

PR #3422, opened 2020-10-02
Disable implicit for-clause optimization with syntax property.

samth:
Related to #3396.

cc @sorawee

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/for.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/for.scrbl
@@ -91,6 +91,14 @@ unreachable, and if the @racket[for] body can no longer reference an
 constructor supports additional sequences that behave like lists and
 streams in this way.
 
+If a @racket[seq-expr] is a quoted literal list, vector, exact integer,
+string, byte string, immutable hash, or expands to such a literal,
+then it may be treated as if a sequence transformer such as
+@racket[in-list] was used, unless the @racket[seq-expr] has a true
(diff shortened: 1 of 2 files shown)


mflatt:
This is good as far as I can tell. It might be better to use the term "quoted" in place of "literal", but I'm not sure.

pkgs/racket-doc/scribblings/reference/for.scrbl
```diff
@@ -91,6 +91,14 @@ unreachable, and if the @racket[for] body can no longer reference an
 constructor supports additional sequences that behave like lists and
 streams in this way.
 
+If a @racket[seq-expr] is a literal list, vector, exact integer,
+string, byte string, immutable hash, or expands to such a literal,
+then it may be treated as if a sequence transformer such as
+@racket[in-list] was used, unless the @racket[seq-expr] has a true
```
sorawee, line 97: Can you add "to improve performance" or something along that line to clarify why doing so is desirable?

racket/collects/racket/private/for.rkt
```diff
@@ -418,6 +419,8 @@
                                               (make-rearm)))]
                     [else (eloop #f #'expanded-rhs)]))]
                [_ (eloop #f #'expanded-rhs)]))]
+          ;; when for:no-implicit-optimization is true
+          [[(id ...) rhs] (eloop #f #'rhs)]
```
sorawee, line 423: I guess another really minor issue is, the name `expanded-rhs` might not be appropriate anymore, since `rhs` here is not expanded.

PR #3427, opened 2020-10-06
Fix compiler warning in 32bits due to size mismatch in ptr cast

pmatos:
Casting a ptr to an integer of a different size causes a warning -
this is only noticeable on 32bits. Fix this by casting the ptr first
to uintptr_t.

### DIFF
--- a/racket/src/ChezScheme/c/gcwrapper.c
+++ b/racket/src/ChezScheme/c/gcwrapper.c
@@ -14,6 +14,8 @@
  * limitations under the License.
  */
 
+#include <stdint.h>
+
 #include "system.h"
 #include "popcount.h"
 
@@ -576,7 +578,7 @@ static void check_pointer(ptr *pp, IBOOL address_is_meaningful, ptr base, uptr s
       }
 
       if (address_is_meaningful) {
-        seginfo *ppsi = MaybeSegInfo(ptr_get_segment(pp));
+        seginfo *ppsi = MaybeSegInfo(ptr_get_segment((uintptr_t)pp));
         if ((ppsi != NULL)
             && (ppsi->generation > psi->generation)
             /* space_data includes stacks, which are always swept */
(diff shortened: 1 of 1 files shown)


racket/src/ChezScheme/c/gcwrapper.c
```diff
@@ -576,7 +578,7 @@ static void check_pointer(ptr *pp, IBOOL address_is_meaningful, ptr base, uptr s
       }
 
       if (address_is_meaningful) {
-        seginfo *ppsi = MaybeSegInfo(ptr_get_segment(pp));
+        seginfo *ppsi = MaybeSegInfo(ptr_get_segment((uintptr_t)pp));
```
mflatt, line 581: The `TO_PTR` macro should be used here: `TO_PTR(pp)`.

racket/src/ChezScheme/c/gcwrapper.c
```diff
@@ -854,7 +856,7 @@ void S_check_heap(aftergc, mcg) IBOOL aftergc; IGEN mcg; {
                     pp1 = TO_VOIDP((ptr)((uptr)TO_PTR(pp1) + size_object(p)));
                   } else {
                     /* skip past unmarked */
-                    pp1 = TO_VOIDP((ptr)((uptr)pp1 + byte_alignment));
+                    pp1 = TO_VOIDP((ptr)((uintptr_t)pp1 + byte_alignment));
```
mflatt, line 859: This should be `(uptr)TO_PTR(pp1)` in place of the original `(uptr)pp1`.

mflatt (changes requested):
In case I miss one going the other direction in the future, the `TO_VOIDP` macro goes from a `ptr` to a `void*`.

Thanks!

pmatos:
Thanks @mflatt 
That makes sense. I will make the changes.

mflatt (approved):

PR #3431, opened 2020-10-09
Fix ARMv7 build using Thumb-2

pmatos:
This ensures that the moveq instruction is inside an IT block.
Backward compatible change - still builds on ARMv6 with Thumb.

Note: This breakage is generally not noticeable on a RPi3 because, although the CPU is ARMv7, it is generally running an ARMv6 OS.


pmatos:
CI on arm64 failed with `error: RPC failed; curl 56 GnuTLS recv error (-54): Error in the pull function.` fetching pb. hum, that's unusual. :(

mflatt (approved):

PR #3442, opened 2020-10-15
flat-contract-with-explanation: fix internal error

LiberalArtist:
Previously, `error` would raise an `exn:fail` with the following message,
because it treated the first of the three strings as a format string
and the other two as its arguments:

> error: format string requires 0 arguments, given 2;
> arguments were:
> " is a procedure, to always escape when called"
> " (by calling raise-blame-error with the arguments it was given"`

### DIFF
--- a/racket/collects/racket/contract/private/misc.rkt
+++ b/racket/collects/racket/contract/private/misc.rkt
@@ -1282,6 +1282,7 @@
          [else
           (accept-or-reason (blame-add-missing-party b neg-party))
           (error 'flat-contract-with-explanation
-                 "expected that result of the first argument, when it"
-                 " is a procedure, to always escape when called"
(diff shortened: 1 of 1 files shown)


LiberalArtist:
It is in keeping with the [Error Message Conventions](https://docs.racket-lang.org/reference/exns.html#(part._err-msg-conventions)) that this error message starts with the name of the complaining function, but the fact that the name of the function is `error` makes it quite confusing. Perhaps `error` should be an exception to the rule?

Additionally, the message did not contain any newlines, which is both unpleasant and inconsistent with the guidelines. Here is the actual `exn-message`:
```racket
"error: format string requires 0 arguments, given 2; arguments were: \" is a procedure, to always escape when called\" \" (by calling raise-blame-error with the arguments it was given\""
```

Issue #3161, opened 2020-05-06
raco test should add support to output to file


pmatos:
In #3159 @samth commented regarding the command line:
```
$ raco test ... 2>&1 | tee log
```

that `raco test` could have in-build support to output to file. This is an enhancement request so we don't forget to discuss it.

yangh:
Do we need to extend the scope this feature to all raco sub commands?

pmatos:
Thanks for giving this a go.
I would say this is most relevant to `raco test`, and that would be good to have for a start.

It looks like you are in the right path and I do not see why we would want/need to change `rackunit`. 
Feel free to open a Draft PR with your current changes and we can work out the details together with a better view of all the changes you're proposing.

samth:
I don't think extending the scope to all subcommands is the right thing here. `raco` commands are intended to be entirely separate -- they don't have to treat options the same way at all. So just adding this to `raco test` is the right way to go.

pmatos:
Thanks for confirming that @samth

sorawee:
FWIW: `raco test` and `rackunit` are orthogonal. There are other testing frameworks too, like:

- https://docs.racket-lang.org/overeasy
- https://docs.racket-lang.org/chk/

So modifying `rackunit` would not suffice.

yangh:
@samth @sorawee 
So we still need to parse the --output <file.name> command line option in the raco.rkt, in this way, no need to change any code else where.

Since raco is widely used, we'd better not to introduce to much changes or make any significant behavior changes.

I'll make a PR as soon as I can, after item 1/2/3 is ready.

PR #3450, opened 2020-10-20
raco: Add support to output to file

yangh:
Example:

$ raco test utils.rkt 
raco test: (submod "utils.rkt" test)
3 tests passed

$ raco test --cmd-output output.txt utils.rkt

$ cat output.txt 
raco test: (submod "utils.rkt" test)
3 tests passed

Fixed issue #3161

Signed-off-by: Hong Yang <yangh.cn@gmail.com>

### DIFF
--- a/racket/collects/raco/raco.rkt
+++ b/racket/collects/raco/raco.rkt
@@ -1,7 +1,8 @@
 #lang racket/base
 (require "command-name.rkt"
          "all-tools.rkt"
-         racket/string)
+         racket/string
+         racket/list)
 
 (module test racket/base)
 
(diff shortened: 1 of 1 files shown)


racket/collects/raco/raco.rkt
```diff
@@ -38,13 +39,48 @@
   ((executable-yield-handler) result) ; to enable GUI commands
   (exit result))
 
+(define (parse-cmd-output-option cmdline)
+  (let* ([tout-idx (index-of cmdline "--cmd-output")]
+         [cmd-is-not-help (not (and (pair? cmdline)
+                                    (equal? "help" (car cmdline))))])
```
sorawee, line 45: I thought that at this point, `raco help` is converted to `raco ... --help` already, no?

racket/collects/raco/raco.rkt
```diff
@@ -38,13 +39,48 @@
...
+    (if (and tout-idx
+             (< tout-idx (sub1 (length cmdline)))
+             cmd-is-not-help)
+        (string->path (list-ref cmdline (add1 tout-idx)))
```
sorawee, line 49: This is unsafe. Users might miss the corresponding parameter. E.g., `raco test --cmd-output`. We want a proper error in that case, not a `list-ref` error.

sorawee: Oh, I just saw that there's a check above that prevents this error.

But that doesn't really solve the problem. There should still be a proper error.

racket/collects/raco/raco.rkt
```diff
@@ -38,13 +39,48 @@
...
+        (append (take cmdline tout-idx)
+                (drop cmdline (add1 tout-file-idx))))))
+
+(define (open-cmd-output-port cmd-output-file)
+  (if (path? cmd-output-file)
+      (with-handlers ([exn:fail:filesystem:errno?
+                       (lambda (e)
+                         (displayln e)
+                         (displayln "Failed to open output file, fallback to standard output")
+                         (current-output-port))])
```
sorawee, line 66: In general I think it's better to write error to `current-error-port`.

I'm not sure if I like the fallback behavior. Users intentionally include `--cmd-output` because they expect it to output to a file. Letting the program continue running in the fallback mode defies that expectation. Perhaps just letting the exception getting thrown might be better?

sorawee:
Wouldn't it be better to modify `pkgs/compiler-lib/compiler/commands/test.rkt` so that it only applies to `raco test`?

yangh:
I'll make another PR based on the comments from @sorawee

PR #3451, opened 2020-10-21
raco: test: Add support to output to file

yangh:
Example:

 raco test --output output.txt utils.rkt

Fixed issue #3161

### DIFF
--- a/pkgs/compiler-lib/compiler/commands/test.rkt
+++ b/pkgs/compiler-lib/compiler/commands/test.rkt
@@ -151,6 +151,7 @@
       (define stdin (if empty-input?
                         (open-input-bytes #"")
                         (current-input-port)))
+
       (unless quiet?
         (when responsible
           (fprintf stdout "raco test:~a @(test-responsible '~s)\n"
@@ -1135,12 +1136,16 @@
            [(positive? (summary-failed sum1)) 1]
            [else 0])))
 
- (if (string? default-output-file)
-     (call-with-output-file*
-      (string->path default-output-file)
-      (lambda (o)
-        (parameterize ([current-output-port o]
-                       [current-error-port o])
+ ;; Save the stdout/stderr into a file.
+ (define (test-main-with-output file-name)
+   (call-with-output-file*
+      (string->path file-name)
+      (lambda (out)
+        (parameterize ([current-output-port out]
+                       [current-error-port  out])
           (test-main)))
-       #:exists 'truncate)
+       #:exists 'truncate))
+
+ (if (string? default-output-file)
+     (test-main-with-output default-output-file)
      (test-main)))
--- a/pkgs/compiler-lib/compiler/commands/test.rkt
+++ b/pkgs/compiler-lib/compiler/commands/test.rkt
@@ -1100,7 +1100,7 @@
   (set! default-error-port default-output-port)]
  #:args file-or-directory-or-collects-or-pkgs
  (parameterize ([current-output-port default-output-port]
-                  [current-error-port default-error-port])
+                [current-error-port default-error-port])
    (define file-or-directory
      (maybe-expand-package-deps file-or-directory-or-collects-or-pkgs))
    (unless (= 1 (length file-or-directory))
--- a/pkgs/compiler-lib/compiler/commands/test.rkt
+++ b/pkgs/compiler-lib/compiler/commands/test.rkt
@@ -43,6 +43,8 @@
 
 (define default-timeout #f) ; #f means "none"
 (define default-mode #f) ; #f => depends on how many files are provided
+(define default-output-port (current-output-port))
+(define default-error-port (current-error-port))
 
 (define single-file? #t)
 
@@ -150,7 +152,6 @@
       (define stdin (if empty-input?
                         (open-input-bytes #"")
                         (current-input-port)))
-
       (unless quiet?
         (when responsible
           (fprintf stdout "raco test:~a @(test-responsible '~s)\n"
@@ -1092,41 +1093,48 @@
  [("--table" "-t")
   "Print a summary table"
   (set! table? #t)]
+ #:once-each
+ [("--output" "-o") filename
+  "Save stdout and stderr to file"
+  (set! default-output-port (open-output-file (string->path filename) #:exists 'truncate))
+  (set! default-error-port default-output-port)]
  #:args file-or-directory-or-collects-or-pkgs
  (let ()
-   (define file-or-directory
-     (maybe-expand-package-deps file-or-directory-or-collects-or-pkgs))
-   (unless (= 1 (length file-or-directory))
-     (set! single-file? #f))
-   (when (and (eq? configure-runtime 'default)
-              (or (and (not single-file?)
-                       (not (memq default-mode '(process place))))
-                  (not (null? submodules))))
-     (set! configure-runtime #f))
-   (define sum
-     ;; The #:sema argument everywhre makes tests start
-     ;; in a deterministic order:
-     (map/parallel (lambda (f #:sema s)
-                     (test-top f
-                               #:check-suffix? check-top-suffix?
-                               #:sema s))
-                   file-or-directory
-                   #:sema (make-semaphore)))
-   (when table?
-     (display-summary sum))
-   (unless (or (eq? default-mode 'direct)
-               (and (not default-mode) single-file?))
-     ;; Re-log failures and successes, and then report using `test-log`.
-     ;; (This is awkward; is it better to not try to use `test-log`?)
-     (for ([s (in-list sum)])
-       (for ([i (in-range (summary-failed s))])
-         (test-log! #f))
-       (for ([i (in-range (- (summary-total s)
-                             (summary-failed s)))])
-         (test-log! #t))))
-   (test-log #:display? #t #:exit? #f)
-   (define sum1 (call-with-summary #f (lambda () sum)))
-   (exit (cond
-           [(positive? (summary-timeout sum1)) 2]
-           [(positive? (summary-failed sum1)) 1]
-           [else 0]))))
+   (parameterize ([current-output-port default-output-port]
+                  [current-error-port default-error-port])
+     (define file-or-directory
+       (maybe-expand-package-deps file-or-directory-or-collects-or-pkgs))
+     (unless (= 1 (length file-or-directory))
+       (set! single-file? #f))
+     (when (and (eq? configure-runtime 'default)
+                (or (and (not single-file?)
+                         (not (memq default-mode '(process place))))
+                    (not (null? submodules))))
+       (set! configure-runtime #f))
+     (define sum
+       ;; The #:sema argument everywhre makes tests start
+       ;; in a deterministic order:
+       (map/parallel (lambda (f #:sema s)
+                       (test-top f
+                                 #:check-suffix? check-top-suffix?
+                                 #:sema s))
+                     file-or-directory
+                     #:sema (make-semaphore)))
+     (when table?
+       (display-summary sum))
+     (unless (or (eq? default-mode 'direct)
+                 (and (not default-mode) single-file?))
+       ;; Re-log failures and successes, and then report using `test-log`.
+       ;; (This is awkward; is it better to not try to use `test-log`?)
+       (for ([s (in-list sum)])
+         (for ([i (in-range (summary-failed s))])
+           (test-log! #f))
+         (for ([i (in-range (- (summary-total s)
+                               (summary-failed s)))])
+           (test-log! #t))))
+     (test-log #:display? #t #:exit? #f)
+     (define sum1 (call-with-summary #f (lambda () sum)))
+     (exit (cond
+             [(positive? (summary-timeout sum1)) 2]
+             [(positive? (summary-failed sum1)) 1]
+             [else 0])))))
--- a/pkgs/compiler-lib/compiler/commands/test.rkt
+++ b/pkgs/compiler-lib/compiler/commands/test.rkt
@@ -1098,7 +1098,7 @@
   "Save stdout and stderr to file"
   (set! default-output-file filename)]
  #:args file-or-directory-or-collects-or-pkgs
- (define (test-main)
+ (define (test)
    (define file-or-directory
      (maybe-expand-package-deps file-or-directory-or-collects-or-pkgs))
    (unless (= 1 (length file-or-directory))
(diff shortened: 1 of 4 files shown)


yangh:
Tested with the following script:

```
$ cat test-raco-test.sh 
#!/bin/sh

do_test () {
	rm -f output.txt

	echo "> Test: $@"
	$@
	echo "> Ret: $?"
	echo "> Output lines: `wc -l output.txt`"
	cat output.txt
        echo ""
}

do_test raco test utils.rkt 
do_test raco test --output output.txt utils.rkt 
do_test raco test --output output.txt --process utils.rkt 
do_test raco test --output output.txt --place utils.rkt 
do_test raco test --output output.txt --jobs 2 utils.rkt 

do_test raco test --output
do_test raco test --output /tmpx/none-exist.txt utils.rkt 

```
Output
```
$ sh test-raco-test.sh 
> Test: raco test utils.rkt
raco test: (submod "utils.rkt" test)
3 tests passed
> Ret: 0
wc: output.txt: No such file or directory
> Output lines: 
cat: output.txt: No such file or directory

> Test: raco test --output output.txt utils.rkt
> Ret: 0
> Output lines: 2 output.txt
raco test: (submod "utils.rkt" test)
3 tests passed

> Test: raco test --output output.txt --process utils.rkt
> Ret: 0
> Output lines: 2 output.txt
raco test: (submod "utils.rkt" test)
3 tests passed

> Test: raco test --output output.txt --place utils.rkt
> Ret: 0
> Output lines: 2 output.txt
raco test: (submod "utils.rkt" test)
3 tests passed

> Test: raco test --output output.txt --jobs 2 utils.rkt
> Ret: 0
> Output lines: 2 output.txt
raco test: 0 (submod "utils.rkt" test)
3 tests passed

> Test: raco test --output
raco test: the "--output" option needs 1 argument, but 0 provided
> Ret: 1
wc: output.txt: No such file or directory
> Output lines: 
cat: output.txt: No such file or directory

> Test: raco test --output /tmpx/none-exist.txt utils.rkt
open-output-file: cannot open output file
  path: /tmpx/none-exist.txt
  system error: No such file or directory; errno=2
  context...:
   /usr/share/racket/collects/racket/cmdline.rkt:153:76
   /usr/share/racket/collects/racket/cmdline.rkt:464:10: call-handler
   "/usr/share/racket/pkgs/compiler-lib/compiler/commands/test.rkt": [running body]
   temp35_0
   for-loop
   run-module-instance!
   "/usr/share/racket/collects/raco/raco.rkt": [running body]
   temp35_0
   for-loop
   run-module-instance!
   "/usr/share/racket/collects/raco/main.rkt": [running body]
   temp35_0
   for-loop
   run-module-instance!
   perform-require!
> Ret: 1
wc: output.txt: No such file or directory
> Output lines: 
cat: output.txt: No such file or directory

```

pkgs/compiler-lib/compiler/commands/test.rkt
```diff
@@ -1092,8 +1093,14 @@
  [("--table" "-t")
   "Print a summary table"
   (set! table? #t)]
+ #:once-each
+ [("--output" "-o") filename
+  "Save stdout and stderr to file"
+  (set! default-output-port (open-output-file (string->path filename) #:exists 'truncate))
```
sorawee, line 1099: I think you need to close the port at some point, right?

It might be worth it to switch `open-output-file` to `with-output-to-file`. That is, make the code below (`(let () ...)`) a thunk, and pass this thunk to an invoker. When `--output` is not set, the invoker will invoke the thunk right away. However, if `--output` is set, then the invoker will supply the thunk to `with-output-to-file`.

yangh: Good idea, the port should be closed.

I assume that if the user specified a --output when invoking raco test, the stderr also needs to be pipelined into the same file, in this case, `with-output-to-file` doesn't work as far as I know.

sorawee: You can `parameterize` the `current-error-port` to be `current-output-port`.

yangh: `with-output-to-file` doesn't work here, I implemented it with `call-with-output-file*`.

sorawee:
Just by reading the code (I have not tried running it), this looks good to me. Thanks!

yangh:
One more test case for the error message
```
$ raco test --output output.txt utils.rkt 
$ cat output.txt 
raco test: (submod "utils.rkt" test)
--------------------
ERROR
name:       check-false
location:   utils.rkt:20:2

hex->dec: arity mismatch;
  expected: 1
  given: 2
  arguments...:
   "123"
   '()
--------------------
1/4 test failures
```

pkgs/racket-doc/scribblings/raco/test.scrbl
```diff
@@ -179,15 +179,19 @@ The @exec{raco test} command accepts several flags:
...
  }
 ]
 
 @history[#:changed "1.1" @elem{Added @DFlag{heartbeat}.}
          #:changed "1.4" @elem{Changed recognition of module suffixes to use @racket[get-module-suffixes],
                                which implies recognizing @filepath{.ss} and @filepath{.rkt}.}
          #:changed "1.5" @elem{Added @DPFlag{ignore-stderr}.}
-         #:changed "1.6" @elem{Added @DPFlag{arg} and @DPFlag{args}.}]
+         #:changed "1.6" @elem{Added @DPFlag{arg} and @DPFlag{args}.}
+         #:changed "1.7" @elem{Added @DFlag{output} or @Flag{o}.}]
```
pmatos, line 194: I would expect this connective to be `and`, as in `Added output and o`.

pkgs/compiler-lib/compiler/commands/test.rkt
```diff
@@ -1092,8 +1093,12 @@
  [("--table" "-t")
   "Print a summary table"
   (set! table? #t)]
+ #:once-each
+ [("--output" "-o") filename
+  "Save stdout and stderr to file"
+  (set! default-output-file filename)]
  #:args file-or-directory-or-collects-or-pkgs
- (let ()
+ (define (test-main)
```
pmatos, line 1101: What's the reason behind calling this `test-main`, it's quite surprising to me as it is not a test.

pkgs/compiler-lib/compiler/commands/test.rkt
```diff
@@ -1129,4 +1134,18 @@
    (exit (cond
            [(positive? (summary-timeout sum1)) 2]
            [(positive? (summary-failed sum1)) 1]
-           [else 0]))))
+           [else 0])))
+
+ ;; Save the stdout/stderr into a file.
+ (define (test-main-with-output file-name)
```
pmatos, line 1140: Same `test-...`?

pmatos (changes requested):
Just a couple of nit picks.

pmatos:
Given @samth initially suggested this change, I am adding him to reviewers.

pkgs/racket-doc/scribblings/raco/test.scrbl
yangh: Good catch, history has different means to the itemize.

pkgs/compiler-lib/compiler/commands/test.rkt
yangh: Any suggestions for the name? (main)?

My thoughts, the test.rkt provides the main function of test, specificly, `raco test`, that's why I calling it as `test-main`

pmatos:
Assign myself to merge and close once all reviews are in.

pmatos (approved):
LGTM

pkgs/racket-doc/scribblings/raco/test.scrbl
```diff
@@ -179,15 +179,19 @@ The @exec{raco test} command accepts several flags:
         as a whitespace-delimited list of arguments to add. To specify
         multiple arguments using this flag within a typical shell,
         @nonterm{arguments} must be
-        enclosed in quotation marks.
+        enclosed in quotation marks.}
+
+ @item{@DFlag{output} or @Flag{o} @nonterm{filename}
+       --- Save all stdout and stderr output into @nonterm{filename}.
```
samth, line 185: This should mention that it overwrites `filename` if it exists.

samth:
Looks good past that one comment.

pkgs/racket-doc/scribblings/raco/test.scrbl
```diff
@@ -179,15 +179,19 @@ The @exec{raco test} command accepts several flags:
...
  }
 ]
 
 @history[#:changed "1.1" @elem{Added @DFlag{heartbeat}.}
          #:changed "1.4" @elem{Changed recognition of module suffixes to use @racket[get-module-suffixes],
                                which implies recognizing @filepath{.ss} and @filepath{.rkt}.}
          #:changed "1.5" @elem{Added @DPFlag{ignore-stderr}.}
-         #:changed "1.6" @elem{Added @DPFlag{arg} and @DPFlag{args}.}]
+         #:changed "1.6" @elem{Added @DPFlag{arg} and @DPFlag{args}.}
+         #:changed "1.7" @elem{Added @DFlag{output} and @Flag{o}.}]
```
yangh, line 194: Which revison should be refered in this history? the racket version? 7.9.0.3? @pmatos

sorawee: It should be 1.8. It should correspond to https://github.com/racket/racket/blob/master/pkgs/compiler-lib/info.rkt.

sorawee:
Another thing that might be worth doing is adding a test to `pkgs/compiler-test/tests/compiler/test/` to test this new feature.

yangh:
> Another thing that might be worth doing is adding a test to `pkgs/compiler-test/tests/compiler/test/` to test this new feature.

Implemented in the last patch:

Check the result when --output <file> added to raco test:
1. Make sure the output to standard is empty.
2. Make sure the output into to file is what's excepted.

pkgs/racket-doc/scribblings/raco/test.scrbl
```diff
@@ -179,15 +179,20 @@ The @exec{raco test} command accepts several flags:
         as a whitespace-delimited list of arguments to add. To specify
         multiple arguments using this flag within a typical shell,
         @nonterm{arguments} must be
-        enclosed in quotation marks.
+        enclosed in quotation marks.}
+
+ @item{@DFlag{output} or @Flag{o} @nonterm{file}
+       --- Save all stdout and stderr output into @nonterm{file}.
+       The target @nonterm{file} will be overwrote if it exists already.
```
pmatos, line 186: nit pick : s/overwrote/overwritten

pmatos (changes requested):

pkgs/racket-doc/scribblings/raco/test.scrbl
yangh: Thanks.

yangh:
> Another thing that might be worth doing is adding a test to `pkgs/compiler-test/tests/compiler/test/` to test this new feature.

Hi sorawee

I've implemented test cases for standard output, but for the stderr output, I have no idea how to test it gracefully, some options in my mind:

1. add an output-error.rkt, which includes a test case that could always cause error.
2. write some test code (with error) into a tempary file at runtime, run raco test with it.

for option 1, is there a side effect of it? such would cause racoket to build or test error during the continuous integration or installation of distribution?

pkgs/compiler-test/tests/compiler/test/runtime.rkt
```diff
@@ -1,33 +1,53 @@
...
+               post-check-result)
+    (error 'runtime "test failed\n  module: ~s~a\n  expected: ~s\n  got: ~s\n  post-check: ~s"
            mod
            (if (null? extra-args) "" (format "\n  extra args: ~s" extra-args))
            expect
-           last-line)))
+           last-line
+           post-check-result)))
+
+(define (output-file-check except file)
```
sorawee, line 34: Do you mean to name the variable `expect`?  

And again, it's probably better to close the port, or use `with-input-from-file` / `call-with-input-file*`.

pkgs/compiler-test/tests/compiler/test/runtime.rkt
```diff
@@ -1,33 +1,53 @@
...
-(define (try mode mod expect . extra-args)
+;; Return last line of the input, or #f if it's empty.
+(define (input-last-line input)
+  (for/fold ([prev #f]) ([s (in-lines input)])
+      (if (or (eof-object? s)
+              (equal? s "1 test passed"))
+          prev
+          s)))
+
+(define (try mode mod expect post-check . extra-args)
```
sorawee, line 16: One minor nit: there's really no need to create a new parameter `post-check`. You can reuse the parameter `expect`, since `expect` and `post-check` are completely disjoint. Then, you can check if `expect` is a procedure, and if it is, apply it to the output. Otherwise, use `equal?` to test it against the output.

sorawee: (And add a comment above that `expect` is either a procedure or a string.)

sorawee:
> for option 1, is there a side effect of it? such as to cause build or test error during the continuous integration or installation of distribution of racket?

It's probably fine for GitHub Actions, since the newly added file is only called by `runtime.rkt` which discards the exit code. It's probably NOT fine for [DrDr](http://drdr.racket-lang.org/) since DrDr seems to run the test module of every file. I have no idea how DrDr works, so it's probably the best to ask @samth.

samth:
How DrDr works is described here: http://drdr.racket-lang.org/help

A solution here is to have a `test2` submodule that prints to stderr, and use the `--submodule` flag to `raco test`.

pkgs/compiler-test/tests/compiler/test/runtime.rkt
yangh: My bad :( Will be fixed in the new commit.

pkgs/compiler-test/tests/compiler/test/runtime.rkt
yangh: Thanks, sorawee, fixed in the new commit.

yangh:
> How DrDr works is described here: http://drdr.racket-lang.org/help
> 
> A solution here is to have a `test2` submodule that prints to stderr, and use the `--submodule` flag to `raco test`.

Thanks, now it's tesred via sub module.

yangh:
@sorawee @pmatos Could you please help to review the last patch again? Thanks

sorawee:
It looks good to me (I actually have more nits, but I don't think it will be worth it / productive to change the PR any further).

pmatos:
LTGM as well. I am happy to merge this.

pmatos:
Thanks for the contribution, time and patience @yangh 
And with this you are Racket News Issue 40 Call to Racket Champion. Well done! :)

yangh:
Thanks, @sorawee @pmatos @samth, really appreciate your time and patience to review my dumb/ugly code... If any further issue found, I'd like to help to fix it.

PR #3460, opened 2020-10-23
make char-numeric? docs more specific

stchang:
It now matches the docs for other `char-X` functions.

I used the categories here: http://www.fileformat.info/info/unicode/category/index.htm

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/chars.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/chars.scrbl
@@ -187,8 +187,8 @@ Returns @racket[#t] if @racket[char]'s Unicode general category is
 
 @defproc[(char-numeric? [char char?]) boolean?]{
 
-Returns @racket[#t] if @racket[char]'s Unicode general category is
-@UCat{Nd}, @UCat{Nl}, or @UCat{No}, @racket[#f] otherwise.}
+Returns @racket[#t] if @racket[char] has the Unicode ``Numeric_Value''
+property.}
 
 @defproc[(char-symbolic? [char char?]) boolean?]{
 
--- a/pkgs/racket-doc/scribblings/reference/chars.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/chars.scrbl
@@ -187,8 +187,8 @@ Returns @racket[#t] if @racket[char]'s Unicode general category is
 
 @defproc[(char-numeric? [char char?]) boolean?]{
 
-Returns @racket[#t] if @racket[char] has the Unicode ``Numeric''
-property.}
+Returns @racket[#t] if @racket[char]'s Unicode general category is
+@UCat{Nd}, @UCat{Nl}, or @UCat{No}, @racket[#f] otherwise.}
 
 @defproc[(char-symbolic? [char char?]) boolean?]{
 


mflatt:
The Unicode `Numeric` property is not defined just in terms of categories. The output of

```
#lang racket/base

(for* ([r (in-list (make-known-char-range-list))]
       [i (in-range (car r) (add1 (cadr r)))])
  (define c (integer->char i))
  (define cat (char-general-category c))
  (define n? (char-numeric? c))
  (define cat? (and (memq cat '(nd nl no)) #t))
  (unless (eq? n? cat?)
    (printf "~s ~s\n" (string c) cat)))      
```

shows the special cases:

```
"參" lo
"拾" lo
"兩" lo
"零" lo
"六" lo
"陸" lo
"什" lo
"廾" lo
```

stchang:
Ah I see. Is this following some standard, or is it Racket-specific?

mflatt:
It's following R6RS.

stchang:
Ok thanks. This came up when I was updating RacketScript to match Racket. Right now Racketscript `char-numeric?` is only returning #t for Nd, Nl, and No categories. Is there an easy way to compute the other cases, e.g., some other category? searching for "unicode numeric property" doesnt seem to turn up any formal specification

mflatt:
See "Numeric" here: https://unicode.org/reports/tr44/

It seems that "Numeric" by itself isn't a property, but "Numeric_Type" and "Numeric_Value" are. I don't know whether the Unicode spec evolved or whether R6RS used "Numeric" when it should more property say "Numeric_Value" (probably the latter). Racket BC's implementation is specifically derived from field 6, where a non-empty field value means that the code point counts as numeric.

stchang:
Thanks, that's exactly what I was looking for. 

Sadly, js regexps dont seem to be 100% compliant with the unicode standard because they report that as "Invalid property name", so I'll have to approximate for now.

Should I update the Racket docs to say "Numeric_Value" instead of "Numeric"?

glebm:
Thanks @mflatt. Do you remember why it is defined like this?

The `Numeric_Value` property is ambiguous for most CJ characters above due to high degree of homography in these languages.

For example, `六` *could* be a number but it highly depends on the context.

`char-numeric?` returning `#t` for something like `六` is unexpected.
This method isn't helpful when parsing CJ numbers either .

I think the definition should be changed to `Nd` `Nl` `No` to be less surprising.

mflatt:
In case it's a source of confusion: There appears to be some implicit conversion between the program and pasting into GitHub or copying back out, at least on my machine. The character "六" above is not supposed to be U+516D (the regular "six" character, which does not have the Numeric_Value property), but U+F9D1. I get U+516D when I copy and paste back into Racket in a terminal window.

I'm not at all sure of the intent of having a separate code point in this case, but I doubt that the code point U+F9D1 was given the numeric value 6 accidentally by the Unicode authors. :) So, if the Unicode standard says that U+F9D1 should be treated like all the other unusual numbers, I'm inclined to stick with that for `char-numeric?`. For any application that wants the union of Nl, Nd, and No, there's always `char-general-category`.

Meanwhile, I am now mostly convinced that the R6RS use of "Numeric" is a typo, and we should fix the Racket documentation to say "Numeric_Value".

glebm:
> So, if the Unicode standard says that U+F9D1 should be treated like all the other unusual numbers

The standard does not say this. 

> I doubt that the code point U+F9D1 was given the numeric value 6 accidentally by the Unicode authors

This is copied directly from UniHan (`kPrimaryNumeric=6` in this case). See https://unicode.org/reports/tr38/#N1024D:

> Finally, we have three fields, kAccountingNumeric, kOtherNumeric, and kPrimaryNumeric to indicate the numerical values an ideograph **may have**. Traditionally, ideographs were used both for numbers and words, and so **many ideographs have (or can have)** numeric values. The various kinds of numeric values are specified by these three fields.

(bolding mine)

(U+F9D1 is a compatibility code point for U+516D, there is no semantic difference, it's a writing variant))

glebm:
Very subjective statements:

* The current definition of `char-numeric?` isn't useful for anything.
* This also makes it dangerous, because most people who work with Unicode would assume this to be a shorthand for `N[ldo]`.
* I'm not aware of any other library that assigns meaning to the presence of `Numeric_Value` attribute.
* I'm willing to bet that most uses of `char-numeric?` in existing code are incorrect!

glebm:
I was wrong on the third point, Python `str.isnumeric` also defines it like this
https://docs.python.org/3/library/stdtypes.html?highlight=isnumeric#str.isnumeric

mflatt:
You're right that I was too glib and imprecise before. I meant that Unicode defines the notion of Numeric_Value, and it's not unreasonable for a predicate to reflect the concept of having a Numeric_Value property. Apparently, U+F9D1 and U+516D are not entirely equivalent, since Unicode makes a distinction between them with respect to Number_Value.

I also had in mind that `char-numeric?` defined in terms of general categories seems unlikely to be safer or more useful than one defined in terms of Number_Value. I have only used `char-numeric?` myself when restricted to ASCII characters. It's difficult for me to imagine a use that needs `(char-numeric? #\¾)` with U+BE and `(char-numberic? #\〸)`  with U+3038 to be true, but not `(char-numberic? #\六)` with U+F9D1. Maybe we just agree that `char-numeric?` isn't all that useful — but I'm far from convinced that one based on general categories is more useful, and I'm leery of introducing a gratuitous incompatibility with R6RS.

stchang:
I've updated this pr

mflatt:
Thanks! I had meanwhile updated the docs to define it in terms of ``Numeric_Type''. I'm happy to change again, though, if it's not right.

glebm:
After seeing the Python function and the amount of confusion it's caused on their bug tracker, I'm afraid I've been barking up the wrong tree. This will need to go through the Unicode Technical Committee.

The Racket and Python functions provide good evidence of issues with the Unicode standard.

Do you happen to have a reference for R6RS specification of `char-numeric?`? I could only find 2 matches in the doc (http://www.r6rs.org/final/r6rs.pdf) but they don't specify the function.

mflatt:
The R6RS specification is here: https://docs.racket-lang.org/r6rs/r6rs-lib-std/r6rs-lib-Z-H-2.html?q=char-numeric#node_idx_24

In PDF form, you'd need the Libraries document: http://www.r6rs.org/final/r6rs-lib.pdf

stchang:
Ah ok, didnt realize you already fixed it. Closing this then

PR #3461, opened 2020-10-23
Use locale.h if not xlocale.h header

pmatos:
Otherwise we have a missing definition for locale_t when
RKTIO_USE_XLOCALE but !RKTIO_USE_XLOCALE_HEADER.

### DIFF
--- a/racket/src/rktio/rktio_private.h
+++ b/racket/src/rktio/rktio_private.h
@@ -14,6 +14,8 @@
 #ifdef RKTIO_USE_XLOCALE
 # ifdef RKTIO_USE_XLOCALE_HEADER
 #  include <xlocale.h>
+# else
+#  include <locale.h>
 # endif
 #endif
 


pmatos:
ARM64 BC does not compile due to missing `locale_t`. I was wondering what's the difference... Well, `pthreads` define `locale_t` on X86_64. But why are we not enabling pthreads on ARM64? Well, it turns out we don't have parallel by default on arm64.

pmatos:
@mflatt any reason we don't enable parallel by default for aarch64 on bc? we don't have jit, but any reason not to enable places support?

mflatt (approved):

mflatt:
I'm not sure the right memory fences are in place for places on Arm (or whether any are needed).

pmatos:
> I'm not sure the right memory fences are in place for places on Arm (or whether any are needed).

Lets run some tests and see what explodes. :)

PR #3463, opened 2020-10-24
Do not run ARM tests on PR

pmatos:
Due to the speed and volume of available ARM HW reduce tests on
ARM during PRs and leave them for push-only. Run ubsan only on a
daily basis.


samth (approved):

pmatos:
Thanks @samth 
Just waiting to see if the ARM fast build is fine.

PR #3466, opened 2020-10-27
Fix CFLAGS ordering to ensure users CFLAGS take precedence

pmatos:


mflatt (approved):
I'm fine with this change. But I can't help thinking that the real problem is that `ChezScheme/configure` and all the `Mf-...` files have hardwired flags, and those should be reorganized so that they're merely defaults.

pmatos:
> I'm fine with this change. But I can't help thinking that the real problem is that `ChezScheme/configure` and all the `Mf-...` files have hardwired flags, and those should be reorganized so that they're merely defaults.

That's a good point and I agree. I will merge this for now and open an issue so that we don't forget about this.

PR #3472, opened 2020-10-29
Rename define-simple-macro to define-syntax-parse-rule

sorawee:
This PR is based on @lexi-lambda's #1310.
It resolves the conflicts in the mentioned PR and
adjusts the version so that it's up-to-date.

Closes #1310

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/define.scrbl
@@ -39,7 +39,7 @@ Defines a macro named @racket[macro-id]; equivalent to the following:
 (fn2 a #:b 'c)
 ]
 
-@history[#:added "7.9.0.3"]
+@history[#:added "7.9.0.22"]
 }
 
 @defform[(define-syntax-parser macro-id parse-option ... clause ...+)]{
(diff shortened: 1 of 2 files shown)


sorawee:
Can we merge this for v8.0?

samth:
I think we need @rmculpepper's thoughts here.

rmculpepper:
I'm ambivalent. In my view, this replaces one okay-but-not-great name with another okay-but-not-great name. On the other hand, other people seem to like this new name better. So if people really think this is an improvement, then I'm willing to go along with the name change.

jackfirth:
I think this is a solid improvement. I agree that the new name is okay-but-not-great, but I think the current name is _really_ bad, to the point where it makes the API come off as untrustworthy. So I'd really like to see this merged in for 8.0. I think @rmculpepper should be the final decider on whether this get merged, so I'll assign it to him.

Metaxal:
IMHO,
Pros:
* consistent name (big pro). 

Cons:
* there's no `syntax-parse-rules` (by analogy to `syntax-rules`), only `syntax-parse`. So not fully consistent either.
* pretty long name, not a good selling point for making something the default.
* not very informative either.  Imagine you come to Racket, and you've heard about its powerful macro system. The first thing you're told to learn is `define-syntax-parse-rule`. "Uh, what the heck does that even mean?" You can't even tell how the words should be grouped: is it (define-syntax)-(parse-rule) or define-(syntax-parse)-rule? That's a real put-off. By contrast `define-simple-macro` is very friendly, despite its naming inconsistency.

PR #3473, opened 2020-10-29
Improve (sinh x) : x around 0

bdeket:
New PR because I accidentally deleted my old branch/PR #3323

now using Horner scheme, since it's slightly more accurate.


samth:
This looks good to me. Any thoughts, @mflatt @pavpanchekha @soegaard ?

mflatt:
I don't feel very well informed — I had to google "hyperbolic sine" again — but this looks good to me.

pavpanchekha:
Yeah, this is good. I say we merge.

@bdeket's work is really great to find these issues and fix them. It would be cool to generate some of the functions in our math library through exact techniques to guarantee 1-ulp for each of them. I might reach out to some researchers in this area and see if they're interested in trying their work out.

bdeket:
Just to be precise, this implementation doesn't give 1-ulp precision (below .1 at a first glance it is ok, but I didn't exhaustively check it. Between .5 up to 1.0 it certainly goes above 1 but not over 2).
The math library has already an implementation that promises 1-ulp precision, and if I remember correctly, I could not find any bigger errors with it.
This is just some low-hanging fruit that improves accuracy for small numbers by a lot.

pavpanchekha:
Yes, I do see that. A better implementation might handle 0.1–1.0 with a different polynomial, and we could apply some techniques to verify the results as well. It's a lot of work!

I think it's reasonable to only guarantee 1-ulp precision for the `math` library (though it would be really swell if the `math` library did 1/2-ulp!) but it would be nice if the standard `racket` operators were more precise about what they do guarantee.

Neither of those impact this PR. This PR is a clear accuracy win without introducing much more complexity or making things much slower (I would imagine).

Issue #3478, opened 2020-11-01
set operations don't use ~e in their error messages


jbclements:
**What version of Racket are you using?**
e.g., 7.8.0.900 [cs]

**What program did you run?**
#lang racket

(define l (for/hash ([i 100])
            (values i
                    "ohuth.ntaho.nhountaho.nu")))
;; messages very very long:
(set-empty? l)
(set-count l)

**What should have happened?**

The printed representation of the hash should have been truncated to 200-ish characters. Instead, the full value is printed. Is this a problem with generics? Maybe the fix is in the code for hashes? I spent 3 minutes looking, but got scared off when I saw the generics there.

**If you got an error message, please include it here.**

**Please include any other relevant details**
e.g., the operating system used or how you are running the code.

PR #3479, opened 2020-11-01
generics: adjust generated error messages

mflatt:
For #3478.

The error messages generated by `define-generic` currently do not follow the normal format. Besides using `~v` where `~e` should be used, the labelled error clauses are not indented by two spaces. And if extra arguments are provided, they are not really formatted right under an `other arguments...:` label.

The patch here adds indentation and changes the way that other arguments are reported in the message. For the other arguments, it preserves the argument names, but gives each a separate error clause with the argument name followed by `argument...`.

For example, with
```
#lang racket
(require racket/generic)

(define-generics something
  (both-something? something x))

(both-something? 1 2)
```

Before:
```
both-something?: contract violation:
expected: something?
given: 1
argument position: 1st
other arguments...:
  x: 2
  context...:
   ....
```

After:
```
both-something?: contract violation:
  expected: something?
  given: 1
  argument position: 1st
  x argument...: 2
  context...:
   ....
```

I'm not sure this is a good idea. for one thing, `define-generic` is still generating a lot of code just for errors. It's not clear to me that it's worth all that just to give names to arguments, and it might be better to just use `raise-argument-error`. But the original author of the `define-generic` error code was clearly interested in including this information for the error, so that's why I've tried to keep it while repairing the format.

As always with changing error messages, there's a danger that a test suite somewhere is relying on exactly the old message.

### DIFF
--- a/racket/collects/racket/private/generic.rkt
+++ b/racket/collects/racket/private/generic.rkt
@@ -441,10 +441,10 @@
        (define/with-syntax (opt-name ...) opt)
        (define/with-syntax ((optkw-key optkw-val) ...) opt-kw)
        (define/with-syntax arg-labels/restargs ; labels for args
-         #'((symbol->string 'req-name) ... 
-            (string-append (symbol->string 'opt-name) " (optional)") ...
-            (string-append "#:" (keyword->string 'reqkw-key)) ...
-            (string-append "#:" (keyword->string 'optkw-key) " (optional)")  ...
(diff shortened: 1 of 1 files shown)


rfindler:
Looking at the expansion it certainly seems like it does not need to expand into that much code!

mflatt:
It's worth looking more at the generated code, but I'll merge this much for now.

PR #3483, opened 2020-11-04
Fix typo: recurs -> recurse

iambrj:


mfelleisen:
No it really is "recur".  The word "recurse" means "cursing many times". :-)

iambrj:
@mfelleisen the original word used was "recurs" (note the presence of "s" and absence of "e", not "recur"), so I presumed it was meant to be "recurse".

Also [Oxford lexico defines recurse](https://www.lexico.com/en/definition/recurse) as the mathematical procedure, not cursing repeatedly :slightly_smiling_face:

rfindler:
I agree with Matthias about it being correct already but not that the typo
fix would be to "recurses" not "recurse"

Maybe "recur" is actually a good change here. The phrase still makes sense
in that case and doesn't invite the "missing 'e', thus typo" line of
thought.

Robby

mfelleisen:
(AE: [to recur](https://www.merriam-webster.com/dictionary/recur))

Issue #3487, opened 2020-11-06
[CS] some unused bindings don't get optimized away


Bogdanp:
**What version of Racket are you using?**
CS HEAD

**What program did you run?**

When I run

```racket
(require racket/match)

(struct st (x))
(define a (st 42))

(collect-garbage)
(collect-garbage)
(collect-garbage)

(time
 (for ([x 100000000])
   (match a
     [(? st?) 'ok])))

(collect-garbage)
(collect-garbage)
(collect-garbage)

(time
 (for ([x 100000000])
   (match a
     [(st _) 'ok])))

(collect-garbage)
(collect-garbage)
(collect-garbage)

(time
 (for ([x 100000000])
   (let ([a1 (st-x a)])
     'ok)))
```

I get the following timings:

```
cpu time: 162 real time: 174 gc time: 0
cpu time: 230 real time: 242 gc time: 0
cpu time: 174 real time: 186 gc time: 0
```

**What should have happened?**

I would have expected the unused binding in the second match clause to be optimized away, but it doesn't seem to be. When I run the same program on [this branch](https://github.com/racket/racket/compare/master...Bogdanp:optimize-struct-match), which changes `match` to eliminate unused wildcard bindings, I get the following timings:

```
cpu time: 148 real time: 159 gc time: 0
cpu time: 152 real time: 164 gc time: 0
cpu time: 177 real time: 189 gc time: 0
```

According to @samth, the binding seems to be optimized away on Racket BC. It seems like CS ought to be able to apply the same optimization.

mflatt:
I think the main CS issue here is that `unsafe-struct-ref` is slower than a safe reference (while `unsafe-struct*-ref` is faster), I expect that can be improved easily.

I see that BC optimizes away `unsafe-struct-ref`, but I'm not certain that it should, since an impersonator can interpose on `unsafe-struct-ref`.

Bogdanp:
That makes sense, thanks. In that case should `match` also be changed so as to not generate these references altogether or should clauses like `(st _)` trigger an impersonator on the first field? I'd find the latter surprising and it seems like the documentation for match agrees that it shouldn't bind anything in this case:

> _ — matches anything, without binding any identifiers.

So the reason match isn't doing this already might be because it was previously relying on BC to do that work for it.

mflatt:
[...]

PR #3488, opened 2020-11-07
match: eliminate field refs due to _ in constructor patterns

Bogdanp:
Related to #3487.  This makes `_` in constructor patterns behave the same way as it does in simple patterns, where the binding is [already being elided](https://github.com/racket/racket/blob/128174594e06d789db736f5ef355055333fa02f3/racket/collects/racket/match/compiler.rkt#L167-L170).

Should the binding in `(st b)` from my example also be removed? I feel like it'd be surprising for something like that to potentially have a side-effect, but don't feel strongly about that case, especially since the docs for match explicitly state that a binding is created.

### DIFF
--- a/racket/collects/racket/match/compiler.rkt
+++ b/racket/collects/racket/match/compiler.rkt
@@ -74,7 +74,11 @@
                                             (Row-vars-seen row)))
                                 rows)
                            esc)])
-      #`[question (let ([tmps (accs #,x)] ...) body)]))
+      (define-values (used-tmps used-accs)
+        (remove-unused-tmps #'(tmps ...) #'(accs ...) #'body))
+      (with-syntax ([(used-tmps ...) used-tmps]
+                    [(used-accs ...) used-accs])
+        #`[question (let ([used-tmps (used-accs #,x)] ...) body)])))
   (cond
     [(eq? 'box k)
      (compile-con-pat (list #'unsafe-unbox*) #'box? (compose list Box-p))]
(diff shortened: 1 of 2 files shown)


Bogdanp:
I pushed another change here to fix the source location issue @capfredf mentioned in Slack. I pushed it here because it was convenient for me, but I'm happy to split it out into a separate PR if necessary. After this change, errortrace correctly reports the location of bad clauses:

Screenshot 2020-11-11 at 11 28 04

racket/collects/racket/match/compiler.rkt
```diff
@@ -55,9 +59,7 @@
                                                   (Row-vars-seen row)))
                                       rows)
                                  esc)])
-      (if (procedure? predicate-stx)
-          #`[#,(predicate-stx x) rhs]
-          #`[(#,predicate-stx #,x) rhs])))
+      #'[lhs rhs]))
   (define (compile-con-pat accs pred pat-acc)
     (with-syntax* ([(tmps ...) (generate-temporaries accs)]
                    [(accs ...) accs]
```
Bogdanp, line 65: I considered adding syntax location to `question` here as well (GitHub won't let me comment on the right line, so you have to click expand to see it), but that seemed like it would be less useful since the location of `pred` would likely be from within this module.

samth:
The source location change is definitely right.

I'm of two minds about the other change. It's definitely something we want to fix, and this definitely works, but the approach worries me just a bit. In particular, it feels like there should be a way to integrate this more smoothly into the overall `match` algorithm. Perhaps `compile` should take additional arguments for syntax for generating the values being matched. I could see that being useful in general.

On the other hand, I'm not planning to write that code in the immediate future, and what @Bogdanp wrote already works. So I'm inclined to merge this soon.

capfredf:
This is a good fix overall, but I just wanted to add:
```
#lang racket
(define a 10)
(match 1
  [(app a 10) #t])
```
This code has a similar issue to what I reported. As far as I can tell, @Bogdanp's fix doesn't seem to solve the source location issue in `app` pattern yet.

Bogdanp:
Thanks @capfredf, I've pushed another change to fix that as well.

@samth, I agree that it would be better to not generate the bindings in the first place than the approach I took here. I'll take a stab at that approach this weekend and report back.

Bogdanp:
@samth please take a look at 1cc0a51a8fe7794a983b56fc3292dbbb1d431518 . I think this implementation is better than the last because it doesn't need to perform a search (and because it now also eliminates unused bindings in vector constructor patterns, which I'd missed the first time around). I don't see a good way to move this further down into `compile` without losing the benefit of only accessing each field once for a set of rows, which seems to be the reason why this is split the way that it is in the first place with `gen-clause` generating the field bindings.

samth:
This is great!

PR #3502, opened 2020-11-16
Add workflow to scribble build guide

pmatos:
This should avoid breaking changes being committed.

### DIFF
--- a/.github/workflows/scribble_build-guide.yml
+++ b/.github/workflows/scribble_build-guide.yml
@@ -1,0 +1,29 @@
+name: Scribble Racket Build Guide
+
+on:
+  push:
+    paths:
+      - ".github/workflow/scribble_build-guide.yaml"
+      - "pkgs/racket-build-guide/**"
+  pull_request:
+    paths:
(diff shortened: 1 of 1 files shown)


pmatos:
@samth something like this should be useful to avoid the problem you pointed our earlier, what do you think?

samth (approved):

.github/workflows/scribble_build-guide.yml
```diff
@@ -0,0 +1,29 @@
...
+      - "pkgs/racket-build-guide/**"
+
+jobs:
+  scribble:
+    runs-on: ubuntu-18.04
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
```
xxyzz, line 20: Does it really needs to fetch all the commits?

pmatos: ahah - well caught. For some reason, I thought that it was just the last one. But for that I need 1, which is the default so I can remove the line. Thanks.

PR #3504, opened 2020-11-16
Fix mutator listing in struct-auto-info. Closes #3501.

### LINKED ISSUE #3501: define-struct creates struct-auto-info incorrectly
7.9 [cs]

**What program did you run?**

```
#lang racket 
(require (for-syntax syntax/parse racket/struct-info))
(struct a-struct (a [b #:mutable] [c #:auto] [d #:auto #:mutable]))
(define-syntax (from-id-get-auto-info stx)
  (syntax-parse stx
    [(_ name:id)
     #`'#,(struct-auto-info-lists (syntax-local-value #'name))]))
(from-id-get-auto-info a-struct)
```

It should not have `set-a-struct-b!`.

N/A

Bug discovered by @kurinoku.

kurinoku:
Closes #3501.

### DIFF
--- a/racket/collects/racket/private/define-struct.rkt
+++ b/racket/collects/racket/private/define-struct.rkt
@@ -593,15 +593,18 @@
                              0
                              fld-size))
                    
-                   (define-values (sets field-to-mutator-directives)
+                   (define-values (sets field-to-mutator-directives sets-auto-count)
                      (let loop ([fields fields])
                        (cond
-                         [(null? fields) (values null null)]
+                         [(null? fields) (values null null 0)]
                          [(not (or mutable? (field-mutable? (car fields))))
                           (loop (cdr fields))]
                          [else
-                          (define-values (other-sets other-directives)
+                          (define-values (other-sets other-directives count)
                             (loop (cdr fields)))
+                          (define count* (if (field-auto? (car fields))
+                                             (+ count 1)
+                                             count))
                           (define this-set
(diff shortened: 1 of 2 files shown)


racket/collects/racket/private/define-struct.rkt
```diff
@@ -613,7 +617,8 @@
                                   (cons (field-to-selector/mutator-directive (car fields)
                                                                              this-set
                                                                              #f)
-                                        other-directives))])))
+                                        other-directives)
+                                  count**)])))
```
sorawee, line 621: I didn't see `count**` defined. Did I miss anything?

sorawee: Ah, nvm, I was blind.

sorawee:
This looks good to me. Some comments:

- Could you add a test to https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/struct.rktl?
- You could also avoid the accumulator `count`. Instead, return `0` in the base case, and in the recursive case, add 1 to the returned value when the condition applies.

racket/collects/racket/private/define-struct.rkt
kurinoku: At L607

samth (commented):
Looks good with a test.

samth:
Looks good. Can you squash this into one commit?

PR #3505, opened 2020-11-16
Check that user correctly regenerated build.md

pmatos:


pmatos:
@mflatt As we discussed on slack. This job will fail if user (me, for example) forgets to generate build.md after changing build guide. :)

Hopefully we'll have more of these for other files that require regeneration. Another thing to my list. :)

PR #3509, opened 2020-11-18
net-lib/ftp upload and download enhancement

dzoep:
Allow ftp-upload-file and ftp-download-file to accept ports instead of only file paths.

### DIFF
--- a/pkgs/net-doc/net/scribblings/ftp.scrbl
+++ b/pkgs/net-doc/net/scribblings/ftp.scrbl
@@ -99,7 +99,7 @@ When @racket[local-dir-or-port] is an @racket[output-port?], it downloads
 @racket[file] from the server's current directory and writes its content to
 provided @racket[output-port?]. The data is written to the port as it is being
 received from the server and if the download is interrupted, incomplete data is
-written to the port.
+written to the port. The port is closed after finishing the transfer.
 
 If @racket[progress-proc] is not @racket[#f], then it is called with a
 function @racket[_get-count] that returns two values: the number of bytes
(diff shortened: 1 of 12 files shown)


pkgs/net-doc/net/scribblings/ftp.scrbl
```diff
@@ -127,10 +136,15 @@ is intended to limit polling.
                                       #f])
          void?]{
 
-Upload @racket[file-path] to the server's current directory using the same name. 
-If the file already exists in the local directory, it is replaced.
-The @racket[progress-proc] argument is used in the same way as in @racket[ftp-download-file],
-but to report uploaded bytes instead of downloaded bytes.}
+When @racket[port] is @racket[#f], upload @racket[file-path] to the server's
+current directory using the same name.  If the file already exists in the local
```
mflatt, line 140: It looks like the original documentation had a mistake. This line should say "remote" instead of "local", right?

pkgs/net-doc/net/scribblings/ftp.scrbl
```diff
@@ -127,10 +136,15 @@ is intended to limit polling.
...
+directory, it is replaced.  The @racket[progress-proc] argument is used in the
+same way as in @racket[ftp-download-file], but to report uploaded bytes instead
+of downloaded bytes.
+
+If @racket[port] is an @racket[input-port?], the content of that port is
+streamed as upload to the server and stored under given @racket[file-path]
+name.}
```
mflatt, line 147: The port is closed after the content is uploaded, right?

mflatt (changes requested):

pkgs/net-doc/net/scribblings/ftp.scrbl
dzoep: I didn't spot it either. Definitely it should say "remote".

pkgs/net-doc/net/scribblings/ftp.scrbl
dzoep: It is true, the transfer-data closes the both "to" and "from" ports after it finishes given transfer. This paragraph should probably be amended accordingly.

PR #3512, opened 2020-11-19
Cptypes: Add special case for zero?

gus-massa:
Add a special case for `zero?` It has more subcases than I expected.
 
IIUC a call to `zero?` is expanded to a local check `(eq? x 0)` and (`fixnum? x)` and then a call to    

https://github.com/racket/racket/blob/master/racket/src/ChezScheme/s/5_3.ss#L2481-L2488

It would be nice to just inline this implementation of `zero?` and let cptypes remove as much as possible, but I guess a naive expansion of all the instances of this function and all the similar function would increase the size of the compiled code too much.


mflatt:
LGTM

PR #3518, opened 2020-11-24
Create `racket/place/dynamic` to reduce dependencies.

samth:
Also adjust implementation of th-places slightly to avoid startup
time dependencies.

Still needs documentation.


mflatt:
Looks good on a first glance. Is the documentation change to add `racket./place/dynamic` (i.e., no change to existing bindings or functionality)?

samth:
Yeah, the documentation change is just to note that this module exists and that various things are provided by it.

samth:
I've put in something for the documentation, but I'm not sure exactly what we want to say, and I've run into my usual inability to make `declare-exporting` do what I want in any non-standard case.

samth:
@mflatt any thoughts on what I need to change about `declare-exporting` here?

mflatt:
I did something and pushed. Feel free to adjust further, of course.

mflatt:
Probably `private/place.rkt` should have been `place/private/dynamic.rkt`. I forgot that `place/private/` already exists.

nwg:
@samth @mflatt I see @samth's branch is deleted, unless i'm misunderstanding. Does anyone mind if i pick this back up? I was going to use '#%racket but i realized i would kinda just be reproducing work here.

nwg:
Thanks i'm excited about this.

samth:
To link the strands together, this was merged in https://github.com/racket/racket/commit/117300621288e0d230c97bfa4012d1c98d09de1c

nwg:
Ok. I tested this in my application and performance is much better. Everything seems to work fine.

PR #3524, opened 2020-11-29
match: eliminate vector refs due to ``_ ddk''

### LINKED ISSUE #3464: [Feature Request] optimize pattern matching on a vector
Currently, pattern matching with the pattern `(vector x _ ...)` or `(vector _ ...)` will invoke `vector->list` on the whole matching value. However, this is unnecessarily expensive. For example, `(vector x _ ...)` only needs to check that the value is a vector of length at least one. After that, it can call `unsafe-vector-ref` to bind `x` to its value.

sorawee:
Currently, the pattern matcher always call `vector->list`
on the input vector if a ddk is detected.
However, when there is exactly one ddk and users wish not to
bind an identifier to the ddk, the whole conversion is not needed.
Instead, we can selectively `vector-ref` the prefix and the suffix
of the vector and match them directly against patterns.

Also strengthen an existing test.

Closes #3464

### DIFF
--- a/pkgs/racket-test/tests/match/examples.rkt
+++ b/pkgs/racket-test/tests/match/examples.rkt
@@ -1,6 +1,7 @@
 #lang scheme/base
 
 (require racket/match
+         (prefix-in s: racket/set)
          scheme/mpair
          scheme/control scheme/foreign
          (only-in racket/list split-at)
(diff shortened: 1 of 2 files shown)


sorawee:
Note that there are further improvement opportunities.

1. We could use the `unsafe-` variant of `vector-ref`.
2. We could elide more vector refs on `_` patterns.

samth (approved):

samth:
Do you want to add the `unsafe` or should we merge this now?

sorawee:
I added `unsafe-vector-ref`, and the CI returns green.

PR #3527, opened 2020-11-29
raco-pkg: show full path instead of broken front page

### LINKED ISSUE #3499: After updating to Racket 7.9 Raco is unable to install packages
7.9

**What program did you run?**
I updated from Racket 7.8 to 7.9 and, as usual, started by reinstalling the packages I was using before. I ran `raco pkg install beautiful-racket`

Normally, the installation process would start at this point.

I got the following error message:
```
$ raco pkg install beautiful-racket
Resolving "beautiful-racket" via https://download.racket-lang.org/releases/7.9/catalog/
raco pkg install: cannot find package on catalogs
  package: beautiful-racket
```

After seeing the error message above I visited `https://download.racket-lang.org/releases/7.9/catalog/` and saw the following:
```
> ((uncaught-exception-handler)    (*(+(*)(*(+(*)(*)(*)(*)(*))(+(*)(*)(*)(*)(*))(+(*)(*)(*)(*))))(+(*)(*)(*)(*)))) uncaught exception: 404
```
Seems like the `7.9` package catalog is unavailable for some reason.

I'm running `Ubuntu 20.04.1 LTS`, but this seems to be OS-independent.

sorawee:
`raco pkg` currently shows a message like:

Resolving "pollen" via https://download.racket-lang.org/releases/7.9/catalog/

However, this URL is not meant to be visited, giving 404 to visitors,
causing them to think that the package server is down, even though
that might not actually be the case.

The problem is exacerbated by the now-fixed #3499, but
I have seen this problem even before the 7.9 release (at least 5 times),
so it seems worth fixing.

In general, we should not show 404 URLs to users, anyway.

Note that I didn't change the local database and local directory cases
because I don't know how to best display the information.
However, since they are local, users at least won't (incorrectly) think that
the server is down.

### DIFF
--- a/racket/collects/pkg/private/catalog.rkt
+++ b/racket/collects/pkg/private/catalog.rkt
@@ -136,10 +136,12 @@
     (or (add-to-cache
          pkg cache
          (for/or ([i (in-list (pkg-catalogs))])
-           (define (consulting-catalog suffix)
+           (define (consulting-catalog suffix #:url [url #f])
              (if download-printf
-                 (download-printf "Resolv~a ~s via ~a\n" suffix pkg (url->string i))
-                 (log-pkg-debug "consult~a catalog ~a" suffix (url->string i))))
+                 (download-printf "Resolv~a ~s via ~a\n"
+                                  suffix pkg (url->string (or url i)))
+                 (log-pkg-debug "consult~a catalog ~a"
+                                suffix (url->string (or url i)))))
            (source->absolute-source
             i
             (select-info-version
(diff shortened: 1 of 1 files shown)


mflatt:
Although, I like this idea, we should just add an "index.html" to the catalog site, since people may tray that after looking at `raco pkg config`. If we add an "index.html", is this change still needed?

sorawee:
Nope. It's no longer needed if the front page is fixed (created).

But we should also modify https://docs.racket-lang.org/pkg/catalog-protocol.html to include the front page as a resource to be served.

mflatt:
I've changed `distro-build` to create an `index.html` file, and I've uploaded the generated file to existing catalogs like https://download.racket-lang.org/releases/7.9/catalog/ . The generated `index.html` could be fancier, but I went with simple because it was simplest to implement.

I didn't end up changing `raco pkg catalog-copy` or the catalog-protocol documentation, so far. Partly, that was because a change to `raco pkg` would not have affected a `distro-build`-created catalog, which is created by `distro-build` directly. Partly, it's because serving a catalog out of S3 is a somewhat unusual arrangement, and a typical HTTP service would synthesize a directory list and not need an explicit `index.html`.

sorawee:
Thanks!

PR #3528, opened 2020-11-30
RacketCS: preserve immutability when encoding structs as chez records

yjqww6:
This enables more optimizations in cp0.
For example:
```racket
(let ()
  (struct A (a b))
  (define x (A 1 2))
  (+ (A-a x) (A-b x)))
```
Without this change, cp0 outputs
```scheme
...
(let ([x_7 (#3%$record struct:A_1 1 2)])
  (let ([app_16 (#3%$object-ref 'scheme-object x_7 9)])
    (#2%+
     app_16
     (#3%$object-ref 'scheme-object x_7 17))))
```
while with these changes it just outputs `3`.

### DIFF
--- a/racket/src/cs/rumble/struct.ss
+++ b/racket/src/cs/rumble/struct.ss
@@ -501,8 +501,21 @@
                                                parent-rtd*
                                                prefab-uid #f #f
                                                (+ init-count auto-count)
-                                               ;; Reporting all as mutable, for now:
-                                               (sub1 (general-arithmetic-shift 1 (+ init-count auto-count))))]
+					       (if (eq? insp 'prefab)
+						   (sub1 (general-arithmetic-shift 1 (+ init-count auto-count)))
+						   (let loop ([i 0] [mask 0])
+						     (cond
+						      [(fx= i init-count)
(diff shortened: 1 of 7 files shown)


mflatt:
This looks great!

For the loops that construct a mask, I worry that using `memq` within the loop creates a quadratic algorithm. (Some classes create structure types with many more fields than you'd otherwise expect.) Could the mask be constructed instead by starting with all bits set or unset and then folding over the list of [im]mutable fields?

yjqww6:
updated.
Actually when I wrote that loop, I followed the pattern from https://github.com/racket/racket/blob/c3c5a33c5387d21875893ac481a9046d9b2f5ca8/racket/src/cs/rumble/prefab.ss#L328-L338
Does it have similar problem?

mflatt:
That loop is definitely not ideal, and improvement would be welcome. Maybe it's not causing a problem because classes create lots of mutable fields, not immutable fields.

yjqww6:
Oh, it seems `mutables->immutables` also contains a "vector-memv". I would leave them as-is in this PR.

mflatt:
Yes, that's reasonable. I expect to merge this change soon.

PR #3532, opened 2020-12-02
guide: make "Simple Definitions and Expressions" easier to read

### LINKED ISSUE #3530: Doc suggestion: use or at least link to string-prefix and string-suffix in the Racket Guide
In this document:
`racket/pkgs/racket-doc/scribblings/guide/simple-syntax.scrbl `
string-length and substring are used to check for string prefixes and suffixes.

Given that the names string-prefix and string-suffix are self-explanatory I think the guide would be easier to read and understand if they were used. (Or at least provide links to them so that people can see straight away that these convenience functions are available.)

sorawee:
Closes #3530


sorawee:
[Rendered doc](https://github.com/racket/racket/files/5629815/simp-def-expr.pdf)

PR #3540, opened 2020-12-07
RacketCS: reduce boot file size by adding #%$app/no-inline to some raise functions

yjqww6:
This seems significantly reduce the size of my racket.boot by about 3MB. (from 29627168 bytes to 26723296 bytes)

BTW: If I append an primitive abort call to `raise-arguments-error` and `raise-argument-error`, the size would be further reduced by about 114KB (by making codes more cptypes friendly or preventing more inlining?), though I don't know whether it is still a good idea.

### DIFF
--- a/racket/src/cs/rumble/error.ss
+++ b/racket/src/cs/rumble/error.ss
@@ -131,7 +131,7 @@
     (raise-argument-error 'raise-arguments-error "symbol?" who))
   (unless (string? what)
     (raise-argument-error 'raise-arguments-error "string?" what))
-  (do-raise-arguments-error who what exn:fail:contract more))
+  (#%$app/no-inline do-raise-arguments-error who what exn:fail:contract more))
   
 (define (do-raise-arguments-error who what exn:fail:contract more)
   (raise
@@ -224,16 +224,16 @@
 (define raise-argument-error
   (case-lambda
     [(who what arg)
-     (do-raise-argument-error 'raise-argument-error "given" who what #f arg #f)]
+     (#%$app/no-inline do-raise-argument-error 'raise-argument-error "given" who what #f arg #f)]
     [(who what pos arg . args)
-     (do-raise-argument-error 'raise-argument-error "given" who what pos arg args)]))
+     (#%$app/no-inline do-raise-argument-error 'raise-argument-error "given" who what pos arg args)]))
 
 (define raise-result-error
   (case-lambda
     [(who what arg)
-     (do-raise-argument-error 'raise-result-error "result" who what #f arg #f)]
+     (#%$app/no-inline do-raise-argument-error 'raise-result-error "result" who what #f arg #f)]
     [(who what pos arg . args)
-     (do-raise-argument-error 'raise-result-error "result" who what pos arg args)]))
+     (#%$app/no-inline do-raise-argument-error 'raise-result-error "result" who what pos arg args)]))
 
 (define (do-raise-type-error e-who tag who what pos arg args)
   (unless (symbol? who)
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

I'll merge this much for now, but on the question of adding a primitive abort: Maybe all the error functions should be written as
```
(define (raise-arguments-error who what . more)
  (raise (#%apply format-arguments-error who what more)))
```
so that the `raise` tends to get inlined and made visible for other optimizations. And then maybe the `#%$app/no-inline`s in "check.ss" should be removed so that the `raise` gets inlined in those places.

samth:
Did you try adding something similar to the other error functions in error.ss? In particular, raise-result-arity-error seems potentially significant.

yjqww6:
@samth Unfortunately, this would increases the size by 6KB. I guess `raise-result-arity-error` itself is large enough, which would not be inlined by default.

samth:
I was imagining adding a do-raise-result-arity-error variant -- is that what you tried?

yjqww6:
Yes.

mflatt:
Just in case: In the `raise-arguments-error` idea above, I used `#%apply` to avoid moving the work of converting `more` to a list to the caller. I think Chez Scheme will do the right thing with the combination of `#%apply` and a final argument that is a rest arg when inlining. But I'm not certain and not sure that it's relevant to `raise-result-arity`, and maybe you already experimented with that difference.

yjqww6:
@mflatt My understanding is: for a  procedure like `(lambda (x . args) body)`, if it is inlined, the construction of `args` would be the caller side. And the construction can only be eliminated if the list is unused or only accessed by combinations of `car` and `cdr`. Looping on `args` is not ok, because `cp0-outer-unroll-limit` is 0 default.

mflatt:
The list could be eliminated if it's only used by `apply`. For example:
```
> (expand/optimize '(apply do-f 1 (list 2 3)))
(do-f 1 2 3)
```
That's useful if it turns out that `do-f` accepts 3 arguments specifically, but it's also useful if it keeps a rest-arg list construction in `do-f`.

Unfortunately, `apply` doesn't see though a local binding right now:
```
> (expand/optimize '(let ([args (list 2 3)]) (apply do-f 1 args)))
(let ([#{args cojfsqk9jnmr6e3jqj8ij4nuq-2} (#2%list 2 3)])
  (#2%apply do-f 1 #{args cojfsqk9jnmr6e3jqj8ij4nuq-2}))
```
and that means it's not seen by `apply` on a rest arg after inlining:
```
> (expand/optimize '(let () (define (f a . args) (apply do-f a args)) (f 1 2 3)))
(let ([#{args cojfsqk9jnmr6e3jqj8ij4nuq-1} (#3%list 2 3)])
  (#2%apply do-f 1 #{args cojfsqk9jnmr6e3jqj8ij4nuq-1}))
```

yjqww6:
Oh, I see, one more case when the list could be eliminated.
I think `apply` could see through local variables. The failed examples are just because `do-f` is unbound, otherwise it should work.
```scheme
> (expand/optimize '(let ([g do-f]) 
                      (define (f a . args) (apply g a args)) 
                      (f 1 2 3)))
(let ([#{g hvmcgbwjisvlq2p7sayxg8ljy-31} do-f])
  (#{g hvmcgbwjisvlq2p7sayxg8ljy-31} 1 2 3))
```

Though it would still fail if `args` is referenced more than once.

mflatt:
Oh, that's interesting. I'm not immediately sure why `do-f` needs to be bound, but that's good enough for the potential use case in functions like `raise-argument-error`.

mflatt:
I started looking at a Racket-level test suite for optimizations, and that indirectly led me to look more at the problem of exposing whether functions escape.

Trying to position a `raise` call so that it will be exposed by inlining does not work so well. It seems to work better to add a `#%$app/no-return` along the same lines as `#%$app/no-inline`. That propagates the information to the place where the compiler wants it, which is on the application, and it gives schemify a way to communicate to the compiler.

In any case, I'm still experimenting.

samth:
If I understand correctly that `#%app/no-return` would indicate that the application did not return, then this would be worth exposing at the Racket level too -- Typed Racket would be able to provide this information.

yjqww6:
Another approach I haved tried is adding a `$unreachable` to chez, which is just a `void` but marked with `bottom` type as result. 
https://github.com/yjqww6/racket/commit/98458165d76a971bb707de1b5daf0c9601e611a9

PR #3543, opened 2020-12-08
Fix bug in example for in-indexed

jackfirth:
The perils of writing examples without running them..


mflatt:
I ran the example before merging, but didn't pay attention beyond "doesn't crash". :)

PR #3546, opened 2020-12-09
ChezScheme: add cross-library inlining support for procedures with improper formals

yjqww6:
Currently, procedures like
```scheme
(define (f a . rest)
   (apply list a rest))
```
are not cross-library inlinable.

I see the original implementation does not support this because some case-lambda clauses might be dropped for cross-library inlining. (from comments)

I think it is not rare that all the clauses are preserved, and in such cases it would still be safe to perform inlining.

### DIFF
--- a/racket/src/ChezScheme/mats/8.ms
+++ b/racket/src/ChezScheme/mats/8.ms
@@ -9302,6 +9302,41 @@
     '(#t . #t))
   (equal? (let () (import (testfile-clo-3a)) (h)) (void))
   (not (let () (import (testfile-clo-3a)) (g)))
+
+  ; testing support of procedures with improper formals
+  (begin
+    (with-output-to-file "testfile-clo-4a.ss"
+      (lambda ()
+        (pretty-print
+         '(library (testfile-clo-4a)
+            (export f g)
+            (import (chezscheme))
+            (define (f a . rest)
+              (apply list a rest))
+            (define g
+              (case-lambda
+                [(a) "foo"]
+                [(a . rest) (apply list a rest)])))))
+      'replace)
+    #t)
+  (begin
+    (load-library "testfile-clo-4a.ss"
+                  (lambda (x) (parameterize ([optimize-level 2] [enable-cp0 #t] [#%$suppress-primitive-inlining #f] [current-eval compile])
+                                (eval x))))
+    #t)
+  (equivalent-expansion?
+   (parameterize ([optimize-level 2] [enable-cp0 #t] [#%$suppress-primitive-inlining #f])
+     (expand/optimize
+      '(lambda (x y z)
+         (import (testfile-clo-4a))
+         (list
+          (f x y z)
+          (g x y z)))))
+   '(begin
+      (#3%$invoke-library '(testfile-clo-4a) '() 'testfile-clo-4a)
+      (lambda (x y z)
+        (#2%list (#2%list x y z)
+                 ((#3%$top-level-value 'g) x y z)))))
 )
 
 (mat lots-of-libraries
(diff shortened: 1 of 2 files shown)


mflatt:
Good timing! I needed this for e02c417de05bc731d17b4fa59a7c80eeb68725c7.

The combination exposed a problem in #3531 that's fixed by 0de549800e8bb5dcbf900807984d6b15938eb623.

yjqww6:
> Good timing! I needed this for [e02c417](https://github.com/racket/racket/commit/e02c417de05bc731d17b4fa59a7c80eeb68725c7).
> 
> The combination exposed a problem in #3531 that's fixed by [0de5498](https://github.com/racket/racket/commit/0de549800e8bb5dcbf900807984d6b15938eb623).

I didn't notice that `residualize-ref` should be used for modifications on `$record-ref`, thanks!

yjqww6:
@mflatt, I just found this is not a good choice for `unsafe-struct`, because it needs to run cp0 twice to optimize such a procedure call to the expected form. So the first cptypes pass would not known much. The second cptypes pass knows it would create a struct of some type, and remove dead branches, but there is no third cp0 pass to eliminate unused wrapper procedures.

For a local struct definitions like
```racket
(let () (struct A (a b) #:authentic) (lambda (x) (define y (struct-copy A x)) (+ (A-a y) (A-b y))))
```
the cp0 output currently looks like
```scheme
(lambda (instance-variable-reference .top-level-bind!1
         .top-level-require!2 .mpi-vector3 .syntax-literals4
         .namespace5 .phase6 .self7 .inspector8
         .bulk-binding-registry9 .set-transformer!10)
  (let ([struct:A_16 (#2%make-record-type-descriptor* 'A #f #f
                       #f #f 2 0)])
    (let ([props (#2%list (#2%cons prop:authentic #t))]
          [insp (current-inspector)])
      (struct-type-install-properties! struct:A_16 'A 2 0 #f props
        insp #f '(0 1) #f 'A #f))
    (#3%$value
      (#3%make-wrapper-procedure
        (lambda (g1)
          (if (#3%record? g1 struct:A_16)
              (#2%void)
              (#3%$record-oops 'A-a g1 struct:A_16))
          (#3%$object-ref 'scheme-object g1 9))
        2
        (#3%cons struct:A_16 0)))
    (#3%$value
      (#3%make-wrapper-procedure
        (lambda (g0)
          (if (#3%record? g0 struct:A_16)
              (#2%void)
              (#3%$record-oops 'A-b g0 struct:A_16))
          (#3%$object-ref 'scheme-object g0 17))
        2
        (#3%cons struct:A_16 1)))
    (lambda (x_28)
      (let ([y_29 (begin
                    (if (#3%record? x_28 struct:A_16)
                        (#2%void)
                        ($app/no-return
                          raise-argument-error
                          'struct-copy
                          "A?"
                          x_28))
                    (let ([app_12 (#3%$record-ref x_28 0)])
                      (#3%$record
                        struct:A_16
                        app_12
                        (#3%$record-ref x_28 1))))])
        (let ([app_14 (#3%$record-ref y_29 0)])
          (#2%+ app_14 (#3%$record-ref y_29 1)))))))
```
Define `unsafe-struct` as a macro to force inlining would eliminate these things.

mflatt:
Interesting. Easier than a macro would be just `(define unsafe-struct #%$record)`, since `unsafe-struct` is internal and `(procedure-name unsafe-struct)` doesn't need to say `'unsafe-struct`.

But I see that doesn't actually work with the current cp0, which surprised me. Improving cp0 here is a small repair that would bring things in line with my expectations generally.

So, I'll make those changes.

PR #3550, opened 2020-12-10
CI: prevent forks from running jobs for self-hosted

sorawee:
CC: @pmatos, @jackfirth


pmatos:
LGTM - thanks @sorawee

PR #3556, opened 2020-12-13
Format the name of continuation prompt tags

shhyou:
This change recovers the Racket-BC formatting style of continuation prompt tags for Racket CS:

```
> (default-continuation-prompt-tag)
#<continuation-prompt-tag:default>
> (make-continuation-prompt-tag 'myprompt)
#<continuation-prompt-tag:myprompt>
```

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/cont.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/cont.scrbl
@@ -94,12 +94,16 @@ handler, which accepts a single thunk to apply (with the prompt
 intact).}
 
 @defproc*[([(make-continuation-prompt-tag) continuation-prompt-tag?]
-           [(make-continuation-prompt-tag [sym symbol?]) continuation-prompt-tag?])]{
+           [(make-continuation-prompt-tag [name symbol?]) continuation-prompt-tag?])]{
 
 Creates a prompt tag that is not @racket[equal?] to the result of any
 other value (including prior or future results from
-@racket[make-continuation-prompt-tag]). The optional @racket[sym]
-argument, if supplied, is used when printing the prompt tag.}
+@racket[make-continuation-prompt-tag]). The optional @racket[name]
+argument, if supplied, specifies the name of the prompt tag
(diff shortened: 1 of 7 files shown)


mflatt:
Thanks for this patch!

But I'm reluctant to add more printing specialization at the Rumble layer. It seems better in the long run to have that in the "io" layer.

What do you think about installing a `prop:object-name` value, instead, and then changing "io/print/main.rkt" to print tags using `print-named`? For consistency, it would make sense to extend `object-name` in BC, too.

shhyou:
Cool! I didn't know about `io/print`. It's indeed a better place.

mflatt:
I didn't mean to force this to "ready to review", but it looks good to me, and I tried it in my checkout.

shhyou:
Thanks! I edited the docs and these are all the changes I have.

pkgs/racket-doc/scribblings/reference/cont.scrbl
```diff
@@ -94,12 +94,15 @@ handler, which accepts a single thunk to apply (with the prompt
...
 @defproc*[([(make-continuation-prompt-tag) continuation-prompt-tag?]
-           [(make-continuation-prompt-tag [sym symbol?]) continuation-prompt-tag?])]{
+           [(make-continuation-prompt-tag [name symbol?]) continuation-prompt-tag?])]{
 
 Creates a prompt tag that is not @racket[equal?] to the result of any
 other value (including prior or future results from
-@racket[make-continuation-prompt-tag]). The optional @racket[sym]
-argument, if supplied, is used when printing the prompt tag.}
+@racket[make-continuation-prompt-tag]). The optional @racket[name]
+argument, if supplied, specifies the name of the prompt tag.}
```
mflatt, line 102: How about adding "for printing or `object-name`"? Otherwise, it it may seem like the name is meant to have some other effect.

shhyou: Added.

PR #3561, opened 2020-12-15
cs: correct a result-arity error test

sorawee:


mflatt:
Thanks - I had removed the `printf` for testing locally but didn't `git add` correctly.

PR #3566, opened 2020-12-16
Use build-path in a collapse-module-path test.

shhyou:
Fixes the test failure triggered in racket#3563.

The PR racket#3563 runs `tests/syntax` on Windows.


shhyou:
I ran the test using the file [from the checkout](https://github.com/shhyou/racket/commit/33eb9691cd0d421b1125d25d9ea3b3aec632add6) and tried the [racket3m build on Windows](https://github.com/shhyou/racket/runs/1561517336). This change seems to work, but I can't test it locally.

PR #3568, opened 2020-12-17
Enable inlining for `define-inline` calls as arguments.

samth:
A seemingly-unintentional choice made the following not behave
as expected:

    (define-inline (f x) (+ x 1))
    (f (f 2))

because the `(f 2)` was not inlined.

Reported by @mflatt and Liwei Chou.


mflatt:
LGTM

Issue #3535, opened 2020-12-04
Locally defined structs are slow in RacketCS


yjqww6:
**What version of Racket are you using?**
7.9 [cs]

**What program did you run?**
```racket
#lang racket
(require racket/unsafe/ops)
(define-syntax-rule (test)
  (begin
    (struct A (a b c d) #:authentic)
    (define (build n)
      (cond
        [(> n 0)
         (define a (build (sub1 n)))
         (A a a a a)]
        [else 1]))
    (define (f x)
      (cond
        [(A? x) (unsafe-fx+ (f (A-a x))
                            (f (A-b x))
                            (f (A-c x))
                            (f (A-d x)))]
        [else x]))

    (define t (build 13))
    (collect-garbage)
    (time (f t))))

(test)

(let () (test))
```

**What should have happened?**
I expect they have similar performance, but the local version is 3-4x slower.
```
cpu time: 172 real time: 193 gc time: 0
67108864
cpu time: 677 real time: 689 gc time: 0
67108864
```
I consider this behavior comes from the lift pass of schemify. The cp0 output is
module-level (seems optimal):
```scheme
[f.1 (lambda (x_17)
       (if (#3%record? x_17 struct:A.1)
           (let ([app_21 (f.1 (#3%$object-ref
                               'scheme-object
                               x_17
                               9))])
             (let ([app_22 (f.1 (#3%$object-ref
                                 'scheme-object
                                 x_17
                                 17))])
               (let ([app_23 (f.1 (#3%$object-ref
                                   'scheme-object
                                   x_17
                                   25))])
                 (#3%fx+
                  app_21
                  app_22
                  app_23
                  (f.1 (#3%$object-ref
                        'scheme-object
                        x_17
                        33))))))
           x_17))]
```
local (struct operations are not inlined):
```scheme
[f_31 (lambda (A-a_25 A-b_26 A-c_27 A-d_28 A?_24 x_41)
        (if (A?_24 x_41)
            (let ([app_34 (f_31 A-a_25 A-b_26 A-c_27
                                A-d_28 A?_24
                                (A-a_25 x_41))])
              (let ([app_35 (f_31 A-a_25 A-b_26 A-c_27
                                  A-d_28 A?_24
                                  (A-b_26 x_41))])
                (let ([app_36 (f_31 A-a_25 A-b_26 A-c_27
                                    A-d_28 A?_24
                                    (A-c_27 x_41))])
                  (#3%fx+
                   app_34
                   app_35
                   app_36
[...]

PR #3571, opened 2020-12-18
ChezScheme: Add a pass to lift well-known closures

yjqww6:
Related to #3535 .
Compared with lambda lifting passes on rumble and schemify layers, this pass is just before closure conversion and after free variables, loops and well-known lambdas discovered, which probably opens up more optimization opportunities.
For chez bootstrapping, the numbers of gc collections and bytes allocated are reduced by about 18%, but the changes on build time are insignificant.
no lift:
```
(time (for-each (lambda (...) ...) ...))
    638 collections
    11.861290102s elapsed cpu time, including 4.161334000s collecting
    11.917565000s elapsed real time, including 4.172969000s collecting
    10216836448 bytes allocated, including 10214525280 bytes reclaimed
```
lift:
```
(time (for-each (lambda (...) ...) ...))
    525 collections
    11.675721869s elapsed cpu time, including 4.086355000s collecting
    11.699825000s elapsed real time, including 4.098280000s collecting
    8382610368 bytes allocated, including 8380233600 bytes reclaimed
```
More benchmarks might be needed, though.

### DIFF
--- a/racket/src/ChezScheme/rktboot/constant.rkt
+++ b/racket/src/ChezScheme/rktboot/constant.rkt
@@ -86,7 +86,8 @@
   prelex-was-flags-offset
   prelex-sticky-mask
   prelex-is-mask
-  scheme-version)
+  scheme-version
+  code-flag-lift-barrier)
 
 (provide record-ptr-offset)
 (define record-ptr-offset 1)
--- a/racket/src/ChezScheme/rktboot/scheme-lang.rkt
+++ b/racket/src/ChezScheme/rktboot/scheme-lang.rkt
@@ -104,6 +104,7 @@
          $compile-profile
          compile-profile
          $optimize-closures
+         $lift-closures
          $profile-block-data?
          run-cp0
          generate-interrupt-trap
@@ -932,6 +933,7 @@
 (define $compile-profile (make-parameter #f))
 (define compile-profile $compile-profile)
 (define $optimize-closures (make-parameter #t))
+(define $lift-closures (make-parameter #t))
 (define $profile-block-data? (make-parameter #f))
 (define run-cp0 (make-parameter error))
 (define generate-interrupt-trap (make-parameter #t))
--- a/racket/src/ChezScheme/rktboot/scheme-lang.rkt
+++ b/racket/src/ChezScheme/rktboot/scheme-lang.rkt
@@ -333,7 +333,8 @@
          priminfo-libraries
          $c-bufsiz
          $foreign-procedure
-         make-guardian)
+         make-guardian
+         $lambda/lift-barrier)
 
 (module+ callback
   (provide set-current-expand-set-callback!))
@@ -1290,3 +1291,7 @@
     [() #f]
     [(v) (void)]
     [(v rep) (void)]))
+
+(define-syntax $lambda/lift-barrier
+  (syntax-rules ()
+    [(_ fmls body ...) (lambda fmls body ...)]))
--- a/racket/src/ChezScheme/rktboot/scheme-lang.rkt
+++ b/racket/src/ChezScheme/rktboot/scheme-lang.rkt
@@ -704,6 +704,7 @@
            [(prelex-was-flags-offset) prelex-was-flags-offset]
            [(prelex-sticky-mask) prelex-sticky-mask]
            [(prelex-is-mask) prelex-is-mask]
+           [(code-flag-lift-barrier) code-flag-lift-barrier]
            [else (error 'constant "unknown: ~s" #'id)])]))
 
 (define $target-machine (make-parameter (string->symbol target-machine)))
--- a/racket/src/ChezScheme/s/cmacros.ss
+++ b/racket/src/ChezScheme/s/cmacros.ss
@@ -869,13 +869,14 @@
 
 ;; Flags that matter to the GC must apply only to static-generation
 ;; objects, and they must not overlap with `forward-marker`
-(define-constant code-flag-system           #b0000001)
-(define-constant code-flag-continuation     #b0000010)
-(define-constant code-flag-template         #b0000100)
-(define-constant code-flag-guardian         #b0001000)
-(define-constant code-flag-mutable-closure  #b0010000)
-(define-constant code-flag-arity-in-closure #b0100000)
-(define-constant code-flag-single-valued    #b1000000)
+(define-constant code-flag-system           #b00000001)
+(define-constant code-flag-continuation     #b00000010)
+(define-constant code-flag-template         #b00000100)
+(define-constant code-flag-guardian         #b00001000)
+(define-constant code-flag-mutable-closure  #b00010000)
+(define-constant code-flag-arity-in-closure #b00100000)
+(define-constant code-flag-single-valued    #b01000000)
+(define-constant code-flag-lift-barrier     #b10000000)
 
 (define-constant fixnum-bits
   (case (constant ptr-bits)
--- a/racket/src/ChezScheme/s/compile.ss
+++ b/racket/src/ChezScheme/s/compile.ss
@@ -559,6 +559,7 @@
                    [$compile-profile ($compile-profile)]
                    [generate-interrupt-trap (generate-interrupt-trap)]
                    [$optimize-closures ($optimize-closures)]
+                   [$lift-closures ($lift-closures)]
                    [enable-cross-library-optimization (enable-cross-library-optimization)]
                    [generate-covin-files (generate-covin-files)]
                    [enable-arithmetic-left-associative (enable-arithmetic-left-associative)]
--- a/racket/src/ChezScheme/s/compile.ss
+++ b/racket/src/ChezScheme/s/compile.ss
@@ -70,7 +70,7 @@
     (with-output-language (Lsrc Expr)
       ($c-make-closure
         ; pretending main is a library routine to avoid argument-count check
-        (let ([x `(case-lambda ,(make-preinfo-lambda #f #f (lookup-libspec main)) (clause () 0 ,x))])
+        (let ([x `(case-lambda ,(make-preinfo-lambda #f #f (lookup-libspec main) #f (constant code-flag-lift-barrier)) (clause () 0 ,x))])
           ($np-compile x #f))))))
 
 (define c-set-code-quad!
@@ -647,7 +647,7 @@
         (with-output-language (Lsrc Expr)
           (define (lambda-chunk lsrc)
             ; pretending main is a library routine to avoid argument-count check
-            `(case-lambda ,(make-preinfo-lambda #f #f (lookup-libspec main))
+            `(case-lambda ,(make-preinfo-lambda #f #f (lookup-libspec main) #f (constant code-flag-lift-barrier))
                (clause () 0 ,lsrc)))
           (define (visit lsrc e* rchunk*)
             (define (rchunks) (cons (make-visit-chunk (lambda-chunk lsrc)) rchunk*))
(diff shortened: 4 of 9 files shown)


mflatt:
This looks fantastic. I can confirm reduced allocation at the Racket level for things like `racket-cl racket` (to recompile the `racket` module from source). There's no speedup on my machine, but also no slowdown.

More usefully, I can discard the `define/lift` macro in "rumble/define.ss" and just use `define`, and there's no allocation penalty for `equal?` like there was before this new pass. Just being able to get rid of that macro makes this improvement worthwhile.

I'll investigate the further change of not using schemify's lift pass.

mflatt:
I'm seeing good results for run time and total allocation. But code space increases, and I don't know why.

The ".so" sizes in "racket/src/build/cs/c" are not really representative, but they may point in a useful direction. This is what I get before your changes (on ARM Mac OS):
```
laptop$ ls -l *.so
-rw-r--r--  1 mflatt  wheel    44808 Dec 18 16:22 chezpart.so
-rw-r--r--  1 mflatt  wheel   890751 Dec 18 16:23 expander.so
-rw-r--r--  1 mflatt  wheel   664879 Dec 18 16:23 io.so
-rw-r--r--  1 mflatt  wheel    68608 Dec 18 16:23 linklet.so
-rw-r--r--  1 mflatt  wheel    31309 Dec 18 16:23 main.so
-rw-r--r--  1 mflatt  wheel  3495469 Dec 18 16:23 racket.so
-rw-r--r--  1 mflatt  wheel   127875 Dec 18 16:23 regexp.so
-rw-r--r--  1 mflatt  wheel   889608 Dec 18 16:22 rumble.so
-rw-r--r--  1 mflatt  wheel   488934 Dec 18 16:23 schemify.so
-rw-r--r--  1 mflatt  wheel   238930 Dec 18 16:23 thread.so
```
This is after, with `define/lift` discarded and schemify's lift pass dropped for the content of "racket/src/cs/schemified":
```
-rw-r--r--  1 mflatt  wheel    44808 Dec 18 13:45 chezpart.so
-rw-r--r--  1 mflatt  wheel  1293884 Dec 18 14:56 expander.so
-rw-r--r--  1 mflatt  wheel   771420 Dec 18 14:56 io.so
-rw-r--r--  1 mflatt  wheel    70463 Dec 18 14:56 linklet.so
-rw-r--r--  1 mflatt  wheel    34310 Dec 18 14:56 main.so
-rw-r--r--  1 mflatt  wheel  3516055 Dec 18 14:56 racket.so
-rw-r--r--  1 mflatt  wheel  3516055 Dec 18 14:56 racket.so
-rw-r--r--  1 mflatt  wheel   168519 Dec 18 14:56 regexp.so
-rw-r--r--  1 mflatt  wheel   917918 Dec 18 14:56 rumble.so
-rw-r--r--  1 mflatt  wheel   622209 Dec 18 14:56 schemify.so
-rw-r--r--  1 mflatt  wheel   285311 Dec 18 14:56 thread.so
```
The increase for "expander.so" seems particularly significant, and it doesn't change much if I turn schemify's lift pass back on for "racket/src/cs/schemified". The "racket.so" size doesn't change much overall, and that's the only one that matters in the end; but the variation is surprising, and it's strange that it's independent of whether schemify has already lifted.

I see similar results for compiled Racket code. The code size after starting for DrRacket, for example, goes up from 127 MB to 134 MB. That's not a big difference, but it's not clear why there should be any difference.

yjqww6:
That's quite surprising. Probably more lambdas became well-known after cp0 inlining. I will do more experiments.

yjqww6:
I tried heuristically avoid lifting lambdas with too much free local variables (or, will definitively take too much extra arguments), some results:

master
```
    44700 12 19 12:46 ../src/build/cs/c/chezpart.so
   919819 12 19 12:47 ../src/build/cs/c/expander.so
   676165 12 19 12:47 ../src/build/cs/c/io.so 
    69075 12 19 12:47 ../src/build/cs/c/linklet.so
    31696 12 19 12:47 ../src/build/cs/c/main.so
  3572916 12 19 12:47 ../src/build/cs/c/racket.so
   131285 12 19 12:47 ../src/build/cs/c/regexp.so
   899541 12 19 12:46 ../src/build/cs/c/rumble.so
   502415 12 19 12:47 ../src/build/cs/c/schemify.so
   244492 12 19 12:46 ../src/build/cs/c/thread.so
```

https://github.com/yjqww6/racket/commit/3859334df8af2da5b86ab1b078fd4c4edcc5e07b on https://github.com/yjqww6/racket/tree/lift3
```
    44700 12 19 12:41 ../src/build/cs/c/chezpart.so
  1009277 12 19 12:42 ../src/build/cs/c/expander.so
   745666 12 19 12:41 ../src/build/cs/c/io.so 
    70239 12 19 12:41 ../src/build/cs/c/linklet.so
    31635 12 19 12:42 ../src/build/cs/c/main.so
  3584329 12 19 12:42 ../src/build/cs/c/racket.so
   136299 12 19 12:41 ../src/build/cs/c/regexp.so
   922454 12 19 12:41 ../src/build/cs/c/rumble.so
   505016 12 19 12:41 ../src/build/cs/c/schemify.so
   257093 12 19 12:41 ../src/build/cs/c/thread.so
```
Though this may not be a good idea since it is no longer reliable.

mflatt:
Ah, I forgot about the `lambda` that goes around the body in the expansion of `library`. That could explain why there are functions with too many free variables to produce "expander.so", and also why they tend to go away with whole-program optimization so that "racket.so" ends up about the same size (but I haven't looked much at WPO, so I'm guessing a little about its output).

The best approach here may be to mark the `lambda` for a library so that lifting doesn't go past it. We'd prefer to just create closures when that `lambda` is called.

yjqww6:
```scheme
> (library (f)
    (export f)
    (import (chezscheme))
    (define (f x) x))
output of cpnanopass:
(case-lambda
  [clause
   ()
   0
   (case-lambda
     [clause
      ()
      0
      (letrec ([#{f mib92292l7ig3qvh7238684mo-58} (case-lambda
                                                    [clause
                                                     (#{x mib92292l7ig3qvh7238684mo-59})
                                                     1
                                                     #{x mib92292l7ig3qvh7238684mo-59}])])
        (begin
          (#[#{primref a0xltlrcpeygsahopkplcn-2} $set-top-level-value! 263585 #f]
            '#{f mib92292l7ig3qvh7238684mo-60}
            #{f mib92292l7ig3qvh7238684mo-58})
          '#<void>))])])
```
It seems right. `library` forms expand to two level of lambdas. I would look into how library expands.

yjqww6:
I see the outer lambda is from https://github.com/racket/racket/blob/0e4f57f44c940e100f8529abbe7f13be507b5ed5/racket/src/ChezScheme/s/compile.ss#L73, and the inner is from https://github.com/racket/racket/blob/0e4f57f44c940e100f8529abbe7f13be507b5ed5/racket/src/ChezScheme/s/syntax.ss#L2707.
Since `c-compile` unconditionally introduces a lambda, we could simply go inside one more level.

mflatt:
I'm not sure I understand. Just going inside an extra `lambda` would work well for `library` forms and also Racket's `linklet` output, but it seems like it wouldn't work well for other things, including the WPO form of the Racket core code.

I had imagined that the `build-lambda` on line 2070 would get a new code flag (alongside flags like `code-flag-single-valued`), and the the lift pass would use that flag to mean that the `lambda` form is a lift barrier. It would also be helpful to have a `$lambda/lift-barrier` form as a target for `linklet` expansion., where `$lambda/lift-barrier` is just like `lambda`, but it adds the lift-barrier code flag.

yjqww6:
I see `compile-whole-program` calls `$np-compile` directly, not go through `c-compile`. So probably we should just recognize the lambda introduced by `c-compile` and go past it.

mflatt:
If it's easier, something constrained to outer `lambda`s would be fine, of course. It could even be specified as an argument to `compile` instead of a form like `$lambda/lift-barrier`, if that works for dealing with libraries.

yjqww6:
Before working on any actual solution, let's figure out all the details about the compilation pipeline:
* `compile` -> `c-compile` -> `$np-compile`
  `c-compile` here will wrap codes with a lambda. 
  * For linklet compilation, there will be two levels of lambdas, and we need to go past one or start with the second.
  * For library compilation, two levels.
  * For normal scheme code, there is just one level, but starting lifting from inner lambda is still ok(probably more common in the presence of `cheat-eval`).
*  `compile-file-help1` -> `$np-compile`
  similar to `c-compile`, `lambda-chunk` in `compile-file-help1` will wrap codes with a lambda.
  * For library compilation, two levels.
  * For `compile-whole-program`, one level.

These wrappers are introduced from
https://github.com/racket/racket/blob/b7c0130a7585def3ac9cd839166da2e34391db1c/racket/src/ChezScheme/s/compile.ss#L647-L650
https://github.com/racket/racket/blob/0e4f57f44c940e100f8529abbe7f13be507b5ed5/racket/src/ChezScheme/s/compile.ss#L73
https://github.com/racket/racket/blob/0e4f57f44c940e100f8529abbe7f13be507b5ed5/racket/src/ChezScheme/s/syntax.ss#L2707-L2708

Since this pass lifts closures to some outer `closures` forms, the simplest approach maybe just marking these wrappers, simply going past all of them, and starting lifting when a `closures` form without any marked lambdas is encountered.

mflatt:
That all sounds right.

yjqww6:
Oh, something like `$lambda/wrapper` is still needed for linklet compilations, so no longer simpler than the barrier approach.

A better approach might be starting where local variables exists outside the `closures` from.

mflatt:
I may misunderstand what you mean by "where local variables" start, but I really think the programmer is going to have to pick. The schemify lifting process (and its precedent in BC) explicitly treats the linklet level differently, because it benefits from a different tradeoff, and that makes me think that there must be two kinds of `lambda`. It's ok if one kind is constrained to not be within the other kind, though.

yjqww6:
Now I believe that some variant of `lambda` is necessary, and the barrier approach is more elegant for use.

yjqww6:
Now I implemented the barrier approach, and the results seem good
```
   44741 12 20 12:11 ../src/build/cs/c/chezpart.so
  919829 12 20 12:11 ../src/build/cs/c/expander.so
  676184 12 20 12:11 ../src/build/cs/c/io.so
   69191 12 20 12:11 ../src/build/cs/c/linklet.so
   35024 12 20 12:11 ../src/build/cs/c/main.so
 3577544 12 20 12:12 ../src/build/cs/c/racket.so
  131335 12 20 12:11 ../src/build/cs/c/regexp.so
  900244 12 20 12:11 ../src/build/cs/c/rumble.so
  503017 12 20 12:11 ../src/build/cs/c/schemify.so
  244437 12 20 12:11 ../src/build/cs/c/thread.so
```
```
(time (for-each (lambda (...) ...) ...))
    527 collections
    11.167024161s elapsed cpu time, including 3.898805000s collecting
    11.202257000s elapsed real time, including 3.908883000s collecting
    8436799520 bytes allocated, including 8434580400 bytes reclaimed
```

mflatt:
This looks great - thanks! I'll try again today to merge.

mflatt:
Overall, things look good. For example, DrRacket's footprint shrinks slightly. I can also confirm that using `$lambda/lift-barrier` for linklets matters, reducing closure sizes in DrRacket to about what it was before instead of almost 10% larger.

Code size in DrRacket's footprint increases by 2-3%, but this is primarily a result of not using schemify's lifter for oversize linklets. By not lifting, there are fewer and large subtrees that get compiled and unpcked on demand. Smaller bytecode due to fewer unpacked subtrees compensates for the increased code size.

yjqww6:
I see new lifting changes broke a test,
https://github.com/racket/racket/blob/33fd201947a121fded031ff32a91215903c65d73/pkgs/racket-test/tests/racket/contract/stronger.rkt#L719-L726
because `mk` would get inlined now(so things rely on `procedure-closure-contents-eq?` no longer work). A possible fix might be Adding `(set! mk mk)` to suppress inlining.

PR #3573, opened 2020-12-19
Split the testing steps in the Windows CI workflow

shhyou:
* cmd.exe only returns the error level of the last executed command and does not support fail-fast. So testing should be done in individual steps.

  See <https://docs.github.com/en/free-pro-team@latest/actions/reference/workflow-syntax-for-github-actions> for more information.

* Also installs db-test dependencies.

A core test in the subprocess section is failing on Windows, so the CI check won't pass.

### DIFF
--- a/.github/workflows/ci_win.yml
+++ b/.github/workflows/ci_win.yml
@@ -45,24 +45,60 @@ jobs:
       shell: cmd
       run: |
         call racket\raco.exe pkg install --auto --no-docs racket-test unstable-flonum-lib net-test
-    - name: Test
+    - name: Run tests/racket/test
       shell: cmd
-      run: |
-        call racket\raco.exe test -l tests/racket/test
-        call racket\racket.exe -l tests/racket/contract/all
-        call racket\raco.exe test -l tests/json/json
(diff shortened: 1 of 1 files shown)


samth:
LGTM but can we fix the failing test first (or did you already do that?).

sorawee:
@shhyou already did that in #3575.

shhyou:
Sorry, I have intended to post an update. Turns out there are more test failures on Windows than the one I fixed and I haven't figured out what is the problem.

If needed, the commit can be reverted so at least CI can tell us if the build succeeds or not.

- The `foreign` tests can't find the C compiler on Windows.
- CS: in `unicode`, lots of tests are failing.
- CS: in `date`, `(seconds->date -86400)` throws an exception. So far, it looks like `(time-utc->date (make-time 'time-utc 0 -86400))` [here](https://github.com/racket/racket/blob/8cd96ec5df3ab2e434e64860757a0dadfd14e057/racket/src/cs/rumble/time.ss#L122) doesn't work for Chez Scheme on Windows.

This list could be incomplete.

PR #3579, opened 2020-12-22
Fix a contract test

yjqww6:
https://github.com/racket/racket/runs/1592098995#step:11:69
This test assumes `mk` is not inlined so that `procedure-closure-contents-eq?` could return `#t` for closures from `one` and `two`, which is no longer true in current RacketCS(related to #3571).
https://github.com/racket/racket/blob/33fd201947a121fded031ff32a91215903c65d73/pkgs/racket-test/tests/racket/contract/stronger.rkt#L719-L726

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/stronger.rkt
+++ b/pkgs/racket-test/tests/racket/contract/stronger.rkt
@@ -720,6 +720,7 @@
         (struct/dc s
                    [a (>=/c c)]
                    [b (a) (>=/c a)]))
+      (set! mk mk) ;suppress inlining
       (define one (mk 1))
       (define two (mk 2))
       (,test #f trust/not-stronger? one two)


rfindler:
Hi @yjqww6 -- thank you for looking into this but I'm not yet convinced that this is the right fix. I'm sorry I have not had the time to look more carefully yet.

yjqww6:
@rfindler A proof might be inlining it manually in racketcs 7.9
```racket
#lang racket
(struct s (a b))
(define one (struct/dc s 
             [a (>=/c 1)] 
             [b (a) (>=/c a)])) 
(define two (struct/dc s 
             [a (>=/c 2)] 
             [b (a) (>=/c a)])) 
(contract-stronger? one two) ;#f
(contract-stronger? two one) ;#f
```
vs
```racket
#lang racket
(struct s (a b))
(define (mk c)
  (struct/dc s 
             [a (>=/c c)] 
             [b (a) (>=/c a)]))
(define one (mk 1)) 
(define two (mk 2)) 
(contract-stronger? one two) ;#f
(contract-stronger? two one) ;#t
```

yjqww6:
More detailedly, `struct/dc` here creates closures like `(lambda (a) (>=/c a)`, and
`struct/dc-stronger?` compares these closures using `procedure-closure-contents-eq?`:
https://github.com/racket/racket/blob/46a191df038244199f83f2bb1235beb8ca137f9a/racket/collects/racket/contract/private/struct-dc.rkt#L673-L676
After inlining, these closures are created from different lambda expressions and would not share same code object.
I think `procedure-closure-contents-eq?` is just a leaky abstraction for such high level things.

rfindler:
Thanks for helping out with this, @yjqww6 ! Let me explain a little bit more why I am hopeful that this commit is the wrong approach.

The way I'm looking at this problem isn't that `procedure-closure-contents-eq?` and inlining aren't cooperating quite in exactly the same as they used to be (although this is true), nor do I think that this extra inlining is a bad idea (it is probably a good idea). 

In my mind, the issue is at the `contract-stronger?` and `struct/dc` layer. In particular, if we don't get contract stronger to work for these two then other programs (like the ones in [this publication](https://users.cs.northwestern.edu/~robby/pubs/papers/ifl2007-fgr.pdf)) will take a LOT longer to run which is a bad thing. So instead of disabling the test case or expect people to sprinkle `set!` into their program isn't really viable. IMO.

rfindler:
> I think `procedure-closure-contents-eq?` is just a leaky abstraction for such high level things.

For the record, I agree completely. And the fact that we got away with using it here for so long is, well, best perhaps characterized as "lucky". (Or "unlucky" depending on your POV. ;)

yjqww6:
Yes, it would best to make `contract-stronger?` work without relying on such indeterminate things.

rfindler:
Okay, I have finally taken a moment to look critically at the code and I see what's going on here and I now completely and totally agree that your commit, @yjqww6 , is the right one. I'm sorry for dragging this out and I'll revert my revert. My apologies!

rfindler:
Oh, wait-- it appears that my revert of the commit didn't actually make it to master. It must be here somewhere. I'll find it and get rid of it.

PR #3581, opened 2020-12-23
cs: fix {u,s}16vector-{ref,set!} when memory is not `bytes?`

dvanhorn:
When a {u,s}16vector points to memory that's not bytes (e.g. from ffi)
then referencing or setting the memory results in a Chez error:

```
foreign-set!: unrecognized type 'int16
```

The fix is to change the type argument to `'integer-16` and
`'unsigned-16`.


dvanhorn:
There's probably a better way to write the test cases I added, but this was the best I could come up with.

mflatt:
Thanks!

I pushed a different variant of the test after merging, since `ffi/vector` didn't really have tests before.

PR #3582, opened 2020-12-23
Order keys when writing JSON output.

### LINKED ISSUE #3537: JSON output differs between BC and CS
Consider the following program (from the `net-jwt` test suite): 
```
(define hs256-header #hasheq((alg . "HS256") (test . "foo")))
(jsexpr->bytes hs256-header)
```

This produces different answers on Racket CS and Racket BC, because the `json` library relies on the (non-deterministic) hash iteration order.

We could 
- declare this not a bug 
- change the JSON library to sort things
- change Racket CS to match the Racket BC behavior (probably unlikely)

samth:
More significant on Racket CS because ordering is often different
than creation order.

Closes #3537.

### DIFF
--- a/racket/collects/json/main.rkt
+++ b/racket/collects/json/main.rkt
@@ -144,16 +144,20 @@
           [(hash? x)
            (write-bytes #"{" o)
            (define first? #t)
-           (for ([(k v) (in-hash x)])
-             (unless (symbol? k)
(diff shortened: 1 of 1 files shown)


mflatt:
This looks right to me.

mflatt:
... but maybe there should be a test somewhere?

mfelleisen:
I just pushed one.

PR #3584, opened 2020-12-23
Check numeric argument to `real->decimal-string` (fixes #3565)

### LINKED ISSUE #3565: real->decimal-string takes rationals
The `real->decimal-string` function is documented as taking a `real?`, although (a) it doesn't check this, and (b) it only works for rational numbers.  This leads to bad error messages when given inputs outside its domain: non-numeric inputs trigger a contract violation on `inexact->exact`, complex inputs trigger a contract violation on `abs`.  It also leads to errors on inputs that are `real?` but not `rational?`, e.g. `+nan.0`, `+inf.0`, etc. break `inexact->exact`.

I'm happy to fix, but not sure what the right thing to do here is.  Should the contract be changed to `rational?` and the implementation updated to check?  This leads to a mismatch with the name of the function and its contract, but I think that mismatch already exists and there should be better error checking and reporting.

dvanhorn:
Check that the argument to `real->decimal-string` is a real number that can be converted to an exact, i.e. it's rational.

This commit adds a check to the implementation and updates the contract in the documentation, but leaves the function name alone.

### DIFF
--- a/racket/collects/racket/private/string.rkt
+++ b/racket/collects/racket/private/string.rkt
@@ -17,7 +17,7 @@
   (define (real->decimal-string n [digits 2])
     (unless (rational? n)
       (raise-argument-error 'real->decimal-string "rational?"
-                            digits))  
+                            n))
     (unless (exact-nonnegative-integer? digits)
       (raise-argument-error 'real->decimal-string "exact-nonnegative-integer?"
                             digits))
--- a/racket/collects/racket/private/string.rkt
+++ b/racket/collects/racket/private/string.rkt
@@ -15,6 +15,9 @@
 
   ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
   (define (real->decimal-string n [digits 2])
+    (unless (rational? n)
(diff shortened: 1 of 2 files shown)


sorawee:
Can we add a note to the documentation that the seemingly incorrect signature is in fact intentional?

dvanhorn:
Sure, sounds good.  I'll update in a bit.

racket/collects/racket/private/string.rkt
```diff
@@ -15,6 +15,9 @@
 
   ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
   (define (real->decimal-string n [digits 2])
+    (unless (rational? n)
+      (raise-argument-error 'real->decimal-string "rational?"
+                            digits))
```
sorawee, line 20: @dvanhorn this should be `n`, right?

dvanhorn:
Yep.

dvanhorn:
OK, both issues have been addressed.  Let me know if the documentation change seems reasonable.

PR #3585, opened 2020-12-23
Various URL improvements

dvanhorn:
Fix up some URLs

- fix up stale links that are available elsewhere
- use wayback for some stale documents that are gone
- use https for servers that don't redirect
- use htdp.org instead of www.htdp.org
- fix old planet server URL

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/other.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/other.scrbl
@@ -36,6 +36,6 @@ includes documentation for packages in that catalog, updated daily.
 For more information about packages, see see @other-manual['(lib
 "pkg/scribblings/pkg.scrbl")].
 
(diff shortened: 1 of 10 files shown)


pkgs/racket-doc/scribblings/guide/other.scrbl
```diff
@@ -36,6 +36,6 @@ includes documentation for packages in that catalog, updated daily.
 For more information about packages, see see @other-manual['(lib
 "pkg/scribblings/pkg.scrbl")].
 
-@link["http://planet.plt-racket.org/"]{@|PLaneT|} serves packages that
+@link["https://planet.racket.org/"]{@|PLaneT|} serves packages that
```
samth, line 39:
```suggestion
@link["https://planet.racket-lang.org/"]{@|PLaneT|} serves packages that
```

pmatos:
@Samth do you have some time to look into this? We shouldn't let this rot.

Issue #3578, opened 2020-12-22
Failing Tests for RacketCS on Windows


shhyou:
**What version of Racket are you using?**

Welcome to Racket v7.9.0.17 [cs].
commit: [50f373f](https://github.com/racket/racket/tree/50f373f4d62308b4bfcceaceb0219b707db489df).

**What program did you run?**

```
  call .\racket\src\worksp\msvcprep.bat x86_amd64
  racket\raco.exe test -l tests/racket/test
```

**If you got an error message, please include it here.**

1. `unicode` tests

```
SUMMARY ----------------------------
Performed 853124 expression tests (814359 value expressions, 38765 exn expressions)
and 77530 exception field tests.

Errors were [61]:
(Section (got expected (call)))
((unicode) (complete error (status complete)))
((unicode) (#"\360\220\200\200" #"\364\220\200\200" ((UTF-16-round-trip #"\0\330\0\334") #"\360\220\200\200")))
((unicode) (complete aborts (status complete)))
((unicode) (6 3 (ds-count 6)))
((unicode) (complete aborts ((status #"\355\240\200") complete)))
((unicode) (complete aborts (status complete)))
((unicode) (6 3 (ds-count 6)))
((unicode) (complete aborts ((status #"\355\255\277") complete)))
((unicode) (complete aborts (status complete)))
((unicode) (6 3 (ds-count 6)))
((unicode) (complete aborts ((status #"\355\256\200") complete)))
((unicode) (complete aborts (status complete)))
((unicode) (6 3 (ds-count 6)))
((unicode) (complete aborts ((status #"\355\257\277") complete)))
((unicode) (complete error (status complete)))
((unicode) (#"\360\220\200\200" #"\355\240\200\355\260\200" ((UTF-16-round-trip #"\0\330\0\334") #"\360\220\200\200")))
((unicode) (complete continues (status complete)))
((unicode) (complete error (status complete)))
((unicode) (#"\360\220\217\277" #"\355\240\200\355\277\277" ((UTF-16-round-trip #"\0\330\377\337") #"\360\220\217\277")))
((unicode) (complete continues (status complete)))
((unicode) (complete error (status complete)))
((unicode) (#"\363\257\260\200" #"\355\255\277\355\260\200" ((UTF-16-round-trip #"\177\333\0\334") #"\363\257\260\200")))
((unicode) (complete continues (status complete)))
((unicode) (complete error (status complete)))
[...]

PR #3587, opened 2020-12-24
Foreign functions: on Windows, retrieve _errno from MSVCRT

shhyou:
For foreign functions, `errno` on Windows might not be the one from `msvcrt.dll`. Racket BC specifically falls back to `_errno` on Windows: <https://github.com/racket/racket/blob/f6b7f9389546a730962240ee6bdda3e3a33ff438/racket/src/bc/foreign/foreign.c#L4558-L4576>.

With this PR, `get-errno` in rumble uses `racket_errno` from `cs/c/boot.c` only when the platform is not Windows.

The test failures are discovered in #3578.

### DIFF
--- a/racket/src/cs/rumble/foreign.ss
+++ b/racket/src/cs/rumble/foreign.ss
@@ -2042,11 +2042,17 @@
 ;; function is called with interrupts disabled
 (define get-errno
   (cond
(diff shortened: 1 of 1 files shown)


mflatt:
This works for me - thanks!

PR #3588, opened 2020-12-28
Implement seconds->date in the io layer

shhyou:
This PR implements `seconds->date` in the `io` layer using `rktio_seconds_to_date` to avoid the platform-dependent behavior of `time-utc->date` in Chez Scheme.

Related to #3578.


racket/src/cs/io.sls
```diff
@@ -212,6 +212,25 @@
...
+         (ftype-ref rktio_date_t (day) p)
+         (ftype-ref rktio_date_t (month) p)
+         (ftype-ref rktio_date_t (year) p)
+         (ftype-ref rktio_date_t (day_of_week) p)
+         (ftype-ref rktio_date_t (day_of_year) p)
+         (ftype-ref rktio_date_t (is_dst) p)
+         (ftype-ref rktio_date_t (zone_offset) p)
+         (if (eqv? tzn NULL)
+             #f
+             (address->ptr tzn)))))
```
mflatt, line 233: Is this intermediate vector needed, or would it be easy to go directly to a `date*` and avoid an extra allocation?

shhyou:
I moved the conversion of `rktio_date_t` to `date*` into `io.sls`. There is no intermediate vector anymore. However, the conversion function refers to `bytes->string/utf-8`. Is this okay or do you have any suggestion?
<https://github.com/racket/racket/pull/3588/files#diff-5e2e9e8932043b9b4e588ee6a7cc49e84188627fe9e1b91950bd77532f324c2bR245>

The previous implementation broke ARM32 build, presumably due to the range check using `fixnum?`. In BC, the range check is done by `scheme_get_int_val`. In the current implementation, I [used](https://github.com/racket/racket/pull/3588/files#diff-5e2e9e8932043b9b4e588ee6a7cc49e84188627fe9e1b91950bd77532f324c2bR218) `in-date-range?` [from rumble](https://github.com/racket/racket/pull/3588/files#diff-878bff27effb2963167a529fff73630362b566b576900deb6b74689bb9d9ae94L139). Let's see if that works on ARM32.

mflatt:
Thanks for the improvements! Instead of `bytes->string/utf-8`, you could use Chez Scheme's `utf8->string` here, and that's probably better in this case.

shhyou:
Done!

racket/src/io/time/main.rkt
```diff
@@ -0,0 +1,37 @@
...
+         dt]
+        [(equal? dt rktio_seconds_to_date-error-kind)
+         (raise-arguments-error who "integer is out-of-range"
+                                "integer" si)]
+        [else
+         (error who "conversion error\n  error: ~a; ~a"
+                (vector-ref dt 1)
+                (bytes->string/utf-8
+                 (rktio_to_bytes
+                  (rktio_get_last_error_string rktio))))]))]))
```
mflatt, line 37: Don't use `rktio_get_last_error_string`, because some other Racket thread (in the same place) may have run between the call to `rktio_seconds_to_date*` and trying to get an error. That's why all the `rktio_...` wrappers return a value for an error, where the vector is gathered as part of the atomic region. Use `raise-racket-error` from "../host/error.rkt" to turn an error vector into an exception.

racket/src/io/time/main.rkt
```diff
@@ -0,0 +1,37 @@
...
+(define/who seconds->date
+  (case-lambda
+   [(s) (seconds->date s #t)]
+   [(s local?)
+    (check who real? s)
+    (let* ([s (inexact->exact s)]
+           [si (floor s)]
+           [get-gmt (if local? 0 1)]
+           [nsecs (floor (* (- s si) 1000000000))]
+           [dt (rktio_seconds_to_date* rktio si nsecs get-gmt)])
```
mflatt, line 25: The general convention is to call `rktio_...` things in atomic mode, so that the operation cannot be interrupted. Any needed explicitly decallocation should happen in the atomic region, but in this case, that's built into `rktio_seconds_to_date*`. Atomic mode could be put in `rktio_seconds_to_date*`, instead.

shhyou:
Thanks for the explanation! Let me fix the code and rebase.

shhyou:
Updated [here](https://github.com/racket/racket/pull/3588/files#diff-5e2e9e8932043b9b4e588ee6a7cc49e84188627fe9e1b91950bd77532f324c2bR222).

PR #3594, opened 2020-12-30
doc: elaborate the behavior of datum->syntax

sorawee:


pkgs/racket-doc/scribblings/reference/stx-ops.scrbl
```diff
@@ -230,6 +230,8 @@ needed to strip lexical and source-location information recursively.
           syntax?]{
 
 Converts the @tech{datum} @racket[v] to a @tech{syntax object}.
+If @racket[v] is already a @tech{syntax object}, then there is no conversion,
+and @racket[v] is returned unmodified.
 The contents of pairs, vectors, and boxes, the fields of @tech{prefab}
 structures, and the values of immutable hash tables are recursively converted.
```
shhyou, line 236: I feels like the sentence is a bit disconnected here. What do you think if it has this structure:
> If @racket[v] is already a @tech{syntax object}, ...
> Otherwise, if @racket[v] is a pair, a vector, a box, ... the contents are recursively converted

sorawee: Yes, that sounds better. I will make a change. Thanks!

sorawee: Well, @samth just merged it (thanks!), so this improvement will be in a separate PR.

samth: Sorry, I thought you had already done that.

PR #3595, opened 2020-12-31
Add syntax-mapped-symbols

yjqww6:
`syntax-mapped-symbols` returns symbols of accessible bindings at each phase.
The benefits autocompletion tools because computing available names in module level would be trivial and efficient:
For example, 
```
> (map (lambda (x) (cons (car x) (length (cdr x))))
            (syntax-case (expand '(module a racket/base)) () [(_ _ m _) (syntax-mapped-symbols #'m)]))
'((1 . 6) (0 . 1527))
```

Alternative approach might be adding bulk binding support and an option to list all phases to `syntax-debug-info`.


pkgs/racket-doc/scribblings/reference/stx-ops.scrbl
```diff
@@ -488,5 +488,15 @@ limited to---the following keys:
 
 @history[#:added "6.3"]}
 
+
+@defproc[(syntax-mapped-symbols [stx syntax?])
+         (listof (cons/c (or/c exact-integer? #f) (listof symbol?)))]{
+Return a association list mapping @tech{phase level} values(where
```
abmclin, line 494: typo here, change **a** to **an**

pkgs/racket-doc/scribblings/reference/stx-ops.scrbl
```diff
@@ -488,5 +488,15 @@ limited to---the following keys:
...
+@racket[#f] corresponds to the @tech{label phase level}) to names
```
abmclin, line 495: Should names be symbols?

yjqww6: I am not sure, but I will change it to "symbols" since you think it is better.

pkgs/racket-doc/scribblings/reference/stx-ops.scrbl
```diff
@@ -488,5 +488,15 @@ limited to---the following keys:
 
 @history[#:added "6.3"]}
 
+
+@defproc[(syntax-mapped-symbols [stx syntax?])
+         (listof (cons/c (or/c exact-integer? #f) (listof symbol?)))]{
+Return an association list mapping @tech{phase level} values(where
```
sorawee, line 494: need a space after "values"

mflatt:
I'm uncertain about this idea.

Unless I'm missing something (which is very possible), this function would expose many identifiers that are not actually or meant to be accessible, because one or more of the scopes is private. That is, as an autocompletion input, it would suggest too many names in a way that exposes implementation details.

At the same time, I think it could miss potential bindings available in a syntactic position, because an identifier can expand in a way that ends up with fewer scopes than a different identifier appearing in the same position.

More generally, I'm not sure that "what are all the names potentially work here?" is a question that can be answered reliably. The name that you use for an identifier can influence the set of available names in all sorts of ways. Granted, autocompletetion needs some sort of heuristic in that world, but I'm not sure this is a good heuristic.

The example in the original comment is more about the bindings available from a module, but there are other (well defined) reflection tools already for that.

yjqww6:
The autocompletion example is only for bindings available in module-level (from a fully expanded module, which is available from drracket background expansion), and how to find out visible local names is another problem.
For "reflection tools", I assume you mean `namespace-mapped-symbols`. It works, but it also requires declaring and instantiating the module, which could take long time.
For example, consider the following program that expands a simple typed racket module
```racket
#lang racket

(define ns (make-base-namespace))
(current-namespace ns)

(define (f)
  (define a '(module a typed/racket
               (define f car)
               (provide f)))
  (collect-garbage)
  (define b (time (expand a)))
  (time (eval b))
  (time (dynamic-require ''a #f))
  (time (void (namespace-mapped-symbols (module->namespace ''a)))))

(f)
```

```
cpu time: 810 real time: 810 gc time: 316
cpu time: 174 real time: 174 gc time: 49
cpu time: 51 real time: 51 gc time: 33
cpu time: 4 real time: 4 gc time: 0
```
It needs hundreds of ms to make that module available for `namespace-mapped-symbols`. One might think it is not siginificant since expansion take much more time, but
```racket
#lang racket

(define ns (make-base-namespace))
(current-namespace ns)

(define (f)
  (define a '(module a typed/racket
               (define f car)
               (provide f)))
  (collect-garbage)
  (define b (time (expand a)))
  (parameterize ([current-namespace (make-base-namespace)])
    (time (eval b))
    (time (dynamic-require ''a #f))
    (time (void (namespace-mapped-symbols (module->namespace ''a))))))

(f)
(f)
```
```
cpu time: 822 real time: 822 gc time: 320
cpu time: 276 real time: 276 gc time: 91
cpu time: 23 real time: 23 gc time: 5
cpu time: 6 real time: 6 gc time: 2
cpu time: 268 real time: 268 gc time: 82
cpu time: 261 real time: 261 gc time: 69
cpu time: 32 real time: 32 gc time: 14
cpu time: 6 real time: 6 gc time: 2
```
In the second pass the expansion is much faster, and declaring modules is expensive now. Even worse, the gc time is quite significant, which might become a problem in drracketcs.

The only efficient approach I found is walking through the expanded module, finding out all `#%require`s, and computing available names. This should typically cost less than 10ms but the implementation is non-trivial, especially for submodules, and duplicate with the expander.

Since these names are already available inside module scopes, I think it is reasonable to provide apis for them.

mflatt:
I agree that `namespace-mapped-symbols` is the right tool. I had in mind `module->exports`, `module-compiled-exports`, and related functions.

yjqww6:
`module->exports` is what the `drcomplete` plugin currently used for external requires, and the heuristic is whether the imported identifier contains macro scopes. It seems ok in practice, but the implementation is incomplete, especially for submodules.

mflatt:
In e98526fbb36e26fc297e22f5ac52b0e14e260726, I added a similar function,`syntax-bound-symbols`. I had forgotten about this PR! And when I decided that I'd have to add something like this, I was surprised to find that the core already exists within the expander as `syntax-mapped-names`.

The function I added was a little different than the one in this PR, because it retrieves names only at one designated phase. Meanwhile, I now see that the internal `syntax-mapped-names` already addresses the concerns that I had before, because it only reports names that are bound with a subset of the given syntax object's scopes (so it can't expose extras, and it won't miss any). 

While any of the interned names it reports could be found by guessing the name, it now occurs to me that gensyms can get exposed by `syntax-bound-symbols`. Probably `syntax-bound-symbols` should filter those out.

I'm not sure that `syntax-bound-symbols` is a better name, but I was trying to connect better to `identifier-binding`.

mflatt:
Meant to add: for my use, a phase-specific lookup is better. I would not object to adding the version here that reports names across phases, and maybe it's mostly a matter of picking sufficiently consistent and distinct names.

yjqww6:
I would like a `syntax-bound-phases`,  though my original problem seems no longer trivial with the existence of binding spaces.

mflatt:
@yjqww6 I like the idea of `syntax-bound-phases`.

Here's an adaptation of your implementation to add it, where I put the code in a different place to keep the implementation shift-multi-scope shifting together:

  https://github.com/mflatt/racket/tree/boundphases

Does that look right to you?

yjqww6:
@mflatt, this looks great!

mflatt:
Merged - thanks!

Issue #3403, opened 2020-09-20
Discrepancy between match behaviour and documentation


JustinHuPrime:
**What version of Racket are you using?**
e.g., 7.2 [3m]

**What program did you run?**
```racket
(match '(+ 1 2 3)
  [`(+ ,@a) #:when ((listof number?) a) (list "match" a)]
  [_ "no match"])
```

**What should have happened?**
I expect a syntax error - it is not permissible to use unquote-splicing followed by just an identifier, according to the documentation for match patterns. Instead, I get `'("match" (1 2 3))`

**Please include any other relevant details**
e.g., the operating system used or how you are running the code.

Using DrRacket's definitions window, on Ubuntu Linux (20.04)

jackfirth:
I think this is an error in the docs. None of the documented rules for quasipatterns cover that case:

> `,pat` - match pat
> `,@(list lvp ...)` - match lvps, spliced
> `,@(list-rest lvp ... pat)` - match lvps plus pat, spliced
> `,@'qp` - match list-matching qp, spliced

What happens if you try `,@'a`?

JustinHuPrime:
> I think this is an error in the docs. None of the documented rules for quasipatterns cover that case:
> 
> > `,pat` - match pat
> > `,@(list lvp ...)` - match lvps, spliced
> > `,@(list-rest lvp ... pat)` - match lvps plus pat, spliced
> > `,@'qp` - match list-matching qp, spliced
> 
> What happens if you try `,@'a`?

The #:when clause has an error: `a: unbound identifier in: a`

jackfirth:
That does make sense, since it's trying to match a list whose tail matches the literal symbol `'a` instead of whose tail is bound to a pattern variable. I think that quote before `qp` in the docs is a typo and should be removed.

JustinHuPrime:
I think the typo is that the quote should be a quasiquote - the following match works as expected if the single quote in the docs were a backtick:
```racket
(match '(+ 1 2 3)
  [`(+ ,@`(,a ,b ,c)) (list "match" a b c)]
  [_ "no match"])
```

jackfirth:
Oh that would make sense. Then I think fixing that clause and adding another clause for the bare identifier case would be appropriate.

JustinHuPrime:
I'm not sure the bare identifier case should exist at all - this snippet fails on a syntax error:

```racket
#lang racket
(match '(+ 1 2 3)
  [`(+ ,@a ,@b) (list "match" a)]
[...]

PR #3597, opened 2021-01-02
Allow variables in unquote-splicing terms

pavpanchekha:
This PR attempts to fix https://github.com/racket/racket/issues/3403 by allowing variables to be unquote-spliced inside patterns. It does so by extending the `null-terminated?` function inside `parse-unquote`.

Broadly speaking, `,@v` inside a quasi-quoted pattern is equivalent to `,@(list v ...)`. That means ambiguous cases, like `(quasiquote (,@v ,@w))` will be resolved with greedy matching. Since the ambiguity already exists with `(quasiquote (,v ... ,w ...))` it does not seem harmful.

The value of this change is greater symmetry between quasiquotation for constructing objects, and patterns for matching them.

The fix works by adding variable patterns to `null-terminated?`. I think that in general, `null-terminated?` should admit all patterns that _could_ match a list, instead of just the ones that _only_ match a list. That means that `(app f)` patterns and many others should also be valid, though those are harder to implement, so I haven't tried those yet. They seem less likely to be useful, because they aren't symmetric to a constructor term. I think this PR should be merged without those being implemented.

### DIFF
--- a/pkgs/racket-test/tests/match/examples.rkt
+++ b/pkgs/racket-test/tests/match/examples.rkt
@@ -786,7 +786,7 @@
             (list es en)]))
 
    (comp '((1 2 3 4) (6 7) (9))
-         (match (range 10)
+         (match '(0 1 2 3 4 5 6 7 8 9)
            [`(0 ,@a 5 ,@b 8 ,@c)
             (list a b c)]))
 
--- a/pkgs/racket-test/tests/match/examples.rkt
+++ b/pkgs/racket-test/tests/match/examples.rkt
@@ -785,6 +785,11 @@
            [`(begin ,es ... ,en)
             (list es en)]))
 
+   (comp '((1 2 3 4) (6 7) (9))
+         (match (range 10)
+           [`(0 ,@a 5 ,@b 8 ,@c)
+            (list a b c)]))
+
    (comp '(a b c)
 	 (let ()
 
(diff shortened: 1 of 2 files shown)


samth:
This looks good, but can you add some tests?

pavpanchekha:
Would love to. Where do the tests live?

samth:
https://www.github.com/racket/racket/tree/master/pkgs%2Fracket-test%2Ftests%2Fmatch%2Fexamples.rkt

pavpanchekha:
I added a test. I could do more but it doesn't look like that's the style from the other tests.

samth:
That test seems to fail to compile.

pavpanchekha:
Yeah, my bad, I didn't notice that the tests are in `scheme/base` instead of `racket`. I mean, it makes sense of course.

pavpanchekha:
There's still a failure, but it doesn't look related (it's a timing test for regexes). Don't know if that test is supposed to be flaky or what, but @samth I think you can trigger a rerun of the test.

shhyou:
Yes, the regexp test failure is unrelated and is likely to disappear if ran multiple times.

samth:
Can you rebase again, which should make the failure go away?

shhyou:
Merged in b56276c6704f9ce97b7.

PR #3599, opened 2021-01-03
list: optimize in-combinations and combinations

sorawee:
This PR in a sense reverts f83cec1b0452b and attempts to directly fix
the bug that the commit tries to address with an approach similar to the
original one.

The problem with the aforementioned commit is that, Gosper's hack
only works efficiently when the length of `l` is small enough that
the number representation can fit into a fixnum
(so that all bit operations take constant time).
When the length of `l` is large, the number representation could
become a bignum with length proportional to the length of `l`.
This is not ideal because it causes the time complexity of the algorithm
to be `O({|l| choose k} |l|)` instead of `O({|l| choose k} k)`, which
would be a significant performance degradation when `|l|` is much
larger than `k`.

Before the PR:

```
#lang racket

(time (for ([x (in-combinations (range 300) 3)]) (void)))
```

produces:

```
cpu time: 7435 real time: 7574 gc time: 32
```

After the PR, it produces:

```
cpu time: 239 real time: 246 gc time: 30
```

### DIFF
--- a/racket/collects/racket/list.rkt
+++ b/racket/collects/racket/list.rkt
@@ -678,9 +678,8 @@
                   ;; then we change it to #(3 5 6 7)
                   [index-to-change
                    (define val-to-change (add1 (vector-ref k* index-to-change)))
-                   (vector-set! k* index-to-change val-to-change)
-                   (for ([i (in-range (add1 index-to-change) k)]
-                         [v (in-naturals (add1 val-to-change))])
+                   (for ([i (in-range index-to-change k)]
+                         [v (in-naturals val-to-change)])
                      (vector-set! k* i v))]
                   ;; Otherwise, there's no incrementable index. E.g.,
                   ;; N = 10 and we have #(6 7 8 9), so we quit enumeration
--- a/racket/collects/racket/list.rkt
+++ b/racket/collects/racket/list.rkt
@@ -677,10 +677,11 @@
                   ;; E.g., if N = 10 and we have #(3 4 8 9)
                   ;; then we change it to #(3 5 6 7)
                   [index-to-change
-                   (vector-set! k* index-to-change
-                                (add1 (vector-ref k* index-to-change)))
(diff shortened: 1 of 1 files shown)


sorawee:
CC: @bennn

sorawee:
This new implementation also guarantees that the enumeration will follow the lexicographic order (w.r.t. the positions in the original list). I don't know if it's worth strengthening the specification in the documentation, though.

samth:
While you're working on this code, it seems worth it to extend the interface to allow vectors as input.

sorawee:
Currently the function resides in `racket/list`. Ideally, it should be in `racket/sequence` and supports arbitrary sequence. However, let's do that in.a separate PR.

samth:
The difference with vectors is that it's already a vector-based algorithm. And moving to a different library seems like not a good change.

sorawee:
There are two separate issues:
- Which kind of collection should the function accept? Here're possible answers:
  - list only (current implementation, doesn't exploit the fact that the algorithm is vector-based)
  - vector only (not backward compatible)
  - list or vector only (seems convoluted to me)
  - any sequence
- Where should the functions reside? 

@samth, you seem to think that it should still reside in `list.rkt`. That does seem to make sense, since removing it from `racket/list` might cause backward compatibility issues. However, what do you think about additionally providing it from `racket/sequence`? And what's your thought on the kind of collection the function should accept?

shhyou:
I think the issue here is that `in-combinations`, `combinations`, `in-permutations`, `permutations`, `argmin`, `argmax` and `cartesian-product` are all provided from `racket/list` and only work on lists. Some of these algorithms can handle arbitrary collections and some can handle any ordered sequences. Therefore it makes sense to discuss the generalization separately and in a larger context.

samth:
Ok, I'm fine with delaying that discussion. How good is the test suite here? Should we trust it's adequate testing?

sorawee:
The existing test suite (by @bennn) is pretty decent already.

PR #3602, opened 2021-01-04
json: make ordering of emitted object keys opt-in

LiberalArtist:
The commit <https://github.com/racket/racket/commit/6369646>
added support for emitting objects' key-value pairs in a deterministic order.

This commit adds an optional keyword argument `#:sort-keys?`
to `write-json`, `jsexper->string`, and `jsexpr->bytes`
to control the new deterministic mode.
It restores the previous default behavior,
in which key-value pairs are emitted in an unspecified order.

Related to https://github.com/racket/racket/issues/3537
Related to https://github.com/racket/racket/commit/bff1a3c
Related to https://github.com/racket/racket/pull/3582
Related to https://github.com/racket/racket/commit/f6b7f93

### DIFF
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -4,6 +4,12 @@
 
 @(define website @link["http://json.org"]{JSON web site})
 @(define rfc @link["http://www.ietf.org/rfc/rfc4627.txt"]{JSON RFC})
+@; FIXME:
+@;  - RFC 4627 was obsoleted by RFC 8259,
+@;    which more explicitly discusses object keys.
+@;  - Should probably mention ECMA-404.
+@;    https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf
+@;  - Consider using bibliography forms.
 
 @(begin (require scribble/eval)
         (define ev (make-base-eval))
@@ -67,6 +73,7 @@ the @rfc for more information about JSON.
 
 @defproc[(write-json [x jsexpr?] [out output-port? (current-output-port)]
                      [#:null jsnull any/c (json-null)]
+                     [#:sort-keys? sort-keys? any/c #f]
                      [#:encode encode (or/c 'control 'all) 'control])
          any]{
   Writes the @racket[x] @tech{jsexpr}, encoded as JSON, to the
@@ -80,20 +87,32 @@ the @rfc for more information about JSON.
   the range of @tt{U+10000} and above are encoded as two @tt{\uHHHH}
   escapes, see Section 2.5 of the @|rfc|.
 
+  If @racket[sort-keys?] is given and is not @racket[#false],
+  the key--value pairs comprising a JSON object are emitted
+  in a deterministic order, as with @racket[hash-map]
+  when called with a non-false third argument.
+  Otherwise, and by default, the order of key--value pairs
+  is unspecified.
+  Interoperable uses of JSON can not rely on the order
+  of key--value pairs:
+  see @secref["The_JS-Expression_Data_Type"] for further discussion.
+
 @examples[#:eval ev
-  (with-output-to-string
-    (λ () (write-json #hasheq((waffle . (1 2 3))))))
-  (with-output-to-string
-    (λ () (write-json #hasheq((와플 . (1 2 3)))
-                      #:encode 'all)))
+  (write-json #hasheq((waffle . (1 2 3))))
+  (write-json #hasheq((b . #t) (a . #t))
+              #:sort-keys? #t)
+  (write-json #hasheq((와플 . (1 2 3)))
+              #:encode 'all)
 ]
 }
 
 @defproc[(jsexpr->string [x jsexpr?]
                          [#:null jsnull any/c (json-null)]
+                         [#:sort-keys? sort-keys? any/c #f]
                          [#:encode encode (or/c 'control 'all) 'control])
          string?]{
-  Generates a JSON source string for the @tech{jsexpr} @racket[x].
+  Like @racket[write-json], but
+  generates a JSON source string for the @tech{jsexpr} @racket[x].
 
 @examples[#:eval ev
   (jsexpr->string #hasheq((waffle . (1 2 3))))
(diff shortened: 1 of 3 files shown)


samth:
Can you say more about when you'd want to use this?

LiberalArtist:
My main concern with promising to sort keys by default is that it is generally incorrect to rely on the ordering of keys in JSON objects. Providing strong guarantees about ordering by default seems like it would encourage unnecessary and even unintentional reliance on ordering, which would break when communicating with other JSON implementations. 

The three JSON standards differ somewhat in details on ordering. In the [json.org](https://www.json.org/json-en.html) definition, "An object is an unordered set of name/value pairs." The [ECMA](https://www.ecma-international.org/publications/files/ECMA-ST/ECMA-404.pdf) standard is more permissive, saying, "The JSON syntax … does not assign any significance to the ordering of name/value pairs. These are all semantic considerations that may be defined by JSON processors or in specifications defining specific uses of JSON for data interchange." (For context, this standard also "does not define any specific meaning to the ordering of the values" in arrays, but it does warn that ordering of array elements is "often" significant.) But [RFC 8259](https://tools.ietf.org/html/rfc8259#section-4), in addition to defining an object as "an unordered collection of zero or more name/value pairs", gives this pragmatic note:
>   JSON parsing libraries have been observed to differ as to whether or
>   not they make the ordering of object members visible to calling
>   software.  Implementations whose behavior does not depend on member
>   ordering will be interoperable in the sense that they will not be
>   affected by these differences.

A secondary concern is around performance. I expect [`try-sort-keys`](https://github.com/racket/racket/blob/c52113397c52bfe4164ae816a00fd4639b29c3c6/racket/src/cs/rumble/hash.ss#L457-L464) is reasonably speedy and, if anything, will be optimized more over time, but there is some overhead in allocating and sorting an intermediate alist. Since many programs won't need deterministic ordering (they haven't had it until now), it seems like there should at least be a way to opt out of that overhead.

mfelleisen:
I would prefer two changes: 

1. retain the sorted default 
2. change the keyword to `#:standard-conform?`

Alternatively, use a parameter for the library. I understand that looking up the parameter costs some time, but it's a single function call per write or string conversion. 

;; - - - 

Here is how you encounter the ordering problem. 

Write tests for functions that compare the data representations of two JSON objects. 
Since the test inputs and outputs are used to communicate/test programs in other PLs
that use JSON, you wrap `check-equal?` with a macro that writes the test inputs and
outputs to a file. That way you can experiment with these tests at the command-line. 
You also add them to a git repo. 

Every time you run the test, you get different orderings .. and you must double check the file.

rmculpepper:
I think we should not add extra complexity to Racket's `json` library in the hopes that it might help fuzz-test some unspecified other JSON implementation.

Racket's representation of JSON objects as hashes already makes it impossible to represent objects with duplicate keys and impossible to directly rely on the order that they keys appear in the serialized form. So I'm not sure how standards conformance is an issue; either mode conforms.

The performance argument might be valid, but it would need measuring. Unless the difference is significant, I think simplicity is the more important virtue.

mfelleisen:
Does this mean “leave as is” (my highest preference) or “remove sorting” (anti-preference)?

rmculpepper:
I mean that I prefer leaving the code as it is, unless @LiberalArtist or someone else shows that the performance cost of sorting is really significant (and that would surprise me). And even in that case I think sorting should be the default behavior.

mfelleisen:
OK!

LiberalArtist:
Here's a quick and dirty benchmark, as a git branch so others can try it: https://github.com/LiberalArtist/racket/commit/69ad0f12499220cb09b405a1b69d1749dec02239

My results:
```
$ ../../../../racket/bin/racket benchmark.rkt
sort-keys? = #t
cpu time: 217407 real time: 221192 gc time: 1619
sort-keys? = #f
cpu time: 168703 real time: 171867 gc time: 425
$ ../../../../racket/bin/racket benchmark.rkt
sort-keys? = #t
cpu time: 208770 real time: 209424 gc time: 1483
sort-keys? = #f
cpu time: 161015 real time: 161750 gc time: 364
```

A better benchmark might use knowledge about the Racket CS implementation to make sure a realistic amount of sorting is being done.

LiberalArtist:
I agree that:

1. Racket's `json` library fully conforms to all of the standards, before and after 6369646 and with or without this PR;
2. Unnecessary complexity is bad; and
3. Deterministic output is a very useful feature.

I'm not sure I completely understand the comment about testing. Part of what I was trying to say is that, even aside from the order of object keys, most JSON values have many possible external representations due to optional whitespace, string escapes, and number syntax. For a Racket analogy, consider `'(λ (x) x)` and `(quote(|λ|[    x    ]\x))`.

The status quo in Racket 7.9 is unspecified, non-deterministic ordering of object keys in output. IIUC, the problem is that, while the order of hash iteration has always been unspecified, Racket BC tends to choose the same order often enough that programs incorrectly came to rely on it: these programs break with Racket CS, which tends to choose different iteration orders.

The change in 6369646 provides an obvious (lexical) deterministic order. If that becomes the default, Racketeers will probably rely on it even more than some did on the often-consistent order with Racket BC: then we're probably committed to it forever for compatibility, so we might as well document it.

rmculpepper:
Thanks for the clarification. I think we should commit to the behavior introduced in 6369646, and as you say, that means we should document it. I think that behavior should be the default (rather than opt-in).

The benchmark does not convince me that there is a need for an opt-out option, unless you know of JSON applications that have single objects with ten thousand keys that get serialized at a high rate. (But if so, please say so!) I tried the benchmark but with the JSON file at http://pre-release.racket-lang.org/status/data/repo_racket_racket.json instead and got about 3% overhead. On the other hand, the code impact is small so I'm not strongly opposed to the option.

samth:
I 100% in agreement with @rmculpepper's last comment.

LiberalArtist:
> The benchmark does not convince me that there is a need for an opt-out option, unless you know of JSON applications that have single objects with ten thousand keys that get serialized at a high rate. (But if so, please say so!) I tried the benchmark but with the JSON file at http://pre-release.racket-lang.org/status/data/repo_racket_racket.json instead and got about 3% overhead. On the other hand, the code impact is small so I'm not strongly opposed to the option.

I had looked over my code and concluded—but had apparently not written here—that, while I do have JSON objects with tens of thousands of keys (representing words used in a corpus of texts), in those cases I was already serializing the objects in advance, so I didn't mind the performance impact—and I really do like determinism, as long as it's reasonably affordable. So I was persuaded that the new (now current) behavior is a good default, and, while the code impact is indeed small, I think it can be left until someone more immediately wants to use it.

PR #3606, opened 2021-01-05
schemify: inline non-authentic struct predicate

yjqww6:
Currently non-authentic struct predicates are not inlined, and the overhead seems quite significant.
For example, on my machine the following program
```racket
#lang racket
(struct A ())

(define (f x)
  (for/fold ([a #t])
            ([_ (in-range 100000000)])
    (A? x)))
(define a #f)
(set! a (A))

(collect-garbage)
(time (f a))
```
costs
380ms current
200ms patched
140ms authentic

### DIFF
--- a/racket/src/schemify/schemify.rkt
+++ b/racket/src/schemify/schemify.rkt
@@ -838,17 +838,28 @@
                                     prim-knowns knowns imports mutated simples unsafe-mode?)]
                 [else #f]))
             (define (inline-struct-predicate k s-rator im args)
-              (define type-id (and (known-struct-predicate-authentic? k)
-                                   (pair? args)
+              (define type-id (and (pair? args)
                                    (null? (cdr args))
                                    (inline-type-id k im add-import! mutated imports)))
               (cond
-                [type-id
+                [(not type-id) #f]
+                [(known-struct-predicate-authentic? k)
                  (define tmp (maybe-tmp (car args) 'v))
                  (define ques `(unsafe-struct? ,tmp ,(schemify type-id 'fresh)))
                  (wrap-tmp tmp (car args)
                            ques)]
-                [else #f]))
+                [else
+                 (define tmp (maybe-tmp (car args) 'v))
+                 (define schemified-type-id (schemify type-id 'fresh))
+                 (define tmp-type-id (maybe-tmp schemified-type-id 'v))
+                 (define ques `(if (unsafe-struct? ,tmp ,tmp-type-id)
+                                   #t
+                                   (if (impersonator? ,tmp)
+                                       (unsafe-struct? (impersonator-val ,tmp) ,tmp-type-id)
+                                       #f)))
+                 (wrap-tmp tmp (car args)
+                           (wrap-tmp tmp-type-id schemified-type-id 
+                                     ques))]))
             (define (inline-field-access k s-rator im args)
               ;; Inline the selector with an `unsafe-struct?` test plus `unsafe-struct*-ref`.
               (define type-id (and (pair? args)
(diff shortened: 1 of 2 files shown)


mflatt:
I've wondered about whether to do this. How much does code size increase with the change?

racket/src/schemify/schemify.rkt
```diff
@@ -838,17 +838,28 @@
...
+                [(known-struct-predicate-authentic? k)
                  (define tmp (maybe-tmp (car args) 'v))
                  (define ques `(unsafe-struct? ,tmp ,(schemify type-id 'fresh)))
                  (wrap-tmp tmp (car args)
                            ques)]
-                [else #f]))
+                [else
+                 (define tmp (maybe-tmp (car args) 'v))
+                 (define schemified-type-id (schemify type-id 'fresh))
+                 (define tmp-type-id (maybe-tmp schemified-type-id 'v))
```
mflatt, line 854: I don't think a temporary is ever needed here, because the type identifier has to be an immutable identifier for the predicate to be considered known.

yjqww6:
Code size from `./racket` -> `(collect-garbage)` -> `(dump-memory-stats)` (ta6osx)
master
```
code                      15,371   13,455,664  |     15,376   13,456,528
```
patched
```
 code                      15,374   13,536,352  |     15,379   13,537,216
```
I would try DrRacket later. If the code size indeed grows too much, probably https://github.com/yjqww6/racket/commit/0ef382bd68265c1c49762c89b74b6139a34b5b72 is better.

mflatt:
That other approach does look better, in any case. For what it's worth, I'd move the `(known-struct-predicate-authentic? k)` around the use of `tmp` instead of mostly duplicating the clause.

yjqww6:
Code sizes from `./racket` -> `(require drracket)` (and waiting for the window) -> `(collect-garbage)` -> `(dump-memory-stats)`, rebuild using `./raco setup -c` and `./raco setup -D`
master
```
 code                     152,249  130,140,832  |    152,254  130,141,696
 total                  7,309,850  506,834,492  |  7,477,242  515,688,380
```
patched
```
 code                     151,986  130,247,712  |    151,991  130,248,576
 total                  7,285,189  506,000,292  |  7,452,573  514,845,860
```
https://github.com/yjqww6/racket/commit/0ef382bd68265c1c49762c89b74b6139a34b5b72
```
 code                     151,971  130,125,136  |    151,976  130,126,000
 total                  7,284,634  505,877,148  |  7,452,018  514,722,716
```
Though this might not be an ideal approach and varies across each runs, code size seems not to be a problem.

benchmark (in another machine):
250ms master
110ms patched
110ms authentic
200ms https://github.com/yjqww6/racket/commit/0ef382bd68265c1c49762c89b74b6139a34b5b72

contracted case:
```racket
#lang racket
(struct A (a) #:mutable)

(define (f x)
  (for/fold ([a #t])
            ([_ (in-range 100000000)])
    (A? x)))

(define/contract a
  (struct/c A integer?) (A 0))

(collect-garbage)
(time (f a))
```
450ms master
330ms patched
220ms https://github.com/yjqww6/racket/commit/0ef382bd68265c1c49762c89b74b6139a34b5b72

Another program:
```racket
#lang racket
(require racket/unsafe/ops)

(struct A (a b c d))
(define (build n)
  (cond
    [(> n 0)
     (define a (build (sub1 n)))
     (A a a a a)]
    [else 1]))
(define (f x)
  (match x
    [(A a b c d)
     (unsafe-fx+
      (f a)
      (f b)
      (f c)
      (f d))]
    [else 1]))

(define t (build 13))
(collect-garbage)
(time (f t))
```
600ms master
310ms patched
250ms authentic
450ms https://github.com/yjqww6/racket/commit/0ef382bd68265c1c49762c89b74b6139a34b5b72

mflatt:
Looks like this is the best option - thanks!

PR #3607, opened 2021-01-06
cs: add support for cross-compiling to iOS

### LINKED ISSUE #3605: cs: iOS cross-compilation error
`HEAD`

**What program did you run?**

```sh
mkdir racket/src/build
cd racket/src/build
../configure --host=aarch64-apple-darwin --enable-ios=iPhoneOS --enable-racket="$(which racket)"
make
```

Compilation should succeed. Instead, I see:

```
gcc  -g -O2 -Wall -DOS_X -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -c -DAARCH64 -I../boot/tarm64osx -I../zlib -I../lz4/lib -I/opt/X11/include/  statics.c
ln -s "../../../../../../cs/c/../../ChezScheme"/c/segment.c segment.c
gcc  -g -O2 -Wall -DOS_X -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -c -DAARCH64 -I../boot/tarm64osx -I../zlib -I../lz4/lib -I/opt/X11/include/  segment.c
ln -s "../../../../../../cs/c/../../ChezScheme"/c/alloc.c alloc.c
gcc  -g -O2 -Wall -DOS_X -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -c -DAARCH64 -I../boot/tarm64osx -I../zlib -I../lz4/lib -I/opt/X11/include/  alloc.c
ln -s "../../../../../../cs/c/../../ChezScheme"/c/symbol.c symbol.c
gcc  -g -O2 -Wall -DOS_X -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -c -DAARCH64 -I../boot/tarm64osx -I../zlib -I../lz4/lib -I/opt/X11/include/  symbol.c
ln -s "../../../../../../cs/c/../../ChezScheme"/c/intern.c intern.c
gcc  -g -O2 -Wall -DOS_X -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -c -DAARCH64 -I../boot/tarm64osx -I../zlib -I../lz4/lib -I/opt/X11/include/  intern.c
ln -s "../../../../../../cs/c/../../ChezScheme"/c/gcwrapper.c gcwrapper.c
gcc  -g -O2 -Wall -DOS_X -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -c -DAARCH64 -I../boot/tarm64osx -I../zlib -I../lz4/lib -I/opt/X11/include/  gcwrapper.c
ln -s "../../../../../../cs/c/../../ChezScheme"/c/gc-011.c gc-011.c
ln -s "../../../../../../cs/c/../../ChezScheme"/c/gc.c gc.c
gcc  -g -O2 -Wall -DOS_X -Wpointer-arith -Wall -Wextra -Werror -Wno-implicit-fallthrough -c -DAARCH64 -I../boot/tarm64osx -I../zlib -I../lz4/lib -I/opt/X11/include/  gc-011.c
In file included from gc-011.c:22:
./gc.c:1350:19: error: unknown register name 'x12' in asm
                  LOCKED_DECR(addr, b);
                  ^
../boot/tarm64osx/scheme.h:310:43: note: expanded from macro 'LOCKED_DECR'
                        : "cc", "memory", "x12", "x7");\
                                          ^
1 error generated.
```

It looks like `enable_ios` is being ignored by CS's `configure` script.

I'm running macOS 10.15.7 with XCode 12.2.

Bogdanp:
Closes #3605 .

This gets everything to compile cleanly for iOS. The result can then be linked and run o a real device, but memory allocation fails because `MAP_JIT` is not supported on iOS so more work is required to get everything working. This seems like a good first step.

### DIFF
--- a/racket/collects/compiler/cm.rkt
+++ b/racket/collects/compiler/cm.rkt
@@ -15,9 +15,6 @@
          make-compile-lock
          compile-lock->parallel-lock-client
 
-         ;; re-used in setup/setup-core and setup/private/pkg-deps
-         cross-multi-compile?
-
          (contract-out
           [current-path->mode
            (parameter/c (or/c #f (-> path? (and/c path? relative-path?))))]))
--- a/racket/collects/compiler/cm.rkt
+++ b/racket/collects/compiler/cm.rkt
@@ -15,6 +15,9 @@
          make-compile-lock
          compile-lock->parallel-lock-client
 
+         ;; re-used in setup/setup-core and setup/private/pkg-deps
+         cross-multi-compile?
+
          (contract-out
           [current-path->mode
            (parameter/c (or/c #f (-> path? (and/c path? relative-path?))))]))
--- a/racket/src/ChezScheme/c/Mf-arm64osx
+++ b/racket/src/ChezScheme/c/Mf-arm64osx
@@ -16,7 +16,7 @@
 m ?= arm64osx
 Cpu ?= AARCH64
 
-mdinclude = -I/opt/X11/include/ 
+mdinclude = -I/opt/X11/include/
 o = o
 mdsrc ?= arm32le.c
 mdobj ?= arm32le.o
@@ -39,7 +39,7 @@ ${Scheme}: ${Kernel} ${KernelLinkDeps} ${Main}
 	$C -o ${Scheme} ${Main} ${Kernel} ${KernelLinkLibs} ${LDFLAGS} ${LIBS}
 
 ../zlib/configure.log:
-	(cd ../zlib; ${SetConfigEnv} CFLAGS="${CFLAGS}" ./configure)
+	(cd ../zlib; ${SetConfigEnv} CPPFLAGS="${CPPFLAGS}" CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}" ./configure)
 
 ../lz4/lib/liblz4.a: ${LZ4Sources}
-	(cd ../lz4/lib; ${SetConfigEnv} CFLAGS="${CFLAGS}" ${MAKE} liblz4.a)
+	(cd ../lz4/lib; ${SetConfigEnv} CPPFLAGS="${CPPFLAGS}" CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}" ${MAKE} liblz4.a)
--- a/racket/src/ChezScheme/c/Mf-arm64osx
+++ b/racket/src/ChezScheme/c/Mf-arm64osx
@@ -39,7 +39,7 @@ ${Scheme}: ${Kernel} ${KernelLinkDeps} ${Main}
 	$C -o ${Scheme} ${Main} ${Kernel} ${KernelLinkLibs} ${LDFLAGS} ${LIBS}
 
 ../zlib/configure.log:
-	(cd ../zlib; ${SetConfigEnv} CPPFLAGS="${CPPFLAGS}" CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}" ./configure)
+	(cd ../zlib; ${SetConfigEnv} CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}" ./configure)
 
 ../lz4/lib/liblz4.a: ${LZ4Sources}
-	(cd ../lz4/lib; ${SetConfigEnv} CPPFLAGS="${CPPFLAGS}" CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}" ${MAKE} liblz4.a)
+	(cd ../lz4/lib; ${SetConfigEnv} CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}" ${MAKE} liblz4.a)
--- a/racket/src/ChezScheme/c/alloc.c
+++ b/racket/src/ChezScheme/c/alloc.c
@@ -34,9 +34,6 @@ void S_alloc_init() {
             S_G.bytes_of_generation[g] = 0;
             for (s = 0; s <= max_real_space; s++) {
               S_G.main_thread_gc.base_loc[g][s] = FIX(0);
-#if defined(WRITE_XOR_EXECUTE_CODE)
-              S_G.main_thread_gc.base_bytes[g][s] = 0;
-#endif
               S_G.main_thread_gc.next_loc[g][s] = FIX(0);
               S_G.main_thread_gc.bytes_left[g][s] = 0;
               S_G.bytes_of_space[g][s] = 0;
@@ -257,9 +254,6 @@ ptr S_find_more_gc_room(thread_gc *tgc, ISPC s, IGEN g, iptr n, ptr old) {
   new_bytes = nsegs * bytes_per_segment;
 
   tgc->base_loc[g][s] = new;
-#if defined(WRITE_XOR_EXECUTE_CODE)
-  tgc->base_bytes[g][s] = new_bytes;
-#endif
   tgc->sweep_loc[g][s] = new;
   tgc->bytes_left[g][s] = (new_bytes - n) - ptr_bytes;
   tgc->next_loc[g][s] = (ptr)((uptr)new + n);
--- a/racket/src/ChezScheme/c/alloc.c
+++ b/racket/src/ChezScheme/c/alloc.c
@@ -33,8 +33,11 @@ void S_alloc_init() {
         for (g = 0; g <= static_generation; g++) {
             S_G.bytes_of_generation[g] = 0;
             for (s = 0; s <= max_real_space; s++) {
-              S_G.main_thread_gc.base_loc[g][s] = FIX(0); 
-              S_G.main_thread_gc.next_loc[g][s] = FIX(0); 
+              S_G.main_thread_gc.base_loc[g][s] = FIX(0);
+#if defined(WRITE_XOR_EXECUTE_CODE)
+              S_G.main_thread_gc.base_bytes[g][s] = 0;
+#endif
+              S_G.main_thread_gc.next_loc[g][s] = FIX(0);
               S_G.main_thread_gc.bytes_left[g][s] = 0;
               S_G.bytes_of_space[g][s] = 0;
             }
@@ -224,6 +227,9 @@ static void close_off_segment(thread_gc *tgc, ptr old, ptr base_loc, ptr sweep_l
     /* in case this is during a GC, add to sweep list */
     si = SegInfo(addr_get_segment(base_loc));
     si->sweep_start = sweep_loc;
+#if defined(WRITE_XOR_EXECUTE_CODE)
+    si->sweep_bytes = bytes;
+#endif
     si->sweep_next = tgc->sweep_next[g][s];
     tgc->sweep_next[g][s] = si;
   }
@@ -251,9 +257,21 @@ ptr S_find_more_gc_room(thread_gc *tgc, ISPC s, IGEN g, iptr n, ptr old) {
   new_bytes = nsegs * bytes_per_segment;
 
   tgc->base_loc[g][s] = new;
+#if defined(WRITE_XOR_EXECUTE_CODE)
+  tgc->base_bytes[g][s] = new_bytes;
+#endif
   tgc->sweep_loc[g][s] = new;
   tgc->bytes_left[g][s] = (new_bytes - n) - ptr_bytes;
   tgc->next_loc[g][s] = (ptr)((uptr)new + n);
+
+#if defined(WRITE_XOR_EXECUTE_CODE)
+  if (s == space_code) {
+    /* Ensure allocated code segments are writable. The caller should
+       already have bracketed the writes with calls to start and stop
+       so there is no need for a stop here. */
+    S_thread_start_code_write(tgc->tc, 0, 1, NULL);
+  }
+#endif
 
   if (tgc->during_alloc == 1) maybe_queue_fire_collector(tgc);
 
@@ -272,6 +290,9 @@ void S_close_off_thread_local_segment(ptr tc, ISPC s, IGEN g) {
   close_off_segment(tgc, tgc->next_loc[g][s], tgc->base_loc[g][s], tgc->sweep_loc[g][s], s, g);
 
   tgc->base_loc[g][s] = (ptr)0;
+#if defined(WRITE_XOR_EXECUTE_CODE)
+  tgc->base_loc[g][s] = 0;
+#endif
   tgc->bytes_left[g][s] = 0;
   tgc->next_loc[g][s] = (ptr)0;
   tgc->sweep_loc[g][s] = (ptr)0;
--- a/racket/src/ChezScheme/c/prim5.c
+++ b/racket/src/ChezScheme/c/prim5.c
@@ -879,10 +879,10 @@ static ptr s_set_code_byte(p, n, x) ptr p, n, x; {
     I8 *a;
     ptr tc = get_thread_context();
 
-    S_thread_start_code_write(tc, 0);
     a = (I8 *)TO_VOIDP((uptr)p + UNFIX(n));
+    S_thread_start_code_write(tc, 0, a);
     *a = (I8)UNFIX(x);
-    S_thread_end_code_write(tc, 0);
+    S_thread_end_code_write(tc, 0, a);
 
     return Svoid;
 }
@@ -891,10 +891,10 @@ static ptr s_set_code_word(p, n, x) ptr p, n, x; {
     I16 *a;
     ptr tc = get_thread_context();
 
-    S_thread_start_code_write(tc, 0);
     a = (I16 *)TO_VOIDP((uptr)p + UNFIX(n));
+    S_thread_start_code_write(tc, 0, a);
     *a = (I16)UNFIX(x);
-    S_thread_end_code_write(tc, 0);
+    S_thread_end_code_write(tc, 0, a);
 
     return Svoid;
 }
@@ -903,10 +903,10 @@ static ptr s_set_code_long(p, n, x) ptr p, n, x; {
     I32 *a;
     ptr tc = get_thread_context();
 
-    S_thread_start_code_write(tc, 0);
     a = (I32 *)TO_VOIDP((uptr)p + UNFIX(n));
+    S_thread_start_code_write(tc, 0, a);
     *a = (I32)(Sfixnump(x) ? UNFIX(x) : Sinteger_value(x));
-    S_thread_end_code_write(tc, 0);
+    S_thread_end_code_write(tc, 0, a);
 
     return Svoid;
 }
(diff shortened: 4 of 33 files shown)


racket/src/cs/c/configure.ac
```diff
@@ -611,9 +639,11 @@ fi
 
 ############## other inferred machine properties ################
 
-AC_CHECK_SIZEOF([void *])
+if test "${enable_ios}" = "" ; then
```
Bogdanp, line 642: I'm not sure why this check fails yet and it would be good to get it to work before merging. The size that is detected is 0 and I'm not sure why.

racket/src/ChezScheme/c/version.h
```diff
@@ -326,7 +326,11 @@ typedef int tputsputcchar;
 #endif
 #if defined(__arm64__)
 # define S_MAP_CODE  MAP_JIT
```
mflatt, line 328: Does moving this to the non-`TARGET_OS_IPHONE` branch below help?

racket/src/ChezScheme/c/segment.c
```diff
@@ -88,14 +88,14 @@ void *S_getmem(iptr bytes, IBOOL zerofill, UNUSED IBOOL for_code) {
 
   if ((addr = malloc(bytes)) == (void *)0) out_of_memory();
 
-  debug(printf("getmem(%p) -> %p\n", bytes, addr))
+  debug(printf("getmem(%ld) -> %p\n", bytes, addr))
```
mflatt, line 91: A more portable repair here is probably to use `TO_VOIDP` on the `addr` argument and keep `%p`.

racket/src/cs/c/configure.ac
```diff
@@ -192,6 +192,32 @@ if test "$ARFLAGS" = '' ; then
   ARFLAGS=rc
 fi
 
+if test "${enable_ios}" != "" ; then
```
mflatt, line 195: Could more of this be in a file in "racket/src/ac" to be included by both BC and CS configure scripts?

racket/src/ChezScheme/c/Mf-arm64osx
```diff
@@ -39,7 +39,7 @@ ${Scheme}: ${Kernel} ${KernelLinkDeps} ${Main}
 	$C -o ${Scheme} ${Main} ${Kernel} ${KernelLinkLibs} ${LDFLAGS} ${LIBS}
 
 ../zlib/configure.log:
-	(cd ../zlib; ${SetConfigEnv} CFLAGS="${CFLAGS}" ./configure)
+	(cd ../zlib; ${SetConfigEnv} CPPFLAGS="${CPPFLAGS}" CFLAGS="${CFLAGS}" LDFLAGS="${LDFLAGS}" ./configure)
```
mflatt, line 42: The Racket makefile currently folds `CPPFLAGS` into `CFLAGS` to pass to the Chez Scheme makefiles — and so on to the zlib and lz4 makefiles — because distinguishing them is a pain, and I don't think CS will need the distinction in the long run that BC needed. Does combing them into `CFLAGS` not work here, though?

mflatt (commented):
Thanks for working on this!

racket/src/ChezScheme/c/version.h
Bogdanp: Yup, that does help get past the allocation problem! Now it aborts because of a normal return from `S_generic_invoke`:

```
Thread 1 Queue : com.apple.main-thread (serial)
#0	0x00000001086a0150 in 0x1086a0150 ()
#1	0x00000001022aa414 in S_call_help ()
#2	0x00000001022a904c in boot_call at /Users/bogdan/sandbox/racket-ios/racket/src/build/cs/c/ChezScheme/tarm64osx/c/scheme.c:353
#3	0x00000001022a8f38 in load at /Users/bogdan/sandbox/racket-ios/racket/src/build/cs/c/ChezScheme/tarm64osx/c/scheme.c:903
#4	0x00000001022a8c00 in Sbuild_heap at /Users/bogdan/sandbox/racket-ios/racket/src/build/cs/c/ChezScheme/tarm64osx/c/scheme.c:1160
#5	0x00000001022b305c in racket_boot at /Users/bogdan/sandbox/racket-ios/racket/src/cs/c/boot.c:156
#6	0x00000001021ee088 in racket_init at /Users/bogdan/sandbox/rbggame/shared/game.c:19
#7	0x00000001021eefe4 in AppDelegate.application(_:didFinishLaunchingWithOptions:) at /Users/bogdan/sandbox/rbggame/ios/rbggame/rbggame/AppDelegate.swift:21
#8	0x00000001021efff8 in @objc AppDelegate.application(_:didFinishLaunchingWithOptions:) ()
#9	0x00000001acc92540 in -[UIApplication _handleDelegateCallbacksWithOptions:isSuspended:restoreState:] ()
#10	0x00000001acc945a8 in -[UIApplication _callInitializationDelegatesWithActions:forCanvas:payload:fromOriginatingProcess:] ()
#11	0x00000001acc9a004 in -[UIApplication _runWithMainScene:transitionContext:completion:] ()
#12	0x00000001ac2f1950 in -[_UISceneLifecycleMultiplexer completeApplicationLaunchWithFBSScene:transitionContext:] ()
#13	0x00000001acc95db0 in -[UIApplication _compellApplicationLaunchToCompleteUnconditionally] ()
#14	0x00000001acc961dc in -[UIApplication _run] ()
#15	0x00000001acc9ba74 in UIApplicationMain ()
#16	0x00000001bdb2d5bc in UIApplicationMain(_:_:_:_:) ()
#17	0x00000001021f07d8 in static UIApplicationDelegate.main() ()
#18	0x00000001021f0750 in static AppDelegate.$main() ()
#19	0x00000001021f081c in main ()
#20	0x00000001a9f1d6c0 in start ()
```

I'll be able to dig into this a little more on Saturday.

Bogdanp: The problem here was actually just that the code segments weren't executable. Explicitly flipping between W and X in 56916d6459dd96378882c8df8a3d2c4e16d54fd6 gets things much farther. `petite.boot` and `scheme.boot` both boot up properly now, but I get an issue when loading `racket.boot`:

```
fasl-read: incompatible fasl-object machine-type 'ta6osx found in #<binary input port bytevector>
```

so something about the build must still be incorrect.

Bogdanp: Ah, actually, the issue is with how I'm generating the C module from my Racket code.  Racket itself boots up properly after the changes in https://github.com/racket/racket/commit/56916d6459dd96378882c8df8a3d2c4e16d54fd6 ! 

I'm using this command to generate a C mod from my Racket code:

   racket -C -G ~/sandbox/racket-ios/racket/etc/ -X ~/sandbox/racket-ios/racket/collects -l- raco ctool --c-mods shared/embed.c core/main.rkt

Unless I'm doing something wrong, there must be some kind of bug in `cext-lib` that's causing it to produce output for the source machine.

mflatt: For cross-compilation to work with CS, cross-compiler "patch" files need to be specified through the `--cross-compiler` argument to `racket`. I see that this is not explained in the documentation, and it's really not worked out fully; its there just enough to cross-build distributions, right now.

For a distribution build, the second argument to `--cross-compiler` points to "racket/src/build/cs/c". I think the two files needed from there will be "compile-xpatch.tarm64osx" and "library-xpatch.tarm64osx", so those could be copied to a different directory to be given to `--cross-compilers`. Probably, the `make install` step should copy the patch files to `racket/lib`.

mflatt: Swapping pages between `PROT_WRITE` and `PROT_EXEC` is not going to be a good long-term solution. For one thing, it's slow. More importantly, it can't work right with threads, because one thread might be trying to run some (immobile) code while another thread needs to write on the same pages.

My vague impression ws that W&X is allowed on iOS with the right entitlements and maybe flags like `MAP_JIT`. Is that just not right?

mflatt: On further reflection, I'm not clear on why `raco ctool` didn't complain about a missing `tarm64osx` cross compiler. Maybe `tarm64osx` as the target machine is not recorded correctly in "~/sandbox/racket-ios/racket/etc/config.rktd", in which case using `--compile-machine tarm64osx` may work around the problem. Or maybe Racket just doesn't complain about a missing cross compiler like it should.

Bogdanp: > My vague impression ws that W&X is allowed on iOS with the right entitlements and maybe flags like MAP_JIT. Is that just not right?

As far as I understand, WX via `MAP_JIT` is only allowed on iOS using a blessed entitlement that only Apple has access to and which they themselves use in Safari. You can probably add it to a project and get a debug build to work, but that build would be rejected by the App Store review process. It used to be possible to [map the same page twice with different virtual addresses and permissions](https://saagarjha.com/blog/2020/02/23/jailed-just-in-time-compilation-on-ios/), but that was patched in iOS 14. Recently, iOS 14.2 introduced a new capability called "Extended Virtual Addressing" that supposedly allows JIT compilation, but it's undocumented and it appears to require the `get-task-allow` entitlement to work, which is not allowed on the App Store. 

It's unfortunate, but it seems like swapping page protection flags is the only viable option at the moment. 

> More importantly, it can't work right with threads, because one thread might be trying to run some (immobile) code while another thread needs to write on the same pages.

It's definitely not ideal, but I think applications could load all code before launching places (no `dynamic-require`, no `eval`) and avoid futures and still get some of the benefits of threading. Do I have that wrong? Are there more language features that would be limited by this approach?

Thank you for the pointers re. cross-compilation! I'll dig into that tomorrow.

mflatt: With places, I worry that there are cases where an `eval` might show up or code might be lazily unpacked (due to a large module whose outer layers are interpreted), so being sure that everything works is be very difficult. But since allocation will be place-local even in those cases, probably things will work *almost* always.

Bogdanp: Re. cross-compilation, what do you think of the change in https://github.com/racket/racket/pull/3607/commits/75124e3ed1b55d2bf7e6406e7fb69ef96c4b437f ? Is it too much of a hack? The best fix here would probably be to port the fasl reader in "fasl.c" to Scheme so that it can be patched for cross-compilation like the writer is, but that seems like a significant amount of work. Another approach might be to re-structure `compiler/embed` so that it always uses sources to lay things out, but that won't work if sources aren't available (which might be an OK compromise).

If the changes to `compiler/embed` seem OK, I'll add a change to update the "Inside Racket CS" docs with a section on how to cross compile module images.

Re. threading, I've added a doc note describing the limitations. Alternatively, we could disable places for iOS cross-builds (though my preference is to leave them on despite the limitations).

mflatt: While there is already a second reader for Chez Scheme fasl files that is written in Scheme (it's in "strip.ss", also used for "vfasl.ss"), I don't think it's just a matter of reading the code. Some compile-time code must be run to extract the paths — that is, `eval` is used here. Unless I misunderstand, that means there's just no way to use compiled code for a different platform. I haven't understood every detail of the change for embedding, but it seems like the right idea.

Leaving places enabled seems ok. Unfortunately, it's not necessarily just about places, because some relatively low-level libraries may create threads for other reasons. The recent readline change is an example, although that example is not so relevant to iOS.

I wonder whether the chance for W/X flipping to go bad with threads could be reduced further by flipping only for segments in new enough generations. In a GC, there's an upper bound on the segments that might be modified. In fasl reading and other code-allocation cases, only generation-0 segments can be modified (except vfasl loading, which might be in direct-to-static mode). Or maybe it's so much simpler to work at the chunk level that it's not worth trying to constrain to certain segments.

Bogdanp: > Some compile-time code must be run to extract the paths — that is, eval is used here. 

Ah! That makes sense and I even went over the code that does this while making the change but did not make this connection. The approach seems more reasonable now. I'll adjust the comments to reflect this.

The change looks much bigger than it is because I had to move some functions around for re-use. The relevant bits are mainly just the ones starting with [this comment](https://github.com/racket/racket/commit/75124e3ed1b55d2bf7e6406e7fb69ef96c4b437f#diff-a40a04285d3ca3db269357d8c60fdbb2464dc302831b3d9079f540a56fc79340R747). Aside from replacing the host code with the target code, the module for that code must also be "prepared" (re-named, submodules potentially removed) in the same ways that the host code was.

> I wonder whether the chance for W/X flipping to go bad with threads could be reduced further by flipping only for segments in new enough generations. In a GC, there's an upper bound on the segments that might be modified. In fasl reading and other code-allocation cases, only generation-0 segments can be modified (except vfasl loading, which might be in direct-to-static mode). Or maybe it's so much simpler to work at the chunk level that it's not worth trying to constrain to certain segments.

I'll take a look at this and report back.

Bogdanp: What do you think of https://github.com/racket/racket/pull/3607/commits/d84c2eabd68d578626e42ab41c57b48d714ddb00 ? The code could be improved (in particular, I'd want to get rid of `is_unused_seg` or at least add a special-case when a chunk has no used segments), but I wanted to first see if this is along the lines of what you were thinking.

racket/src/ChezScheme/c/prim5.c
```diff
@@ -878,63 +878,63 @@ static char *s_getwd() {
 static ptr s_set_code_byte(p, n, x) ptr p, n, x; {
     I8 *a;
 
-    S_thread_start_code_write();
+    S_thread_start_code_write(0, static_generation);
```
mflatt, line 881: This could be `static_generation-1`. But, also, this seems extremely expensive if code is ever compiled on the flay. Probably the `c-mkcode` loop in "compile.ss" needs to be made non-interruptable and enable/disable writing just once.

mflatt (commented):
Some thoughts on individual calls to `S_thread_start_code_write`, in case this fine-grained approach is a good idea.

racket/src/ChezScheme/c/prim5.c
mflatt: The `c-mkcode` loop is already in a critical section combined with allocating the code, so now I think this can just be 0.

mflatt: ... although it still seems expensive to keep toggling.

racket/src/ChezScheme/c/version.h
mflatt: I think the changes are on the right track for this idea. And it looks possible to further confine toggling: in cases where toggling is confined to generation 0, that could mean generation 0 for segments that are owned by the current thread (as reflected by the `creator_gc` field in `seginfo`), because it's always enabling writes to just-allocated memory.

I'm still not sure this is worthwhile, since it seems so difficult to make it something you can really reason about and rely on. But maybe it works in practice, and maybe it's worthwhile to keep shrinking the problem surface like this.

Minor: I'd get rid of the `mingen` argument, since it's always 0.

mflatt: Idea for an even further refinement: Starting code write for generation 0 really means the current code allocation segments for the current thread (starts as `base_loc[0][space_code]`) plus any newly allocated segments (via `S_find_more_gc_room`). Ending code write for 0 needs to turn on execution for the current allocation segments plus other newly allocated segments. The `sweep_next` chain in `seginfo`, which currently serves a role only for segments allocated during a collection, could be used to track the latter.

Bogdanp: I was able to limit flipping to segments belonging to the current thread for generation 0, but had trouble getting the second suggestion to work. I'll give it another go this weekend since I'll have more time then to dig into what exactly is going wrong. 

As a quick stress-test for the two approaches, I set up a project that runs `eval` on a loop inside a place and lets the user `eval` arbitrary code on the main thread. I haven't been able to get it to crash once with the changes to confine toggling, whereas I can easily get it to crash with the coarse-grained approach.

mflatt: A thought on constraining generation-0 writes to active allocation segments: The functions `$set-code-word!`, etc., and `$make-relocation-table!` will be called many times for a given code object, so may not be obvious what "current code allocation segment" means for each individual call. A better strategy for those functions may be to enable writing for just the specific segment containing the address that's an argument to those functions.

For fasl reading, meanwhile, the calls to start and end bracket all allocation and corresponding modification of code, so it makes sense to track the current allocation region there. And for GC, the existing strategy should be ok.

Here are the two remaining problems I see:
 * `__collect_safe` callbacks during `__collect_safe` callouts just can't work. Fortunately, `__collect_safe` functionality (`#:blocking? #t` at the Racket level) is not needed often, but it is needed for the GUI library on Mac OS, and I can imagine that it would be useful for similar reasons on iOS.
 * A function that is compiled and then immediately passed to another Scheme thread runs the risk of becoming non-executable when the original allocating thread allocates more. If a program controls threads and sharing enough, it may be able to avoid doing that, at worst with a `(collect-garbage)` just before passing the new code to the other thread.

Bogdanp: Thank you for the suggestions! I've added segment hinting in https://github.com/racket/racket/pull/3607/commits/bed4988b29d44091a0bdad50c373e84e8da920d1 (with portability fixes in https://github.com/racket/racket/pull/3607/commits/4b385623333decd976037ce26cc6844ee1a69ce9) and the ability to flip just the current allocation region in https://github.com/racket/racket/pull/3607/commits/75468bc10e2ff0a9df076564734ef0e2766579ff . For hinting, I had to use the effective address rather than the code pointers since the offsets could cause the modifications to be on different segments. For the current region changes, I think I could've avoided adding `sweep_bytes` to `seginfo` and walked the segment up to the forward marker to determine its length, but this approach seemed more clean.

> __collect_safe callbacks during __collect_safe callouts just can't work. Fortunately, __collect_safe functionality (#:blocking? #t at the Racket level) is not needed often, but it is needed for the GUI library on Mac OS, and I can imagine that it would be useful for similar reasons on iOS.

Would it make sense for now to make these fail with an explicit error when `WRITE_XOR_EXECUTE_CODE` is defined?

racket/collects/compiler/cm.rkt
```diff
@@ -15,6 +15,9 @@
          make-compile-lock
          compile-lock->parallel-lock-client
 
+         ;; re-used in setup/setup-core and setup/private/pkg-deps
+         cross-multi-compile?
```
mflatt, line 19: Probably this function should be put into its own module, so the pubic interface of `compiler/cm` doesn't grow, and so it can be used without importing all the rest of `compiler/cm`.

racket/src/ChezScheme/c/alloc.c
```diff
@@ -33,8 +33,11 @@ void S_alloc_init() {
         for (g = 0; g <= static_generation; g++) {
             S_G.bytes_of_generation[g] = 0;
             for (s = 0; s <= max_real_space; s++) {
-              S_G.main_thread_gc.base_loc[g][s] = FIX(0); 
-              S_G.main_thread_gc.next_loc[g][s] = FIX(0); 
+              S_G.main_thread_gc.base_loc[g][s] = FIX(0);
+#if defined(WRITE_XOR_EXECUTE_CODE)
+              S_G.main_thread_gc.base_bytes[g][s] = 0;
```
mflatt, line 38: Can `base_bytes` be computed as `next_loc - base_loc + bytes_left`?

racket/src/ChezScheme/c/segment.c
```diff
@@ -568,18 +575,154 @@ static void contract_segment_table(uptr base, uptr end) {
...
+   chances of a page being flipped while a thread is executing code
+   off of it.
+*/
 
-void S_thread_start_code_write(void) {
+void S_thread_start_code_write(ptr tc, IGEN maxg, IBOOL current, void *hint) {
+#if defined(WRITE_XOR_EXECUTE_CODE)
+  enable_code_write(tc, maxg, 1, current, hint);
+#else
+  (void)tc;
```
mflatt, line 594: Normally, an `UNUSED` annotation is written on the function argument to declare unused variables. In this case, of course, it's only unused when `WRITE_XOR_EXECUTE_CODE`, so I'd define a new `UNUSED_EXCEPT_W_XOR_X` (or something with a better name) and use that.

racket/src/ChezScheme/c/version.h
mflatt: I'm not sure how practical it will be to make those fail. Using `__collect_safe` with a foreign call is not necessarily broken. I guess using `_collect_safe` with a callback could be disabled; although that's not actually a problem when called though a callback without `__collect_safe`, putting `_collect_safe` on a callback declares an intent. But the `WRITE_XOR_EXECUTE_CODE` mode is not curently reflected to the Scheme layer where `foreign-callback` lives. I' inclined to leave this to documentation.

PR #3610, opened 2021-01-07
bc: make srcloc->string compatible with cs

### LINKED ISSUE #1371: read-syntax returns syntax with negative column position
The following program seems to return -1, which breaks the documented contract of `syntax-column`, and it seems like `set-port-next-location!` should cause the next column to be `#f` rather than `-1`, but I'm not sure what the correct behavior should be.

```
#lang racket

(define p (open-input-string "foo"))
(port-count-lines! p)
(set-port-next-location! p #f #f #f)
(define stx (read-syntax 'dummy p))
(syntax-column stx)
```

sorawee:
Currently, in Racket BC, the following program:

```
(srcloc->string (make-srcloc "x.rkt" #f #f 90 #f))
(srcloc->string (make-srcloc "x.rkt" #f 80 #f #f))
(srcloc->string (make-srcloc "x.rkt" #f #f #f #f))

(srcloc->string (make-srcloc "x.rkt" #f 80 90 #f))
(srcloc->string (make-srcloc "x.rkt" 70 #f 90 #f))
(srcloc->string (make-srcloc "x.rkt" 70 80 #f #f))
(srcloc->string (make-srcloc "x.rkt" 70 80 90 #f))
```

results in:

```
"x.rkt::90"
"x.rkt::80"
"x.rkt::-1"
"x.rkt::80"
"x.rkt:70:90"
"x.rkt:70:80"
"x.rkt:70:80"
```

This output is very confusing and inconsistent.
When we see "x.rkt::90", we can never be sure if it's a srcloc
whose position is 90, or a srcloc whose column is 90.
The same applies for "x.rkt:70:90".

Moreover, the srloc "x.rkt::-1" is weird and is arguably incorrect
(see #1371).

Code-wise, the line
https://github.com/racket/racket/blob/e71c217758f744d3c8dfb30dd152b8c7d08a2735/racket/src/bc/src/error.c#L2151
is a dead code, because if col < 0, then col will be set to pos + 1
which is guaranteed to be non-negative. This offers another evidence
that the current srcloc->string is incorrect.

In Racket CS, the program results in:

```
"x.rkt::90"
"x.rkt::#f"
"x.rkt::#f"
"x.rkt::90"
"x.rkt::90"
"x.rkt:70:80"
"x.rkt:70:80"
```

which is consistent and makes sense.

This PR makes the behavior of srcloc->string in Racket BC matches
that of Racket CS.

Fixes #1371

### DIFF
--- a/pkgs/racket-test-core/tests/racket/read.rktl
+++ b/pkgs/racket-test-core/tests/racket/read.rktl
@@ -1472,6 +1472,15 @@
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 ;; srcloc->string
 
+(test "x.rkt::#f" srcloc->string (make-srcloc "x.rkt" #f #f #f #f))
+(test "x.rkt::90" srcloc->string (make-srcloc "x.rkt" #f #f 90 #f))
+(test "x.rkt::#f" srcloc->string (make-srcloc "x.rkt" #f 80 #f #f))
+
+(test "x.rkt::90" srcloc->string (make-srcloc "x.rkt" #f 80 90 #f))
+(test "x.rkt::90" srcloc->string (make-srcloc "x.rkt" 70 #f 90 #f))
+(test "x.rkt:70:80" srcloc->string (make-srcloc "x.rkt" 70 80 #f #f))
+(test "x.rkt:70:80" srcloc->string (make-srcloc "x.rkt" 70 80 90 #f))
+
 (test "x.rkt:10:11" srcloc->string (make-srcloc "x.rkt" 10 11 100 8))
 (test "x.rkt::100" srcloc->string (make-srcloc "x.rkt" #f #f 100 8))
 (test "x.rkt::100" srcloc->string (chaperone-struct (make-srcloc "x.rkt" #f #f 100 8)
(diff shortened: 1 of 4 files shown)


soegaard:
Are there tools that needs to be updated due to this change?
(Maybe racket-mode?)
@greghendershott

mflatt:
This change looks like the right direction to me.

For the result that have `#f`, I think the right choice is probably not to have `::#f` and just return the path. (This has bugged me for a while, but I never investigated.) Can we make that change to both BC and CS?

sorawee:
@soegaard IIUC, I think it will not affect Racket Mode. Racket Mode primarily operates on srcloc that has both line and column information, but this PR concerns the cases where one of the information is missing.

@mflatt done!

greghendershott:
I did a thumbs-up on @sorawee 's comment but for people reading in email; I agree this shouldn't affect Racket Mode.

greghendershott:
p.s. OT:

My bane is things that change useful absolute pathnames into not-useful relative ones.

- So the only thing I "hate" about `srcloc->string` is when it does the `<prefix>/` thing. I have hacks attempting to prevent that and also attempting to "undo" it. None can be 100% effective.

- Separately, at some deeper level (both BC and CS AFAICT) Racket will truncate `"/path/slightly/too/long/file.rkt"` to `".../slightly/too/long/file.rkt"`. I have no hacks for that.

PR #3614, opened 2021-01-09
doc: fix unbalanced braces and merge excessive history elements

sorawee:


sorawee:
CC: @samth 

Should this fix be added to the release branch?

Issue #3609, opened 2021-01-06
Subprocess signal handling


willghatch:
Subprocesses spawned with the `subprocess` function inherit Racket's signal handling.  I'm not sure if there is a reason we want non-default signal handlers in the child process.  But at least for Rash I do generally want default signal handlers.  Eg. right now when a child process gets SIGPIPE, instead of exiting with status 141 (which I can detect and potentially handle as a non-failing pipeline stage), the process exits with status 1 and prints a message about a broken pipe to stderr.

The only signal handling I'm aware of that needs to happen with regards to fork/exec for a subprocess is SIGCHLD, which the parent needs to handle, not the child.

So unless there is a reason for child processes to inherit some of Racket's signal handling, I propose that `rktio_process` should set signal handlers to `SIG_DFL` before running `exec`.  If there is a reason for the child to inherit Racket's signal handling, I propose that we add some kind of option for it.

mflatt:
@willghatch Does the commit above look like the right idea, especially the part in "rktio_signal.c"?

mflatt:
I'm starting to think that this is too complicated, after all, and it's not so bad for rktio to just know the signal-ID range for each supported platform.

Originally, I had in mind saving the startup state (i.e., maybe some process started Racket with ignored signals, for example), but moved towards setting all signals to their default and setting the signal mask to empty. And after going that direction, maybe there's no need to track a specific set of signals.

Even more simply, it looks like just iterating from 1 to `SIGPROF` is going to work fine in practice. At worst, the range from 1 to `SIGPROF` is too narrow for some platform, and then we're just back to the current state until it's fixed for that platform.

willghatch:
I think tracking the original state is the best solution, since without it Racket can't compose with eg. `nohup`, which I know people use and generally expect to work.  I think c697f11 is going in the right direction for that, simply needing a second argument to get the original value.  Is there something that makes that impractical?  If adding another argument to restore the original value works, that's my favorite solution.  If there is something that makes that impractical, then I guess the simpler solution is better for its simplicity.

mflatt:
Saving the state and restoring it works fine — `nohup` is a good example — and I had that in an iteration before c697f11 . So, it's easy to go back, and it will be nice that the work there wasn't wasted.

PR #3616, opened 2021-01-09
rktio_process: set child signal handlers to default

willghatch:
This is a sketch of a solution for issue #3609.

I tried to find all the signals that Racket installs handlers for, and set them to `SIG_DFL` before the `execve` call.

There are some improvements that could still be made here:

* There are various `#define` flags that determine whether various signals are actually overridden.  We should probably use those same flags to determine which signals are reset.
* Rather than setting each handler to `SIG_DFL`, we could set them to the values they had before they were overridden.  This is a bigger change, since we would need to capture those values in all the various places they are set by `signal` or `sigaction` and then have them available here at the function call site.

But before trying to do any of that I thought I would start with this since I haven't yet heard anything as far as a thumbs up/down for this kind of change.  At any rate, the current behavior is undocumented and, in my opinion, a bug.

### DIFF
--- a/racket/src/rktio/rktio_private.h
+++ b/racket/src/rktio/rktio_private.h
@@ -416,6 +416,7 @@ char *rktio_strndup(char *s, intptr_t len);
 
 #ifdef RKTIO_SYSTEM_UNIX
 void rktio_set_signal_handler(int sig_id, void (*proc)(int));
+void rktio_set_default_signal_handlers_for_subprocess(void);
 #endif
 void rktio_forget_os_signal_handler(rktio_t *rktio);
 
--- a/racket/src/rktio/rktio_process.c
+++ b/racket/src/rktio/rktio_process.c
@@ -1532,6 +1532,10 @@ rktio_process_result_t *rktio_process(rktio_t *rktio,
         _exit(1);
       }
 
+      /* Reset signal handling for new process */
+      rktio_set_default_signal_handlers_for_subprocess();
+
+
       /* Exec new process */
 
       {
--- a/racket/src/rktio/rktio_signal.c
+++ b/racket/src/rktio/rktio_signal.c
@@ -105,4 +105,60 @@ void rktio_set_signal_handler(int sig_id, void (*proc)(int))
   sa.sa_handler = proc;
   sigaction(sig_id, &sa, NULL);
 }
+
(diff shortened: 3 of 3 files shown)


mflatt:
Setting signal handlers to default seems clearly the right idea.

The approach here is ok. Another possibility is to have a function that rktio provides for registering signal IDs, and anything that sets a signal handler could (be required to) call that function to register the ID. The latter seems better to me, but it's a lot more work, and Chez Scheme would probably need a function to list the signal IDs that it sets so they could be bounced back to rktio.

Either way, I'd put the implementation in `rktio_signal.c`, and then have `rktio_process.c` use that.

willghatch:
Here is a slightly improved version of the simple approach -- just set all handlers used by Racket to `SIG_DFL`.

I agree that it would be better to track and set the original signal handlers.  For RacketBC I think it would be enough just to have `rktio_set_signal_handler` update an array of handlers for each first use with a new signal ID, then funnel all signal handling through that function.  For CS, though, we need to essentially write that logic twice (IE write the same thing in Chez Scheme), then keep them in sync, maybe by checking the CS array each time before updating the Racket array to see if CS actually had the first signal change.

Anyway, for whatever reason CS doesn't seem to be building for me right now, so I'm not going to try to do the more involved version that involves changing Chez Scheme at the moment.

This version, at any rate, solves most of my problems with Racket signal handling.  The only thing I know of that relies on setting signal handlers and expecting them to propagate to children in practice is the `nohup` command.  This patch does not fix the lack of propagation with `nohup`, but `nohup` really doesn't work with Racket anyway, because Racket sets `SIGHUP` to generate `exn:break:hang-up` regardless of what the original signal handler was.  Maybe Racket could fix this interaction with `nohup` by checking if it's set to `SIG_IGN` and NOT setting its own `SIGHUP` handler.

We could use this partial solution now with a plan to write the other one later, probably with some documentation about what happens with signal handling.  Right now the documentation says nothing about signals, and gives subprocesses an undocumented set of handlers.  This version will set signals handled by Racket to `SIG_DFL` for subprocesses, but should have a disclaimer that in the future Racket will propagate its initial signal handlers to subprocesses.  Should I write this documentation (IE will we use this version for now), or should I wait until I write the better version?

mflatt:
Closed by  2cbaeb7.

PR #3618, opened 2021-01-10
cptypes: fix arity in primitive reduction

gus-massa:
Fix the arity in primitive reduction. This argument is used only in primitives with a rest argument and it is important only if the last argument is different. For example

    (expand/optimize '(lambda (x y z) (append x y z) (vector? y)))

the last expression `(vector?)` was not reduced because the bad arity made cptypes not mark `y` as a `null-or-pair`, because while analyzing `y` it looked somewhat like `(append x y)`. (Anyway, `z` was not dropped, the final expression was correct.)

This is only a missed optimization opportunity, but I guess it's possible to find a case with another primitive where this cause an optimization bug.

It would be nice to merge this in the release branch for v8.0.


samth:
cc @racket/release

samth:
@jbclements I think this is ready to cherry-pick onto the release branch.

PR #3621, opened 2021-01-11
Skip keywords in invalid-option-placement

### LINKED ISSUE #3603: Function header issues
@LeifAndersen 

(First, thanks for writing this, it was very helpful!)

Apologies for filing an issue rather than a PR, I'm a little short on time :-/  
I might write a PR at some later point if someone else doesn't do it first, but first maybe someone has feedback on this issue.

## Issue 1 
https://github.com/racket/racket/blob/dc76e1b7e0f48c185d95d5e212a2f2a4dc7f831a/racket/collects/syntax/parse/lib/function-header.rkt#L55
This is too strict. The following should compile:
```racket
(syntax-parse #'(#:a [a 1] #:b b)
   [a:formals #'(a a.params)])
```
but instead fails with `default-value expression missing`.

I fixed this in a separate package by passing keywords too and checking order only if not keyword:
https://github.com/Metaxal/define2/blob/5462981e51a554f18e4f22ef7d560842156c0035/formals.rkt#L98

## Issue 2 
https://github.com/racket/racket/blob/dc76e1b7e0f48c185d95d5e212a2f2a4dc7f831a/racket/collects/syntax/parse/lib/function-header.rkt#L20
Probably this should use [`check-duplicates`](https://docs.racket-lang.org/reference/pairs.html?q=check-duplicates#%28def._%28%28lib._racket%2Flist..rkt%29._check-duplicates%29%29) with `#:key syntax-e`.

xxyzz:
Close #3603

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
@@ -276,7 +276,7 @@ Note that the literal-set uses the names @racket[#%plain-lambda] and
   [(_ (fml:formal) body ...+) #'(fml fml.name fml.kw fml.default)])
 ]
 
-@defstxclass[splicing-formals-no-rest #:splicing]{
+@defstxclass[formals-no-rest #:splicing]{
  Like @racket[formals] but without dotted-tail identifier.
  @defattribute[params syntax?]{
   The list of parameters.
--- a/pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
@@ -277,7 +277,7 @@ Note that the literal-set uses the names @racket[#%plain-lambda] and
 ]
 
 @defstxclass[splicing-formals-no-rest #:splicing]{
- Like formals but without dotted-tail identifier.
+ Like @racket[formals] but without dotted-tail identifier.
  @defattribute[params syntax?]{
   The list of parameters.
  }
--- a/pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
@@ -275,3 +275,10 @@ Note that the literal-set uses the names @racket[#%plain-lambda] and
 (syntax-parse #'(lambda (#:kw [kw 42]) kw)
   [(_ (fml:formal) body ...+) #'(fml fml.name fml.kw fml.default)])
 ]
+
+@defstxclass[splicing-formals-no-rest #:splicing]{
+ Like formals but without dotted-tail identifier.
+ @defattribute[params syntax?]{
+  The list of parameters.
+ }
+}
--- a/pkgs/racket-test/tests/stxparse/test.rkt
+++ b/pkgs/racket-test/tests/stxparse/test.rkt
@@ -1062,6 +1062,14 @@
 
 ;; from Laurent Orseau, issue #3603 (1/2021)
 (require syntax/parse/lib/function-header)
-(test-case "skip keywords in invalid-option-placement"
+(check-equal?
+ (syntax->datum
   (syntax-parse #'(#:a [a 1] #:b b)
-    [a:formals #'(a a.params)]))
+    [fmls:formals #'(fmls fmls.params)]))
+ '((#:a (a 1) #:b b) (a b)))
+
+(check-equal?
+ (syntax->datum
+  (syntax-parse #'(a #:b [b 1] c #:d d [e 2] #:f [f 3] [g 4] . rest)
+    [fmls:formals #'(fmls fmls.params)]))
+ '((a #:b [b 1] c #:d d [e 2] #:f [f 3] [g 4] . rest) (a b c d e f g rest)))
--- a/pkgs/racket-test/tests/stxparse/test.rkt
+++ b/pkgs/racket-test/tests/stxparse/test.rkt
@@ -1086,3 +1086,17 @@
    (syntax-parse #'(a [b 1] c)
      #:context 'me
      [fmls:formals #'(fmls fmls.params)])))
+
+(check-exn
+ #rx"me: duplicate argument identifier"
+ (lambda ()
+   (syntax-parse #'(a . a)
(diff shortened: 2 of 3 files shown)


racket/collects/syntax/parse/lib/function-header.rkt
```diff
@@ -52,60 +51,21 @@
 ;; invalid-option-placement : (Listof Id) (Listof Syntax/#f) -> Id/#f
```
Metaxal, line 51: This comment is out of sync with the code. Also on line 55.

pkgs/racket-test/tests/stxparse/test.rkt
```diff
@@ -1059,3 +1059,9 @@
 (convert-compile-time-error
  (syntax-parse #'(1 2 'bar 4 5 'bar 'foo)
    [((~seq (~between x:nat 2 2) ... z) ...+ expr) (void)]))
+
+;; from Laurent Orseau, issue #3603 (1/2021)
+(require syntax/parse/lib/function-header)
+(test-case "skip keywords in invalid-option-placement"
+  (syntax-parse #'(#:a [a 1] #:b b)
+    [a:formals #'(a a.params)]))
```
Metaxal, line 1067: It would be good to add a few other cases, just to make sure, such as mixing keywords and positional arguments like `#'(a #:b [b 1] c #:d d e . rest)`. (Yes, I know it's horrible, but it's still allowed by the grammar of `lambda` and should be checked for backward compatibility.)

Metaxal (commented):
Thank you for this PR! Some comments below.

racket/collects/syntax/parse/lib/function-header.rkt
```diff
@@ -49,63 +48,25 @@
...
       name))
   ;; Skip through mandatory args until first optional found, then search
   ;; for another mandatory.
-  (let loop ([names names] [defaults defaults])
+  (let loop ([kws kws] [names names] [defaults defaults])
     (cond [(or (null? names) (null? defaults))
            #f]
-          [(eq? (car defaults) #f) ;; mandatory
-           (loop (cdr names) (cdr defaults))]
+          [(or (car kws) ;; keyword
```
Metaxal, line 68: This is probably not correct, is it?
If it is, can you add a test case that would break without it?

Metaxal: Oh I see.
Here are 2 test cases:
This one should pass: `(a #:b [b 1] c)`
This one should fail with `"default value missing"`: `(a [b 1] c)`

Metaxal:
Wonderful, that looks good to me now. Thank you!

Metaxal:
While you're at it, and *if* you are willing, you could avoid the unnecessary code duplication in `formals` by defining a new splicing-syntax-class, say `splicing-formals-no-rest` that matches simply `(~seq arg:formal ...)`and doesn't care about `rest-id` but does almost all the error checking, and then `formals` would be simplified to something like
```racket
(define-syntax-class formals
  (pattern (~or* (args:formals-no-rest)
                 (args:formals-no-rest . rest-id:id))
           #:attr params #'((~@ . args.names) (~? rest-id))
           #:fail-when (and (attribute rest-id)
                            (member #'rest-id (syntax->list #'args.names))
                            #'rest-id)
                       "duplicate argument identifier"))
```

Unless of course people think it's better in a separate PR, or someone opposes to that.

xxyzz:
> defining a new splicing-syntax-class, say `splicing-formals-no-rest` that matches simply `(~seq arg:formal ...)`
 
But that will get an error: `pattern: head pattern not allowed here at: (~seq arg:formal ...) in: (pattern ...argument name" #:fail-when (check-duplicates (attribute arg.kw) (lambda (x y) (and x y (equal? ...`

Metaxal:
It's because you need to use [`define-syntax-splicing-class`](https://docs.racket-lang.org/syntax/stxparse-specifying.html?q=define-syntax-splicing-class#%28form._%28%28lib._syntax%2Fparse..rkt%29._define-splicing-syntax-class%29%29) instead of `define-syntax-class`.

xxyzz:
Thanks! One more question, I don't understand why there is a `#'rest-id` in `and`?

Metaxal:
This way if there is a duplicate identifier, it raises an exception that will know *where* the fault is precisely. DrRacket will highlight the corresponding `rest-id` identifier in red (do try it!).

Ideally you would add a test that does test the location of the error message, but I haven't tried that myself and it looks a little cumbersome.

Metaxal:
It looks cleaner indeed, thanks! :)

jackfirth:
Is this ready to merge?

xxyzz:
Could someone review this please?

pkgs/racket-doc/syntax/scribblings/parse/lib.scrbl
```diff
@@ -274,4 +274,11 @@ Note that the literal-set uses the names @racket[#%plain-lambda] and
                                  (~? fml.default #f))])
 (syntax-parse #'(lambda (#:kw [kw 42]) kw)
   [(_ (fml:formal) body ...+) #'(fml fml.name fml.kw fml.default)])
-]
\ No newline at end of file
+]
+
+@defstxclass[splicing-formals-no-rest #:splicing]{
+ Like formals but without dotted-tail identifier.
```
sorawee, line 280: IIUC, this should be `@racket[formals]`.

sorawee:
It looks good to me now, but @rmculpepper probably should be the final reviewer and merger.

pmatos:
Can we get this merged? If @rmculpepper is not available, maybe @mflatt can speed this up?

rmculpepper:
Sorry, I keep forgetting to look at this.

I'd prefer to drop the `splicing-` prefix and just call it `formals-no-rest`.

Please add `#:attributes (params)` to the new version of `formals`. I think it's probably a good idea to require all "library" syntax classes to have an explicit `#:attributes` clause.

Is the dependency on `racket/dict` still needed, or can it be dropped now?

pmatos:
re-pinging @rmculpepper  post-commits by @xxyzz

rmculpepper:
Merged, thanks!

PR #3622, opened 2021-01-12
doc: fix usage of @racket in shared

sorawee:


mflatt:
This was merged - not sure why GitHub lost track.

PR #3623, opened 2021-01-12
net/ftp: implement ftp-current-directory

dzoep:
The current implementation of net/ftp lacks support for PWD command - RFC 959, section 4.1.3, response described in section 4.2.2/4.2.3, expected response as per section 5.4.

This command returns the current directory as first term on the response line right after the response code and space character. The directory name is guaranteed to be delimited by double quotes immediately before and after the name.

### DIFF
--- a/pkgs/net-lib/net/ftp.rkt
+++ b/pkgs/net-lib/net/ftp.rkt
@@ -1,6 +1,6 @@
 #lang racket/base
 
-(require racket/date racket/file racket/port racket/tcp racket/list)
+(require racket/date racket/file racket/port racket/tcp racket/list racket/string)
 
 (provide ftp-connection?
          ftp-cd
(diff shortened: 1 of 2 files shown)


samth:
@racket/release @dzoep would like this to appear in the release, so we should discuss it.

mfelleisen:
The changes make sense relative to the commit msgs. 

Questions: 
 — did net/ftp come with tests? 
 — where are tests for the new code?

dzoep:
That is a good point - the tests are in net-tests/tests/net/ftp.rkt
I added the test for the PWD command, reasonable server output, and expected user output. All tests including the new one pass (now it is 116 of them):

racket/pkgs/net-test/tests/net$ ../../../../racket/bin/racket ftp.rkt 
116 tests passed

mfelleisen:
Thanks. I’ll mention it to Sam.

samth:
@dzoep can you rebase and squash this?

samth:
Closing in favor of #3625.

PR #3625, opened 2021-01-13
net/ftp: ftp-current-directory

dzoep:
* implement ftp-current-directory
* add test for ftp-current-directory
* 
The current implementation of net/ftp lacks support for PWD command - RFC 959, section 4.1.3, response described in section 4.2.2/4.2.3, expected response as per section 5.4.

This command returns the current directory as first term on the response line right after the response code and space character. The directory name is guaranteed to be delimited by double quotes immediately before and after the name.


samth:
@jbclements I think this is ready to cherry-pick onto the release branch.

PR #3627, opened 2021-01-17
add make-struct-type-property/generic and make-generic-struct-type-property

### LINKED ISSUE #1647: generics: expose generic-property and generic-method-table?
Should we expose and document `generic-property` and `generic-method-table`, two of the internal macros in `racket/private/generic-methods.rkt`?

An example usage would be
```racket
(struct name [field ...]
  #:property (generic-property gen:interface)
  (generic-method-table gen:interface
    method-def ...))
```

These macros are currently being used by both Rosette and Typed Rosette for different reasons. Rosette needs them for lifting structs to work as symbolic values, and Typed Rosette needs them to expand to after typechecking generic method implementations.

This was first asked in `generics: optional scope argument for private macros` (https://github.com/racket/racket/pull/1644).

AlexKnauth:
Closes #1647, adds `make-struct-type-property/generic` and `make-generic-struct-type-property`, which allow struct-type-properties to inherit from generic-interfaces, in a way suggested by @LiberalArtist to avoid needing to expose the internal method-table.

```racket
(make-struct-type-property/generic
  name-expr
  maybe-guard-expr
  maybe-supers-expr
  maybe-can-imporsonate?-expr
  option
  ...)
maybe-guard-expr =
                 | guard-expr
maybe-supers-expr =
                  | supers-expr
maybe-can-impersonate-expr =
                           | can-impersonate-expr?
option = #:property prop-expr val-expr
       | #:methods gen-interface-id [method-def ...]
```
Produces 3 values just like `make-struct-type-property`.

```racket
(make-generic-struct-type-property
  gen-interface
  method-def
  ...)
```
Produces just the property value.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/generic.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/generic.scrbl
@@ -299,4 +299,56 @@ a @tech{chaperone} of @racket[val-expr] otherwise.
 
 }
 
+@defform[(make-struct-type-property/generic
+           name-expr
+           maybe-guard-expr
+           maybe-supers-expr
+           maybe-can-impersonate?-expr
+           property-option
+           ...)
+  #:grammar
+  ([maybe-guard-expr (code:line) guard-expr]
+   [maybe-supers-expr (code:line) supers-expr]
+   [maybe-can-impersonate?-expr (code:line) can-impersonate?-expr]
+   [property-option (code:line #:property prop-expr val-expr)
+                    (code:line #:methods gen:name-id method-defs)]
+   [method-defs (definition ...)])
+  #:contracts
+  ([name-expr symbol?]
+   [guard-expr (or/c procedure? #f 'can-impersonate)]
+   [supers-expr (listof (cons/c struct-type-property? (-> any/c any/c)))]
+   [can-impersonate?-expr any/c]
+   [prop-expr struct-type-property?]
+   [val-expr any/c])]{
+Creates a new structure type property and returns three
+values, just like @racket[make-struct-type-property] would:
+
+@itemize[
+ @item{a @tech{structure type property descriptor}}
+ @item{a @tech{property predicate} procedure}
+ @item{a @tech{property accessor} procedure}
+]
+
+Any struct that implements this property will also implement
+the properties and @tech{generic interfaces} given in the
+@racket[#:property] and @racket[#:methods] declarations.
+The property @racket[val-expr]s and @racket[method-def]s are
+evaluated eagerly when the property is created, not when
+it is attached to a structure type.
+}
+
+@defform[(make-generic-struct-type-property
+            gen:name-id
+            method-def
+            ...)]{
+Creates a new structure type property and returns the
+@tech{structure type property descriptor}.
+
+Any struct that implements this property will also implement
+the @tech{generic interface} given by @racket[gen:name-id]
+with the given @racket[method-def]s. The @racket[method-def]s
+are evaluated eagerly when the property is created, not when
+it is attached to a structure type.
+}
+
 @close-eval[evaluator]
--- a/pkgs/racket-test/tests/generic/struct-type-property.rkt
+++ b/pkgs/racket-test/tests/generic/struct-type-property.rkt
@@ -1,0 +1,38 @@
+#lang racket/base
+
+(require racket/generic racket/match racket/dict rackunit)
+
+(test-case "silly-dict"
+  (define prop:silly-dict
+    (make-generic-struct-type-property
+     gen:dict
+     (define (dict-ref self k [default #f])
+       'i-am-a-dict)))
+  (struct example ()
+    #:transparent
+    #:property prop:silly-dict #f)
+  (check-equal? (dict-ref (example) 42) 'i-am-a-dict))
+
+(test-case "container-repr"
+  (define-generics repr (->sexpr repr)
+    #:defaults
+    ([number? (define (->sexpr n) n)]
+     [symbol? (define (->sexpr s) `',s)]))
+  (define-values [prop:container container? container-ref]
+    (make-struct-type-property/generic
+     'container
+     #:methods gen:repr
+     [(define/generic gen->sexpr ->sexpr)
+      (define (->sexpr self)
+        (match ((container-ref self) self)
+          [(cons constructor contents)
+           (cons constructor (map gen->sexpr contents))]))]))
+  (struct foo (a b c)
+    #:property prop:container
+    (lambda (self)
+      (list 'foo (foo-a self) (foo-b self) (foo-c self))))
+  (check-equal? (->sexpr (foo 1 2 3)) '(foo 1 2 3))
+  (check-equal? (->sexpr (foo 'a 'b 'c)) '(foo 'a 'b 'c))
+  (check-equal? (->sexpr (foo 'a 1 (foo 'b 2 'empty)))
+                '(foo 'a 1 (foo 'b 2 'empty))))
+
--- a/racket/collects/racket/generic.rkt
+++ b/racket/collects/racket/generic.rkt
@@ -3,6 +3,7 @@
          racket/contract/combinator
          "private/generic.rkt"
          "private/generic-methods.rkt"
+         "private/struct-type-property.rkt"
          (for-syntax racket/base racket/syntax syntax/stx))
 
 ;; Convenience layer on top of racket/private/generic.
@@ -19,7 +20,9 @@
          chaperone-generics
          impersonate-generics
          redirect-generics
-         generic-instance/c)
+         generic-instance/c
+         make-struct-type-property/generic
+         make-generic-struct-type-property)
 
 (begin-for-syntax
 
--- a/racket/collects/racket/private/struct-type-property.rkt
+++ b/racket/collects/racket/private/struct-type-property.rkt
@@ -1,0 +1,131 @@
+#lang racket/base
+
+(provide make-struct-type-property/generic
+         make-generic-struct-type-property)
+
+(require racket/function
+         "generic-methods.rkt"
+         (for-syntax racket/base
+                     syntax/stx))
+
+(begin-for-syntax
+  ;; stx-expr? : Stx -> Bool
+  (define (stx-expr? v)
+    (not (keyword? (if (syntax? v) (syntax-e v) v))))
+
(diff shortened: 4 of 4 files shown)


willghatch:
This looks like it will solve all of my issues with generics.  (Like various others, I have projects that want to dynamically create structs with `make-struct-type` that implement generic interfaces.)

rocketnia:
This looks great. My abstract concern about API lock-in seems to be covered.

To be more specific, this makes it possible to create generic interfaces and be safe in the knowledge that people can abstract over them with a better interface design later on, using `#:derive-property`. Struct type properties were safe from lock-in in this way, being able to abstract over each other using their `supers` feature. Generic interfaces could abstract over structure type properties, but not over each other.

Now that these new utilities make it possible for a generic interface to be used where a structure type property is expected, they'll be just as possible to abstract over.

pkgs/racket-doc/scribblings/reference/generic.scrbl
```diff
@@ -288,4 +288,51 @@ a @tech{chaperone} of @racket[val-expr] otherwise.
...
+   [method-defs (definition ...)])
+  #:contracts
+  ([name-expr symbol?]
+   [guard-expr (or/c procedure? #f 'can-impersonate)]
+   [supers-expr (listof (cons/c struct-type-property? (-> any/c any/c)))]
+   [can-impersonate?-expr any/c]
+   [prop-expr struct-type-property?]
+   [val-expr any/c])]{
+Creates a new structure type property and returns three
+values just like @racket[make-struct-type-property] would:
```
LiberalArtist, line 313: I think there should be a comma between `values` and `just`.

LiberalArtist:
Thanks @AlexKnauth for moving this forward!

Re your [comment](https://github.com/racket/racket/issues/1647#issuecomment-755041110) on #1647:
> I like the suggestion for make-generic-struct-type-property by @LiberalArtist. I would be inclined to choose that over @rocketnia's version because @LiberalArtist's version only evaluates the method definitions once on property-creation, while @rocketnia's version evaluates the method definitions each time the property is attached to a struct-type.

That is the reason I'd written my example the way I did, but as I've continued to think about @rocketnia's version, I can see a case for making a binding for the given property value in scope during the evaluation of the `method-def`s, perhaps as an additional option.

I don't think performance considerations need to decide the question: attaching properties to struct types is a relatively rare operation, probably done just once at startup most of the time, and I think the compiler should do enough lifting to avoid unnecessary closure allocation.

But I think there would be slightly more expressiveness, though I'm not convinced it's really useful. Using the accessor procedure to get at the value requires that you have an instance of the struct. Also, the accessor is undefined until the property is created, which is a further limitation if you are evaluating the `method-def`s only once. Many cases could be handled by @rocketnia's [suggestion](https://github.com/racket/racket/issues/1647#issuecomment-488154505) of refactoring “into a separate structure type property which performs that computation in its guard procedure,” but I don't think it would let you use the property value to compute the procedure that goes into the method table (as opposed to inside the body of a method). However, I haven't come up with anything interesting that this would accomplish.

All of that said, I would lean toward **not** adding complexity now: I definitely don't want to delay closing this almost-four-year-old issue!

Two more concrete comments:

1. I think we should document that the the `method-def`s are evaluated during the creation of the property, since it's observable anyway. (This wouldn't preclude later extending `option` with a case like `#:methods gen-interface-id #:given given-val-id [method-def ...]` that would have different rules.)
2. In `define-generics`’s `#:derive-property prop-expr prop-value-expr` clause, “when `prop-value-expr` is executed, each `method-id` is bound to its specific implementation for the structure type.” Should `supers-expr` and/or `#:property prop-expr val-expr` behave the same way? I would lean toward not: if we're evaluating the `method-def`s only once, there's no advantage. (This would be different if we waited until the property is attached to a struct type—another reason to postpone that until someone actually wants it.) Regardless, it might be worth clarifying in the documentation, since this will be on the same page as `define-generics`/`#:derive-property`.

pkgs/racket-doc/scribblings/reference/generic.scrbl
AlexKnauth: I have added a comma here, thanks

AlexKnauth:
(Re: 1) I have documented that the the method-defs are evaluated during the creation of the property, good idea. I've added that to this PR

(Re: 2 “when prop-value-expr is executed, each method-id is bound to its specific implementation for the structure type.”) For generic interfaces, there are names for multiple methods that are specialized. For struct-type properties, there is an anonymous "single value", that's easily produced by `(prop-ref self)`, so I don't think that's worth it.

rocketnia:
> For generic interfaces, there are names for multiple methods that are specialized. For struct-type properties, there is an anonymous "single value", that's easily produced by `(prop-ref self)`, so I don't think that's worth it.

Looks like maybe there's a miscommunication? There are multiple named methods being defined in the `#:methods` block of a `make-struct-type-property/generic`. These specialized method definitions could shadow the dispatching method definitions in the lexical environment of `supers-expr` and/or `#:property prop-expr val-expr`. If they did, that would be consistent with the precedent of `define-generics`.

Personally, I don't know why `define-generics` does that, and it seems unintuitive to me. I wouldn't mind if `make-struct-type-property/generic` was different in this regard.

AlexKnauth:
Yes, but in one case there's a generic-interface extending a property (that makes all of the named methods available to the property's value-expression), and in the other case there's a property extending a generic interface. It makes sense to bind the names of the methods in the first case, as `define-generics` already does. But this PR is for the second case, where it doesn't seem worth it to bind the property-value... to me anyway

rocketnia:
> Yes, but in one case there's a generic-interface extending a property (that makes all of the named methods available to the property's value-expression), and in the other case there's a property extending a generic interface.

Hmm, I started to write that I saw the analogy more like this, and then I realized my mistake:

* In the case of `define-generics`, there's a `#:methods` block defining method specializations and some `#:derive-property prop-expr prop-value-expr` expressions that extend properties.
* In the case of `make-struct-type-property/generics`, there's a `#:methods` block defining method specializations and some `supers-expr` and `#:property prop-expr val-expr` expressions that extend properties.

My mistake is that there's no `#:methods` block in `define-generics`. It's more like there's a signature of a `#:methods` block. So the methods aren't right there at the call site, and being able to access them actually has a meaningful impact on the programmer's ability to express performant property implementations. I think I get the point of it now.

And you're right. The information that's unknown at a `define-generics` call site is what the method specializations are. In this case, the information that's unknown at a `make-struct-type-property/derived` call site is what the property value is. That does make these analogous in an interesting way -- or at least, a way that would be interesting if we cared what the value of the property was or if we couldn't just use `prop-ref` to get that information. :) I see now, thank you.

I guess what @LiberalArtist was saying was that if someday the method specializations could depend on the property value, then  the information that's unknown at a `make-struct-type-property/derived` call site is what the property value is *and* what the method specializations are. At that point the `supers-expr` and `#:property prop-expr val-expr` expressions might benefit from having access to the method specializations.

Hmm, that could be worth considering someday, if that kind of dependency on the property value is ever allowed. It's hard to say access to the method specializations would add much expressiveness. I don't think it's quite reducible to using `(prop-ref self)` and recomputing the method specializations from that result, since `self` isn't generally available during these expressions and the `#:methods` block may have side effects like generativity that inhibit recomputation. If there is any advantage to it, it's probably better served by allowing arbitrary bindings (not just method specializations) to travel from the `#:methods` block to the property specifications. At any rate, I guess that's a bridge that can be crossed if and when we come to it.

LiberalArtist:
> I guess what @LiberalArtist was saying was that if someday the method specializations could depend on the property value, then the information that's unknown at a `make-struct-type-property/derived` call site is what the property value is _and_ what the method specializations are. At that point the `supers-expr` and `#:property prop-expr val-expr` expressions might benefit from having access to the method specializations.
> 
> … At any rate, I guess that's a bridge that can be crossed if and when we come to it.

Yes, that's what I meant and I agree with the conclusion. If someday we do add a mode or create a new set of macros to allow the method specializations to depend on the property value, presumably we'll be doing it because someone has found a reason to actually *want* to do that—I still haven't come up with one. If that hypothetical day ever arrives, I think we'll be in a better position to know what would be useful than we are right now. In the mean time, I don't think we're closing off any future options by providing the functionality now for which we do know real use-cases.

pkgs/racket-doc/scribblings/reference/generic.scrbl
```diff
@@ -288,4 +288,53 @@ a @tech{chaperone} of @racket[val-expr] otherwise.
...
+values, just like @racket[make-struct-type-property] would:
+
+@itemize[
+ @item{a @tech{structure type property descriptor}}
+ @item{a @tech{property predicate} procedure}
+ @item{a @tech{property accessor} procedure}
+]
+
+Any struct that implements this property will also implement
+the properties and generic-interfaces given in the
```
LiberalArtist, line 322: I would write `@tech{generic interfaces}` rather than `generic-interfaces`.

pkgs/racket-doc/scribblings/reference/generic.scrbl
```diff
@@ -288,4 +288,53 @@ a @tech{chaperone} of @racket[val-expr] otherwise.
...
+            ...)]{
+Creates a new structure type property and returns the
+@tech{structure type property descriptor}.
+
+Any struct that implements this property will also implement
+the generic interface given by @racket[gen:name-id] with the
+given @racket[method-def]s. The @racket[method-def]s are
+evaluated eagerly when the property is created, not when
+attached to a structure.
+}
```
LiberalArtist, line 338: I would either:

1. Make the same edit I suggested above; or
2. Rewrite it referring to `make-struct-type-property/generic`—maybe something like this:
```
Like @racket[make-struct-type-property/generic], but simplified
to implement a single @tech{generic interface} @racket[gen:name-id]
and to return only a single value, the @tech{structure type property descriptor}
for the new property.
```

LiberalArtist (commented):
Aside from these minor potential documentation suggestions, LGTM.

mflatt:
Is this still in progress? I would be good to wrap it up well in advance of the next release.

AlexKnauth:
Oh, sorry, I can do the improvements suggested by @LiberalArtist soon

AlexKnauth:
@LiberalArtist does this look good now?

AlexKnauth:
@mflatt Should I merge, or wait?

LiberalArtist:
LGTM! Thanks for staying on top of this, @AlexKnauth.

PR #3629, opened 2021-01-17
ci: cross-compile to iOS for PR

Bogdanp:
This ensures that cross-compilation to iOS for Racket CS works on PR. To test that things work at runtime, we could set up a free account at one of the iOS CI services and have it run daily builds. I'll look into that next weekend.

The PR workflows seemed like a convenient place to put this, but I'm happy to move it elsewhere if there's a better place (cc @pmatos).

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -66,3 +66,35 @@ jobs:
         raco test -l tests/db/all-tests
         raco test -c tests/stxparse
         raco test -c tests/syntax
+    - name: Tarball
+      run: tar -cvjf racketcs-macos-x64_git${{ github.sha }}.tar.bz2 racket
+    - uses: actions/upload-artifact@v2
+      with:
+        name: racketcs-macos-x64_git${{ github.sha }}
+        path: racketcs-macos-x64_git${{ github.sha }}.tar.bz2
+
+  build-ios:
+    runs-on: macos-latest
+    needs: buildtest-macos
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/download-artifact@v2
+      with:
+        name: racketcs-macos-x64_git${{ github.sha }}
+        path: ${{ github.workspace }}
+    - name: Untar host Racket
+      run: |
+        mkdir host-racket
+        tar -xvjf racketcs-macos-x64_git${{ github.sha }}.tar.bz2 -C host-racket --strip-components 1
+    - name: Build iOS Racket
+      run: |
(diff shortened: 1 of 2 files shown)


racket/src/ChezScheme/c/alloc.c
```diff
@@ -284,9 +284,6 @@ void S_close_off_thread_local_segment(ptr tc, ISPC s, IGEN g) {
   close_off_segment(tgc, tgc->next_loc[g][s], tgc->base_loc[g][s], tgc->sweep_loc[g][s], s, g);
 
   tgc->base_loc[g][s] = (ptr)0;
-#if defined(WRITE_XOR_EXECUTE_CODE)
```
pmatos, line 287: Shouldn't this be part of another PR? I would like to isolate CI patches, instead of having them together with changes that potentially break code.

.github/workflows/ci-pr.yml
```diff
@@ -66,3 +66,35 @@ jobs:
...
+        tar -xvjf racketcs-macos-x64_git${{ github.sha }}.tar.bz2 -C host-racket --strip-components 1
+    - name: Build iOS Racket
+      run: |
+        set -euo pipefail
+        mkdir -p racket/src/build-ios
+        pushd racket/src/build-ios
+        ../configure \
+          --host=aarch64-apple-darwin \
+          --enable-ios=iPhoneOS \
+          --enable-racket=${{ github.workspace}}/host-racket/bin/racket \
```
pmatos, line 97: space after `github.workspace` for consistency.

.github/workflows/ci-pr.yml
```diff
@@ -66,3 +66,35 @@ jobs:
...
+          --enable-scheme=${{ github.workspace}}/host-racket/src/build/cs/c
```
pmatos, line 98: same - space after `github.workspace` for consistency.

pmatos (changes requested):
I don't have a way to test it, but overall looks good to me.
Also - should this also be part of the push setup, instead of just PR?

Bogdanp:
Thanks for the review! I've fixed the spacing and added a job for push as well. I'll test it on my fork (I don't think there's a way to test it here w/o merging, is there?) and then circle back here when it passes.

Bogdanp:
The `push` workflow [passes](https://github.com/Bogdanp/racket/runs/1721529887?check_suite_focus=true) so I've squashed the commits.

pmatos (approved):
lgtm

PR #3647, opened 2021-01-23
build: fix `make install` for cross builds

Bogdanp:
W/o this change, some targets in the "parent" Makefile (like `copytree-run`) may fail because they try to run the target Racket executable instead of the host.


mflatt:
Thanks!

PR #3654, opened 2021-01-25
CS: Reimplement Lattice of cptypes

gus-massa:
(@mflatt already saw some small parts to help fixing bugs, but I repeat them for everyone. It's possible that this version has new bugs never seen before.)
 
I'm trying to clean up the type lattice in cptypes and the implementation of `pred-{imples?,imples-not?,union,intersect}`. For example `pred-union` calls like 20 times `pred-implies?`, and I want to reduce it to only one pass.

While reimplementing them, I modified these functions slightly, to make them more set friendly. The idea is to assign a set to each "predicate":

    'bottom -> empty set, i.e. errors
    <something> -> some other set
    'ptr -> all single values expressions
    #f -> a result that may be single or multiple valued.

so

    'bottom ⊂ <something> ⊂ 'ptr ⊂ #f

In the current implementation `'bottom `and `#f` have cases that don't follow this rule, so I changed them. These produces some changes, and an implicit effect it is easier for `'bottom` to bubble out of an expression 

There should not be too many changes, but for example

     (box? (make-vector (+ 3 'banana))

was reduced to

     (begin (make-vector (+ 3 'banana)) #f)     ; because it's a vector anyway

and now it is reduced to

     (begin (make-vector (+ 3 'banana)) #t)     ; because the set's of possible results is empty, so all of them are boxes

and perhaps in a second pass it has more reductions.

In particular, it was necessary to fix the signature of `generate-temporaries`. Previously it was producing a fake error that was ignored, but now this fake error is not ignored. This fix in in the first commit, it was submitted upstrem but I guess it's better to merge it now here.

In the second commit, I moved the functions to it's own file. Because they are quite easy to isolate, and also I'd like to test them independently in the future. (And hopefully, one day autogenerate most of this.)

The rest are the main part of the implementation. It's easier for debuging to keep them separate, but they have too much garbage, so my idea is to squash everything before merging.

The last but one is to use the new features of `rtd-ancestors `. I was thinking about making an new primitive for `rtd-ancestors` but I think it's better to keep it secret in case we later change the implementation to one that use `O(N log(N))` memory instead of `O(N^2)`. It has a strange part for cross-compiling. I hope it is correct.

The last one is a tiny change in Mf-base. I don't undestand all the details, but I think that `fxmap` is in the wrong line. (I probably put `fxmap` in the wrong line a long time ago.)

As I said, my idea is to squash all of this in a single commit, because the intermediate parts don't make sense independently.

### DIFF
--- a/racket/src/ChezScheme/mats/cptypes.ms
+++ b/racket/src/ChezScheme/mats/cptypes.ms
@@ -558,11 +558,14 @@
 
 (mat cptypes-type-implies?
   (test-chain '((lambda (x) (eq? x 0)) fixnum? #;exact-integer? real? number?))
+  (test-chain* '((lambda (x) (or (eq? x 0) (eq? x 10))) fixnum? #;exact-integer? real? number?))
   (test-chain* '(fixnum? integer? real?))
   (test-chain* '(fixnum? exact? number?)) ; exact? may raise an error
   (test-chain* '(bignum? exact? number?)) ; exact? may raise an error
   (test-chain* '((lambda (x) (eq? x (expt 256 100))) bignum? real? number?))
   (test-chain '((lambda (x) (eqv? 0.0 x)) flonum? real? number?))
+  (test-chain* '((lambda (x) (or (eqv? x 0.0) (eqv? x 3.14))) flonum? real? number?))
+  (test-chain* '((lambda (x) (or (eq? x 0) (eqv? x 3.14))) real? number?))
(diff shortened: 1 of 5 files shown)


mflatt:
LGTM

gus-massa:
Merged in https://github.com/racket/racket/commit/0ff8d18be50b44c7595c909706068a241020e372

PR #3670, opened 2021-01-31
Added dictionary contracts

wilbowma:
Added advanced dictionary contracts for recognizing dictionaries with
contracted keys and values. In particular:
+ homogeneous-dictof, where all keys and values satisfy contracts uniformly
+ heterogeneous-dictof, where keys and values satisfy matching pairs of
  contracts, and keys can be optional.
+ dict-mutability/c, a convenience for recognizing mutable dicts based on an
  input
+ dict-immutability/c, a convenience for recognizing immutable dicts based on an
  input

Design informed by discussions with Jonathan Chan, Philip McGrath, Ben Greenman, Robby
Findler, Jack Firth, and George Neuner.
https://groups.google.com/g/racket-users/c/EuIcgVj2ILk/m/wNarkW7xBwAJ

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/dicts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/dicts.scrbl
@@ -188,7 +188,7 @@ values all satisfy the @racket[value-contract].
 @defproc[(heterogeneous-dictof [#:mandatory-keys? mandatory-keys? boolean? #t]
                                [#:exact-keys? exact-keys? boolean? #f]
                                [#:immutable? immutable? boolean? #t]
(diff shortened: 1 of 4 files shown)


sorawee:
IIUC #3196 is the same as `homogeneous-dictof`, right?

wilbowma:
Looks similar, although it looks like that PR has done/is doing the work to deal with non-flat contracts.

--

wilbowma:
Spoke with Robby a bit about how this contract fails for mutable dicts and I need to use impersonators. Will take a look when I have more time.

PR #3671, opened 2021-01-31
ChezScheme: use types to eliminate some dirty stores

yjqww6:
The current implementation could only eliminate dirty stores for constant immediate objects and fixnum operations. This adds an `$immediate` primitive which could be introduced in cptypes, which implies the rhs of an assignment is an immediate object so that there is no need to touch the remember set.

A benchmark:
```racket
#lang racket
  
(collect-garbage)

(time (let ([v (make-vector 1000000)])
        (for* ([_ (in-range 100)]
               [i (in-range 1000000)])
          (vector-set! v i 0))))

(collect-garbage)

(time (let ([v (make-vector 1000000)])
        (for* ([_ (in-range 100)]
               [i (in-range 1000000)])
          (vector-set! v i (eq? i 100)))))

(collect-garbage)

(time (let ([v (make-vector 1000000)])
        (for* ([_ (in-range 100)]
               [i (in-range 1000000)])
          (vector-set! v i i))))
```

master
```
cpu time: 257 real time: 270 gc time: 6
cpu time: 436 real time: 450 gc time: 5
cpu time: 296 real time: 309 gc time: 5
```

patched
```
cpu time: 259 real time: 273 gc time: 5
cpu time: 324 real time: 338 gc time: 5
cpu time: 259 real time: 272 gc time: 5
```


mflatt:
This looks good to me. I did not run exhaustive tests, but a build works.

@gus-massa - any opinion on the "cptypes.ss" changes?

Interestingly, I'm seeing a 50x(!) improvement on the last case on ARM Mac OS. That may be from avoiding the write fence, but it doesn't seem like the difference should be that big.

I think it would be better to change the definition of `$immediate` in "prims.ss" to something like
```
(define-who $immediate
  (lambda (x)
    (if ($immediate? x)
         x
         ($oops who "~s is not an immediate value" x))))
```
Even though `$immediate` is meant only for unsafe use, this works as a kind of documentation for what it's supposed to be, and it may support some kind of debugging checks in the long run.

yjqww6:
I ran that program in M1 Mac with one more case:
```racket
(collect-garbage)

(time (let ([v (make-vector 1000000)])
        (for* ([_ (in-range 100)]
               [i (in-range 1000000)])
          (vector-set! v i v))))
```
```
cpu time: 99 real time: 110 gc time: 1
cpu time: 263 real time: 274 gc time: 1
cpu time: 4315 real time: 4324 gc time: 1
cpu time: 266 real time: 277 gc time: 2
```
vs
```
cpu time: 99 real time: 110 gc time: 1
cpu time: 131 real time: 141 gc time: 1
cpu time: 101 real time: 112 gc time: 1
cpu time: 268 real time: 279 gc time: 2
```

The result is indeed strange.

mflatt:
Removing either the fence or the dynamic fixnum check (i.e., always add to remember) makes the surprising slowdown go away. My best guess is that tightly alternating writes and fences hit some pessimistic case at the processor level, but that's really a guess.

If the guess happens to be right, it may be worth duplicating the assignment under the fixnum test when a fence is needed, so a write fence and remember operation always happen together.

gus-massa:
Two minor changes:

* https://github.com/racket/racket/pull/3671/commits/7c37aaaa2f56c51d418bba9f42655e8687f829c0#diff-0ca8f3da40b57592cc70236bbad2d43d378152832939f8d0040ce23d45708ac1R701 it should be `(not (number? x))` instead of `(not (fixnum? x))`, because the constant can be for example a flonum, of a number that is a fixnum in one platform and a bignum in other.

* https://github.com/racket/racket/pull/3671/commits/7c37aaaa2f56c51d418bba9f42655e8687f829c0#diff-81b8816a8166e5d95d92cee7c10845b088d2462429caac4d7feaba92eeec4587R2167 the signature should not be `mifoldable` because that depends on the platform.  (With the original version I think there would be no problems, but with the new version with a check, cross compiling in a 32bits version code for a 64bit version may cause a problem if you are very unlucky and cptypes add `$immediate` because the argument is a non trivial expression and then in the second pass `cp0` simplifies it to `2^48`. In this case `cp0` will eat the error, but I'm worried that other weird combination may cause a problem.)

yjqww6:
@gus-massa 
For the first case I think `(not (fixnum? x))` is ok because `$immediate?` has already ruled out other number types.

The second seems indeed a problem. Originally I added `mifoldable` to ensure no constants are wrapped by `$immediate`. Now with the check added alternative approaches might be needed (probably folding it manually in cp0 cases).

gus-massa:
* About the first problem: You are right `fixnum?` is the correct function there.

* About the second: Instead of a special case in the long list of conditions in https://github.com/racket/racket/pull/3671/files/1d9ece51e1e73ecd8182f417c7c180e9f1bb0af7..0d2842491a30212eeeb8f3bb6085ce8dee7fb869# it is better to add a `define-inline`. A mix of the new implementation and something similar to https://github.com/racket/racket/blob/0d2842491a30212eeeb8f3bb6085ce8dee7fb869/racket/src/ChezScheme/s/cp0.ss#L2530-L2535

gus-massa:
* In the first item, you are correct, `fixnum?` is the correct function there.

* In the second item, instead of an additional clause in the long list https://github.com/racket/racket/pull/3671/commits/0d2842491a30212eeeb8f3bb6085ce8dee7fb869#diff-2c071acb2b61f6b200dd6d37976dfcbd678143f5c43c2fd3efdd4d0e24b0cc22R5236-R5244 it's better to add a `define-inline`, something like a mix of the new implementation and https://github.com/racket/racket/blob/master/racket/src/ChezScheme/s/cp0.ss#L2530-L2535

yjqww6:
A drawback of the `define-inline` approach is that `value-visit-operand!` would lift bindings so that the resulting code may look like:
`(some-dirty-store-prim target (let (...) ($immediate ...)))`
This is why the original implementation contains a `build-wrapper*` to fixup this case (`fold-primref2` also uses `value-visit-operand!`). 
We could use the `define-inline` approach and bring back `build-wrapper*`, but that would make the logic separated. This is not a problem, but maybe `$immediate` is special enough to be a cp0 case.

gus-massa:
I slightly prefer something like the other version with `build-wrapper*`, so it's more automatic in case `cp0-call` is used in other constructions. (IIUC, now the only difference is in expressions like `(apply $immutable <something>)` but it doesn't look like an expected case.). But whatever version @mflatt likes is fine for me.

mflatt:
I'm happy with the current version, and we can try a different approach if needed in the future. I expect to merge tomorrow (Utah time).

PR #3679, opened 2021-02-08
cs: optimize unsafe struct operations

yjqww6:
This intends to make pattern matching on structs faster. It works by ensuring rtds of impersonators themselves share same rtd, so the checks of `unsafe-struct-ref/set!` could be simplified.

A simple benchmark:
```racket
#lang racket
(struct A (a b c d))
           
(define (f n)
  (cond
    [(= n 0) 1]
    [else
     (define a (f (sub1 n)))
     (A a a a a)]))

(define t (f 13))

(define (g x)
  (let g ([x x])
    (match x
      [(A a b c d)
       (g a) (g b) (g c) (g d)]
      [else (void)])))

(collect-garbage)
(time (g t))
```
on my machine:
master
```
cpu time: 340 real time: 346 gc time: 0
```
here
```
cpu time: 246 real time: 251 gc time: 0
```
bc
```
cpu time: 215 real time: 216 gc time: 0
```

This makes the performance of pattern matching on struct closer to bc, but still slower. Another drawback is that it defeats optimization introduced in #3531. So this might be not good enough currently.
Also, the `define-impersonator-record` might need more syntax checks.

### DIFF
--- a/racket/src/cs/rumble/box.ss
+++ b/racket/src/cs/rumble/box.ss
@@ -6,8 +6,8 @@
 
 ;; ----------------------------------------
 
-(define-record box-chaperone chaperone (ref set))
-(define-record box-impersonator impersonator (ref set))
+(define-impersonator-record box-chaperone chaperone (ref set))
+(define-impersonator-record box-impersonator impersonator (ref set))
 
 (define (box? v)
   (or (#%box? v)
@@ -136,10 +136,10 @@
            [else (loop next val)]))]))]))
 
 (define (set-box-impersonator-hash!)
-  (record-type-hash-procedure (record-type-descriptor box-chaperone)
+  (record-type-hash-procedure (impersonator-descriptor box-chaperone)
                               (lambda (c hash-code)
                                 (hash-code (box (unbox c)))))
-  (record-type-hash-procedure (record-type-descriptor box-impersonator)
+  (record-type-hash-procedure (impersonator-descriptor box-impersonator)
                               (lambda (i hash-code)
                                 (hash-code (box (unbox i))))))
 
--- a/racket/src/cs/rumble/control.ss
+++ b/racket/src/cs/rumble/control.ss
@@ -1564,8 +1564,8 @@
 (define-record-type (continuation-mark-key create-continuation-mark-key authentic-continuation-mark-key?)
   (fields (mutable name))) ; `mutable` ensures that `create-...` allocates
 
-(define-record continuation-mark-key-impersonator impersonator (get set))
-(define-record continuation-mark-key-chaperone chaperone (get set))
+(define-impersonator-record continuation-mark-key-impersonator impersonator (get set))
+(define-impersonator-record continuation-mark-key-chaperone chaperone (get set))
 
 (define make-continuation-mark-key
   (case-lambda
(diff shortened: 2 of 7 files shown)


mflatt:
This is a nice idea, but we may be able to get close to the same machine code with a simpler change.

Right now, the expansion of `impersonator?` checks that the object is tagged (unnecessary for `unsafe-struct-ref`), that the tag is an rtd (also unnecessary), and that the ancestry depth of the impersonator record type is no more than the argument record (could be made unnecessary since the impersonator record type has no supertype). So, with a suitable variant of `record?` — say, `record-instance?` — that requires its first argument to be a record, and with an adjustment to the rtd-ancestry vector, I think we could eliminate three branch instructions. At the rumble layer, we'd just change `(impersonator? x)` to `(#3%record-instance? x impersonator-rtd)`.

This approach may involve three indirections and a comparison, instead of just two indirections and a comparison, but the extra indirection might be a small difference. Also, the associated compiler improvements would apply somewhat more generally. I'll give it a try, and we can see whether it works and how performance compares.

yjqww6:
Actually I have a similar implementation (without assumption on ancestry vector, though) before trying this one:
https://github.com/racket/racket/compare/master...yjqww6:struct-predicate
Though it also benefits some safe operations ,its performance is no good as this one:
```
cpu time: 293 real time: 297 gc time: 0
```
Maybe removing the ancestry depth check helps, I didn't have a try.

mflatt:
I'll perform more careful experiments (including on x86_64), but so far on an M1:
```
original                   cpu time: 245 real time: 248 gc time: 0
$record-is-a?              cpu time: 221 real time: 223 gc time: 0
$record-is-a? + ancestry   cpu time: 214 real time: 216 gc time: 0
this pr                    cpu time: 207 real time: 210 gc time: 0
```
The third variant with a revised ancestry representation (where `$record-is-a?` is `record-instance?`) is https://github.com/racket/racket/compare/master...mflatt:ancestry-redux

yjqww6:
The third on my machine (a 7700HQ):
with one more benchmark on normal accessors
```racket
(define (k x)
  (let g ([x x])
    (when (A? x)
      (g (A-a x))
      (g (A-b x))
      (g (A-c x))
      (g (A-d x)))))
(collect-garbage)
(time (k t))
```

master
```
cpu time: 335 real time: 339 gc time: 0
cpu time: 298 real time: 302 gc time: 0
```
third (the running time is not stable, the following two is most frequent):
```
cpu time: 256 real time: 261 gc time: 0
cpu time: 325 real time: 329 gc time: 0
```
```
cpu time: 242 real time: 246 gc time: 0
cpu time: 410 real time: 415 gc time: 0
```

mflatt:
I expect the slowdown for regular selectors for "third" is due to dropping the direct rtd test and always going through the ancestry vector. It's possible that the right choice is to keep that shortcut, although the shortcut is pessimistic for predicates that fail (common enough when the predicate is not a contract) or when they're applied to subtype instances.

I'm seeing similar results on other machines, where the "third" implementation is close to the performance of this PR, but not quite as good.

Maybe the right choice is something like this PR plus the revised ancestry representation (because I think it has other benefits, pending a choice of shortcut or not). But instead of duplicating `define-record` as here, we could add a new option to `define-record` at the Chez Scheme level to supply the meta-rtd.

mflatt:
I see that `$record-is-a?` is a good idea independent of other changes, because cptypes can substitute `$record-is-a?` for `record?` when it knows that the first argument is a record (as you implemented), and that substitution is possible many times in the compilation of "cpnanopass.ss".

I'll work on a merger of the ancestry change and `$record-is-a?` (I still like `record-instance?` as the name, along with `$sealed-record-instance?`), but I'll leave the implementation of impersonators alone.

I still think this PR is a good direction if the meta-rtd is made an option for `define-record` at the Chez Scheme level.

yjqww6:
I agree that `record-instance?/$record-is-a?`(I think either is ok, the `is-a` I used comes from cpnanopass ) is worthwhile change in general, I remember the size of racket.boot decreased by 40kb when I experimented on it. 
For the idea of this pr, I am sill wondering how to bring back #3531.

mflatt:
Does this direction still look worthwhile, or it is superseded by other changes?

yjqww6:
Some related cp0 optimizations are more difficult with this representation, so this seems not worthwhile unless new features make them straightforward. I will close it for now.

PR #3683, opened 2021-02-12
Clarify the purpose of `flat-contract-predicate`

jackfirth:
The docs currently state that `flat-contract-predicate` is merely a holdover for backwards compatibility, and that flat contracts can be used directly as predicates now. This isn't correct: _some_ flat contracts can be used directly as predicates, but not all. Ordinary racket values that double as contracts such as numbers and symbols can't be used as predicates without using `flat-contract-predicate`. Additionally, this pull request directs users to `coerce-flat-contract` if they're seeking to build contract combinators that explicitly convert ordinary racket values into flat contracts.

Context: racket/scribble#293.


rfindler:
Thanks!

Issue #3678, opened 2021-02-06
Make `display-xml` more technically correct


default-kramer:
The documentation for `display-xml` says "newlines and indentation make the output more readable, though less technically correct when whitespace is significant." For example, the xexpr `'(a (b "c"))` is displayed as

```
<a>
  <b>
    c
  </b>
</a>
```

I was able to make a change such that this will print like

```
<a>
  <b>c</b>
</a>
```

My change is pretty limited in scope. The contents of `<b>` are not indented because **all** of its children are pcdata. This means that elements with mixed content will remain technically incorrect. Despite this limited scope, I think this change is still useful. I use Racket as a faster way to hand-craft test XML files, none of which have featured mixed content so far. Being able to display those files with both indentation and correct whitespace is nice.

Should I create a PR for this change?

jeapostrophe:
I think it's a good idea, although I think it should probably be a parameter and not the default for backwards compatibility. :'(

sorawee:
Should there also be a list of tags parameter that controls whether the content inside specified tags should be indented? I experienced a similar problem in @greghendershott's custom XML operation in his Markdown package (see https://github.com/greghendershott/markdown/issues/75). In that context, the type of tags dictates whether it's safe to indent (see https://developer.mozilla.org/en-US/docs/Web/HTML/Inline_elements)

sorawee:
Hmm. Actually the bug that I experienced was slightly different. In that one, it's about how:

```
<p>
  Hello<label></label>.
</p>
```

and 

```
<p>
  Hello
  <label></label>
  .
</p>
```

are different (the below one has an extra space between "Hello" and "."). So it's not "inside specified tag", but rather, the entire tag.

default-kramer:
@jeapostrophe Good point, I can see now that the current behavior might be exactly what some callers want. For example `'(html (body (p "This is line 1." "This is line 2")))` works great with `display-xml`, even if it is depending on "technically incorrect" behavior.

@sorawee If we want to allow the caller to customize the behavior, I think the most powerful solution is a callback that accepts the current node and returns two values: 1) whether the current node should be indented, and 2) a new callback to be used for the child nodes. I'll try this out and maybe some other strategies.

sorawee:
@default-kramer now that #3686 is merged, can this issue be closed? Or are there still any outstanding issues?

default-kramer:
@sorawee Oops, I forgot about this. Yes, it can be closed.

PR #3686, opened 2021-02-15
adding different indentation to XML

default-kramer:
Working on issue #3678.

I still need to update the documentation and add some more tests, but I want to get your thoughts on this approach first. (This PR should be marked as a draft in Github.)

The 4 indentation styles are:

- `'none` No indentation. Behaves the same as `write-xml`.
- `'classic` Indents every node. Keeps the same behavior that `display-xml` has today.
- `'scan` If any child node is "whitespace-sensitive", then none of the children are indented.
- `'peek` If the first child node is "whitespace-sensitive", then none of the children are indented.

Both `peek` and `scan` work well for XML that represents boring DTOs. More precisely, I mean XML that meets both of these conditions:

1. The only whitespace-sensitive nodes are pcdata nodes.
2. A pcdata node never has a non-pcdata sibling. (No mixed content.)

### HTML Support?
Currently `whitespace-sensitive?` is defined as `pcdata?`. It is possible that allowing the caller to provide their own predicate could help indent HTML better. For example, the caller might have a list of HTML tags to be considered whitespace-sensitive, as @sorawee suggested in #3678. This might work well enough in practice.

However, I'm not sure if "correct HTML indentation" can be well-defined in the general case. For example, `div` is not commonly whitespace-sensitive, but it's easy to write CSS that makes this untrue. So I'm not sure if it is worth adding this functionality and attempting to explain the caveats, or if we should just say "use `#:indentation 'none` when correctness matters."

### DIFF
--- a/pkgs/racket-doc/xml/xml.scrbl
+++ b/pkgs/racket-doc/xml/xml.scrbl
@@ -244,26 +244,76 @@ Like @racket[syntax:real-xml], but it reads an XML element like
 @defproc[(write-xml [doc document?] [out output-port? (current-output-port)])
          void?]{
 
-Writes a document to the given output port, currently ignoring
-everything except the document's root element.}
+Same as @(racket display-xml) with @(racket #:indentation 'none).}
 
 @defproc[(write-xml/content [content content/c] [out output-port? (current-output-port)])
          void?]{
 
-Writes document content to the given output port.}
+Same as @(racket display-xml/content) with @(racket #:indentation 'none).}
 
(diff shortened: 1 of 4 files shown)


jeapostrophe:
I like this approach, thank you.

default-kramer:
Hi @jeapostrophe , I probably didn't make it obvious that this PR is ready for review and possible merging now. (If you already know but have been busy, no worries, it's not urgent.)

jeapostrophe:
Awesome, thank you! I did miss that previous indication... I'm not sure I got an email about it. Sorry!

PR #3689, opened 2021-02-16
Simplify function definitions

jackfirth:
This is an automated change generated by [`resyntax`](https://github.com/jackfirth/resyntax).

This change simplifies function definitions using shortcuts for `define` forms whose right-hand-side is either a lambda expression or a `case-lambda` expression equivalent to a function with a default argument.


racket/collects/ffi/unsafe/define.rkt
```diff
@@ -8,21 +8,19 @@
...
-                 (apply
-                  string-append
-                  "arguments:"
-                  (let loop ([args args])
-                    (if (null? args)
-                        null
-                        (cons (format " ~e"
-                                      (car args))
-                              (loop (cdr args)))))))))))
+(define (((make-not-available id)) . args)
```
samth, line 11: This might be less clear in the new version -- maybe 0-argument functions should be a special case?

samth (commented):
This broadly looks good to me but it changes a lot of code so I'm not sure who the right person to review it is.

racket/collects/ffi/unsafe/define.rkt
jackfirth: I think that makes sense. Opened https://github.com/jackfirth/resyntax/issues/61 for this.

jackfirth:
I could split it up into separate pull requests, would that be easier?

rfindler:
I looked through all of the contract files and the changes seem good to me. I didn't check that the test suites still pass (although it seems very likely they would). If someone has it installed, running `racket -l tests/racket/contract/all` runs them.

samth:
The contract tests run in the CI so we can see that they pass.

jackfirth:
Closing this pull request for a few reasons:

1. It's too large. I think these kinds of pull requests should probably be in the range of a couple hundred lines instead, at least while Resyntax is still in such an experimental state.
1. I'd like to fix jackfirth/resyntax#61 first.
1. It's been a while and now there's merge conflicts.

PR #3692, opened 2021-02-19
racket/src/{bc,cs/c}/configure fix arg stripping

tgbugs:
rktio doesn't know how to handle the --collectsdir argument and will
stop the build process if it is passed to racket/src/configure, this
fixes the immediate issue but the underlying issue remains and other
unhandled arguments are likely to break rktio configure as well


mflatt:
Thanks for the patch! But I'll adjust the source of the configures scripts and rebuild them. The other argument that could cause problems in the same way is `--appsdir`.

PR #3694, opened 2021-02-21
cs: optimize pattern matching on authentic structs

yjqww6:
This eliminates the branch of `unsafe-struct-ref` on cptypes level. 

A benchmark:
https://gist.github.com/yjqww6/d20134c86a749cf19a1946e8a2aaecbc

Results (on a M1 Mac):
master
```
cpu time: 1458 real time: 1481 gc time: 261
420000
```
this
```
cpu time: 1088 real time: 1107 gc time: 333
420000
```

Alternative approach may be making `match` generate `unsafe-struct*-ref` directly.

### DIFF
--- a/racket/src/ChezScheme/mats/cptypes.ms
+++ b/racket/src/ChezScheme/mats/cptypes.ms
@@ -696,6 +696,18 @@
       (define-record-type ,my-rec (fields a))
       (define-record-type ,my-sub-rec (fields a) (parent ,my-rec) (sealed #t))
       (lambda (x) (and (my-rec? x) (list 'ok (#3%$sealed-record-instance? x (record-type-descriptor ,my-sub-rec)))))))
+
+  ;; obviously incompatible rtds
+  ;; the third pass is needed to eliminate #3%$value
+  (parameterize ([run-cp0 (lambda (cp0 x) (cp0 (cp0 (cp0 x))))])
+    (cptypes-equivalent-expansion?
+     `(let ()
+        (define-record I (a))
+        (define A (make-record-type-descriptor* 'a #f #f #f #f 1 0))
+        (lambda (x) (and ((record-predicate A) x) (I? x))))
+     `(begin
+        (make-record-type-descriptor* 'a #f #f #f #f 1 0)
+        (lambda (x) #f))))
 )
 
 (mat cptypes-unsafe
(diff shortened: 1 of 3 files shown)


samth:
Is authentic-ness available in static structure information? If so, match could generate that pretty easily. Even so, this seems like a good idea.

yjqww6:
> Is authentic-ness available in static structure information? If so, match could generate that pretty easily. Even so, this seems like a good idea.

It doesn't currently. Also, the authentic-ness is not really known in compile time, because it could be determined dynamically:
```
> (let ()
    (define a prop:authentic)
    (struct A (a) #:property a #t)
    (struct-type-authentic? struct:A))
#t
```

samth:
Right, we're not going to have a complete check, but a sound approximation (roughly just using `#: authentic`) ought to be enough to capture the important optimizations.

mflatt (approved):
Looks good to me (didn't actually try running it, yet), so waiting for a review by @gus-massa to check/confirm.

gus-massa:
It looks good, but I never read the details of the tricks to reduce records in `cp0`, so I'm confused by

    (define (rtd-obviously-incompatible? x y)
      (let ([x-flds (rtd-flds x)]
            [y-flds (rtd-flds y)])
        (or (and (fixnum? x-flds) (not (fixnum? y-flds)))
            (and (not (fixnum? x-flds)) (fixnum? y-flds)))))

I tried to create manually a few `rtd`s but I got `vector`s instead of `fixnum`s.

About the three passes for `cp0`, I found a similar case in https://github.com/cisco/ChezScheme/pull/421

yjqww6:
@gus-massa , that function is a simplified version of
https://github.com/racket/racket/blob/63aa90c7487c361ce4786f79ae2d5fca1984da78/racket/src/ChezScheme/s/cp0.ss#L3960-L3971

`rtd`s with anonymous fields (created by `make-record-type-descriptor*`) would have `fixnum` in `flds`. That's the case of racket structs.
`impersonator`s are defined with `define-record`, and have named fields (list of vectors in `flds`).

gus-massa:
LGTM

gus-massa (approved):

pmatos:
Note that this change causes the solo chez workflows to fail.

yjqww6:
> Note that this change causes the solo chez workflows to fail.

The failure should be due to outdated expected error message in https://github.com/racket/racket/commit/0523a5311ce95ab9658a585e59c907433db9fdf9 .

mflatt:
I'll fix the tests (that I broke) in a separate push.

PR #3695, opened 2021-02-23
doc: fix contract for read-bytes-avail!/enable-break

capfredf:


pkgs/racket-doc/scribblings/reference/string-input.scrbl
```diff
@@ -254,7 +254,7 @@ is not reached.}
                                          [in input-port? (current-input-port)]
                                          [start-pos exact-nonnegative-integer? 0]
                                          [end-pos exact-nonnegative-integer? (bytes-length bstr)])
-         (or/c exact-positive-integer? eof-object? procedure?)]{
+         (or/c exact-nonpositive-integer? eof-object? procedure?)]{
```
sorawee, line 257: Do you mean nonnegative?

capfredf: Yes, thank you

mflatt:
Thanks!

PR #3697, opened 2021-02-24
Improve flat-murec-contract with random generation

camoy:
This PR adds support for generation, strength checking, and equivalence checking to `flat-murec-contract`. It does this by changing the expansion of `flat-murec-contract` to use the same contract struct as `flat-rec-contract` (which supports these features) instead of predicates.


rfindler:
Thanks!

PR #3698, opened 2021-02-25
Fail random generation of range contracts with bad bounds

camoy:
Contracts that describe ranges of values (`between/c`, `integer-in`, `char-in`) didn't fail gracefully when generating random instances in the case of improper bounds (where lower is greater than upper). For `integer-in` and `char-in`, it raised an internal error, and for `between/c` it actually generated values that didn't satisfy the contract.


rfindler:
Oops. Kind of wished that there had been an error signaled in the constructor when these were added, back in the day.

rfindler:
Thanks!

PR #3705, opened 2021-02-28
cs: add maybe predicates to cptypes

gus-massa:
Add `maybe-*` predicates to cptypes. Actually, use a more general decomposition with three parts

* an `$immediate` that is not a `fixnum`, 
* a `$record` 
* other types like `string`, `vector`, ...

(Actually in the 132 order.)

This is not as general as an arbitrary union, but it is enough for the common cases, and also to handles the common objects in Racket that are implemented in CS as the corresponding object and a record for the impersonator. For example a `racket:maybe-vector` can be decomposed as the union of `#f`, `<record impersonator>`, `cs:vector`. 

It would be nice to rewrite (later) [rumble/impersonators.ss](https://github.com/racket/racket/blob/master/racket/src/cs/rumble/impersonator.ss) to be more friendly with this decomposition.

The first part includes `char`, it's strange but it makes sense at the low level. It does not include `fixnum`, but it's easier to keep the fixnum with the other kinds of numbers and also the separations is more natural at the low level.

Each part use the same value for bottom: `'bottom`, but a different value for top: `$immediate/nofixnum`, `$record`, `normalptr`. I don't like the name of the last one, so any suggestion is welcome.

When the predicate uses only one of the three parts, then it is not wrapped, so it's necessary to have three different tops. (And having different names makes debugging easier.) If necessary, It's easy to rewrite the code to force the missing parts filled with `'bottom`, but it the makes writing the code in cptypes more tricky because it's not possible to use symbols for the usual types.

I also don't like the name of the wrapper `pred-or` too much, but `pred-union` was confusing, any suggestion is welcome.

I had a prototype before the previous change for authentic structs, so I had to copy the changes manually. It would be nice if @yjqww6 can take a look at that part.

The second commit is an attempt to add a new primitive `$immediate/nofixnum?`. It makes sense because `$immediate?` is implemented in [CS/cpnanopass.ss](https://github.com/racket/racket/blob/master/racket/src/ChezScheme/s/cpnanopass.ss) as an `or` of the "real" immediates and the fixnums. So I think it's better to expand it before and let cptypes simplify it partially if possible. I failed adding the new primitive, so it's commented. 

Ideas for the near future:

* Add special cases for `maybe-char` and `char/eof`.
* Add all the missing `maybe-*` that appear in primdata.ss .
* More support for other predicates, for example `null?`, so we get reductions like
    `(when (list? x) (unless (null? x) (pair? x)))` ;=> 
    `(when (list? x) (unless (null? x) #t))`
 * Rewrite rumble/impersonators.ss ?

### DIFF
--- a/racket/src/ChezScheme/s/cptypes-lattice.ss
+++ b/racket/src/ChezScheme/s/cptypes-lattice.ss
@@ -28,21 +28,28 @@
 ; remember to check (implies? x bottom) before (implies? x something)
 
 (module cptypes-lattice
-  (predicate-implies?
+  (primref-name/nqm->predicate
+   ptr-pred
+   $fixmediate-pred
+   true-pred ; anything that is not #f
+   true-rec  ; only the #t object
+   false-rec
+   void-rec
+   null-rec
+   eof-rec
+   bwp-rec
+   predicate-is-ptr?
+   predicate-implies?
    predicate-disjoint?
    predicate-intersect
    predicate-union
    make-pred-$record/rtd
    make-pred-$record/ref)
 
-  (include "base-lang.ss")
-  (with-output-language (Lsrc Expr)
-    (define true-rec `(quote #t)))
-
-  ; don't use rtd-* as defined in record.ss in case we're building a patch
-  ; file for cross compilation, because the offsets may be incorrect
-  (define rtd-ancestors (csv7:record-field-accessor #!base-rtd 'ancestors))
-  (define rtd-flds (csv7:record-field-accessor #!base-rtd 'flds))
+  (define-record-type pred-or
+    (fields imm nor rec)
+    (nongenerative #{pred-or nlomo7xtc1nguv2umpzwho0dt-0})
+    (sealed #t))
 
   (define-record-type pred-$record/rtd
     (fields rtd)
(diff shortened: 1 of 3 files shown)


racket/src/ChezScheme/s/cptypes-lattice.ss
```diff
@@ -54,6 +55,35 @@
...
+  (define maybe-number-pred (make-pred-or false-rec 'number 'bottom))
+  (define maybe-fixnum-pred (make-pred-or false-rec 'fixnum 'bottom))
+
+  ; This can be implemented with implies?
+  ; but let's use the straightforward test.
+  (define (predicate-is-ptr? x)
+    (and (pred-or? x)
+         (eq? (pred-or-imm x) '$immediate/nofixnum)
+         (eq? (pred-or-imm x) 'normalpred)
+         (eq? (pred-or-imm x) '$record)))
```
yjqww6, line 80: should be `pred-or-nor` and `pred-or-rec` respectively?

yjqww6:
I confirmed previous optimization for authentic structs still works, and it should since chez ci passed.

racket/src/ChezScheme/s/cptypes-lattice.ss
gus-massa: Thanks. Fixed.

gus-massa: Fixed also `'normalpred` -> `'normalptr`.

mflatt:
I think it's unhelpful (and it has confused me sometimes) that "immediate" sometimes refers to `type-immediate` values and sometimes it refers to those values plus fixnums.

Maybe we should make `$immediate?` refer to just `type-immediate`, introduce an `$immediate-or-fixnum?` macro for the few places where `$immediate?` is currently used, rename `$immediate` to `$immediate-or-fixnum`, and also change `IMMEDIATE` as used in the kernel.

If that sounds like a good idea, would it be better or worse if I do that before you merge these changes?

gus-massa:
I think it's better that you rename the functions first.

mflatt:
Since "immediate-or-fixnum" is a lot to say/read/type, how about "fixmediate"?

gus-massa:
I changed the names to use the new meaning of `$immediate?`.

mflatt:
LGTM

PR #3710, opened 2021-03-03
Avoid hard-coded paths for utilities in build scripts

LiberalArtist:
This helps build Racket in environments like Nix and GNU Guix where these tools don't have their usual paths.

Most of these changes apply to Chez Scheme: @samth suggested submitting here first, then to the upstream repository. I've put the Chez changes in a separate commit from the Racket-only changes.


LiberalArtist:
Rebased on master after the [tests had passed](https://github.com/racket/racket/actions?query=branch%3Abuild-command-paths-wip) for 85e6cce.

PR #3711, opened 2021-03-04
raco pkg new: wrap long output line

LiberalArtist:
The line was 84 characters, too long for many terminals.


sorawee:
OK to merge?

PR #3713, opened 2021-03-05
version/check: use https; actually enforce timeout; guard reading params

LiberalArtist:


racket/collects/version/check.rkt
```diff
@@ -93,6 +104,11 @@
 (define (check-version)
   (with-timeout timeout check-version-raw))
 
+(define simulate-timeout-for-interactive-testing? #f)
```
samth, line 107: Might be nice to read this from an environment variable.

LiberalArtist: Yes, I'll do that.

rmculpepper: How about making it a parameter that is only set by the testing submodule?

LiberalArtist: I think it's also an issue for clients of this module: I would not have expected calling `check-version` to potentially block for 30 seconds!

LiberalArtist: I'm thinking now about adding a `check-version-evt` and/or making the timeout an optional argument, but I also don't want to make this code more complex, especially since it doesn't seem to have a lot of eyeballs on it.

samth: That seems like overkill, I think either my idea or @rmculpepper's is fine and you should just pick whichever seems better.

LiberalArtist:
It looks like somehow my comment didn't go through, but I've made the changes and squashed. Opinions about the environment variable name are welcome.

PR #3714, opened 2021-03-05
add unsafe-unreachable

yjqww6:
The motivation is that I found `match-define` is usually slower than plain accessors in unsafe mode because it needs to check the objects. 
`unsafe-unreachable` could somewhat address this problem, for example:
```racket
#lang racket
(require (only-in '#%unsafe unsafe-unreachable))
(struct A (a b) #:authentic)

(define (f x)
  (match x
    [(list (A a b) c d)
     (+ a b c d)]
    [else (unsafe-unreachable)]))
````
The generated scheme code for `f` is optimal. All the predicates are optimized away.
```scheme
(lambda (x_7)
  (let ([unsafe-car8_13 (#3%car x_7)])
    (let ([unsafe-cdr9_14 (#3%cdr x_7)])
      (let ([temp11_15 (#3%$record-ref
                        unsafe-car8_13
                        0)])
        (let ([temp12_16 (#3%$record-ref
                          unsafe-car8_13
                          1)])
          (let ([unsafe-car16_19 (#3%car
                                  unsafe-cdr9_14)])
            (#2%+
             temp11_15
             temp12_16
             unsafe-car16_19
             (#3%car
              (#3%cdr unsafe-cdr9_14)))))))))
```
But I am not sure how to make use of this in `match-define` currently. Though making `match`s default to this when compiling the module in unsafe mode might be straightforward, it seems backward incompatible. Probably unsafe variants of `match`s are necessary.


samth:
Another place this could be useful is in the schemify translation of `let-values`, which includes an error case. That might be enough by itself to simplify some of the expansion of `match-define`, as I mentioned on the mailing list.

racket/src/ChezScheme/s/cptypes.ss
```diff
@@ -1430,10 +1438,14 @@ Notes:
                       (predicate-implies? ret3 'bottom)) ;check bottom first
                  (values ir 'bottom pred-env-bottom #f #f)]
                 [(predicate-implies? ret2 'bottom) ;check bottom first
-                 (values (make-seq ctxt `(if ,e1 ,e2 ,void-rec) e3)
+                 (values (if (unsafe-unreachable? e2)
```
samth, line 1441: I worry a little about whether we need to look deeper to find the `unreachable`, or propagate unreachability outward.

yjqww6:
The overhead of multiple values could be eliminated by the `np-push-mrvs` pass, so the error case from `let-values` should not be a problem.

racket/src/ChezScheme/s/cptypes.ss
yjqww6: IIUC, it seems the only easy case is `(begin (unsafe-unreachable) ...)`

mflatt:
I'm uncertain about what `(unsafe-unreachable)` means. What would the documentation say?

racket/src/ChezScheme/s/cptypes.ss
samth: The tricky part for match is calling a function whose body is `(unsafe-unreachable)`.

mflatt: Would cptypes have already changed `(begin ($unreachable) ....)` in `e2` into `($unreachable)`, since `$unreachable` is marked as aborting?

samth:
Just for reference, here's what the Clang docs say: https://clang.llvm.org/docs/LanguageExtensions.html (search for `builtin_unreachable`).

yjqww6:
I think it might looks like 
___
The compiler could assume that control flow never reaches `(unsafe-unreachable)`. 
The behavior of program is unpredictable if it is anyway reached.

For example, if a branch of `if` starts with `(unsafe-unreachable)`, Racket could 
simply evaluate the `test-expr`, and then the other branch with the assumption
that `test-expr` would be evaluated to the corresponding value.

racket/src/ChezScheme/s/cptypes.ss
yjqww6: Right, it is already simplified to `($unreachable)`.

yjqww6: A function whose body is `(unsafe-unreachable)` should be small enough to inline, so I think it is not a problem.

mflatt:
Thanks for the clarifications.

I'd phrase it this way: the contract on `unsafe-unreachable` is `(case->)`, which is a contract that is never satisfied by any call, and "unsafe" has its usual meaning: if a contract check would fail, anything at all can happen afterward.

racket/src/ChezScheme/s/cptypes.ss
mflatt: Should we improve cp0 so that an application form`(e1 .... <aborting-call>  e2 ...)` turns into <aborting-call> (given that the parts of an application are allowed to be evaluated in any order)? And then that would take care of more `($unreachable)` cases?

mflatt:
I guess `(case->)` isn't the right contract, since `procedure-arity` will claim arity 0. Maybe `(->d () #:pre #f any)`.

mfelleisen:
`(->i () #:pre () #false any)` please

rfindler:
Maybe it is supposed to be a contract that just blows up when you evaluate
it, something like none/c.

gus-massa:
It looks like a good idea, but I'm not sure why it is in `cptypes` instead of `cp0`. The only problem I see if it were in cp0 is that it may be removed too early, for examples in

    (unless (pair? x)
      (unsafe-unreachable))
    (pair? x)

What about a safe version for debuging, like `fx+` and `unsafe-fx+`? Is this a good idea?

How is this compiled in `cpnanopass`? If the unsafe version does not emit any code, it can useful in `schemify`, so be used after `raise-error-something` so Chez Scheme understands it will not return, without increasing the size of the executable.

racket/src/ChezScheme/s/cptypes.ss
gus-massa: I prefer to keep the shallow version in the current proposal, and (later) make `cptypes` more aggressive to cut expressions when it detects an error.

yjqww6:
@gus-massa , yes. If it is implemented in cptypes, `(lambda (x) (if (pair? x) (car x) (unsafe-unreachable)))` could be optimized to `(lambda (x) (unsafe-car x))` rather than `(lambda (x) (car x))`.

`(unsafe-unreachable)` would be compiled to raise a exception if it is not eliminated. `schemify` already recognizes non-return functions and expands them with `$app/no-return`.

I am not sure whether safe variant is really needed, since we could write something like
```racket
(if (variable-reference-from-unsafe? (#%variable-reference))
    (unsafe-unreachable)
    <raise an exception>)
```

yjqww6:
More aggressively, `schemify` could replace functions raise `exn:fail:contract` with `(unsafe-unreachable)` when compiling in unsafe mode, but unfortunately `exn:misc:match` extends `exn:fail` rather than `exn:fail:contract`.

rfindler:
Just don't do it in the test suite. ;P

gus-massa:
I think `$app/no-return` can be eliminated in a future change.  (Not now. It's just an idea to think about.) Just split in something like

    (define-sytax-rule ($app/no-return f x ...))
      (begin (f x...) (#3%$unreachable)))

I like to minimizing the number of primitives

---

I don't propose that the safe version of `unsafe-reachable` is another primitive in Racket, just a normal function that raises an error. When I use `unsafe-fx+`, I like to test it with the safe replacement, and then switch to the unsafe version (and switch back if necessary.). I try to avoid marking the whole module as unsafe, so I prefer the possibility of using a unsafe function in a single spot.

---

IIUC currently only `#3%$unreachable` is magic, but `#2%$unreachable` just raises an error at run time. I like this.
Anyway, my  proposal is that `#3%$unreachable` also disappears in `cpnanopass`, to make the compiled code smaller (and more dangerous).

yjqww6:
I make `(#3%$unreachable)` compiled as `(void)` now, which would become no-op for effect positions.

mflatt:
Is this a good place to use the word "assert" (not already used in Racket), so that the Racket-level functions would be `assert-unreachable` and `unsafe-assert-unreachable`?

I can further imagine an `assert` form where `(assert #f)` turns into `(unsafe-assert-unreachable)` in unsafe mode (by way of `(unless #f (unsafe-assert-unreachable))`. That is, `assert` would be specified as raising a contract exception if its argument produces false. The only reason to consider it right at the moment is to decide whether using the word "assert" here looks like a good idea.

yjqww6:
Any opinions on which module they should be in? `racket/unreachable` and `racket/unsafe/ops` respectively?

samth:
I think changing TR to produce exn:fail:contract (and match too, re @yjqww6's comment earlier) is a good idea.

mflatt:
If `assert-unreachable` is a good name, should it take `raise-arguments-error`-like arguments, so in safe mode it's like `raise-arguments-error`, and in unsafe mode it can turn into `(#3%$unreachable)`? And then `match` is changed to use `assert-unreachable` as a default "else" clause?

What would happen to `exn:misc:match?`? Would it be an alias for `exn:fail:contract?`? Does it seem likely that any existing code relies on a distinction?

samth:
I think there are a few questions here:

1. Should Typed Racket's assert and match's match failure errors be instances of exn:fail:contract?
2. Should assert-unreachable be able to construct informative error messages?
3. Should match use assert-unreachable?

I think the answers for 1 and 2 are definitely yes, but for 3 it depends on the details of the error messages and existing error handling.

yjqww6:
I prefer that `assert-unreachable` has no arguments, because that would be easier to switch between safe and unsafe variants.

I think `match` could use `variable-reference-from-unsafe?`, and raise `exn:misc:match` in safe mode.

mflatt:
I meant to suggest that both `assert-unreachable` and `unsafe-assert-unreachable` have arguments, but the latter is equivalent to just `(begin arg ... (#3%$unreachable))`.

Meanwhile, the intent of having `assert-unreachable` is that forms like `match` wouldn't have to handle unsafe mode explicitly (if it makes sense to treat the default case as an assertion). It would be handled by using `assert-unreachable`.

yjqww6:
It is possible that `arg ...` are too complex to be eliminated. In that case, the optimization would not work.

yjqww6:
I suggest `assert-unreachable` and `unsafe-assert-unreachable` have no argument, and provide a macro like
```racket
(define-syntax-rule (with-assert-unreachable form)
  (if (variable-reference-from-unsafe? (#%variable-reference))
      (unsafe-assert-unreachable)
      form))
```
So someone could write
```racket
(with-assert-unreachable <raise an exception>)
```

mflatt:
I'd be happy with that.

mflatt:
Thanks for the doc updates and other changes!

Unless you object, I'll try making a few changes while merging. The substantitive one would be to make `assert-unreachable` more primitive, so that `(#%declare #:unsafe)` turns it into `unsafe-assert-unreachable`, and I'd try doing that by adding `unsafe-unreachable` as a public binding in Chez Scheme (instead of private-to-the-implementation `$unreachable`).

yjqww6:
I am fine with that.

mflatt:
Merged as b4d05e7a410e78154dc973b74c5b48c3441c7226

samth:
I tried this out for match in https://github.com/samth/racket/commit/9f642e96d1a02f9be5055e8f1aa6d10736025487

I looked at the results for this program:

```
#lang racket

(#%declare #:unsafe)

(struct op (a b) #:authentic)

(define (f x)
  (match-define (cons (op a b) r) x)
  (+ a b (f r)))
```
Just putting `with-assert-unreachable` inside the definition of `match:error` was insufficient to get the optimizations; I had to put it around the calls to `match:errror`. After that, the structure access looked good, but the `call-with-values` was still there after cp0/cptypes.

yjqww6:
`runtime.rkt` itself is compiled in safe mode, so I think putting `with-assert-unreachable` into `match:error` has no effect.

We could inspect the output of later passes by
```
> (require ffi/unsafe/vm)
> (vm-eval '(($primitive $np-tracer) 'np-push-mrvs))
```
I expect `call-with-values` to be eliminated after this pass since the taken clause is statically known.

samth:
I think the relevant output is 

```
    (#{f i3cm2mbiucfn72wfezyrsr9to-47}
      (case-lambda
        [clause
         (#{x_7 i3cm2mbiucfn72wfezyrsr9to-63})
         (#{a_8 i3cm2mbiucfn72wfezyrsr9to-64} #{b_9 i3cm2mbiucfn72wfezyrsr9to-65}
           #{r_10 i3cm2mbiucfn72wfezyrsr9to-66}
           #{args i3cm2mbiucfn72wfezyrsr9to-67}
           #{t i3cm2mbiucfn72wfezyrsr9to-68}
           #{unsafe-car7_14 i3cm2mbiucfn72wfezyrsr9to-69}
           #{unsafe-cdr8_15 i3cm2mbiucfn72wfezyrsr9to-70}
           #{temp10_16 i3cm2mbiucfn72wfezyrsr9to-71})
         #f
         1
         (overflow-check
           (trap-check
             #f
             (mlabel
               (begin
                 (set! #{unsafe-car7_14 i3cm2mbiucfn72wfezyrsr9to-69}
                   (mref #{x_7 i3cm2mbiucfn72wfezyrsr9to-63} zero 7 uptr))
                 (set! #{unsafe-cdr8_15 i3cm2mbiucfn72wfezyrsr9to-70}
                   (mref #{x_7 i3cm2mbiucfn72wfezyrsr9to-63} zero 15 uptr))
                 (set! #{temp10_16 i3cm2mbiucfn72wfezyrsr9to-71}
                   (mref
                     #{unsafe-car7_14 i3cm2mbiucfn72wfezyrsr9to-69}
                     zero
                     9
                     uptr))
                 (set! #{r_10 i3cm2mbiucfn72wfezyrsr9to-66}
                   #{unsafe-cdr8_15 i3cm2mbiucfn72wfezyrsr9to-70})
                 (set! #{b_9 i3cm2mbiucfn72wfezyrsr9to-65}
                   (mref
                     #{unsafe-car7_14 i3cm2mbiucfn72wfezyrsr9to-69}
                     zero
                     17
                     uptr))
                 (set! #{a_8 i3cm2mbiucfn72wfezyrsr9to-64}
                   #{temp10_16 i3cm2mbiucfn72wfezyrsr9to-71})
                 (goto #{mv i3cm2mbiucfn72wfezyrsr9to-72}))
               (#{mv i3cm2mbiucfn72wfezyrsr9to-72}
                 (begin
                   (set! #{t i3cm2mbiucfn72wfezyrsr9to-68}
                     (mvcall
                       #{dcl i3cm2mbiucfn72wfezyrsr9to-73}
                       #f
                       #{r_10 i3cm2mbiucfn72wfezyrsr9to-66}
                       ()))
                   (mvcall #f #<literal +> #{a_8 i3cm2mbiucfn72wfezyrsr9to-64}
                     #{b_9 i3cm2mbiucfn72wfezyrsr9to-65}
                     #{t i3cm2mbiucfn72wfezyrsr9to-68} ()))))))]))
```

which indeed looks like the multiple values are gone.

Would be nice to (a) integrated this into `PLT_LINKLET_SHOW_CP0` somehow and (b) have a more readable printing, but I don't know how much work either of those entails.

samth:
With `(print-gensyms #f)` I get:

```
    (f (case-lambda
         [clause
          (x_7)
          (a_8 b_9 r_10 args t unsafe-car7_14 unsafe-cdr8_15
               temp10_16)
          #f
          1
          (overflow-check
            (trap-check
              #f
              (mlabel
                (begin
                  (set! unsafe-car7_14 (mref x_7 zero 7 uptr))
                  (set! unsafe-cdr8_15 (mref x_7 zero 15 uptr))
                  (set! temp10_16 (mref unsafe-car7_14 zero 9 uptr))
                  (set! r_10 unsafe-cdr8_15)
                  (set! b_9 (mref unsafe-car7_14 zero 17 uptr))
                  (set! a_8 temp10_16)
                  (goto mv))
                (mv (begin
                      (set! t (mvcall dcl #f r_10 ()))
                      (mvcall #f #<literal +> a_8 b_9 t ()))))))]))
```

which is pretty easy to read and now I want to document these forms.

PR #3719, opened 2021-03-07
cptypes: discard call form around an error

mflatt:
For example, `(f (error 'x "x") y x)` turns into `(error 'x "x")`.

@gus-massa @yjqww6 Does this seem worthwhile? Is there a better way to implement it?

I didn't cover all cases in `define-specialize/unrestricted` handler, because they didn't seem especially likely. I experimented briefly with changing cp0 instead of cptypes, but cptypes has more information and already handles other cases.

### DIFF
--- a/racket/src/ChezScheme/mats/cp0.ms
+++ b/racket/src/ChezScheme/mats/cp0.ms
@@ -2614,7 +2614,8 @@
                  (begin (#%write 'f) #\y)
                  (begin (#%write 'g) ($zzz-ok))))
              (begin (#%write 'h) 1)))))
-  (expansion-matches?
+  (parameterize ([enable-type-recovery #f])
+   (expansion-matches?
         '(begin (write 'a)
            ((begin (write 'b) string-ref)
             (begin (write 'c)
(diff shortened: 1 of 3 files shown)


yjqww6:
It seems possible to put the logic into `map-Expr/delayed` and `Expr/call` and use continuation/cps to escape instead of duplicating it many times, although there may be performance problems.

mflatt:
Thanks for the comments. I'll end up merging a variant of this, because it's part of getting "optimize.rktl" going. We can replace the implementation with a better one later, of course.

The changes to support error unwrapping ended up getting tangled with changes to better handle continuation marks. I could split those apart, but I probably won't bother.

mflatt:
Merged as 75ba7ff5bc5d05d77f891699ff98ebb71f8085b6, keeping discussion open here for a little while.

gus-massa:
LGTM. Sorry for the delay. 

`Expr/call` is weird because the results are not the results of the expressions an in the other `Expr/???`, so it needs the additional result. A time ago, I wondered about returning as the type a special record `(pred-procedure r t tt ft)` that includes all the information in case the function is call, but it has too many `fxmap`s that are tricky to use correctly if they are moved around too much. I hope to fix this someday, but I'll probably try first some easier cases of subcases of `procedures`.  (Also, it would be nice to add something like the `app` contexts in `cp0`.)

PR #3722, opened 2021-03-09
Add maybe-char to cptypes

gus-massa:
chars are immediates, so the previous change doesn't add automatically the combinations like `maybe-char`. Add also `eof/char` that is commonly used and has the same problem.

I notice a small problem with the names: `eof/char` means `eof or char` but `$immediate/true` means `$immediate and true`. All the other use or `/` in primdata.ss means also `or` so I think it's better to rename `$immediate/true` that is used only internally in cptypes. But I don't like the alternatives like `$true-immediate` or `true-$immediate`, so I didn't rename it. Does anyone have a better alternative?

### DIFF
--- a/racket/src/ChezScheme/mats/cptypes.ms
+++ b/racket/src/ChezScheme/mats/cptypes.ms
@@ -1192,8 +1192,17 @@
     '(lambda (x) (when (or (not x) (vector? x)) (when x (vector? x))))
     '(lambda (x) (when (or (not x) (vector? x)) (when x #t))))
   (cptypes-equivalent-expansion?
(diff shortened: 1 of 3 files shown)


mflatt:
This looks fine to me, but I didn't try running it.

I like `true-immediate`. FWIW, I read `$immediate+true` as the union of immediate and true.

(I read `$immediate` as "`immediate` that is private to Chez Scheme but public across internal components", so there's not necessarily any need to keep `$` when using the word `immediate`. But I grant that the choice is not obvious when you want to connect to the `$immediate?` predicate.)

PR #3723, opened 2021-03-09
mv get/build-val-first-projection & get/build-late-neg-projection doc

### LINKED ISSUE #2883: Docs incorrectly claim `get/build-late-neg-projection` is in `racket/contract/combinator`
The docs say `get/build-late-neg-projection` is provided from

* `racket/contract/combinator`,
* `racket/contract`, and
* `racket`.

However, it's actually not provided from `racket/contract/combinator`. Instead, it's provided from `racket/contract/base` (and this fact is undocumented).

I've been working around this by requiring it from `racket/contract`, since that's the most specific require spec that's consistent with the documentation.

I'm not sure what I consider "expected behavior" here. It looks like `racket/contract/combinator` is meant for people who are implementing new contract combinators, so it *would* be a great home for `get/build-late-neg-projection`. On the other hand, changing the documentation might have fewer backwards-compatibility issues than changing the behavior.

xxyzz:
close #2883

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -2564,28 +2564,6 @@ contracts.  The error messages assume that the function named by
   the value cannot be coerced to a contract.
 }
 
-@defproc[(get/build-val-first-projection [c contract?])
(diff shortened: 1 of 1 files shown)


rfindler:
Thanks!

PR #3729, opened 2021-03-15
Chez Scheme: add suport for lists in cptypes

gus-massa:
With `list-assuming-immutable?` and the internal construct `immutable-list` the compiler can assume that some lists will
not be mutated.

This is used only to remove some `list?` checks.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/basic.rktl
+++ b/pkgs/racket-test-core/tests/racket/basic.rktl
@@ -540,7 +540,7 @@
   (test #t k:interned-char? #\()
   (test #t k:interned-char? #\ )
   (test #t k:interned-char? '#\newline)
-  (test #f k:interned-char? #\u100)
+  (test (not (eq? 'chez-scheme (system-type 'vm))) k:interned-char? #\u100)
   (test #f k:interned-char? 7)
   (test #f k:interned-char? #t)
   (test #f k:interned-char? #t)
--- a/racket/src/ChezScheme/mats/cptypes.ms
+++ b/racket/src/ChezScheme/mats/cptypes.ms
@@ -210,6 +210,18 @@
   (cptypes-equivalent-expansion?
     '(lambda (x) (when (and (fixnum? x) (zero? x)) x))
     '(lambda (x) (when (and (fixnum? x) (zero? x)) 0)))
+  (cptypes-equivalent-expansion?
+    '(lambda (x f) (when (list-assuming-immutable? x) (f x) (list-assuming-immutable? x)))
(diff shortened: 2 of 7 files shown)


mflatt:
This change has Racket-level implications, right? If so, consider adding a test at that level now that "optimize.rktl" is back.

gus-massa:
I fixed a typo in the description of the first commit, but it's unchanged.

I uncommented a few test and added two additional checks in `test-implies` so when it test that 

    null? => list?

it will also test

    null? => null?
    list? => list?

I don't want to add this automatic test to `test-predicate` in the test block above, because many of the tests are very difficult to track or not yet tracked in CS. 

Anyway, I got an error because it `k:interned-char?` is not tracked (and difficult to track). Anyway^2, I think that all the chars are interned, and `k:interned-char?` should be a platform specific function, so I "fixed" it.

Anyway^3, I get another error, because now `k:interned-char?` is equivalent to `char?`, so `test-implies`
 complains that 

    char? => k:interned-char?

but this test is easy to be removed, if this is the correct behavior in CS.

gus-massa:
I squashed everything, because I had too many small interleaved fixups.

I added an special case for `cdr` so

    (lambda (x)
      (when (list-assuming-immutable? x)
        (list? (cdr x))  ;==> #t
      ))

racket/src/ChezScheme/mats/cptypes.ms
```diff
@@ -584,6 +597,9 @@
   (test-disjoint* '(list? record? vector?))
   (not (test-disjoint* '(list? null?)))
   (not (test-disjoint* '(list? pair?)))
+  (not (test-disjoint* '(list-assume-immutable? null?)))
+  (not (test-disjoint* '(list-assume-immutable? pair?)))
+  (not (test-disjoint* '(list-assume-immutable? list?)))
```
mflatt, line 602: Should these be `list-asuming-immutable?` (with "ing")?

pkgs/racket-test-core/tests/racket/basic.rktl
```diff
@@ -540,7 +540,7 @@
   (test #t k:interned-char? #\()
   (test #t k:interned-char? #\ )
   (test #t k:interned-char? '#\newline)
-  (test #f k:interned-char? #\u100)
+  (test (not (eq? 'chez-scheme (system-type 'vm))) k:interned-char? #\u100)
```
mflatt, line 543: Looks like this should be without `not`.

mflatt:
Looks good, aside from small test corrections.

Issue #3730, opened 2021-03-15
Error installing Linux packages in CI for Solo Chez Build in 32 bits


gus-massa:
Since last week, there is an error in the 32 bits builds of Chez Scheme in the CI.

    The following packages have unmet dependencies:
    libssl-dev:i386 : Depends: libssl1.1:i386 (= 1.1.1-1ubuntu2.1~18.04.8) but it is not going to be installed

https://github.com/racket/racket/runs/2107521053?check_suite_focus=true#step:5:75

(I tried the obvious fix, but I got even more errors ;( .)

pmatos:
@gus-massa thanks - I will look at this.

pmatos:
This has been failing since March 10 and I didn't notice. My apologies.
It turns out I should've changed the default base image to ubuntu-20.04 as the 18.04 was deprecated awhile ago and I didn't. Github did remove the image March 10 and package installation fails on 20.04.

I should also be using a container there to avoid changes in the base VM image causing issues. Pushing a fix soon.

pmatos:
Oh dear - fixing this opens another can of works, which is the fact that chez tests don't like to run inside a container. I have seen something like this before and I think the problem is not necessarily Chez but how github passes variable information into the container. Maybe for now we should run directly in the VM - I will test.

PR #3731, opened 2021-03-16
Run workflow directly in VM

pmatos:
Related to #3730

### DIFF
--- a/.github/workflows/chez-build.yml
+++ b/.github/workflows/chez-build.yml
@@ -17,7 +17,6 @@ on:
 jobs:
   build-linux:
     runs-on: ubuntu-20.04
-    container: ubuntu:20.04
 
     strategy:
       fail-fast: false
@@ -30,8 +29,8 @@ jobs:
     steps:
       - name: Download base dependencies
         run: |
-          apt-get update
-          apt-get install -y make git gcc
+          sudo apt-get update
+          sudo apt-get install -y make git gcc
       - uses: actions/checkout@v2
         with:
           fetch-depth: 50
@@ -42,13 +41,13 @@ jobs:
       - name: Install libs for 32-bit
         if: matrix.mach == 'i3le' || matrix.mach == 'ti3le'
         run: |
-          dpkg --add-architecture i386
-          apt-get update
-          apt-get install -y gcc-multilib lib32ncurses-dev libssl-dev:i386
+          sudo dpkg --add-architecture i386
+          sudo apt-get update
+          sudo apt-get install -y gcc-multilib lib32ncurses-dev libssl-dev:i386
       - name: Install libs for 64-bit
         run: |
-          apt-get update
-          apt-get install -y libncurses5-dev libssl-dev libx11-dev
+          sudo apt-get update
+          sudo apt-get install -y libncurses5-dev libssl-dev libx11-dev
       - name: Build Chez with PB boot files
         working-directory: racket/src/ChezScheme
         run: |


pmatos:
Interested in seeing if this solves the issues we are seeing or if it's really a problem in the tests introduced since the test started failing March 10.

pmatos:
There are some actual issues in 32 bits although running the tests directly on the vm does fix the containerization issues.
@mflatt any ideas what's going on with these failures (unique to 32bits): https://github.com/racket/racket/runs/2121096966?check_suite_focus=true#step:10:199 ?

mflatt:
Looking into the first one, I do see a real bug with cp0 enabled:
```
> (time=?  (make-time 'time-duration 989666791 -2) #f)
#<time-duration -1.010333209>
```
I will investigate more.

mflatt:
The problem there is a bad spec of `make-time` in "primdata.ss". It claims that the nanosecond part is always a fixnum, but that's not true on a 32-bit platform.

mflatt:
It looks like all the failures may be the same problem. I've pushed a repair as a5db4bf376771baac66fbb98d63574387914fe94.

gus-massa:
Should this be fixed too?


https://github.com/racket/racket/blob/a5db4bf376771baac66fbb98d63574387914fe94/racket/src/ChezScheme/s/primdata.ss#L863

mflatt:
Yes - thanks!

pmatos:
Rerunning the tests on rebased branch - thanks for the fix.

pmatos:
@mflatt the date issue seems to be affecting 64bits now. https://github.com/racket/racket/pull/3731/checks?check_run_id=2122559726#step:10:198

gus-massa:
I fixed the test in 64 bits, but now there are a lot of (unrelated) errors about files. For example:

https://github.com/racket/racket/runs/2143976239?check_suite_focus=true#step:11:262

    > 6.mo:Bug in mat filesystem-operations clause 81 at line 2947, char 3 of 6.ms
    > 6.mo:Bug in mat filesystem-operations clause 82 at line 2948, char 3 of 6.ms

https://github.com/racket/racket/blob/993b86ffa364ca5f1080d9fa3ec4cf1d61abbe8b/racket/src/ChezScheme/mats/6.ms#L2947-L2948

xxyzz:
> but now there are a lot of (unrelated) errors about files.

That might be caused by the container. If this pr get rebased again, all tests should pass.

pmatos:
> I fixed the test in 64 bits, but now there are a lot of (unrelated) errors about files. For example:
> 
> https://github.com/racket/racket/runs/2143976239?check_suite_focus=true#step:11:262
> 
> ```
> > 6.mo:Bug in mat filesystem-operations clause 81 at line 2947, char 3 of 6.ms
> > 6.mo:Bug in mat filesystem-operations clause 82 at line 2948, char 3 of 6.ms
> ```
> 
> https://github.com/racket/racket/blob/993b86ffa364ca5f1080d9fa3ec4cf1d61abbe8b/racket/src/ChezScheme/mats/6.ms#L2947-L2948

Really unsure why this is happening atm. Thanks for the fix.

pmatos:
> > but now there are a lot of (unrelated) errors about files.
> 
> That might be caused by the container. If this pr get rebased again, all tests should pass.

Ah - after reading your comment, I see why the tests are failing.

pmatos:
I have now rebased my PR.

pmatos:
This looks good now. Thanks all.

PR #3736, opened 2021-03-20
CS: enable more tests in optimize.rktl

gus-massa:
A few optimization have been implemented in previous commits, so reenable more tests for CS.

Four of them are due to small improvements in cptypes and other parts of Chez Scheme. 

I added a few more (disabled in CS) because `pair`s don't have chaperones, so the original tests with `pair`s are easier than the new tests with `vector`s.

There are other three test that just check that the optimizer is not doing a wrong reduction. They are important in BC because a a similar case has a reduction. In CS the compiler does not have the reduction, so the chances of a bad reduction is minimal, but I they can be enabled just in case.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -444,7 +444,7 @@
                               [y (random)])
                        (list x x y y))
                     #f)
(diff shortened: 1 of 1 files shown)


mflatt:
LGTM - thanks!

PR #3739, opened 2021-03-22
Chez Scheme: fix too much propagation of numbers in cptypes

gus-massa:
In this example, after the check with `eq?` or `eqv?` the value of  the variables `x` and `y` are known, so `cptypes` replaced the reference to the variable with it. 

But it can interact badly with the `eq?` test, for example in this case the first `(eq? x y)` can be `#f` and the second be reduced by the compiler to `#t`.

In spite this is technically correct because `eq?` cannot be used to compare numbers reliably, this behavior is too confusing and it's better to avoid it.

    (define (one)
      (let ([r (random 2)])
        (if (= r 0) (one) r)))

    (define (f x y)
      (define first-comparison (eq? x y))
        (when (and (eqv? x 7.0)
               (eqv? y 7.0))
          (define second-comparison (eq? x y))
          (eq? first-comparison second-comparison)))

    (f (+ 6.0 (one)) (+ 6.0 (one)))))

I had to check that the number is a `fixnum?` and a `target-fixnum?` to avoid problems cross compiling from a 32 bits machine to a 64 bits machine, where the number can be different `bignums?` while compiling, and the same `fixnum?` at run time. (Another possibility is to remove the `mifoldable` flag in the signature of `eq?` and add and add a custom version of fold that understand the cross compilation problems.) 

---

The second commit is a small improvement to `eq?` and `eqv?`. Improve `eq?` and `eq?` handling to reduce expressions like
`(eq? (newline) (void)) => (begin (newline) #t)`. It "fix" a few test in optimize.rktl.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -7110,5 +7110,25 @@
          h)))))
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Regression test related to propagation of flonums
+;; from eqv? tests in CS
+
+(test #t
+      (lambda ()
+        ; optimizer unfriendly 1
+        (define (one)
+          (let ([r (random 2)])
+            (if (= r 0) (one) r)))
+
+        (define (f x y)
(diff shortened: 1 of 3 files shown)


mflatt:
LGTM

PR #3740, opened 2021-03-23
Optimize splitting strings on whitespace

lojic:
Use an optimized function, instead of regexp-split, when string-split
is called with the default separator and trim setting.

### DIFF
--- a/racket/collects/racket/string.rkt
+++ b/racket/collects/racket/string.rkt
@@ -148,7 +148,9 @@
           [else (loop beg (add1 end))])))
 
 (define (string-split str [sep none] #:trim? [trim? #t] #:repeat? [+? #f])
-  (if (and (string? str) (eq? sep none) trim?)
+  (if (and (string? str)
+           (or (eq? sep none) (eq? sep #px"\\s+"))
+           trim?)
       (internal-split-whitespace str)
       (internal-split 'string-split str sep trim? +?)))
 
--- a/racket/collects/racket/string.rkt
+++ b/racket/collects/racket/string.rkt
@@ -148,7 +148,7 @@
           [else (loop beg (add1 end))])))
 
 (define (string-split str [sep none] #:trim? [trim? #t] #:repeat? [+? #f])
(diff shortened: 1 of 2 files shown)


racket/collects/racket/string.rkt
```diff
@@ -124,8 +124,33 @@
...
+           (let ([pos (add1 end)])
+             (if (= beg end)
+                 (loop pos pos)
+                 (cons (substring str beg end)
+                       (loop pos pos))))]
+          [else (loop beg (add1 end))])))
+
 (define (string-split str [sep none] #:trim? [trim? #t] #:repeat? [+? #f])
-  (internal-split 'string-split str sep trim? +?))
+  (if (and (eq? sep none) trim?)
```
samth, line 151: I think this should check that `str` is a `string?` so that the other branch handles the error case.

lojic: Ok, I'll make the change.

lojic: I've pushed the change.

mflatt:
Thanks for this PR!

After considering this a little, I wonder if it makes sense to change `(eq? sep none)` to `(or (eq? sep none) (eq? sep #px"\\s+"))`. Literal regexps are interned, so an `eq?` test would make an explicitly supplied default regexp fast in most cases without significantly slowing down the check.

lojic:
> Thanks for this PR!
> 
> After considering this a little, I wonder if it makes sense to change `(eq? sep none)` to `(or (eq? sep none) (eq? sep #px"\\s+"))`. Literal regexps are interned, so an `eq?` test would make an explicitly supplied default regexp fast in most cases without significantly slowing down the check.

I think that makes sense. I didn't do that originally because I was concerned it might slow things down, but the default case will short circuit the `or`, and with the interning, it should be fine. I'll make the change.

lojic:
I've pushed the change to trigger the fast path if `sep` is either unspecified or the default regexp.

PR #3741, opened 2021-03-24
Add `for[*]/mlist[s]`.

NoahStoryM:
Like `for[*]/list[s]`, but for mutable lists.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/for.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/for.scrbl
@@ -154,6 +154,11 @@ element.
   (error "doesn't get here"))
 ]}
 
+@defform[(for/mlist (for-clause ...) body-or-break ... body)]{
+Like @racket[for/list], but for @tech{mutable lists}.
+@history[#:added "8.0.0.12"]}
+
+
 @defform/subs[(for/vector maybe-length (for-clause ...) body-or-break ... body)
               ([maybe-length (code:line)
                              (code:line #:length length-expr)
@@ -161,7 +166,7 @@ element.
               #:contracts ([length-expr exact-nonnegative-integer?])]{
 
 Iterates like @racket[for/list], but results are accumulated into
-a vector instead of a list. 
+a vector instead of a list.
 
 If the optional @racket[#:length] clause is specified, the result of
 @racket[length-expr] determines the length of the result vector.  In
@@ -267,8 +272,7 @@ is accumulated into a result with @racket[*].
 ]}
 
 
-@defform[(for/lists (id ... maybe-result)
-                    (for-clause ...)
+@defform[(for/lists (id ... maybe-result) (for-clause ...)
            body-or-break ... body)
          #:grammar
          ([maybe-result (code:line) (code:line #:result result-expr)])]{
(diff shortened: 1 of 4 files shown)


jackfirth:
Did you have a concrete use case for this? Mutable lists and pairs are generally discouraged in Racket, I'm not sure making them easier to use is a good idea.

NoahStoryM:
> Did you have a concrete use case for this? Mutable lists and pairs are generally discouraged in Racket, I'm not sure making them easier to use is a good idea.

Hi jackfirth!


The first reason I wanted to add for/mlist was to make the use of mlist and list more consistent -- both are the basic Linked Structure in Racket, and many list operators should be able to correspond to a mlist version. In Racket, we can implement these operators by recursion, and adding support for for/mlist really only makes mlist easier to use.

I quite agree that, define list as a kind of immutable data structure is a good idea. After I learned SICP, my understanding for the assignment is that, the assignment introduces state, so the equal things at a moment may be different later, and therefore assignment introduces time to our code. It destroyed the substitution model, so in most cases we don't encourage the use of the assignment.

However, Racket/Scheme does not eliminate assignment as radically as pure functional languages. In my opinion, programming is essentially a way to use a language to express models in our mind, or to describe the world.

In the real world we almost always deal with time-related problems, even the order of Mathematical calculation can also be seen as a model of time (a lot of algorithms use assignment to control the order of calculation. Although some algorithms use the assignment, but they can promise no external environment changes, and the same input has the same output, so you can use it safely). Assignment is always an important option in my mind when dealing with states (or time), because it simply describes my understanding of those states.

mlist is just a Linked Structure that you can change values to. mList is not recommended, but if it's discouraged to use mlist for the reason that there are too few operations on mlists, so programmers have to define these operators themselves, I don't think it's a good idea. If a programmer doesn't understand the danger of assignment (and I'm sure all Racket programmers understand it), there's no way to prevent him from writing bad code. Else if a programmer understands the problems associated with assignment and still decides to use it, why not make his experience a little more comfortable?

mflatt:
I can see the argument for `for/mlist`, but it seems like it belongs with the mlist functions in`compatibiity/mlist` (in the "compatibility-lib" package in the https://github.com/racket/compatibility repo) instead of `racket/for`.

NoahStoryM:
At first I also thought about adding `for/mlist` to the `compatibility-lib` package, but considering the consistency between `mlist` and `list`, I felt that the definitions of `for/list` and `for/mlist` should also be consistent, they should probably be implemented in the same way. And on the other hand `for/mlist` does not need to use the operators in the `compatibility-lib`, it just uses `mcons` (and defines `alt-mreverse` like `alt-reverse`), so I think it makes sense to define `for/mlist` in `racket/for` as well.

Defining `for/mlist` in the `compatibility-lib` also makes sense, but I'm not sure how to ensure that the implementations of `for/list` and `for/mlist` are always consistent in that way.

Can you give me some advice?

mflatt:
The `for/fold/derved" forms are good for implementing new `for` variants. As an example, see the implementation of `for/set` in "racket/collects/racket/private/set-types.rkt". That doesn't let you reuse everything that does into `for/list`, but it handles the main work.

NoahStoryM:
Thanks, I'll learn it later. Feel free to close the PR if you feel it's better to define `for/mlist` in `compatibility-mlist`. (And I plan to add more useful operators for `compatibility-mlist`)

PR #3745, opened 2021-03-29
rumble,error: fix error case when given a field with no value

### LINKED ISSUE #3744: cs: invalid memory reference due to bad call to `raise-arguments-error`

Commit b320a1a114 .

**What program did you run?**

In `a.rkt`:

```racket
#lang racket/base

(provide
 f)

(define (f x)
  (raise-arguments-error 'f "invalid" x)) ;; missing the "field" argument for `x`
```

In `b.rkt`:

```racket
#lang racket/base

(require "a.rkt")

(f "foo")
```

Compile `a.rkt`:

    raco make a.rkt

Run `b.rkt`:

    racket b.rkt

A contract error should be raised, instead I get:

```
$ racket b.rkt
  context...:
   body of "/path/to/b.rkt"
```

Bogdanp:
Closes #3744

### DIFF
--- a/racket/src/cs/rumble/error.ss
+++ b/racket/src/cs/rumble/error.ss
@@ -150,7 +150,7 @@
         [(null? more) '()]
         [(string? (car more))
          (cond
-          [(null? more)
+          [(null? (cdr more))
            (raise-arguments-error 'raise-arguments-error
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #3746, opened 2021-03-29
xexpr: add `current-unescaped-tags` to disable escaping for certain tags

Bogdanp:
I don't much like the name `current-unescaped-tags`, but couldn't come up with something better. Suggestions welcome!


jeapostrophe:
This looks good to me, thank you!

PR #3748, opened 2021-03-30
Implementation of 3x3 and 4x4 Toom-Cook multiplication.

samth:
This speeds up `(factorial 1000000)` (using factorial from math/number-theory)
by about 3x, and the conversion of the result to a string by about 2x.

Benchmark:

    #lang racket
    (require math/number-theory)
    (define n (time (factorial 1000000)))
    (define s (time (number->string n)))
    (string-length s)

Current Racket CS:
cpu time: 19135 real time: 19137 gc time: 372
cpu time: 33416 real time: 33418 gc time: 463

Current Racket BC (GMP is really fast):
cpu time: 1465 real time: 1465 gc time: 51
cpu time: 3661 real time: 3659 gc time: 3

This PR:
cpu time: 6173 real time: 6172 gc time: 168
cpu time: 17846 real time: 17847 gc time: 377

Cutoff between Karatsuba and Toom3 estimated by mflatt.
Cutoff between Toom3 & Toom4 guessed.

### DIFF
--- a/racket/src/ChezScheme/mats/5_3.ms
+++ b/racket/src/ChezScheme/mats/5_3.ms
@@ -1690,6 +1690,16 @@
                                  (and (exact? x) (exact? y))
                                  (or (inexact? x) (inexact? y)))
                              (g (+ j 1)))))))))
+    (let ([sb* (foreign-procedure
+		"(cs)mul" (scheme-object scheme-object) scheme-object)])
+      ;; (expt 2 100000) is big enough that all multiplication algorithms
+      ;; are exercised
+      ;; we add a power of 3 so that the number isn't too simple
+      (eqv? (sb* (+ 1 (expt 3 50) (expt 2 100000))
+		 3)
+	    (* (+ 1 (expt 3 50) (expt 2 100000))
+	       3)))
+      
     (error? ; #f is not a fixnum
       (* 3 #f))
     (error? ; #f is not a fixnum
--- a/racket/src/ChezScheme/s/5_3.ss
+++ b/racket/src/ChezScheme/s/5_3.ss
@@ -2388,7 +2388,103 @@
                               [(fx= x 1) (unless (number? y) (nonnumber-error who y)) y]
                               [else ($negate who y)])]
                             [else (integer* x y)])
-                           (let ()
+                           (let ([slim 32]
+				 [klim 100]
+				 [t3lim 512])
+                             ; both of the following functions were adapted from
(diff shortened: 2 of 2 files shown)


mflatt:
`bitwise-bit-field` may be a better way to get the mid and lo segments, like this (if I didn't break it):
```
 [x-hi (ash x (fx* -2 k))]
 [y-hi (ash y (fx* -2 k))]
 [x-mid (bitwise-bit-field x k (fx* 2 k))]
 [y-mid (bitwise-bit-field y k (fx* 2 k))]
 [x-lo (bitwise-bit-field x 0 k)]
 [y-lo (bitwise-bit-field y 0 k)]
```

On my machine, 512 was a better cutoff than 100 before the `bitwise-bit-field` change, but the difference mostly went away with the change.

samth:
Is there a reason not to use bitwise-bit-field for all 3 segments (and also in karatsuba)?

mflatt:
The numbers are signed, right? In that case, for the high segment, you want to keep the sign — so you don't want a finite bit field.

Using `bitwise-bit-field` seems better for the low part in `karatsuba`, too.

samth:
What's the easiest way to access `(constant bigit-bits)` and `$bignum-length` from the Chez top-level? Is that possible?

mflatt:
I don't think you can access `(constant ...)`, but you can use `#%$bignum-length`.

samth:
This is now ready for review, although I haven't written new tests. This was tested quite extensively with the following code: https://gist.github.com/92d24b44184fbcdace3215b193848e64 It's much easier to test when I can just compare with the behavior of `*`.

mflatt:
It would probably make sense to give the threshold values names, since they're used multiple times.

For testing in Chez Scheme "5_3.ms", you could use
```
(define sb*
  (foreign-procedure "(cs)mul" (scheme-object scheme-object) scheme-object))
```
to get schoolbook multiplication for comparison.

How much / when does toom4 help?

samth:
With just Toom3, the benchmark in the commit message was about 9 seconds, and adding Toom4 got it down to 6 seconds.

I'll add a test the uses schoolbook multiplication, but unfortunately doing a lot of tests will be slow. Maybe I should write some tests with multiplication and division to check that expected arithmetic laws hold.

I would also like to avoid the duplication of the tests for various sizes, but I haven't yet come up with something nice.

samth:
I think this is now ready.

mflatt:
LGTM

Issue #3747, opened 2021-03-29
raco pkg install: Don't rely on `.git` suffix to recognize Git repos


sschwarzer:
**Is your feature request related to a problem? Please describe.**
I have a package repository hosted on Sourcehut, at https://git.sr.ht/~sschwarzer/todo-txt . When I try
```
$ raco pkg install https://git.sr.ht/~sschwarzer/todo-txt
```
I get an error
```
Downloading checksum for todo-txt
Cloning remote directory https://git.sr.ht/~sschwarzer/todo-txt
raco pkg install: could not find MANIFEST for package source
  source: https://git.sr.ht/~sschwarzer/todo-txt
  possible solution:
   If the URL is intended to refer to a Git repository, use
     https://git.sr.ht/~sschwarzer/todo-txt.git
   so that the URL ends in ".git"
```
However, Sourcehut repos can't be accessed with a `.git` suffix:
```
$ raco pkg install https://git.sr.ht/~sschwarzer/todo-txt.git
Querying Git references for todo-txt at https://git.sr.ht/~sschwarzer/todo-txt.git
pkg: Git checkout initial protocol failed;
 the given URL might not refer to a Git repository
  given URL: https://git.sr.ht/~sschwarzer/todo-txt.git
  context...:
   /sd/racket/current/collects/pkg/private/stage.rkt:762:25
   /sd/racket/current/collects/net/git-checkout.rkt:336:0: parse-initial-refs
   /sd/racket/current/collects/net/git-checkout.rkt:73:8
   /sd/racket/current/collects/net/git-checkout.rkt:54:2: retry-loop
   /sd/racket/current/collects/pkg/private/stage.rkt:59:2: lookup-normally
   /sd/racket/current/collects/pkg/private/stage.rkt:107:0: stage-package/info
   /sd/racket/current/collects/pkg/private/install.rkt:141:0: install-packages
   /sd/racket/current/collects/pkg/private/install.rkt:925:4
   /sd/racket/current/collects/racket/contract/private/arrow-val-first.rkt:555:3
   /sd/racket/current/collects/racket/file.rkt:437:8
   /sd/racket/current/collects/racket/file.rkt:426:0: call-with-file-lock
   /sd/racket/current/collects/pkg/main.rkt:216:16
   body of (submod "/sd/racket/current/collects/pkg/main.rkt" main)
   /sd/racket/current/collects/raco/raco.rkt:41:0
   body of "/sd/racket/current/collects/raco/raco.rkt"
[...]

PR #3749, opened 2021-03-31
add support for `git+https://` package sources

mflatt:
This is a first cut at the suggestion in #3747.

The essential ingredients are the change in source parsing to recognize `git+https://`, avoiding the addition of ".git" for that kind of source in `normalize-url` (which turns out to be a matter of *not* changing `normalize-url`), and documentation updates.

For completeness and consistency, there's also a `--type git-url` option that can force a parsing of `https://` more like `git+https://`, although I think the `--type` option has turned out not to be that useful. The possibility of `--type` triggers some extra internal plumbing to track, but it's not too bad.

### DIFF
--- a/racket/collects/pkg/name.rkt
+++ b/racket/collects/pkg/name.rkt
@@ -4,7 +4,8 @@
          racket/format
          racket/string
          racket/path
-         net/url)
+         net/url
+         "private/git-url-scheme.rkt")
 
 (provide
  package-source-format?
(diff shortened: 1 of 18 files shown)


sorawee:
The `raise-arguments-error` commit seems to be accidentally included here.

racket/collects/pkg/name.rkt
```diff
@@ -136,8 +137,8 @@
         (regexp-match? rx:package-name s))
     (values (validate-name s complain #f) 'name)]
    [(and (eq? type 'clone)
-         (not (regexp-match? #rx"^(?:https?|git(?:hub)?)://" s)))
-    (complain "repository URL must start 'http', 'https', 'git', or 'github'")
+         (not (regexp-match? #rx"^(?:https?|git(?:hub|[+]https?)?)://" s)))
+    (complain "repository URL must start 'http', 'https', 'git', 'git+http', 'git+hhtps', or 'github'")
```
samth, line 141:
```suggestion
    (complain "repository URL must start 'http', 'https', 'git', 'git+http', 'git+https', or 'github'")
```

mflatt:
Merged as 2606ae3d8ea95e1b644f9400e156ec2ed781a582

PR #3754, opened 2021-04-02
Add print-value-columns parameter.

mikesperber:


mflatt:
This looks good to me. I would be inclined to clarify in the docs that the default print handler does not currently use the parameter (but other handlers may use it).

samth:
Should there be some tests here? I guess they would only be that the parameter exists and accepts only numbers.

mikesperber:
Tried to do what @mflatt and @samth suggested.

rfindler:
@greghendershott This new parameter has me wondering about Emacs mode integration and whether or not Racket mode can do something like what DrRacket's default global-port-print handler is doing to print values returned to the REPL with a window-appropriate width (in DrR if you change the width of the window it uses a different pretty-printing width internally; the value used isn't visible to the program; it just affects the port that `print` prints to).

There may also something useful that xrepl can do here too, I'm not sure.

mflatt:
@mikesperber Thanks! This seems better, and it helps explain why "values" is in the name.

Did you mean to change the contract on `print-syntax-width` in the docs?

One one more small request: add `@history[#:added "8.0.0.13"]` at the end of the just (just before `}`).

Probably `racket/init` should do something with the parameter in the handler that it installs for `current-print`, but that doesn't need to be your problem.

Do you have push permission on this repo, or should I merge when you're ready?

mikesperber:
@mflatt Updated the patch as per your comments. I don't have push permission, though.

PR #3756, opened 2021-04-03
Clarify docs form contract-out and define/contract

Metaxal:
A small readability improvement for the documentation about `contract-out` and `define-contract`.

Happy to make changes if, in particular, the terminology is incorrect.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1864,14 +1864,24 @@ is provided from the module. In
 addition, clients of the module must live up to the contract specified
 by @racket[contract-expr] for each export.
 
-The @racket[contract-out] form treats modules as units of
-blame. The module that defines the provided variable is expected to
-meet the positive (co-variant) positions of the contract. Each module
-that imports the provided variable must obey the negative
-(contra-variant) positions of the contract. Each @racket[contract-expr]
-in a @racket[contract-out] form is effectively moved to the end of the
-enclosing module, so a @racket[contract-expr] can refer to variables
-that are defined later in the same module.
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/contracts.scrbl
```diff
@@ -1864,14 +1864,24 @@ is provided from the module. In
...
+@margin-note{Technically, the @racket[contract-out] form treats modules as
+units of blame. The module that defines the provided variable is expected to
+meet the positive (co-variant) positions of the contract. Each module that
+imports the provided variable must obey the negative (contra-variant)
+positions of the contract.}
+The contracts of @racket[contract-out] are placed at the boundary of the
+enclosing module. This means that, for example, if the module provides a form
+@racketid[my-form] with @racket[contract-out], the contract is checked only
+when @racketid[my-form] is called from outside the module, but when called
+from within the module. When @racketid[my-form] is called from outside the
```
otherjoel, line 1876: Was this supposed to be “but **not** when called from within the module”?

Metaxal: Thanks for catching this. I thought I had already fixed it, but I forgot to push :/

Metaxal:
ping?

samth:
cc @rfindler

rfindler:
This looks like a good change to me, except that this sentence got lost:

```
Each @racket[contract-expr] in a @racket[contract-out] form is effectively moved to the end of the enclosing module, so a @racket[contract-expr] can refer to variables that are defined later in the same module.
```

That sentence is about the order of evaluation of the expressions inside a contract, not about blame, so probably it needs to come back and be put somewhere. Perhaps with a bit more explanation?

PR #3758, opened 2021-04-04
Use export-orig-stx instead of export-out-sym when appropriate

greghendershott:
Forms like rename-out supply the export-id syntax as export-orig-stx,
as well as the symbolic value as export-out-sym. In that case, expand
using export-orig-stx. That way, the piece of syntax representing the
export-id will have srcloc and other properties. This may be useful to
tools that deal with renaming provides.

This commit tries to be conservative by doing the old behavior unless
(eq? (syntax-e (export-orig-stx export)) (export-out-sym export)).

This commit doesn't update documentation for `export-orig-stx`.
Instead the spirit of this commit is to treat this as a private
implementation detail of `rename-out`, as opposed to stating some new
official API for renaming provide transformers in general.

### DIFF
--- a/racket/collects/racket/private/reqprov.rkt
+++ b/racket/collects/racket/private/reqprov.rkt
@@ -763,7 +763,13 @@
                                        (export-local-id export)
                                        (quasisyntax/loc out
                                          (rename #,(export-local-id export)
-                                                 #,(export-out-sym export))))]
+                                                 #,(if (eq? (syntax-e (export-orig-stx export))
+                                                            (export-out-sym export))
+                                                       (export-orig-stx export)
+                                                       (datum->syntax
+                                                        #f
+                                                        (export-out-sym export)
+                                                        (export-orig-stx export))))))]
                                   [mode (export-mode export)])
                               (let ([phased
                                      (cond


greghendershott:
This re <https://racket.slack.com/archives/C06V96CKX/p1617476121233900?thread_ts=1617474202.233600&cid=C06V96CKX>.

Caveat:

I tried this temporiarily modifying `reqprov.rkt` in my 8.0.0.12 snaphot.

"It worked".

Then I caught up with the Racket repo, to make this PR.

However I'm still waiting for that to build, much less re-try it manually for that or run any automated tests.

I figured I'd go ahead and do the PR for the sake of getting feedback if I understood the suggestion and am on the right track, or not.

mflatt:
This looks like the right idea to me.

When the symbols don't match, would it be useful to apply a source location in `orig-stx` to `out-sym`?

greghendershott:
> When the symbols don't match, would it be useful to apply a source location in orig-stx to out-sym?

Do you mean something like:

- Walk `(syntax->list orig-stx)` recursively to find an identifier that does match `out-sym`.
- If found, use that syntax object (and its srcloc and props) as the export-id.
- Else at least apply the srcloc for the entire `orig-stx`, to `out-sym`.  (On the rationale that, although that srcloc might be N/A, it could still be better than no srcloc at all).

?

greghendershott:
Like say:

```diff
modified   racket/collects/racket/private/reqprov.rkt
@@ -763,11 +763,7 @@
                                        (export-local-id export)
                                        (quasisyntax/loc out
                                          (rename #,(export-local-id export)
-                                                 #,(if (eq? (syntax-e
-                                                             (export-orig-stx export))
-                                                            (export-out-sym export))
-                                                       (export-orig-stx export)
-                                                       (export-out-sym export)))))]
+                                                 #,(export-out-stx export))))]
                                   [mode (export-mode export)])
                               (let ([phased
                                      (cond
@@ -786,6 +782,13 @@
                  (syntax/loc stx
                    (begin new-out ...)))))]))]))
 
+  (define-for-syntax (export-out-stx export)
+    (or (let loop ([stx (export-orig-stx export)])
+          (cond [(eq? (syntax-e stx) (export-out-sym export)) stx]
+                [(ormap loop (syntax->list stx))]
+                [else #f]))
+        (export-out-sym export)))
+
   (define-for-syntax (combine-prop b a)
     (if a (if b (cons a b) a) b))
```

mflatt:
I only had in mind the "else". So, either the symbol matches immediately, as in your code, or the source location is transferred.

My guess is that searching inside `orig-stx` probably isn't useful, but I'm not sure. There's also some chance that searching could take too long (e.g., take quadratic time if a bunch of identifiers are extracted from an `orig-stx` term whose size is proportional to the number of identifiers).

greghendershott:
Ah OK.

My quick diff left out the "else" --- the only part you had in mind. :)

So IIUC I think you mean something like

```diff
   (define-syntax (provide-trampoline stx)
     (syntax-case stx ()
       [(_ out ...)
        (letrec ([transform-simple
                  (lambda (out)
                    (let ([exports (expand-export out null)])
                      (map (lambda (export)
                             (let ([base
                                    (if (eq? (syntax-e (export-local-id export))
                                             (export-out-sym export))
                                        (export-local-id export)
                                        (quasisyntax/loc out
                                          (rename #,(export-local-id export)
                                                  #,(if (eq? (syntax-e
                                                              (export-orig-stx export))
                                                             (export-out-sym export))
                                                        (export-orig-stx export)
-                                                       (export-out-sym export)))))]
+                                                       (datum->syntax
+                                                        (export-orig-stx export)
+                                                        (export-out-sym export)
+                                                        (export-orig-stx export))))))]
                                   [mode (export-mode export)])
                               (let ([phased
```

sorawee:
This is long ago, so I might misremember the details, but myself in the past thought that "using a uniform approach for all provide transformers looks like a bad idea" and that "a better approach is that each provide transformer should explicitly attach `disappeared-use` manually to `export-local-id`, so that it can precisely control how arrow will be drawn". Is that relevant to this PR?

Both quotes are from #3386, which was my attempt to add arrows from definitions to `all-defined-out`. I abandoned the PR (due to the above reason) in favor of #3391.

greghendershott:
In my motivating example, `rename-out` takes the export-id as a piece of syntax from the original user program, _effectively_ does a `syntax->datum`, and inserts that symbol in the expanded syntax. I think that specifically is a fairly straightforward situation to change.

But I have no real mileage with the `disappeared-use` property, so I might be failing to understand and appreciate the connection.

mflatt:
@greghendershott Yes, that's what I'd do, except that I imagined using `#f` for the first argument to `datum->syntax` (just getting source location, not scopes). As always here, I'm not sure it matters.

greghendershott:
@mflatt Thanks for the review! I force-pushed commit 66a8436, with that change, and rewriting the commit message.

PR #3760, opened 2021-04-06
Add `license` field for package info.rkt files

LiberalArtist:
I propose adding a `license` field to the package metadata in `info.rkt` files.

[SPDX license expressions](https://spdx.dev/ids/) are used to encode license metadata in a variety of projects and package systems, including [Cabal](https://cabal.readthedocs.io/en/latest/cabal-package.html#pkg-field-license), [Cargo](https://doc.rust-lang.org/cargo/reference/manifest.html#the-license-and-license-file-fields), [RubyGems](https://guides.rubygems.org/specification-reference/#license=), [NPM](https://docs.npmjs.com/cli/v7/configuring-npm/package-json#license), Python's proposed [PEP 639](https://www.python.org/dev/peps/pep-0639/), and the [Linux kernel](https://www.kernel.org/doc/html/latest/process/license-rules.html#license-identifier-syntax). For Racket, I propose an s-expression syntax for SPDX expressions.

# Updated Summary

> See <https://github.com/racket/racket/pull/3760#issuecomment-820800099>. See also the original proposal below. 

An *SPDX License S-Expression* is a datum with the following grammar:

```                                   
  <license-sexp> = <license-id>                            
                 | (<license-id> WITH <exception-id>)
                 | (<license-sexp> AND <license-sexp>)               
                 | (<license-sexp> OR <license-sexp>)
```

Where:
  - `license-id` is an identifier from the [SPDX License List](https://spdx.org/licenses/index.html); and
  - `exception-id` is an identifier from the [https://spdx.org/licenses/exceptions-index.html"]{SPDX License Exceptions} list (like `Classpath-exception-2.0`).

For example:
https://github.com/racket/racket/blob/d20a1c7f117e385dfbd852d75545b351f8b66166/pkgs/base/info.rkt#L28-L29 

The s-expression grammar is designed so that `(format "~s" license-sexp)` produces a string conforming to the grammar in [Appendix IV](https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/) and [Appendix V](https://spdx.github.io/spdx-spec/appendix-V-using-SPDX-short-identifiers-in-source-files/) of the SPDX Specification v2.2.0, which is specified in terms of character sequences.

Concretely, this pull request documents the proposed field and adds it to the `info.rkt` files of the packages in this repository, as well as those added by `raco pkg new`. It does not add any other tools, leaving those for other packages and pull requests. (I plan to make some.)

# Original Proposal

```
  <license-datum> = (SPDX-License-Identifier: <compound-expr>)        
                  | <compound-expr>                                   
                                                                  
  <compound-expr> = <versioned-license-id>                            
                  | (<versioned-license-id> WITH <license-exception-id>)
                  | (<compound-expr> AND <compound-expr>)               
                  | (<compound-expr> OR <compound-expr>)  
```
For example:
```racket
(define license
  '(SPDX-License-Identifier: (Apache-2.0 OR MIT)))
```

If written in the usual way (i.e. what `write` would produce—no unusual escape sequences), a s-expression matching this grammar should also conform to the syntax in [Appendix IV](https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/) and [Appendix V](https://spdx.github.io/spdx-spec/appendix-V-using-SPDX-short-identifiers-in-source-files/) of the SPDX Specification v2.2.0, which is specified in terms of character sequences. This represents a slightly different implied AST than the specification, which has a `license-id` and a `+` operator rather than a `versioned-license-id`. The `versioned-license-id` approach seems especially justified since the current SPDX specification (see [here](https://spdx.dev/ids/), under “Allowing later versions of a license”) handles versioning differently for GNU and non-GNU licenses: you write `AFL-2.0` or `AFL-2.0+` but `GPL-3.0-only` or `GPL-3.0-or-later` (and neither `GPL-3.0` nor `GPL-3.0+` are allowed). So a tool handling versioning properly really needs the license list metadata, not just the expression syntax.

(Making optional parentheses mandatory is presumably a less remarkable change.)

--------

Concretely, this pull request documents the proposed field and adds it to the `info.rkt` files for the packages in this repository: it doesn't add any tools. A linter in `raco pkg` might be a nice addition, but it raises the question of whether to include the SPDX data and validate that the `versioned-license-id` and `license-exception-id` are defined. If it goes in this repository, I'm inclined to think not, but it might be useful for, say, the pkgs.racket-lang.org website.

### DIFF
--- a/pkgs/racket-doc/pkg/scribblings/pkg.scrbl
+++ b/pkgs/racket-doc/pkg/scribblings/pkg.scrbl
@@ -1342,6 +1342,56 @@ The following @filepath{info.rkt} fields are used by the package manager:
        set up (plus collections for global documentation indexes and
        links).}
 
+ @item{@definfofield{license} --- a @deftech{license s-expression}
+  specifying the package's license. A license s-expression represents an @deftech{SPDX}
+  @hyperlink["https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/"]{
+   license expression} as a datum with the quoted form:
+
+  @racketgrammar[#:literals (AND OR WITH) license-sexp
+                 license-id
+                 (license-id WITH exception-id)
+                 (license-sexp AND license-sexp)
+                 (license-sexp OR license-sexp)]
+
+  @margin-note{See @elemref["spdx-plus-operator"]{further details below}
+   about @racket[_license-id] and the @litchar{+} operator.}
+
+  where a:
+
+  @itemize[
+ @item{@racket[_license-id] is a short-form identifier from the
+    @hyperlink["https://spdx.org/licenses/index.html"]{SPDX License List},
+    e.g@._ @racketvalfont{LGPL-3.0-or-later}, @racketvalfont{Apache-2.0},
+    or @racket[BSD-3-Clause]; and an}
+ @item{@racket[_exception-id] is an identifier from the
+    @hyperlink["https://spdx.org/licenses/exceptions-index.html"]{
+     SPDX License Exceptions} list, e.g@._ @racketvalfont{Classpath-exception-2.0}.}]
+
+  For example, packages in the main Racket distribution
+  define @racketidfont{license} as:
+  @racketblock[(define license
+                 '(Apache-2.0 OR MIT))]
+
+  The grammar of @tech{license s-expressions} is designed so that
+  @racket[(format "~s" license)] produces a string conforming to the grammar in
+  @hyperlink["https://spdx.github.io/spdx-spec/appendix-IV-SPDX-license-expressions/"]{
+  Appendix IV} and
+  @hyperlink["https://spdx.github.io/spdx-spec/appendix-V-using-SPDX-short-identifiers-in-source-files/"]{
+  Appendix V}
+  of the SPDX Specification v2.2.0,
+  which is specified in terms of character sequences.
+
+  @elemtag["spdx-plus-operator"]{If the @litchar{+} operator is used,}
+  it must be written as part of the @racket[_license-id],
+  e.g@._ @racketvalfont{AFL-2.0+}.
+  Note that the @hyperlink["https://spdx.dev/ids/"]{SPDX Workgroup has deprecated}
+  (under ``Allowing later versions of a license'') the use of the @litchar{+}
+  operator with GNU licenses: thus, one writes
+  @racketvalfont{AFL-2.0} or @racketvalfont{AFL-2.0+} but
+  @racketvalfont{GPL-3.0-only} or @racketvalfont{GPL-3.0-or-later}
+  (and neither @racket[GPL-3.0] nor @racket[GPL-3.0+] are correct).
+ }
+
  @item{@definfofield{distribution-preference} --- either
        @racket['source], @racket['built], or @racket['binary],
        indicating the most suitable distribution mode for the package
@@ -1367,7 +1417,8 @@ The following @filepath{info.rkt} fields are used by the package manager:
 ]
 
 @history[#:changed "6.1.0.5" @elem{Added @racketidfont{update-implies}.}
-         #:changed "6.1.1.6" @elem{Added @racketidfont{distribution-preference}.}]
+         #:changed "6.1.1.6" @elem{Added @racketidfont{distribution-preference}.}
+         #:changed "8.1.0.8" @elem{Added @racketidfont{license}.}]
 
 @; ----------------------------------------
 
--- a/pkgs/at-exp-lib/info.rkt
+++ b/pkgs/at-exp-lib/info.rkt
@@ -8,3 +8,6 @@
 (define pkg-authors '(eli mflatt))
 
 (define version "1.2")
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/base/info.rkt
+++ b/pkgs/base/info.rkt
@@ -24,3 +24,6 @@
 (define pkg-desc "Racket libraries that are currently always available")
 
 (define pkg-authors '(mflatt))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/compiler-lib/info.rkt
+++ b/pkgs/compiler-lib/info.rkt
@@ -14,3 +14,6 @@
 (define pkg-authors '(mflatt))
 
 (define version "1.10")
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/compiler-test/info.rkt
+++ b/pkgs/compiler-test/info.rkt
@@ -20,3 +20,6 @@
 		     "dynext-lib"
 		     "mzscheme-lib"))
 (define update-implies '("compiler-lib"))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/compiler/info.rkt
+++ b/pkgs/compiler/info.rkt
@@ -8,3 +8,6 @@
 (define pkg-desc "Racket compilation tools, such as `raco exe'")
 
 (define pkg-authors '(mflatt))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/net-doc/info.rkt
+++ b/pkgs/net-doc/info.rkt
@@ -14,3 +14,6 @@
                      "web-server-doc"
                      "web-server-lib"))
 (define update-implies '("net-lib"))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/net-lib/info.rkt
+++ b/pkgs/net-lib/info.rkt
@@ -7,3 +7,6 @@
 (define pkg-desc "implementation (no documentation) part of \"net\"")
 
 (define pkg-authors '(mflatt))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/net-test/info.rkt
+++ b/pkgs/net-test/info.rkt
@@ -16,3 +16,6 @@
                      "sandbox-lib"
                      "web-server-lib"))
 (define update-implies '("net-lib"))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/net/info.rkt
+++ b/pkgs/net/info.rkt
@@ -8,3 +8,6 @@
 (define pkg-desc "Networking libraries")
 
 (define pkg-authors '(mflatt))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/racket-benchmarks/info.rkt
+++ b/pkgs/racket-benchmarks/info.rkt
@@ -15,3 +15,6 @@
 
 (define pkg-desc "Racket benchmarks")
 (define pkg-authors '(eli jay mflatt robby samth stamourv))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/racket-build-guide/info.rkt
+++ b/pkgs/racket-build-guide/info.rkt
@@ -14,3 +14,6 @@
 (define pkg-desc "Racket build and contribution documentation")
 
 (define pkg-authors '(mflatt))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/racket-doc/info.rkt
+++ b/pkgs/racket-doc/info.rkt
@@ -44,3 +44,6 @@
 (define pkg-authors '(eli jay matthias mflatt robby ryanc samth))
 
 (define version "1.1")
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/racket-index/info.rkt
+++ b/pkgs/racket-index/info.rkt
@@ -26,3 +26,6 @@
                             "scribblings/main/user/compiled/release_scrbl.zo"
                             "scribblings/main/user/compiled/search_scrbl.zo"
                             "scribblings/main/user/compiled/start_scrbl.zo"))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/racket-lib/info.rkt
+++ b/pkgs/racket-lib/info.rkt
@@ -20,3 +20,6 @@
 (define pkg-desc "Combines platform-specific native libraries that are useful for base Racket")
 
 (define pkg-authors '(eli jay matthias mflatt robby ryanc samth))
+
+(define license
+  '(Apache-2.0 OR MIT))
--- a/pkgs/racket-test-core/info.rkt
+++ b/pkgs/racket-test-core/info.rkt
@@ -18,3 +18,6 @@
 (define pkg-desc "Minimal core version of Racket test suites")
 
 (define pkg-authors '(eli jay matthias mflatt robby ryanc samth))
+
+(define license
+  '(Apache-2.0 OR MIT))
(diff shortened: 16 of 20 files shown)


jackfirth:
Haven't fully parsed the design of the grammar proposed here, but I'd like to say I'm definitely in support of having structured standardized license information available in package `info.rkt` metadata.

pkgs/racket-doc/pkg/scribblings/pkg.scrbl
```diff
@@ -1342,6 +1342,33 @@ The following @filepath{info.rkt} fields are used by the package manager:
...
+                  ]
+  where a @racket[_versioned-license-id] is a short-form identifier from the
+  @hyperlink["https://spdx.org/licenses/index.html"]{SPDX License List},
+  potentially with a version suffix (e.g. @racket[GPL-3.0-or-later], @racket[AFL-2.0+], or
+  @racket[LGPL-2.0-only]), and a @racket[_license-exception-id] is an identifier from the
+  @hyperlink["https://spdx.org/licenses/exceptions-index.html"]{SPDX License Exceptions}
+  list (e.g. @racket[Classpath-exception-2.0]).
+  
+  Writing the @racket[_license-datum] using @racket[quote], without internal newlines,
+  and in the form beginning with @racket[SPDX-License-Identifier:] my facilitate
```
greghendershott, line 1364: tiny typo -- "my" => "may" ?

greghendershott:
I wonder about the goal of allowing existing extra-Racket tools to parse this. 

If that's an important goal, then shouldn't the `SPDX-License-Identifier:` be mandatory not optional? Otherwise, it feels weird having that in the RHS and I'd suggest something like `(define spdx-version <compound-expr>)`?

Similarly if not for that goal, we'd probably use prefixed s-expressions, not infix, like the rest of info.rkt?

I also worry some such tool would try to be clever and look only in comments, and end up ignoring this in info.rkt because it's not in a comment. :)

---

I love the idea of being able to do things like:

- Filter packages by license on pkgs.racket-lang.org
- Run a checker (is there a license? is it compatible transitively with dependencies?).

I wonder if whoever does that work, someday, should have latitude to determine the grammar? e.g. Maybe parsing infix expressions isn't rocket science, but it's more work they need to do, and maybe they'd rather not. Or maybe there's some other consideration that only surfaces when trying to implement these things.

---

p.s. All of the above isn't a super strong opinion! I love that you're making a proposal in this space!

pkgs/racket-doc/pkg/scribblings/pkg.scrbl
LiberalArtist: Thanks; will fix.

LiberalArtist:
> I wonder about the goal of allowing existing extra-Racket tools to parse this.
> 
> If that's an important goal, then shouldn't the `SPDX-License-Identifier:` be mandatory not optional? Otherwise, it feels weird having that in the RHS and I'd suggest something like `(define spdx-version <compound-expr>)`?
…
> 
> I also worry some such tool would try to be clever and look only in comments, and end up ignoring this in info.rkt because it's not in a comment. :)

Very good points. I guess, broadly speaking, this is my attempt to balance at least two considerations:

 1. Given that the SPDX standard exists, it would be nice at least not to create barriers to interoperability.
 2. I don't actually know of any language-independent SPDX tools that I particularly want to run, so I'm reluctant to mandate boilerplate.

I am very interested in discussion about balancing these and other considerations, though!

I've been thinking of the `SPDX-License-Identifier:` wrapper as somewhat like the `package-content-state` "info.rkt" field: it's useful for tools, but it can be generated automatically, and, as package authors, we don't typically write `(define package-content-state '(source #f))` explicitly. In the case of `SPDX-License-Identifier:`, writing it explicitly does have the ancillary benefit of serving as a sort of documentation.

Concretely, I experimented with the [`reuse`](https://github.com/fsfe/reuse-tool) command-line tool. (I exclude this from "language-independent SPDX tools that I particularly want to run" because there would be a lot more to do to have a package meet the [REUSE Specification](https://reuse.software/spec/).) It has trouble with extra closing parentheses on the same line as the license declaration—I think that's a bug on their part, and there are other bugs—but it does recognize the SPDX-License-Identifier if written like this:
```racket
(define license
  '(SPDX-License-Identifier: (Apache-2.0 OR MIT)
   ))
```

A problem with something like `(define SPDX-License-Identifier: '(Apache-2.0 OR MIT))` is that the `'` would break the SPDX grammar.

> Similarly if not for that goal, we'd probably use prefixed s-expressions, not infix, like the rest of info.rkt?

In that case—and I could be persuaded that this the right approach—I'd be more inclined to just drop the `SPDX-License-Identifier:` from the grammar but keep the rest the same. If I consider a syntax like:
```racket
(define license
  '(and MIT (or (+ AFL-2.0) (with GPL-3.0-or-later Classpath-exception-2.0))))
```
It seems just different enough from the SPDX License Expression DSL to make me have to think about the translation. 

Practically, even if external tools would't detect the information in the info.rkt file without the `SPDX-License-Identifier:`, the compatibility is still useful to me as a Racket tool-writer: I can convert the license expression to a totally SPDX-compatable string using just:
```racket
(format "~s" '(MIT AND (AFL-2.0+ OR (GPL-3.0-or-later WITH Classpath-exception-2.0))))
```

The other package managers I listed above have a `license` field in their metadata that takes just the SPDX License Expression (i.e. no `SPDX-License-Identifier:` tag) as a string. They seem to call the field `license` rather than refer to SPDX in the field name because they have varying levels of support for licenses that are not on the SPDX license list, e.g. for packages that are not free and open source software.

Maybe another alternative would be to change the way the way this is presented in the docs. The grammar given first could be just:
```                                   
  <license-expr> = <license-id>                            
                 | (<license-id> WITH <exception-id>)
                 | (<license-expr> AND <license-expr>)               
                 | (<license-expr> OR <license-expr>)  
```

And later, the docs could say that, optionally, the `license-expr` may be wrapped in a list with  `SPDX-License-Identifier:`.

This grammar includes another change I might prefer anyway, using `license-id` rather than `versioned-license-id`, with the idea of presenting a note later about `+` later. My impression is that `+` is very rarely used in practice: you're no longer supposed to use it for GNU licenses (apparently their were discussions settling on `GPL-3.0-or-later` to address certain idiosyncrasies), and it doesn't really make sense either for non-copyleft licenses or for copyleft licenses like the Mozilla Public License, which always includes the equivalent of "or, at your option, any later version." I don't think I've ever seen `+` except in projects still using deprecated versions of the SPDX standard and license list (which originally did use e.g.  `GPL-3.0+`).

> I wonder if whoever does that work, someday, should have latitude to determine the grammar? e.g. Maybe parsing infix expressions isn't rocket science, but it's more work they need to do, and maybe they'd rather not. Or maybe there's some other consideration that only surfaces when trying to implement these things.

I'm hoping to write some of the kinds of tools you mention!

I've been thinking idly about a good way of encoding license metadata as part of generally trying to be a good license citizen. A particular impetus is that I've been working a bit (arguably too much) on adding Racket support to the Guix package manager, along the lines of Guix's support for the other package managers I listed above. It has a mandatory `license` field, which it would be nice to be able to fill in it automatically.

I'm also interested in a checker and support on pkgs.racket-lang.org.

One of my reasons for proposing the grammar before writing tools is that there seems to be a question of where such tools should live. A basic tool that just knows the grammar would be trivial, but most interesting tools would need the SPDX license list and exception list, either using the JSON files from https://github.com/spdx/license-list-data at runtime or perhaps converting them to `jsexpr?`s at compile time. It seems like that belongs in a package, not in core Racket, and probably not a package developed in this repository—except that it might be nice for it to potentially hook into `raco setup` somehow, like `--check-pkg-deps`. My inclination is that the package would be best somewhere under the Racket organization: all of this will be most useful if it has community buy-in and discoverability, so that package authors actually use it.

LiberalArtist:
@greghendershott I gave removing the optional `SPDX-License-Identifier:` wrapper a try, and I think I like this way better:

  - I think it makes the documentation simpler and clarifies the focus on the actual license expression.
  - If you're writing a Racket tool to add a `SPDX-License-Identifier:` wrapper when it was implicit, you could at least as easily just e.g. `(printf ";; SPDX-License-Identifier: ~s\n" '(Apache-2.0 OR MIT))`, which has the advantage of actually being recognized by tools that exist today.
  - Arguably, it may even be more correct: the `SPDX-License-Identifier:` tag is usually used to give the license information for a particular file, and the license(s) of the "info.rkt" file could in principle be a subset of the licenses that apply to the package as a whole.

The grammar is now:
```                                   
  <compound-expr> = <license-id>                            
                  | (<license-id> WITH <exception-id>)
                  | (<compound-expr> AND <compound-expr>)               
                  | (<compound-expr> OR <compound-expr>)  
```

There's an un-squashed version of these changes over at <https://github.com/LiberalArtist/racket/tree/spdx-info-rkt-hist-v1>.

LiberalArtist:
And here's an updated example:
```racket
(define license
  '(Apache-2.0 OR MIT))
```

greghendershott:
Sorry I didn't mean to do a drive-by comment and not reply later; I got sucked into other things.

These changes seem good!

The idea of a license being per-file is a good point. (Similarly it's not unusual to have copyright vary among source files.)

I'm not an IP lawyer. Would one advise, "You really ought to state the copyright and license in a comment in each and every source file"?

I guess the idea of the project/package-wide license is that it's:

1. A default to use in case any specific file lacks one.
2. It's a human's judgment of the most-permissive license compatible with all the per-file licenses.

Is that right?

LiberalArtist:
> Sorry I didn't mean to do a drive-by comment and not reply later; I got sucked into other things.

I likewise got sucked into other things, and I've also been ruminating over these questions:

> The idea of a license being per-file is a good point. (Similarly it's not unusual to have copyright vary among source files.)
> 
> I'm not an IP lawyer. Would one advise, "You really ought to state the copyright and license in a comment in each and every source file"?
> 
> I guess the idea of the project/package-wide license is that it's:
> 
>     1. A default to use in case any specific file lacks one.
> 
>     2. A human's judgment of the most-permissive license compatible with all the per-file licenses.

(N.B. I am also not any kind of lawyer.)

In brief, my answer is that the `license` field in the package's `info.rkt` file should answer the question, "what does the package's maintainer say the license is?"

As package maintainers, we often answer that question in prose in `README.md` or some similar place. This proposal lets us also answer the question in machine-readable form in `info.rkt`.

As package users, that answer isn't the end of all license-related questions. At the extreme, the package maintainer could be lying! Or even just mistaken. And there may be things you have to do to actually _comply_ with the applicable licenses, like preserving the verbatim text:

> In no event shall Brent Benson be liable to any party for direct, indirect, special, incidental, or consequential damages arising out of the use of this software and its documentation, even if Brent Benson has been advised of the possibility of such damage. … The software provided hereunder is on an "as is" basis, and Brent Benson has no obligation to provide maintenance, support, updates, enhancements, or modifications.

Still, it's useful to know what the package's maintainer says the license is. It could let you know that you're facing the kind of unfortunate situation about which [sincere lawyers](https://sfconservancy.org/blog/2016/feb/25/zfs-and-linux/) can come to [opposite conclusions](https://www.softwarefreedom.org/resources/2016/linux-kernel-cddl.html). I suspect a more likely scenario is discovering that someone forgot to put their package under any license at all.

Beyond that, I expect different kinds and amounts of legal scrutiny would be prudent under different sets of circumstances. (Again, this is not legal advice, which I, an non-attorney, would not be capable of giving if I wanted to.) If you are creating free, open-source software, maybe you're at the proverbial 90% solution to satisfying the license terms just by publishing all of your source code. On the other hand, if you're a large, wealthy corporation that would make an attractive target for litigation and that is releasing non-free software, maybe you want to retain counsel to thoroughly review all of the code you use.

Some organizations suggest that per-file licensing and automated auditing tools can help with some of these scenarios. In particular, the [REUSE Software](https://reuse.software/) standard and tools from the Free Software Foundation Europe advocate a very rigorous and structured way of recording that information. I guess there's some debate over the extent to which that is useful or not: apparently there was a [literal organized debate](https://archive.fosdem.org/2020/schedule/event/debate_license_compliance/) on this topic at FOSDEM 2020. (I have not (yet) watched much of it.)

But I don't think the usefulness of this proposal depends on which side is right in that debate. In practice, very few projects I've contributed to—maybe 1? maybe 0?—have per-file license and copyright information. Regardless of what one thinks the following steps should be, a structured, standardized way for a package's maintainers to communicate what they say its license is seems like a step in the right direction.

LiberalArtist:
I've continued to be happy with this syntax, so I've updated the documentation in light of this discussion and added a summary [above](https://github.com/racket/racket/pull/3760#issue-609522131). If there's consensus, I'd love for this to make it into 8.2.

@greghendershott and @jackfirth, since you commented on this earlier, I'd especially appreciate your thoughts.

sorawee:
I think it would be nice to change `raco pkg new` to add this new field as well.

jbclements:
Branch is today. That doesn’t mean it can’t make it in, but it does make it somewhat harder….

John

LiberalArtist:
Re @sorawee: Good catch! I've done it in https://github.com/racket/racket/pull/3760/commits/5c1fa41e65d078642fcb98e03d602ce04c34c802

Re @jbclements: Ah, that was an off-by-one error on my part.

sorawee:
I would love to see this in the 8.3 release. As I understand, 8.3 is scheduled for next month, so it would be nice to have this merged soon.

LiberalArtist:
The above-referenced pull requests should (I hope!) add license metadata to all packages in the release catalog except for the binary packages in <https://github.com/racket/libs>. (There are two non-formulaic ones that no one has reviewed yet: <https://github.com/racket/swindle/pull/1> and <https://github.com/racket/slatex/pull/1>.)

I've started to look at the binary packages, but I wanted to get feedback before I proceed about, basically, what level of detail seems useful—and it seems like <https://github.com/racket/libs> has issues disabled, so I'll post here.

Here's what I might put in the `info.rkt` file for `draw-aarch64-macosx-3`:
```racket
(define license
  '((Apache-2.0 OR MIT) ;; Racket code
    AND
    (MIT ;; libffi, most of pixman, part of fontconfig
     AND
     (MIT-Modern-Variant ;; HarfBuzz, part of fontconfig
      AND
      (LGPL-2.1-or-later ;; Glib, FriBidi, Pango, libintl
       AND
       (LGPL-2.0-or-later ;; some libintl files
        AND
        ((LGPL-2.1-only OR MPL-1.1) ;; Cairo
         AND
         (HPND-sell-variant ;; fontconfig, part of pixman
          AND
          (Unicode-DFS-2016 ;; part of fontconfig
           AND
           (FTL ;; part of pixman (may have been replaced)
            AND
            (Libpng ;; libpng (obviously)
             AND
             (IJG ;; libjpeg
              AND
              BSD-3-clause)))))))))))) ;; libuuid

;; additionally, fontconfig/src/fcmd5.h and 
;; fontconfig/src/ftglue.[ch] are placed in the public domain
;; using non-standardized language.
```

Does that seem like too much?

On the one hand, it seems like quite a lot to me: certainly it's more detail than I need for anything I have in mind. It could easily be collapsed to just `'(LGPL-2.1-only OR (Apache-2.0 OR MIT))`, or any point in between.

On the other hand, I think this is probably a very rare and extreme case. If we accept the near-absurdity here, we avoid any uncertainty later over whether we have summarized too much.

I'm leaning toward including all the details, but I'd be glad to hear other opinions.

PR #3761, opened 2021-04-06
Support pattern matching of stream and stream*

camoy:
This PR adds support for pattern matching on streams via the `stream` and `stream*` forms. (The change in `define-forms.rkt` is to break a cyclic dependency.)

cc @samth

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sequences.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sequences.scrbl
@@ -1152,15 +1152,19 @@ stream, but plain lists can be used as streams, and functions such as
 
  @history[#:added "8.0.0.12"]}
 
-@defform[(stream expr ...)]{
+@defform[(stream e ...)]{
   A shorthand for nested @racket[stream-cons]es ending with
-  @racket[empty-stream].
+  @racket[empty-stream]. As a match pattern, @racket[stream]
+  matches a stream with as many elements as @racket[e]s,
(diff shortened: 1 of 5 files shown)


pkgs/racket-doc/scribblings/reference/sequences.scrbl
```diff
@@ -1152,15 +1152,19 @@ stream, but plain lists can be used as streams, and functions such as
 
  @history[#:added "8.0.0.12"]}
 
-@defform[(stream expr ...)]{
+@defform[(stream e ...)]{
```
sorawee, line 1155: Why is `expr` changed to `e`? `expr` follows the [Notation for Syntactic Form](https://docs.racket-lang.org/reference/notation.html#%28part._.Notation_for_.Syntactic_.Form_.Documentation%29), so I think it's better to keep the original name.

camoy: It might make sense to keep as `expr`, but I changed it because now those can also be patterns too so I thought `expr` might be misleading.

samth: Typically the documentation for constructors doesn't worry about this; for example `list` or `cons`.

sorawee: A `deftogether` with one case for constructor and another case for pattern could also work, IMO.

PR #3766, opened 2021-04-08
rktio: fix output of newlines in text files in Windows

gus-massa:
Fix typo so the output files opened in `'text` mode in Windows use `CRLF` for newlines.


mflatt:
Oops - I should have looked more closely before mergin. I think the derived "io.scm" needs to be rebuilt for this to take effect, and also a test should be added, but I'll do that.

PR #3767, opened 2021-04-08
update openssl tests

rmculpepper:
This PR fixes problems running the openssl tests on Ubuntu 20.04, which includes a build of OpenSSL which increases the minimum accepted private key size and which disables TLS 1.0 and 1.1. With these changes, the tests run cleanly on Ubuntu 20.04.

There is a new script for generating the test certificates.


mflatt:
LGTM

PR #3768, opened 2021-04-09
doc: @racket[parameterizing] -> @racket[parameterize]-ing

sorawee:


rfindler:
I think we've been leaving the hyphen out in edits like this one?

PR #3776, opened 2021-04-11
generator documentation edit

maueroats:
Directly specify the meaning of the `formals` argument instead of refering to what it is not ("non-keyword, non-optional").


maueroats:
Wrong base, includes garbage commit.

PR #3777, opened 2021-04-12
expander: improve the error message for begin

### LINKED ISSUE #2696: Improve error message for missing last expression in begin
Missing an expression at the end of `define` body is a somewhat common problem (in StackOverflow):

- https://stackoverflow.com/questions/56568358/racket-scheme-nested-define
- https://stackoverflow.com/questions/16378703/a-strange-error-in-racket-no-expression-after-a-sequence-of-internal-definition
- https://stackoverflow.com/questions/56503105/how-to-fix-this-error-begin-possibly-implicit-no-expression-after-a-sequence
- https://stackoverflow.com/questions/17332236/differences-between-common-lisp-and-scheme-lexical-closures
- https://stackoverflow.com/questions/33403937/scheme-tail-recursion-bitson-count

A small example is:

```
(define (foo)
  1
  (define x 1))
```

which results in

```
begin (possibly implicit): no expression after a sequence of internal definitions in:
  (begin 2 (define x 1))
  2
  (define x 1)
```

While the error message makes sense at some level, it could be confusing (Is `2` an internal definition? Why are forms in `begin` shown twice -- one inside `begin` and one independently?).

I propose changing the error message to the following:

For a non-empty `begin` expression like `(begin 2 (define x 1))`:

```
begin (possibly implicit): required the last clause to be an expression; found:
 (define x 1) in:
 (begin 2 (define x 1))
```

For an empty `begin` expression like `(begin)`:

```
begin: required at least one expression
```

(The error message for `(define (foo))` is perfect already, I think. `(lambda ())` could be improved, but it's at least not confusing)

But if anyone has a better idea, feel free to suggest.

sorawee:
With this PR, the error message of:

```
#lang racket

(define (foo)
  1
  (define x 1))
```

becomes:

```
string:3:2: begin (possibly implicit): the last form is not an expression
  at: (define x 1)
  in: (begin 1 (define x 1))
```

in Racket, and

```
begin (possibly implicit): the last form is not an expression in: (define x 1)
```

in DrRacket.

Previously, it produces:

```
begin (possibly implicit): no expression after a sequence of internal definitions
  in: (begin 1 (define x 1))
  location...:
   t.rkt:3:0
   t.rkt:4:2
   t.rkt:5:2
```

in Racket, and 

```
begin (possibly implicit): no expression after a sequence of internal definitions in:
  (begin 2 (define x 1))
  2
  (define x 1)
```

in DrRacket, which is confusing.

Related to #2696


sorawee:
On the other hand, I just realized that blaming the last form might not be always accurate. E.g.,

```
(define (foo)
  (begin)
  (begin))
```

The fault might not be due to the last `(begin)`. I still find this new error message to be an improvement, however.

mflatt:
I would be happy with improvement here, but as you show, "the last form must be an expression" is not quite accurate. The last form could be `(begin)` or it could be something that expands to `(begin (define x 1) x)` (which has an expression, but isn't one).

How about "the last form is not an expression"? That would be true and suggestive, but without making a complicated claim about what the syntax needs to be.

B1Z0N:
Hello, is this going to proceed? Would be very grateful!

sorawee:
Thanks for the reminder! I already forgot I had this. Will work on it by the end of this week.

sorawee:
@mflatt This is ready for review. I will separately submit another PR to improve the documentation.

PR #3786, opened 2021-04-14
generator argument formals specified directly

maueroats:
Instead of specifying generator arguments "by analogy" or by saying what they are not, just write what they are.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sequences.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sequences.scrbl
@@ -1395,11 +1395,13 @@ values from the generator.
   @racket[#f] otherwise.
 }
 
-@defform[(generator formals body ...+)]{
-  Creates a @tech{generator}, where @racket[formals] is like the
-  @racket[formals] of @racket[case-lambda] (i.e., the
-  @racket[_kw-formals] of @racket[lambda] restricted to non-optional
-  and non-keyword arguments).
+@defform/subs[(generator formals body ...+)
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/sequences.scrbl
```diff
@@ -1395,11 +1395,13 @@ values from the generator.
...
-  @racket[formals] of @racket[case-lambda] (i.e., the
-  @racket[_kw-formals] of @racket[lambda] restricted to non-optional
-  and non-keyword arguments).
+@defform/subs[(generator formals body ...+)
+              ([formals (id ...)
+                        (id ...+ . rest-id)
+                        rest-id])]{
+  Creates a @tech{generator}, where @racket[formals] specify the arguments.
+  Keyword and optional arguments are not supported. This is the same as the
+  @racket[formals] of @racket[case-lambda] without any matching ocurring.
```
samth, line 1404: I would avoid the term "matching" here, since it's not used for this elsewhere.

maueroats: Is there techically correct term for how `case-lambda` decides which case to choose? Perhaps "without the cases"? This was one of the points that threw me on the first reading - I expected generator to include matching on different invocation patterns after reading the original docs.

samth: Maybe a phrase like "a single case-lambda clause" would be useful.

PR #3788, opened 2021-04-15
change json parser to be more strict about whitespace (matching the spec)

rfindler:
Looks like other languages get upset about whitespace characters and [this place](http://seriot.ch/parsing_json.php) also declares this to be an error, so let's think about changing it.

### DIFF
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -162,6 +162,9 @@ the @rfc for more information about JSON.
       "false42" (code:comment "invalid JSON text sequence")
       (λ () (read-json))))
 ]
+
+@history[#:changed "8.1.0.2" @list{Adjusted the whitespace handling to reject whitespace that isn't either
+           @racket[#\space], @racket[#\tab], @racket[#\newline], or @racket[#\return].}]
 }
 
 @defproc[(string->jsexpr [str string?] [#:null jsnull any/c (json-null)])
--- a/racket/collects/json/main.rkt
+++ b/racket/collects/json/main.rkt
@@ -188,7 +188,7 @@
           (read-char i)
           (skip-whitespace)]
          [(char-whitespace? ch)
-          (err "found whitespace that is not allowed by the json spec\n  char: ~s"
+          (err "found whitespace that is not allowed by the JSON specification\n  char: ~s"
                ch)]
          [else ch])]
       [else ch]))
--- a/racket/collects/json/main.rkt
+++ b/racket/collects/json/main.rkt
@@ -174,12 +174,23 @@
     (define-values [l c p] (port-next-location i))
     (raise-read-error (format "~a: ~a" who (apply format fmt args))
                       (object-name i) l c p #f))
+  (define (json-whitespace? ch)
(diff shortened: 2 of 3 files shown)


mfelleisen:
I thought you wanted to make this backwards compatible. ~~ Should the change be documented?

rfindler:
I'm not sure what is the right thing wrt to backwards compatibility here. Definitely if we know of code that depends on the non-spec behavior I'd want to continue supporting it, but in the absence of that knowledge, I'm less clear on what's the right thing to do.

rfindler:
And thanks for noticing the lack of documentation; pull request updated.

racket/collects/json/main.rkt
```diff
@@ -174,12 +174,23 @@
...
-      [(and (char? ch) (char-whitespace? ch))
-       (read-char i)
-       (skip-whitespace)]
+      [(char? ch)
+       (cond
+         [(json-whitespace? ch)
+          (read-char i)
+          (skip-whitespace)]
+         [(char-whitespace? ch)
+          (err "found whitespace that is not allowed by the json spec\n  char: ~s"
```
samth, line 191:
```suggestion
          (err "found whitespace that is not allowed by the JSON specification\n  char: ~s"
```

mfelleisen:
I could imagine exporting a parameter for switching from one whitespace parsing to another. 

Should we just merge and wait for anyone to yell?

pkgs/racket-doc/json/json.scrbl
```diff
@@ -162,6 +162,9 @@ the @rfc for more information about JSON.
       "false42" (code:comment "invalid JSON text sequence")
       (λ () (read-json))))
 ]
+
+@history[#:changed "8.1.0.2" @list{Adjusted the whitespace handling to reject whitespace that isn't either
```
mfelleisen, line 166: Should we point to json.org?

samth:
These JSON documents were already rejected by widely-used JSON parsers and would almost certainly have to be hand-generated, so I think they're very unlikely. I would be in favor of just doing this change and adjusting if someone has a problem.

LiberalArtist:
My vote would be to add an optional keyword argument like `#:permissive-whitespace?` with a default of the current behavior. If parsers are accepting `#\page` as whitespace, I'd bet someone out there is generating JSON with form-feed characters in it, and I'd hate to find out that I'm communicating with one of them by getting an exception.

(It still wouldn't be as bad as the time a Python dependency decided their JSON parser was too slow, so they switched to one written in C, and it segfaulted.)

rfindler:
@LiberalArtist my feeling is like @samth 's. Much more popular languages are going to reject those ones and having a permissive parser is also problematic as it means we won't realize that the JSON we have is going to be rejected by others until possibly too late (as people won't notice the flag).

That said, I generally very much agree with the basic sentiment behind your comment! I find this to be a special case.

LiberalArtist:
As I look at those results more closely, *none* of the parsers are simply accepting `#\page` as whitespace: a very few are accepting it in one of two specific locations, and none is even accepting it in both. So I'm persuaded, @rfindler.

mfelleisen:
Let's merge.

PR #3797, opened 2021-04-22
Supply original stx to make-import

greghendershott:
Although commit 5b0f0cee23 was sufficient to make `rename-in` expand
to `rename` with srcloc for the original name, it turns out it did not
address a similar issue with `only-in` renaming.

Do so by supplying the original (not renamed) identifier syntax as the
`orig-stx` for `make-import`.

### DIFF
--- a/racket/collects/racket/private/reqprov.rkt
+++ b/racket/collects/racket/private/reqprov.rkt
@@ -517,7 +517,7 @@
                                                               (import-mode import)
                                                               (import-req-mode import)
                                                               (import-orig-mode import)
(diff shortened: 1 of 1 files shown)


greghendershott:
(Context/motivation: As with commit 5b0f0ce and commit b7078dc272, the idea is to make sure there is srcloc for the original name in the fully expanded code (in the `rename` sub-forms of `#%require` and `#%provide`). That way, things like check-syntax could draw an arrow from the new to the old name, and tools that use the arrows would see the relation.)

PR #3798, opened 2021-04-23
Fix Errortrace Information for Invalid `->[*]` Contracts

shhyou:
# Fix Errortrace Information for Invalid `->[*]` Contracts
## The Issue

The source location of contract construction errors from `errortrace` points to the contract system instead of the invalid contract in the user's code. Here is an example.

```racket
#lang racket
(-> (lambda () #f)
    any)
```

```
->: contract violation
  ...
  errortrace...:
   ....../racket/contract/private/arrow-val-first.rkt:864:10: #%module-begin
```

The `->*` contract also has the same issue. Currently, the test case only checks the `errortrace` information of `->` at top-level instead of in the modules.

## Cause of the Issue

Here is what `->` expands to:

```racket
;; In racket/contract/private/arrow-val-first.rkt
#`(let #,let-bindings
    #,(quasisyntax/loc stx    ;; stx is the user's code
        (build-very-simple--> ...
```

`errortrace` tries to avoid annotating macro-generated code. Therefore when `->` appears in a module, the source location in the expanded code `(build-very-simple--> ...` is never picked up by `errortrace`.

The existing test case checks the error message of `(-> <invalid-contract> any)` at top-level. This passes because `errortrace` receives partially expanded result of `->` instead of the original source code.

## Fix A (This PR)

This PR simply duplicates the source location on the outermost syntax object.
Because the syntax properties on the original syntax object automatically propagates
to expansion result, the new source location is correctly retrieved by `errortrace`.

Nonetheless, the source location on the inner syntax object is still necessary for the top-level case.

```racket
;; In racket/contract/private/arrow-val-first.rkt
(quasisyntax/loc stx
  (let #,let-bindings
    #,(quasisyntax/loc stx    ;; stx is the user's code
        (build-very-simple--> ...
```

## Fix B

<https://github.com/shhyou/racket/commit/bce1331f2e5681dbd6bee933de6ccb3e7bb58cf4>
(incomplete)

Another approach is to forward the `errortrace` annotation on the original
syntax object (if any) to the `(build-very-simple-->` part that has the source
location information.

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/errortrace.rkt
+++ b/pkgs/racket-test/tests/racket/contract/errortrace.rkt
@@ -1,5 +1,45 @@
 #lang racket/base
-(require "test-util.rkt")
+(require "test-util.rkt"
+         racket/list)
+
+(define (test-errortrace-has-name #:name name
+                                  #:target-name [target-name name]
+                                  expr0 . exprs)
+  (define all-exprs (append (list expr0) exprs))
+  (define init-exprs (drop-right all-exprs 1))
+  (define expr (last all-exprs))
+  (define test-case-name (string->symbol (format "checking-arrow-src-locs.~a" name)))
+
+  (for ([index (in-naturals)]
+        [init-expr (in-list init-exprs)])
+    (define sp (open-input-string (format "~s\n" init-expr)))
+    (define init-stx (read-syntax (string->symbol (format "~a.~a" name index)) sp))
+    (close-input-port sp)
+    (contract-eval init-stx #:test-case-name test-case-name))
+
+  (define sp (open-input-string (format "~s\n" expr)))
+  (define stx (read-syntax name sp))
+  (define exn
+    (with-handlers ((exn:fail? values))
+      (contract-eval stx)))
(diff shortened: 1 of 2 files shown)


rfindler:
If I'm understanding properly, then I think that Fix B is probably not going to work well. Errortrace avoids annotation on code that is already compiled (which things in the contract system tend to be) so it won't see that property.

I think the approach in this PR is probably a good idea. Well, if it isn't, I'd like to know :) as I'm using it a lot in redex too!

shhyou:
Thanks!

I'm happy with either of the fix. As fix B also passes the tests, it should also work when the contract system itself is already compiled. To be more accurate, fix B is about adding `'errortrace:annotate` the output of the `->` macro if the source syntax object has that property. In other words, it explicitly marks the output of `->` as `errortrace`-able. Therefore whether the contract system itself is compiled does not affect the behavior.

Sadly this is only fixing the errors of the contracts themselves, not the error messages of contract violations.

rfindler:
Oh, I see what you mean. Indeed, I did confuse myself! Thanks.

PR #3799, opened 2021-04-24
fix(docs): corrects typos in project documentation

plan-do-break-fix:
Scope of changes is restricted to docs only.


gus-massa:
Thanks for the report. Merged in https://github.com/racket/racket/commit/0cdede88c379703d69be089695359cf8a0dde8a3

It's too late for the 8.1 release, but it will be included in the 8.2 release.

PR #3806, opened 2021-04-29
docs: add `define-match-expander` example

9999years:
The first example for `define-match-expander` destructures *syntax lists*, making it hard to tell (at least for me!) what parts are dealing with syntax because of the example is matching syntax lists and what parts are dealing with syntax because `define-match-expander` takes a syntax parser as input.

I've added an example which deals with value structs to hopefully make this a little bit clearer and a bit closer to the "minimal working example" of the macro.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -2,8 +2,10 @@
 @(require "mz.rkt" "match-grammar.rkt" racket/match)
 
 @(define match-eval (make-base-eval))
-@examples[#:hidden #:eval match-eval (require racket/match racket/list)]
-@examples[#:hidden #:eval match-eval (require (for-syntax racket/base))]
+@examples[#:hidden #:eval match-eval (require racket/match
+                                              racket/list
+                                              (for-syntax racket/base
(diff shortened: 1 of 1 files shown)


samth:
This is a nice example, but I think some explanatory text would be helpful. In particular, just describing what the two expanders do, and the structs are for would be enough.

9999years:
How's this language look?

> For example, we can use `define-match-expander` to add syntax sugar for destructuring structs, shown here with a result type that can contain either a failure or a success value. Note the [rename transformer](https://docs.racket-lang.org/reference/syntax-model.html#%28tech._rename._transformer%29) used as the `other-transformer` sub-expression so that `success` (and `failure`) can be used to construct `success` structs normally outside of patterns as well as destructuring within patterns.

I also renamed the sub-expressions so that we don't have to refer to them as "the first `proc-expr`" and "the second `proc-expr`".

9999years:
I no longer wish to be involved with the Racket community as long as Matthias Felleisen is involved, due to [his abusive conduct towards Matthew Butterick](https://beautifulracket.com/appendix/why-i-no-longer-contribute-to-racket.html) and the core team's tacit defense of him.

Issue #3805, opened 2021-04-28
lazy-require instantiates module twice


m4burns:
**What version of Racket are you using?**
8.0 [cs], 8.1.0.2 [cs] snapshot 20210427-6e3a94b6ac

**What program did you run?**
main.rkt:
```
#lang racket

(module bar racket
  (require racket/lazy-require)
  (provide bar)
  (lazy-require
    ["foo.rkt" (foo)])
  (define bar foo))

(module baz racket
  (require racket/lazy-require)
  (provide baz)
  (lazy-require
    ["foo.rkt" (foo)])
  (define baz foo))

(require (submod "." bar)
         (submod "." baz))

(for-each thread-wait
  (map thread (list baz bar)))
```

foo.rkt:
```
#lang racket

(provide foo)

(define (foo)
  (printf "hello~n"))
```

**What should have happened?**
```
$ racket main.rkt
hello
hello
```

**If you got an error message, please include it here.**
```
$ racket main.rkt
hello
define-values: assignment disallowed;
 cannot re-define a constant
  constant: foo
  in module:"/home/marc/foo.rkt"
  context...:
   body of "/home/marc/foo.rkt"
   /usr/local/racket/collects/racket/lazy-require.rkt:89:13: get-sym
   /usr/local/racket/collects/racket/promise.rkt:65:10
   /usr/local/racket/collects/racket/promise.rkt:45:2
   /usr/local/racket/collects/racket/lazy-require.rkt:106:6
```

**Please include any other relevant details**
Linux x86_64

It seems that calling two functions bound by `lazy-require` in a race on two threads can cause a module to be instantiated twice. I found this when trying to establish database connections in two distinct threads at roughly the same time - the connection constructors for each database implementation are `lazy-require`d.

m4burns:
When I encountered this, the error message was different:

```
instantiate-linklet: mismatch;
 reference to a variable that is not exported;
 possibly, bytecode file needs re-compile because dependencies changed
  name: maybe-add-state-undo
  exporting instance: "/usr/local/racket/collects/syntax/parse/private/residual.rkt"
  importing instance: "/usr/local/racket/collects/syntax/parse/private/lib.rkt"
  context...:
   body of "/usr/local/racket/collects/syntax/parse/private/lib.rkt"
   .../racket/lazy-require.rkt:89:13: get-sym
   /usr/local/racket/collects/racket/promise.rkt:65:10
   /usr/local/racket/collects/racket/promise.rkt:45:2
[...]

PR #3809, opened 2021-04-29
Fix lazy-require race with reentrant lock

m4burns:
This is a proposed fix for #3805 . The reentrant lock implementation is copied from `racket/src/expander/namespace/registry.rkt`.

This patch causes the example code in #3805 to work as expected and resolves the issue in our application.

There may be a contrived use of `lazy-require` that would deadlock with this patch:
1. Module A lazy-requires `b` from module B
2. Module B lazy-requires `c` from module C
3. During instantiation, module B starts and waits on a new thread that invokes `c`
4. Deadlock when `b` is invoked from module A

Questions:

- Does this approach make sense?
- Should the reentrant lock implementation be available in `racket/base` or somewhere else it could be required by both `registry.rkt` and `lazy-require.rkt`?
- Where should a test be added?

### DIFF
--- a/racket/collects/racket/lazy-require.rkt
+++ b/racket/collects/racket/lazy-require.rkt
@@ -97,10 +97,44 @@
                 (syntax-property (quote-syntax aux-name)
                                  'not-provide-all-defined #t))) ...)))]))
 
+(struct reentrant-lock (box))
+
+(define (make-reentrant-lock) (reentrant-lock (box #f)))
+
+(define (call-with-reentrant-lock r proc)
+  (define lock-box (reentrant-lock-box r))
+  (let loop ()
+    (define v (unbox lock-box))
+    (cond
+     [(or (not v)
+          (sync/timeout 0 (car v) (or (weak-box-value (cdr v)) never-evt)))
+      (define sema (make-semaphore))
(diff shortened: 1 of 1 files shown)


mflatt:
I think this is the right idea, but the lock should be the same one that's already in place for a namespace and used by on-demand instantiation of ready modules. That lock is specific to a module registry, which also seems like a good idea here.

This begs the question of whether `dynamic-require` should use the lock automatically, not to mention all the other functions that resolve module paths or otherwise affect the namespace. That would be a big change, and I'm not completely sure it would work out. On-demand things like resolving lazy requires and instantiating ready modules seem like a special catgory that particularly benefits from a lock.

So, I'm adding `namespace-call-with-registry-lock`, and I'll update `lazy-require` to use it mostly as here (but more tightly around `dynamic-require`). We'll see how that works.

m4burns:
Thanks @mflatt ! I did try taking out a reentrant lock, distinct from the namespace lock, in the body of `dynamic-require`. This resulted in deadlock during the initial `raco setup` multithreaded build.

By the way, the CS rewrite is excellent. It was easy to find the implementation of `dynamic-require` and the existing reentrant lock.

m4burns:
Alright, d9c128fe656516777b7d653ab2d18519f6603e4f resolves the issue in our application. Thanks again @mflatt !

PR #3814, opened 2021-05-03
cs,thread: fix `thread-receive-evt` retry proc

### LINKED ISSUE #3813: cs,thread: `thread-receive-evt` crashes Racket after `thread-resume`
9e2f50e11e

**What program did you run?**

```racket
#lang racket/base

(require racket/match)

(define thd #f)

(define (maybe-start!)
  (unless thd
    (set! thd (thread/suspend-to-kill
               (lambda ()
                 (let loop ()
                   (sync
                    (handle-evt
                     (thread-receive-evt)
                     (lambda (_)
                       (match (thread-receive)
                         [`(marco ,ch)
                          (channel-put ch 'polo)
                          (loop)]))))))))))

(define (test)
  (define cust (make-custodian))
  (parameterize ([current-custodian cust])
    (maybe-start!)
    (thread-wait
     (thread
      (lambda ()
        (thread-resume thd (current-thread))
        (define ch (make-channel))
        (thread-send thd (list 'marco ch))
        (displayln (channel-get ch)))))
    (custodian-shutdown-all cust)))

(test)
(test)
```

The output should be:

```
polo
polo
```

instead, it's:

```
polo
result arity mismatch;
 expected number of values not received
  expected: 2
  received: 1
  in: local-binding form
  arguments...:
   #<void>
internal error: terminated in atomic mode!
```

If I replace my usages of the thread mailbox with a channel, everything works fine, but I would have expected the mailbox to work in this case, too.

Bogdanp:
Fixes #3813

### DIFF
--- a/pkgs/racket-test-core/tests/racket/sync.rktl
+++ b/pkgs/racket-test-core/tests/racket/sync.rktl
@@ -1041,6 +1041,23 @@
 ;; make sure it's ok for rewind to be the first action:
 (test (void) thread-wait (thread (lambda () (thread-rewind-receive '(1 2 3)))))
 
+(let* ([t (thread/suspend-to-kill
+           (lambda ()
+             (let loop ()
+               (sync
+                (handle-evt
+                 (thread-receive-evt)
+                 (λ (_)
+                   (channel-put (thread-receive) 'ok)
+                   (loop)))))))]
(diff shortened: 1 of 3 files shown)


mflatt:
Thanks!

PR #3818, opened 2021-05-06
Improve arithmetic using code from Gambit.

### LINKED ISSUE #3816: Recommend ratnum algorithms from Section 4.5.1 of Knuth's TAOCP
This isn't really a bug, just a recommendation.

Welcome to Racket v8.1.0.5 [cs].

**What program did you run?**
Please include a short example that triggers the bug
```
heine:~/programs/racket> /usr/local/racket/bin/mzscheme
Welcome to Racket v8.1.0.5 [cs].
> (define (phi n)
    (do ((i 0 (+ i 1))
         (r 1 (+ 1 (/ r))))
        ((= i n) r)))
> (define a (time (phi 100000)))
cpu time: 10142609 real time: 10152424 gc time: 40
```

It can run quite a bit faster.

Gambit's ratnum routines are based on the basic algorithms from Section 4.5.1 of Knuth's TAOCP.  They're written in Scheme, are fairly short, and should be easy to adapt to the new Racket CS runtime.  If you want to use them as is (which seems possible, with Chez's naming convention, etc.), please do.  They start at

https://github.com/gambit/gambit/blob/ead2ba415b5c4cae9bb0f4505e3e359a97f00ad7/lib/_num.scm#L10722

The same program in Gambit is
```
heine:~/programs/racket> gsi
Gambit v4.9.3-1390-g9267589b

> (define (phi n)
  (do ((i 0 (+ i 1))
       (r 1 (+ 1 (/ r))))
      ((= i n) r)))
> (define a (time (phi 100000)))
(time (phi 100000))
    0.363497 secs real time
    0.363390 secs cpu time (0.355362 user, 0.008028 system)
    184 collections accounting for 0.027910 secs real time (0.027315 user, 0.000600 system)
    132009416 bytes allocated
    2579 minor faults
    no major faults
```
Brad

samth:
Closes #3816.


samth:
Unfortunately, just using an un-normalized ratnum appears to cause infinite loops. I'll try that a different way.

mflatt:
Does it make sense to add a few tests in "5_3.ms", or do the tests already there cover the interesting cases? I'm happy for more/most tests to be in the Racket test suite, but basic code coverage is a good goal for the Chez Scheme tests.

For the Racket tests, I recommend loading "gambit-number.rktl" from "number.rktl".

Otherwise, this looks good as far as I can tell.

samth:
Also the code isn't quite right yet, `(lcm 1/2 3)` hangs.

samth:
The only remaining bugs are sin/tan in Racket BC.

samth:
That should fix all the bugs, but I'm having trouble getting the float/double distinction write in `scheme_complex_sin`. @mflatt, am I missing something obvious?

samth:
I think [this xform failure](https://github.com/racket/racket/pull/3818/checks?check_run_id=2549424047#step:4:5470) on windows was because I didn't declare `sinh`, but I've found the place.

mflatt:
I don't think that has anything to do with `sinh`. It looks more like it's because of `signbit`, which expands in a way that xform doesn't recognize.

Instead of `signbit`, use `scheme_minus_zero_p`.

samth:
This should be done (assuming tests pass everywhere).

PR #3823, opened 2021-05-07
Fix typo in base64.scrbl

gus-massa:
A recent commit changed the name of the argument but missed one spot.

Anyway, I'm worried because the in the fixed version this will link to the docs of the function `newline` instead of to nowhere or to the position of the argument.


mflatt:
LGTM. The identifier `newline` should not be mislinked, since that part of Scribble uses normal scoping.

PR #3824, opened 2021-05-07
use unaccented e in eof/length

dalev:
Fix a type-o in the signature of `get-string-n!`.


mfelleisen:
(Hi Dale, what a surprise. Is Jane St. using Chez/Racket now :-)

dalev:
Hi Matthias!  Ha, not to my knowledge, just playing around a bit with the shiny CS-powered parentheses here at home :)

mflatt:
Thanks!

mflatt:
No action is needed on your part. I'm not sure whether there's something wrong with that test, or whether it's just that a package changed (because DrDr doesn't notice the difference between a package changing and the main repo changing).

Thanks for checking!

dalev:
Got it, thanks!  (I guess the "blame theorem" hasn't yet been extended to github/automated tests workflow @mfelleisen ;)

PR #3826, opened 2021-05-08
add warning about regexp-match-exact?

rmculpepper:
This subtlety just bit me again, so here's a reminder for the docs.

I assume this is the intended behavior, based on section 9.7 (Alternation) of the Racket Guide. The behavior seems to be consistent between CS and BC.


mflatt:
I like this improvement, but I think it also points to a bug in the initial description:

> Like regexp-match?, but #t is only returned when the entire content of input matches pattern.

Probably that should say

> Like regexp-match?, but #t is only returned when the first found match is to the entire content of input.

along with your addition.

PR #3829, opened 2021-05-11
Fix warning about `git checkout HEAD^2`

pmatos:
Currently CI is warning that this step is no longer needed, so remove
it.


pmatos:
Warning here: https://github.com/racket/racket/runs/2550937182?check_suite_focus=true#step:4:7

samth (approved):

Issue #3835, opened 2021-05-19
`inclusive-range` error inside `for/list`


tomyun:
**What version of Racket are you using?**
8.1 [cs]

**What program did you run?**
```
> (define n 10)

> (for/list ([i (range 0 n)]) i)
'(0 1 2 3 4 5 6 7 8 9)

> (for/list ([i (inclusive-range 0 n)]) i)
; add1: contract violation
;   expected: number?
;   given: 'n
; [,bt for context]
```

**What should have happened?**
Expected something like below.
```
> (for/list ([i (inclusive-range 0 10)]) i)
'(0 1 2 3 4 5 6 7 8 9 10)
```

gus-massa:
Thank for the report.

I pushed a fix https://github.com/racket/racket/commit/aa7a5b9509dfd31fe08ee8a966695c548a16aa79

PR #3836, opened 2021-05-19
Fix in-inclusive-range when last bound is not an explicit number

gus-massa:
Fix bug reported in https://github.com/racket/racket/issues/3835


pkgs/racket-test-core/tests/racket/for.rktl
```diff
@@ -16,16 +16,24 @@
...
+(test-sequence [(5 6)] (in-range (five) 7))
+(test-sequence [(3 4)] (in-range 3 (five)))
+(test-sequence [(0 5)] (in-range 0 10 (five)))
 (test-sequence [(3.0 4.0 5.0)] (in-range 3.0 6.0))
 (test-sequence [(3.0 3.5 4.0 4.5 5.0 5.5)] (in-range 3.0 6.0 0.5))
 (test-sequence [(3.0 3.1 3.2)] (in-range 3.0 3.3 0.1))
 (test-sequence [(6 7)] (in-inclusive-range 6 7))
 (test-sequence [(3 4 5 6)] (in-inclusive-range 3 6))
 (test-sequence [(7 6 5 4)] (in-inclusive-range 7 4 -1))
+(test-sequence [(5 6 7)] (in-range (five) 7))
```
sorawee, line 34: I think you meant this to be `in-inclusive-range`, right?

gus-massa: Fixed. Thanks!

PR #3837, opened 2021-05-19
Fix error formatting for non-empty-listof contract

camoy:
The error message for `non-empty-listof` had extra quotes due to a duplicated call to `format`:

```
; foo: broke its own contract
;   promised: "(and/c list? pair?)"
;   produced: '()
```

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/list.rkt
+++ b/pkgs/racket-test/tests/racket/contract/list.rkt
@@ -47,6 +47,12 @@
   (test/pos-blame 
    'nelistof5
    '(contract (non-empty-listof integer?) (list #f #t) 'pos 'neg))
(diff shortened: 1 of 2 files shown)


samth:
Is it possible that this failure is caused by these changes: https://github.com/racket/racket/pull/3837/checks?check_run_id=2623685999#step:9:42

camoy:
Looks like the test that failed is a timing test on a contract that uses `cons/c`. This change only pertains to `non-empty-listof` so it should have no impact. Perhaps this is a flaky test that needs a bit more generous timeout?

rfindler:
May I have a test case, please? Or if you prefer I can add one.

rfindler:
Thanks!

PR #3838, opened 2021-05-19
hash fns: use bad-pos raise-argument-error arg

stchang:
Is there a reason not to use the `bad-pos` version of `raise-argument-error` in Racket CS?

Because without it, these fns print a slightly different err msg in Racket CS compared to Racket BC.

### DIFF
--- a/racket/src/cs/rumble/hash.ss
+++ b/racket/src/cs/rumble/hash.ss
@@ -129,7 +129,7 @@
          (let ([ht (impersonator-val ht)])
            (mutable-hash? ht)))
     (impersonate-hash-set! ht k v)]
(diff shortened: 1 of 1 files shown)


stchang:
Does anyone object to this change? Otherwise I will merge.

mfelleisen:
Go for it!

mflatt:
LGTM.

In general, CS error messages rarely try to show "other arguments" in the way that BC does. Partly, that's because the information is less available for primitives that are supplied by Chez Scheme directly. Partly, it's just because it's a pain and doesn't seem to matter all that much. But it makes sense to improve the ones that are easy like this.

PR #3847, opened 2021-05-24
patterns.scrbl:  Clarified phase 0 use of require for syntax-parser

dstorrs:
Modified the documentation at https://docs.racket-lang.org/syntax/stxparse-patterns.html?q=syntax-parser#%28part._.Ellipsis-head_.Patterns%29 to mention the need for (require syntax/parse) when using syntax-parser instead of (require (for-syntax syntax/parse))


sorawee:
I think it's odd to note about `(require syntax/parse)` there. There are uses of `syntax-parse` at run-time throughout that page (the earliest one is [here](https://docs.racket-lang.org/syntax/stxparse-patterns.html?q=syntax-parser#%28part._.Single-term_.Patterns%29) IIUC). If you really want to make the change, I think the note should be at the top of the page. You can also use margin note.

Issue #3839, opened 2021-05-20
Memory leaks found by asan


samth:
The address sanitizer CI builds don't find any errors, but do find memory leaks. You can see a recent run here: https://github.com/racket/racket/suites/2790555937/artifacts/61953948 (that's a zip file with a bunch of text files listing leaks). All of them are here; I don't think it changes much run-to-run: https://github.com/racket/racket/actions/workflows/ci-asan.yml

Here's a representative set of results. This one has a lot of leaks from inside sqlite somewhere, and another signficant leak of resources allocated by iconv from `rktio_converter_open`.

```

=================================================================
==6824==ERROR: LeakSanitizer: detected memory leaks

Direct leak of 1624 byte(s) in 29 object(s) allocated from:
    #0 0x7fcb11d99330 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x55da74318332 in S_make_mutex (/__w/racket/racket/racket/bin/racket+0x57332)
    #2 0x40841032  (<unknown module>)

Direct leak of 864 byte(s) in 18 object(s) allocated from:
    #0 0x7fcb11d99330 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x55da7431ac52 in S_make_condition (/__w/racket/racket/racket/bin/racket+0x59c52)
    #2 0x408413d2  (<unknown module>)

Direct leak of 256 byte(s) in 1 object(s) allocated from:
    #0 0x7fcb11d99330 in __interceptor_malloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe9330)
    #1 0x55da745e3c77 in rktio_get_current_directory (/__w/racket/racket/racket/bin/racket+0x322c77)
    #2 0x42d38cf2  (<unknown module>)

Direct leak of 72 byte(s) in 3 object(s) allocated from:
    #0 0x7fcb11d99518 in calloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe9518)
    #1 0x55da745e7745 in rktio_system_fd (/__w/racket/racket/racket/bin/racket+0x326745)
    #2 0x42d323eb  (<unknown module>)

Direct leak of 48 byte(s) in 2 object(s) allocated from:
    #0 0x7fcb11d99518 in calloc (/usr/lib/x86_64-linux-gnu/libasan.so.5+0xe9518)
[...]

PR #3853, opened 2021-05-27
Ignore some objects in LeakSanitizer.

samth:
Really this should ignore fewer than this, but it's not yet clear
where the calls should go.

Related to #3839.

### DIFF
--- a/racket/src/rktio/lsan-suppressions.supp
+++ b/racket/src/rktio/lsan-suppressions.supp
@@ -1,0 +1,15 @@
+# LeakSanitizer suppressions for Racket CS.
+# These are intentional, process-lifetime allocations that are referenced
+# only from the Chez Scheme GC heap (which LSan does not scan), so LSan
+# reports them as leaks. Each matches a frame in the allocation stack.
+
+# --- rktio persistent state (held by the rktio_t / Racket) ---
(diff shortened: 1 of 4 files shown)


samth:
This is effectively doing the same as having a suppression file for these function calls, which isn't really what we want. But I don't know where the right place to make calls to `__lsan_ignore_object()` is, especially since I think the objects are "leaked" inside Racket code.

samth:
Another problem is that building with ASan doesn't work on an recent ubuntu install; I had to do this development inside a docker container using the `racket/racket-ci` docker image.

mflatt:
The fundamental misatch is related to roots, so I'd expect the right change to be using `__lsan_register_root_region` (maybe just before exiting) in Chez Scheme, not changing rktio.

samth:
Where would I call that (and with what sorts of arguments)?

mflatt:
I had imagined adding an exit callback that registers all GC pages. But on further reflection, pointers to C objects are likely to be represented as fixnums or bignums in a way that won't look like a pointer to LeakSanitizer.

So, maybe a better strategy is to have some way (maybe an environment variable) to change exit to force a custodian shutdown. Then, we'd have to deal with whatever is left.

PR #3857, opened 2021-05-30
"any" nested lambda, lexical scope

jasonhemann:
Someone parsing these words closely might misunderstand this sentence in isolation to mean literally *any* nested lambda.


mflatt:
I think the sentence should pin down which nested `lambda` forms are appropriate. How about "any nested `lambda` forms in the body"?

jasonhemann:
That's a further improvement! I was actually wondering about `(define y (λ (x) (λ (x) x)))`: is this section presuming Barendregt naming convention?

jasonhemann:
I suppose it's not terrible as written.

mflatt:
Well, I thought you had a good point about substitution, but hadn't gotten back to another suggestion.

How about a footnote or margin note to explain the assumption that different bindings have different names, where it's effectively the job of the macro expander to make that true, or something like that?

PR #3863, opened 2021-05-30
Copyediting syntax model in Reference

jasonhemann:
I make some small edits to the syntax-model section of the Racket reference, up through 1.2.3.5.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/syntax-model.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax-model.scrbl
@@ -311,7 +311,7 @@ syntactic-form names refer to the bindings defined in
 
 In a fully expanded program for a namespace whose @tech{base phase} is
 0, the relevant @tech{phase level} for a binding in the program is
-@math{N} if the bindings has @math{N} surrounding
+@math{N} if the binding has @math{N} surrounding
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/syntax-model.scrbl
```diff
@@ -520,16 +520,16 @@ core syntactic forms are encountered:
  @item{When a @racket[define], @racket[define-values],
        @racket[define-syntax], or @racket[define-syntaxes] form is
        encountered at the top level or module level, a binding is
-       added @tech{phase level} 0 (i.e., the @tech{base environment}
+       added to @tech{phase level} 0 (i.e., the @tech{base environment}
```
sorawee, line 523: I think the original one makes more sense to me.

mfelleisen: IMO, this is a question of grammar not understanding.

sorawee: Ah, my bad! I missed the word "is" and read the sentence as "a binding added to phase level", which didn't make sense to me. But with "is", adding "to"/"into" is definitely the right thing to do.

PR #3869, opened 2021-06-07
rename LICENSE -> LICENSE.txt

LiberalArtist:
The top-level LICENSE file seems to have been missed in the great
renaming of <https://github.com/racket/racket/commit/ecf9054>.

In addition, there are some subtle differences between `LICENSE.txt` and `racket/src/LICENSE.txt` and more notable ones between those two and `pkgs/racket-index/scribblings/main/license.scrbl`. I wonder if it would make sense to generate the `LICENSE.txt` files along the lines of `build.md`. For example:

  - https://github.com/racket/racket/blob/106545343a371271be0efad72193a090ae9f553b/racket/src/LICENSE.txt#L7-L8
    vs.
    https://github.com/racket/racket/blob/106545343a371271be0efad72193a090ae9f553b/LICENSE#L6-L7

  - https://github.com/racket/racket/blob/106545343a371271be0efad72193a090ae9f553b/LICENSE#L41
    vs.
    https://github.com/racket/racket/blob/106545343a371271be0efad72193a090ae9f553b/racket/src/LICENSE.txt#L42

  - Different stray references to the Boehm GC were left in each set of files. The `LICENSE.txt` files have (modulo relative paths): https://github.com/racket/racket/blob/106545343a371271be0efad72193a090ae9f553b/racket/src/LICENSE.txt#L55-L57
    while `license.scrbl` has:
    https://github.com/racket/racket/blob/106545343a371271be0efad72193a090ae9f553b/pkgs/racket-index/scribblings/main/license.scrbl#L47-L49


mflatt:
We left `LICENSE` without `.txt` on the grounds that GitHub recognizes the `LICENSE` filename specifically. Does GitHub treat `LICENSE.txt` the same as `LICENSE`?

LiberalArtist:
I think so. [Github says here](https://docs.github.com/en/github/creating-cloning-and-archiving-repositories/creating-a-repository-on-github/licensing-a-repository#determining-the-location-of-your-license):

> Most people place their license text in a file named `LICENSE.txt` (or `LICENSE.md` or `LICENSE.rst`) in the root of the repository

I just tried adding a `LICENSE.txt` to https://github.com/LiberalArtist/racket-config-tethered-experiment, and it shows the "view license" button on the right.

Is there any other special treatment?

sorawee:
AFAICT, GitHub supports both `LICENSE` and `LICENSE.txt`. But FWIW, if you start a new repo with a license file, it will be named `LICENSE`, so perhaps `LICENSE` is more widely used.

mflatt:
It's possible that we kept `LICENSE` just because it's more common, and maybe @samth remembers more. In any case, I'd be happy with renaming.

samth:
I don't have any memory of why we made that choice in particular, but I'm also ok with renaming.

The idea of centralizing that text seems like an excellent idea as well.

mflatt:
I didn't try to centralize the text, but I corrected (hopefully!) the noted problems.

PR #3870, opened 2021-06-08
make-temporary-file: accept keyword arguments

LiberalArtist:
When `#:base-dir` and `#:copy-from` are passed as keyword arguments,
rather than by position, `make-temporary-file` can still use source
location information to generate a template, rather than having to
fall back to `"rkttmp~a"`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
@@ -1,5 +1,6 @@
 #lang scribble/doc
 @(require "mz.rkt"
+          scribble/core
           (for-label framework/preferences
                      racket/runtime-path
                      launcher/launcher
@@ -680,7 +681,7 @@ directory.}
 
 @defparam*[current-directory-for-user path path-string? (and/c path? complete-path?)]{
 
-Like @racket[current-directory], but use only by
+Like @racket[current-directory], but for use only by
 @racket[srcloc->string] for reporting paths relative to a
 directory.
 
@@ -1390,43 +1391,125 @@ will not create it.
 
 
 @defproc[(make-temporary-file [template string? "rkttmp~a"]
-                              [copy-from-filename (or/c path-string? #f 'directory) #f]
-                              [directory (or/c path-string? #f) #f])
-         path?]{
+                              [#:copy-from copy-from (or/c path-string? #f 'directory) #f]
+                              [#:base-dir base-dir (or/c path-string? #f) #f]
+                              [compat-copy-from (or/c path-string? #f 'directory) copy-from]
+                              [compat-base-dir (or/c path-string? #f) base-dir])
+         (and/c path? complete-path?)]{
 
-Creates a new temporary file and returns a pathname string for the
-file.  Instead of merely generating a fresh file name, the file is
+Creates a new temporary file and returns its path.
+Instead of merely generating a fresh file name, the file is
 actually created; this prevents other threads or processes from
 picking the same temporary name.
 
-The @racket[template] argument must be a format string suitable
-for use with @racket[format] and one additional string argument (where
-the string contains only digits). If the resulting string is a
-relative path, it is combined with the result of
-@racket[(find-system-path 'temp-dir)], unless @racket[directory] is
-provided and non-@racket[#f], in which case the
-file name generated from @racket[template] is combined with
-@racket[directory] to obtain a full path.
+The @racket[template] argument must be a format string
+suitable for use with @racket[format] and one additional
+string argument (which will contain only digits). By
+default, if @racket[template] produces a relative path, it
+is combined with the result of
+@racket[(find-system-path 'temp-dir)] using
+@racket[build-path]; alternatively, @racket[template] may
+produce an absolute path, in which case
+@racket[(find-system-path 'temp-dir)] is not consulted. If
+@racket[base-dir] is provided and non-@racket[#false],
+@racket[template] must not produce a @tech{complete} path,
+and @racket[base-dir] will be used instead of
+@racket[(find-system-path 'temp-dir)]. Using
+@racket[base-dir] is generally more reliable than including
+directory components in @racket[template]: it avoids subtle
+bugs from manipulating paths as string and eleminates the
+need to sanitize @racket[format] escape sequences.
 
-The @racket[template] argument's default is only the string @racket["rkttmp~a"]
-when there is no source location information for the callsite of
-@racket[make-temporary-file] (or if @racket[make-temporary-file] is
-used in a higher-order position). If there is such information, then the template
-string is based on the source location.
+On Windows, @racket[template] may produce an absolute path
+which is not a complete path (see @secref["windowspaths"])
+when @racket[base-dir] is absent or @racket[#f] (in which
+case it will be resolved relative to
+@racket[(current-directory)]) or if @racket[base-dir] is a
+drive specification (in which case it will be used as with
+@racket[build-path]). If @racket[base-dir] is any other kind
+of path, it is an error for @racket[template] to produce an
+absolute path.
 
-If @racket[copy-from-filename] is provided as path, the temporary file
+When the @racket[template] argument is not provided, if
+there is source location information for the callsite of
+@racket[make-temporary-file], a template string is generated
+based on the source location: the default is
+@racket["rkttmp~a"] only when no source location information
+is available (e.g@._ if @racket[make-temporary-file] is used
+in a higher-order position).
+
+If @racket[copy-from] is provided as path, the temporary file
 is created as a copy of the named file (using @racket[copy-file]). If
-@racket[copy-from-filename] is @racket[#f], the temporary file is
-created as empty. If @racket[copy-from-filename] is
-@racket['directory], then the temporary ``file'' is created as a
-directory.
+@racket[copy-from] is @racket[#f], the temporary file is
+created as empty. As a special case, for backwards compatibility,
+if @racket[copy-from] is @racket['directory],
+then the temporary ``file'' is created as a directory:
+for clarity, prefer @racket[make-temporary-directory] for creating
+temporary directories.
 
 When a temporary file is created, it is not opened for reading or
-writing when the pathname is returned. The client program calling
+writing when the path is returned. The client program calling
 @racket[make-temporary-file] is expected to open the file with the
 desired access and flags (probably using the @racket['truncate] flag;
 see @racket[open-output-file]) and to delete it when it is no longer
-needed.}
+needed.
+
+The by-position arguments @racket[compat-copy-from] and
+@racket[compat-base-dir] are for backwards compatibility:
+if provided, they take precedence over the @racket[#:copy-from] and
+@racket[#:base-dir] keyword variants.
+Supplying by-position arguments prevents @racket[make-temporary-file]
+from generating a @racket[template] using the source location.
+
+@history[
+ #:changed "8.4.0.3"
+ @elem{Added the @racket[#:copy-from] and @racket[#:base-dir] arguments.}
+ ]}
+
+@defproc[(make-temporary-directory [template string? "rkttmp~a"]
+                                   [#:base-dir base-dir (or/c path-string? #f) #f])
+         (and/c path? complete-path?)]{
+
+ Like @racket[make-temporary-file], but
+ creates a directory, rather than a regular file.
+
+ As with @racket[make-temporary-file], if the
+ @racket[template] argument is not provided, a template
+ string is generated from the source location of the call to
+ @racket[make-temporary-directory] when possible: the default
+ is @racket["rkttmp~a"] only when no source location
+ information is available.
+
+@history[
+ #:added "8.4.0.3"
+ ]}
+
+@deftogether[
+ (@defproc[(make-temporary-file* [prefix bytes?]
+                                 [suffix bytes?]
+                                 [#:copy-from copy-from (or/c path-string? #f) #f]
+                                 [#:base-dir base-dir (or/c path-string? #f) #f])
+           (and/c path? complete-path?)]
+   @defproc[(make-temporary-directory* [prefix bytes?]
+                                       [suffix bytes?]
+                                       [#:base-dir base-dir (or/c path-string? #f) #f])
+            (and/c path? complete-path?)])]{
+
+ Like @racket[make-temporary-file] and
+ @racket[make-temporary-directory], respectively, but, rather
+ than using a template for @racket[format], the path is based
+ on @racket[(bytes-append prefix generated suffix)], where
+ @racket[generated] is a byte string chosen by the
+ implementation to produce a unique path. If there is source
+ location information for the callsite of
+ @racket[make-temporary-file*] or
+ @racket[make-temporary-directory*], @racket[generated] will
+ incorporate that information. The resulting path is combined
+ with @racket[base-dir] as with @racket[make-temorary-file].
+
+ @history[
+ #:added "8.4.0.3"
+ ]}
 
 @defproc[(call-with-atomic-output-file [file path-string?] 
                                        [proc ([port output-port?] [tmp-path path?]  . -> . any)]
--- a/pkgs/racket-test-core/tests/racket/file.rktl
+++ b/pkgs/racket-test-core/tests/racket/file.rktl
@@ -352,14 +352,211 @@
 (err/rt-test (open-output-file (build-path (current-directory) "baddir" "x"))
 	    exn:fail:filesystem?)
 
+;; Tests for make-temporary-{file,directory}{,*}
+
+;; arities
+(arity-test make-temporary-file 0 3)
+(test 'make-temporary-file object-name make-temporary-file)
+(let-values ([(required accepted) (procedure-keywords make-temporary-file)])
+  (test '() 'make-temporary-file-no-required-keywords required)
+  (test '(#:base-dir #:copy-from) 'make-temporary-file-accepts-keywords accepted))
+(arity-test make-temporary-file* 2 2)
+(test 'make-temporary-file* object-name make-temporary-file*)
+(let-values ([(required accepted) (procedure-keywords make-temporary-file*)])
+  (test '() 'make-temporary-file*-no-required-keywords required)
+  (test '(#:base-dir #:copy-from) 'make-temporary-file*-accepts-keywords accepted))
+(arity-test make-temporary-directory 0 1)
+(test 'make-temporary-directory object-name make-temporary-directory)
+(let-values ([(required accepted) (procedure-keywords make-temporary-directory)])
+  (test '() 'make-temporary-directory-no-required-keywords required)
+  (test '(#:base-dir) 'make-temporary-directory-accepts-keywords accepted))
+(arity-test make-temporary-directory* 2 2)
+(test 'make-temporary-directory* object-name make-temporary-directory*)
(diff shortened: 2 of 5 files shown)


samth:
This might be a good opportunity to make an explicit `#:directory?` optional argument.

LiberalArtist:
Maybe there's a better name for `#:copy-from`—something along the lines of `#:initialize`, maybe?

I used `#:base-dir`, which is what the formal argument was called in the implementation, rather than `directory`, which was the name previously used in the documentation, because having a `#:directory` argument that was *not* the one that could be `'directory` seemed especially confusing. Probably `#:directory?` would be less confusing, but it still seems less than ideal. Also, I don't know what should happen if you do `(make-temporary-file #:copy-from "some-file" #:directory? #t)`.

I've wanted to do this for the sake of the nice file name for a while, but I was reminded by your discussion with @benknoble on Slack. Maybe I should see if I can clarify the documentation around "combined with", too.

I also like @jackfirth's idea of just adding a `make-temporary-directory` version for clarity.

racket/collects/racket/file.rkt
```diff
@@ -134,60 +135,112 @@
...
+       (identifier? #'x)
+       #'make-temporary-directory/proc]
+      [(_)
+       #`(app make-temporary-directory/proc
+              '#,(infer-temporary-file-template stx))]
+      [(_ #:base-dir base-dir)
+       (not (keyword? #'base-dir))
+       #`(app make-temporary-directory/proc
+              '#,(infer-temporary-file-template stx)
+              #:base-dir #'base-dir)]
```
LiberalArtist, line 180: The current `make-temporary-file` takes care to use `#%app` with the context of the source syntax, but I'm not sure I've fully considered any implications for using keyword arguments.

LiberalArtist (commented):
I've pushed some work in progress on `make-temporary-directory`. I haven't added tests yet, and I still want to tweak the docs a bit.

racket/collects/racket/file.rkt
```diff
@@ -134,60 +135,112 @@
...
+                                 #:base-dir [_base-dir #f]
+                                 [copy-from _copy-from]
+                                 [base-dir _base-dir])
+      (do-make-temporary-file 'make-temporary-file
+                              template copy-from base-dir))
+    (define (make-temporary-directory [template "rkttmp~a"]
+                                      #:base-dir [base-dir #f])
+      (do-make-temporary-file 'make-temporary-directory
+                              template 'directory base-dir))
+    (define (do-make-temporary-file who template copy-from base-dir)
```
jackfirth, line 229: Instead of abstracting over these two functions from the outside in with a single `do-make-temporary-file` function, I suggest making the implementations of `make-temporary-file` and `make-temporary-directory` separate and pulling out some of the common chunks into helper functions.

LiberalArtist: I'm not sure I'm understanding what you have in mind here. The only actual difference based on `copy-from` is: https://github.com/racket/racket/blob/27e40598bf7187040fb2df463c801c63e5f594e5/racket/collects/racket/file.rkt#L277-L281
Presumably `make-temporary-file` still needs to handle the `'directory` case for compatibility, so I'm not sure what else to split out.

LiberalArtist:
Right now, the documentation only specifies that:
 1. if `template` produces a "relative path"
 2. it is "combined with" some other path (`(or base-dir (find-system-path 'tmp-dir))`)
 3. "to obtain a full path".

The current implementation explicitly allows things that don't satisfy `relative-path?`. However, the expected behavior somewhat ambiguous: "full path" is not defined, and, on Windows, an _absolute_ path without a drive specification is not a _complete_ path:
```racket
#lang racket
(require rackunit)
(define absolute-no-drive
  (bytes->path #"\\absolute\\no\\drive" 'windows))
(check-false
 (relative-path? absolute-no-drive))
(check-true
 (absolute-path? absolute-no-drive))
(check-false
 (complete-path? absolute-no-drive))
```
Here are some examples of the current behavior:
```racket
#lang racket

(define template "\\Users\\philip\\absolute-tmp~a")

(make-temporary-file template #f "C:")

(find-system-path 'temp-dir)

(parameterize ([current-directory "R:\\some\\dir\\"])
  (make-temporary-file template))

(make-temporary-file template #f "C:\\some\\dir\\")
```
```
#<path:C:\Users\philip\absolute-tmp16237057451623705745768>
#<path:C:\Users\philip\AppData\Local\Temp\>
"\\Users\\philip\\absolute-tmp16237057451623705745769"
..\..\Program Files\Racket\collects\racket\file.rkt:172:9: build-path: absolute path cannot be added to a base path
  absolute path: "\\Users\\philip\\absolute-tmp16237057451623705745770"
  base path: #<path:C:\some\dir\>
```
Note that `"\\Users\\philip\\absolute-tmp16237057451623705745769"` is a string, not a path! The created file is on the `R:` drive, i.e. relative to `(current-directory)` rather than `(find-system-path 'temp-dir)`.

That case seems like it should be fixed to use `path->complete-path` explicitly.

I am less sure about the last case. I could imagine someone wanting it to resolve the drive relative to the `base-dir`, but I'm inclined to just leave it as an error, and I'm reluctant to add too much Windows-specific logic.

LiberalArtist:
I've pushed an update to the docs with my current understanding of what `make-temporary-file` does or should do. I have some more tests to add later.

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
```diff
@@ -1294,43 +1294,75 @@ will not create it.
 
 
 @defproc[(make-temporary-file [template string? "rkttmp~a"]
-                              [copy-from-filename (or/c path-string? #f 'directory) #f]
-                              [directory (or/c path-string? #f) #f])
+                              [#:copy-from copy-from/kw (or/c path-string? #f 'directory) #f]
+                              [#:base-dir base-dir/kw (or/c path-string? #f) #f]
+                              [copy-from (or/c path-string? #f 'directory) copy-from/kw]
+                              [base-dir (or/c path-string? #f) base-dir/kw])
```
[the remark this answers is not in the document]
LiberalArtist: I've taken this suggestion, perhaps even more radically than you had in mind. I've relegated the old positional arguments to here:
https://github.com/racket/racket/blob/cc5db4d54bbcce737c37e8e3fba8931ad26b92b8/pkgs/racket-doc/scribblings/reference/filesystem.scrbl#L1362-L1375
I like it overall, as long as the deprecation doesn't seem too aggressive.

LiberalArtist:
Ok, I think that's everything. I've rebased and squashed: there is an un-squashed version at https://github.com/LiberalArtist/racket/tree/make-temporary-file-kw-hist.

samth (approved):
This looks great to me, especially the significantly expanded test suite.

samth:
Ah, Windows: https://github.com/racket/racket/pull/3870/checks?check_run_id=2843678835#step:8:51
```

ERROR: make-temporary-file: malformed template
  expected: a format string accepting 1 string argument
  given: "C:\\Users\\RUNNER~1\\AppData\\Local\\Temp\\absolute~a"
```

pkgs/racket-test-core/tests/racket/file.rktl
```diff
@@ -349,14 +349,123 @@
...
+
+(define rx:tmp-file #rx"make-temporary-file")
+(define rx:tmp-dir #rx"make-temporary-directory")
+(err/rt-test (make-temporary-file "bad\0~a") exn:fail:contract? rx:tmp-file)
+(err/rt-test (make-temporary-directory "bad\0~a") exn:fail:contract? rx:tmp-dir)
+(err/rt-test (make-temporary-file "bad~x") exn:fail:contract? rx:tmp-file)
+(err/rt-test (make-temporary-directory "bad~x") exn:fail:contract? rx:tmp-dir)
+(let* ([temp-dir (find-system-path 'temp-dir)]
+       [absolute-template
+        (path->string (build-path temp-dir "absolute~a"))])
```
samth, line 415: This is the part that causes the error in the windows tests -- using a path as a string isn't guaranteed (and here doesn't) produce a usable format string, and might have other problems as well.

LiberalArtist:
I'd initially started to just try to escape the template well enough for the tests to pass—testing the behavior when a template produces an absolute path seemed to require building a template from an absolute path—but kept considering the general problem that an API based on `format` brings in all the pitfalls of conflating strings and paths that Racket generally works hard to avoid. (Likewise I think it would have been better for many of these edge cases to just have been errors, but now that would be a breaking change.)

Instead, I added `make-temporary-file*` and `make-temporary-directory*`, which take a prefix and a suffix and combine them with `bytes-append`. (Almost all of this I did back in June, before going mostly offline for a while.) If this approach seems viable, I'll add more tests. I also think I might prefer for the prefix and suffix to be mandatory by-position arguments, rather than keyword arguments: `#""` is fewer characters than either `#:prefix` or `#:suffix`.

racket/collects/racket/file.rkt
```diff
@@ -134,100 +139,410 @@
...
+                           (path->relative-string/library source)
+                           (lambda (_ s) (string-upcase s)))]
+          [(string? source) source]
+          [else #f]))
+  (define str-loc
+    (cond [(and line col) (format "-~a-~a" line col)]
+          [pos (format "--~a" pos)]
+          [else ""]))
+  (define combined-str (string-append (or str-src "rkttmp") str-loc))
+  (define sanitized-str (regexp-replace* #rx"[<>:\"/\\|]" combined-str "-"))
```
LiberalArtist, line 383: It seems like this could have the same problem that came up in CI if the source file's path happened to contain a `~`: probably `sanatized-str` should handle that, too.

LiberalArtist:
Another issue is that temporary files are generally supposed to be created with mode `0600` to mitigate against various attacks. I think it's probably out of scope for this PR, but I don't think Racket exposes any way to control a file's permissions on creation or to set `umask` or something. It would be possible for `make-temporary-file` to use `file-or-directory-permissions`, but I think doing it in two steps might create a race.

LiberalArtist:
@mflatt re <https://github.com/racket/racket/commit/41880b3816fa6dbb926bb545b75f5597beb9f842>, I cannot overstate the extent to which I am not a Windows person, but I think using [`CreateDirectoryW`](https://docs.microsoft.com/en-us/windows/win32/api/fileapi/nf-fileapi-createdirectoryw) rather than `_wmkdir` would allow some support for permissions on Windows.

LiberalArtist:
The [SEI CERT](https://wiki.sei.cmu.edu/confluence/display/c/FIO21-C.+Do+not+create+temporary+files+in+shared+directories) guidelines say that temporary files should use unpredictable filenames. In particular, they report that the FreeBSD implementation of `mkstemp` uses secure randomness to generate the filename.

Currently, Racket adds some (non-cryptographically-secure) randomness to the time to produce digits for the file name. I suppose instead we could use something based on `crypto-random-bytes`, though I'm not sure how much benefit this would give since we already avoid race conditions between generating the name and creating the file.

racket/collects/racket/file.rkt
```diff
@@ -134,100 +139,410 @@
...
+                       ;; try again with a new name
+                       (loop (- s (random 10))
+                             (+ ms (random 10))
+                             (add1 tries)))])
+      (if copy-from
+          (if (eq? copy-from 'directory)
+              (make-directory pth)
+              (copy-file copy-from pth))
+          (close-output-port (open-output-file pth)))
+      pth)))
```
LiberalArtist, line 218: As I look at this again, it seems there's no limit on the number of tries other than for `ERROR_ACCESS_DENIED` on Windows. It seems like it might be better to have some very high limit on tries in general—100? 1000?—so that, if something is going badly wrong, we eventually raise an exception, rather than looping forever.

LiberalArtist:
Some temporary file libraries use mechanisms like `O_TMPFILE` on Linux, `FILE_FLAG_DELETE_ON_CLOSE` on Windows, or unlinking the file after creation on generic Unix to have the OS clean up automatically, even after an abnormal exit, or perhaps for some additional level of security. I think, though, that Racket should not overload these functions to support those kinds of temporary files. They seem notably different from the named temporary files currently created by `make-temporary-file`, and I think a Racket interface for creating such files would need to think through a bunch of OS and filesystem portability issues.

LiberalArtist:
I'm wondering what to do about permissions in the `#:copy-from` case. There's an argument for copying the permissions of the original, or at least having the option to do so, but there's also the security consideration for favoring mode `0600`. Currently, `copy-file` doesn't support a `#:permissions` argument: should it?

If so, I think it might be a reasonable compromise between security and compatibility for `make-temporary-file{*,}` to get a `#:permissions` argument that defaults to `#f`, meaning `0600` or the permissions of the `#:copy-from` file, if any. (The potential to mess with execute permissions seems like the most notable compatibility issue.) Since the `-directory` functions are completely new and don't have a `#:copy-from` case, I think their `#:permissions` could just default to `0600`.

Also, it looks to me from https://github.com/racket/racket/blob/41880b3816fa6dbb926bb545b75f5597beb9f842/racket/src/rktio/rktio_fs.c#L1740-L1756 that, on non-Windows systems, permissions are set only after the file has been copied, which seems like it might be problematic. Another enhancement would be using functions that avoid copying data from the kernel to userspace and back, like `copyfile`, `copy_file_range`, or `sendfile`, on systems that support them (Mac, Linux, apparently Solaris, and I think at least some BSDs). That's mostly unrelated, but some of those functions provide different mechanisms for setting or copying permissions.

racket/collects/racket/file.rkt
```diff
@@ -134,100 +139,410 @@
...
+      [(_ #:base-dir base-dir)
+       (not (keyword? #'base-dir))
+       #`(app make-temporary-directory/proc
+              '#,(infer-temporary-file-template stx)
+              #:base-dir base-dir)]
+      [(_ . whatever)
+       #'(app make-temporary-directory/proc . whatever)])))
+
 (define-syntax (make-temporary-file stx)
   (with-syntax ([app (datum->syntax stx #'#%app stx)])
```
LiberalArtist, line 421: I think this (existing) line is actually a no-op, because the [docs](https://docs.racket-lang.org/reference/stxops.html#%28def._%28%28quote._~23~25kernel%29._datum-~3esyntax%29%29) say that "if `v` is already a syntax object, then there is no conversion, and `v` is returned unmodified."

benknoble: Surely it's not a no-op: this looks like it's adjusting the `#%app` syntax object to be in the syntax scope/hygiene (terminology unclear to me) of the original `stx`.

LiberalArtist: That would be true if it were `'#%app` (as I had assumed until this evening), but it is actually `#'#%app`, so, IIUC, it will not.

It seems very possible this line of code was _supposed_ to do that, but I was never clear on exactly _why_ it seemed to be trying to do that, so, if I'm right that it is not actually doing anything now, I'm inclined not to make it start doing anything.

LiberalArtist:
# Roadmap

Other than the one discussed below, the tests now pass, and the documentation corresponds to the implementations.

Before merging, I think we should consider what to do about `#:permissions` when given a path as `#:copy-from`, as I [mentioned above](https://github.com/racket/racket/pull/3870#issuecomment-957971651). Currently, it is ignored, but that seems less than ideal. Some possibilities seem to be:

  - Add a `#:permissions` argument to `copy-file`, with `#f` meaning the current behavior (copy the permissions). I like this idea, but it might (or might not!) take some work. For `RKTIO_SYSTEM_UNIX`, the current `rktio_copy_file_start` implementation, `copy_file_range`, and `fcopyfile` all operate on file descriptors, but the Windows `CopyFileW` uses paths, and I'm not sure what the right thing to do there would be.
  - Remove `#:permissions` from this PR and make a new PR to sort out these questions.
  - Keep `#:permissions`, but document that the behavior when `#:permissions` is not `#f` and `#:copy-from` is a path may change.
  - Keep the current behavior forever.
  - Work around this non-atomically in `make-temporary-file`. That seems like a bad option.

Whatever we do, it would be good to have more tests for the `*` variants and `#:permissions`: I'll try to add some.

Here are some things I don't think should be part of this PR, but might be worth doing:

- Use `crypto-random-bytes`, like FreeBSD, per SEI CERT: https://github.com/racket/racket/pull/3870#issuecomment-957779099
- Maybe use `CreateDirectoryW` to support `make-directory` with permissions on Windows: https://github.com/racket/racket/pull/3870#issuecomment-957718280
- Give up eventually: https://github.com/racket/racket/pull/3870#discussion_r741177829

# Failing test

I do not (yet) understand the new Windows-specific failures, which seem to originate here:

https://github.com/racket/racket/blob/abc259abdaa2529e69f5fb0e468a564dae0a75fd/pkgs/racket-test-core/tests/racket/stx.rktl#L1856-L1915

with the error:
```
ERROR: open-output-file: error opening file
  path: C:\Users\RUNNER~1\AppData\Local\Temp\PKGS-racket-test-core-t----racket-stx.rktl-1861-16_16415034751641503475522
  system error: Access is denied.; win_err=5
```
The special treatment of `win_err=5` here makes me suspicious that something mysterious may be going on:
https://github.com/racket/racket/blob/3413b87357dc804b02f6a6310c69e061833d03b8/racket/collects/racket/file.rkt#L213-L227

LiberalArtist:
The combination of mysterious platform-specific test failures and unresolved questions about the correct behavior convinced me that `#:permissions` belongs in its own pull request.

I've added more tests to this one, and it should now be ready to merge.

LiberalArtist:
The new test failure related to `custom-elem-contents` seems unrelated.

LiberalArtist:
I've opened https://github.com/racket/racket/pull/4126 (as a draft) so as not to loose track of `#:permissions`.

samth (approved):

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
```diff
@@ -1390,43 +1391,131 @@ will not create it.
...
+ @racket[make-temporary-directory], respectively, but, rather
+ than using a template for @racket[format], the path is based
+ on @racket[(bytes-append prefix generated suffix)], where
+ @racket[generated] is a byte string chosen by the
+ implementation to produce a unique path. If there is source
+ location information for the callsite of
+ @racket[make-temporary-file*] or
+ @racket[make-temporary-directory*], @racket[generated] will
+ incorporate that information. The resulting path is combined
+ with @racket[base-dir] as with @racket[make-temorary-file].
```
mflatt, line 1514: Typo: `temorary` -> `temporary`

mflatt:
This looks generally, good, although I haven't yet tried it in many environments. You've clearly paid attention to many issues!

My only concern is that the documentation for `make-tempoary-file` doesn't reflects the function's actual signature. Instead of either keyword arguments or by-position arguments, the function accepts both at the same time; the by-position defaults are taken from keyword arguments, but the by-position ones take precedence when provided. I think documenting the function that way is better, still with the by-position variants described as being for backward compatibility. (I'm sure there's a precedent for this, but I can't remember where.)

LiberalArtist:
> My only concern is that the documentation for `make-temporary-file` doesn't reflects the function's actual signature. Instead of either keyword arguments or by-position arguments, the function accepts both at the same time; the by-position defaults are taken from keyword arguments, but the by-position ones take precedence when provided. I think documenting the function that way is better, still with the by-position variants described as being for backward compatibility. (I'm sure there's a precedent for this, but I can't remember where.)

That makes sense. I've pushed a revised commit.

PR #3879, opened 2021-06-14
docs: clarify `most-{positive,negative}-fixnum`

LiberalArtist:
For example, `(< (most-negative-fixnum) 0.5 (most-positive-fixnum))`
is `#true`, but `0.5` is not a fixnum.


mflatt:
Thanks!

PR #3880, opened 2021-06-15
rewrote explanation of one example in 16.2.6.2 Phases and Modules

maueroats:
Directly trace through the bindings to explain why the example does not end up in phase 0.
This is a new explanation of why importing a module into two different phases will not (by itself) solve a phase-level mismatch in macro expansion.


maueroats:
Image of rendered version of changed paragraph.
2021-06-14-at-22-39

mflatt:
Thanks! Ok if I squash the addition of those two commas while merging?

maueroats (commented):
Looks great to me.

PR #3886, opened 2021-06-21
ci: fix iOS build

Bogdanp:
The `make install` step no longer copies the xpatch files into the
"lib/" directory as of 30effc9 so we have to do it manually here.


pmatos:
lgtm - will merge

PR #3890, opened 2021-06-22
ci: fix push workflow for iOS

Bogdanp:
I forgot about this workflow so the change in 5ed13f64be only covered the "pull request" workflow.


pmatos:
Thanks

PR #3898, opened 2021-06-29
cs: add make-uninitialized-string and use it within primitives

Bogdanp:
This change makes it so that `substring`, `string-append`, `string-copy` and `list->string` don't have to spend time filling newly-allocated strings. In all of those cases, the newly-allocated strings are immediately filled with the contents of other strings so this should be a safe change.


samth:
Do the byte string functions have the same issue?

mflatt:
Since `make-uninitialized-string` is undocumented, unsafe, and meant for internal use, it should be called `$make-uninitialized-string` and listed in "cmacros.ss" with other internal `$` functions.

Bytevectors in Chez Scheme don't have the same issue, because `make-bytevector` creates an uninitialized bytevector unless a fill is supplied, and Rumble's `bytes-append` takes advantage of that. Racket doesn't expose the raw `make-bytevector` because it's insecure in the sense that memory formerly used by one entity could be exposed to another. The `make-string` function always fills, because not filling would be unsafe: using `string-ref` on an uninitialized string could produce a garbage value. 

I worry a little that `string-append` using `$make-uninitialized-string` can allow a partially initialized string to be exposed through interrupts and Chez Scheme's inspection API. Turning off interrupts would solve that problem but may be a cure-worse-than-the-disease solution.

Bogdanp:
> Since make-uninitialized-string is undocumented, unsafe, and meant for internal use, it should be called $make-uninitialized-string and listed in "cmacros.ss" with other internal $ functions.

Thanks! I'll make that change later today.

> I worry a little that string-append using $make-uninitialized-string can allow a partially initialized string to be exposed through interrupts and Chez Scheme's inspection API. Turning off interrupts would solve that problem but may be a cure-worse-than-the-disease solution.

Does `bytes-append` have the same issue? Or does Rumble protect against this problem? Or is your worry that interrupting on a string could leave it in a bad state where it contains invalid characters, whereas a half-initialized bytes value is still a valid value despite not containing the expected data?

mflatt:
Yes, the difference between `bytes-append` versus `string-append` is Racket versus Chez Scheme and some notion of security versus safety. Racket doesn't provide access to facilities (through its safe API) that would let you look at arbitrary values reachable from a continuation. Meanwhile, Chez Scheme aims for safety (so a string where `string-ref` can give bad values is a problem) without an extra notion of in-process security.

Bogdanp:
@mflatt Instead of "cmacros.ss" did you mean "primdata.ss" by any chance? I don't see many internal function definitions in "cmacros.ss", but I could definitely be missing something.

I've bracketed the parts that deal with uninitialized strings with `with-interrupts-disabled` and I've been using [this program](https://gist.github.com/Bogdanp/7a6a40711466338be732198ac6254c07) to test the differences against the latest 8.2.0.1 snapshot. Here are the results on my machine (CPU time):

```
                   8.2.0.1 SNAP      THIS BRANCH
append short              121ms            132ms
append long               229ms             19ms
append many long         1833ms           1360ms
substring short           148ms            261ms
substring medium         1575ms           1349ms
substring long           1782ms           1163ms
string-copy short         129ms            140ms
string-copy long        21731ms           1401ms
list->string short       1263ms           2242ms
list->string long        2617ms           2116ms
```

For very short strings, the bracketing does add some overhead. I could add a length cutoff to determine which allocation style to use. For example, for strings shorter than 100 characters, we could use `make-string` and no bracketing and `$make-uninintialized-string` with bracketing for everything else. That would come at the cost of duplicating some of this code, though. What do you think?

mflatt:
Yes, sorry, I meant "primdata.ss".

If dropping `with-interrupts-disabled` makes a measurable difference, I think it's ok to just drop it. The more I think about it, I doubt that this would be the only place where an ill-formed value is potentially exposed by exploring a continuation, and it seems not worth worrying about. Either way, a comment about the presence or absence of `with-interrupts-disabled` is probably a good idea.

Thanks!

Bogdanp:
I've dropped the uses of `with-interrupts-disabled` and added a comment. Here are the numbers without bracketing:

```
                   8.2.0.1 SNAP      THIS BRANCH       NO BRACKETING
append short              121ms            132ms               133ms
append long               229ms             19ms               221ms
append many long         1833ms           1360ms              1358ms
substring short           148ms            261ms               140ms
substring medium         1575ms           1349ms              1294ms
substring long           1782ms           1163ms              1400ms
string-copy short         129ms            140ms               121ms
string-copy long        21731ms           1401ms              8000ms
list->string short       1263ms           2242ms              1257ms
list->string long        2617ms           2116ms              2456ms
```

With bracketing, the "append long" and "string-copy long" benchmarks I think are somewhat misleading simply because the bracketing seems to make the GC less likely to act (and those two benchmarks spend a good amount of time in GC otherwise) so these results look better to me overall.

mflatt:
This looks ready to me. Should I squash and merge now?

Bogdanp:
Yup! Thanks.

Issue #3899, opened 2021-06-29
match trusts struct info


rmculpepper:
`match` trusts struct info too much. In particular, it assumes that the reference to the struct predicate identifier evaluates to the predicate for the struct type, and it generates unsafe code based on that assumption.

    > (match (cons 1 2) [(cons x y) (+ x y)])
    3
    > (current-code-inspector (make-inspector (current-code-inspector)))
    > (struct point (x y))
    > (set! point? pair?)
    > (match (cons 1 2) [(point x y) (list x y)])
    '(562949953421312 1618481116086272)

The first use of `match` is necessary to load its modules in the original code inspector. (This should probably be considered a bug in `lazy-require`, if that's what it's using.)

samth:
Is it enough to handle mutable bindings differently, or is there something that doesn't involve set! that can also go wrong?

rmculpepper:
Here's a similar issue using `local-expand`, which works because nothing binds the values definition and the syntax definition together:

    #lang racket/base
    (require (for-syntax racket/base racket/syntax)
             racket/match)

    (define-syntax (bad-struct stx)
      (syntax-case stx ()
        [(_ name (field ...))
         (let ([ee (local-expand #'(struct name (field ...))
                                 (syntax-local-context)
                                 #f)])
           (with-syntax ([(begin _vdef sdef) ee]
                         [n? (format-id #'name "~a?" #'name)]
                         [(n-f ...)
                          (map (lambda (f) (format-id #'name "~a-~a" #'name f))
                               (syntax->list #'(field ...)))])
             #'(begin (define n? pair?)
                      (define (n-f n) #f) ...
                      sdef)))]))

    (bad-struct point (x y))

    (match (cons 1 2)
      [(point x y) (list x y)])

samth:
This seems bad. We constructed a whole hierarchy of static struct info so that `match` could trust only the ones that actually came from `struct` itself, but this example suggests that it's not possible to trust it. The `local-expand` example seems like the sort of thing that taints should be about fixing, but obviously that's not working here.

mflatt:
If `local-expand` were protected, then that example is like something that injects bad uses of unsafe operations.

On the mutability side, maybe `struct` should expand to code that raises an error if a binding is mutable; the check would be optimized away, normally. This kind of change could be a backward-compatibility issue, but maybe not too big of a problem.

[...]

PR #3903, opened 2021-07-01
Check for mutated struct predicates.

samth:
Related to #3899.

### DIFF
--- a/pkgs/racket-test/tests/match/examples.rkt
+++ b/pkgs/racket-test/tests/match/examples.rkt
@@ -7,6 +7,7 @@
          (only-in racket/list split-at)
          (for-syntax scheme/base)
          (prefix-in m: mzlib/match)
+         (prefix-in r: racket/base)
          rackunit)
 
 (define-syntax (comp stx)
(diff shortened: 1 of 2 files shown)


mfelleisen:
I wonder whether we could have a "dirty" bit per struct and then just pay the price for all accessors if any one of them has been mutated.

samth:
I think that would involve a runtime test, whereas this approach will be entirely optimized away.

mfelleisen:
You are generating n run-time tests if I read this code correctly:

```
  #`(if (variable-reference-constant? (#%variable-reference #,pred))
                                (λ (x) (unsafe-struct-ref x #,i))
                                #,acc))
```

 I am proposing to generate one but punish the entire suite of accessors at once.

samth:
That code will always be optimized to just the true or just the false branch at compile time.

samth:
@racket/release should we include this in 8.2?

mflatt:
I think it makes sense to include this in v8.2.

To help confirm that the amount of generated code is ok: a build plot before and after the change shows no difference (i.e., in the noise) for build time or memory use on x86_64 Mac OS.

jbclements:
picked, thanks.

PR #3907, opened 2021-07-07
Add checks to placeholder-set! and placeholder-get

camoy:
These two functions did not check that their argument was actually a placeholder.


mflatt:
Thanks for this repair!

I suggest using the simpler form of argument checking, `(check who placeholder? ph)`, since the test expression and contract can be derived from just `placeholder?` (unless I'm missing something). Would you like to make that change before it's merged?

camoy:
Thanks for the suggestion, should be good now.

mflatt:
Thanks!

PR #3909, opened 2021-07-08
Add module-level contract to make-limited-input-port

zyrolasting:
First PR, so I'm probably missing something. Just point me to the materials I missed and I'll update accordingly (Case in point: How do you all define tests for module-level contracts?)

Consider:

```racket
(define in (make-limited-input-port (open-input-bytes #"_") 1.0))
(read-byte in)
```

`make-limited-input-port` does not fail, and returns a useless input port. The contract for `read-bytes-avail!*` raises a violation when a read actually occurs. At that point one might not have the right party to blame.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/portlib.rktl
+++ b/pkgs/racket-test-core/tests/racket/portlib.rktl
@@ -723,6 +723,11 @@
       (test 1 peek-bytes-avail!* b 1 #f s)
       (test 2 read-bytes-avail!* b s))))
 
+;; module-level contract violation check
+(err/rt-test (make-limited-input-port (open-input-bytes #"_") 1.0)
+             exn:fail:contract?)
+
+
 ;; Make sure that blocking on a limited input port doesn't
 ;; block in the case of a peek after available bytes:
 (let ()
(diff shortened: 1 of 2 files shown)


mflatt:
This patch looks good, but as you suggest, it would be good to have a test to make sure that `(make-limited-input-port (open-input-bytes #"_") 1.0)` raises an exception. The test would go in "pkgs/racket-test-core/tests/racket/portlib.rktl" using `err/rt-test`.

zyrolasting:
Alright, that's up. `raco setup` runs to completion, which appears to be the way to run these tests...?

sorawee:
`raco test -l tests/racket/test` should run the test. If you want to only test `portlib`, you can temporarily modify `tests/racket/all.rktl` to only reference `portlib`, but don't delete the first line -- `(load-relative "loadtest.rktl")`. (Would love to know if there's a better way, but this is what I have been doing...)

zyrolasting:
Thank you! Just ran everything. `Passed all tests.`

mflatt:
Thanks!

samth:
@sorawee the best was is `racket -f portlib.rktl` from in that directory, or `racket -f quiet.rktl portlib.rktl` for less output.

PR #3915, opened 2021-07-16
remove syntax arming, but protect macro-expansion operations

mflatt:
Syntax arming and disarming were part of a design to allow sandboxing untrusted code without unduly constraining the sandboxed code. This commit replaces that approach with a simpler one. The trade-off is that some advanced macro-implementation tools, including `syntax-local-value` and `local-expand`, cannot be referenced directly within a sandbox (i.e., in a context where the code inspector is different than the original one).  The commit message has more detail.

This change is work with @michaelballantyne, who identified problems with the current system and moved the discussion in this direction, plus @rmculpepper, @samth, and @mfelleisen.

The change breaks sandbox tests, and the `sandbox-lib` package will need to be updated to work right again. On the whole, this seems very likely an improvement for `sandbox-lib`. It's possible that other packages will break, and we'll need to check specifically if/when we merge the change.


mflatt:
Merged.

PR #3918, opened 2021-07-19
doc: use processor-count from racket/future

sorawee:
There's a broken link at https://www.cs.utah.edu/plt/snapshots/current/doc/raco/api_parallel-build.html#%28def._%28%28lib._setup%2Fparallel-build..rkt%29._parallel-compile-files%29%29

The cause seems to be due to how both `racket/future` and `racket/place` provide `processor-count`, so Scribble got confused. It looks like if Scribble is smarter, then this PR wouldn't be necessary. So feel free to not merge the PR if you think it's not the right solution.

### DIFF
--- a/pkgs/racket-doc/scribblings/raco/make.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/make.scrbl
@@ -6,7 +6,7 @@
                      racket/future
                      racket/promise
                      racket/file
-                     racket/place
+                     (except-in racket/place processor-count)
                      compiler/cm
                      compiler/cm-accomplice
                      setup/parallel-build


mflatt:
I think the problem is that `processor-count` is not documented correctly, and it the build was previously just lucky that it found the binding via `racket/future` instead of `racket/place`. So, the best fix is something on the `processor-count` side.

Possible solutions:

 * make `processor-count` documented as from `'#%futures` — normally the right choice for bindings that are reexported by multiple libraries, but it has the drawback that old links break
 * make `racket/place` get the binding from `racket/future` instead of `#%futures` — would work fine in practice, but a remaining problem is that the documentation for `racket/place` doesn't mention that it provides the binding
 * document `processor-count` separately for `racket/place`, and note in the documentation that it's the same binding as `racket/future` — preserves old links and more naturally documents how `processor-count` is provided from both libraries; in a Scribble document that imports from both `racket/future` and `racket/place`, however, the link will go to one or the other randomly

I think the third solution is probably right to fix the documentation for `processor-count`.

Meanwhile, "make.scrbl" really shouldn't require `racket/future`, because it doesn't need any of the other bindings from that library. I think the `racket/future` import was probably there originally to refer to `processor-count`, and then `processor-count` got added to `racket/place` (probably on purpose) without a needed adjustment to the documentation.

PR #3919, opened 2021-07-19
Fix typos

sschwarzer:
In `performance.scrbl`, that's how I understand the sentence. Please
change it if something else is meant.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/performance.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/performance.scrbl
@@ -573,15 +573,13 @@ argument instead.
 
 @section[#:tag "Reachability and Garbage Collection"]{Reachability and Garbage Collection}
 
-In general, Racket re-uses the storage for a value when the
-garbage collector can prove that the object is unreachable from
-any other (reachable) value. Reachability is a low-level, 
-abstraction breaking concept (and thus one must understand many
-details of the runtime system's implementation to accurately predict
-when values are reachable from each other),
-but generally speaking one value is reachable from a second one when 
-there is some operation to recover the original value from the second
-one.
+In general, Racket re-uses the storage for a value when the garbage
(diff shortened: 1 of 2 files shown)


pkgs/racket-doc/scribblings/guide/performance.scrbl
```diff
@@ -577,8 +577,8 @@ In general, Racket re-uses the storage for a value when the
 garbage collector can prove that the object is unreachable from
 any other (reachable) value. Reachability is a low-level, 
 abstraction breaking concept (and thus one must understand many
-details of the runtime system's implementation to accurate predicate
-precisely when values are reachable from each other),
+details of the runtime system's implementation to accurately predict
```
samth, line 580:
```suggestion
details of the runtime system's implementation to accurately predict precisely
```

sschwarzer: Hi Sam,

I thought about this, too. However, in my opinion "accurate" and "precise" are so close in meaning that nothing is won if you use both words; it looks rather redundant. (When deciding which of "accurately" or "precisely" to remove, I stuck with "accurate" because it sounds better with "predict.")

I can accept the suggestion if you want me to :-) , but I think keeping the sentence simple(r) would be better for most readers.

sschwarzer: Updated the paragraph after discussion with Sam. I tried to express the gist of the paragraph, intending to make it a bit easier to read.

sorawee: Not sure if @samth wants to include the word "precise" because it's a technical word ("precise GC"). Though precise GC is also known as accurate/exact GC...

sorawee: Oh... I'm two minutes too late. Looks like you talked to each other already...

Issue #3925, opened 2021-07-22
dynamic-rerequire does not always reload a module if the module's dependencies are modified


djholtby:
**What version of Racket are you using?**
8.2 [cs], 8.1 [cs], 7.9 [bc]

**What program did you run?**
`dir/one.rkt`
```
#lang racket/base
(require "two.rkt")
(provide alpha)
(define alpha beta)
```

`main.rkt`
```
#lang racket/base

(require racket/rerequire)
(module+ test (require rackunit))

(define (get-alpha)
  (dynamic-rerequire "dir/one.rkt")
  (dynamic-require "dir/one.rkt" 'alpha))

(define (set-beta! v)
   (with-output-to-file "dir/two.rkt" #:exists 'replace
     (lambda ()
      (printf "#lang racket/base\n(provide beta)\n(define beta ~v)\n" v))))

(module+ test
  (set-beta! 2)
  (check-eqv? (get-alpha) 2)
  (set-beta! 3)
  (check-eqv? (get-alpha) 3))
```

Results: second check fails (alpha is still 2)

**What should have happened?**

The second call to `dynamic-rerequire` by way of `get-one` should have triggered a reload of "two.rkt" and "one.rkt", resulting in `alpha` being equal to 3.

**Other Details**

This happens in Dr.Racket but not when invoking `racket main.rkt` 

Looking into it, there was a "beta.zo" being created when running in Dr.Racket (8.2 [cs]) and the issue ends up being due to compiled files and their interaction with the loader.  

It also happens if the dependency is in a submodule (looks to be reported as #971) in which case it happens from command line also.

sorawee:
This is fixed, IIUC. Closing.

Issue #971, opened 2015-07-13
`dynamic-rerequire` does not transitively reload dependencies within submodules


mbutterick:
I believe there’s a discrepancy in `dynamic-rerequire` between the loading and reloading procedures. The loader will find dependencies inside submodules; the reloader will not.

Test case: save this as "one.rkt"

``` racket
#lang racket/base

(module inner racket/base
  (require "two.rkt")
  (provide proc))

(require 'inner)
(provide proc)
```

Test case runner:

``` racket

#lang racket
(require rackunit racket/file racket/rerequire)

(define (change-required-file path proc-value)
  (display-to-file (format "#lang racket/base
(provide proc)
(define (proc) ~v)" proc-value) path #:exists 'replace)
  (sleep 1) ; to make sure mod time changes so rerequire notices it
  (file-or-directory-modify-seconds path (current-seconds)))

;; create "two.rkt" with `proc` function that evaluates to "foo"
(change-required-file "two.rkt" "foo")
;; even though "two.rkt" is inside a submodule, rerequire will transitively load it the first time
(check-true (and (member (path->complete-path "two.rkt") (dynamic-rerequire "one.rkt")) #t))
;; change "two.rkt"
(change-required-file "two.rkt" "zam")
;; this will error: rerequire should transitively reload "two.rkt", but doesn't
(check-false (empty? (dynamic-rerequire "one.rkt"))) 
```

My best theory so far is that the bug is inside [`check-latest`](https://github.com/plt/racket/blob/b4c52698c0128539e5eb318b3082eba4a992687f/racket/collects/racket/rerequire.rkt#L93). At this point in the code, it’s possible for `p`, the result of `resolved-module-path-name`, to be a submodule path (i.e., a list of symbols). If so, `check-latest` ought to recurse into this submodule to keep checking dependencies. Instead, it tries to make `p` into a path by doing `(car p)`. This seems wrong — the result will always be a symbol, and thus ignored by the rest of `check-latest`, which actually does the reloading.

mbutterick:
Here’s a further exploration of this idea with a [new version of `check-latest`](https://gist.github.com/mbutterick/3f7f22fa92e748c43ab7#file-rerequire-rkt-L108). So far this seems to work — it cures the test case shown above — but I’m sure it contains some lame error so I’m not yet upgrading it to a pull request.

[...]

PR #3926, opened 2021-07-22
Fix rerequire reload on changed dependencies

djholtby:
Fix for #3925 (maybe #971)

mflatt pointed me to `resolve-module-path-index` being needed to rebase the module-path-index attached to the exports of  .zo files (which is always relative to themselves)

### DIFF
--- a/pkgs/racket-test/tests/racket/rerequire.rkt
+++ b/pkgs/racket-test/tests/racket/rerequire.rkt
@@ -67,7 +67,7 @@
                                    dirpath))])
             (delete-file file)))
 
-        (define (change-required-file proc-value)
+        (define (create-or-change-required-file proc-value recreate?)
           (display-to-file
            (string-append
             "#lang racket/base\n"
@@ -76,7 +76,7 @@
             (format "(define (proc) ~v)" proc-value))
            two-path
            #:exists 'replace)
-          (sleep 1) ; to make sure mod time changes so rerequire notices it
+          (when recreate? (sleep 1)) ; to make sure mod time changes so rerequire notices it
           (file-or-directory-modify-seconds
            two-path
            (current-seconds)))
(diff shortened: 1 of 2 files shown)


djholtby:
While I was poking at it I noticed it was reloading files that have submodules multiple times, bf68c45 adds a second hash table to solve this so that each that needs a reload is loaded only once.

mflatt:
Thanks for this repair!

Some tests for this module are in `pkgs/racket-test/tests/racket/rerequire.rkt`. The tests there currently are not at all exhaustive, but I started the test suite when adding a previous repair. Would it make sense to add some tests there?

djholtby:
Very likely -
One change needed is instead of checking for a non-empty list, it should
check specifically that both one.rkt and two.rkt were reloaded (in case
it's only reloading the changed file, but not ones that depend on it, which
is the issue I was having).
Another is to try different combinations of files with only source
available, and files that have been compiled.

I've done both here:
https://github.com/djholtby/racket/blob/additional-rerequire-tests/pkgs/racket-test/tests/racket/rerequire.rkt

I also added a "zero" file in a different directory to make sure it's
changing directories without issue.  Not that I think that makes a
difference, but still.

The last few commits are my descent into madness as I try to make sure it's
not "hard coding" the location of the compiled files.
I settled on using a regexp to look for something that looks vaguely like a
compiled version of the target file.

djholtby:
OK, I found the proper way to locate the compiled file name.  (I think)
Added a reworked test file.  

Without the repair all of the tests fail on Windows 10 and Ubuntu 20.04  (even the "all source" test) at the reload stage, so using 3 files wasn't excessive after all.  
(However doing all 16 permutations of compiled and source is probably excessive).

mflatt:
Thanks!

Issue #3251, opened 2020-06-15
Definition context API change proposal


michaelballantyne:
Starting with <https://github.com/racket/racket/issues/3198>, I've been thinking about the first-class definition context API. Several flaws make most all macros using the API unhygienic in certain situations. To fix the API, I propose:

* Each first-class definiton context should come with an associated outside-edge scope.
* The expander should track sets of use-site scopes to remove for each first-class definition context, rather that in the surrounding expand context.
* We should replace `internal-definition-context-introduce` with operations that manipulate outside-edge and use-site scopes along with inside-edge scopes.

### Outside-edge scopes

See <https://github.com/racket/racket/issues/3198>.

I prefer to integrate the outside-edge scopes into definition context operations rather than create separate operations so that macro authors don't need to know about and apply separate operations for each kind of scope. Instead, one operation should do all the work needed to add syntax to the definition context's conceptual scope.

### Parent inside-edge scopes

`local-expand` and other operations add the inside-edge scope of the immediate definition context, but not parent definition contexts. Currently, macros that use chains of definition contexts sometimes accumulate a list of all the definition contexts and either use `internal-definition-context-introduce` or pass the whole list to `local-expand` to ensure that the parent inside-edge scopes are added.

With the addition of an outside-edge scope, macros will need to apply the outside-edge scope from each definiton context before syntax expands. If the same operation applies the inside-edge scope, macros will not need to perform the additional work to accumulating definition contexts to ensure that parent inside-edge scopes have been applied.

### Use-site scopes

The current strategy of tracking use-site scopes to remove in the expand context and resetting at expression contexts doesn't quite work. Consider this example:

```
#lang racket

(require racket/block)

(define x 'good)

(let ()
  (define-syntax-rule (m1 arg)
    ; I use an explicit `#%expression` because the current 
    ; implementation of block does not check it is in an
    ; expression context.
    (#%expression
     (block
      (define arg 'bad)
      x)))
  (define res (m1 x))
  res) ; => evaluates to 'bad
```

The definition right-hand-side where `(m1 x)` expands is already in an expression context, with a fresh, empty set of use-site scopes. The use-site scope of the expansion is recorded in the set. Expansion in `block`'s first-class definition contexts uses the set from the expand context, so when `block` binds or uses `syntax-local-identifier-as-binding`, the use-site scope is removed.

[...]

Issue #3198, opened 2020-05-21
Internal-definition context hygiene bug


michaelballantyne:
This program:

```
#lang racket
(require racket/block)

(define x 'good)
(define-syntax-rule (m) (displayln x))

(block
 (define x 'bad1)
 (m))

(define c%
  (class object%
    (super-new)
    (define x 'bad2)
    (m)))
(new c%)
```

Displays `'bad1` and `bad2`. I suspect other definition context macros behave similarly.

The underlying problem is that first-class definition contexts add inside-edge scopes but not outside-edge scopes.

We might think to fix this by adding an outside edge scope via `make-syntax-introducer`. Unfortunately that solution interacts badly with the way `'quote-syntax` removes scopes; the change breaks this test from `racket-test-core/tests/racket/macro.rktl`:

```
(module uses-internal-definition-context-around-id racket/base
  (require (for-syntax racket/base
                       racket/block))

  (define-syntax (m stx)
    (let ([x1 #'x]
          [x2 (block
               (define x3 #'x)
               x3)])
      #`(let ([#,x2 1]) #,x1)))

  (define v (m))
  (provide v))
(test 1 dynamic-require ''uses-internal-definition-context-around-id 'v)
```

The test breaks because the scope added via syntax introducer is not added to the expand context's set for `quote-syntax` to remove.

mflatt:
When you say that an outside-edge scope interacts badly with `quote-syntax`, do you mean when `make-syntax-introducer` is used explicitly by `block`?

If so, is the solution to have `make-internal-definition-context` create an outside-edge scope that it records in the same way as the inside-edge scope, and then provide a function for `block` (and similar) to add that scope like the result of `make-syntax-introducer`?

Either way, it sounds like `block` and similar macros have to change to explicitly use an outside-edge scope. Is that right?

michaelballantyne:
> When you say that an outside-edge scope interacts badly with quote-syntax, do you mean when make-syntax-introducer is used explicitly by block?

Yes.

> If so, is the solution to have make-internal-definition-context create an outside-edge scope that it records in the same way as the inside-edge scope, and then provide a function for block (and similar) to add that scope like the result of make-syntax-introducer?

This sounds like a reasonable solution.

> Either way, it sounds like block and similar macros have to change to explicitly use an outside-edge scope. Is that right?

[...]

PR #3927, opened 2021-07-22
API changes supporting better hygiene behavior for first-class definition contexts

michaelballantyne:
This PR implements my proposal from https://github.com/racket/racket/issues/3251. The description below updates the text of that issue. It is written for macro system experts; I plan to write separate documentation for macro authors.

I haven't finished the documentation so the PR is a draft, but the implementation is ready and I'd appreciate comments!

## Summary

The goal of this PR is to make the hygiene behavior of expansion using first-class definition contexts fully in line with the expansion of Racket's core forms, to ensure correct hygiene for forms such as `class` and for DSL expanders. 

The current hygiene behavior supported by definition contexts falls short with respect to three elements: outside-edge scopes, use-site scopes, and the frame-id optimization to avoid unnecessary use-site scopes,

The changes, in summary:

* Each first-class definiton now comes with an associated outside-edge scope.
* The expander tracks sets of use-site scopes to remove with each first-class definition context, rather that in the surrounding expand context.
* `internal-definition-context-introduce` is replaced with operations that manipulate outside-edge and use-site scopes along with inside-edge scopes.
* The expander provides a new primitive operation for applying macro transformers in DSL expanders, replacing `local-apply-transformer`. The operation needs to be primitive in order to cooperate with the new use-site scope behavior. It also allows the DSL expander to specify the binding that the transformer is associated with in order to apply the frame-id optimization and to set the `current-module-code-inspector` appropriately.
* When the context is of type `'expression`, new definition contexts use a new frame-id rather than reuse that from the context.

I also propose to document support for lists of definition contexts in some API arguments as "for backwards compatibility only"; see the "Arguments supporting lists of definition contexts" section.

### Outside-edge scopes

See <https://github.com/racket/racket/issues/3198> for a description of the problem.

#### Solution

This PR associates an outside-edge scope with each first-class definition context, in addition to the present inside-edge scope.

Both scopes are registered with the expand context for removal at `quote-syntax` and addition via `syntax-local-get-shadower`.

The "API additions and changes" section discusses the operations used to manipulate these scopes.

### Use-site scopes

The existing approach to use-site scopes relies on tracking a set of use-site scopes in the expander context. In the `syntax-local-identifier-as-binding` operation or when a `define` form is expanded, those use-site scopes on the binding identifier that are also in the set tracked by the context are removed from the binder. The set is changed when entering certain new expansion contexts.

This approach leads to two problems.

#### Improper removal of use-site scopes that belong to the surrounding context

Consider this example:

```
#lang racket
(define x 'good)
(let ()
  (define-syntax-rule (m1 arg)
    (local [(define arg 'bad)]
      x))
  (define res (m1 x))
  res) ; => evaluates to 'bad
```

The correct expansion would have the `x` reference refer to the outer binding, not the use-site binder, as it would if the `local` expression were instead a `let`.

Right now, the expansion of `(m1 x)` applies a use-site scope. This scope is recorded in the expand context's set of use-site scopes. Because expansion does not newly enter an expression context before creating the binding for the `define` inside `local`, the use-site scope remains in the set and is removed from the binder. This results in the improper capture of the reference

#### Failure to record use-site scopes when they are disabled in the surrounding context

As noted in https://github.com/racket/racket/pull/2237, the expander currently disables the application of use-site scopes in expression contexts entered via `local-expand`. This PR does not change that behavior, but removes one of its secondary consequences.

Consider the interaction between Turnstile or a DSL expander that expands subexpressions fully with `local-expand`, here simulated by `expr`, with a definition context macro such as `local`:

```
#lang racket
(define-syntax (expr stx)
  (syntax-case stx ()
    [(_ e) (local-expand #'e 'expression '())]))
(expr
 (local [(define-syntax-rule (my-define x v) (define x v))
         (my-define x 5)]
   x))
; x: unbound identifier    
```

`local-expand` given `'expression` sets the expand context value for the use-site set to `#f`, disabling both application and tracking of use-site scopes. When `local` uses `local-expand` to expand the definitions, `local-expand` replaces that `#f` with a new set and records the use-site scope applied to the define. However, that new set only lasts as long as the `local-expand`---in the surrounding context where `local` will use `syntax-local-bind-syntaxes`, the value returns to `#f`. This means that the use-site scope is not removed, and the reference is unbound.

#### Solution

The solution to both of these problems is to track use-site scopes created by expansion for each first-class definition context. Expansion using a first-class definiton context sets the expand context use-site scope set to that stored with the definition context. In the first problem above this ensures that only use-site scopes created during the expansion within `local` are removed from the bindings. In the second, it ensures that the scopes are recorded with the definition context rather than only in the nested expand context that is discarded after the definition expansions.

### Frame-ids

The frame-id optimization ensures that use-site scopes are only applied for expansions where they are potentially necessary. Unless a macro definition and a use site are in the same definition context, the use-site should usually have additional scopes that would prevent use-site binders from capturing definition-site references. Generally the expander associates a frame-id with each definition context, and only applies use-site scopes when the frame-id of the macro binding and use match.

Splicing forms complicate this picture, because spliced macro definitions are bound in the surrounding definition context but their right-hand-sides have the scopes of the inner context. Thus there are no scopes to distinguish use-site binders from introduced references in this case. To accommodate this case, first-class definition contexts conservatively always use the same frame-id as their parent first-class definition context (if any) or the surrounding expansion context.

In a DSL expander, the result of this strategy is that the entire expansion uses a single frame-id and the optimization never applies.

#### Solution

This PR enables the frame-id optimization for some DSL expanders via two changes.

First, when the surrounding expand context is an expression context, syntax-local-make-definition-context creates new frame-id rather than reuses that of the context. This is safe because definitions cannot splice out of an expression context.

Second, the new `syntax-local-apply-transformer` discussed in the "API additions and changes" section accepts a `binding-id` argument indicating the binding associated with a macro transformer. The frame-id associated with this binding is used to apply the frame-id optimization when safe.

Note that this approach is not enough to allow a DSL expander for a language with nested scopes not separated by an expression contexts (a statement-oriented language, for example) to take advantage of the frame-id optimization. That would require an additional mechanism with which the expander could promise that no definitions will splice out from a given context, making it safe to use a new frame-id.

## API additions and changes

The PR adds three new operations: `internal-definition-context-add-scopes`, `internal-definition-context-remove-scopes`, and `syntax-local-apply-transformer`. It adds an optional argument to `syntax-local-identifier-as-binding`, as well as a return value from `syntax-local-bind-syntaxes`. I also propose to document `local-apply-transformer` and the use of a list of definition contexts rather than a single definition context in several operations as "for backwards compatibility only".

##### New operations

```
(internal-definition-context-add-scopes intdef-ctx stx) → syntax?
```

Adds the defintion context's inside- and outside-edge scopes to `stx`.

```
(internal-definition-context-remove-scopes intdef-ctx stx) → syntax?
```

Removes the definition context's inside- and outside-edge scopes. Also removes any use-site scopes created by expansions in the definition context. This operation is designed for use by macros that create splicing definition contexts. Removing the use-site scopes ensures that bindings properly splice out of the definition context.

The above two operations should generally replace uses of `internal-definition-context-introduce`, so I propose that the documentation of that operation suggest using the new operations instead.

##### Changed operations

```
(syntax-local-identifier-as-binding id-stx [intdef-ctx]) → identifier?
```

The new optional `intdef-ctx` argument determines which use-site scopes are removed. When it is `#f` (the default), the operation removes the use-site scopes in the set from the expand context. When it specifies a first-class definition context, it removes the use-site scopes in the set stored with that context. Most uses should provide a definition-context argument. Some macros such as `trait` locally expand definitions without binding them in a first-class definition context; in these cases the definition-context argument should be omitted. (Though such macros really should use a definition context for shadowing behavior consistent with the rest of Racket.)

```
(syntax-local-bind-syntaxes
  id-list expr intdef-ctx [extra-intdef-ctxs]) → (listof identifier?)
```

Now returns the identifiers derived from the `id-list` adjusted to match the scopes of the new bindings: inside-edge scopes added, and use-site scopes associated with the definition context removed. This obviates the need to separately construct identifiers matching the bindings using `internal-definition-context-introduce` and `syntax-local-identifier-as-binding`. This purpose change is just to make the API more convenient.

##### New operation

The change to use-site scope tracking means that `apply-as-transformer` can no longer be implemented using `local-expand`. This PR introduces a new, primitive operation in its place with some additional features.

```
(syntax-local-apply-transformer
  transformer binding-id context-v intdef-ctx arg ...) -> any
```

Applies the procedure `transformer` to the `arg ...`s while manipulating the expand context and the scopes on the arguments and return values for hygiene. The arguments and returns may be any value; scopes are manipulated only for those that are syntax objects. The hygiene operations combine the behavior of local-expand and transformer application. The `context-v` is as in `local-expand`, and the intdef-ctx is a first-class definition context or `#f`.

When the `binding-id` is an identifier, it must be bound (but may be out of context). The expand context's frame-id is set to that associated with the binding. The `current-module-code-inspector` is set to the weaker inspector between that associated with the binding and that of the current expand context. When `#f` is provided as `binding-id`, the frame-id optimization is disabled and the `current-module-code-inspector` is set to `#f`, equivalent to an inspector with access to no bindings.

##### Changed operation

`syntax-local-apply-transformer` is designed to replace `local-apply-transformer`. The new operation supports multiple arguments and returns as well as the `binding-id` argument and associated behaviors.

I propose to document `local-apply-transformer` as "for backwards compatibility only". However, I have changed its implementation to call `syntax-local-apply-transformer`.

### Arguments supporting lists of definition contexts

Several of the `local-expand` variants and `syntax-local-*` operations either accept a list of definition contexts as an alternative to a single context, or include an extra argument providing such a list.

With the new approach to use-site scopes, expansion needs to know which definition context to record newly-created use-site scopes with. A list of definition contexts doesn't provide this information.

With the current hygiene model and API, it appears that support for lists of definition contexts in these arguments is not necessary. I propose to document as "for backwards compatibility only":

* A list of definition contexts rather than a single definition context or `#f` in the `intdef-ctxs` argument of `syntax-local-value` and `local-expand` and its variants
* The `extra-intdef-ctxs` argument to `syntax-local-bind-syntaxes

Some (lengthy) historical context describing the original purpose of these behaviors:

One purpose for using a list of definition contexts may have been to add inside-edge scopes (or previously, renamings) for definition contexts representing outer scopes in a macro that uses a chain of definition contexts, as in one that supports `define*`. With my proposed API, macros will use `internal-definition-context-add-scopes` to add outside-edge scopes, and inside-edge scopes will thus be added at the same time. This obviates the need to add inside-edge scopes other than that corresponding to the innermost definition context at `local-expand` or `syntax-local-value`

It looks like `local-expand`'s support for a list of definition contexts was added to support `package` and `define*` under the marks and renamings hygiene model (<https://github.com/racket/racket/commit/2b8b10dd400a833a0b8632411f91b01c48e61796>). I can't reconstruct exactly why it was necessary, but this PR includes an implementation of a block form with support for `define*` to show that the new model supports this situation. I've included a test analogous to the one for the original issue with `package`: <https://github.com/racket/racket/blob/c2412547f8b2d40340c0fe27ed6dca1d0db04b45/pkgs/racket-test-core/tests/racket/macro.rktl#L2833-L2844> (The current implementation of compatibility/package does not use first-class definition contexts.)

Another purpose for using a list of definition contexts is suggested by the conjunction of these tickets:

* <https://github.com/racket/racket/pull/2043>
* <https://github.com/racket/racket/commit/6834e4a12c2294263c2b9b9fb51339c02c50db53>
* <https://github.com/racket/racket/commit/3531cb24f4ca181927133cf33a1245c5660678d8>

In the initial port of the expander from C to Racket, only bindings from immediate definition contexts were added to the expander environment; those from parent definition contexts were not included. It looks like support for a list of definition contexts was added to operations such as `syntax-local-value` to allow access to bindings from all the contexts, rather than just the innermost. Soon after the expander was fixed to match the behavior of the C expander and include bindings from the parent contexts, but support for lists of definition contexts remains.

## Backwards compatibility

All definition context macros should eventually be updated to use the new operations for proper hygiene. Until they are updated, most uses will work as they have previously, but some will break. Updating macros to improve their hygiene behavior could itself break backwards compatiblility with unhygienic uses, of course.

This PR includes the changes needed to maintain the previous behavior for macro definitions in the main respository.

### Uses of `syntax-local-identifier-as-binding`

Macros that do not adjust their use of `syntax-local-identifier-as-binding` to pass the appropriate first-class definition context will break when a macro that abstracts over definition forms is defined and used in the same context. For example, without the included changes to `local` this use would fail because the appropriate use-site scope would not be removed:

```
(local [(define-syntax-rule (my-define x v) (define x v))
        (my-define x 5)]
  x)
```

### Lists of definition contexts

When a macro passes a list of definition contexts to `local-expand`, use-site scopes created during that expansion will be recorded with the surrounding expand context rather than with any of the definition contexts. Bindings with `syntax-local-bind-syntaxes` will not remove these use-site scopes, resulting in the same problem with locally defined abstractions over definition forms as above. These macros should be adapted to only pass the innermost definition context to `local-expand`.

### Splicing macros

Racket's hygiene model doesn't get use-site binder hygiene right for forms such as `splicing-let`:

```
#lang racket

(require racket/splicing)

(define x 'good)

(define-syntax-rule (m arg)
  (splicing-let ([arg 'bad])
    x))

(m x) ; => bad
```

Both the let-bindings and definitions that splice from the body eventually expand to `define`, so the surrounding definition context has no way to know which should splice out of the use-site vs which should be contained to the use-site.

This limitation has consequences for the implementation of splicing macros. The next subsections describe the adaptations to `racket/splicing` that were needed to integrate with the API change.

#### Trampolining splicing macros

Use-site scopes will be (improperly) removed from splicing let-bindings by the surrounding definition context when it expands `define` regardless of whether or not the use-site scopes for the outer context and the body of the splicing-let are kept separate. There is no value in using a new first-class definition context to get a separate use-site scope set.

For splicing macros implemented via trampolining, this means it makes sense to `local-expand` without using a new definition-context for the splicing context. It is useful, however, to use the outside-edge scope from a definition context as the scope for the splicing form. Unlike a scope created by `make-syntax-introducer`, the outside-edge scope will be properly removed from syntax quotations when the splicing form is used in the implementation of a macro.

#### Definition-context splicing macros

Unlike the other splicing forms, `splicing-local` expands its local bindings using a definition context, rather than trampolining. This creates an interesting problem: during local-expansion, use-site scopes would be handled correctly due to the changes in this PR. But then use-site scopes would subsequently be removed from the generated definitions. The generated definitions would then fail to properly shadow the temporary bindings from local-expansion, resulting in an ambiguous binding error.

The (unfortunate) solution is to intentionally get use-site binding hygiene wrong during the local-expansion pass. This is what I do in the included change to splicing-local.

### Definition context uses that don't correspond to new scopes

Some previous uses of definition contexts, such as an earlier implementation of `splicing-syntax-parameterize`, used definition contexts to extend the expander environment but did not intend to create a new scope. The `add-scope?` argument to `syntax-local-make-definition-context` supported this situation. This behavior was only useful in combination with `syntax-local-get-shadower`; otherwise there is no way to access the bindings due to their inside-edge scope.

Tracking use-site scopes on a per-definition-context basis is incompatible with this use of definition contexts. I don't think any macros still use definition contexts in this way, so that should be okay.

I propose we document the `add-scope?` argument as for backwards compatibility only.

### pkg-build results

Running a pkg-build found a couple of definition context macros needing a fix for the `syntax-local-identifier-as-binding` change. The packages and my patch branches:

* scribble - https://github.com/michaelballantyne/scribble/tree/defctx
* list-plus - https://github.com/michaelballantyne/list-plus/tree/defctx

The only other packages needing changes to pass the pkg-build were my `ee-lib` DSL expander library and `minikanren-ee` DSL.


michaelballantyne:
(@mflatt, @rmculpepper, and @lexi-lambda might have opinions)

mflatt:
This looks really great! As far as I can tell, it will be ready to merge as soon as the documentation is updated.

mflatt:
Really minor, and you should feel free to ignore: I'd prefer that the changes in "src/expander" never have a newline after `require`, use DrRacket-style indentation for `if`, and use two `;` for comments on their own line (more friendly to Emacs modes). This is well below the level of textual-style detail that I'd normally comment on, especially for the rest of the code base, but I've tried to keep "src/expander" especially consistent.

lexi-lambda:
I have not taken the time to look through absolutely all of this in as much detail as it probably deserves, but overall, this looks great to me. Here are my immediate thoughts from a quick overview:

## Changes to `syntax-local-identifier-as-binding`

> ```racket
> (syntax-local-identifier-as-binding id-stx [intdef-ctx]) → identifier?
> ```
>
> The new optional `intdef-ctx` argument determines which use-site scopes are removed.

As far as I can tell, this operation does not actually need `(syntax-transforming?)` to be `#t` if a first-class definition context is provided, so lumping it into an existing `syntax-local-*` function seems philosophically wrong to me. I’d propose this functionality be exposed through a new function, instead, `internal-definition-context-identifier-as-binding`. (Now there’s a Racket function name if I’ve ever seen one!)

## Lists of first-class definition contexts

> With the current hygiene model and API, it appears that support for lists of definition contexts in these arguments is not necessary. I propose to document as "for backwards compatibility only"

Without other changes, I think this is too aggressive. Suppose I have a macro with a grammar like this:

```racket
(complicated ([foo-id foo-expr] ...)
             ([bar-id bar-expr] ...)
  #:foo special-foo-expr
  #:bar special-bar-expr
  body ...+)
```

And suppose I want `complicated` to have a rather unusual binding structure:

  * The `foo-id`s should be bound in `special-foo-expr`.

  * The `bar-id`s should be bound in `special-bar-expr`.

  * Both the `food-id`s *and* the `bar-id`s should be bound in the `body`s.

This binding structure is strange because it cannot be linearized: neither the `foo-id`s nor the `bar-id`s are bound in a scope that is “inside” the other one, yet both scopes are present around the `body` forms. With a list of definition contexts, this is easy to express: you just provide both definition contexts when expanding the `body`s. But without that capability, it’s not immediately obvious what to do.

One could argue that this structure is strange enough that the Racket macro system does not need to support it. After all, one could (arbitrarily) make the `foo-id` context a parent of the `bar-id` one, then simply avoid applying the `foo-id` context’s *scope* when expanding `special-bar-expr`. This would cause the `foo-id`s to be in context during the expansion of `special-bar-expr`, but otherwise inaccessible without identifier-smuggling shenanigans.

However, this idea that definition contexts should always be linearizable is not one that appears anywhere in the existing documentation, and it is somewhat unintuitive given the fact that *scopes* are quite explicitly unordered and are designed to be able to handle exactly this sort of strange binding structure. Furthermore, I think arranging for all created definition contexts to be (artificially) linearized could be genuinely inconvenient in the construction of particularly complex macros: consider, for example, a macro that has subforms that belong to two different binding spaces (to use Matthew’s recently introduced terminology). The macro’s helper functions might process these subforms independently, creating two definition contexts that have no particular relationship to one another, but nonetheless must both be brought into scope in case a subform has any cross-binding space references.

I think these examples highlight how there are really two meaningfully distinct interpretations given to `local-expand`’s `intdef-ctxs` argument:

  1. The bindings in the given `intdef-ctx` are in scope during the expansion of this subform, so they must be added to the expansion context’s binding table.

  2. The subform being expanded *belongs* to the given `intdef-ctx`.

An example where the former is true but not the latter is when expanding the body of a form like `local`: local definitions’ bindings must be visible within the body, but the body forms do not actually belong to that definition context.

### Proposal: Decouple binding table extension from transformer application

Given all of the above, I propose a change that may be seen as somewhat radical: I think we should add a new primitive for extending the expansion context’s binding table without requiring a fresh transformer application.

This new primitive might have an API like this:

```racket
(syntax-local-call-with-bindings intdef-ctxs proc) -> any
  intdef-ctxs : (or/c internal-definition-context?
                      (listof internal-definition-context?))
  proc : (-> any)
```

This applies `proc` in a context that is identical to the current one, except that the binding mixins from the provided definition contexts are folded into the current expansion context. Giving the user the option to do this directly has a lot of advantages:

  * Functions like `local-expand` don’t need to worry about handling point 1 from above, only point 2, so they can accept a single definition context instead of a list.

  * Functions like `syntax-local-value` no longer need to worry about internal definition contexts at all.

    (Currently, there are actually a number of functions littered throughout Racket that don’t take an `intdef-ctxs` argument but really ought to, such as the `static` syntax class from `syntax/parse`. Threading definition contexts through all these functions is a huge pain, and `syntax-local-value` ultimately just adds the given contexts’ bindings to the current expansion context, anyway.)

  * Macro authors don’t have to take nearly as much care to pass definition contexts to all the functions in the `syntax-local` zoo that need them. Rather, while running phase 1 code that is morally “inside” the definition context, they can simply parameterize the binding environment appropriately.

This is something of a departure from the design of existing macro system APIs, since historically, the expander has offered no way to modify the expansion context except by trampolining through another macro invocation. However, I think the things we’ve been doing lately really encourage moving in this direction, and I think this is in line with a broader vision of someday having a more modular expander.

Indeed, I would advocate for the addition of further functions that expose other aspects of the expansion context in more direct ways. For example, I’d be interested in a `syntax-local-call-with-introduction-scope` that directly sets the macro-introduction scope used by `syntax-local-introduce` to an arbitrary value, or a `syntax-local-call/capture-lifts` function that captures uses of `syntax-local-lift-expression` that occur while evaluating an arbitrary thunk. Of course, these things don’t have to happen now, they’re just examples of things we could do in the future.

## The `add-scope?` option

> Some previous uses of definition contexts, such as an earlier implementation of `splicing-syntax-parameterize`, used definition contexts to extend the expander environment but did not intend to create a new scope. The `add-scope?` argument to `syntax-local-make-definition-context` supported this situation.
>
> […]
>
> I propose we document the `add-scope?` argument as for backwards compatibility only.

Personally, I set `add-scope?` to `#f` all the time, but not because I want to avoid a new scope being created at all. Rather, I use `internal-definition-context-introduce` to add the scope explicitly, since don’t always want the scope to be automatically added by `local-expand`. The reason is the same one I mentioned in the previous section: I need the bindings to be brought into scope, but the expression to be expanded doesn’t actually belong to the definition context.

Fortunately, my proposed `syntax-local-call-with-bindings` primitive supports this use case much more neatly than the `add-scope?` argument does. In situations where I don’t want the scope to be automatically added, I just wouldn’t pass the definition context to `local-expand` in the first place.

michaelballantyne:
Thanks for the comments @lexi-lambda!

> lumping it into an existing syntax-local-* function seems philosophically wrong to me

I can see where you're coming from here. I think it would make the API slightly harder to use, though, with two names rather than one to remember for similar situations. I also can't think of situations where it would be important to use outside of `syntax-transforming?` contexts. By the time expansion is over the fate of use-site scopes should be resolved. Do you have a situation in mind?

> And suppose I want complicated to have a rather unusual binding structure...

I indeed hadn't thought about non-linearizable / DAG-shaped scoping structure, thanks!
Unless I missed something, the solutions you propose don't address the original problem I ran into with lists of definition contexts: my new approach to use-site scopes needs local-expand to record use-site scopes in a single definition context representing the current (conceptual) scope. Given a list, there isn't a good way to choose which.

I think the right way to support DAG-structured scope with my proposal is to allow a first-class definition context to have multiple parent contexts, by widening the domain of the `parent-ctx` argument to `syntax-local-make-definition-context` to accept a list of contexts. That way there is a single current definition context to which the `local-expand`ed syntax "belongs", but bindings from the parent contexts can be available in the expander environment.

Part of my goal with this design is to make first-class definition contexts map one-to-one with our intuitive or conceptual notion of lexical "scopes". Binding as sets of scopes implements each such conceptual scope with a combination of elements: outside-edge and inside-edge scopes, an environment frame, a frame id, and a set tracking use-site scopes created during expansion within the conceptual scope. With this PR definition contexts bundle all those elements together, and macro authors can use definition contexts and their parent relationships to construct their intended scoping structure. So I'm explicitly trying to come up with a model that avoids your interpretation #1 of the `intdef-ctx` argument. Similarly, the `#:add-scope?` argument only makes sense to me given a lower-level interpretation of definition contexts as primarily representing environment frames.

I'd like to move on from this PR (I've been at it for a long time now), so I'm inclined to merge it without support for DAG-shaped scope for the time being unless there are existing macros that depend on it. Support for multiple parent contexts can be added later. Would that break anything for you?

lexi-lambda:
> I can see where you're coming from here. I think it would make the API slightly harder to use, though, with two names rather than one to remember for similar situations. I also can't think of situations where it would be important to use outside of `syntax-transforming?` contexts. By the time expansion is over the fate of use-site scopes should be resolved. Do you have a situation in mind?

I don’t have a particular use case in mind, but it isn’t clear to me whether or not `syntax-local-identifier-as-binding` has a legitimate purpose *without* providing a definition context argument. Therefore, if the single-argument use is effectively deprecated, it seems conceptually cleaner to just introduce a separate function. However, if it isn’t deprecated, and it has some legitimate use I’m not aware of, I’m much more partial to your counterargument.

> I indeed hadn't thought about non-linearizable / DAG-shaped scoping structure, thanks!

I think my focus on non-linearizable contexts in my previous comment may have a mistake—to be honest, while they are interesting, they’re really only a small portion of the problem I’m alluding to.

Consider `local`. Its grammar is this:

```racket
(local [local-defn ...] body ...+)
```

We want to implement `local` using a first-class definition context for the `local-defn`s, but *not* for the `body` forms, which we can just wrap in an empty `let` and expand as an expression. Clearly, the `local-defn`s *belong* to the definition context, so when we call `local-expand`, we need to pass the definition context as an argument in order to (among other things) accumulate use-site scopes. But when we expand `(let () body ...)`, we aren’t in the definition context anymore, but we still need those bindings to be in scope.

Perhaps this isn’t especially compelling to you because the outer `let` around the `body` forms will inevitably cause the expander  to switch into a separate context, anyway. Fair enough—but then let’s consider a weirder variant, `weird-local`. `weird-local` is just like `local`, but while the `local-defn`s are *in context* in the `body` forms, they aren’t automatically *in scope*. That is, the definition context’s scopes are *not* added to the `body` forms, but identifier-smuggling macros might still reference the created bindings. There is no way to express this sort of thing right now without setting the `add-scope?` argument to `#f` and manually managing introduction of the scope.

Perhaps this still isn’t compelling to you, because you thing `weird-local` is simply too weird to be worth supporting. Again, fair enough—but `weird-local` is ultimately a simplified representative of more complicated yet still quite useful applications of first-class definition contexts.

> Part of my goal with this design is to make first-class definition contexts map one-to-one with our intuitive or conceptual notion of lexical "scopes". Binding as sets of scopes implements each such conceptual scope with a combination of elements: outside-edge and inside-edge scopes, an environment frame, a frame id, and a set tracking use-site scopes created during expansion within the conceptual scope. With this PR definition contexts bundle all those elements together, and macro authors can use definition contexts and their parent relationships to construct their intended scoping structure. So I'm explicitly trying to come up with a model that avoids your interpretation #1 of the `intdef-ctx` argument. Similarly, the `#:add-scope?` argument only makes sense to me given a lower-level interpretation of definition contexts as primarily representing environment frames.

I think the heart of my quibbles is that I disagree with this conclusion. As demonstrated above, even if you treat a first-class definition context as a first-class definition context, there are still situations where you want to bring those bindings into context to expand a subexpression (not necessarily a sub-definition context, so you can’t rely on the `parent-ctx` chain!) that does not belong to the definition context.

`local` is a canonical example of a macro that codifies this idea that a form does not have to be inside a definition context to see its bindings, and in fact a definition context *can* be intuitively viewed as a “collection of bindings” in a similar sense to binding pairs in a `let` form are, and I think that perspective is one that is important to continue to support without necessarily having to trampoline into a sub-expansion context within the definition context using `#%expression` or something like that.

> Unless I missed something, the solutions you propose don't address the original problem I ran into with lists of definition contexts: my new approach to use-site scopes needs local-expand to record use-site scopes in a single definition context representing the current (conceptual) scope. Given a list, there isn't a good way to choose which.

I don’t think this is a problem with anything I presented in my previous comment, because there is still always exactly one definition context that any given form belongs to, which is the one that should be passed to `local-expand`. Other definition contexts’ bindings merely need to be brought into context during the expansion.

> I think the right way to support DAG-structured scope with my proposal is to allow a first-class definition context to have multiple parent contexts, by widening the domain of the `parent-ctx` argument to `syntax-local-make-definition-context` to accept a list of contexts. That way there is a single current definition context to which the `local-expand`ed syntax "belongs", but bindings from the parent contexts can be available in the expander environment.

I wouldn’t be opposed to this idea in general, but I still really think the `parent-ctx` argument makes sense as a convenience feature. I don’t think definition contexts should need to have any hierarchical structure. Scopes don’t, and I don’t think there’s any way to justify definition contexts needing it independently. As far as I can tell, nothing in your design depends on it, and imposing this requirement on macro authors seems to mostly just make things more complicated—relative to giving them a primitive to bring the bindings into scope explicitly—for no gain.

> I'd like to move on from this PR (I've been at it for a long time now), so I'm inclined to merge it without support for DAG-shaped scope for the time being unless there are existing macros that depend on it. Support for multiple parent contexts can be added later. Would that break anything for you?

Would it actually break any programs I’ve written? Probably not. I suspect all my programs would continue to do the same thing because the differences in behavior are not significant enough to actually cause issues in practice. However, it very much *would* cause programs of mine to be “philosophically incorrect,” in that they use features like passing lists of definition contexts to `local-expand` and setting `add-scope?` to `#f` with no migration path to a non-deprecated API.

michaelballantyne:
@lexi-lambda 

> However, if it isn’t deprecated, and it has some legitimate use I’m not aware of, I’m much more partial to your counterargument.

I think there remain legitimate uses without a definition context argument. For example, an abstraction over `define` might want to construct an identifier that will match the final definition binding for use as a key in a free- or bound-identifier table.

> I don’t think this is a problem with anything I presented in my previous comment, because there is still always exactly one definition context that any given form belongs to, which is the one that should be passed to local-expand. Other definition contexts’ bindings merely need to be brought into context during the expansion.

Sorry, I had misread something previously. Now it makes sense.

However,

> `weird-local` is just like `local`, but while the `local-defns` are in context in the body forms, they aren’t automatically in scope... There is no way to express this sort of thing right now without setting the `add-scope?` argument to `#f` and manually managing introduction of the scope.

I think I understand where we were failing to connect now. The above suggests you use the definition context scope as an outside-edge scope: one that should be applied to bindings and references in a context before macro expansion moves anything, but not to syntax that is moved by expansion into the definition context. The default `add-scope?` `#t` behavior corresponds instead to an inside-edge scope: one that should be added to any identifier that ends up in the definition context, whether there originally or by expansion. And a separate outside-edge scope was missing.

With the PR definition contexts now come with both outside- and inside-edge scopes. To get the behaviors you're looking for, a macro should manipulate the outside-edge scope with `internal-definiton-context-add-scopes` rather than the inside-edge with `internal-definition-context-introduce`. `weird-local` should add the outside-edge scope only to the `local-defns` and not the body forms. The inside-edge scope for the definition context isn't needed on the body forms in this (weird) form, but it does no harm because the absence of the outside-edge scope ensures the `local-defns` bindings are out of scope.

lexi-lambda (changes requested):
> I think there remain legitimate uses without a definition context argument. For example, an abstraction over `define` might want to construct an identifier that will match the final definition binding for use as a key in a free- or bound-identifier table.

Alright, that’s fair—I don’t feel as strongly about this point, then.

I do still think that splitting this functionality out into a separate function would be more consistent with the design of the rest of the syntax API. For example, we have `syntax-local-introduce` and `internal-definition-context-introduce`, not an optional `intdef-ctx` argument to `syntax-local-introduce`. When an optional `intdef-ctx` argument *is* present, it usually augments the functionality that the function already performs, not replace it completely.

In your design, it sounds like `syntax-local-identifier-as-binding` does a fundamentally different thing when given an `intdef-ctx` argument; that is, it no longer removes the use-site scopes that it would have without the argument. I think that’s confusing and inconsistent, so I would still prefer the separate function, but it isn’t a strong enough preference that I will die on this hill.

> I think I understand where we were failing to connect now. The above suggests you use the definition context scope as an outside-edge scope: one that should be applied to bindings and references in a context before macro expansion moves anything, but not to syntax that is moved by expansion into the definition context. The default `add-scope?` `#t` behavior corresponds instead to an inside-edge scope: one that should be added to any identifier that ends up in the definition context, whether there originally or by expansion. And a separate outside-edge scope was missing.
> 
> With the PR definition contexts now come with both outside- and inside-edge scopes. To get the behaviors you're looking for, a macro should manipulate the outside-edge scope with `internal-definiton-context-add-scopes` rather than the inside-edge with `internal-definition-context-introduce`. `weird-local` should add the outside-edge scope only to the `local-defns` and not the body forms. The inside-edge scope for the definition context isn't needed on the body forms in this (weird) form, but it does no harm because the absence of the outside-edge scope ensures the `local-defns` bindings are out of scope.

Yes, I think I better understand the disconnect now. Unfortunately, the remaining confusion appears to stem from the fact that—as far as I can tell—your description does not actually line up with the implementation.

Before your PR, the `add-scope?` option controlled whether or not the definition context’s scope was added to the syntax object provided to `local-expand` both before **and** after its expansion. See this line: https://github.com/racket/racket/blob/f1e6ac414707270edd4a6647008a4a903fe7d12e/racket/src/expander/expand/local-expand.rkt#L118

This means the existing behavior essentially treats the definition context’s scope as both an inside- and outside-edge scope.

With your PR, definition contexts now have separate inside- and outside-edge scopes. On paper, that’s a great idea—but the aforementioned line in `do-local-expand` is still present, which means `add-scope?` will now cause what is supposed to be the inside-edge scope to be added to everything expanded in the context *before* it’s expanded, which is obviously wrong!

So at the very least, something doesn’t quite line up between your description of what you think the mental model *should* be and what the implementation actually does. But I don’t know how to fix it: changing the meaning of the `add-scope?` option to do what you describe sounds extremely backwards-incompatible to me. Regardless, I think this PR as-written isn’t quite right.

lexi-lambda (commented):
Alright, Michael and I had a long phone call about this hashing the details out, and he successfully convinced me of both the correctness and backwards-compatibility of this design. Here are the main subtleties discussed:

  * It’s perfectly backwards-compatible to think of the scope historically associated with first-class definition contexts as being exclusively the inside-edge scope. The fact that `local-expand` only adds the inside-edge scope is fine, as that is consistent with the historical (mis)behavior.

  * Historically, the default `#t` value for `add-scope?` has allowed macro authors to rely on `local-expand` to introduce syntax into the definition context. However, going forward, correct use of first-class definition contexts will always require explicit use of `internal-definition-context-add-scopes` to add the outside-edge scope.

  * Given the above point, setting `add-scope?` to `#f` is unnecessary, since the effect I mentioned using it for in previous comments is accomplished by simply omitting the outside-edge scope. The addition of the inside-edge scope in such situations is harmless, since the bindings’ accessibility is determined by other scopes.

  * The fact that parent definition contexts don’t have their inside-edge scope automatically added during expansion is fine, since the “late” addition of the inside-edge scope is only necessary to reference bindings introduced in the current definition context. (Adding parents’ inside-edge scopes would be harmless, but omitting them is a valid optimization.)

  * My proposed `syntax-local-call-with-bindings` operation goes in roughly the right direction, but it probably isn’t quite right, because it doesn’t adjust the way the expansion context tracks use-site scopes, nor does it add the inside-edge scope for forms expanded immediately within the context. However, a more complete version of this operation can already be effectively implemented using `syntax-local-apply-transformer` (since the addition of the inside-edge scope is now harmless).

Having discussed all those details, my fears in my previous comment are assuaged, so I’m mostly satisfied with the changes in this PR as-written. Here are my (comparatively minor) lingering quibbles:

  1. We both agreed that the domain of `internal-definition-context-remove-scopes` ought to be restricted to an identifier for the same reasons `syntax-local-identifier-as-binding` is. We also agreed that it makes more sense to give it a more purpose-oriented name, like `internal-definition-context-identifier-for-splice`, since it’s really only intended to be used to implement splicing forms.

  2. I personally wish `internal-definition-context-add-scopes` also had a more purpose-oriented name that captured the fact that it is intended to be applied to *the original forms* that make up the definition context, not applied to the results of macroexpansion in the context. (That is, its primary intent is to add the outside-edge scope.) This would make its relationship to `internal-definition-context-introduce` much less confusing.

  3. I still have a preference for a separate `internal-definition-context-identifier-as-binding` operation for somewhat philosophical reasons (and I think it could pair nicely with `internal-definition-context-identifier-for-splice`), but I don’t feel especially strongly about this.

---

@michaelballantyne, one thought I had while typing this up is that it could make sense to replace `internal-definition-context-add-scopes` with an operation that is like `internal-definition-context-introduce`, but exclusively for the outside- rather than the inside-edge scope. I think that would satisfy both our goals:

  * It is more consistent with the rest of the API and gives programmers complete control, if they really want it.

  * Macro authors still only need to call that one function, since the inside-edge scope will get added by `local-expand` anyway.

In an ideal world, I think we would want to call such an operation `internal-definition-context-introduce`, but we can’t reuse that name without breaking backwards-compaitibility. So perhaps we could use a name like `internal-definition-context-original-introduce`, which captures the idea that it should be applied to syntax that originates from the definition context.

michaelballantyne:
> It could make sense to replace `internal-definition-context-add-scopes` with an operation that is like `internal-definition-context-introduce`, but exclusively for the outside- rather than the inside-edge scope... Macro authors still only need to call that one function, since the inside-edge scope will get added by `local-expand` anyway.

`local-expand` only adds the `inside-edge` scope of the current definition context. Given a chain of definition contexts, only `internal-definition-context-add-scopes` ensures that the inside-edges from the outer contexts are added.

michaelballantyne:
@lexi-lambda, @mflatt any feelings on:

* `internal-definition-context-splice-binding-identifier` rather than `internal-definition-context-remove-scopes`
* `internal-definition-context-add-initial-scopes` rather than `internal-definition-context-add-scopes`

?

lexi-lambda:
> `internal-definition-context-splice-binding-identifier` rather than `internal-definition-context-remove-scopes`

I don’t like this name because I think it sounds like it should splice an identifier into the given context, but it’s actually used to build an identifier that can splice into a surrounding context. So I’d prefer a name like `internal-definition-context-identifier-for-splice`.

If you don’t like that name because it sounds like it should access an identifier in the context rather than be an operation on an identifier, you could try something like `internal-definition-context-prune-for-splice`, but I still prefer the former suggestion. This is really just bikeshedding, though.

Of perhaps some interest is the (deprecated) `identifier-remove-from-definition-context` function, which uses an alternate naming scheme that starts with `identifier` rather than `internal-definition-context`. In theory, I think you could even justify commandeering `identifier-remove-from-definition-context` to also remove use-site and the outside-edge scopes (which would undeprecate it), since removing the context’s use-site scopes on such an identifier is harmless and would not cause backwards-compatibility issues (and legacy code would never add the outside-edge scope in the first place, so removing it would be a no-op). However, I don’t especially like that function’s name, either, because to me it still suggests that it somehow mutates the definition context, when really all it does is manipulate the identifier’s scopes.

In any case, I’m basically fine with any of those three suggestions.

>  `internal-definition-context-add-initial-scopes` rather than `internal-definition-context-add-scopes`

I’ve been thinking more about this, and honestly, the more I wish we could just change `internal-definition-context-introduce` to manipulate both inside- and outside-edge scopes simultaneously. From a macro author’s perspective, the context is really represented by both scopes, and I can think of essentially no (non-contrived) circumstances in which one would want to explicitly add or remove one scope but not the other. Alas, it would be backwards-incompatible (though I suspect very little code would actually break in practice).

So I am backtracking on my earlier opinion: I don’t think it’s all that helpful to use the word “original” or “initial” in this function’s name. I think I mostly just don’t like how weird `internal-definition-context-add-scopes` is relative to all the other operations that manipulate scopes, all of which have the word `introduce` in their name and allow addition, removal, or flipping.

All this makes me somewhat inclined to completely give up and suggest `internal-definition-context-introduce*`. Yes, it really sucks, but does it suck any less than anything else? I don’t think so—and hey, even `syntax/parse` has `~or*`. Maybe sometimes it’s best to just admit we screwed up.

michaelballantyne:
Documentation is in and I squashed to a single commit with a descriptive message. I think this is ready to merge. @mflatt

I stuck to `internal-definition-context-add-initial-scopes`, but went with `internal-definition-context-splice-binding-identifier`. I acknowledge the imperfection of the name but still prefer it to the others. Hopefully the documentation makes the meaning clear.

mflatt:
This looks generally good to me.

My only request before you merge is about the documentation for `syntax-local-apply-transformer`. The word "hygiene" is not currently used anywhere else in the reference (to avoid having to define it). So, probably it should not be used here, and the documentation for `syntax-local-apply-transformer` should just say what it does in terms of adding and flipping scopes.

For the scribble change, just try to merge it at about the same time, and no version exception is needed on the package server (because, as you say, it's part of the main distribution). But it would be good to bump the version dependency, if that's not done already.

mflatt:
Thanks for the doc update. I've built and tested with the latest commit on my machine, but adjusted to bump the version to 8.2.0.7 (and with the Scribble commit). Should I merge that, assuming no problems discovered by the CI run that's in progress?

michaelballantyne:
@mflatt does this look better?

It doesn't look like the scribble packages specify a minimum version dependency in their info.rkt. Should I add one, or does it make sense as-is?

mflatt:
@michaelballantyne Yes — I think our messages crossed, so see above.

Adding a minimum version for `"base"` to `racket-lib` would be a good improvement for that commit.

michaelballantyne:
@mflatt oh, they did indeed cross! Feel free to merge.

mflatt:
Ok - merged!

michaelballantyne:
Thanks! For future reference, do you bump the version number when adding new expander primitives? I wasn't clear on that vs vm primitives.

mflatt:
Yes. There are a variety of reasons to bump the version number, but one of them is that `#%kernel` has new exports, so `racket/base` needs to be recompiled to propagate them, etc.

Issue #3929, opened 2021-07-22
raco make silently discards --disable-constant flag when using parallel build tasks


djholtby:
**What version of Racket are you using?**
8.2 [cs]

**What program did you run?**
`raco make -j 4 --disable-constant *.rkt`

**What should have happened?**

*.rkt files were compiled with constant enforcement disabled

**What does happen?**

`--disable-constant` is ignored when using `-j` with 2 or more tasks

Can be tested by attempting to `dynamic-rerequire` the compiled modules.  Without disabling constant enforcement, subsequent reloading of the module will result in an error.

**Please include any other relevant details**

It looks like the issues are

* raco make does not pass along the option for --disable-constant to `parallel-compile-files`
* `parallel-compile-files` doesn't pass its options to the workers 
* the workers don't recognize an option for disabling constants

mflatt:
Closed by #3930

PR #3930, opened 2021-07-22
add option to disable constant enforcement for parallel-compile-files, and have raco make use this option when using parallel builds

djholtby:
Proposed solution for #3929


samth:
Is the first commit here intended/related?

djholtby:
> Is the first commit here intended/related?

Hmmm, no.  I had a pr from years ago that I guess I merged instead of deleting.  Simpler times.  Let me see if I can fix that

djholtby:
(apparently not...)

djholtby:
OK I learned a lot about how to use rebase (and also how not to use rebase).

mflatt:
Thanks!

PR #3934, opened 2021-07-23
Avoid excess dependencies in `syntax/parse/lib/function-header`.

samth:
This causes `racket/match` to no longer depend on `racket/contract`.

cc @rmculpepper

### DIFF
--- a/racket/collects/racket/match/define-forms.rkt
+++ b/racket/collects/racket/match/define-forms.rkt
@@ -198,5 +198,5 @@
          [(_ ?header:function-header ?clause ...)
           (quasisyntax
            (define ?header
-             (match*/derived (?? ?header.params) #,stx
+             (match*/derived (~? ?header.params) #,stx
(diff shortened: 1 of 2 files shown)


shhyou:
The `??` after `match*/derived` needs to be `~?`.

racket/collects/racket/match/define-forms.rkt
```diff
@@ -197,7 +196,7 @@
      (define-syntax (define/match stx)
        (syntax-parse stx
          [(_ ?header:function-header ?clause ...)
-          (quasitemplate
+          (quasisyntax
            (define ?header
              (match*/derived (?? ?header.params) #,stx
```
samth, line 201:
```suggestion
             (match*/derived (~? ?header.params) #,stx
```

samth:
@rmculpepper is this ok to merge?

rmculpepper:
LGTM

PR #3935, opened 2021-07-23
Add `pattern-bound-ids`.

samth:


pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -758,14 +758,26 @@ This binding is provided @racket[for-syntax].
...
 
 @defproc[(syntax-local-match-introduce [stx syntax?]) syntax?]{
 For backward compatibility only; equivalent to @racket[syntax-local-introduce].
 
+This binding is provided @racket[for-syntax].
+
 @history[#:changed "6.90.0.29" @elem{Made equivalent to @racket[syntax-local-introduce].}]}
 
 
+@defproc[(pattern-bound-identifiers #:legacy? [legacy? any/c #f] [pat syntax?]) (listof identifier?)]{
```
jackfirth, line 772: Is there a benefit to allowing non-boolean values for `#:legacy?`? It doesn't seem worth making the API harder to understand.

jackfirth: What do you think of renaming this to `match-pattern-bound-identifiers`? Since we've also got `syntax-case` and `syntax-parse` patterns.

pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -758,14 +758,26 @@ This binding is provided @racket[for-syntax].
...
+Parse @racket[pat] as a pattern, and compute the identifiers bound by
+that pattern. If @racket[legacy?] is @racket[#f], use the legacy match
```
jackfirth, line 774: Did you mean for this to be `#t` instead of `#f` here? Also, is there any reason users outside the Racket implementation would _need_ the legacy behavior? We could just leave this undocumented in the public API and only document it with a comment in the implementation.

samth: Yes, I mean `#t`.

pkgs/racket-doc/scribblings/reference/match.scrbl
samth: Allowing generalized booleans seems conventional, which is why I'm doing it.

samth: I like that renaming; I'll try to update soon.

sorawee: For the sake of being consistent with the rest of Racket, I think it should allow non-boolean.

See various string functions that accept keyword arguments like `#:all?`, `#:trim?`, etc. as examples

```
(string-replace str from to [#:all? all?]) → string?
str : string?
from : (or/c string? regexp?)
to : string?
all? : any/c = #t
```

pkgs/racket-doc/scribblings/reference/match.scrbl
samth: As for why you'd use it, the legacy syntax is public API. If you `(require mzlib/match)` you can still use it too. :)

Maybe there's no actual need to support it at all, though.

jackfirth: Nobody's asked for this feature for legacy patterns, and if they do I think it'd probably be better to just make a separate copy of the function and export it from `mzlib/match`. No need to _preemptively_ make legacy APIs.

pkgs/racket-doc/scribblings/reference/match.scrbl
jackfirth: Alright, `any/c` on the basis of consistency seems reasonable to me. It still strikes me as weird though to allow non-booleans in places where there's no expectation that any useful data could be passed through. Probably too late to rock that boat for core Racket APIs.

PR #3938, opened 2021-07-24
rerequire: fix reloading of module paths with up indicators

Bogdanp:
After  14f0f86b793a9fa3b77176df43c09007b2f1f3e0, I noticed that changes to paths containing up indicators were not being detected. After fixing that problem (by simplifying the `npath`s before they're looked up in the cache), I noticed that the changed modules wouldn't reload properly.

The second problem was caused by the switch to `resolve-module-path-index` from `module-path-index-resolve`. The former doesn't simplify the paths it resolves so the `current-module-declare-name` of reloaded modules whose paths weren't simplified would end up being wrong.

```
> (module-path-index-resolve (module-path-index-join "../a.rkt" #f))
#<resolved-module-path:"/Users/a.rkt">
> (resolve-module-path-index (module-path-index-join "../a.rkt" #f))
#<path:/Users/bogdan/../a.rkt>
```

This change fixes both of those problems and removes a couple pieces of dead code from `check-latest`.


mflatt:
Thanks!

PR #3939, opened 2021-07-25
rerequire: fix reloading for diamond dependencies

Bogdanp:
With a dependency graph like

    a -> b
    a -> c
    b -> d
    c -> d

where changes were made to `d`, say `check-latest` checked the
dependencies in this order:

    a b d c d

while checking `b`, it would reload `d` and mark it as loaded, then
reload `b` and move on to `c`.  At `c`, it would check `d` again, but
because it had already marked `d` as loaded, it would skip `d` so `c`
wouldn't get reloaded either.

This change fixes that problem by reusing the reloaded cache to
determine whether or not a dependency of some module was already
reloaded.

### DIFF
--- a/pkgs/racket-test/tests/racket/rerequire.rkt
+++ b/pkgs/racket-test/tests/racket/rerequire.rkt
@@ -202,3 +202,75 @@ CODE
    path1)
 
   (check-equal? ((reload)) 2))
+
+
+(test-case "ordering"
+  (define dir (make-temporary-file "rereq-~a" 'directory))
+  (define path1 (build-path dir "a.rkt"))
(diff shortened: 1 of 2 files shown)


mflatt:
I think the new tests are meant to count as passing, but because `re-loading` messages are printed to stderr, `raco test --drdr` treats the test as failing. Maybe it's just that `#:verbosity 'reload` needs to be removed from the test.

Bogdanp:
Whoops, I meant to reduce the verbosity there but forgot. Should be fixed now. Thanks!

mflatt:
Thanks!

PR #3942, opened 2021-07-27
expander: Fix compilation errors when installed as a package

lexi-lambda:
These modules don’t appear to actually be used in the built expander, but they cause `raco setup` to report errors if `expander` is installed as a package, which I find annoying. (I like having `expander` installed as a package because it lets me use “Open Require Path” in DrRacket to open expander modules.)


mflatt (changes requested):
Probably it's better to just remove these. I think they're leftovers from before the reader was implemented in the expander (or something like that).

PR #3943, opened 2021-07-27
Add links to useful blog posts in CONTRIBUTING.md

jackfirth:
This change adds links to two Racket blog posts for first-time contributors to see: the tutorial on how to contribute to Racket, and the instructions for testing changes against the entire package catalog.


samth:
I think the discussion of pkg-build should say that it's not usually needed/only for changes that might have compatibility problems.

PR #3946, opened 2021-07-29
`define-runtime-module-path-index`: collaborate with compilation

samth:


mflatt:
I agree that there's an issue here: `define-runtime-module-path` creates a compilation-friendly link between modules, and `define-runtime-module-path-index` does not. In many cases, that link is helpful, while forcing module loading is not, so a third point in the spectrum would be good. Meanwhile, some uses of `define-runtime-module-path-index` might not mean to refer to modules that already exist at runtime, so maybe `define-runtime-module-path-index` should not become that third point without a specific option.

For the implementation, I think you probably meant to make the boolean `#t` for `define-runtime-module-path-index`. Also, a module dependency would be closer to `register-external-module` than `register-external-module` — but even `register-external-module` seems wrong, because it deals with absolute paths instead of relative module paths.

Maybe a better approach is for `define-runtime-module-path-index` (or something like it) to generate a submodule that has a `for-label` references to target module paths. That approach would get relative module paths right for compilation, but it would not force the modules referenced via `for-label` to be loaded with the enclosing module.

PR #3950, opened 2021-08-02
net/ftp: close both ports on exception to avoid leaking descriptors

dzoep:
While developing a service that communicates with remote systems via FTP using the net/ftp module, we encountered a situation where the Racket process was suddenly unable to open any new sockets (not just TCP for FTP - but any other sockets for DNS, DB connections etc...). After some extensive logging we noticed the number of open descriptors for the process is steadily increasing. The reason for this was a misbehaving remote server which closed connections randomly and if we waited enough time, it just exhausted the limit of open descriptors for our service.

Platform: Linux (probably all platforms though)

Originally found and preliminary investigation of the issue: Dominik Pantůček

Further investigation and proposed patch: Adam Zaiter <adam.zaiter@trustica.cz> (as a part of his work on this project)

More information from him follows:

## Reproducing the bug
 
Misconfigure the FTP server, such that it is possible to 
establish a connection, but every other action produces a 500 status
code (for example `500 OOPS: cannot read user list`).
Start the FTP server on localhost and connect to it.

`ftp-conn-open` exception will be raised, which is originally `ftp: expected result code 220, got 500 OOPS: cannot read user list`

The exception will trigger every 10 seconds, opening 5 new file 
descriptors (can be checked with `ls /proc/<PID>/fd | wc -l`).
Eventually, the soft limit for open files will be reached and 
exceptions `system error: Too many open files; errno=24` and ` tcp-connect: host not found` will be raised.

When looking at output of `lsof -i -P -n | grep his`, you can see 
TCP sockets in (CLOSE_WAIT) state starting to add up.

## The issue

The problem is in the `net/ftp` package's `ftp-establish-connection` function. 
A TCP connection is established [here](https://github.com/racket/racket/blob/master/pkgs/net-lib/net/ftp.rkt#L155).
Then in `ftp-esablish-connection*`, `ftp-check-response` is called
with the expectation of a 220 status code. When the code is 500,
[this](https://github.com/racket/racket/blob/master/pkgs/net-lib/net/ftp.rkt#L32) exception is raised. However, the TCP connection is 
still opened, and there is no way for the user to close it, as 
the ports are not returned.

## Solution

The solution is to close the ports, **then** raise the exception
to the user as can be seen below:

```racket
(define (ftp-establish-connection* in out username password)
  (with-handlers ([exn:fail? (λ (e)
                                (close-input-port in)
                                (close-output-port out)
                                (raise e))])
  (ftp-check-response in out #"220" void (void))
  (fprintf out "USER ~a\r\n" username)
  (let ([no-password? (ftp-check-response
                        in out (list #"331" #"230")
                        (lambda (line 230?)
                          (or 230? (regexp-match #rx#"^230" line)))
                        #f)])
    (unless no-password?
      (fprintf out "PASS ~a\r\n" password)
      (ftp-check-response in out #"230" void (void))))
  (make-ftp-connection in out)))
```

### DIFF
--- a/pkgs/net-lib/net/ftp.rkt
+++ b/pkgs/net-lib/net/ftp.rkt
@@ -139,17 +139,21 @@
 ;; need PASS command. "or 230? (rege..." means if 230? is true already
 ;; , then do not check the line anymore, it's just true.
 (define (ftp-establish-connection* in out username password)
-  (ftp-check-response in out #"220" void (void))
-  (fprintf out "USER ~a\r\n" username)
-  (let ([no-password? (ftp-check-response
-                       in out (list #"331" #"230")
-                       (lambda (line 230?)
-                         (or 230? (regexp-match #rx#"^230" line)))
-                       #f)])
-    (unless no-password?
-      (fprintf out "PASS ~a\r\n" password)
-      (ftp-check-response in out #"230" void (void))))
-  (make-ftp-connection in out))
+  (with-handlers ([exn:fail? (λ (e)
+                               (close-input-port in)
+                               (close-output-port out)
+                               (raise e))])
+    (ftp-check-response in out #"220" void (void))
+    (fprintf out "USER ~a\r\n" username)
+    (let ([no-password? (ftp-check-response
+                         in out (list #"331" #"230")
+                         (lambda (line 230?)
+                           (or 230? (regexp-match #rx#"^230" line)))
+                         #f)])
+      (unless no-password?
+        (fprintf out "PASS ~a\r\n" password)
+        (ftp-check-response in out #"230" void (void))))
+    (make-ftp-connection in out)))
 
 (define (ftp-establish-connection server-address server-port username password)
   (let-values ([(tcpin tcpout) (tcp-connect server-address server-port)])
--- a/pkgs/net-test/tests/net/ftp.rkt
+++ b/pkgs/net-test/tests/net/ftp.rkt
@@ -81,7 +81,46 @@
                 (thread-wait pasv3-thd)
                 (thread-wait main-thd)
                 (get-output-string cop) => EXPECTED-USER-OUTPUT
-                ))))
+                ))
+        ;; make sure that TCP connections don't leak when attempting to open
+        ;; an FTP session fails
+        (test (for ([immediate-failure? '(#t #f)])
+                (define bad-custodian (make-custodian))
+                (parameterize ([current-custodian bad-custodian])
+                  (define bad-listener (tcp-listen 0))
+                  (define bad-port (let-values ([(_1 p _2 _3) (tcp-addresses bad-listener #t)]) p))
+                  (thread (lambda ()
+                            (let loop ()
+                              (define-values (i o) (tcp-accept bad-listener))
+                              (cond [immediate-failure?
+                                     (fprintf o "500 OOPS: cannot read user list\n")]
+                                    [else
+                                     ; After password
+                                     (fprintf o "220 FTP ready\n")
+                                     (flush-output o)
+                                     (read-bytes-line i 'any)
+                                     (fprintf o "331 Go ahead\n")
+                                     (flush-output o)
+                                     (read-bytes-line i 'any)
+                                     (fprintf o "530 Cannot login\n")])
+                              (close-output-port o)
+                              (close-input-port i)
+                              (loop))))
+                  (for ([i (in-range 10000)])
+                    ((with-handlers ([exn:fail? (lambda (exn)
+                                                  ;(printf "~s\n" (exn-message exn))
+                                                  void)])
+                       (define conn (ftp-establish-connection server bad-port user passwd))
+                       (thread (lambda ()
+                                 (displayln 'established)))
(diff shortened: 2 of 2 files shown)


mflatt:
Thank you for this repair!

There are some tests for `net/ftp` in `pkgs/net-test/tests/net/ftp.rkt`. Do you have time to add a test there, or should I look into it?

dzoep:
> Thank you for this repair!
> 
> There are some tests for `net/ftp` in `pkgs/net-test/tests/net/ftp.rkt`. Do you have time to add a test there, or should I look into it?

I don't think it is possible to test this directly. My last repair (ftp-current-directory) was easy to test (and I think Matthias pointed that out) so I added a simple "local" test: https://github.com/racket/racket/commit/28105b7be6c34ca1c1ab1ec24d571ce220c07aad

But to test this, you need to run a bit "malicious" FTP server and write an infinite loop FTP client in Racket to connect to it many times. Even our fastest test takes minutes to deplete the descriptors. I don't see a simple, reliable and reasonably fast test for this at the moment. Also beware - the default limits may differ between systems (just different Linux distribution is enough).

mflatt:
Ok, I'll take a look.

My experience is that it's easy to exhaust file descriptor resources on machine where our tests normally run. The limit is usually on the order of 1000, which takes little time to consume when you have a cooperatively misbehaving server (and the test seems to have its own mock server). So, you just loop a client for 1000 or 10000 times with a server that always errors, and if that loop complete without error, then the test passes. If the test runs out of file descriptors, then things go bad for the process, and that counts as failure.

dzoep:
Ok, I'll look into spawning the process for test and see if we can do that. Good idea! I'd just like to see this in 8.3 - for now, we need to restart the service every now and then because the real malicious server is not going to be configured properly anytime soon.

mflatt:
I don't think a new process is really needed. Here's a test that shows the bug on my machine and takes a second or two to run:

```
  ;; make sure that TCP connections don't leak when attempting to open
  ;; an FTP session fails
  (define bad-custodian (make-custodian))
  (parameterize ([current-custodian bad-custodian])
    (define bad-listener (tcp-listen 0))
    (define bad-port (let-values ([(_1 p _2 _3) (tcp-addresses bad-listener #t)]) p))
    (thread (lambda ()
              (let loop ()
                (define-values (i o) (tcp-accept bad-listener))
                (fprintf o "500 OOPS: cannot read user list\n")
                (close-output-port o)
                (close-input-port i)
                (loop))))
    (for ([i (in-range 10000)])
      ((with-handlers ([exn:fail? (lambda (exn)
                                    ;; (printf "~s\n" (exn-message exn))
                                    void)])
         (define conn (ftp-establish-connection server bad-port user passwd))
         (lambda () (error "connection didn't fail; server wasn't bad enough")))))
    ;; we should still be able to connect to the mock server
    (with-handlers ([exn:fail? (lambda (exn)
                                 (eprintf "Failed to open one more connection; out of descriptors?\n")
                                 (raise exn))])
      (tcp-connect server bad-port)))
  (custodian-shutdown-all bad-custodian)
```

I think it would be better to try with an immediate 500 and with a failure on the password step, though.

dzoep:
@mflatt - do you mean testing first 10000x immediate 500 and then 10000x 220->331->530? Like in:

```racket
nections don't leak when attempting to open
        ;; an FTP session fails
        (test (let ()
                (define bad-custodian (make-custodian))
                (parameterize ([current-custodian bad-custodian])
                  (define bad-listener (tcp-listen 0))
                  (define bad-port (let-values ([(_1 p _2 _3) (tcp-addresses bad-listener #t)]) p))
                  (thread (lambda ()
                            (let loop ()
                              (define-values (i o) (tcp-accept bad-listener))
                              ; After password
                              (when #f
                                (fprintf o "220 FTP ready\n")
                                (flush-output o)
                                (define s (read-bytes-line i 'any))
                                (fprintf o "331 Go ahead\n")
                                (flush-output o)
                                (define s2 (read-bytes-line i 'any))
                                (fprintf o "530 Cannot login\n")
                                )
                              ; Immediate
                              (when #t
                                (fprintf o "500 OOPS: cannot read user list\n"))
                              (close-output-port o)
                              (close-input-port i)
                              (loop))))
                  (for ([i (in-range 10000)])
                    ((with-handlers ([exn:fail? (lambda (exn)
                                                  ;(printf "~s\n" (exn-message exn))
                                                  void)])
                       (define conn (ftp-establish-connection server bad-port user passwd))
                       (thread (lambda ()
                                 (displayln 'established)))
                       (lambda () (error "connection didn't fail; server wasn't bad enough")))))
                  ;; we should still be able to connect to the mock server
                  (with-handlers ([exn:fail? (lambda (exn)
                                               (eprintf "Failed to open one more connection; out of descriptors?\n")
                                               (raise exn))])
                    (tcp-connect server bad-port)))
                (custodian-shutdown-all bad-custodian)))
```

(Just a concept...)

mflatt:
Yes, that's the sort of test I had in mind.

dzoep:
OK, pushed tests update. Of course, any feedback is appreciated - as usual!

mflatt:
Thanks!

PR #3955, opened 2021-08-09
Check argument contract in delta introducer

michaelballantyne:
Turns this:

```
((make-syntax-delta-introducer #'id1 (datum->syntax #f 'id1))
  #f 'flip)
```

from a segfault into

```
syntax-introducer: contract violation
  expected: syntax?
  given: #f
```


shhyou:
Just curious -- should it be `'syntax-delta-introducer` instead of `'syntax-introducer`?

michaelballantyne:
Sure, that would make at least as much sense.

sorawee:
There's a `(raise-argument-error 'syntax-introducer "(or/c 'add 'remove 'flip)" mode)` below (need to click "expand more code" in the GitHub UI to see it), which I think is why @michaelballantyne chose that name.

I think the correct fix should not be `syntax-delta-introducer`, but rather `make-syntax-delta-introducer`. The variable `who` is already bound to `'make-syntax-delta-introducer`. But for the sake of consistency, the `raise-argument-error` line should also be changed to use `who`.

shhyou:
I see the other `'syntax-introducer` now. Indeed it is correct as well. Using `'make-syntax-delta-introducer` isn't seems strange though, because the error is raised by the syntax introducer that `make-*` produces, not by `make-*` themselves.

mflatt:
I'll merge as-is for now. I agree that both `syntax-introducer` and `syntax-delta-introducer` make sense, and I don't have a strong opinion among those two, while `make-syntax-delta-introducer` doesn't seem like the right choice.

PR #3956, opened 2021-08-10
Clarify that `prop:syntax-class` is in phase 1

shhyou:
Clarify that `prop:syntax-class` is in phase 1. Also replaces occurrences of `make-variable-like-transformer` and `set!-transformer-procedure` by a inline `syntax-parse`.

### DIFF
--- a/pkgs/racket-doc/syntax/scribblings/parse/stxclasses.scrbl
+++ b/pkgs/racket-doc/syntax/scribblings/parse/stxclasses.scrbl
@@ -63,6 +63,8 @@ bound as variables in the body. Syntax classes support optional
 arguments and keyword arguments using the same syntax as
 @racket[lambda]. The body of the syntax-class definition contains a
 non-empty sequence of @racket[pattern] variants.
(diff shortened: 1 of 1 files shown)


rmculpepper:
The first change looks good, thanks.

I don't think the second change is necessary, and the resulting behavior is wrong. If `f` is bound to an `expr-and-stxclass` and `(f 1 2 3)` is expanded in a context where `#%app` has been replaced, then the new code causes the application to use Racket's `#%app` syntax, but `make-variable-like-transformer` (correctly) uses the `#%app` syntax of the original application. I thought this subtlety was documented, but it appears not. Maybe we should add an example to `make-variable-like-transformer`.

shhyou:
Thanks a lot! Using the correct `#%app` is indeed a subtle but important point. I removed the edits to the example and instead added a `require` form.

I also added a pointer to `expr/c` at the beginning of the documentation (at line 66). How does it look?

While ensuring the correctness of the example is important, I do think there's value to having a simple example for the purpose of illustration. I'll make a separate pull request for that.

Issue #3937, opened 2021-07-24
[Feature Request] `stat` and `lstat` APIs


sschwarzer:
**Is your feature request related to a problem? Please describe.**

I want to scan a directory tree to detect where two or more paths refer to the same inode, i. e. building a mapping inode -> list-of-paths. The `stat` system call (which seemingly doesn't have an API in current Racket) returns a struct that contains not only the inode number of a file or directory, but also the number of paths for the inode. So the scan would only need to store paths where the number of paths for an inode is greater than 1.

Being able to use stat would also be faster if you want different directory/file properties at once, so you don't have to make multiple indirect stat calls by calling, say, `file-size` and `file-or-directory-modify-seconds` after another.

**Describe the solution you'd like**

I'd like to see support for the [stat and lstat](https://man7.org/linux/man-pages/man2/stat.2.html) calls as a Racket API. The procedure, maybe called `stat`, would take a path and an optional keyword argument `#:follow-links?` and return a Racket `struct` similar to that of a the (l)stat(2) calls. I'm not sure about the name of the returned `struct`; `stat-result` may be too generic. Maybe `directory/file-info`?

Cross-platform considerations:

- Different platforms may define different fields for the stat (or equivalent call) result. I'm not sure if the result `struct` should have a union of all the fields that are possible on different platforms or if there should be different `stat` result `struct`s for different platforms (with the reasoning that a Racket process should only need one of the `struct` types).

- I'm not sure I understand the implications for distributed places though where you might want to transfer stat information between different machines on different platforms.

- On platforms where no `stat` or `stat`-alike call is supported, the `stat` call could raise an exception or return a partly populated `stat` result `struct` with the information that can be gathered by other API calls.

**Describe alternatives you've considered**

- There's a procedure [file-or-directory-identity](https://docs.racket-lang.org/reference/Filesystem.html#%28def._%28%28quote._~23~25kernel%29._file-or-directory-identity%29%29) which seems to contain inode information. But that still doesn't contain the number of file system entries for this directory/file, so while scanning I'd need to collect path information for _all_ directories/files, not only those where the number of paths for an inode is greater than 1. Besides that, the documentation for `file-or-directory-identity` is so vague that I'm not completely sure if I could rely on it for my purpose.

  Compared to `file-or-directory-identity` I would also welcome the device information in the `stat` result to detect if the file system tree crosses file system boundaries, which I'd like to avoid in my tool.

- I looked for a Racket package that implements the `stat` calls but didn't find any.

[...]

PR #3961, opened 2021-08-17
WIP: Add stat/lstat functionality

sschwarzer:
This is a **work in progress** PR to include stat/lstat-like functionality in Racket (see https://github.com/racket/racket/issues/3937).

@mflatt : I started with the documentation, but so far the comments take up more space than the actual documentation. ;-)

I'd like to think a bit more about the things I wrote in the notes, but if you want, feel free to give feedback. In particular, I'd like your thoughts on
- using strings or symbols for the hash keys. I guess symbols would be more idiomatic?
- using `follow-links?` (as for `find-files`) or `as-link?` (as for `file-or-directory-identity`)
- order of hash keys and their descriptions
- representation of timestamps with sub-second precision


mflatt:
Definitely symbols as keys.

I'd go with the `as-link?` direction instead of `follow-links?`, since providing `#t` feels more like adding `l` to the front of `stat`.

I don't have an opinion on the order of the keys and descriptions in the documentation, if that's what you mean.

For timestamps where you have sub-second resolution,one possibility is to add another key (such as a `...-nanoseconds` key).

sschwarzer:
To keep a record of a discussion with Matthew outside Github:

The list returned by `(file-or-directory-permissions path)` contains the effective permissions for the user that makes the call. The function uses the C `access` function for this purpose. On the other hand, `file-or-directory-stat` is supposed to return static status information on a file or directory, which would be strictly based on the permission bits. So, if the result of `file-or-directory-stat` contains a list with permissions, it would be something like `'(user-read-bit group-read-bit)`. Since users of the function would need `member` or `index-of` to find if a "bit" is set in the list, they might as well use `bitwise-and` to check the bit. This would save time to create the list and to search it, as applying a bit mask should be much faster.

Therefore, (at least for now) the result of `file-or-directory-stat` will only contain the literal permission bits as returned by `(file-or-directory-permissions path 'bits)`. That said, we still should have bit mask constants to interpret the additional bits that aren't read, write or execute bits.

mflatt:
I left one small comment, but this looks good, so far!

There aren't OS variants other than Posix(-like) and Windows to cover.

sschwarzer:
Hi @mflatt , as far as I can tell I'm done with my part and you can take over, especially for the Windows-specific code in `rktio_fs.c`.

Many thanks for answering my many questions here and on Slack! Also thanks for adding the `derived` target. This was very useful. :-)

If you find anything I should fix, or if you have questions, please let me know. Regarding the questions, some of them might be answered in the - hopefully helpful - commit messages.

racket/src/rktio/rktio.h
```diff
@@ -76,11 +76,12 @@ Thread and signal conventions:
 
  - On systems where signal handling is thread-specific, as on Linux,
    then `rktio_init` should be called before any additional threads,
-   so that a suitable inheritable siganl disposition can be
+   so that a suitable inheritable signal disposition can be
    configured.
 
 */
      
+#include <time.h>
```
mflatt, line 84: I don't think this needs to be here. A client who uses `"rktio.h"` doesn't need to know about `<time.h>`, right?

racket/src/rktio/rktio_fs.c
```diff
@@ -701,6 +702,94 @@ rktio_ok_t rktio_set_current_directory(rktio_t *rktio, const char *path)
...
+    return NULL;
+  }
+
+  do {
+    /* No stat/lstat distinction under Windows */
+    stat_result = _wstat64(wp, &stat_buf);
+  } while ((stat_result == -1) && (errno == EINTR));
+
+  if (stat_result) {
+    /* TODO: Error handling / integration with Racket side? */
```
mflatt, line 762: Should have `get_posix_error()` here for the TODO?

racket/src/rktio/rktio_fs.c
```diff
@@ -701,6 +702,94 @@ rktio_ok_t rktio_set_current_directory(rktio_t *rktio, const char *path)
...
+    return NULL;
+  }
+
+  rktio_stat_buf = (struct rktio_stat_t *) malloc(sizeof(struct rktio_stat_t));
```
mflatt, line 766: Since this block of code is (almost?) the same as the Unix one, would it make sense to share more?

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
```diff
@@ -512,6 +512,70 @@ In all modes, the @exnraise[exn:fail:filesystem] on error (e.g., if no
...
 
 
+@defproc[(file-or-directory-stat [path path-string?]
+                                 [as-link? boolean? #f])
+         (hash/c symbol? any/c)]{
+
+@index['("inode")]{Returns} a hash with the following keys and values:
+
+@itemlist[
+ @item{@racketvalfont{'device-id} : device id}
```
mflatt, line 522: For each of these, I think the documentation should describe the representation, such as "an exact nonnegative integer".

mflatt: Instead of `#racketvalfont{'device-id}`, `@racket['device-id]` is better. In this case, `indexed-racket` is probably best. In general, prefer `racket` to `racketvalfont` until you need to control the textual representation, such as for showing octal.

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
```diff
@@ -512,6 +512,70 @@ In all modes, the @exnraise[exn:fail:filesystem] on error (e.g., if no
...
+If @racket[as-link?] is a true value, then if @racket[path] refers to a
+symbolic link, the stat information of the link is returned instead of the stat
+information of the referenced filesystem item.
+
+The mode bits are the bits for permissions and other data, as returned from the
+Posix @tt{stat}/@tt{lstat} functions or the Windows @tt{_wstat64} function,
+respectively. To select portions of the bit pattern, use the constants
+@racket[user-read-bit] etc.
+
+Depending on the operating system and filesystem, the "nanoseconds" timestamps
```
mflatt, line 564: nit: should be
```
``nanoseconds''
```
instead of
```
"nanoseconds"
```

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
```diff
@@ -512,6 +512,70 @@ In all modes, the @exnraise[exn:fail:filesystem] on error (e.g., if no
...
+may have less than nanoseconds precision. For example, in one environment a
+timestamp may be @racketvalfont{1234567891234567891} (nanoseconds precision)
+and in another environment @racketvalfont{1234567891000000000} (seconds
+precision).
+
+Values that aren't available for a platform/filesystem combination may be set
+to @racketvalfont{0}.
```
mflatt, line 571: Shouldn't they just be absent, instead? Or is the problem that `stat` doesn't say which things are available?

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
```diff
@@ -512,6 +512,70 @@ In all modes, the @exnraise[exn:fail:filesystem] on error (e.g., if no
...
+
+If @racket[as-link?] is @racketvalfont{#f} and @racket[path] isn't accessible,
+the @exnraise[exn:fail:filesystem]. This exception is also raised if
+@racket[as-link?] is a true value and @racket[path] can't be resolved, i. e. is
```
mflatt, line 575: nit: `i.e.,` instead of `i. e.`

mflatt (changes requested):
Thanks for this work! I'm happy to make changes along the lines of this review, but there are a few questions here where I'm not sure about something or would just like to make sure I understand.

racket/src/rktio/rktio.h
sschwarzer: It seems I added this in [286e2a3b] when I added the timestamp fields to the C and Racket structs. When I check the manpage `man 2 stat`, it doesn't list `time.h` as requirement. The header is only used in some example code of the manpage.

Since everything still works after removing the `time.h` include, I removed it in [286e2a3b].

racket/src/rktio/rktio_fs.c
sschwarzer: This is the error handling for the Windows code. I don't know what error handling should look like for Windows, so I'd like to leave this TODO to you. If `get_posix_error()` here works also for Windows, it's confusingly named. ;-)

racket/src/rktio/rktio_fs.c
sschwarzer: I wrote something about this in the commit message for [5466b9f9]. You may have missed it because it was further down:
```
    The sections for Posix and Windows that copy values from `stat_buf` to
    `rktio_stat_buf` are similar, but not the same:
    
    - Some comments refer only to Posix, some only to Windows.
    - The struct fields for timestamps in `stat_buf` are different for
      Posix and Windows.
    
    Though it would be possible to put common code lines in a section
    outside the `ifdef`s, it could well be that this refactoring makes the
    code _more_ difficult to follow. Also keep in mind that we just copy
    data from stat structs that are unlikely to change, due to backward
    compatibility considerations. So we might not benefit from being able to
    change common code in one place. Also it might be that the "common" code
    needs to be changed in different ways for Posix and Windows.
    
    Long story short: Probably it's better to keep the Posix and Windows
    implementations separate, but I leave it to you.
```

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
sschwarzer: Yes, _currently_ all hash entries are non-negative integers. At some point there was a plan to include a file type field, whose values would be symbols, and I don't know if we'll add other non-integer fields in the future. Therefore I'm not sure what's better: Be more precise regarding the current implementation or make a potentially wrong "promise" which may cause problems for client code if it relies on the non-negative-integer assumption for the future.

sschwarzer: > For each of these, I think the documentation should describe the representation, such as "an exact nonnegative integer".

Can you give an example where you would put this information? Anyway, I'm not sure it's needed because someone using the stat function has to know what the fields mean and what their possible values are. On the other hand, the Windows values 0 for uid/gid are a bit special and possibly worth documenting.

racket/src/rktio/rktio.h
sschwarzer: I haven't pushed the review changes yet, so the commit id links will be dead. I'll remove this comment after pushing my changes.

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
sschwarzer: Great, I didn't know that/how Scribble supported this. :-) Changed in sschwarzer@bf7f3053 .

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
sschwarzer: Right. There's something on this in the commit message of sschwarzer@eff1b4ce .

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
sschwarzer: Changed in sschwarzer@c0b1d2fb .

pkgs/racket-doc/scribblings/reference/filesystem.scrbl
sschwarzer: > For each of these, I think the documentation should describe the representation, such as "an exact nonnegative integer".

Somewhat related: The contract of the function says about the hash: `(hash/c symbol? any/c)`. _Currently_ all hash values are non-negative integers, but I'm using `any/c` to avoid promising this for future enhancements of the function. Actually, at one point we discussed including a `'type` key whose value would be a symbol like `'directory`. There's still (commented-out) documentation on this.

mflatt:
Some notes after trying a build on Mac OS:

The `st_atim` fields is called `st_atimespec`, etc. Apparently, Mac OS is the only non-standard one these days. Of course, I'm happy to make adjustments there.

This test doesn't pass:
```
  (test #t >= (stat-ref 'access-time-nanoseconds)
              (stat-ref 'modify-time-nanoseconds))
```
It fails because the file is most recently written, and not read. It seems like the failure shouldn't be Mac-specific, but maybe it requires a filesystem that records time in a certain way (including with enough resolution). I'm inclined to repair the test by adding
```
  (void (call-with-input-file temp-file-path read-byte))
```
just after the `display-to-file` that writes the file.

I'd also like to add tests like
```
  (test #t >= (stat-ref 'access-time-seconds)
              (stat-ref 'modify-time-seconds))
```
and tests to make sure that the nanosecond values round down to the seconds values. And a test to make sure there the set of keys matches the documented set (i.e., no extra ones) and that all values (for now) are nonnegative exact integers.

I didn't notice this before, but since the table uses symbol keys, it should be constructed with `hasheq` instead of `hash`.

A bigger thing I didn't notice before is that `file-or-directory-stat` is only implemented for CS. (Maybe I once knew to expect that and had forgotten.) I'll fill in the BC implementation.

mflatt:
Here's a tentative set of updates: https://github.com/mflatt/racket/commit/b67037f48292de4d497ad1ed8237ad390aedd805

mflatt:
Initial Windows experience: `st_rdev` is defined to be the same as `st_dev`, so usually not 0. Probably it's best to set `device_id_for_special_file` to 0 on Windows. Also, there are tests that `block_size` and `block_count` are positive, so I'll adjust those tests to happen only on non-Windows.

mflatt:
Merged!

sschwarzer:
I see that in the documentation you changed "user id" to "user ID" and "group id" to "group ID." I usually see the terms written as "user id" and "group id", but a web search shows that "user ID" is a common term. That said, to me it seems "user ID" usually refers to a user name like `mflatt`, whereas the user id in the stat result is a _numeric_ id. Similarly for the group id.

Changing "user ID" back to "user id" probably is still ambiguous because the distinction between "id" and "ID" is too subtle. Therefore, I suggest we make the wording in the documentation explicit and change it to "numeric user ID" and "numeric group ID."

What do you think? If you agree, should I create a PR or will you do the change yourself? :-)

sschwarzer:
Github says:
>  Closed with unmerged commits

Can I see somewhere what these unmerged commits are? Are these the three "Reduce mode information to permission/bits" etc.?

samth:
I believe that just means that the actual commits weren't merged from Git's perspective, instead the commits ended up with other hashes.

mflatt:
Yes, I should have used GitHub a different way to properly record that it was merged.

Using "numeric group ID" sounds good, and I can make that change.

PR #3962, opened 2021-08-18
json: Parse large exponents without getting stuck

m4burns:
Before this patch, all JSON-parsing forms in the `json` module get stuck on numbers with large exponents:

```
(require json racket/sandbox)
(with-limits 10 #f (string->jsexpr "1e99999999999999"))
; with-limit: out of time
```

Memory consumption appears to grow slowly. This is bad when parsing JSON from untrusted sources, e.g. as a web server might do.

The underlying cause is evaluation of `(expt 10 exp)` where `exp` is an unbounded value read from the input. Since the JSON parser coerces all exponent-containing numbers to `inexact?`, the result will have bounded exponent magnitude or will be +/-infinity. This patch returns +/-infinity or +/- 0 eagerly when the exponent is too big or too small.

It's worth noting that the JSON-parsing forms already break their contract by returning +/-infinity on large enough inputs:

```
(string->jsexpr "1e400")
+inf.0
```

This patch preserves that behaviour.

### DIFF
--- a/pkgs/racket-test/tests/json/json.rkt
+++ b/pkgs/racket-test/tests/json/json.rkt
@@ -115,6 +115,15 @@
         (string->jsexpr @T{-10.34e-3}) => -1.034e-2
         (string->jsexpr @T{-10.34e-03}) => -1.034e-2
         (string->jsexpr @T{-10.34e+31}) => -1.034e32
+        (string->jsexpr @T{ 1e9999999999999999    }) => +inf.0 ; breaks contract
+        (string->jsexpr @T{ 1.0e9999999999999999  }) => +inf.0
+        (string->jsexpr @T{-1e9999999999999999    }) => -inf.0
+        (string->jsexpr @T{-1.0e9999999999999999  }) => -inf.0
+        (string->jsexpr @T{ 1e-9999999999999999   }) =>  0.0
+        (string->jsexpr @T{-1e-9999999999999999   }) => -0.0
+        (string->jsexpr @T{ 0e9999999999999999    }) => 0.0
+        (string->jsexpr @T{ 0e-9999999999999999   }) => 0.0
+        (string->jsexpr @T{ 0.001e310 }) => 1.0e307
         (string->jsexpr @T{ true  }) => #t
         (string->jsexpr @T{ false }) => #f
         (string->jsexpr @T{ null  }) => 'null
--- a/pkgs/racket-test/tests/json/json.rkt
+++ b/pkgs/racket-test/tests/json/json.rkt
@@ -124,6 +124,7 @@
         (string->jsexpr @T{ 0e9999999999999999    }) => 0.0
         (string->jsexpr @T{ 0e-9999999999999999   }) => 0.0
         (string->jsexpr @T{ 0.001e310 }) => 1.0e307
+        (string->jsexpr @T{ 100000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000e-402 }) => 1e-292
         (string->jsexpr @T{ true  }) => #t
         (string->jsexpr @T{ false }) => #f
         (string->jsexpr @T{ null  }) => 'null
--- a/racket/collects/json/main.rkt
+++ b/racket/collects/json/main.rkt
@@ -361,12 +361,16 @@
     ;; evaluate n * 10^exp to inexact? without passing large arguments to expt
     ;; assumes n is an integer
     (define (safe-exponential->inexact n exp)
+      (define result-exp
+        (if (= n 0)
+            exp
+            (+ (log (abs n) 10) exp)))
       (cond
-        [(< exp -309)
+        [(< result-exp -400)
          (cond
            [(>= n 0) 0.0]
            [else -0.0])]
-        [(> exp 309)
+        [(> result-exp 400)
          (cond
            [(= n 0) 0.0]
            [(> n 0) +inf.0]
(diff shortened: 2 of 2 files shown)


rfindler:
This is great, thank you for finding the problem! 

The only change I'd make if I'd written this was to either formulate a bunch of json test cases to get branch coverage in the new function or test the new helper directly.

m4burns:
@rfindler I did add some test cases to `pkgs/racket-test/tests/json/json.rkt` that I believe cover all branches.

I also tested with vs. without the patch on random inputs using this script:

```
#lang racket

(require "main.rkt"
         (prefix-in new: "main-new.rkt"))

(define (gen)
  (define n? (= (random 2) 0))
  (define d (random 1 100))
  (define f? (= (random 2) 0))
  (define of? (= (random 2) 0))
  (define flz (random 1 100))
  (define fd (random 1 100))
  (define e (random -600 600))
  (with-output-to-string
    (thunk
      (when n?
        (display "-"))
      (if of?
          (display "0")
          (for ([i d])
            (display (random 10))))
      (when (or f? of?)
        (display ".")
        (for ([i flz])
          (display "0"))
        (for ([i fd])
          (display (random 10))))
      (display "e")
      (display e))))

(let loop ([n 0])
  (when (= (remainder n 10000) 0)
    (printf "~a~n" n))
  (define s (gen))
  (unless (= (string->jsexpr s) (new:string->jsexpr s))
    (printf "mismatch: ~a~n" s)
    (exit 1))
  (loop (add1 n)))
```

It hasn't found anything in a few million attempts. I'm not sure where to put something like this; it's awkward to commit it since it requires both the old and new version of `json/main.rkt`.

rfindler:
Oh, beautiful! I somehow looked only at the second commit when I was looking for the test cases (my bad).

I think this is ready to go. Any objections to me merging it?

rfindler:
Oh, could you use a slightly different property in your random tests, namely: if json parses it successfully as a number, then racket's read must produce the same number? Or, actually, looking at the code, maybe it always generates numbers that are both racket and json numbers?

m4burns:
@rfindler OK by me! Thanks for the quick responses :)

rfindler:
Okay and if you want to include a variation on that random testing that uses `read`, I think that'd be great but can be a separate pull request.

rfindler:
Oh, whoops: this should probably be one commit. I'll adjust it and push that and close this.

m4burns:
Sounds good. I'm not sure `read` will match `string->jsexpr` in all cases but I'll look into it.

rfindler:
squashed and pushed.

rfindler:
I recall a comment in the json source at some point that said something like "we can almost just use read to read numbers here and simplify this code a lot" or similar which is what led to my (perhaps overly) optimistic comment. :)

mfelleisen:
Just caught up with this mail thread. 

There are many such inputs that can make a JSON reader blow up. I have a whole library (draft forever) that allows to enforce restrictions at all kinds of stages of reading: [SwDev/Testing/communication](https://github.com/mfelleisen/SwDev/blob/master/Testing/communication.rktl). I am wondering how much of this protection a plain reading  library should provide because this doesn't come free but for real applications it's essential. 

And of course also see [JSON is hard](http://seriot.ch/parsing_json.php). 

@m4burns Thanks for working this out. Numbers really are a bad spot.

rfindler:
That SwDev link seems to be private, @mfelleisen . 

re "I am wondering how much of this protection a plain reading library should provide", in the specific case of this bug fix, I think that if producing the read result requires only, some small fixed amount of memory, then the JSON parser should not allocate unbounded amounts attempting to produce it. So I think this library was the perfect place for this fix.

I would be delighted if someone were to go over the test cases in the "JSON is hard" link and bring as many of them as makes sense into our test suites.

samdphillips:
> I would be delighted if someone were to go over the test cases in the "JSON is hard" link and bring as many of them as makes sense into our test suites.

I wrote [this test suite](https://github.com/samdphillips/racket-stream-json/blob/main/stream-json-test/json/stream/test) with the json files from "JSON is hard" a while back to test some (still experimental) JSON parsing tools.  I tooled it to also exercise the Racket json library and the only failures that I saw were due to `read-json` allowing trailing data after reading a json term, which is a perfectly reasonable thing to do IMO.  Changing the behavior of the tester could eliminate those failures though.

Would a PR of a the test runner (cleaned up) and test files be acceptable?

m4burns:
@samdphillips Nice! I threw together a script that did the same thing some time ago (that's actually how I discovered this issue) but yours looks much better. I recall it was simple to adjust the script to match what the tests expect for trailing data.

Such a PR sounds like a good idea to me. @rfindler suggested to add (some of) the tests to the existing suite in `pkgs/racket-test/tests/json/json.rkt`, which I think might be a smaller overall change.

mfelleisen:
@rfindler [https://github.com/mfelleisen/SwDev](https://github.com/mfelleisen/SwDev) is public 

@samdphillips When I read "json is hard" our reader also failed on some boundary Unicode things.  Some students exploited this gap. ~~ Leaving characters on the stream when `read-json` is finished is _totally okay_. 

Thanks.

rfindler:
I agree that a PR would be great! I also think it is essential that `read-json`'s behavior wrt to consuming the entire stream not change. If someone wants to make sure the stream contains exactly one json object, they should check for an eof after calling read-json.

m4burns:
Follow up on testing against `read`: the only mismatch my random tests reveal is that `read` accepts numbers with unnecessary leading zeroes, which `read-json` will just parse as `0` since unnecessary leading zeroes are not permitted in JSON. Otherwise, `read-json` and `read` produced the same result.

I'll follow up with another PR later that adds the random test script.

rfindler:
@m4burns great!

PR #3968, opened 2021-08-19
Adjust documentation for contract generators

camoy:
Two documentation changes:

* The `generator` argument for `flat-named-contract` was not correct. It takes in fuel and returns a thunk of exactly one result.
* The `generator` argument for `build-*-contract-property` can be made more precise. Fuel should always be a nonnegative integer.


rfindler:
Thanks!

PR #3991, opened 2021-09-10
Clarify contribution for `main-distribution` packages

jryans:
This extends the contribution guide for `main-distribution` packages to describe two separate workflows

1. Release / snapshot
2. In-place build

each of which needs slightly different commands.

### DIFF
--- a/pkgs/racket-build-guide/contribute.scrbl
+++ b/pkgs/racket-build-guide/contribute.scrbl
@@ -76,15 +76,31 @@ See the @secref["contribute-guidelines"].
 @section[#:tag "pkg-contribute"]{Distribution-Package Contributions}
 
 If you find yourself changing a file that is in a
-@filepath{share/pkgs} subdirectory, then that file is not part of the
-main Racket Git repository. It almost certainly has its own Git
-repository somewhere else, possibly within
+@filepath{share/pkgs} subdirectory (either installed as part of a
+Racket release or as a product of an in-place build), then that file
+is not part of the main Racket Git repository. It almost certainly has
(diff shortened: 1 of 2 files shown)


pkgs/racket-build-guide/contribute.scrbl
```diff
@@ -76,15 +76,30 @@ See the @secref["contribute-guidelines"].
...
+is not part of the main Racket Git repository. It almost certainly has
+its own Git repository somewhere else, possibly within
 @url{https://github.com/racket}, but possibly in another user's space.
 The name of the directory in @filepath{share/pkgs} is almost certainly
 the package name.
 
-To start working on a package @nonterm{pkg-name}, it's usually best to
-go to the root directory of your Racket repository checkout and run
+To start working on a package @nonterm{pkg-name} from a Racket release
+or snapshot, you first need to edit the package's catalog
```
mflatt, line 88: "edit the package's catalog" is not quite right here. This should be something like "adjust the package installation to use the source specified by the main package catalog".

mflatt (changes requested):
Thanks for this improvement! I have one request for a wording change.

pkgs/racket-build-guide/contribute.scrbl
jryans: Ah thanks, that's a better description of what's happening. 😄 I have updated this to match your wording.

jryans:
@mflatt Thanks for taking a look, review comment addressed. 🙂 Someone will need to re-approve the workflows if a fresh CI pass is desired before merging.

PR #4001, opened 2021-09-24
Fix fixnum/wraparound tests

gus-massa:
Fix a typo in a variable. This was causing an error, but it was call inside an error handler that calls an escape continuation claiming success, so the error was ignored.

https://github.com/racket/racket/blob/d5794de89bb7e44e46621c9fd2385ebb20313f41/pkgs/racket-test-core/tests/racket/fixnum.rktl#L102

Also, it's necessary also swap the code used to complete the bits of the result in case of a wraparound.


sorawee:
Can we add a test that would cause the code before the PR to fail?

gus-massa:
Something like this? I'm not sure if I should put the test there, or down after all the definition of the auxiliary functions.

Another alternative is to split the `binary-table` in two tables, and then clone the code and make a version without the error handler for the wraparound functions.

PR #4002, opened 2021-09-25
Fix build guide when returning to normal

jryans:
This fixes a small error in the build guide. Instead of `--catalog`, `--lookup` should be used to return to the default installation of a package.


shhyou:
Thank you!

PR #4003, opened 2021-09-26
define-ffi-definer: support variables, c-id from run-time expr

rmculpepper:
If `define-foreign` is defined with `define-ffi-definer`, then it can define variables (backed by `make-c-parameter`) as follows:

    (define-foreign counter _int #:variable)

And it can calculate the foreign object name at run time as follows:

    (define-foreign close (_fun _socket -> _int)
      #:c-id ,(if win? 'closesocket 'close))


mflatt:
This looks like a good change, but `history` annotations are needed with the documentation changes (unless they're already there and I'm overlooking them).

PR #4004, opened 2021-09-26
pretty-print adjustment

sorawee:
Adjust so that `send` is formatted like `module` rather than
`syntax-case`.

In particular:

```
(parameterize ([pretty-print-columns 10])
  (for ([x '(module syntax-case)])
    (pretty-print
     `(,x obj meth (begin 1234 5678)))))
```

outputs:

```
'(module obj meth
   (begin
     1234
     5678))
'(syntax-case obj meth
   (begin
    1234
    5678))
```

and we want it to be formatted like the first one
rather than the second one.

Also add λ to the list of supported macros


rfindler:
Sorry if I'm being naive here, but why does swapping between `module` and `syntax-case` determine how the `begin` is formatted?

sorawee:
`pp-syntax-case` prints each form in the body as an expression list. This makes sense because you don't want:

```
(syntax-case stx ()
  [begin
   something
   something])
```

to be formatted as:

```
(syntax-case stx ()
  (begin
    something
    something))
```

On the other hand, `pp-module` prints each form in the body as an expression, so `(begin something something)` is indented with the `begin` style as it should.

rfindler:
Oh! Duh. Of course. Thanks.

PR #4006, opened 2021-09-27
file/glob: fixes for Windows

### LINKED ISSUE #4000: glob.rkt fails unit test on Windows

8.2 [cs]

**What program did you run?**

Load the file `racket/racket/collects/file/private/glob.rkt` from the Racket distribution into Dr Racket on Windows. Run it (you can also eval the same file including the `test` submodule).

It should complete without error.

`check-equal?: path-string->glob:unambiguous [645:8:23] actual value C:\a\b\c differs from expected /a\b\c`

Windows only.

bennn:
- disallow escapes because \ is a path separator
- output correct path separators
- change expectations for some tests (/ vs. C:\ etc.)

fixes #4000

### DIFF
--- a/pkgs/racket-doc/file/scribblings/glob.scrbl
+++ b/pkgs/racket-doc/file/scribblings/glob.scrbl
@@ -37,6 +37,8 @@ a set of path strings using the following @deftech[#:key "glob-wildcard"]{wildca
 }
 ]
 
+@margin-note{On Windows, wildcards cannot be escaped because @tt{\} is a path separator.}
+
 By default, wildcards will not match files or directories whose name begins
 with a period (aka "dotfiles").  To override, set the parameter
 @racket[glob-capture-dotfiles?] to a non-@racket[#f] value or supply a similar
--- a/racket/collects/file/private/glob.rkt
+++ b/racket/collects/file/private/glob.rkt
@@ -42,6 +42,7 @@
   (for-syntax racket/base))
 
 (module+ test ;; faking RackUnit
+  (define windows? (eq? 'windows (system-type 'os)))
 
   (define current-test-case (make-parameter #f))
   (define num-tests (box 0))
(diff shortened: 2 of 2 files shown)


bennn:
@Danl2620 would you test this patch on your machine?

mflatt:
I haven't tried running it, yet, but this looks like a useful and thorough update.

Is it possible that disallowing `\` as an escape character on Windows will break some existing uses, or was `glob` so reliably broken on Windows before that there can't really be existing uses?

bennn:
No, I don't think escapes ever worked because`glob` calls `path->complete-path` before matching. That should turn all Windows "escapes" into path separators, right?

PS looks like `?` and `*` can't appear in a Windows filename [[link]](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions). So the only things to escape would be matching `[]` and `{}`.

mflatt:
> No, I don't think escapes ever worked because`glob` calls `path->complete-path` before matching. That should turn all Windows "escapes" into path separators, right?

Ok, that makes sense.

> PS looks like `?` and `*` can't appear in a Windows filename [[link]](https://docs.microsoft.com/en-us/windows/win32/fileio/naming-a-file#naming-conventions). So the only things to escape would be matching `[]` and `{}`.

Well, yes and no. That page lists conventions you should follow to avoid trouble, but not all of them are actual filesystem or API restrictions. I'm pretty sure it was once true that special characters can be included in a path element using the `\\?\` path form to access filenames using those characters (when mounting an ext3 filesystem, for example). then again, I can't make that work on Windows 10 accessing a VirtualBox shared folder; maybe the OS changed, maybe the filesystem driver changed. Meanwhile, I see that Racket's path functions for Windows incorrectly omit `?` and `*` from the list of special characters, and I'll fix that.

In any case, I agree that the danger is low for changing `glob` for Windows.

PR #4007, opened 2021-09-28
Improve user experience with read-char and read-line at the REPL.

tonyg:
`winny` on IRC (and subsequently on the mailing list [1]) remarked
that, at the REPL,

```racket
Welcome to Racket v8.2.0.8 [cs].
> (read-line)
""
>
```

which is surprising, since it didn't appear to wait for a line of
input.

Guile does this differently, with its `read-eval-print-loop`
apparently consuming any whitespace after the `read` expression and
before starting the `eval`.

This patch performs a similar trick. In *interactive* contexts
(namely, by action of the default `current-read-interaction`), if
`read-syntax` answers non-`eof`, a procedure
`discard-line-terminators` peeks for and consumes a *single* CR, LF or
CRLF line terminator.

Non-interactive contexts are not affected.

This is very much a special-case in order to improve user experience:
I feel like, because it's an amendment to the REPL and the top-level
is hopeless [2], it's fair game to introduce exceptional handling like
this.

[1]: https://groups.google.com/g/racket-users/c/qUIFqWkkvFs/m/AERXYmfGBgAJ
[2]: https://gist.github.com/samth/3083053


winny-:
Shout out to @tonyg  for looking into how to address this issue and coming up with a PR!

An added bonus, this appears to work in the Emacs Racket REPL.

This PR appears to fix the common cases:

### yn prompter

(8.0 always results in `#f` unless you do something like `(begin (yn) (yn))` to eat the newline)

```
> (define (yn) (match (read-char) [(or #\y #\Y) #t] [_ #f]))
> (yn)
y
#t
>
```

### Reading a line to greet

(8.0 always prints `Hello !` unless you eat the newline, e.g. `(begin (read-char) (greet))`.)

```
> (define (greet) (printf "Hello ~a!\n" (read-line)))
> (greet)
A Racket User
Hello A Racket User!
>
```

### Reading two lines interactively

(8.0 gives back `'("" "hello")` as it doesn't eat the first newline.)

```
> (list (read-line) (read-line))
hello
world
'("hello" "world")
```

This seems (to me?) less surprising than the 8.0 behavior.


## Some Odd behaviors - maybe ignore?

In this section I compare and contrast with Guile's handling of the trailing newline at top level and just behavior in general.  My understanding is this change was inspired by how Guile addresses the edgecases this PR addresses.  These behaviors might be worth glossing over, as top level is pretty hopeless :slightly_smiling_face: :

### double read-line

```
> (read-line)(read-line)
"(read-line)"
>
```

(For what it's worth this is how Guile behaves as well, maybe not worth making another special case for?)

### double read-char

```
> (read-char)(read-char)
#\(
#<procedure:read-char>
; readline-input:41:21: read-syntax: unexpected `)` [,bt for context]
>
```

(Same as above; Guile behaves the same, perhaps not worth making special case for?)

### double read-char separated by whitespace

As found in this PR:

```
> (read-char) (read-char)
#\space
u
#\u
>
```

And in Racket 8.0:

```
> (read-char) (read-char)
#\space
#\newline
> 
```

Interestingly one gets the following in Guile:

```
scheme@(guile-user)> (use-modules (ice-9 rdelim))
scheme@(guile-user)> (read-char) (read-char)
$1 = #\(
$2 = #<procedure read-char (#:optional _)>
While reading expression:
In procedure read_inner_expression: #<unknown port>:2:24: unexpected ")"
```

I wonder if the new behavior of returning the space, then the next read character is the least surprising?  Just my opinion.

Hope these example cases help.

tonyg:
Thanks for the examples. The Guile behaviour is to eat all (available??) trailing whitespace after an expression, I think, whereas I chose instead to take at most a single LF/CR/CRLF. I think for the purposes of the REPL the LF/CR/CRLF-eating is sufficient, but Guile's behaviour is also totally reasonable and would make a fine alternative, I reckon.

mflatt:
This change seems good to me. I was worried at first about the interaction with non-S-expression modes, but I finally see that this approach of adjusting the default `current-read-interaction` should generally avoid that issue.

It looks like the documentation needs a `history` note.

tonyg:
> It looks like the documentation needs a `history` note.

Ah, right -- something along the lines of "Prior to version X, the behaviour lacked newline-chomping"? I'll amend the patch. Should anything else go in the history note?

samth:
Well, you'd probably use `@history` to generate text like that for you, but basically yes.

tonyg:
OK, @mflatt (and @samth?), I've added a `@history` note and it's ready for review again. Thanks!

mflatt:
Thanks! That's a more detailed history note than typical, but more is certain fine.

Although I think this is almost certainly a good idea, I recommend that we merge after the v8.3 branch, so we try this for a while before shipping it.

mflatt:
This change turns out to interact badly with `xrepl`. For example, try this:
```
,ls
```
Before this PR's change, `xrepl` used the leftover newline as an indication that no more input is available.
Now, since the newline is gone, it waits for more input.

I discovered the interaction because I've been working on `expeditor` as a replacement for `readline`, making `xrepl` work nicely with it. The `expeditor` implemenation is based on Chez Scheme's expression editor. It doesn't suffer from the problem that this PR addresses, because it ends up separating command input from the current input port.

If `expeditor` pans out and gets added to the distribution, in which case that will be what `racket` users see by default, then maybe the change from this PR should be reverted to reduce its unexpected effects. (I don't yet have a strong opinion, though.)

PR #4013, opened 2021-10-07
Check for initial eof in `file->X` functions.

### LINKED ISSUE #4012: file->string seems to fail in some cases
8.2 [cs]

**What program did you run?**
`(file->string some-file)`  running in a loop on a network share where another process is writing some-file continuously (and possibly erasing it).

Don't know because I don't know the content of the file at this very moment, but I'd expect the content of the file at some point in the past (maybe empty?) considering the documentation :)
We're clearly into undefined behaviour territory though, so I need to do validation afterwards, so not crashing would be enough.

```
string-append: contract violation
  expected: string?
  given: #<eof>
```

Windows

samth:
Closes #4012.

### DIFF
--- a/pkgs/racket-test/tests/racket/stress/file-string-eof.rkt
+++ b/pkgs/racket-test/tests/racket/stress/file-string-eof.rkt
@@ -1,0 +1,27 @@
+#lang racket
+
+(define (spin-a-while)
+  (let loop ([n (random 1000000)])
+    (if (zero? n)
+        'done
+        (loop (sub1 n)))))
+
+(for ([i 100])
+  (define f (make-temporary-file))
+  (with-output-to-file f #:exists 'truncate
+    (lambda () (write 1)))
+  (define t1 (thread (lambda ()
+                       (spin-a-while)
+                       (for ([i 100])
+                         (unless (string? (file->string f))
+                           (error "oops"))))))
+  (define t2 (thread (lambda ()
+                       (spin-a-while)
+                       (with-output-to-file f #:exists 'truncate
+                         (lambda ()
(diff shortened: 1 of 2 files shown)


samth:
I wrote the following test, but it never failed for me:

```
#lang racket

(for ([i 100])
  (thread
   (lambda ()
     (define f (make-temporary-file))
     (with-output-to-file f #:exists 'truncate
       (lambda () (write 1)))
     (define t1 (thread (lambda () (file->string f))))
     (define t2 (thread (lambda () (with-output-to-file f #:exists 'truncate
                                     (void)))))
     (thread-wait t1)
     (thread-wait t2))))
```

mflatt:
I think you have an extra `thread` above that causes the program to exit before it really does anything.

With the following variant, I'm able to get "oops".  I haven't been able to provoke a `string-append` error, through, even on Windows.

```
#lang racket

(define (spin-a-while)
  (let loop ([n (random 1000000)])
    (if (zero? n)
        'done
        (loop (sub1 n)))))

(for ([i 100])
  (define f (make-temporary-file))
  (with-output-to-file f #:exists 'truncate
    (lambda () (write 1)))
  (define t1 (thread (lambda ()
                       (spin-a-while)
                       (for ([i 100])
                         (unless (string? (file->string f))
                           (error "oops"))))))
  (define t2 (thread (lambda ()
                       (spin-a-while)
                       (with-output-to-file f #:exists 'truncate
                         (lambda ()
                           (spin-a-while)
                           (display "now there's data")
                           (flush-output))))))
  (thread-wait t1)
  (thread-wait t2)
  (delete-file f))
```

PR #4021, opened 2021-10-18
Use correct `format` in error case.

### LINKED ISSUE #4018: Incorrect error message for `file-or-directory-identity` on non-existant file
8.2 [cs]

**What program did you run?**
`racket - e (file-or-directory-identity "garbage")`, assuming `garbage` does not exist.

An error showing that a file does not exist error.

```
format: should only be used as a fallback
  format string: "error obtaining identity for path\n  path: ~a"
  argument: #<path:/home/sage/garbage>
  context...:
   body of top-level
```

samth:
Fixes #4018.


samth:
It seems that there were never any tests for `file-or-directory-identity`, so I added a couple. I want to make sure they pass on all platforms before I merge this, though.

PR #4025, opened 2021-10-20
Add the optional parameter try-order? to hash->list.

scolobb:
Make `hash->list` forward the optional argument `try-order?` to `hash->map`, which is anyway the workhorse behind `hash->list`.

See also this discussion on Racket Users : https://groups.google.com/g/racket-users/c/mwiS9sN6ElA


scolobb:
> If you're up for it, it would be great to get `hash-keys` and `hash-values`, too!

Sounds good to me!  I added a new commit adding `try-order?` to `hash-values`, in the same way as for `hash->list`.

I have some doubts concerning `hash-keys` though: it is currently [not implemented using `hash-map`](https://github.com/racket/racket/blob/4f43400bbfe00adb9f507338a5d5c2a53c13bbfd/racket/collects/racket/private/hash.rkt#L4-L12), but the implementation doesn't seem too far away from [`hash-map`](https://github.com/racket/racket/blob/4f43400bbfe00adb9f507338a5d5c2a53c13bbfd/racket/src/cs/rumble/hash.ss#L444-L466) at a first glance.

Is there a reason not implement `hash-keys` in terms of `hash-map`?  If there isn't any, I'll definitely do the change.

scolobb:
> It may be out of scope for this PR, but I've sometimes wondered if `orderable?` and `orderable<?` from https://github.com/racket/racket/blob/4f43400bbfe00adb9f507338a5d5c2a53c13bbfd/racket/src/cs/rumble/hash.ss#L500-551 ought to be exposed as new primitives. On the one hand, this functionality is sort of arbitrary, but it's also super useful for common cases like a hash table of symbols or integers, and the general principle is to define these sort of types via a predicate, even if partly just for documentation purposes.

I think it makes sense to expose these if we add `try-order?` to many functions: these new primitives would help to know what kind of behavior to expect from `try-order?` on a particular collection.

I am not sure how to proceed though.  I think starting a different PR would be a good idea, as merging that change could be decorrellated from merging this one, but I don't know the best practices for the Racket repository yet.

And thank you for your comments @LiberalArtist by the way!

samth:
Is this ready to merge?

scolobb:
> Is this ready to merge?

I believe the two commits are good for merging, but I'm still not sure about `hash-keys`.  Since the implementation of `hash-keys` seems quite similar to `hash-map`, I would naturally replace the current implementation of `hash-keys` by a call to `hash-map` (like in `hash-values`).  If there are no reasons to not do this replacement, I would add a third commit to make this PR a complete bundle.

What do you think?

LiberalArtist:
My guess is that `hash-keys` is trying to avoid accessing the values as a performance optimization, though I don't know if that's correct or, if so, if it's still applicable. But I think a conservative change for now could do something like:
```racket
(define-syntax-rule (define/optional-try-order (name table try-order?)
                      body0 body ...)
  (define name
    (let* ([name (λ (table try-order?)
                   body0 body ...)]
           [name
            (case-lambda
              [(table)
               (unless (hash? table)
                 (raise-argument-error 'name "hash?" 0 table))
               (name table #f)]
              [(table try-order?)
               (unless (hash? table)
                 (raise-argument-error 'name "hash?" 0 table try-order?))
               (name table try-order?)])])
      name)))

(define/optional-try-order (hash-keys h try-order?)
  (if try-order?
      (hash-map h (λ (k v) k) #t)
      (let loop ([pos (hash-iterate-first h)])
       (if pos
           (let ([k (hash-iterate-key h pos unsafe-undefined)]
                 [r (loop (hash-iterate-next h pos))])
             (if (eq? k unsafe-undefined)
                 r
               (cons k r)))
         null))))
```

scolobb:
Thank you for answer @LiberalArtist ! Busy times for me right now, but I'll try to get around to this PR this week.

scolobb:
Phew, finally got around to implementing your suggestion @LiberalArtist , thanks again!

Besides being busy lately, I also hesitated quite a lot about the following two things:

1. Only avoiding the values when `try-order?` is `#f` or not given: I tried figuring out how I could directly do the kind of sorting `hash-map` does in the case in which `try-order?` is `#t`, and understood your question about `orderable?` and `orderable<?` :smile: I think falling back to `hash-map` is the simplest way to go for now.

2. Defining `define/optional-try-order`: Defining a macro seemed an overkill to me, until I started trying to write out your solution without syntax transformers :smile: I realized then that your syntax rule reduces boilerplate and duplicate code, and is actually very elegant.

So, at the end, I just copied and pasted your code :innocent:

scolobb:
Hm, this error message gets me:

```
/__w/racket/racket/racket/collects/racket/private/hash.rkt:21:2: define/optional-try-order: use does not match pattern: (define/optional-try-order (name table try-order?) body0 body ...)
  in: (define/optional-try-order (hash-keys h try-order?) (if try-order? (hash-map h (λ (k v) k) #t) (let loop ((pos (hash-iterate-first h))) (if pos (let ((k (hash-iterate-key h pos unsafe-undefined)) (r (loop (hash-iterate-next h pos)))) (if (eq? k unsafe-u...
  location...:
   /__w/racket/racket/racket/collects/racket/private/hash.rkt:21:2
  context...:
   /__w/racket/racket/racket/collects/syntax/wrap-modbeg.rkt:46:4
   /__w/racket/racket/racket/collects/setup/main.rkt:240:24
   loop
   body of "/__w/racket/racket/racket/collects/setup/main.rkt"
```

When I write the same code in a separate file with `#lang racket` or even `#lang racket/base`, the definition of `hash-keys` works.  Am I missing some requires?

LiberalArtist:
It looks like you need to require `(for-syntax "ellipses.rkt")` (i.e. `racket/private/ellipses`).

`define/optional-try-order` should be useful for `hash-values` and `hash->list`, too.

scolobb:
Thanks a lot @LiberalArtist! It looks like requiring `ellipses` `for-syntax` fixed it.  I'll refactor the whole PR to use `define/optional-try-order` now :crossed_fingers:

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -570,26 +570,41 @@ with the following order (earlier bullets before later):
...
          #:changed "7.1.0.7" @elem{Added guarantees for @racket[try-order?].}]}
 
-@defproc[(hash-keys [hash hash?])
+@defproc[(hash-keys [hash hash?] [try-order? any/c #f])
          (listof any/c)]{
 Returns a list of the keys of @racket[hash] in an unspecified order.
 
-See @racket[hash-map] for information about modifying @racket[hash]
-during @racket[hash-keys]. @see-also-concurrency-caveat[]}
+If @racket[try-order?] is true, then the order of keys and values passed to
```
mflatt, line 577: It looks like this sentence came from the `hash-map` documentation, but there's no `proc` here. I think just dropping the sentence, leaving the next one (that directs a reader toward `hash-map`) is probably fine.

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -570,26 +570,41 @@ with the following order (earlier bullets before later):
...
+@racket[proc] is normalized under certain circumstances.  See @racket[hash-map]
+for further explanations on @racket[try-order?] and on information about
+modifying @racket[hash] during
+@racket[hash-keys]. @see-also-concurrency-caveat[]
 
-@defproc[(hash-values [hash hash?])
+@history[#:changed "8.3.0.2" @elem{Added the @racket[_try-order?] argument.}]}
```
mflatt, line 583: If you're changing other things, updating the history note to say "8.3.0.11" would be good.

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -570,26 +570,41 @@ with the following order (earlier bullets before later):
...
+
+@defproc[(hash-values [hash hash?] [try-order? any/c #f])
          (listof any/c)]{
 Returns a list of the values of @racket[hash] in an unspecified order.
 
-See @racket[hash-map] for information about modifying @racket[hash]
-during @racket[hash-values]. @see-also-concurrency-caveat[]}
+If @racket[try-order?] is true, then the order of keys and values passed to
```
mflatt, line 589: Same here as for `hash-keys`, but in this case, it might make sense to clarify that the order is based on the associated keys, not the returned values.

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -570,26 +570,41 @@ with the following order (earlier bullets before later):
...
+@history[#:changed "8.3.0.2" @elem{Added the @racket[_try-order?] argument.}]}
 
-@defproc[(hash->list [hash hash?])
+@defproc[(hash->list [hash hash?] [try-order? any/c #f])
          (listof (cons/c any/c any/c))]{
 Returns a list of the key--value pairs of @racket[hash] in an unspecified order.
 
-See @racket[hash-map] for information about modifying @racket[hash]
-during @racket[hash->list]. @see-also-concurrency-caveat[]}
+If @racket[try-order?] is true, then the order of keys and values passed to
```
mflatt, line 601: Here too, of course.

mflatt (changes requested):

scolobb:
Thank you @mflatt for your comments!  As you suggest, I removed the bogus references to `proc`, fixed the sentences to explain what exactly the ordering is done with respect with, and updated the Racket version.

mflatt:
Thanks for the quick updates!

PR #4041, opened 2021-11-02
Unicode 14.0.0

97jaz:
This change replaces the existing Unicode 7 data files with ones
from Unicode 14 and adjusts tests to account for changes.

[Edit] This PR changes Chez to use Unicode 14 data; it does not affect Racket BC.


97jaz:
@mflatt This is meant to update Chez to use Unicode 14 without adding any new functionality yet. At first, I though the `string-titlecase` code in `5_4.ss` would need to be updated, since the word boundary definition has changed between [Unicode 7](https://www.unicode.org/reports/tr29/tr29-25.html#Word_Boundary_Rules) and [Unicode 14](https://unicode.org/reports/tr29/#Word_Boundary_Rules).

But after looking at the changes, I don't think they would affect titlecasing for the same reason that, e.g., ignoring the `CR × LF` rule doesn't affect it -- as noted in [this comment](https://github.com/racket/racket/blob/master/racket/src/ChezScheme/s/5_4.ss#L436-L438). That is, the changes to word boundaries are all in service of not generating too many boundaries in sequences of non-letters.

Does that seem right to you? Also, do you know of other places in Chez that might be affected by a change in Unicode version?

racket/src/ChezScheme/unicode/extract-info.ss
```diff
@@ -286,15 +291,17 @@
...
         (pretty-print
           `(module ($char-constituent? $char-subsequent? $char-upper-case? $char-lower-case? $char-title-case? $char-alphabetic?
                     $char-numeric? $char-whitespace? $char-cased? $char-case-ignorable? $char-category
                     $wb-aletter? $wb-numeric? $wb-katakana? $wb-extend? $wb-format? $wb-midnum? $wb-midletter?
                     $wb-midnumlet? $wb-extendnumlet? $char-combining-class $char-dump
                     ; UNICODE 7.0.0
-                    $wb-hebrew-letter? $wb-single-quote? $wb-double-quote? $wb-regional-indicator?)
+                    $wb-hebrew-letter? $wb-single-quote? $wb-double-quote? $wb-regional-indicator?
+                    ; UNICODE 14.0.0
+                    $wb-wsegspace? $wb-zwj?)
```
97jaz, line 304: These aren't used yet, so they probably shouldn't be in this PR. I thought I would need them to change the word boundary detection algorithm used in `string-titlecase` but ended up not, as I mentioned in a previous comment.

Both will be needed for [grapheme cluster boundary detection](https://unicode.org/reports/tr29/#Grapheme_Cluster_Boundary_Rules) however, as will an additional data file to provide the `Extended_Pictographic` character property.

mflatt:
@97jaz I've forgotten whatever I used to know about titlecasing. And I can't think of anything that would be affected other than the things that are specifically about characters.

racket/src/ChezScheme/unicode/extract-info.ss
97jaz: Sorry, that's not quite right. The `WSegSpace` property isn't used in the grapheme boundary rules, only the word boundary ones. So I suppose we don't need it at all (unless we want/need to implement the full word boundary detection algorithm).

97jaz:
Thanks @mflatt. I realized that the changes should probably start in the Chez repo, so I'm closing this in favor of https://github.com/racket/ChezScheme/pull/41, which also doesn't include the whitespace-only changes.

mflatt:
Ok - we can work in either repo. I would say that this one is more preferred, though, especially since this is something where we'd want to migrate CS and BC at the same time.

97jaz:
@mflatt Makes sense; I can move it back to this repo.

PR #4043, opened 2021-11-02
Unicode 14

97jaz:
Updating Racket to use Unicode 14


97jaz:
@mflatt I'm trying to implement the approach for strings that you described in the [Rhombus discussion](https://github.com/racket/rhombus-brainstorming/discussions/188#discussioncomment-1578230) -- in fact, I'd like to experiment both with this approach and also one like @samth  described -- but I'd like your opinion on the appropriate place to implement different parts of this.

In this PR, I've added a couple of commits on top of the ones that just update Chez and BC to use Unicode 14 data. The newer commits:
- add the additional Unicode data that's needed to do grapheme breaking
- implement grapheme breaking on strings

Rather than add new Unicode data to each of Chez and BC separately (and then add new string primitives to Racket), I added the data and the grapheme breaking functions directly to `racket/base`, so they can be used by either BC or CS. That all seems to work fine. However, it does mean that Unicode data files are now consumed in yet another place. Also, it means that grapheme I/O also needs to be implemented in the standard library. If it's done in the `io` layer, then `io` will need to pull in the new Unicode data data from `racket/base`, which results in two copies of the same character info table sitting in memory (IIUC).

`read-grapheme` can be implemented easily on top of `read-char` and `peek-char`, though it does require reading a char after peeking at it in cases where the grapheme is larger than a single char.[1] A `peek-grapheme` procedure is more troublesome. I can't think of an efficient way to do that in the Racket library layer without basing it on byte I/O and duplicating UTF-8 decoding. (Though, even if it were implemented in `io`, I think it would require some refactoring to avoid that duplication there, too.) The crux of the problem is that you need to peek ahead more than one char, but you can't do that without knowing the number of bytes that comprised the previous chars. And you just can't know that given the current `peek-char` API (unless you know the byte stream is valid UTF-8).

[1] An implementation of `read-grapheme`:
```racket
(define (read-grapheme [in (current-input-port)])
    (define c1 (read-char in))

    (cond
      [(eof-object? c1) c1]
      [else
       (let loop ([c1 c1] [c2 (peek-char in)] [accum null] [state #b0000])
         (define-syntax-rule (finish)
           (cond [(null? accum) c1]
                 [else (list->string (reverse (cons c1 accum)))]))
         (cond
           [(eof-object? c2) (finish)]
           [else
            (define gb1 (unsafe-grapheme-break-data c1))
            (define gb2 (unsafe-grapheme-break-data c2))
            (define state* (break-between?/state gb1 gb2 state #f))
            (define bound (boundary-value state*))

            (cond
              [(= BOUNDARY-YES bound) (finish)]
              [else (loop (read-char in)
                          (peek-char in)
                          (cons c1 accum)
                          state*)])]))]))
```

mflatt:
Implementing `string-next-grapheme-length` at the `racket/base` level makes sense. For the grapheme data, though, i imagined it would be better to include that with the other data gathered by built-in functions in CS/BC, exposing it as some query on characters. Having that at the CS/BC level should allow port line counting to use the information (where the counting state needs a little more information alongside the UTF-8-decodig status information that it currently hash).

We will need `peek-grapheme`, but I don't yet see why the UTF-8 decode would need to be duplicated. I imagine it would be implemented with `peek-char` and `char-utf-8-length`.

97jaz:
> i imagined it would be better to include that with the other data gathered by built-in functions in CS/BC

I was afraid you'd say that. :)

> [Edit: Maybe it's about encoding errors? In that case, maybe we want to add something like peek-char/length that returns both a character and the number of bytes read to produce the character.]

Yep, that's what I was referring to. Thanks!

97jaz:
@mflatt I just pushed two commits that add `peek-char/length` for CS and BC, respectively. The CS version was implemented in the `io` layer. In Racket, it shows up in `#%kernel` but not in `racket/base`. The BC version, however, is in `racket/base`. Now, I don't have a strong opinion as to whether it should be in `base` or not, but I do want to make sure it's in the same place in both versions. In the CS version, I listed it in the kernel primitives list. I assume that it's not in `base` just because it was't re-provided. But in the case of BC, I don't know why it just automatically shows up in `base`.

mflatt:
If I'm reading the commit right, you didn't yet bump the Racket version number. So, it may be that `raco setup` doesn't know that certain modules need to be recompiled on the grounds that `#%kernel` exports changed. Meanwhile, maybe BC files happened to be in a state where recompilation happened anyway. It's best to bump the version when adding primitives to avoid this problem.

97jaz:
Okay, that makes sense. I think I also didn't bump the primitive count when I added a primitive in a previous commit, which is why the number goes up by two in this one. Thanks!

97jaz:
It appears that implementing grapheme breaking in `racket/base` won't spare me from needing to implement another version in C, since port line counting in BC is implemented in C.

97jaz:
@mflatt You wrote:
> Change port line counting to count positions and columns in graphemes, not characters — either always or by request.
>
> The "always" option would be backward-incompatible, obviously, but it might make sense to take that leap. The "by request" part would need significant new plumbing in the port layer. We should experiment to see how bad the backward-incompatibility hit would be.

I'm not very familiar with the port layer (yet), but would it be difficult to count graphemes _in addition_ to chars and maintain backwards compatibility that way (since grapheme counting will require assembling the UTF-8 into chars anyways)? The existing procedures, like `port-next-location` would continue to work the same way, but a new procedure could return an additional value, or a location struct that includes the grapheme column (or whatever it would be called).

mflatt:
Tracking both would mean reflecting that double tracking in `make-input-port` and `make-output-port`. Anything that makes new ports with those really would need to be upgraded, anyway. Meanwhile, counting graphemes will be the same as counting characters in simple cases (since the current counting treats "\r\n" as one position). That's why I'm guessing that the more simply backward-incompatible path is the easier way, but I might be overlooking something or just gussing wrong.

97jaz:
@mflatt In [`port-count!`](https://github.com/racket/racket/blob/master/racket/src/io/port/count.rkt#L99-L105), the UTF-8 decoding isn't interested in the decoded characters, just their count and the updated state of the decoder. If we're going to do grapheme breaking, though, we'll need the characters themselves (or, at any rate, their grapheme-breaking properties). Since the comment above the procedure says "in atomic mode," could I safely reuse the the same string for any invocation, instead of needing to allocate a fresh string each time? Another alternative would be to decode a character a time to avoid allocation.

mflatt:
> Since the comment above the procedure says "in atomic mode," could I safely reuse the the same string for any invocation, instead of needing to allocate a fresh string each time?

The buffer would have to be place-local, but that could work. It might be simpler and better to allocate a small buffer as part of `loc`, though.

97jaz:
> It might be simpler and better to allocate a small buffer as part of loc, though.

Yeah, that's a much better idea.

PR #4047, opened 2021-11-05
Improve error message for set! of imported identifier.

samth:


mflatt:
My hesitation here is whether "another module" is the right phrasing in a top-level context. I don't have a better idea, though, and maybe it's fine.

mfelleisen:
I'd use "different" rather than "another". Adding "or surrounding" might help the beginner. (Personally I think of module+ as different from the surrounding one.)

samth:
I think the best way to deal with that is to put the module where the identifier is defined in the error message, but I couldn't quite follow the error message protocol to add extra pieces.

sorawee:
Is this ready for merging?

samth:
It would really be better to do the thing I mentioned but this would be an improvement.

PR #4062, opened 2021-11-12
Quote $PSScriptRoot in msvcprep.ps1

VikoRifeo:
This allows the script to work correctly when the path to it contains whitespace.


mflatt:
Thanks!

PR #4063, opened 2021-11-12
Fix some problems with hash operators.

NoahStoryM:
`hash-clear` and `hash-union!` 's error msg-symbol are `hash-clear!` and `hash-union`:

```
Welcome to Racket v8.3 [cs].
> (require racket/hash)
> (hash-clear (make-hasheq '()))
; hash-clear!: contract violation
;   expected: (and/c hash? immutable?)
;   given: '#hasheq()
; [,bt for context]
> (hash-union! (make-hasheq '([1 . one])) #hasheq([1 . one]))
; hash-union: duplicate values for key 1: 'one and 'one [,bt for context]

```

And when we use a longer mutable hash as rest parameter, `hash-union` will fail:
```
Welcome to Racket v8.3 [cs].
> (require racket/hash)
> (hash-union #hasheq() (make-hasheq '([1 . one])))
; hash-clear!: contract violation
;   expected: (and/c hash? immutable?)
;   given: '#hasheq((1 . one))
; [,bt for context]
```

After fixing the problems:

```
Welcome to Racket v8.3 [cs].
> (require racket/hash)
> (hash-clear (make-hasheq '()))
; hash-clear: contract violation
;   expected: (and/c hash? immutable?)
;   given: '#hasheq()
; [,bt for context]
> (hash-union! (make-hasheq '([1 . one])) #hasheq([1 . one]))
; hash-union!: duplicate values for key 1: 'one and 'one [,bt for context]
> (hash-union #hasheq() (make-hasheq '([1 . one])))
'#hasheq((1 . one))
```


mflatt:
Thanks for this repair!

We should add a test to make sure the implementation doesn't regress. The test should be in "pkgs/racket-test-core/tests/racket/hash.rktl" with other `hash-union` tests. (You can run the tests with `racket -f hash.rktl`.) Could you add tests there?

NoahStoryM:
No problem, I have already added the test.

The problem with the original definition is that `hash-union` tries to expand with the longest hash table in the parameters, but ignores the possibility of mutable hash.

mflatt:
Thanks!

PR #4074, opened 2021-12-01
Update line-input.rkt

kimtg:
cross-platform read-line

fixed console IO

Racket's inherent problem: read-line

On Windows, console IO doesn't recognize \r\n unless you put an appropriate value like 'any in the second argument. 

In order to work properly on Windows, the OS with the most users, there is a burden of always using read-line (read-line (current-input-port) 'any) when using read-line.

C, C++, C#, Python, Java, Go, Clojure, and Common Lisp do not have this problem.

If you fix this, I guarantee that the number of Racket users will increase.

In order to expand the base of programming languages, it is necessary to respond to common sense use by ordinary people, but this basic thing is not possible in Racket.


LiberalArtist:
In Racket, line endings for Windows typically are handled by passing a `#:mode 'text` argument to procedures like `open-input-file` and `open-output-file`: https://docs.racket-lang.org/reference/file-ports.html#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._open-input-file%29%29

kimtg:
> In Racket, line endings for Windows typically are handled by passing a `#:mode 'text` argument to procedures like `open-input-file` and `open-output-file`: https://docs.racket-lang.org/reference/file-ports.html#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._open-input-file%29%29

File IO is no problem. I'm talking about console input (keyboard input).

winny-:
This might be related #4007.

sorawee:
[Another user](https://www.reddit.com/r/Racket/comments/rez0cv/how_to_make_a_readline_work_in_drracket_and_the/) from Reddit mentioned this unexpected behavior as well.

It sounds to me that the default `current-input-port` from stdin is at fault here: on Windows, it should do filtering in the same manner as the `'text` mode for `open-input-file`.

mflatt:
I agree that `'any` would have been a better default, and it's the default of `in-lines`. Unfortunately, however, changing an established default can break existing programs.

For example, someone may run a script on Unix that uses `read-line` to input CRLF lines (where the CR is retained), possibly update the line, and then print it back out with a LF. The end result will happen to have CRLF output, currently. Whether or not relying on the current behavior was a good implementation, changing the default for `read-line` would break that script.

So, I think the right approach is to use `'any` in new things and in future languages, but leaving the current behavior of `read-line` from `racket` for backward compatibility.

(Changing `current-input-port` to be in text mode creates all sorts of other problems, because it depends on whether you wanted text input or binary input. We tried things like that in the early days.)

sorawee:
Several people had a discussion about this. One possible compromise is to add `read-line*` that is just like `read-line`, but defaults the mode to `'any`? The documentation of `read-line` can even encourage users to switch to use `read-line*` instead. What do you think, @kimtg?

kimtg:
> Several people had a discussion about this. One possible compromise is to add `read-line*` that is just like `read-line`, but defaults the mode to `'any`? The documentation of `read-line` can even encourage users to switch to use `read-line*` instead. What do you think, @kimtg?

> I agree that `'any` would have been a better default, and it's the default of `in-lines`. Unfortunately, however, changing an established default can break existing programs.
> 
> For example, someone may run a script on Unix that uses `read-line` to input CRLF lines (where the CR is retained), possibly update the line, and then print it back out with a LF. The end result will happen to have CRLF output, currently. Whether or not relying on the current behavior was a good implementation, changing the default for `read-line` would break that script.
> 
> So, I think the right approach is to use `'any` in new things and in future languages, but leaving the current behavior of `read-line` from `racket` for backward compatibility.
> 
> (Changing `current-input-port` to be in text mode creates all sorts of other problems, because it depends on whether you wanted text input or binary input. We tried things like that in the early days.)

read-line*: It's a good idea if you can't give up read-line backwards compatibility.

I brought up the issue of read-line not working natively on Windows about 5 years ago on the Racket mailing list. Even then, the issue was dismissed as a backwards compatibility issue. I understand that the background of the Racket project is to (partially) break Scheme backwards compatibility. (eg immutable list) However, it is unfortunate and very sad that the generality of read-line cannot be embraced again.

spdegabrielle:
This issue bit another Racket user last week. So it seems appropriate to try to mitigate it.

I'm not sure what the best approach;
* Suggest explicit mode of 'any with read-line https://github.com/jackfirth/resyntax/issues/211
* add `read-line*` that is just like `read-line`, but defaults the mode to `'any` https://github.com/racket/racket/pull/4074#issuecomment-992668244
* (and/or) adjust the documentation to warn users of this issue

Are there any other options that have been missed?

cc:@metaxal

kimtg:
Why does read-line behave like Unix in DrRacket on Windows?

samth:
@kimtg I'm not sure what you mean. `read-line` in DrRacket behaves very differently from any command-line Racket program.

kimtg:
> @kimtg I'm not sure what you mean. `read-line` in DrRacket behaves very differently from any command-line Racket program.

read-line in DrRacket: newline is Unix type (LF) even in Windows. Why is it LF and not CR LF in Windows? I think it is inconsistent.

When you run the same program, the read-line behavior is different when you run it with DrRacket and when you run it with just Racket!

rfindler:
We might need some more details here, but just in the hope of adding a little clarification: DrRacket programs and Racket programs should behave the same in this regard, if they are given the same input port -- that is, if you open a file and use read-line on that, or if you open a network port and use read-line on that, it shouldn't matter if you're working in DrRacket or in the command-line racket.

The place where things might differ is the port that's connected to `current-input-port` in the two settings, as those are wholly different implementations and, as Sam points out, behave differently (intentionally, for UI reasons). 

Anyway, if you have a small program that illustrates what you think is wrong, perhaps posting that program will help us understand your concern better.

kimtg:
> We might need some more details here, but just in the hope of adding a little clarification: DrRacket programs and Racket programs should behave the same in this regard, if they are given the same input port -- that is, if you open a file and use read-line on that, or if you open a network port and use read-line on that, it shouldn't matter if you're working in DrRacket or in the command-line racket.
> 
> The place where things might differ is the port that's connected to `current-input-port` in the two settings, as those are wholly different implementations and, as Sam points out, behave differently (intentionally, for UI reasons).
> 
> Anyway, if you have a small program that illustrates what you think is wrong, perhaps posting that program will help us understand your concern better.

I wrote the program below and saw that it worked well on DrRacket, so I proudly made an executable file and gave it to a friend, but it doesn't work properly (on Windows). If you are not going to change the default value of read-line to (read-line (current-input-port) 'any), it is better to have read-line operate platform-specifically in DrRacket to alert you to risks early.

```
#lang racket
(displayln "Division Exercise")
(let loop ([score 0])
  (define b (random 1 10))
  (define c (random 100))
  (define a (* b c))
  (printf "\nScore: ~a\n" score)
  (printf "~a / ~a = " a b)
  (define line (read-line))
  (unless (eof-object? line)
    (cond [(equal? c (string->number line))
           (printf "Correct!\n")
           (loop (add1 score))]
          [else (printf "Wrong! Correct answer is ~a.\n" c)
                (loop score)])))
```

```
Division Exercise

Score: 0
205 / 5 = 41
Wrong! Correct answer is 41.

Score: 0
408 / 6 =
```

PR #4075, opened 2021-12-02
Use monotonic time for timeouts and delays

m4burns:
I noticed random crashes when starting a Racket service at boot on a Raspberry Pi. These crashes were happening just after NTP set the system clock (Raspberry Pi has no RTC). The root cause is that many routines used to generate delays or timeouts use `current-inexact-milliseconds`. This can cause a lengthy hang when the system clock moves back, or premature timeout and shortened delays when the system clock moves forward.

This PR:

- Attempts to replace `current-inexact-milliseconds` with `current-inexact-monotonic-milliseconds` in all locations I could find that seem to really want monotonic milliseconds.
- Adds an optional second argument to `alarm-evt` for using monotonic milliseconds (defaults to UTC milliseconds).
- Changes the already-documented API of `unsafe-poll-ctx-milliseconds-wakeup` to use monotonic milliseconds.

Marking this WIP until tests pass and someone who knows the Racket runtime better than me takes a look. I'm not very confident these changes are safe or that breaking the API of `unsafe-poll-ctx-milliseconds-wakeup` is acceptable.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/evts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/evts.scrbl
@@ -307,13 +307,15 @@ with timeouts that have not yet expired. The system-idle event's
 ]}
 
 
-@defproc[(alarm-evt [msecs real?]) evt?]{
+@defproc[(alarm-evt [msecs real?] [monotonic? boolean? #f]) evt?]{
 
 Returns a @tech{synchronizable event} that is not @tech{ready for synchronization} when
-@racket[(current-inexact-milliseconds)] would return a value that is
+@racket[(_milliseconds)] would return a value that is
 less than @racket[msecs], and it is @tech{ready for synchronization} when
-@racket[(current-inexact-milliseconds)] would return a value that is
-more than @racket[msecs]. @ResultItself{alarm event}.
+@racket[(_milliseconds)] would return a value that is
+more than @racket[msecs]. The value of @racket[_milliseconds] is
+@racket[current-inexact-milliseconds] when @racket[monotonic?] is @racket[#f],
+or @racket[current-inexact-monotonic-milliseconds] otherwise. @ResultItself{alarm event}.
 
 @examples[#:eval evt-eval
   (define alarm (alarm-evt (+ (current-inexact-milliseconds) 100)))
--- a/pkgs/compiler-test/tests/compiler/zo-test.rkt
+++ b/pkgs/compiler-test/tests/compiler/zo-test.rkt
@@ -158,9 +158,10 @@ exec racket -t "$0" -- -s -t 60 -v -R $*
 (define timing-thread
   (thread
    (λ ()
-     (sync 
-      (alarm-evt (+ (current-inexact-milliseconds)
-                    (* 1000 (time-limit)))))
+     (sync
+      (alarm-evt (+ (current-inexact-monotonic-milliseconds)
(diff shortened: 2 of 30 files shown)


m4burns:
@mflatt This seems like your kind of thing :)

m4burns:
Looks like the changes for BC would be a bit more complex.

m4burns:
BC changes were actually not bad at all! Ran most of the tests locally, hopefully everything passes in CI..

m4burns:
Tests are passing, ready for review and feedback.

mflatt:
This is really nicely done! I'll try it out, but reading the diff, everything that I could think of is covered.

I find one use of `unsafe-poll-ctx-milliseconds-wakeup` in my current installation, and that's in `gui-lib`. We should grep all registered packages, and if it shows up nowhere else (which I think is likely), we can just change it. If other packages use `unsafe-poll-ctx-milliseconds-wakeup`, then maybe it needs a bridge along the same lines as `alarm-evt`.

pkgs/racket-doc/scribblings/reference/evts.scrbl
```diff
@@ -307,13 +307,15 @@ with timeouts that have not yet expired. The system-idle event's
 ]}
 
 
-@defproc[(alarm-evt [msecs real?]) evt?]{
+@defproc[(alarm-evt [msecs real?] [monotonic? boolean? #f]) evt?]{
 
 Returns a @tech{synchronizable event} that is not @tech{ready for synchronization} when
-@racket[(current-inexact-milliseconds)] would return a value that is
+@racket[(milliseconds)] would return a value that is
```
mflatt, line 313: Using `_milliseconds` here and each other use would typeset in italics, making more clear that it's a metavariable.

sorawee, line 313: What is `milliseconds` here? As I understand, this is not a function in Racket, right?

sorawee: Ah, it's defined below, my bad.

mflatt:
I've grepped all packages, and the only use of `unsafe-poll-ctx-milliseconds-wakeup` is in `gui-lib`.

pkgs/racket-doc/scribblings/reference/evts.scrbl
m4burns: It's typeset in italic now to avoid confusion.

m4burns: Done

m4burns:
I made the suggested change, added a history note in the docs for `unsafe-poll-ctx-milliseconds-wakeup`, and squashed into one commit. Going to test DrRacket on OS X to make sure this doesn't break `gui-lib`.

m4burns:
DrRacket seems to work fine with these changes on OS X at least

mflatt:
`gui-lib` would only need to change for Unix/Gtk, and I'm happy to make that change.

Now that I've built on my machine, I notice that `alarm-evt` is documented to accept only a `boolean?` second argument, but it actually accepts `any/c`. Also, a `history` note here should mention the new argument.

I'm happy to merge and then make those small documentation updates, or I can wait for you before merging. Just let me know.

m4burns:
Easier for you to make them after merge, thanks! Go ahead.

PR #4077, opened 2021-12-03
cs: mutable hash chaperone-of? intensional equality

AlexKnauth:
Changes `chaperone-of?` so that on mutable hash tables, the two tables underneath must be `eq?` one to be a chaperone of the other.

This should bring Racket CS in line with what Racket BC already does on this.


mflatt:
Thanks! But can you revise to avoid all the whitespace changes? (It would be better if the trailing whitespace weren't there, but changing it makes history more difficult to visit.)

AlexKnauth:
Okay

.gitignore
```diff
@@ -28,6 +28,7 @@ TAGS
 *.swo
 *.swp
 .gdb_history
+/.vscode/
```
rfindler, line 32: This doesn't look like it belongs in a commit with this message?

rfindler:
Should there be some test cases to capture the correct behavior?

.gitignore
AlexKnauth: It just came up while I was editing. I use VS Code to edit C files, and it created that directory which should be ignored by git and not included it the commit.

AlexKnauth:
I added 2 new test cases in `pkgs/racket-test-core/tests/racket/chaperone.rktl`, should there be more?

.gitignore
rfindler: I agree that directory should not be included in the commit but it isn't and I'm not talking about that. I'm saying that if you want to add that to the .gitignore file, it would be better to do it in a separate commit.

(This isn't a bid deal.)

AlexKnauth: Ah okay, I can separate them.

mflatt:
Thanks for the update! It's ok with me to include the ".gitignore" change here.

pkgs/racket-doc/scribblings/reference/chaperones.scrbl
```diff
@@ -142,7 +142,7 @@ chaperones to @racket[v2].
 For values that include no chaperones or other impersonators,
 @racket[v1] and @racket[v2] can be considered chaperones of each other
 if they are @racket[equal?], except that corresponding mutable
-vectors, boxes, strings, byte strings, @tech{mutable pairs}, and
+vectors, boxes, hashes, strings, byte strings, @tech{mutable pairs}, and
```
mflatt, line 145: Just noticed this small point: the reference terminology is "hash tables" instead of "hashes".

mflatt:
The added test case is good, but it's probably a good idea to add ones  for the `eq`, `eqv`, `weak`, and `ephemeron` variants (since those sometimes have different internal representations) while you're here.

rfindler:
Ah, sorry; I missed the test cases that you'd included @AlexKnauth . 

I've built with this commit on my machine to run the contract system tests and they all pass.

AlexKnauth:
Okay, I've added more tests, fixes "hashes" to "hash tables", and separated the ".gitignore" change.

mflatt:
Thanks!

PR #4081, opened 2021-12-06
Add a global search notice to local doc

sorawee:
This PR adds a notice to local docs that search results are local.
It also fixes a JS error about the missing `redo_search_global`.


mflatt:
It looks like this is a copy of code in "main/search.scrbl". The user-scope "main/user/search.scrbl" might just have been overlooked when the note was added for installation-scope documentation, so putting this in a module used by both places seems worthwhile to help avoid that kind of mistake in the future.

sorawee:
Done! And the CI passes, too.

mflatt:
Thanks!

PR #4084, opened 2021-12-06
Fix #4065: also search "package name"

### LINKED ISSUE #4065: Docs search doesn’t index package name
Found a bug in the docs code 
 searching the docs doesn’t find https://docs.racket-lang.org/search/index.html?q=Satore because the package name isn’t in the scribble
 
Presumably the package name isn’t indexed 

n/a - this happens in the browser I think
Unless the index is built at html compile time

**What program did you run?**
Please include a short example that triggers the bug

https://docs.racket-lang.org/search/index.html?q=Satore 

Should have found Satore package docs

e.g., the operating system used or how you are running the code.

sorawee:
This PR adjusts searching so that a regular search term `<term>` also
functions like `T:<term>`, which searches the "package name" (it is
actually the manual directory name), except that:

- In `T:<term>`, the term must match exactly. Here, partial match is OK.
- In `T:<term>`, related results (bindings, etc.) are also presented.
  Here, only the "main" result is presented.

### DIFF
--- a/pkgs/racket-index/scribblings/main/private/search.js
+++ b/pkgs/racket-index/scribblings/main/private/search.js
@@ -413,6 +413,16 @@ function UrlToManual(url) {
 // mostly for context queries.
 
 function CompileTerm(term) {
+  function regularMaker() {
+    var compare_words = CompileWordCompare(term);
+    return function(x) {
+      var r = Compare(term,x[0]);
+      // only bindings can be used for rexact matches
+      if (r >= C_rexact) return (x[3] ? r : C_exact);
+      if (r > C_words3) return r;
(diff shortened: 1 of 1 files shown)


sorawee:
Before:
Screen Shot 2021-12-06 at 2 55 00 AM

After:
Screen Shot 2021-12-06 at 2 55 08 AM

which is similar to `T:satore`:
Screen Shot 2021-12-06 at 2 55 36 AM

Partial match is also OK:
Screen Shot 2021-12-06 at 2 55 22 AM

unlike `T:sator`
Screen Shot 2021-12-06 at 2 55 44 AM

PR #4088, opened 2021-12-10
Docs: AArch64 supports parallel places

otherjoel:


LiberalArtist:
I thought 3M/BC doesn't support aarch64-macosx natively at all.

otherjoel:
You’re right, I goofed this up.

LiberalArtist:
It might still be worth rewording to sound more positive than "only", though, since CS is now the default and IIUC has parallel places on all supported platforms. Maybe something like (if this is accurate)
```
Parallel support for places is enabled on all platforms that support
Racket @tech{CS}, the default implementation of Racket.
The @tech{3m} implementation also supports parallel execution of places
by default on Windows, Linux x86/x86_64, and Mac OS x86/x86_64. To
enable support for other platforms with @tech{3m}, use @DFlag{enable-places} with
@exec{configure} when building Racket. The @racket[place-enabled?]
function reports whether places run in parallel.
```
From my quick search, it looks like `--enable-places` is only accepted by `configure` for BC, which makes sense, since Chez Scheme strongle recommends enabling threads.

It might be worth describing support for places on other platforms with BC as "experimental" or something, since IIRC it was off by default because it had some problems, but "experimental" might not be the right word, since it's not an experiment anyone is pursuing.

PR #4089, opened 2021-12-10
update other-editors guide with modern vim references.

benknoble:
After some Discord chatting, I realized this page was woefully out of date. Here are some attempts at updates.

/cc @spdegabrielle

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/other-editors.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/other-editors.scrbl
@@ -136,78 +136,86 @@ popular among Racketeers as well.
 
 @section{Vim}
 
-Many distributions of Vim ship with support for Scheme, which will
-mostly work for Racket. You can enable filetype detection of Racket
+Many distributions of Vim ship with support for Scheme, which will mostly work
+for Racket. As of @hyperlink["https://github.com/vim/vim/commit/1aeaf8c0e0421f34e51ef674f0c9a182debe77ae"]{version 7.3.518},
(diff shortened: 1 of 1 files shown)


benknoble:
I did not have a great time testing the docs locally, but there' s probably a better process. `raco scribble --dest oe other-editors.scrbl && open oe/other-editors.html` showed the red squigglies under `#lang` and `racket` and `at-exp`. I am probably missing `for-label` references. Not sure which ones to add.

sorawee:
FWIW, this is what I see when building with `raco setup scribblings/guide`: [24.3 Vim.pdf](https://github.com/racket/racket/files/7696104/24.3.Vim.pdf)

benknoble:
> FWIW, this is what I see when building with `raco setup scribblings/guide`: [24.3 Vim.pdf](https://github.com/racket/racket/files/7696104/24.3.Vim.pdf)

Great, that looks like the references were all correct then! 😄

mflatt:
Thanks for the update.

Nit: the capitalization of `Vim` looks inconsistent. Should it be `Vim` everywhere or `vim` everywhere?

benknoble:
> Nit: the capitalization of `Vim` looks inconsistent. Should it be `Vim` everywhere or `vim` everywhere?

I… am not sure. I habitually write vim, but https://www.vim.org/ consistently writes Vim, so I need to correct both the document and the habit :)

benknoble:
@mflatt corrected. I also added a section on REPLs.

PR #4091, opened 2021-12-12
net/http-client: fix handling of 1xx, 204 and 304 responses

Bogdanp:
Without this change, keep-alive requests to servers that return these
status codes end up blocking the connections forever.  These response
types should never contain any content according to:

* https://datatracker.ietf.org/doc/html/rfc2616#section-10.1
* https://datatracker.ietf.org/doc/html/rfc2616#section-10.2.5
* https://datatracker.ietf.org/doc/html/rfc2616#section-10.3.5

Related to Bogdanp/racket-http-easy#15

### DIFF
--- a/pkgs/net-test/tests/net/http-client.rkt
+++ b/pkgs/net-test/tests/net/http-client.rkt
@@ -327,7 +327,7 @@
     '()
     #""])
 
-  ;; Content-less keep alive tests.
+  ;; Body-less keep alive tests.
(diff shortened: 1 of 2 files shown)


Bogdanp:
Apologies for the force-push noise. I figured it was worth making this a little more lenient in case the server does return a body with one of these status codes. This is ready to be reviewed now.

mflatt:
Thanks!

PR #4093, opened 2021-12-13
tests/future: run test under --timeout 90

AlexKnauth:


samth:
Merging this led to a failure here: https://github.com/racket/racket/runs/4509970206?check_suite_focus=true#step:24:20281

AlexKnauth:
I'm borrowing a Linux machine to test times on, and have configured it to be the closest I could get to the configuration in the first failing CI run.

I'm going to start trying things like this:
```time raco test --timeout 90 -l tests/future/future`

AlexKnauth:
For `touch-and-final.rkt` it usually takes less than 3m30s.

But for `future.rkt`, while it usually takes less than 2m10s on the Linux machine I'm running this on, it has gone up to 6 minutes in some CI runs.

Should I try increasing the timeout to 450 seconds, or 7m30s, for `future.rkt`, and 300 seconds for everything else?

samth:
I think changing just that file is the best thing here, and being very generous makes sense. There's a big range between 90 seconds and 6 hours.

AlexKnauth:
I increased the timeouts in https://github.com/racket/racket/commit/b5aa6f7b083cf976f48ac1b9beba13594862d359, and got them to pass.

PR #4096, opened 2021-12-15
Generalize for(*) macros to work with multiple values

sorawee:
**NOTE**: this is a draft PR for discussion. Please do not merge.

This PR allows each iteration of `for(*)/list` to accumulate more or less than one element. Other `for(*)` macros, like `for/hash` and `for/vector` could also be generalized in the same way.

This PR is in the same spirit as #2483 (which is right now still unmerged), but this PR generalizes the existing macros rather than creating entirely new set of macros.

The PR is backward compatible for `#lang racket`, and there is no performance
degradation. Existing uses of `for/list`, which produces only one value in
the body, will generate the same code (compared to before the PR) after
CP0 performs inlining.

The PR requires Typed Racket to be updated, however.

To be done: writing tests, adjusting docs, generalizing other `for` macros.

### DIFF
--- a/racket/collects/racket/private/for.rkt
+++ b/racket/collects/racket/private/for.rkt
@@ -1824,11 +1824,23 @@
     (lambda (x) x)
     (lambda (x) `(,#'begin ,x ,#'(values))))
 
+  ;; rev-append :: list?, list? -> list?
+  (define (rev-append xs ys)
+    (if (null? xs)
+        ys
+        (rev-append (cdr xs) (cons (car xs) ys))))
+
   (define-for-variants (for/list for*/list)
     ([fold-var null])
     (lambda (x) `(,#'alt-reverse ,x))
     (lambda (x) x)
-    (lambda (x) `(,#'cons ,x ,#'fold-var)))
+    (lambda (x)
+      #`(call-with-values
+         (λ () #,x)
+         (case-lambda
+           [() fold-var]
+           [(e) (cons e fold-var)]
+           [xs (rev-append xs fold-var)]))))
 
   (define (grow-vector vec)
     (define n (vector-length vec))


mfelleisen:
A bit of topic or perhaps full on: See John Stone's Algorithms book. (Every also is generalized to accommodate multiple values. Rewrite the book with these new constructs and it may look halfway decent -- and perhaps inspire and/or justify these things.)

rfindler:
I think that this might actually be a performance hit. I know that this pattern of `call-with-values` is optimized because of the contract system, but I don't think it is as fast as just not doing the `call-with-values`. 

Here's my attempt at an experiment.

```
#lang racket

(define (f1 t) (cons (t) 1))

(define (f2 t)
  (call-with-values
   t
   (case-lambda
     [() 1]
     [(x) (cons x 1)]
     [xs 2])))

(define (try f)
  (define (t) 1)
  (time
   (for ([x (in-range 10000000)])
     (f t) (f t) (f t) (f t) (f t)
     (f t) (f t) (f t) (f t) (f t)
     (f t) (f t) (f t) (f t) (f t)
     (f t) (f t) (f t) (f t) (f t)
     (f t) (f t) (f t) (f t) (f t)
     (f t) (f t) (f t) (f t) (f t)
     (f t) (f t) (f t) (f t) (f t))))

(unless (zero? (random 1))
  (set! f1 #f)
  (set! f2 #f)
  (set! try #f))

(try f1)
(try f2)
```

This is the output on my machine:

```
cpu time: 1433 real time: 1482 gc time: 14
cpu time: 2157 real time: 2248 gc time: 16
```

mflatt:
The patch uses `call-with-values` and an immediate `lambda`, which is like using `(lambda () (t))` in place of `t` in your benchmark. For CS with that change, the difference should be small. For BC, though, the difference will still be large, because BC only knows how to optimize the case with a simple-arity receiver.

rfindler:
Yes, it is!

sorawee:
@mfelleisen thanks for the recommendation! I ordered it from the school library, and will read it.

@mflatt that's a great point that Racket BC won't be able to optimize `call-with-values` away. How important is it to preserve BC's performance? Should I not pursue this approach? Do you have any suggestion?

mflatt:
@sorawee I am ready to discount BC performance for something useful. Producing multiple values to control how many items are added to a list is a neat trick, but it doesn't seem so useful to me that I would be inclined to add it. I use `(apply append (for/list ....))` in some places, and I am happy enough with that. Still, I will yield to other opinions.

Minor: In CS, I think there will a small cost in code size for existing uses `for/list` cases where CS cannot tell that the body produces a single value. It's probably fine, and we could measure the change in DrRacket's code footprint as a sanity check.

rfindler:
Seems neato to me!

iitalics:
for/list is probably the most ubiquitous kind of for/* besides just for. so i would be extremely cautious about introducing a performance regression for an extra feature that is not used very often. wouldn't inlining break down if you did (for/list [...] (function)) where (function) is not inlined?

sorawee:
Abandoning the PR for the issue that @iitalics raised.

PR #4103, opened 2021-12-18
Cptypes: Specialize min, max, =, >, ...

gus-massa:
**ChezScheme: check again if first part of seq is omitable:**

This is not 100% necessary and it's one of the commits where the explanation is longer than the change. 

For example in 

    (lambda (x) (when (fixnum? x) (zero? x) 'banana)))

`cp0` does not remove `(zero? x)` because it can raise an error. `cptypes` assumes that all the the first part of a `sequence` has a side effect, because otherwise it would have been removed. But `cptypes` can prove that it will not raise an error and transform it into the unsafe version, but `cptypes`will not check if it has side effects after the reduction. In the second run, `cp0` will notice that it can be removed, so the final result is 

        (lambda (x) (when (fixnum? x) 'banana)))

With this change, after analyzing (and reducing) `(zero? x)`, `cptypes` will check if it's possible to remove it using a shallow check (`drop`) that has a small amount of fuel to avoid quadratic behavior.

This second change is most of the times wasteful, but not very expensive. The advantage is that it can remove some expressions in a single run, and is more similar to the handling of other constructions.

When I wrote the code (1 year ago?) I preferred to avoid this second check, but now I like it. (I'm not sure what I'll think next year.)

---

**ChezScheme: specialize = and max**

This add a specialization in `cptypes` for some primitives that never raise errors when all the arguments are `fixnum?` or `flonums?`, for example `min`, `max`, `=`, `>`, ... , so

    (lambda (x) (when (fixnum? x) (min x 7)))

is reduced to

    (lambda (x) (when (fixnum? x) (#3%fxmin x 7)))

In particular, this re-enable some of the tests in optimize.rktl,

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -3016,14 +3016,13 @@
   (test-use-unsafe-fxbinary 'fxior 'unsafe-fxior)
   (test-use-unsafe-fxbinary 'fxxor 'unsafe-fxxor)
 
-  (unless (eq? 'chez-scheme (system-type 'vm)) ; cptypes doesn't currently convert to fixnum ops
-    (test-use-unsafe-fxbinary '= 'unsafe-fx=)
-    (test-use-unsafe-fxbinary '< 'unsafe-fx<)
-    (test-use-unsafe-fxbinary '> 'unsafe-fx>)
-    (test-use-unsafe-fxbinary '<= 'unsafe-fx<=)
-    (test-use-unsafe-fxbinary '>= 'unsafe-fx>=)
-    (test-use-unsafe-fxbinary 'min 'unsafe-fxmin)
-    (test-use-unsafe-fxbinary 'max 'unsafe-fxmax))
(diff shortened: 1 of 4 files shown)


gus-massa:
**ChezScheme: Filter the argument type in unsafe primitives**

In an expression like

    (#3%zero? <maybe-fixnum>)

the compiler can assume the argument is correct, so the handler for the specialization will get only `<fixnum>` instead of `<maybe-fixnum>`.

It also detect a few more possible errors, for example to reduce

    (vector (unbox x) (car x))   ==> (begin (unbox x) (car x))

And also, remove an unnecessary check in the specialization of `cdr`.

mflatt:
These changes seem fine to me: they seem useful, and I'm not seeing anything that would go wrong.

mflatt:
I think something has gone wrong here with primitives that have zero arguments. The cptypes pass crashes on this example at the Chez Scheme level:

```
(lambda ()
  (flround
   1
   (let ([app_4 (flmax)])
     (values app_4 (values)))))
```

The problem is related to handling `(values)`, where `make-seq*` is called with an empty list of arguments because a bottom type is involved.

PR #4104, opened 2021-12-19
realms for error-message adjustment

mflatt:
_The changes here seem too big to include in v8.4 — not conceptually all that large, but the changes touch a lot of code. Also, it's not obviously the right idea for solving the problem at hand, although it's the best I've been able to find._

The change in this PR is aimed primarily at the problem of error reporting when a language is built on top of Racket and wants to use different names for the primitives; in that case, Rackety error reporting from a primitive is not right for the new language, but wrapping primitives in an obvious way would introduce extra checks and otherwise interfere with optimization. At the same time, the changes here can help libraries (Racket or otherwise) that wrap a not-so-primitive function where it's inconvenient or inefficient to perform redundant checks on arguments, but where tracking context to adjust error reporting can be made cheap relative to the operation.

The first new ingredient is the notion of a "realm". Roughly, everywhere that Racket deals with procedure names, it now also has a realm. That means attaching a realm to a procedure, but also means generalizing error-constructing functions like `raise-argument-error` that take a symbol name; now there's `raise-argument-error*`, which takes a name and a realm. Realms are not just for function-name conventions, but also for the notation used to describe contracts on function arguments and results (and maybe more things we discover in the future).

The main use of a realm is to recognize function names and other error-message text to rewrite them. Granted, rewriting as a way of adjusting error messages isn't pretty, but it has been an effective strategy for adjusting Racket error messages in the HtDP teaching languages or for adjusting Chez Scheme error messages in Racket CS. Those rewriters work because they sit on a known and relatively static base. A language like Rhombus, however, needs to not only use Racket's primitives, but also interact with other Racket libraries; with error-message rewriting, there's a danger that error messages from the libraries could be mistaken for things that need to be rewritten. Information that an error message is from a realm like `'racket/primitive` helps an error-message rewriter avoid adjusting error messages that it shouldn't adjust.

So, that's the strategy here for primitives: make them recognizable through a `'racket/primitive` realm, and then the run-time configuration for a language like Rhombus can set a `current-error-adjuster` parameter to rewrite messages. The error adjuster is used only on an error path, so it has no performance effect on primitives as used by expanded Rhombus code.

A global parameter doesn't compose nicely for multi-languages applications. It also doesn't take into account the context around an error, which might involve trips through different language layers. For very-primitive primitives like `car` or `vector-ref`, cheaply and precisely tracking language boundaries is at best difficult and maybe just impractical. For the near future, having primitive errors presented in a designated main language (determined the same way that value printing is currently determined) seems practical and useful.

Still, this PR doesn't give up entirely on racking context. For many functions that a language or library would like to wrap with a different name or error convention, a continuation mark has a perfectly acceptable cost. So, alongside the global
`current-error-adjuster` parameter, there's `error-message-adjuster-key` to be used as a continuation-mark key. When an error message is rewritten, the stack of adjusters associated to the current continuation is applied first, and then the value of `current-error-adjuster` is used last. This arrangement is similar to Racket's stack of exception handlers plus its global unhandled-exception handler.

Functions that create error messages have to opt into adjustment. Functions like `raise-argument-error` and `raise-argument-error*` will call adjusters. In general, an error-message creator can use `error-message->adjusted-string` and/or `error-contract->adjusted-string` to trigger adjustment. The `raise-syntax-error` and `raise-read-error` (which is internal?) functions do not trigger adjustment, on the grounds that they're static errors that are already in context, and the problem is with dynamic errors where the cause may span realms.

This PR does not address the problem of how support for an arbitrary language X manages to translate a message whose realm is Y. Maybe in some future there's a universal language of names and contracts so that X only needs to be able to translate from there, and maybe a contract subsystem can jut do that kind of thing automatically.

Smaller details:

 * Most procedures get their names by inference in the compiler. To configure the realm for procedures created in a module, use the new `#:realm` declaration form with `#%declare`, as in `(#%declare #:realm rhombus)`. More generally, there's a `current-compile-realm` parameter, and declaring a realm in a module effective overrides that parameter for the module's code. A module retains its realm, and `module-compiled-realm` or `module->realm` can extract it.

 * Functions like `procedure-rename` now take an extra optional argument for the realm. The keyword-argument layer has to accomodate that, too.

 * The `continuation-mark-set->context` function takes an optional argument to report realm information alongside name and source-location information. That information could potentially be used to detect realm crossings, although inlingin and such will make the results approximate.

 * The `make-struct-field-accessor`, `make-struct-field-mutator`, and `make-struct-type-property` functions take new arguments to configure the name, realm, and argument contract of accessors and mutators (for the errors that they potentially report). This creates ripple effects in compiler passes at the schemify, expander-flattener, and BC level that need to recognize pure constructions and inline error generation.

 * As of the initial PR, the BC implementation is not complete, but most of the pieces are in place.

 * It's probably going to take a while to properly sort out the `'racket/primtive` versus `'racket` realms. The intent is that `'racket/primitive` is relatively static and doesn't have any overlapping names within the realm. Also, there are likely many more places that will need to use `error-message->adjusted-string`, and it will take time to get to all those; fortunately, since this is "only" about error messages, it adapts well to partial support.

 * A lot of the remaining work is writing tests (and fixing bugs they expose).


samth:
This is very comprehensive, but I worry about continuing even further down the road of representing all information about errors in strings. Perhaps it is already too late (perhaps by 25 years) but given that this PR contemplates a large amount of code change, maybe it's worth introducing more structure to avoid repeated parsing of error messages.

Some things of course need to be treated as strings (or things like strings), such as `error-primitive-name->symbol-handler`. But `error-primitive-contract->string-handler`, for example, seems like a case where the string format makes things harder. Similarly, `error-primitive-message->string-handler` could take more structure. 

For the latter, one idea would be to structure the message into sections, as a lot of messages already do (for example, the `expected:` and `given:` lines produced by `raise-contract-error`). 

Maybe rewriting is needed because the Chez Scheme layer will need rewriting anyway, or because it's too hard to change all the current locations or to keep the in sync, but I feel like it's worth at least considering a different direction.

mflatt:
@samth I agree with your concern. One thing I tell myself is that this approach would adapt to strategies that rely less on strings. In the same way that a contract rewriter gets used before putting the result into a string that then goes to a message-string rewriter, there could be new protocols for generating some of the other existing strings in the first place. To put it another way, changes that rely less on strings would (probably?) not skip any of the steps here, but instead add more steps.

samth:
That seems sensible to me, but I wonder two things specific to the PR:

- if we build more infrastructure that relies on strings (eg code that uses `error-primitive-message->string-handler`) now does that make more work for doing something more structured later?
- if we're (really you're) changing all this now and introducing new API, is it the right time to go further and change more/introduce more new API?

I don't have a firm opinion on either question yet.

mflatt:
For that discussion, maybe it would help if I say more about the protocol implemented in this PR.

The value of `current-error-adjuster` or a value associated with `error-adjuster-key` is a procedure that takes a mode symbol. A mode symbols is currently one of `'message`, `'name`, or '`contract`, but the intent is that new mode symbols will be added in the future. An adjuster function returns `#f` if its doesn't want or know how to handle a requested mode. Otherwise, and adjuster returns a function that works in a mode-specific way. For the `'contract` mode, the arguments and results are a string and a realm symbol. For the `'name` mode, the arguments and results are a symbol and a realm symbol. For the `'message` mode, the arguments and results are a name symbol or `#f`, a name-realm symbol, a message string, and a message-realm symbol.

The `error-primitive-message->string-handler` function currently tries `'message` mode, but if an adjuster doesn't support that, it also tries `'name` mode, which is represents a subset of the `'message` mode behavior. That illustrates how the set of modes is not only meant to be extensible, but also potentially implement varying levels of specificity and fallbacks.

The `raise-argument-error` function uses all three support modes. It uses `'contract` mode to potentially rewrite an individual contract string. Then, it puts that contract in a message, and it uses the other two modes to arrive at a final error message.

Currently, the default value of `current-error-adjuster` complains if it's given a mode that is not one of the currently defined modes. Maybe it just should return `#f` in that case, and then the set of modes can be extended by layers other than the primitive layer? (If the protocol is on the the right track at all, that is.)

mfelleisen:
This last para suggests that you may wish to compose these adjusters. Imagine hosted languages such as those proposed by Michael B.  The ultimate LOP program in this direction isn't a composition of #lang languages but layers upon layers of small languages within a single module.  These layers will not know about each other; they just support extension points. It's up to the "composer" to create glue. So `#f` seems appropriate and passing on adjustment responsibilities will hopefully create the right error message. (That's orthogonal to Sam's string-vs-struct comment. But I will say his comment almost points into the algebraic effect direction.)

gus-massa:
I think Chez Scheme already has a structured representation of the messages of the errors, but I never played too much with it.

PR #4105, opened 2021-12-19
cs: don't mix-hash-code for impersonator case

AlexKnauth:
In the `impersonator?` case of `equal-hash-loop`, don't use `mix-hash-code` on that recursive call.

Impersonators of the same object can be `equal?` to that object and each other, their hash codes should also be equal without mixing getting in the way.


AlexKnauth:
I'm having trouble coming up with a test that succeeds after this change but fails before this change that doesn't involve `equal-always?`... I'll keep looking

mflatt:
I agree with the change, and I also can't construct an example. I think that may be dead code as far as impersonators go, because all the primitive impersonators have an `equal?` implementation.

PR #4112, opened 2021-12-29
place: allow `resolved-module-path?` args to `dynamic-place`

Bogdanp:
I think this is meant to work and the existing check is incomplete.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/places.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/places.scrbl
@@ -150,7 +150,7 @@ are simulated using @racket[thread].}
 }
 
 
-@defproc[(dynamic-place [module-path (or/c module-path? path?)]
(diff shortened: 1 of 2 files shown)


mflatt:
I think this might work currently with Racket CS, but despite the predicate at the C level in the BC implementation, I don't think it can work there currently, because a resolved module path is not a valid place message. Because of the way that resolved module paths are meant to be interned, it's not clear that it works right in CS, either.

How useful would it be to make this work, as opposed to having a client convert the resolved module path back to an S-expression module path?

Bogdanp:
I was trying to get `dynamic-place` to work within an executable that references a module using `define-runtime-module-path`. It didn't click with me that I could resolve the module path then quote it, and everything would work until I read your second paragraph. That seems to work fine, so I don't think this is necessary after all. Thanks!

PR #4113, opened 2021-12-29
Chez Scheme: add reduction for unbox and cadr in cp0

gus-massa:
Reduce

    (unbox (box <x>)) ==> <x>
    (cadr (list <x> <y> <z>)) ==> (begin <x> <z> <y>)

I'm not sure how usual is this kind of patterns, but this "fix" two of the optimizations checks in optimize.rktl. There are other two checks that still fail because cp0 needs a third pass to simplify them. For example:

    #lang racket
    (unbox (box 5))

==>

    #lang ChezScheme
    (let ([temp (box 5)])
      (if (box? temp)
        (unbox temp)
        (app/noinline slow-version-of-unbox temp)))

==> fists pass (cptypes)

    #lang ChezScheme
    (let ([temp (box 5)])
      (#3%unbox temp))

==> second pass (cp0)

    #lang ChezScheme
    (#3%unbox (box 5))

==> "third" pass (cp0)

    #lang ChezScheme
    5


sorawee:
Shouldn't it be the following?

(cadr (list <x> <y> <z>)) ==> (begin <x> (begin0 <y> <z>))

This matters when `<y>` and `<z>` have side-effect.

olopierpa:
Il giorno 29 dicembre 2021, alle ore 18:12, gus-massa ***@***.***> ha scritto:

>Reduce
>(cadr (list <x> <y> <z>)) ==> (begin <x> <z> <y>) 

Maybe I'm missing the context, but shouldn't the evaluation order be preserved?

gus-massa:
This is the problem of mixing Racket code and Chez Scheme code. I should have been more explicit.

In Chez Scheme the arguments can be evaluated in any order, so the code in Racket is transformed by `schemify` to code in Chez Scheme that enforces the order. The transformation is something like

    #racket
    (cadr (list (display 1) (display 2) (display 3)))

==> schemify

    #lang ChezScheme
    (cadr (let ([x1 (display 1)])
            (let ([x2 (display 2)])
              (list x1 x2 (display 3)))))

==> cp0*2

    #lang ChezScheme
    (begin
      (display 1)
      (let ([x2 (display 2)])
        (display 3)
        x2))

mflatt:
LGTM

Issue #4108, opened 2021-12-23
Raco unable to install packages on macOS due to SSL/TLS certificate issues


yonkeltron:
**What version of Racket are you using?**
8.3 [cs]

I installed this on my macOS device with `brew install --cask drracket` and everything works great until I tried to install a package on my macOS device. I was UNABLE to replicate the issue when running the same command inside a Docker container on the machine in question.

**What program did you run?**

I ran raco as per the below:

```bash
$ raco pkg install --auto beautiful-racket
```

This exited with an error related to SSL/TLS certificates. I work in a large corporate environment where encrypted traffic is cracked at the edge. Raco did not respect my system's certificate chain and I'm unable to figure out how to specify the correct certificate bundle it should use.

**What should have happened?**

Raco should have been able to install the package on macOS as it can easily do inside the Docker container running on the same machine.

**If you got an error message, please include it here.**

```bash
$ raco pkg install --auto beautiful-racket
Resolving "beautiful-racket" via https://download.racket-lang.org/releases/8.3/catalog/
Resolving "beautiful-racket" via https://pkgs.racket-lang.org
Downloading repository https://github.com/mbutterick/beautiful-racket.git?path=beautiful-racket commit b0d133f4ba4149ea506d8b0dbd3845c915efdd06
ssl-connect: connect failed (error:1416F086:SSL routines:tls_process_server_certificate:certificate verify failed)
  context...:
   /Applications/Racket v8.3/collects/openssl/mzssl.rkt:606:0: error/network
   /Applications/Racket v8.3/collects/openssl/mzssl.rkt:1590:0: wrap-ports
   /Applications/Racket v8.3/collects/racket/contract/private/arrow-val-first.rkt:555:3
   /Applications/Racket v8.3/collects/net/http-client.rkt:67:0: http-conn-open!
   /Applications/Racket v8.3/collects/net/http-client.rkt:274:0: http-conn-open
   /Applications/Racket v8.3/collects/racket/contract/private/arrow-val-first.rkt:555:3
   /Applications/Racket v8.3/collects/net/url.rkt:202:0: http://getpost-impure-port
   /Applications/Racket v8.3/collects/net/url.rkt:305:0: get-pure-port/headers
   /Applications/Racket v8.3/collects/racket/contract/private/arrow-val-first.rkt:555:3
   /Applications/Racket v8.3/collects/net/git-checkout.rkt:245:0: initial-connect
[...]

PR #4117, opened 2022-01-05
openssl: use SecTrustCopyAnchorCertificates by default on Mac OS

rmculpepper:
This might fix the issue described in #4108. If it does, it's probably preferable to adding additional keychains and trying to filter them based on trust settings.

Docs need history notes.


mflatt:
This looks good to me, and it works on my machine.

Issue #4066, opened 2021-11-14
`trace-define` fails in `local`


jasonhemann:
I thought I could use `trace-define` in place of any `define`. Should I expect this ASL program to work? Right now it signals an error.

```
(require racket/trace)

(define (foo xs) 
    (local [(trace-define (inner xs)
              (if (null? xs) #t (inner (cdr xs))))]
      (inner xs)))
```

I do require the `racket/trace` module here. Using `trace-define` in `local` in `#lang racket` is also an error. 

```
; /Applications/Racket v8.2/collects/racket/trace.rkt:323:63: local: not a definition
;   at: (trace inner)
;   in: (local ((trace-define (inner xs) (if (null? xs) #t (inner (cdr xs))))) (inner xs))
```

michaelballantyne:
It looks like `(trace-define (inner xs) ...)` expands to:

```
(begin
  (define (inner xs) ...)
  (trace inner))
```

This works in definition contexts that allow a mix of definitions and expressions, like `let`. But `local` only allows definitions, so it raises an error at the `(trace inner)` expression.

I think it could work to modify the `trace-define` macro to instead expand to:

```
(begin
  (define (inner xs) ...)
  (define _ (trace inner))) 
```

sorawee:
Fixed. Closing.

PR #4118, opened 2022-01-06
Forces trace-define to expand trace as definition

jasonhemann:
To address #4066 .


LiberalArtist:
I wonder what the right thing is for macro authors to do about this in general. Does checking for a ["liberal expansion" context](https://docs.racket-lang.org/reference/stxtrans.html#%28tech._liberal._expansion%29) cover this? Or maybe there should be a test for a stratified expansion context?

PR #4120, opened 2022-01-08
Chez Scheme: add predicate substract in cptypes

gus-massa:
When a test is false, sometimes `cptypes` can remove that possibility and get a more precise type to
use in future checks.

For example in

    (lambda (x)
      (when (list? x)
        (unless (null? x)
          (pair? x))))

`cptypes` can prove that `(pair? x)` is `#t`.

And in

    (lambda (x)
      (let [t (get-u8 x])
        (unless (eof-object? t)
          (fixnum? t))))

`cptypes` can prove that `(fixnum? x)` is `#t`.

Up to now, this was done only to exclude `#f` from a type, and transform for example `'maybe-char` in `'char`. 

Anyway, this implementation is restricted to cases where the result is representable in the type lattice, so some optimization opportunities are missed. (I expect to complete the system in a few months.)

In the second commit, this is extended to `(eq? x <cnt>)`. The implementation not straighforward, because I was thinking for a similar case for `(fx< 3 x 5)`, and it would be nice to drop one of the inequalities. I'm not sure if it's worth to add a case for `(eq? a b c ...)`.

The third commit changes the internal representation of types composed by `immediates`. The previous representation has 6 cherry picked combinations and need like 36 special cases to calculate `union`/`intersect`/`substract`. The new representation use bits, so the calculation is easier, but there are some special cases to combine `chars`.


mflatt:
LGTM

PR #4123, opened 2022-01-10
Update README.md to include other racket channels

spdegabrielle:


mflatt:
I'd be happy to merge this, but what do you think about moving the mailing-list link to the end of the list of channels and dropping "development"? The other channels are much more active.

spdegabrielle:
> I'd be happy to merge this, but what do you think about moving the mailing-list link to the end of the list of channels and dropping "development"? The other channels are much more active.

I'd suggest removing the mailing list.

mflatt:
> I'd suggest removing the mailing list and IRC.

That sounds fine to me.

spdegabrielle:
I also file the link to go to Building, Distributing, and Contributing to Racket instead of build.md

spdegabrielle:
and updated build.md to point to the definitive document

sorawee:
Can you revert changes to `build.md`, at least for now?

`build.md` is automatically generated by https://github.com/racket/racket/blob/master/.github/workflows/scribble_build-guide.yml. So if we really want to take it out, the above file should be removed as well. There are also various mentioning of `build.md` (e.g., in CONTRIBUTING.md) that must be fixed too.

But I'm not sure if we should remove this file.

One possibility that will point readers to the web version, while still providing the Markdown version (which could be useful, perhaps because the reader has no internet access, and does not have a working Racket yet), is write something along the line "The online version can be found at https://docs.racket-lang.org/racket-build-guide/" at https://github.com/racket/racket/blob/master/pkgs/racket-build-guide/racket-build-guide.scrbl, and somehow makes it so that only the Markdown renderer will render it.

spdegabrielle:
will do another PR for readme only

PR #4128, opened 2022-01-16
Remove \#* .\#*

spdegabrielle:
Remove
\#*
.\#*
As per https://racket.discourse.group/t/default-gitignore-for-racket/534/12?u=spdegabrielle

### DIFF
--- a/racket/collects/pkg/private/new.rkt
+++ b/racket/collects/pkg/private/new.rkt
@@ -115,8 +115,6 @@ EOS
       (with-output-to-file ".gitignore"
         (lambda () (display #<<EOS
 *~
-\#*
-.\#*
 .DS_Store
 compiled/
 /doc/


racket/collects/pkg/private/new.rkt
```diff
@@ -115,8 +115,6 @@ EOS
       (with-output-to-file ".gitignore"
         (lambda () (display #<<EOS
 *~
-\#*
```
sorawee, line 118: Why is `\#*` removed? IIUC, that's the format for backup files in DrRacket, and has nothing to do with the issue that @greghendershott was raising.

sorawee: Actually, disregard what I wrote above. `#*` files are actually "autosave files" and they function just like `.#*` for Emacs, so I guess it makes sense to remove them both.

spdegabrielle:
@sorawee i am closing this unmerged as there is a lack of agreement on what the right thing to do is.

PR #4138, opened 2022-01-24
cs: tpb build fixes

Bogdanp:
These are fixes for some problems I encountered while trying out https://github.com/racket/racket/commit/4480e643daa5ec13c4d41c4e002223d86c1fa645 on my x86-64 MBP. There may be better approaches to each of these, in which case you should feel free to ignore this PR!

With these changes in place, I can configure a build with

```
$ ../configure --enable-pb --enable-racket=auto
```

and end up with

```
$ ./cs/c/racketcs -l racket/base -i
Welcome to Racket v8.4.0.4 [cs].
> (system-type 'vm)
'chez-scheme
> (system-type 'arch)
'pb
> (require racket/future)
> (futures-enabled?)
#t
```

### DIFF
--- a/racket/src/ChezScheme/c/atomic.h
+++ b/racket/src/ChezScheme/c/atomic.h
@@ -97,7 +97,7 @@ FORCEINLINE int S_cas_any_fence(int load_acquire, volatile void *addr, void *old
 # define CAS_LOAD_ACQUIRE(a, old, new) S_cas_any_fence(1, a, old, new)
 # define CAS_STORE_RELEASE(a, old, new) S_cas_any_fence(0, a, old, new)
 #elif (__GNUC__ >= 5) || defined(__clang__)
-# define CAS_ANY_FENCE(a, old, new) __sync_bool_compare_and_swap(a, old, new)
+# define CAS_ANY_FENCE(a, old, new) __sync_bool_compare_and_swap((uptr *)a, (uptr)old, (uptr)new)
 #elif defined(_MSC_VER)
 # if ptr_bits == 64
 #  define CAS_ANY_FENCE(a, old, new) (_InterlockedCompareExchange64((__int64 *)(a), (__int64)(new), (__int64)(old)) == (__int64)(old))
--- a/racket/src/ChezScheme/rktboot/make-boot.rkt
+++ b/racket/src/ChezScheme/rktboot/make-boot.rkt
@@ -41,6 +41,13 @@
   (printf "~a\n" msg)
   (flush-output))
 
+(define target-arch
+  (case target-machine
+    [("pb32" "tpb" "tpb32")
+     (begin0 'pb
+       (status (format "Target arch for ~a machine is pb" target-machine)))]
+    [else (string->symbol target-machine)]))
+
 (define sources-date
   (for/fold ([d 0]) ([dir (in-list (list here-dir
                                          nano-dir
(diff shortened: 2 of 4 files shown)


mflatt:
Thanks! I'll take a closer look soon.

The `adr` instruction has only 20 bits available for the displacement, so I don't think that change can be right. But maybe the displacement is currently in bytes, and probably it could be in instructions, in which case the displacement could drop the low two bits. (I worked around the problem in initial experiments by using `--disable-wpo`.)

Bogdanp:
Thanks! Re. the `asm-return-address` change, I somehow missed that the reg was packed in the instruction as well, so I'm going to go ahead and drop that commit since it's definitely wrong. 

Experimenting some more, another problem I noticed is that the `so-suffix` in rumble defaults to `.so`, so loading ffi libs at the Racket level doesn't currently work on macOS.

racket/src/ChezScheme/c/atomic.h
```diff
@@ -97,7 +97,7 @@ FORCEINLINE int S_cas_any_fence(int load_acquire, volatile void *addr, void *old
 # define CAS_LOAD_ACQUIRE(a, old, new) S_cas_any_fence(1, a, old, new)
 # define CAS_STORE_RELEASE(a, old, new) S_cas_any_fence(0, a, old, new)
 #elif (__GNUC__ >= 5) || defined(__clang__)
-# define CAS_ANY_FENCE(a, old, new) __sync_bool_compare_and_swap(a, old, new)
+# define CAS_ANY_FENCE(a, old, new) __sync_bool_compare_and_swap((uptr *)a, (uptr)old, (uptr)new)
```
mflatt, line 100: Either `uptr` or `ptr` works. But since this is a C macro, it's best to have parentheses around `a`, `old`, and `new`, so I'll merge a variant that does that.

racket/src/ChezScheme/rktboot/make-boot.rkt
```diff
@@ -255,7 +262,7 @@
               [(constant-case architecture [else e ...])
                (loop #`(begin e ...))]
               [(constant-case architecture [(arch ...) e ...] . _)
-               (memq (string->symbol target-machine) (syntax->datum #'(arch ...)))
+               (memq target-arch (syntax->datum #'(arch ...)))
```
mflatt, line 265: I'll push a more general repair here, which is to use the `architecture` constant as defined earlier by the machine's ".def" file.

racket/src/cfg-cs
```diff
@@ -2272,6 +2272,9 @@ if test "${enable_racket}" = "auto" ; then
 fi
 RUN_LOCAL_RACKET="local/cs/c/racketcs"
 CONFIGURE_LOCAL_RACKET=--enable-cs
+if test "${enable_pb}" = "yes" ; then
+  CONFIGURE_LOCAL_RACKET="${CONFIGURE_LOCAL_RACKET} --enable-pb"
```
mflatt, line 2276: I see why this makes the build work, but I think the right repair involves corrections to the main CS `configure.ac` and `Makefile.in`. (The general constraint on `--enable-racket=auto` mode is that the host Racket build must work without any configuration, so that means the host build shouldn't require `--enable-pb` itself, so that's why this workaround isn't really the right approach.)

mflatt:
I've merged changes that address the same issues as here. I haven't yet addressed the problem with reporting information about actual platform for queries like `(system-type)` or `(system-type 'so-suffix)`.

A `configure --enable-pb --enable-racket=auto` is now treated like a cross compilation. Along those lines, the automatically created `racket` runs in native mode instead of with `--enable-pb`. A plain `configure --enable-pb` should work, though, as long as `pb` bootfiles are available. The `pb` bootfiles are in in a source distribution or fetched by `make pb-fetch` in the Git repo root; that's the long-term route for when BC is eventually discontinued, anyway.

Bogdanp:
Great! Thank you!

PR #4139, opened 2022-01-25
Simplify `for/fold/derived` examples

jackfirth:
This updates the examples for `for/fold/derived` and `for*/fold/derived` to use `define-syntax-parse-rule` instead of `syntax-case`. This makes the macro examples much simpler and easier to read.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/for.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/for.scrbl
@@ -1,6 +1,8 @@
 #lang scribble/doc
 @(require "mz.rkt"
-          (for-label syntax/for-body))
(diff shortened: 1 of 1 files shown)


sorawee:
Is this the first time `syntax/parse` is used in the Racket reference? It seems to me that `syntax/parse` (or anything outside `#lang racket`) was purposefully avoided in the reference.

- If my understanding is correct, then perhaps we should not introduce `syntax/parse` here.
- But if not, there probably are a lot more places to make similar simplification.

jackfirth:
I don't think it was purposefully avoided. Rather, most of the reference examples simply predate the existence of `syntax-parse`.

PR #4146, opened 2022-01-30
doc: revamp explanation of implicit begin

sorawee:
Thanks to @rmculpepper for the initial improvement.

I end up explaining it a bit differently in a way that I think flows a bit better. Let me know if there's any inaccuracy. 

cc: @mflatt, @rmculpepper

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/begin.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/begin.scrbl
@@ -34,24 +34,16 @@ tail position with respect to the @racket[begin] form.
 (print-triangle 4)
 ]
 
-Many forms, such as @racket[lambda] or @racket[cond] support a
-sequence of expressions even without a @racket[begin]. Such positions are
-sometimes said to have an @deftech{implicit begin}.
-
-@defexamples[
-(define (print-triangle height)
-  (cond
-    [(positive? height)
-     (display (make-string height #\*))
-     (newline)
-     (print-triangle (sub1 height))]))
-(print-triangle 4)
-]
-
 The @racket[begin] form is special at the top level, at module level,
-or as a @racket[body] after only internal definitions. In those
-positions, instead of forming an expression, the content of
+or as a @racket[_body] (e.g. of @racket[let] or @racket[lambda];
+see @tech{internal definition} for more details).
+Such @racket[begin] form has a slightly different grammar:
+
+@specform[(begin def-or-expr ...)]{}
+
+In those positions, instead of forming an expression, the content of
 @racket[begin] is spliced into the surrounding context.
+The content could contain definitions.
 
 @defexamples[
 (let ([curly 0])
@@ -63,6 +55,22 @@ positions, instead of forming an expression, the content of
 
 This splicing behavior is mainly useful for macros, as we discuss
 later in @secref["macros"].
+
+The whole @racket[_body] sequence is sometimes said to have an @deftech{implicit begin}.
+The restriction of an @tech{implicit begin} follows the restriction of the @racket[_body] sequence:
+the sequence should end with an expression.
+
+@defexamples[
+(code:comment @#,t{Implicit @racket[begin]})
+(let ()
+  (define num 1)
+  (add1 num))
+(code:comment @#,t{Explicit @racket[begin]})
+(let ()
+  (begin
+    (define num 1)
+    (add1 num)))
+]
 
 @;------------------------------------------------------------------------
 @section{Effects After: @racket[begin0]}
--- a/pkgs/racket-doc/scribblings/guide/define.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/define.scrbl
@@ -194,10 +194,10 @@ equivalent to a @racket[define-values] form with a single @racket[_id].
 
 When the grammar for a syntactic form specifies @racket[_body], then
 the corresponding form can be either a definition or an expression.
-A definition as a @racket[_body] is an @defterm{internal definition}.
+A definition as a @racket[_body] is an @deftech{internal definition}.
 
 Expressions and internal definitions in a @racket[_body] sequence can
-be mixed, as long as the last @racket[_body] is an expression.
+be mixed, as long as the sequence ends with an expression.
 
 For example, the syntax of @racket[lambda] is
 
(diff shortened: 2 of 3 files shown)


pkgs/racket-doc/scribblings/guide/begin.scrbl
```diff
@@ -34,24 +34,16 @@ tail position with respect to the @racket[begin] form.
 (print-triangle 4)
 ]
 
-Many forms, such as @racket[lambda] or @racket[cond] support a
-sequence of expressions even without a @racket[begin]. Such positions are
-sometimes said to have an @deftech{implicit begin}.
+The @racket[begin] form is special at the top level, at module level,
+or as a @racket[_body] (e.g. of @racket[let] or @racket[lambda];
+see @tech{internal definition} for more details).
```
rfindler, line 39: detail, not details

pkgs/racket-doc/scribblings/guide/begin.scrbl
```diff
@@ -34,24 +34,16 @@ tail position with respect to the @racket[begin] form.
...
-     (newline)
-     (print-triangle (sub1 height))]))
-(print-triangle 4)
-]
+@specform[(begin def-or-expr ...)]{}
 
-The @racket[begin] form is special at the top level, at module level,
-or as a @racket[body] after only internal definitions. In those
-positions, instead of forming an expression, the content of
+In those positions, instead of forming an expression, the content of
```
rfindler, line 44: these, not those

pkgs/racket-doc/scribblings/guide/begin.scrbl
```diff
@@ -34,24 +34,16 @@ tail position with respect to the @racket[begin] form.
 (print-triangle 4)
 ]
 
-Many forms, such as @racket[lambda] or @racket[cond] support a
-sequence of expressions even without a @racket[begin]. Such positions are
-sometimes said to have an @deftech{implicit begin}.
+The @racket[begin] form is special at the top level, at module level,
+or as a @racket[_body] (e.g. of @racket[let] or @racket[lambda];
+see @tech{internal definition} for more details).
+Such @racket[begin] form has a slightly different grammar:
```
rfindler, line 40: Add "a", ie "Such a ..."

pkgs/racket-doc/scribblings/guide/begin.scrbl
```diff
@@ -64,6 +56,22 @@ positions, instead of forming an expression, the content of
 This splicing behavior is mainly useful for macros, as we discuss
 later in @secref["macros"].
 
+The whole @racket[_body] sequence is sometimes said to have an @deftech{implicit begin}.
```
rfindler, line 59: Why sometimes? Probably best to always or never (at least in the documentation we control).

pkgs/racket-doc/scribblings/guide/begin.scrbl
```diff
@@ -64,6 +56,22 @@ positions, instead of forming an expression, the content of
...
+The restriction of an @tech{implicit begin} follows the restriction of the @racket[_body] sequence:
+the sequence should end with an expression.
+
```
rfindler, line 62: An implicit begin is restricted in the same way as a @racket[_body]; it must end with an expression.

pkgs/racket-doc/scribblings/guide/define.scrbl
```diff
@@ -194,10 +194,10 @@ equivalent to a @racket[define-values] form with a single @racket[_id].
 
 When the grammar for a syntactic form specifies @racket[_body], then
 the corresponding form can be either a definition or an expression.
-A definition as a @racket[_body] is an @defterm{internal definition}.
+A definition as a @racket[_body] is an @deftech{internal definition}.
```
rfindler, line 197: "A definition in", I'd say.

pkgs/racket-doc/scribblings/reference/syntax.scrbl
```diff
@@ -2787,13 +2787,14 @@ For backward compatibility only; equivalent to @racket[syntax-local-introduce].
 @defform*[[(begin form ...)
            (begin expr ...+)]]{
 
-The first form applies when @racket[begin] appears at the top level,
-at module level, or in an internal-definition position (before any
-expression in the internal-definition sequence). In that case, the
+The first form applies when @racket[begin] appears at the @tech{top-level context},
```
rfindler, line 2790: Appear in contexts, not at contexts.

pkgs/racket-doc/scribblings/reference/syntax.scrbl
```diff
@@ -2787,13 +2787,14 @@ For backward compatibility only; equivalent to @racket[syntax-local-introduce].
...
+at @tech{module context}, or in an @tech{internal-definition context}. In that case, the
 @racket[begin] form is equivalent to splicing the @racket[form]s into
 the enclosing context.
+The @tech{internal-definition context} itself is said to have an
+@deftech{implicit begin} of this first form.
```
rfindler, line 2795: "of the first form" is correct, but how about "The internal-definition context itself is treated as a begin containing definitions" or something more like that?

rfindler (commented):
Thanks for this! I just noticed a few places where the words should be slightly different.

pkgs/racket-doc/scribblings/guide/begin.scrbl
mflatt: Also, I'm not sure it's clear that the metavariable `body` here refers to internal-definition contexts, so maybe `The whole @racket[_body] sequence of an internal-definition context`?

pkgs/racket-doc/scribblings/guide/define.scrbl
sorawee: Are you treating the metavariable `body` as the whole sequence? IIUC, elsewhere, the metavariable `body` is treated as a single form in the sequence (hence the usage of `@racket[body]s` in various places). So I think "a definition as a @racket[_body]" is correct already.

pkgs/racket-doc/scribblings/guide/begin.scrbl
sorawee: One complication is that the Racket Guide doesn't really define what internal-definition context is, so I've been avoiding it. But perhaps we should define it, perhaps in the section "Internal Definitions".

PR #4149, opened 2022-02-03
Allow raise-syntax-error to raise an exception inheriting exn:fail:syntax

sorawee:


pkgs/racket-doc/scribblings/reference/exns.scrbl
```diff
@@ -519,7 +520,18 @@ message. If not @racket[""], it should normally start with a newline
 and two spaces to add extra fields to the message (see
 @secref["err-msg-conventions"]).
 
-@history[#:changed "6.90.0.18" @elem{Added the @racket[message-suffix] optional argument.}]}
+If specified, the @racket[exn] argument should be a constructor or function that
+consumes a list of syntax, a string, and a continuation mark set;
+and it should produce a @racket[exn:fail:syntax?].
```
bennn, line 525: Looks good to me, but maybe these lines should link to `exn:fail:syntax` instead of giving all the details:

> .... or function that has the same signature as the `exn:fail:syntax` constructor.

PR #4150, opened 2022-02-05
doc: fix argument order

sorawee:
Fix a mistake made in #4149

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/exns.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/exns.scrbl
@@ -458,10 +458,15 @@ for error-message adjustments.
                              [sub-expr any/c #f]
                              [extra-sources (listof syntax?) null]
                              [message-suffix string? ""]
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/exns.scrbl
```diff
@@ -521,7 +526,7 @@ and two spaces to add extra fields to the message (see
 @secref["err-msg-conventions"]).
 
 If specified, the @racket[exn] argument should be a constructor or function that
-consumes a list of syntax, a string, and a continuation mark set;
+consumes a string, a continuation mark set, and a list of syntax;
 and it should produce a @racket[exn:fail:syntax?].
```
bennn, line 530: Looks good to me, but maybe these lines should link to `exn:fail:syntax` instead of giving all the details:

> .... or function that has the same signature as the `exn:fail:syntax` constructor.

sorawee: Thanks, done!

PR #4152, opened 2022-02-08
Add boxes to place-message

timjervis:
These changes are designed to add `box`es to `place-message`s, treating boxes in that context in the same way `vector`s are treated. That is, boxes are allowed, and mutable boxes are automatically converted to immutable boxes.

This follows a short conversation on discourse:

[https://racket.discourse.group/t/vectors-boxes-and-places/648](https://racket.discourse.group/t/vectors-boxes-and-places/648)

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/places.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/places.scrbl
@@ -352,10 +352,11 @@ messages:
 
  @item{@tech{paths} (for any platform);}
 
- @item{@tech{pairs}, @tech{lists}, @tech{vectors}, and immutable
+ @item{@tech{pairs}, @tech{lists}, @tech{box}es, @tech{vectors}, and immutable
        @tech{prefab} structures containing message-allowed values,
-       where a mutable vector is automatically replaced by an
-       immutable vector and where @tech{impersonators} of vectors and
+       where a mutable box is automatically replaced by an
+       immutable box, a mutable vector is automatically replaced by an
+       immutable vector and where @tech{impersonators} of boxes, vectors and
        @tech{prefab} structures are copied;}
 
  @item{@tech{hash tables} where mutable hash tables are automatically
@@ -376,9 +377,9 @@ messages:
  @item{values produced by @racket[shared-flvector],
        @racket[make-shared-flvector], @racket[shared-fxvector],
        @racket[make-shared-fxvector], @racket[shared-bytes], and
-       @racket[make-shared-bytes].}
-
-]}
+       @racket[make-shared-bytes].}]
+       
+@history[#:changed "8.4.0.7" @elem{include @racket[boxes].}]}
 
 @deftogether[(
 @defthing[prop:place-location struct-type-property?]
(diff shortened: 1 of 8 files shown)


samth:
That's a test for Racket BC, which doesn't use the `thread` implementation you changed. The relevant implementation of `place-message-allowed?` for Racket BC is here: https://github.com/racket/racket/blob/1b644903dad0bc7cab6d21d221a8400479a89f7d/racket/src/bc/src/place.c#L2611-L2628

racket/src/thread/place-message.rkt
```diff
@@ -245,24 +261,36 @@
...
                                              ph2
                                              (as ph3 (hasheq 'a 1 'b ph3))
                                              '#s(pre 4 5)))
                              ph)))
   (test #f (place-message-allowed-direct? stateful-cyclic))
   (test #t (place-message-allowed? stateful-cyclic))
-  (test stateful-cyclic (un-message-ize (message-ize stateful-cyclic)))
-
+  (test stateful-cyclic (un-message-ize (message-ize stateful-cyclic
+                                                     (lambda ()
```
samth, line 281: Why did you add the second argument here?

mflatt:
Changing BC is especially painful in this case, so let me know if hat becomes an obstacle.

racket/src/thread/place-message.rkt
timjervis: I assumed `message-ize` referred to the definition on line 94 which appears to require a second argument. Have I missed something?

samth: Ah, the problem is that this test code is not run by default and thus was totally broken before your change. Sigh.

timjervis: That's actually encouraging, to me at least - I figured the same.

timjervis:
> 

Thanks Matthew. I've pushed a version that appears to work on my system.

Apologies, I completely neglected BC. Also, as you can probably tell, this is my first pull request. Thanks for all your help!

pkgs/racket-doc/scribblings/reference/places.scrbl
```diff
@@ -376,9 +377,9 @@ messages:
  @item{values produced by @racket[shared-flvector],
        @racket[make-shared-flvector], @racket[shared-fxvector],
        @racket[make-shared-fxvector], @racket[shared-bytes], and
-       @racket[make-shared-bytes].}
+       @racket[make-shared-bytes].}]
 
-]}
+@history[#:changed "8.4.0.7" @elem{include @racket[boxes].}]}
```
mflatt, line 382: Capitalized "Include" to be consistent with other notes. Also, I think you meant `tech` instead of `racket` here, but I've make it plain "boxes".

mflatt (changes requested):
One small doc edit requested. Otherwise, this looks good, and I'll try a test build to double-check.

pkgs/racket-doc/scribblings/reference/places.scrbl
timjervis: Looks good - thanks!

mflatt (approved):

PR #4157, opened 2022-02-12
io: fix bad argument order to do-peek-string!

### LINKED ISSUE #4156: `peek-string!` yields invalid memory reference
8.4 [CS]

**What program did you run?**
```rkt
(define str (open-input-string "cat"))
(define buf (make-string 50 #\_))
(peek-string! buf 0 str)
```

No error.

```
  context...:
   body of "/home/camoy/tmp/peek.rkt"
```

I think this is likely a CS bug. I tried on an old pre-CS Racket and it worked alright. (I found this bug by converting TR types for the base library functions into contracts and using `contract-exercise`.)

Bogdanp:
Fixes #4156


mflatt:
Thanks!

PR #4158, opened 2022-02-12
Chez Scheme: add reductions for bitwise operations in cptypes

gus-massa:
### Isolate exact-integers in cptypes lattice

Add an additional field to `pred-or` to store the information about `exact-integer` separately from the other number types. This ensures that when a check `(fixnum? x)` is false, that information is not lost. In particular, in

    (unless (fixnum? x)
      (unless (fixnum? x)
        (something x)))

the second check will be eliminated.

My idea is to add in the future (1 month?) a more fine grained classification of `exact-integers`, so it also helps to keep them in a different part. 

Also, this includes fixes that enable some missed reduction opportunities.

### Add reductions for bitwise operations

 **In particular reduce**

    (bitwise-and <fixnum> <fixnum>)
    ==>
    (#3%fxand <fixnum> <fixnum>)

**Also**

    (bitwise-and (f) <fixnum>)
    ==>
    (let ([x (f)])
      (if (fixnum? x))
          (#3%fxand x <fixnum>)
          (bitwise-and x <fixnum>)))

This is a reduction in the wrong direction, because it makes the code longer instead of shorted, but `rumble` already has a similar trick with a macro. The difference is that the macro can do this with any number of arguments, but my version only does this with one argument to ensure the second pass of `cptypes` does not repeat it, because the types system in `cptypes` can't remember `(not (and (fixnum? x) (fixnum? y)))`.

**Know that the result of**

    (bitwise-and x <positive-fixnum>)
    (bitwise-ior x <negative-fixnum>)

is a `fixnum`. I added a check in optimize.rktl for `bitwise-ior` but this is a new feature that is not implemented in BC. I can add it to BC, but I think it's almost-frozen.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/optimize.rktl
+++ b/pkgs/racket-test-core/tests/racket/optimize.rktl
@@ -3028,7 +3028,7 @@
   (test-use-unsafe-fxbinary 'fxmin 'unsafe-fxmin)
   (test-use-unsafe-fxbinary 'fxmax 'unsafe-fxmax))
 
-(unless (eq? 'chez-scheme (system-type 'vm)) ; cptypes doesn't currently convert to fixnum ops
+(unless (eq? 'chez-scheme (system-type 'vm)) ; in these cases, cptypes converts zero? to fxzero?
   (test-comp '(lambda (vx)
                 (let ([x (string-length vx)])
                   (zero? x)))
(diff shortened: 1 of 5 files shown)


mflatt:
This sounds great. I agree about not updating BC.

gus-massa:
I forgot to ask. The new `let` form 

    (let ([temp (f)]) (something temp temp))

is implemented in Chez Scheme as

    ((lambda (temp) (something temp temp)) (f))

and the `lambda`  has some `preinfo`.

I'm using an empty `preinfo`. Is it possible/useful to add something interesting to it, for example that it is single valued?

PR #4162, opened 2022-02-17
Add `read-syntax-accept-graph` to allow graph structure in `read-syntax`

lexi-lambda:
Syntax objects made from plain S-expressions cannot contain cycles, so allowing graph structure in `read-syntax` mode is questionably useful. However, R7RS specifies that graph structure syntax should be permitted directly in program text, which is awkward to support in `#lang r7rs` without cooperation from the reader; see lexi-lambda/racket-r7rs#17.

As a compromise, this PR adds a new `read-syntax-accept-graph` parameter that enables graph structure parsing in `read-syntax` mode, but cycles are explicitly represented using `placeholder?` values in the result. Such syntax objects are not directly useful, but they ease the implementation of `#lang r7rs`, since it can arrange for the placeholders to be eliminated during macroexpansion.


racket/src/expander/Makefile
```diff
@@ -101,5 +101,5 @@ expander-maybe:
 	$(RACO) make bootstrap-run.rkt
 	$(MAKE) ../bc/src/startup.inc
 
-../bc/src/startup.inc: compiled/bootstrap-run_rkt.zo
+../bc/src/startup.inc: compiled/8.4.0.7/compiled/bootstrap-run_rkt.zo
```
samth, line 104: This looks wrong.

lexi-lambda:
@mflatt Just pinging you about this.

mflatt (approved):
Has it been a week already? Looks good.

PR #4171, opened 2022-02-27
Chez Scheme: add integer?/exact?/inexact? to cptypes

gus-massa:
[WIP, 90% done]

### Add `integer?` to cptypes

Similarly, improve the support for `exact?`/`inexact?` and other subsets of numbers. The idea is to use a bit field to store the information. I'm using a also bit field for `$immediates` and other for `numbers`. In particular, now `(and (integer? x) (exact? x))` is equivalent to `(or (fixnum? x) (bignum? x))`. 

The bit field for `$immediate` excludes `fixnums` and `chars`, and in the future I will add other objects that are singletons, like `""`, `immutable-""`, `0.0`, `-0,0` so I will rename the bit field as `singleton`.

The bit field for `numbers` excludes `exact-numbers` and includes `chars` and `symbols`. I want to rename it, but I don't know a good name for `not-singletons`. Any suggestion?

And there is another slot for `exact-numbers` because I'm planning to add more granularity in the near future.

And there is another slot for `$records` and a last slot for all the other stuff that probably need some special case.

In the classification of `numbers`, I added `flzero?` as one of the subsets, but after thinking about it, I don't expect it to be very useful because `fl` operation usually don't raise errors and `(fl/ <x> <not-zero>)`  can still produce `+/-ínf.0`, so I'll probably join `flzero?` and the other `flonum*s` in a single set. Perhaps someone of the typed-racket team has more insight about when tracking `flzero?` may be useful.

### Move info about predicates from cptypes to primdata

On one hand, this makes the code to handle `predicates` more general, so it doesn't assume predicates are total and can handle `exact?`, `inexact?`, ...  

On the other hand, it moves the mapping from functions to the name of the predicate from cptypes.ss to primdata.ss, so each predicate can be annotated there instead of using another separate list.

Also, this commit fixes the type of `ports`, because I thought they were `opaque records`, but they are not `records`.

### add tests of predicates to primvars

This is an idea to check predicates also in primvars.ms. I tried it, but I got an error, in spite the code is very similar to the code that test that the wrong arguments cause an error. If someone spot the error I'll fix it, otherwise I'll just drop this change.


gus-massa:
I renamed the internal slot for `immediate` to `singleton` because I expect to add more things to it later (two months?), for example `""` or `0.0` and other singletons (modulo `eqv?`) that are not immediates. 

I renamed the slot for not-singletons (like numbers, chars, symbols, ..) to `flock`. I couldn't find the standard technical name.  From https://www.google.com/search?q=singleton+antonyms I got a few bad options and I like `flock`, but I'd change it if there is a better name.

As in the initial post, the last commit just adds more automatic test using info in primvars.ss, but it has an error that I can't find. I'd be nice to merge it, but it's not necessary so my idea is to just drop it.

mflatt:
The word "flock" is ok with me, but for what it's worth, I keep reading it as "f(ile) lock" (https://man7.org/linux/man-pages/man2/flock.2.html).

samth:
That was also my initial reading.

For the test failure, does it also occur on the current HEAD?

gus-massa:
From the list of antonyms, I tried to avoid `group`, `set`, ... because they have technical meaning. So `flock` is a bad option too. What about `swarm` or `herd`?

I rebased my initial PR, so it's using the current version of master (or a very recent one). The problem is some strange level of quoting in primvars.ms near line 690

    (let ([call `(,name ',good.bad)])

If I remove the qoute, I get problems with the test with boxes like `&a` that are not self-quoting. If I keep the quote, instead of the function `values` I get the symbol `'values`. I tried to fix it before my first post, but after a while I gave up.

rfindler:
A murder (of crows) is a terrible idea. Don't do that.

Robby

;P

gus-massa:
I like `multiplet`. It's strange that I didn't think about this, because I use it a lot in quantum chemistry https://en.wikipedia.org/wiki/Multiplet

I fond now also `multiton` and `multipleton` but they are slightly different and I don't like how they sound https://en.wikipedia.org/wiki/Multiton_pattern

If `multiplet` is somehow confusing, I like `swarm`.

(As a not native speaker, I had no idea what `slew` means, so I had to google it.)

gus-massa:
I renamed `flock` to `multiplet` and I dropped the last commit that only adds some tests.

PR #4173, opened 2022-02-28
ffi: add _delay ctype form

rmculpepper:
This adds a `_delay` form for ctypes, similar to `recursive-contract` for contracts.

Before merging:
- The documentation should probably be moved to a different section, but I'm not sure which section would be best.
- Remove FIXME from merge commit (or better, merge by squashing).


mflatt (approved):
This looks fine to me. I don't have a better idea for where to put the documentation, although maybe a link from the `define-cstruct` form's documentation would be helpful.

PR #4180, opened 2022-03-10
Enable USE_ZLIB for emscripten ChezScheme

TheUnlocked:
Sets the default configuration for `--emscripten` builds of Chez Scheme to use `-s USE_ZLIB=1` instead of the normal zlib build process. Compiling with `--emscripten` should now just work out of the box.


mflatt:
Thanks! I'm reworking all the configure scripts and makefiles, but I'll make sure this effect gets propagated.

Issue #3672, opened 2021-01-31
`raco pkg install <git repo>` doesn't look for `main` branch


mbutterick:
`main` is supplanting `master` as the default branch of Git repos. [The current code](https://github.com/racket/racket/blob/46a191df038244199f83f2bb1235beb8ca137f9a/racket/collects/net/git-checkout.rkt#L385) only looks for `master`, and fails even if `main` exists.

mflatt:
I've changed `raco pkg` so that when a package reference does not specify a branch, it uses the default branch for a repo as reported by the server.

Unfortunately, this change will take a while (i.e., some number of release cycles) to percolate through the package infrastructure and the installations on Racketeers' machines. Using `#main` in the package reference will necessary for a while, yet.

pmatos:
@mbutterick / @mflatt can this be closed?

LiberalArtist:
Now that 8.1 and cff766ab84dcc5d9dc969930611e0b8c150bfaf6 are here, would it make sense for the package catalog server to add explicit branch names automatically? (Or perhaps just when talking to pre-8.1 clients?)

mflatt:
I'm reluctant to add that kind of auto-conversion to the package server. It seems safest (and certainly simpler) to leave this in the hands of package authors.

PR #4190, opened 2022-04-01
Use default branch for GitHub package sources

jryans:
Commit cff766ab84dcc5d9dc969930611e0b8c150bfaf6 changed **Git** package sources to use the default branch (instead of assuming `master` explicitly), but the more specialised **GitHub** code path was missed at the time. This fixes the GitHub path as well.

As with Git package sources at the time of the previous commit above, this is technically a backward-incompatible change to the interpretation of GitHub package sources, but explicit branch specification continues to work the same. For the foreseeable future, to support recent versions, packages in a branch other than `master` will still need to be specified using the branch name, such as including `#main` at the end of the package source. Eventually, relevant versions of Racket will support the new default.

Related to https://github.com/racket/racket/issues/3672
Related to https://github.com/racket/pkg-index/issues/32

### DIFF
--- a/pkgs/racket-test/tests/pkg/tests-network.rkt
+++ b/pkgs/racket-test/tests/pkg/tests-network.rkt
@@ -96,4 +96,8 @@
 
  (try-git-repo-using-default-branch
   "remote/git with default branch"
-  "https://github.com/racket/test-pkg-2.git"))
+  "https://github.com/racket/test-pkg-2.git")
+
+ (try-git-repo-using-default-branch
+  "remote/github with default branch"
+  "git://github.com/racket/test-pkg-2"))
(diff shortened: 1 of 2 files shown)


mflatt:
I think I originally had in mind not changing the way GitHub sources are treated, viewing them as a legacy form where a change to the repo should be combined with a change to the package registration. But I'm not sure that's the right idea, and the documentation certainly doesn't reflect that idea, so maybe it's better to go with this change.

jryans:
> I think I originally had in mind not changing the way GitHub sources are treated, viewing them as a legacy form where a change to the repo should be combined with a change to the package registration. But I'm not sure that's the right idea, and the documentation certainly doesn't reflect that idea, so maybe it's better to go with this change.

Ah okay, I wasn't entirely sure if it was just missed or skipped intentionally with the earlier change to Git package sources... If GitHub packages sources were defined as _only_ the `github://` type, I could understand a legacy treatment like that, but given it also includes `git://github.com` as well (which seems to still be quite widely used), it would be a bit harder to stomach, I think.

Anyway, thanks for the review and merge. 😄

PR #4179, opened 2022-03-09
replacing the makefiles and other build scripts

mflatt:
This is still very much work in progress, but it's far enough that I can now predict where the path leads.

The Problem
-----------

Running `make` currently triggers a Rube-Goldberg-style contraption: makefiles that run scripts that generate twisty sets of makefile macros that trick `make` into running compilers and other makefiles and... It's fragile, difficult update, and everything has to be done twice to cover Unix and Windows. See also [Non-recursive Make Considered Harmful: Build Systems at Scale](https://dl.acm.org/doi/10.1145/3241625.2976011); our makefiles don't currently include `$$$$`, but probably because I've found only worse ways to do things.

For me, the final straw with the current build system was 1de96effcc3327958276c6833900e51b41153e50, which was fallout from 0f43e3efaa651c16eff20c849a8c3f3c14f85eaa and then 803b5ff7d106be2919128a9497ca75b4663b1f9b. Although the commit message for 1de96effcc3327958276c6833900e51b41153e50 doesn't really explain, the all-day work there was figuring out that "../nanopass" needed to be "../nanopass::." in this copy of what "Mf-base" does. Meanwhile, the work of adapting Chez Scheme's implementation to Arm64 Windows was straightforward and maybe an hour's work, but updating the build system to add that option took all day — and the Racket part outside of Chez Scheme promises to be much worse. The Emscripten port 636a4669db802887fbfe87b7390eaf4b55d89e0d was a similar experience.

The Solution
------------

As in [Non-recursive Make Considered Harmful: Build Systems at Scale](https://dl.acm.org/doi/10.1145/3241625.2976011), we can build a better `make`. This commit mostly imitates Shake as described there, but in a Racket style and called "Zuo" (based on the Chinese word for "make"). The Zuo implementation started at https://github.com/mflatt/zuo, but here it's in the intended long-term home, `racket/src/zuo`.

The goal is for Racket to be be just as easy for end-users to build as now after converting scripts from `make` and `/bin/sh` to `zuo`. So, `make` should work in Racket's top-level directory, and `configure` plus `make` plus `make install` should work in a source directory or distribution, all with no new pre-installed packages required. The new makefiles will be small, however, just ensuring that the `zuo` executable is built and then handing off to Zuo scripts.

For that first step, building `zuo` can be just

```
 cc -o zuo zuo.c
```

on Unix or

```
 cl zuo.c
```

[...]

PR #4203, opened 2022-04-12
Guix-related build system enhancements

LiberalArtist:
Distributions like Guix may delete the "zlib" directory
(e.g. because they intend to configure with `ZLIB=-lz`).

See discussion at https://github.com/racket/racket/pull/4179#issuecomment-1095825003.


mflatt:
I didn't notice this PR until after pushing a different repair. This repair is more complete in the way that it reports an error if `ZLIB=...` is needed, though, and similar checking would be useful for `lz4`, so I plan to look more at this.

racket/src/ChezScheme/build.zuo
```diff
@@ -139,18 +139,28 @@
...
+    (define (do-doc name install-dir-key)
       (mkdir-p (at-dir name))
       (define mf (~a "m=" m "\n"
                      "srcdir=" (find-relative-path (at-dir name) (at-source name)) "\n"
+                     ;; FIXME: fales if configured path includes "#"
+                     "installdir=" (hash-ref config install-dir-key) "\n"
                      (file->string (at-source (~a "makefiles/Makefile-" name ".in")))))
       (display-to-file mf (at-dir name "Makefile") :truncate)
       (shell/wait "make" target
                   (hash 'dir (at-dir name))))
```
LiberalArtist, line 160: @mflatt I've pushed a rebased branch with the start of the Chez configure changes from https://github.com/racket/racket/pull/4179#issuecomment-1104231034. I've dropped https://github.com/LiberalArtist/racket/commit/483239d9b5c16506b23b1ae0d62dd21841062d1a, at least for now, since it no longer directly applies, but https://github.com/LiberalArtist/racket/releases/tag/pr-4203-v0 should preserve it from garbage collection.

I've used names consistent with the existing Chez `configure` flags, rather than the Autoconf conventions, which would have used e.g. `--docdir=` rather than `--installdoc=`.

There are at few things I'm wondering about:

 0. Is there a mechanism to add quoting around a path destined to be written into a makefile? I tried configuring with a path like `/foo/#bar/baz`, but `make` interpreted the `#` as beginning a comment, so the files were installed to `/foo/`. (I sometimes use paths like this because they are covered by the `.gitignore` file.)
 0. I think there probably should be a `.gitignore` rule for `main.zuo` files generated by `configure`, but I don't have a great sense of how to best to structure one or where to put it. I could imagine putting a directory-specific `.gitignore` file in each directory where a `main.zuo` could be generated or adding the full paths to the top-level `.gitignore`, or maybe it would work to add a general `main.zuo` rule at the top level and put a comment in the placeholder `racket/src/zuo/main.zuo` that it has to be force-added.
 0. I noticed that Zuo is invoked with `	$(ZUO) $(workarea) install-docs MAKE="$(MAKE)"`, but the sub-build is invoked using `"make"`.
 0. I haven't tried building with Guix yet, but I'm suspecting that it will be a problem that this doesn't seem to pass general make flags on to the document `make` invokation: we pass `STEXLIB=`, because we've removed the bundled `stex` directory and replaced it with one we have elsewhere. I could support `STEXLIB` specially via `configure`, but is there a mechanism for propagating flags to `make` more generally?

mflatt: Thanks for these improvements!

> Is there a mechanism to add quoting around a path destined to be written into a makefile?

You can use shell quoting, but escaped by one level if you're calling `configure` in a shell. So, something like `--installdoc="/foo/\"#\"bar/baz"`. Zuo scripts specifically parse information from `Mf-config`/`Makefile` as shell sequences (or, at least, that's how it's meant to work).

> I think there probably should be a .gitignore rule for main.zuo files generated by configure

I have so far resisted trying to exclude more build files. My model in recent times is that you either use a build directory or you're opting into a context where you don't care about mingling build and source files. But probably I can be convinced.

> I noticed that Zuo is invoked with $(ZUO) $(workarea) install-docs MAKE="$(MAKE)", but the sub-build is invoked using "make".

I agree that the Zuo scripts need to be fixed to not use "make".

> I haven't tried building with Guix yet, but I'm suspecting that it will be a problem that this doesn't seem to pass general make flags on to the document make invokation

Based on earlier experiments, I think GNU Make will set environment variables so that `make` arguments get passed along, mostly. And this might be a place where it makes sense to assume/require GNU make.

LiberalArtist: > > I haven't tried building with Guix yet, but I'm suspecting that it will be a problem that this doesn't seem to pass general make flags on to the document make invokation
> 
> Based on earlier experiments, I think GNU Make will set environment variables so that `make` arguments get passed along, mostly. And this might be a place where it makes sense to assume/require GNU make.

Delightfully, this indeed worked! At least for Guix, assuming GNU make is fine.

LiberalArtist:
I've hit an error in the build that's essentially identical to the one in https://github.com/racket/racket/pull/4179#issuecomment-1104231034 (the full new log is at  
[zp4xc6d97hd4alyjprlihmqja9ym6m-chez-scheme-for-racket-9.5.7.7.drv.gz](https://github.com/racket/racket/files/8526788/zp4xc6d97hd4alyjprlihmqja9ym6m-chez-scheme-for-racket-9.5.7.7.drv.gz)). Looking at this excerpt:
```
Output written on csug.pdf (558 pages, 1710398 bytes).
Transcript written on csug.log.
touch csug.thirdrun
sort libsrecorded | uniq > libsrecorded.sort
sort libslisted | uniq > libslisted.sort
diff libsrecorded.sort libslisted.sort
(cd /gnu/store/a0l6pc7xxwhl026i5lmrdvd167xyla8d-stex-1.2.2-1.5405149//lib/stex; tar -cf - gifs) | tar -xpf -
(cd /gnu/store/a0l6pc7xxwhl026i5lmrdvd167xyla8d-stex-1.2.2-1.5405149//lib/stex; tar -cf - math) | tar -xpf -
sed -e 's/(putprop (quote /(putprop (quote |TSPL|:/' ../../csug/tspl4/tspl.haux | \
         sed -e 's;url) ";url) "http://scheme.com/tspl4/;' > tspl.haux
sed -e 's;"\(.*\)\.html#;"http://scheme.com/tspl4/\1.html#;' ../../csug/tspl4/out.hidx | \
         sed -e 's/"")$/"t")/' > in.hidx
cat tspl.aux >> csug.aux
cat tspl.rfm >> csug.rfm
cat tspl.idx >> csug.idx
cat tspl.haux > csug.haux
touch csug.hprefirstrun
(if [ ! -e math/csug ] ; then mkdir -p -m u=rwx,g=srx,o=rx math/csug; ln -s ../Makefile ../mathmacros math/csug; fi)
mkdir: cannot create directory ‘math/csug’: Permission denied
ln: target 'math/csug': No such file or directory
make[1]: *** [/gnu/store/a0l6pc7xxwhl026i5lmrdvd167xyla8d-stex-1.2.2-1.5405149//lib/stex/Mf-stex:131: csug.hfirstrun] Error 1
make[1]: Leaving directory '/tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/ta6le/csug'
```
The problem seems to be that this command:
```
(cd /gnu/store/a0l6pc7xxwhl026i5lmrdvd167xyla8d-stex-1.2.2-1.5405149//lib/stex; tar -cf - math) | tar -xpf -
```
creates the `math` directory with the permissions of the original—which is read-only on Guix, so attempting to create `math/csug` fails.

The problem didn't show up before because Guix was building in the source `csug` directory, rather than in `ta6le/csug/`, so the `math` directory already existed.

I think `stex` is to blame for this problem: I'll try patching Guix's copy.

racket/src/ChezScheme/build.zuo
LiberalArtist: > > Is there a mechanism to add quoting around a path destined to be written into a makefile?
> 
> You can use shell quoting, but escaped by one level if you're calling `configure` in a shell. So, something like `--installdoc="/foo/\"#\"bar/baz"`. Zuo scripts specifically parse information from `Mf-config`/`Makefile` as shell sequences (or, at least, that's how it's meant to work).
> 

Actually, that makes installation fail altogether.

Here are two scenarios:

 1. With `$ ../configure "--installprefix=/tmp/chez-zuo/#sub"`

    in `ta6le/Mf-config`:
    ```
    InstallBin=/tmp/chez-zuo/#sub/bin
    InstallLib=/tmp/chez-zuo/#sub/lib
    InstallMan=/tmp/chez-zuo/#sub/share/man/man1
    InstallCSUG=/tmp/chez-zuo/#sub/share/doc/csug9.5
    InstallReleaseNotes=/tmp/chez-zuo/#sub/share/doc/csv9
    ```
    in `ta6le/csug/Makefile`:
    ```
    m=ta6le
    srcdir=../../../csug
    installdir=/tmp/chez-zuo/#sub/share/doc/csug9.5
    # srcdir, installdir, and m are added before
    ```
    Running `make install-doc` installs to `/tmp/chez-zuo/`, without the subdirectories.
 3. With `../configure "--installprefix=/tmp/chez-zuo/\"#\"sub"`

    in `ta6le/Mf-config`:
    ```
    InstallBin=/tmp/chez-zuo/"#"sub/bin
    InstallLib=/tmp/chez-zuo/"#"sub/lib
    InstallMan=/tmp/chez-zuo/"#"sub/share/man/man1
    InstallCSUG=/tmp/chez-zuo/"#"sub/share/doc/csug9.5
    InstallReleaseNotes=/tmp/chez-zuo/"#"sub/share/doc/csv9
    ```
    in `ta6le/csug/Makefile`:
    ```
    m=ta6le
    srcdir=../../../csug
    installdir=/tmp/chez-zuo/"#"sub/share/doc/csug9.5
    # srcdir, installdir, and m are added before
    ```
    Running `make install-doc` 
    ```
    make[2]: Leaving directory '/home/philip/code/tmp/racket/racket/src/ChezScheme/tmp/ta6le/csug/math/csug'
    touch csug.mathrun
    chmod 644 csug.pdf
    ../../../csug/../makefiles/installsh -m 2755 -d /tmp/chez-zuo/"
    bash: -c: line 1: unexpected EOF while looking for matching `"'
    bash: -c: line 2: syntax error: unexpected end of file
    make[1]: *** [Makefile:21: install] Error 2
    make[1]: Leaving directory '/home/philip/code/tmp/racket/racket/src/ChezScheme/tmp/ta6le/csug'
    shell command failed
     in map
     in build-one
     in loop
     in module->hash
    make: *** [Makefile:85: install-docs] Error 1
    ```

LiberalArtist: On the special characters issue, [the Autoconf manual](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/Special-Chars-in-Variables.html) lists several restrictions on the values of output variables, including a ban on `#`, and says that the restrictions "apply both to the values that `configure` computes, and to the values set directly by the user."

mflatt: I didn't think through the fact that `make` parsing would matter in this case, but it also would not have occurred to me that `make` doesn't follow `sh` quoting rules. Remarkably enough, every Unix `make` variant I've tried works the same way: `#` anywhere ends the line unless it's preceded by `\`, and `\#` escapes the `#`. Naturally, `nmake` on Windows is different.

Probably Zuo's `config-file->hash` should follow the Unix `make` rule for `#` and `\#`, and the small `nmake` mismatch seems unlikely to create new issues. Then, it would be a matter of writing a `configure` invocation that uses a shell encoding of a makefile encoding of a shell encoding, which is possibly `../configure "--installprefix=/tmp/chez-zuo/\"\\#\"sub"`.

LiberalArtist: That makes sense to me, but would the Zuo code writing to makefiles then also need to re-encode `#`? Alternatively, I was surprised by it (given their high tolerance for painfully weird configurations), but I personally could live with [Autoconf's rule](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/Special-Chars-in-Variables.html) against using such characters in arguments to `configure`. Dealing with arcane quoting rules is easily my least-favorite part of traditional build scripting.

LiberalArtist:
I've pushed several commits (to be cleaned up and squashed eventually) that make Zuo's `configure` more compatible with Autoconf-style options, particularly the ones I mentioned in https://github.com/racket/racket/pull/4179#issuecomment-1095684253, mostly by ignoring a whole bunch of inapplicable ones.

It had not occurred to me until just now that `racket` actually uses Autoconf, and thus you must have had a strategy for dealing with the fact that it writes in `build/cs/c/Mf-config` e.g. `libdir = ${exec_prefix}/lib`, expecting the user to be able to override `exec_prefix` or `prefix` at `make` time to change it, if `libdir` wasn't set explicitly at `configure` time, which led me to discover `shell-subst` in `racket/src/lib.zuo`. If that could be promoted perhaps to `zuo/shell`, Zuo could probably use a very concise `configure.ac`.

Another issue is that running e.g. `cofigure CFLAGS=-g` currently overrides the default `CFLAGS=-O2`, rather than adding to them, which I think is the convention.

Currently Zuo's `configure` simply ignores `--build=`, `--host=`, `--target=`, `CC_FOR_BUILD=`, `CONFIG_SHELL=`, and `SHELL=`. At least the first set seem like they might be worth handling in `zuo/c`. I could also imagine the latter set being used, perhaps in `zuo/shell`, but it would probably require thinking through the fact that (at least GNU) make has [special treatment for `SHELL`](https://www.gnu.org/software/make/manual/html_node/Choosing-the-Shell.html#Choosing-the-Shell) and what the interaction ought to be with [`CONFIG_SHELL`](https://www.gnu.org/savannah-checkouts/gnu/autoconf/manual/autoconf-2.71/html_node/config_002estatus-Invocation.html#index-CONFIG_005fSHELL-1).

racket/src/ChezScheme/build.zuo
mflatt: Yes, and this is a reason to avoid writing makefiles from Zuo scripts. For example, the Racket CS Zuo scripts do not try to write a Chez Scheme configuration "makefiles", and instead they pass variables to other Zuo scripts in Chez Scheme. There is some makefile writing in the Chez Scheme Zuo scripts, but only a little, and probably that could be improved. Anyway, I can sympathize with the autoconf "don't do that" approach.

LiberalArtist: That makes sense, and I think I can avoid doing so here, too.

Tangentially, one format that avoids this is [`environment.d` configuration files](https://www.freedesktop.org/software/systemd/man/environment.d.html#Configuration%20Format) as used by `systemd`. They don't support any kind of quoting at all, and the don't support newlines in variable values, but they do support `$x`, `${x}`, `${a:-b}`, `${a:+b}`, line comments, and blank lines.

mflatt:
Using `autoconf` for Zuo would make sense. As you can probably tell, I was trying to avoid `autoconf` for Zuo, but maybe avoiding it doesn't actually make things any simpler.

I'm pretty sure that providing `CFLAGS` to an `autoconf`-generated `configure` will set `CFLAGS`, not augment. A `configure` script might reasonably augment `CFLAGS` with flags like `-D`, but I don't think it should add flags like `-O2` or `-g` or `-Wall` (except through a flag like `--enable-warning-flags`).

LiberalArtist:
I can also understand wanting to avoid Autoconf if possible!

From a quick experiment, it looks like putting this in `configure.ac`:
```autoconf
AC_INIT([Zuo],
        [1],
        [https://github.com/racket/racket/issues],
        [],
        [https://github.com/racket/zuo])
AC_CONFIG_SRCDIR(zuo.c)
AC_PROG_MAKE_SET()
AC_CONFIG_FILES([Makefile])
AC_OUTPUT()
```
And this at the top of `Makefile.in`:
```
# @PACKAGE_NAME@ @configure_input@
@SET_MAKE@

srcdir = @srcdir@

prefix=@prefix@
exec_prefix=@exec_prefix@
datarootdir=@datarootdir@
datadir=@datadir@

bindir=@bindir@
pkgdatadir=@datadir@/@PACKAGE_TARNAME@

CC = @CC@
CFLAGS = @CFLAGS@
CPPFLAGS = @CPPFLAGS@
LDFLAGS = @LDFLAGS@
LIBS = @LIBS@
```
is enough to replicate the current script (without any auxiliary files) other than `--embed="<lib> ..."` and `--big`, which presumably would need to be converted to `--enable-` style.

mflatt:
Thanks — starting with that, I converted Zuo's `configure`.

LiberalArtist:
Great! I've rebased.

I'd started looking at this two (but hadn't made any of the changes on the Zuo side), and the first commit in the new series has a couple of minor adjustments based on what I'd tried: mostly fixing `--disable-embed`, but I also found that generating `main.zuo` via `AC_CONFIG_FILES` seemed to cooperate better with things like `config.status`. Most of the diff is because I didn't have the same version of Autoconf readily at hand, but I'm sure I could set it up if you prefer.

mflatt:
It looks like the first three commits at this point are probably ready to merge ("configure.ac" corrections, `check` target, and Chez Scheme `configure --help` repairs). Is `configure --installdoc=` also ready, or are you still working there?

LiberalArtist:
> It looks like the first three commits at this point are probably ready to merge ("configure.ac" corrections, `check` target, and Chez Scheme `configure --help` repairs).

Yes, I think so

> Is `configure --installdoc=` also ready, or are you still working there?

I'm still adjusting that: I think I can avoid having to write to the makefiles from Zuo there.

LiberalArtist:
I've pushed a working version of the Chez `configure --installdoc=` change that I think is ready to merge. With this (plus a two-line patch to Stex I'll shortly send upstream), I've managed to build all of `main-distribution` using Guix—provided I skip the Chez Scheme test suite, which had been having problems. I'll investigate that next.

Some miscellaneous things I haven't done:

- I haven't tried to change the Zuo scripts to use `MAKE` instead of `"make"`.
- I haven't tried to change other cases where `ChezScheme/build.zuo` and other scripts write to makefiles. 
- While Zuo now knows that documentation needs to be rebuild if Stex files change, Stex's own makefiles don't always track the dependencies closely enough. (On the other hand, since I seem to be 1 of 4 people to have contributed to Stex since 2016, that's probably a fairly low priority.)
- I haven't touched Racket's Autoconf setup, but I imagine there should be a way to have it configure Zuo for the `build` system (in the `build`/`host`/`target` sense) even when cross-compiling Racket. (But I know approximately as little about Autotools as possible.) For Guix, we'd still want to bring our own Zuo anyway, though.
- The only way to use an installed Stex up to this point has been running `make STEXLIB=` with GNU Make or some other `make` that will propagate it as needed. I have only adjusted the Zuo script to recognize an `STEXLIB` passed that way: I haven't tried to support setting `STEXLIB` via `configure` or other mechanisms. There would be some uniformity in doing so, but it wouldn't be useful for Guix packaging unless it were also supported upstream, since we try to share as much as possible between the package definitions for the two Chez Schemes.

LiberalArtist:
Here's a log from the failing Chez test suite: [x0kbkfv1rp27xma3b3fy3sbxidfxrz-chez-scheme-for-racket-9.5.7.7.drv.gz](https://github.com/racket/racket/files/8549000/x0kbkfv1rp27xma3b3fy3sbxidfxrz-chez-scheme-for-racket-9.5.7.7.drv.gz)

Excerpts like these:
```
< 7.mo:Bug in mat verify-loadability clause 92 at line 5099, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/mats/7.ms
< 7.mo:Error in mat verify-loadability clause 93: "strip-fasl-file: failed for testfile-clK0.so: no such file or directory" at line 5102, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/mats/7.ms
```

```
< 8.mo:Error in mat eval-when-library clause 7: "load-library: file "testfile-ewl1.so" not found in source directories" at line 9761, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/mats/8.ms
< 8.mo:Error in mat eval-when-library clause 13: "loading testfile-ewl3.ss did not define library (testfile-ewl3)" at line 9778, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/mats/8.ms
< 8.mo:Error in mat eval-when-library clause 14: "loading testfile-ewl4.ss did not define library (testfile-ewl4)" at line 9779, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/mats/8.ms
< 8.mo:Error in mat eval-when-library clause 15: "loading testfile-ewl5.ss did not define library (testfile-ewl5)" at line 9780, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/mats/8.ms
```

```
< 8.mo:Error in mat top-level-program clause 23: "load-program: file "testfile-tlp5.so" not found in source directories" at line 10837, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/mats/8.ms
< 8.mo:Error in mat library-meta clause 5: "load: file "testfile-lm-b2.so" not found in source directories" at line 11020, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/mats/8.ms
```
make me think that files may not be getting moved to the right places, or something like that.

mflatt:
I don't have any good ideas about what might be happening, but I would be interested to see a couple of `make.out` files after the test run. Look for `make.out` in the log, but I think one file should be `/tmp/guix-build-chez-scheme-for-racket-9.5.7.7.drv-0/source/racket/src/ChezScheme/ta6le/mats/output-compile-0-f-f-f-simple/make.out`.

LiberalArtist:
Here is the whole `ChezScheme` directory from the build (which just barely compresses small enough to attach here): [ChezScheme.tar.gz](https://github.com/racket/racket/files/8575518/ChezScheme.tar.gz)

mflatt:
Thanks for the build-directory tarball. I wasn't able to figure out anything, unfortunately. The failures look like they could be interference among parallel runs writing to the same directory, but `mat-file` from `mat.ss` is given a different directory for each run. Maybe it's something about the `dir` argument there, or how `source-directories` and `library-directories` are updated?

LiberalArtist:
Thanks for looking! I will try adjusting the Chez package to use an out-of-source build and see if it clarifies anything. I can also try turning off parallel tests, though that would presumably be very slow.

PR #4214, opened 2022-04-22
Fix the build.md link in CONTRIBUTING.md

robertpostill:
The link was ever so slightly off and returned a 404.


robertpostill:
I might have found another way to raise this if I'd realised it would cause a 38 minute build 😳

mflatt:
Thanks for the repair!

I see that GitHub makes `/build.md` work. I would have written `../build.md`, since that relative path makes sense at the source level, but is `/build.md` better?

robertpostill:
The way I saw it picking a path that was descriptive of the build.md location in the repository means it matters less where CONTRIBUTING.md lives which made me think it was the more flexible approach.  

Happy to go either way 😄

mflatt:
Ok - thanks!

PR #4220, opened 2022-04-29
Chez Scheme: Add x->immutable-x to cp0 and cptypes

gus-massa:
Fold expressions like `(string->immutable-string "")` and when the argument is immutable.

Also, propagate the immutable versions of `""`, `#()` and `#vu8()`.


gus-massa:
About `cpprim`, I looked now and saw a very similar optimization. IIRC it's not sure that all the optimization in `cp0` will be applied, so it's better to keep the version in `cpprim`, I think it's better to have this in `cp0`, but it will be used very seldom, so if this may cause problems that part may be skipped. I think the schemified racket code already has immutable strings. 

Is this a problem when `rktboot` is interpreting Chez Scheme code? When it writes the boot files? When it reads the boot files? (Can `rktboot` read boot files?)

What happens with code like this:?

    #lang chez-scheme
    (let ([x ""]
          [y (string->immutable-string ""])
      (set! x x)
      (set! y y)
      (eq? x y))

mflatt:
While `rktboot` runs the Chez Scheme compiler to compile Chez Scheme, it makes `immutable-string?`, `immutable-vector?` and `immutable-bytevector?` always return `#f`. The old justification was that immutable objects are never used during compilation of Chez Scheme itself.

But that's not true after this patch, and it seems like it will be a problem: cp0 running during `rktboot` would recognize the `(string->immutable-string "")` in the implementation of `string->immutable-string`, then simplify to a constant that `immutable-string?` still says `#f` to, which means that `fasl` during `rktboot` would not recognize an immutable string, so the generated boot file would have the wrong constant within the implementation of `string->immutable-string`.

It turns out that the approximation in `rktboot` also means that the `eq?` test against a literal `""` (intended as a mutable string) fails, because that literal `""` turns out to be represented as an immutable string. So, the cp0 conversion does not fire, and there's no bad constant in the boot file. This is subtle, and it does not seem like a good idea to rely on that.

Probably the right answer is for `rktboot` to manage immutability more correctly. I'll have to look into that more.

gus-massa:
Is this problem fixed if I remove the `(define-inline 2 string->immutable-string ...)` and similar special cases in `cp0`? I don't expect them to be used very often, so they can wait to be merged, and it's better to err on the safe side.

I'd like to merge the rest of the PR, or there are still problems? Is `cpprim` used to make the boot files or only to run the code? Perhaps I can use a workarround. For example `cpprim` checks that the length of the string is `0` instead of comparing with `""` and `immutable-""`

mflatt:
Yes, the current commit without the cp0 changes should be fine.

I haven't found an easy fix for `rktboot`. If the cp0 folding of `X->immutable-X` doesn't seem so useful, then it would be nice to skip it and not have to find a general solution for `rktboot`.

PR #4221, opened 2022-05-02
Add #:groups, #:group-sep and #:decimal-sep keywords to ~r

Metaxal:
It can be difficult to read very large integers, and I often find myself grouping digits with my finger or mouse pointer on the screen. I figured it would be smarter to tackle the problem at the core. Digits can be grouped not only 3 by 3, but also more generally by groups of varying sizes (for example the Indian system groups the first 2 digits, then groups digits 3 by 3).

Additionally, this PR also introduces the `#:decimal-sep` option to `~r`, because not all languages or conventions use `.` as a decimal separator and it makes sense to allow for changing the decimal separator if the group separator can be changed too.

This PR does _not_ group digits after the decimal separator, because i) default floats have only small precision that don't require much grouping (by contrast to integers) ii) because I don't really have a use case for it myself and iii) because it took me long enough already to getting around to finish this small PR.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/format.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/format.scrbl
@@ -231,7 +231,10 @@ marker is @racket["..."].
                 (or/c #f string? (-> exact-integer? string?))
                 #f]
                [#:min-width min-width exact-positive-integer? 1]
-               [#:pad-string pad-string non-empty-string? " "])
+               [#:pad-string pad-string non-empty-string? " "]
+               [#:groups groups (listof exact-positive-integer?) '()]
+               [#:group-sep group-sep string? ""]
+               [#:decimal-sep decimal-sep string? "."])
          string?]{
 
 Converts the rational number @racket[x] to a string in either
@@ -281,6 +284,26 @@ decimal point are used, and the decimal point is never dropped.
 (~r 50 #:precision 2)
 (~r 50 #:precision '(= 2))
 (~r 50 #:precision '(= 0))
+]}
+
+@item{@racket[decimal-sep] specifies what decimal separator is printed.
+
+@examples[#:eval the-eval
+(~r 123.456)
+(~r 123.456 #:decimal-sep ",")
+]}
+
+@item{@racket[groups] controls how digits of the integral part of the number 
+   are separated into groups.
+   Rightmost numbers of @racket[groups] are used to group rightmost digits of the integral part.
+   The leftmost number of @racket[groups] is used repeatedly to group leftmost digits.
+   @racket[group-sep] specifies which separator to use between digit groups.
+       
+
+   @examples[#:eval the-eval
+(~r 1234567890 #:groups '(3) #:group-sep ",")
+(~r 1234567890 #:groups '(3 2) #:group-sep ",")
+(~r 1234567890 #:groups '(1 3 2) #:group-sep "_")
 ]}
 
 @item{@racket[min-width] --- if @racket[x] would normally be printed
@@ -402,6 +425,8 @@ the resulting string is appended to the significand:
 ]}
 
 ]
+@history[#:changed "8.5.0.3"
+         @elem{Added @racket[#:groups], @racket[#:groups-sep] and @racket[#:decimal-sep].}]
 }
 
 @; ----------------------------------------
--- a/pkgs/racket-doc/scribblings/reference/format.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/format.scrbl
@@ -232,7 +232,7 @@ marker is @racket["..."].
                 #f]
                [#:min-width min-width exact-positive-integer? 1]
                [#:pad-string pad-string non-empty-string? " "]
-               [#:groups groups (listof exact-positive-integer?) '()]
+               [#:groups groups (non-empty-listof exact-positive-integer?) '(3)]
                [#:group-sep group-sep string? ""]
                [#:decimal-sep decimal-sep string? "."])
          string?]{
@@ -426,7 +426,7 @@ the resulting string is appended to the significand:
 
 ]
 @history[#:changed "8.5.0.3"
-         @elem{Added @racket[#:groups], @racket[#:groups-sep] and @racket[#:decimal-sep].}]
+         @elem{Added @racket[#:groups], @racket[#:group-sep] and @racket[#:decimal-sep].}]
 }
 
 @; ----------------------------------------
--- a/pkgs/racket-doc/scribblings/reference/format.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/format.scrbl
@@ -297,7 +297,7 @@ decimal point are used, and the decimal point is never dropped.
    are separated into groups.
    Rightmost numbers of @racket[groups] are used to group rightmost digits of the integral part.
    The leftmost number of @racket[groups] is used repeatedly to group leftmost digits.
-   @racket[group-sep] specifies which separator to use between digit groups.
+   The @racket[group-sep] argument specifies which separator to use between digit groups.
        
 
    @examples[#:eval the-eval
@@ -425,7 +425,7 @@ the resulting string is appended to the significand:
 ]}
 
 ]
-@history[#:changed "8.5.0.3"
+@history[#:changed "8.5.0.5"
          @elem{Added @racket[#:groups], @racket[#:group-sep] and @racket[#:decimal-sep].}]
 }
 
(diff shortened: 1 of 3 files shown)


Metaxal:
cc @rmculpepper

racket/collects/racket/format.rkt
```diff
@@ -428,14 +462,34 @@
...
+                                   (string->list group-sep))))]
+        [else
+         (let-values ([(q r) (quotient/remainder N base)])
+           (loop q (cons (get-digit r upper?) digits)))]))))
+
 (define (number->fraction-string N base upper? precision)
   (let ([s (number->string* N base upper?)])
     (string-append (make-string (max 0 (- precision (string-length s))) #\0) s)))
 
+(define A-10 (- (char->integer #\A) 10))
```
sorawee, line 485: Any reason why this part is changed? I find the old code less confusing -- `A-10` doesn't have a descriptive meaning by itself.

sorawee:
Hmm. I just saw this comment:

> ;; MAYBE TO DO:
> ...
> ;;  - decimal separators (see "man 7 locale", lconv numeric fields)

If this can be supported without too much effort, I think that might be a better way to go.

racket/collects/racket/format.rkt
Metaxal: This is just a little classic optimization. Printing numbers can be a little time sensitive, and the old code was (quite likely) doing some unnecessary work.

Metaxal:
> Hmm. I just saw this comment:
> 
> > ;; MAYBE TO DO:
> > ...
> > ;;  - decimal separators (see "man 7 locale", lconv numeric fields)
> 
> If this can be supported without too much effort, I think that might be a better way to go.

I know it would take me significant effort compared to the current PR at least, if that's what you ask. Also, `"."` was hard coded in a couple of places in the previous code, so I think it's unlikely to be a smooth change, in particular since the code doesn't call C `printf` and friends.

sorawee:
I benchmarked the code, because I wanted to see the effect of the "little classic optimization" for fun :). Turns out "the optimization" has no effect. But the change in this PR does significantly slow the function down ~6x.

Method: 
1) Copy https://github.com/racket/racket/blob/b6b3d168b2f8c0bf0df527e03fc06151f90a5219/racket/collects/racket/format.rkt to a new file
2) Change `get-digit` to:

```
(define (get-digit d upper?)
  #;(cond [(< d 10) (integer->char (+ d char0))]
        [else (integer->char (+ d (if upper? A-10 a-10)))])

  (cond [(< d 10) (integer->char (+ d (char->integer #\0)))]
        [else (integer->char (+ (- d 10) (char->integer (if upper? #\A #\a))))]))
```

3) Add the following at the end

```
(require (prefix-in racket: racket/format))

(collect-garbage)
(collect-garbage)
(collect-garbage)
(time (for ([i 100000])
        (~r (expt 2 1000))))

(collect-garbage)
(collect-garbage)
(collect-garbage)
(time (for ([i 100000])
        (racket:~r (expt 2 1000))))
```

4) Run it. Switch the commented expression in `get-digit` and run it again.

Results:

Two runs with the optimization

```
cpu time: 2537 real time: 2606 gc time: 6
cpu time: 434 real time: 451 gc time: 1

cpu time: 2550 real time: 2620 gc time: 7
cpu time: 434 real time: 450 gc time: 1
```

Two runs without the optimization

```
cpu time: 2550 real time: 2617 gc time: 7
cpu time: 430 real time: 446 gc time: 1

cpu time: 2546 real time: 2615 gc time: 6
cpu time: 432 real time: 448 gc time: 1
```

As mentioned, there's no significant difference between the optimized or unoptimized version, but there's a significant difference in the version before the PR and the version after the PR.

This is of course expected since the PR adds new functionality. But it would be better if there is a "fast path" for the default configurations, to avoid the 6x slowdown in the common cases.

sorawee:
Here's a drop-in replacement for `number->string*/whole-positional` which should improve its performance. Note: I haven't thoroughly tested it.

```
;; number->pre-string :: number? number? boolean? -> (or/c string? (listof char?))
;; returns either string or list to avoid potential round trip conversion
(define (number->pre-string N base upper?)
  (cond [(memv base '(2 8 10 16))
         (let ([s (number->string N base)])
           (if (and (= base 16) upper?)
               (string-upcase s)
               s))]
        [(zero? N) "0"]
        [else
         (let loop ([N N] [digits null])
           (cond [(zero? N) digits]
                 [else (let-values ([(q r) (quotient/remainder N base)])
                         (loop q (cons (get-digit r upper?) digits)))]))]))

(define (number->string* N base upper?)
  (define pre-string (number->pre-string N base upper?))
  (cond
    [(string? pre-string) pre-string]
    [else (apply string pre-string)]))

(define (number->string*/whole-positional N base upper? groups group-sep)
  (cond
    [(zero? N) "0"]
    [(equal? group-sep "") (number->string* N base upper?)]
    [else
     (define pre-string (number->pre-string N base upper?))
     (define digits
       (cond
         [(string? pre-string) (string->list pre-string)]
         [else pre-string]))
     (apply string
            (separate-groups (reverse digits)
                             (reverse groups)
                             (string->list group-sep)))]))
```

Metaxal:
Thanks for the benchmark and catching the large slowdown factor!

I thought I had already made a fast pass, but I missed the most important case. I'll try to find sometime before the weekend to fix it.

Metaxal:
Thank @sorawee , I've updated the code based on a simplification of your idea. It speeds it up significantly indeed when it can call the base `string->number`. It also simplifies the code, which is always good.

I haven't reverted the get-digit implementation, let me know if that's an issue.

pkgs/racket-doc/scribblings/reference/format.scrbl
```diff
@@ -402,6 +425,8 @@ the resulting string is appended to the significand:
 ]}
 
 ]
+@history[#:changed "8.5.0.3"
+         @elem{Added @racket[#:groups], @racket[#:groups-sep] and @racket[#:decimal-sep].}]
```
sorawee, line 429: Typo: `#:group-sep`

sorawee:
Re `get-digit`: it was a minor nit, definitely not a hill I'm willing to die on. So if you don't want to change it, it's your choice.

Should `#:groups` be default to non-empty list? Should there also be a check that `#:groups` is a non-empty list? Providing non-empty `#:group-sep` alone right now results in an obscure error.

Another possible approach is to check that `#:groups` is non-empty _when_ `#:group-sep` is non-empty.

Metaxal:
Good idea. The contract for groups is now `non-empty-listof exact-positive-integer?` and defaults to `'(3)` as per the International System of Units.

racket/collects/racket/format.rkt
```diff
@@ -414,28 +429,62 @@
...
+
 (define (number->string* N base upper?)
   (cond [(memv base '(2 8 10 16))
          (let ([s (number->string N base)])
            (if (and (= base 16) upper?)
                (string-upcase s)
                s))]
         [(zero? N)
-         (string #\0)]
+         "0"]
```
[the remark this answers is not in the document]
Metaxal: That makes sense now! Although yeah, mutating the result of `~r` does sound like a terrible idea.

racket/collects/racket/format.rkt
```diff
@@ -414,28 +429,62 @@
...
+                                (reverse groups)
+                                (string->list group-sep)))))))
+
 (define (number->fraction-string N base upper? precision)
   (let ([s (number->string* N base upper?)])
     (string-append (make-string (max 0 (- precision (string-length s))) #\0) s)))
 
+(define A-10 (- (char->integer #\A) 10))
+(define a-10 (- (char->integer #\a) 10))
+(define char0 (char->integer #\0))
```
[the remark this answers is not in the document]
Metaxal: Oh right, I see.

Metaxal:
All @mflatt 's comments above have been addressed.

pkgs/racket-doc/scribblings/reference/format.scrbl
```diff
@@ -425,7 +425,7 @@ the resulting string is appended to the significand:
 ]}
 
 ]
-@history[#:changed "8.5.0.3"
+@history[#:changed "8.5.0.5"
```
sorawee, line 428: It's actually 8.5.0.7 now... (https://github.com/racket/racket/blob/master/pkgs/base/info.rkt#L17)

mflatt: Well, close enough

PR #4226, opened 2022-05-05
Allow showing arbitrary pass results.

samth:
Here's an example of the output:

```
$ PLT_LINKLET_SHOW_PASSES="cpnanopass np-recognize-let" racket -e 1
;; linklet ---------------------
(linklet
  ((.top-level-bind! .top-level-require!)
    (.mpi-vector .syntax-literals)
    (.namespace .phase .self .inspector .bulk-binding-registry
      .set-transformer!))
  ()
  1)
;; schemified ---------------------
(lambda (instance-variable-reference .top-level-bind!1
         .top-level-require!2 .mpi-vector3 .syntax-literals4
         .namespace5 .phase6 .self7 .inspector8
         .bulk-binding-registry9 .set-transformer!10)
  1)
;; cpnanopass np-recognize-let ---------------------
output of cpnanopass:
(case-lambda
  [clause
   ()
   0
   (case-lambda
     [clause
      (#{instance-variable-reference n4kgr6aybd814a9h7f37s1lxn-0} #{\x2E;top-level-bind!1 n4kgr6aybd814a9h7f37s1lxn-1}
        #{\x2E;top-level-require!2 n4kgr6aybd814a9h7f37s1lxn-2}
        #{\x2E;mpi-vector3 n4kgr6aybd814a9h7f37s1lxn-3}
        #{\x2E;syntax-literals4 n4kgr6aybd814a9h7f37s1lxn-4}
        #{\x2E;namespace5 n4kgr6aybd814a9h7f37s1lxn-5}
        #{\x2E;phase6 n4kgr6aybd814a9h7f37s1lxn-6}
        #{\x2E;self7 n4kgr6aybd814a9h7f37s1lxn-7}
        #{\x2E;inspector8 n4kgr6aybd814a9h7f37s1lxn-8}
        #{\x2E;bulk-binding-registry9 n4kgr6aybd814a9h7f37s1lxn-9}
        #{\x2E;set-transformer!10 n4kgr6aybd814a9h7f37s1lxn-10})
      11
      '1])])
output of np-recognize-let:
(case-lambda
  [clause
   ()
   0
   (case-lambda
     [clause
      (#{instance-variable-reference n4kgr6aybd814a9h7f37s1lxn-0} #{\x2E;top-level-bind!1 n4kgr6aybd814a9h7f37s1lxn-1}
        #{\x2E;top-level-require!2 n4kgr6aybd814a9h7f37s1lxn-2}
        #{\x2E;mpi-vector3 n4kgr6aybd814a9h7f37s1lxn-3}
        #{\x2E;syntax-literals4 n4kgr6aybd814a9h7f37s1lxn-4}
        #{\x2E;namespace5 n4kgr6aybd814a9h7f37s1lxn-5}
        #{\x2E;phase6 n4kgr6aybd814a9h7f37s1lxn-6}
        #{\x2E;self7 n4kgr6aybd814a9h7f37s1lxn-7}
        #{\x2E;inspector8 n4kgr6aybd814a9h7f37s1lxn-8}
        #{\x2E;bulk-binding-registry9 n4kgr6aybd814a9h7f37s1lxn-9}
        #{\x2E;set-transformer!10 n4kgr6aybd814a9h7f37s1lxn-10})
      11
      '1])])
;; compiled ---------------------
done
```

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/compiler.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/compiler.scrbl
@@ -119,6 +119,11 @@ forms or adjust the way forms are displayed:
          form after transformation by Chez Scheme's front-end
          optimizer}
 
+   @item{@envvar-indexed{PLT_LINKLET_SHOW_PASSES} --- show the
+         intermediate form of a schemified linklet after the specified
+         passes (listed space-separated) in Chez Scheme's internal
+         representation}
+
    @item{@envvar-indexed{PLT_LINKLET_SHOW_ASSEMBLY} --- show the
          compiled form of a schemified linklet in Chez Scheme's
          abstraction of machine instructions}
--- a/racket/src/cs/linklet.sls
+++ b/racket/src/cs/linklet.sls
@@ -200,6 +200,15 @@
   (define jit-demand-on? (getenv "PLT_LINKLET_SHOW_JIT_DEMAND"))
   (define literals-on? (getenv "PLT_LINKLET_SHOW_LITERALS"))
   (define known-on? (getenv "PLT_LINKLET_SHOW_KNOWN"))
+  (define passes-on (cond [(getenv "PLT_LINKLET_SHOW_PASSES")
+                           =>
+                           (lambda (e)
+                             (let lp ([p (open-string-input-port e)])
+                               (define pass (read p))
+                               (if (symbol? pass)
+                                   (cons pass (lp p))
+                                   '())))]
+                          [else '()]))
   (define cp0-on? (getenv "PLT_LINKLET_SHOW_CP0"))
   (define assembly-on? (getenv "PLT_LINKLET_SHOW_ASSEMBLY"))
   (define show-on? (or gensym-on?
@@ -210,6 +219,7 @@
                        literals-on?
                        known-on?
                        cp0-on?
+                       (pair? passes-on)
                        assembly-on?
                        (getenv "PLT_LINKLET_SHOW")))
   (define show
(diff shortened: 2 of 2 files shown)


mflatt:
This looks like a good idea. A few comments:

* It would be nice to support multiple passes at a time. (That's especially useful for seeing what pass does.) So, maybe parse the environment variable using `read` in a loop to extract a sequence of symbols?
* The`parameterize` form should work on `$np-tracer`, since the Chez Scheme representation of parameters is not distinct from other procedures. Separately, it might matter that `$np-tracer` is not thread-safe, but I'm ok with ignoring that issue for now.
* Minor: untabifying would be more consistent with the current file content.
* Nit: I recommend `cond` over nested `if`s here.

rmculpepper:
I would really like it if access to this information could be controlled within Racket. One implementation is a parameter containing a list of symbols indicating which passes to print. The parameter could be initialized from the environment variables above, for compatibility.

samth:
This is now ready. I wasn't sure how to use `cond` in this case.

mflatt:
Seems ok. FWIW, I'd parse the environment variable just once on startup (changing `passes-on?` to `passes-on` as a list that's empty by default).

samth:
I fixed that, but now I realize that this plays poorly with `PLT_LINKLET_SHOW_ASSEMBLY` in a way that I don't know how to fix. In particular, the header lines appear together, rather than in the correct spot:

```
;; cp0 ---------------------
(lambda (instance-variable-reference .top-level-bind!1
         .top-level-require!2 .mpi-vector3 .syntax-literals4
         .namespace5 .phase6 .self7 .inspector8
         .bulk-binding-registry9 .set-transformer!10)
  1)
;; assembly ---------------------
;; cpnanopass ---------------------
output of cpnanopass:
(case-lambda
  [clause
   ()
   0
   (case-lambda
     [clause
      (#{instance-variable-reference n8vokmsf9od8xe3o4z60nci9x-0} #{\x2E;top-level-bind!1 n8vokmsf9od8xe3o4z60nci9x-1}
        #{\x2E;top-level-require!2 n8vokmsf9od8xe3o4z60nci9x-2}
        #{\x2E;mpi-vector3 n8vokmsf9od8xe3o4z60nci9x-3}
        #{\x2E;syntax-literals4 n8vokmsf9od8xe3o4z60nci9x-4}
        #{\x2E;namespace5 n8vokmsf9od8xe3o4z60nci9x-5}
        #{\x2E;phase6 n8vokmsf9od8xe3o4z60nci9x-6}
        #{\x2E;self7 n8vokmsf9od8xe3o4z60nci9x-7}
        #{\x2E;inspector8 n8vokmsf9od8xe3o4z60nci9x-8}
        #{\x2E;bulk-binding-registry9 n8vokmsf9od8xe3o4z60nci9x-9}
        #{\x2E;set-transformer!10 n8vokmsf9od8xe3o4z60nci9x-10})
      11
      '1])])

entry.12:
0:       cmpi           (imm 11), %ac0
4:       bne            lf.11(11)
dcl.13:
6:       movi           (imm 8), %ac0
13:      jmp            (disp 0 %sfp)
lf.11:
17:      movi           (imm 4294967295), %ts
27:      jmp            %ts
29:      relocation:    (x86_64-jump 65 (library-code #(libspec doargerr 170)))
29:      <end>

dcl.14:
0:       movi           (literal 0 (closure . #($c-func (...) #f (...) #f))), %ac0
10:      jmp            (disp 0 %sfp)
14:      <end>
;; compiled ---------------------
```

Is this just the best we can do?

mflatt:
I noticed that, and I don't think it matters enough to worry about. It could make sense to print one header for both, though — maybe "passes and assembly".

mflatt:
LGTM

Issue #4228, opened 2022-05-07
racket style guide contract-out example missing a require


HeladoDeBrownie:
the example [starting here](https://github.com/racket/racket/blob/448b77a6629c68659e1360fbe9f9e1ecea078f9c/pkgs/racket-doc/scribblings/style/unit.scrbl#L111) is missing a `(require racket/contract)` or `(require racket/contract/base)` preceding the `provide`, without which racket would complain that `contract-out` is not a provide sub-form

mfelleisen:
See be6ea6d

HeladoDeBrownie:
thank you for the examples. i wasn't 100% on the style to use since it's more important to get the style correct in the style guide. i've made a couple of tweaks to this fix in #4230

PR #4230, opened 2022-05-07
minor fixes to module section of style guide

HeladoDeBrownie:
related to #4228. removes an unneeded `require` and makes more precise why the other `require` is needed

### DIFF
--- a/pkgs/racket-doc/scribblings/style/unit.scrbl
+++ b/pkgs/racket-doc/scribblings/style/unit.scrbl
@@ -78,8 +78,6 @@ implementation:
 
 ;; the module implements a tv server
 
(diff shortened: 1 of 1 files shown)


mfelleisen (commented):
Nah, we leave it like this. Technically you can move the `require` anywhere in the module. I prefer to put those ahead of `provide` that mater to the exports.

HeladoDeBrownie:
the one i removed isn't needed because it's only needed for `contract-out`, which doesn't appear in that example, only the one below

mfelleisen (approved):

mfelleisen:
Oops of course.

PR #4236, opened 2022-05-12
WIP: Add equal-always? with gen:equal-mode+hash

AlexKnauth:
An updated version of https://github.com/racket/racket/pull/4076 with `gen:equal-mode+hash` included.

Things left to do:
 - [ ] Add documentation for the associated hash-tables and sets
 - [ ] In the docs for `prop:equal+hash`, `gen:equal+hash`, and `gen:equal-mode+hash`, document what it means for an implementation to be "honest" and work properly with `chaperone-of?` and `equal-always?`, vs "dishonest", using mutability or mode.


mflatt:
I think this may be ready to merge. Is that right?

AlexKnauth:
Yes.

mflatt:
Two extra commits here: https://github.com/mflatt/racket/commit/0627983d063cef3231125a5151b7db8d0ba3e78f and https://github.com/mflatt/racket/commit/5e94b02715be03e398255ef9e8053a0cf8579c03

Do the doc updates in that second one look ok?

Ok if I squash the commits to one (authored by @AlexKnauth) to merge?

AlexKnauth:
Sure!

AlexKnauth:
Although docs might still be a bit unclear on `gen:equal-mode+hash` not doing anything on mutable structures...

AlexKnauth:
Wait, I just tried that again and the implementation didn't do what I wanted:
```racket
(struct cell [value] #:mutable
  #:methods gen:equal-mode+hash
  [(define (equal-mode-proc self other rec mode)
     (rec (cell-value self) (cell-value other)))
   (define (hash-mode-proc self rec mode)
     (rec (cell-value self)))])

(equal-always? (cell 4) (cell 4))
```

I wanted this to return `#f`, being conservative on this.

AlexKnauth:
After some thinking on Discord, I think I was being too conservative on mutable struct equal-mode+hash. I've updated the tests and documentation examples to include mutable structs that use the mode.

PR #4237, opened 2022-05-16
Index `match` keywords such as `app`.

### LINKED ISSUE #1653: Search for "app match" doesn't result in useful results
I wanted documentation on the the `app` form of `match`. But the search results didn't have anything in it even relating to `match`.  This is likely because `app` is not indexed as it is entirely matched on symbolic name.

https://docs.racket-lang.org/search/index.html?q=app%20match

samth:
Closes #1653.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -2,6 +2,7 @@
 @(require "mz.rkt" "match-grammar.rkt" racket/match)
 
 @(define match-eval (make-base-eval))
+@(define (match-kw s) (index (list s) (racketidfont s)))
 @examples[#:hidden #:eval match-eval (require racket/match racket/list)]
 @examples[#:hidden #:eval match-eval (require (for-syntax racket/base))]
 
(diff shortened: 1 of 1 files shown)


samth:
This produces search results that say "app in reference". It's not clear to me whether it's possible to do better than that.

pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -202,7 +203,7 @@ In more detail, patterns match as follows:
          the patterns @racket[(list-no-order x 1 x)] and
          @racket[(list-no-order x 1 x ...)] both produce syntax errors.}}
 
- @item{@racket[(#,(racketidfont "list-no-order") _pat ... _lvp)] ---
+ @item{@racket[(#,(match-kw "list-no-order") _pat ... _lvp)] ---
```
sorawee, line 206: This is the second indexing. Is it OK? (It could be OK, but I'm not sure)

samth: Ah, that was not intended.

samth:
```suggestion
 @item{@racket[(#,(racketidfont "list-no-order") _pat ... _lvp)] ---
```

PR #4240, opened 2022-05-16
add hash-map/copy and dict-map/copy

AlexKnauth:
So that it can be used instead of enumerating variants of hash tables.

### DIFF
--- a/racket/collects/racket/private/hash.rkt
+++ b/racket/collects/racket/private/hash.rkt
@@ -96,6 +96,8 @@
   (define (hash-map/copy table f)
     (unless (hash? table)
       (raise-argument-error 'hash-map/copy "hash?" table))
+    (unless (and (procedure? f) (procedure-arity-includes? f 2))
+      (raise-argument-error 'hash-map/copy "(procedure-arity-includes/c 2)" f))
     (cond
      [(immutable? table)
       (for/fold ([acc (hash-copy-clear table)])
--- a/racket/collects/racket/private/hash.rkt
+++ b/racket/collects/racket/private/hash.rkt
@@ -93,6 +93,22 @@
        [(hash-eqv? table) (make-hasheqv)]
        [(hash-eq? table) (make-hasheq)])]))
 
+  (define (hash-map/copy table f)
+    (unless (hash? table)
+      (raise-argument-error 'hash-map/copy "hash?" table))
+    (cond
+     [(immutable? table)
+      (for/fold ([acc (hash-copy-clear table)])
+                ([(k1 v1) (in-hash table)])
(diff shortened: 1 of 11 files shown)


racket/collects/racket/private/hash.rkt
```diff
@@ -93,6 +93,22 @@
        [(hash-eqv? table) (make-hasheqv)]
        [(hash-eq? table) (make-hasheq)])]))
 
+  (define (hash-map/copy table f)
+    (unless (hash? table)
+      (raise-argument-error 'hash-map/copy "hash?" table))
+    (cond
+     [(immutable? table)
+      (for/fold ([acc (hash-copy-clear table)])
+                ([(k1 v1) (in-hash table)])
```
sorawee, line 102: Not sure if it matters, but there's a family of `in-XXX-hash`, like `in-immutable-hash`, that could make iteration faster.

AlexKnauth: I can use `in-immutable-hash`, but I don't want to use the others because it's simpler to have the same code path handle mutable, weak, and ephemeron hashes.

racket/collects/racket/private/hash.rkt
```diff
@@ -93,6 +93,22 @@
        [(hash-eqv? table) (make-hasheqv)]
        [(hash-eq? table) (make-hasheq)])]))
 
+  (define (hash-map/copy table f)
+    (unless (hash? table)
+      (raise-argument-error 'hash-map/copy "hash?" table))
+    (cond
```
mflatt, line 99: Should there be a  check here that `f` is a procedure that accepts 2 arguments?

AlexKnauth: The contract in `racket/collects/racket/dict.rkt` has
```racket
 [dict-map/copy
  (->i ([d dict?]
        [proc (d) (-> (dict-key-contract d) (dict-value-contract d)
                      (values (dict-key-contract d) (dict-value-contract d)))])
       [_r dict?])]
```
But code that uses `racket/collects/racket/private/hash.rkt` directly will bypass that contract.

AlexKnauth: Oh, sorry I got halfway through fixing it and my mind switched to the dict version. Yes, the hash version should have it, but it should be unnecessary for the dict version.

AlexKnauth: I've updated `hash-map/copy`.

mflatt:
This is ready to merge, right?

samth:
Can you change the `info.rkt` for the expander to require the new version?

PR #4244, opened 2022-05-18
ci: strip path for correct reporting

sorawee:


sorawee:
There are still paths that are not in the right place. For example:

- `racket/src/cs/c/ChezScheme/ta6le/lz4/lib/lz4.c` (actual path: `racket/src/ChezScheme/lz4/lib/lz4.c`)
- `racket/src/bc/rktio/conftest.c` (autogenerated?)

Not sure what would be a nice way to deal with these cases.

samth:
Chez Scheme creates links in the build process.

`conftest.c` is generated by `configure`. 

I assume that both of those problems were there before.

PR #4245, opened 2022-05-19
hash-map/copy: add optional kind argument, update schemified

AlexKnauth:
To fix a mistake I made in my #4240, this adds an optional `kind` argument to `hash-map/copy`, because some uses of `hash-map/copy` need to produce immutable versions. In particular, the messages for places in "thread.scm" need to use `#:kind 'immutable` with `hash-map/copy` because they are meant to convert to immutable hashes.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -518,10 +518,24 @@ performed in constant time.  If @racket[hash] is a @tech{chaperone},
 then each key is removed one-by-one using @racket[hash-remove].}
 
 
-@defproc[(hash-copy-clear [hash hash?]) hash?]{
+@defproc[(hash-copy-clear
+          [hash hash?]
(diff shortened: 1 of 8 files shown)


rmculpepper:
Why not add an optional argument to `hash-map/copy` instead? Say, `'immutable` make an immutable version, `'mutable` makes a mutable version, and `#f` (default) makes the same kind as the argument.

AlexKnauth:
Okay, it could be something similar to the `#:kind` argument to `set/c` then...
```racket
kind : (or/c 'immutable 'mutable 'weak 'ephemeron #f)
     = #f
```
If `kind` is not supplied or `#f`, produces a hash table of the same kind and mutability as the given `hash`. If `kind` is `'immutable`, `'mutable`, `'weak`, or `'ephemeron`, produces a table that's immutable, mutable with strongly-held keys, mutable with weakly-held keys, or mutable with ephemeron-held keys respectively.

The simplest way I can think of to implement this would be adding the optional `kind` argument to both `hash-copy-clear` and `hash-map/copy`.

AlexKnauth:
Can this be merged soon?

PR #4247, opened 2022-05-19
match: fix tests

sorawee:
Many tests are meant to test the current, modern `match` form,
but it is actually testing the legacy `match` form.
It just happens that these tests are valid for both `match` forms,
and deceptively produce the same matching behaviors.

This PR does the following:

1. Move `match-tests.rkt` to `legacy-match-tests.rkt` and remove
all (one) modern `match` test from the file.
The modern test is introduced in
https://github.com/racket/racket/commit/393582492203608a579a062b3fcece0ad4e18e88

2. Copy `legacy-match-tests.rkt` to `match-tests.rkt` and port these
tests to use the modern `match`. Also, restore the modern test introduced in
https://github.com/racket/racket/commit/393582492203608a579a062b3fcece0ad4e18e88

3. Make `match-exn-tests.rkt` test modern `match`.
This test file is introduced in
https://github.com/racket/racket/commit/026d368a38560b474e13633fdb75aae5b3c3384b
and it is meant to test the modern `match`.

### DIFF
--- a/pkgs/racket-test/tests/match/legacy-match-tests.rkt
+++ b/pkgs/racket-test/tests/match/legacy-match-tests.rkt
@@ -1,26 +1,25 @@
-(module match-tests racket/base
-  (require racket/match rackunit
-           (for-syntax racket/base))
+(module legacy-match-tests mzscheme
+  (require mzlib/match rackunit)
   
-  (provide match-tests)
+  (provide legacy-match-tests)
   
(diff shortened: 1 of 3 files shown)


samth:
Can you fix the conflicts here?

sorawee:
Oh, sorry! I thought GitHub merges branches successfully. Will do.

PR #4252, opened 2022-05-23
expander: add module-cache-clear!

Bogdanp:
This provides a way to clear the module cache from Racket for programs that want to reduce their memory footprint.

I couldn't decide between calling the new function `clear-module-cache!` and `module-cache-clear!`. I only went with the latter because I noticed most of the other functions in `main-primitive` were prefixed.  Happy to change it if anyone feels strongly about it.

For the docs, I didn't find an existing place where this particular cache is mentioned so I added a new section, but maybe there's a more appropriate place to put this.


sorawee:
@Bogdanp I think there is a conflict in `startup.inc`. Currently, your PR removes the recently added equal always feature. Perhaps these files should be regenerated from the latest version.

Bogdanp:
Thanks @sorawee! Looks like I forgot to run `make derived` after rebasing. Should be fixed now.

PR #4254, opened 2022-05-23
Add missing docs for equal-always related list/hash ops

AlexKnauth:
Thanks to @samth for pointing out these missing docs from Typed Racket `test-docs-complete`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/pairs.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/pairs.scrbl
@@ -518,7 +518,10 @@ Returns @racket[(remove v lst equal-always?)].
   (remw 2 (list 1 2 3 4 5))
   (remw '(2) (list '(1) '(2) '(3)))
   (remw "2" (list "1" "2" "3"))
-  (remw #\c (list #\a #\b #\c))]
+  (remw #\c (list #\a #\b #\c))
+  (define b1 (box 5))
+  (define b2 (box 5))
+  (remw b2 (list 0 b1 1 b2 2))]
 
 @history[#:added "8.5.0.3"]}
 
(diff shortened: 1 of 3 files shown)


pkgs/racket-doc/scribblings/reference/pairs.scrbl
```diff
@@ -509,6 +509,20 @@ Returns @racket[(remove v lst eqv?)].
   (remv #\c (list #\a #\b #\c))]}
 
 
+@defproc[(remw [v any/c] [lst list?])
+         list?]{
+
+Returns @racket[(remove v lst equal-always?)].
+
+@mz-examples[
```
sorawee, line 517: Can these examples be a bit more illuminating? It right now doesn't distinguish it from `remove`.

AlexKnauth: If they involved boxes, something like
```racket
> (define b1 (box 5))
> (define b2 (box 5))
> (define l (remw b2 (list b1 b2)))
> (set-box! b1 (box 13))
> l
(list (box 5))
```
where if it was based on `equal?` it would be `(list (box 13))`.

But that seems more complicated than these docs need to be, in my opinion.

AlexKnauth: Might even be the case that a `memw` version would be clearer, if it used a "flattened" version of the `assw` example above.

AlexKnauth: Using order relative to other elements might also make the `remw` example clear enough to not need the mutation in between... what do you think?

Is this version without the mutation clear enough?
```
> (define b1 (box 5))
> (define b2 (box 5))
> (remw b2 (list 0 b1 1 b2 2))
(list 0 (box 5) 1 2)
```

Or is this version with the mutation better?
```
> (define b1 (box 5))
> (define b2 (box 5))
> (define l (remw b2 (list 0 b1 1 b2 2)))
> (set-box! b2 13)
> l
(list 0 (box 5) 1 2)
```

sorawee: That looks great!

AlexKnauth: Okay, I've added better examples and tests for `assw`, `memw`, and `remw`. Not using mutation, but using box identity and list order to demonstrate the difference.

PR #4256, opened 2022-05-25
Zuo: Use SHA-256.

LiberalArtist:
This PR changes Zuo to use SHA-256 rather than SHA-1. It avoids adding any new C code by moving the implementation from `rktio_sha2.c` into `zuo.c`, then using Zuo to generate `rktio_sha2.c`.

Currently, this is a just a proof-of-concept: if it seems like a useful thing to do, some cleanup will be needed. In particular, I have not renamed `string-sha1`, `file-sha1`, `sha1?`, or `no-sha1`, just changed the implementations to actually deal with SHA-256.

(Some projects transitioning from SHA-1 to SHA-256 have also changed to avoid mentioning the hash algorithm in API names, to ease a transition some day from SHA-256 to something else. Personally, I think `string-sha256` is good enough, given that `hash` means something else in Racket/Zuo and SHA-256 seems likely to be viable for a long time.)

I don't know of any currently-existing problem with SHA-1 as used by Zuo, but SHA-256 seems like a better choice as a basis for something new.

### DIFF
--- a/racket/src/zuo/zuo-doc/zuo-build.scrbl
+++ b/racket/src/zuo/zuo-doc/zuo-build.scrbl
@@ -6,7 +6,7 @@
 
 @(define shake-url "https://shakebuild.com/")
 
-@title[#:tag "zuo-build"]{Zuo as a @tt{make} Replacement}
+@title[#:tag "zuo-build"]{Zuo as a @exec{make} Replacement}
 
 @defzuomodule[zuo/build]
 
@@ -26,13 +26,15 @@ and build steps. The library has two layers:
 
 A @tech{target} represents either an input to a build (such as a
 source file) or a generated output, and a target can depend on any
-number of other targets. A target's output is represented by 40-character
-string that is normally a SHA-1 hash; the @racket[build] procedure
+number of other targets. A target's output is represented by a
+string that is normally an SHA-256 hash: more precisely, it is
+represented by a value satisfying the predicate @racket[sha256?].
+The @racket[build] procedure
 records hashes and dependencies in a database located alongside
 non-input targets, so it can avoid rebuilding targets when nothing has
-changed since the last build. Unlike @tt{make}, timestamps are used
-only as a shortcut to avoiding computing the SHA-1 of a file (i.e., if
-the timestamp has not changes, the SHA-1 result is assumed to be
+changed since the last build. Unlike @exec{make}, timestamps are used
+only as a shortcut to avoiding computing the SHA-256 of a file (i.e., if
+the timestamp has not changed, the SHA-256 result is assumed to be
 unchanged).
 
 ``Recursive make'' is encouraged in the sense that a target's build
@@ -55,14 +57,14 @@ skipped if the target isn't needed. Instead, @racket[target] takes a
 are needed. The @racket[_get-rule] procedure returns up to three
 results in a @racket[rule] record: a list of dependencies; the hash of
 an already-built version of the target, if one exists, where
-@racket[file-sha1] is used by default; and a @racket[_rebuild]
+@racket[file-sha256] is used by default; and a @racket[_rebuild]
 procedure that is called if the returned hash, the hash of
 dependencies (rebuilt if needed), and recorded results from a previous
 build together determine that a rebuild is needed.
 
 When a target's @racket[_rebuild] function is called, it optionally
 returns a hash for the result of the build if the target's
-@racket[rule] had one, otherwise @racket[file-sha1] is used to get a
+@racket[rule] had one, otherwise @racket[file-sha256] is used to get a
 result hash. Either way, it's possible that the result hash is the
 same the one returned by @racket[_get-rule]; that is, maybe a
 dependency of the target changed, but the change turned out not to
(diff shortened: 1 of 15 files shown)


mflatt:
It's fine to swap SHA-256 in place of SHA-1 in Zuo, but extracting that implementation from Zuo to use in rktio does not seem like a good idea.

If the code was big enough or was likely to need changes or improvements, we could make another layer that is shared by Zuo and rktio. As it is, just duplicating the implementation seems better.

racket/src/zuo/zuo.c
```diff
@@ -6116,163 +6116,499 @@ zuo_t *zuo_process_wait(zuo_t *pids_i) {
...
+    unsigned state[8];
+    unsigned char buffer[64];
+    int is224;
+} zuo_sha2_ctx_t;
+
+#define ZUO_SHA256_DIGEST_SIZE 32
+
+/* rktio requires certain integer sizes, anyway: */
+typedef unsigned int uint32_sha2_t;
+typedef uintptr_t size_sha2_t;
```
LiberalArtist, line 6133: I wasn't sure if these two typedefs were right for Zuo. For the first, it would seem natural to write:
```c
typedef zuo_uint32_t uint32_sha2_t;
```
For the second, there is a `zuo_intptr_t`, but not a `zuo_uintptr_t`. On Unix, it is indeed just `intptr_t`, but on Windows:
```c
# ifdef _WIN64
typedef long long zuo_intptr_t;
# else
typedef long zuo_intptr_t;
# endif
```
Would it be better to add a `zuo_uintptr_t` and use `unsigned long long` etc., or just to keep `uintptr_t` for `size_sha2_t`?

(I definitely don't claim to understand the portability pitfalls of C integer sizes.)

mflatt: It's better to add `zuo_uintptr_t`. If I remembr correctly, trying to use `intptr_t` and `uintptr_t` created trouble with an MSVC installation that is very old but that I still use sometimes.

racket/src/zuo/zuo-doc/zuo-build.scrbl
```diff
@@ -494,18 +494,21 @@ such a target to one where @racket[args] are the argument when the
...
-The @racket[file-sha1] procedure returns the SHA-1 hash of the content
-of @racket[file], or it returns @racket[no-sha1] if @racket[file] does
+The @racket[file-sha256] procedure returns the SHA-256 hash of the content
+of @racket[file], or it returns @racket[no-sha256] if @racket[file] does
 not exist.
 
-The @racket[sha1?] predicate recognizes values that are either a
-40-character string or @racket[no-sha1].}
+The @racket[sha256?] predicate recognizes values that are either a
+64-character string (thus, @racket[sha256-length]) or @racket[no-sha256].
```
LiberalArtist, line 507: This doesn't cover the concatenation case:
```racket
(define (sha256? s)
  (or (and (string? s)
           (or (= (string-length s) sha256-length)
               (string=? s no-sha256)
               ;; also allow a concatenation of SHA-256s for multi-file targets
               (and (>= (string-length s) sha256-length)
                    (= 0 (modulo (string-length s) sha256-length)))))
      (eq? s phony-sha256)))
```
Should it, or is that intended to be an implementation detail, like `phony-sha256`?

LiberalArtist:
I've pushed an updated commit. One thing I've not tried to address is smooth transitioning from old SHA-1 based `_zuo.db` and `_zuo_tc.db` files. Some options I can imagine are:

 1. Do nothing, hoping the files will be treated as corrupt and deleted.
 2. Use the different lengths to determine if we have an old db, and ignore its contents if so.
 3. Add a version to the beginning of the database, and ignore it if it's wrong.

I think I prefer the last option, which would be able to deal with any future format changes, and would also avoid questions like how long to wait before removing the code to detect and handle SHA-1­–based databases.

racket/src/zuo/zuo-doc/zuo-build.scrbl
mflatt: The documentation should be repaired/updated to reflect the implementation.

mflatt:
I don't think anything needs to be done to smooth the transition at this point in Zuo's lifecycle.

racket/src/zuo/zuo-doc/zuo-build.scrbl
LiberalArtist: I've pushed an update.

mflatt:
I see now why some transition smoothing is needed, at least for the Racket build. I suggest adjusting `file-sha256/cached` so that it doesn't return an ill-formed result:

```
(define (file-sha256/cached path time-cache)
  (let ([timestamp (file-timestamp path)])
    (and timestamp
         (let ([cached (and time-cache
                            (hash-ref time-cache
                                      (string->symbol path)
                                      #f))])
           (if (and cached
                    (equal? (car cached) timestamp)
                    (string? (cadr cached))
                    (= sha256-length (string-length (cadr cached))))
               (cadr cached)
               (string-sha256 (file->string path)))))))
```

LiberalArtist:
That makes sense. I've pushed an amended commit.

PR #4264, opened 2022-05-28
`compose`: ignore `values`.

NoahStoryM:
Before:
```racket
Welcome to Racket v8.5.0.5 [cs].
> (equal? + (compose values + values))
#f
```

After:
```racket
Welcome to Racket v8.5.0.5 [cs].
> (equal? + (compose values + values))
#t
```


racket/collects/racket/private/list.rkt
```diff
@@ -323,31 +323,36 @@
             [else (cons (fcn j)
                         (recr (add1 j) (sub1 i)))])))
 
-  (define-values [compose1 compose]
+  (define-values (compose1 compose)
     (let ()
+      (define (id? f) (equal? values f))
```
mflatt, line 328: I think this needs to be `eq?` instead of `equal?` in case a chaperoned `values` is supplied.

racket/collects/racket/private/list.rkt
```diff
@@ -358,46 +363,58 @@
...
-        ;;   (pipeline1 (let ([fst (car rfuns)]) (lambda (x) (fst (f x))))
-        ;;              (cdr rfuns)))
-        (lambda (x)
-          (let loop ([x x] [f f] [rfuns rfuns])
-            (if (null? rfuns)
-              (f x)
-              (loop (f x) (car rfuns) (cdr rfuns))))))
+        (cond
+          [(andmap id? rfuns) f]
+          [(id? f) (pipeline1 (car rfuns) (cdr rfuns))]
```
mflatt, line 368: This is quadratic time in the case that all functions are identity except the last one in `rfuns`. Maybe special-case an empty `rfuns` after filtering, instead?

Also, as written here, it seems like `f` would not be reduced to single-argument arity like it was before, and maybe also doesn't get renamed as before.

racket/collects/racket/private/list.rkt
```diff
@@ -358,46 +363,58 @@
...
+          [else
+           (let ([rfuns (remove* (list values) rfuns)])
+             ;; use the other composition style in this case, to optimize an
+             ;; occasional arity-1 procedure in the pipeline
+             (if (eqv? 1 (procedure-arity f))
+                 ;; if `f' is single arity, then going in reverse they will *all* be
+                 ;; single arities
+                 (let loop ([f f] [rfuns rfuns])
+                   (if (null? rfuns)
+                       (procedure-rename f 'composed)
```
mflatt, line 396: I can see how this preseves old behavior when multiple functions were provided, but does this change the old behavior when only one function is provided?

racket/collects/racket/private/list.rkt
```diff
@@ -409,17 +426,24 @@
                  (raise-argument-error 'name "procedure?" 0 f g))
                (unless (procedure? g)
                  (raise-argument-error 'name "procedure?" 1 f g))
-               (can-compose name 0 f f '())
-               (simple-compose f g)]
+               (cond
+                 [(id? f) g]
+                 [(id? g) f]
```
mflatt, line 431: I think these shortcuts have the same issue with not changing a function's arity in the case of `compose1`, and probably also the same naming issue.

racket/collects/racket/private/list.rkt
```diff
@@ -409,17 +426,24 @@
...
-                 (if (pair? fs)
-                   (begin (can-compose name i f f0 fs0)
-                          (loop (car fs) (cdr fs) (add1 i) (cons f rfuns)))
-                   (simple-compose (pipeline (car rfuns) (cdr rfuns)) f)))]))))
+                 (cond
+                   [(pair? fs)
+                    (can-compose name i f f0 fs0)
+                    (loop (car fs) (cdr fs) (add1 i) (cons f rfuns))]
+                   [(andmap id? rfuns) f]
+                   [(id? f) (pipeline (car rfuns) (cdr rfuns))]
```
mflatt, line 445: Same quadratic-time issue here, I think.

mflatt (commented):
This looks like a fine idea, but I think there are some things to fix to avoid changing the current behavior and to avoid an unlikely but possible quadratic-time recursion. There will need to be tests to make sure this works as intended and preserves behavior.

racket/collects/racket/private/list.rkt
NoahStoryM: If `eq?` is used here I am concerned that TR may behave unexpectedly.

For example:
```racket
#lang typed/racket/base
(module m racket/base
  (provide (all-defined-out))
  (define (id? f) (eq? values f)))
(require/typed 'm
  [id? (-> Any Boolean)])
(displayln (id? values)) ; #f

```

NoahStoryM:
I'll add tests tomorrow.

racket/collects/racket/private/list.rkt
samth: That seems like the behavior we would expect in that situation.

NoahStoryM: Yes, so I think `equal?` might be appropriate here.

sorawee: What's the motivation for this PR? I personally don't think that it's worth the complexity added to the code, unless you can demonstrate that it makes the function more efficient, and this programming pattern is frequently seen in the wild.

Trying to ad-hoc recognize "identity functions" is bound to fail. You can test for `values`, and perhaps chaperoned `values`. Currently you are missing `identity` from `racket/function`. And what about `(lambda (x) x)`? What about `(lambda xs (apply values xs))`, etc, etc.

NoahStoryM: I guess it's worth it, the reason why `compose` should ignore `values` is because `(compose)` evaluates to `values`, so I'm not going to consider other cases.

Treating `compose` as monoid may be convenient for some scenarios (I guess `Qi` may need it).

samth: dropping a chaperoned version of values is absolutely not the right thing to do. That could cause a check that should happen not to happen.

NoahStoryM: You are right, I'll fix it later.

mflatt (changes requested):
Old version:

```
> ((compose1 + values) 1 2)
result arity mismatch;
 expected number of values not received
  expected: 1
  received: 2
```

After these changes:

```
> ((compose1 + values) 1 2)
3
```

I think the old error needs to be preserved on the grounds that `compose1` is defined to restrict the result of procedures (all but the first one) to a single result value. Of course, `compose` doesn't have that constraint.

Meanwhile, I see that arity reporting wasn't very precise before. This commit makes the arity as reported by `procedure-arity` sometimes more precise than it used to be. That's technically a change, but I think it's probably an ok change.

NoahStoryM:
It's ok for me, I also realize that this change isn't as simple as I thought.

rfindler:
> > it would be an error only if that funciton is applied
> 
> I wonder if this behavior is worth keeping? The original `compose1` seems to generate an unusable function, when should such a function exist?

This kind of reasoning is good, IMO, when you're designing a new API. But when we have one already out there and widely used, this kind of reasoning doesn't really work. In particular, someone may have written code that always uses compose but have a check that effectively keeps the function being applied when it would fail. If the unusable functions weren't created in the beginning and an error raised, that could would've been written differently, but since that check wasn't there before ... well, it wasn't. And now when someone upgrades Racket some program mysteriously breaks (because someone else wrote that code, for example).

Does this make sense?

NoahStoryM:
I can understand that in some cases for compatibility, some improvements should not be adopted immediately (like the hash operators in Typed Racket), and I believe these cases should be based on the correct logic.

I think generating an unusable function is a bug, and even if we keep the original behavior and don't throw exceptions right away, that doesn't mean those exceptions won't be thrown later -- they're just hidden.

rfindler:
> I think generating an unusable function is a bug, and even if we keep the original behavior and don't throw exceptions right away, that doesn't mean those exceptions won't be thrown later -- they're just hidden.

In my opinion (and your mileage may vary) when an existing function has something like that and it has been that way a long time, then we get into the realm of where we're effectively controlling how our users spend their time. That is, there are probably many things wrong with any given codebase. In an ideal world, the programmers working on the codebase and the dictates from their financial (or otherwise) situation should determine what they attend to. But, when we make a change like this, we are reaching out and forcing them to do certain things in certain orders, and that might be a big disadvantage to them.

Yes, for sure the function is broken, and yes, for sure, it would be nice to fix it. But I have been in the situation where I get out some old talk or some old lecture notes and I'd like to show it to someone. And the last time I ran that code was 10 years ago when I last did that presentation or taught that class. And now, the code just breaks in some way that seems very mysterious to me. It is a big drag that I have to now go back to an old version of Racket (well, that doesn't work on my new machine because the machines at the time and the machines now are different architectures or worked on very old versions of my OS or whatever) or I have to stop whatever I'm doing (trying to demo) and spend 20-30 minutes to figure out what this error message means and where it's coming from and remember what were the breaking changes since then to hone in on what I need to do to fix.

This is, IMO, a bad experience for Racket programmers and one I think we should avoid if we possibly can. And it seems to me that, in this case, we can!

Again, this is my $0.02, but it is something I've felt for a long time and it has guided me in my own Racket development. I'd like to see it be something that Racket developers can rely on. (It definitely isn't something that other language's developers can rely on.)

NoahStoryM:
Hi @rfindler, I rarely run programs from more than ten years ago, so I haven't met such a problem, my opinion may be one-sided.

It seems to me that it is more appropriate to run old code in old version language? If the old code needs to be demoed in latest version, why not check it out before the demo starts? (This is a problem if someone wants to demo old code on a whim. But is it common?)

I think the core of the problem is that Racket has found problems, but needs to ignore them for various reasons (such as compatibility). But I think at least the programmers should be made aware of them, I wonder if there is some mechanism in Racket to issue a warning?

NoahStoryM:
Hi @mflatt, I'm not sure if I miss something, it seems that `compose1` checks its arguments even in old version.

old version:

```racket
Welcome to Racket v8.5 [cs].
> (compose cons values)
#<procedure:composed>
> ((compose cons values) 1 2)
'(1 . 2)
> (compose1 cons values)
compose1: contract violation
  expected: (any/c . -> . any/c)
  given: #<procedure:cons>
 [,bt for context]  
```

after change:
```racket
Welcome to Racket v8.5 [cs].
> (compose cons values)
#<procedure:cons>
> ((compose cons values) 1 2)
'(1 . 2)
> (compose1 cons values)
compose1: contract violation
  expected: (any/c . -> . any/c)
  given: #<procedure:cons>
 [,bt for context]
```

rfindler:
Thanks, @NoahStoryM . I would say that backwards compatibility has been one of the things I've long pushed for in Racket and many of the core developers also agree on this point. I do think that there is more we can do to help people with warnings and suchlike in situations like this (perhaps even tools that can detect problems). 

I think that your suggestion about being prepared boils down to pushing this problem back to the client of Racket (ie it is their responsibility to keep up with breaking changes that Racket makes) so I cannot get behind that one!

mfelleisen:
Noah writes "I rarely run programs from more than ten years ago, so I haven't met such a problem, my opinion may be one-sided.” 

Think about the age of libraries. 2htdp/universe was started around 2002, and it is used by a few thousand students every semester. They all run sw from way more than 10 years ago (indeed, some of them run sw from _before_ they were born).

NoahStoryM:
Thanks for your reminders and examples! I can understand the importance of compatibility. But I think this problem may not be just a compatibility problem. If we find some design flaws, and trying to fix these flaws may cause the old code to crash, how should we deal with it? Ignore them, or issue a warning, or force programmers to correct their code? It doesn't make sense to me to find the problem and then ignore it completely.

mflatt:
> Hi @mflatt, I'm not sure if I miss something, it seems that `compose1` checks its arguments even in old version.

Oh, right — good! It looks like I probably got confused between `compose` and `compose1`. Now that the release branch has been made, I'll merge the change.

PR #4273, opened 2022-05-30
Add `(require racket/class)` to code example in class guide chapter

xxyzz:
This guide doesn't mention `racket/class` is not in the `racket/base`, I think this change could help other Racket learners to solve the unbound identifier error.


sorawee:
The entire guide was written under the assumption that you use `#lang racket`: see https://docs.racket-lang.org/guide/intro.html. 

However, if it is anyway a good idea to note that users may need `(require racket/class)` when they use `#lang racket/base`, it probably should be included as a margin note, and these notes should be added throughout the whole guide. https://docs.racket-lang.org/guide/match.html is another page to add the note, for instance.

PR #4278, opened 2022-06-02
add section for Visual Studio Code/Magic Racket

spdegabrielle:
I did not include the other extensions as I have not been able to identify any users of other extension.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/other-editors.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/other-editors.scrbl
@@ -278,3 +278,13 @@ place to get help from Vimmers.
 
 The @hyperlink["https://sublime.wbond.net/packages/Racket"]{Racket package}
 provides support for syntax highlighting and building for Sublime Text.
+@; ------------------------------------------------------------
+
+@section{Visual Studio Code}
+
+The Magic Racket extension provides [Language Server Protocol](https://github.com/jeapostrophe/racket-langserver#readme) 
+support, REPL integration, and syntax highlighting for Racket in Visual Studio Code.
+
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/guide/other-editors.scrbl
```diff
@@ -278,3 +278,13 @@ place to get help from Vimmers.
 
 The @hyperlink["https://sublime.wbond.net/packages/Racket"]{Racket package}
 provides support for syntax highlighting and building for Sublime Text.
+@; ------------------------------------------------------------
+
+@section{Visual Studio Code}
+
+The Magic Racket extension provides [Language Server Protocol](https://github.com/jeapostrophe/racket-langserver#readme)
```
sorawee, line 285: Scribble can't use the Markdown syntax. 

However, I will argue that we should not even make this link. I assume that the target audience of this section is a beginner. If that's the case, we should avoid providing too much information, which could be overwhelming and make them confused.

My vote is to have only one link in this section -- perhaps https://marketplace.visualstudio.com/items?itemName=evzen-wybitul.magic-racket, but make it clear that readers need to follow the instructions on the page.

sorawee (changes requested):

pkgs/racket-doc/scribblings/guide/other-editors.scrbl
spdegabrielle: I agree.

spdegabrielle: @sorawee The change now only includes the link to the extension, in the same format as the sublime text section immediately above. 

And it is now in scribble not markdown.

sorawee:
LGTM

PR #4296, opened 2022-06-11
[json]: improve `write-json`.

NoahStoryM:
This PR makes the following changes to `write-json`:
1. Add `#:format?` and `#:indent` to `write-json`.
````racket
Welcome to Racket v8.5 [cs].
> (require json)
> (displayln
     (with-output-to-string
       (λ () (write-json #hasheq((waffle . (1 2 3))
                                 (와플 . (1 2 3)))))))
{"waffle":[1,2,3],"와플":[1,2,3]}
> (displayln
     (with-output-to-string
       (λ () (write-json #hasheq((waffle . (1 2 3))
                                 (와플 . (1 2 3)))
                         #:format? #t))))
{
	"waffle": [1, 2, 3],
	"와플": [1, 2, 3]
}
> (displayln
     (with-output-to-string
       (λ () (write-json #hasheq((waffle . (1 2 3))
                                 (와플 . (1 2 3)))
                         #:format? #t
                         #:indent "  "))))
{
  "waffle": [1, 2, 3],
  "와플": [1, 2, 3]
}
````

2. Avoid generating JSON strings containing `inexact-integer`.

Before:
```racket
Welcome to Racket v8.5 [cs].
> (require json)
> (jsexpr->string 1)
"1"
> (jsexpr->string 1.)
"1.0"
```
After:
```racket
Welcome to Racket v8.5 [cs].
> (require json)
> (jsexpr->string 1)
"1"
> (jsexpr->string 1.)
"1"
```

### DIFF
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -82,7 +82,7 @@ the @rfc for more information about JSON.
   the range of @tt{U+10000} and above are encoded as two @tt{\uHHHH}
   escapes, see Section 2.5 of the @|rfc|.
 
-  If @racket[format?] is given as @racket[#f], then @racket[indent] is
+  If @racket[format?] is given as @racket[#t], then @racket[indent] is
   used to format JSON.
 
 @examples[#:eval ev
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -67,7 +67,9 @@ the @rfc for more information about JSON.
 
 @defproc[(write-json [x jsexpr?] [out output-port? (current-output-port)]
                      [#:null jsnull any/c (json-null)]
-                     [#:encode encode (or/c 'control 'all) 'control])
+                     [#:encode encode (or/c 'control 'all) 'control]
+                     [#:format? format? boolean? #f]
+                     [#:indent indent string? "\t"])
          any]{
   Writes the @racket[x] @tech{jsexpr}, encoded as JSON, to the
   @racket[out] output port.
@@ -80,18 +82,38 @@ the @rfc for more information about JSON.
   the range of @tt{U+10000} and above are encoded as two @tt{\uHHHH}
   escapes, see Section 2.5 of the @|rfc|.
 
+  If @racket[format?] is given as @racket[#f], then @racket[indent] is
+  used to format JSON.
+
 @examples[#:eval ev
   (with-output-to-string
     (λ () (write-json #hasheq((waffle . (1 2 3))))))
   (with-output-to-string
     (λ () (write-json #hasheq((와플 . (1 2 3)))
                       #:encode 'all)))
+  (write-string
+   (with-output-to-string
+     (λ () (write-json #hasheq((waffle . (1 2 3))
+                               (와플 . (1 2 3)))))))
+  (write-string
+   (with-output-to-string
+     (λ () (write-json #hasheq((waffle . (1 2 3))
+                               (와플 . (1 2 3)))
+                       #:format? #t))))
+  (write-string
+   (with-output-to-string
+     (λ () (write-json #hasheq((waffle . (1 2 3))
+                               (와플 . (1 2 3)))
+                       #:format? #t
+                       #:indent "  "))))
 ]
 }
 
 @defproc[(jsexpr->string [x jsexpr?]
                          [#:null jsnull any/c (json-null)]
-                         [#:encode encode (or/c 'control 'all) 'control])
+                         [#:encode encode (or/c 'control 'all) 'control]
+                         [#:format? format? boolean? #f]
+                         [#:indent indent string? "\t"])
          string?]{
   Generates a JSON source string for the @tech{jsexpr} @racket[x].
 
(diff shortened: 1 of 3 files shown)


mfelleisen:
Change 1 seems to be backwards compatible. 

Change 2 would affect existing code. If you really think this is necessary, enable this behavior via a new keyword.

rfindler:
It looks like the newlines and indentation happen no matter the actual json object? Would it be possible/make sense to notice that things like `[1,2,3]` probably don't need newlines and indentation and not do them (ie do pretty printing)?

NoahStoryM:
> Change 2 would affect existing code. If you really think this is necessary, enable this behavior via a new keyword.

I'm stuck in a weird problem, which seems to be resolved after I do change2 . I'll do further research to see if this change is necessary.

> It looks like the newlines and indentation happen no matter the actual json object?  Would it be possible/make sense to notice that things like [1,2,3] probably don't need newlines and indentation and not do them (ie do pretty printing)?

I'm not sure what you mean, can you provide some examples?

mfelleisen:
After thinking about it, I thought that perhaps #2 could be solved with a parameter.

sorawee:
> I'm not sure what you mean, can you provide some examples?

Pretty printing consumes one more argument, which is the page width limit. It tries to make the output pretty by making it not too tall (e.g. minimize the height) and not too wide (minimize overflow over the page width limit).

Here's an example: let's say the json object is an array with 10000 elements. You probably want each item to be in separate lines. But in your printer, you hard code it so that they are always separated by spaces.

rfindler:
> 
> > It looks like the newlines and indentation happen no matter the actual json object?  Would it be possible/make sense to notice that things like [1,2,3] probably don't need newlines and indentation and not do them (ie do pretty printing)?
> 
> I'm not sure what you mean, can you provide some examples?

Beyond what @sorawee said, there is a pretty good literature on designing and implementing pretty-printers. It may also make sense to look at what other json printers do to see how (if?) they are clever.

I don't mean to turn this into mission creep, btw. If this isn't an interesting direction to pursue, the current pull request is certainly an improvement.

gus-massa:
About #2: I'm using the `json` package to open a file, filter part of entries in the `json` and save the result, so I preffer the `(write-json (read-json))` is as close to the identity as possible, so I can just add some function in between and be sure it will not mess anything in the parts I didn't change.

NoahStoryM:
> >  Change 2 would affect existing code. If you really think this is necessary, enable this behavior via a new keyword.
>
> I'm stuck in a weird problem, which seems to be resolved after I do change2 . I'll do further research to see if this change is necessary.

The problem I'm stuck in seems to be caused by some other factor, so I guess change2 isn't necessary, I'll undo it later.

rmculpepper:
I think it is a bad idea to add parameters that cause this library to behave in new, incompatible ways. (The existing `json-null` parameter was a design mistake, IMO.)

As someone who writes libraries that depend on the `json` library, I don't want my library's clients reconfiguring `json` so that it produces values that my library does not expect. It also increases the likelihood of bugs when a user sets a parameter in one thread but the JSON processing happens in a different thread; see the recent discussion on Discourse (https://racket.discourse.group/t/running-a-procedure-from-delay-does-not-capture-modified-parameters/936).

Making the new features available as ordinary function arguments is fine, and thank you for the effort. I just think that this use of parameters is a common Racket anti-pattern.

mfelleisen:
I get this argument and I can see why numeric formatting could be something that interferes with your server-client architecture. (Of course it could also be an indication that the composition is brittle. If you have such a separation, you must expect that eventually someone writes a different clien and injects bugs.) 


Ryan also writes "I just think that this use of parameters is a common Racket anti-pattern”. I am not sure I’d go that far but I see the principle behind the argument. Thanks for pointing this out.

NoahStoryM:
I can understand the downside of using parameters, do I need to remove `json-inf` and `json-mhash?` parameters and only enable these behaviors via keywords?

samth:
Yes, I think we should not introduce new parameters to this interface (it's also a big problem for Typed Racket).

NoahStoryM:
Considering Typed Racket , should I allow `read-json` to read `+inf.0` and `-inf.0`?

samth:
I'm not sure what you mean. Are those in the grammar for JSON?

NoahStoryM:
In the current json-lib:

```racket
Welcome to Racket v8.5 [cs].
> (require json)
> (string->jsexpr "1e500")
+inf.0
> (jsexpr->string (string->jsexpr "1e500"))
"+inf.0"
> (string->jsexpr (jsexpr->string (string->jsexpr "1e500")))
string::1: string->jsexpr: bad input starting #"+inf.0" [,bt for context]
> (string->jsexpr "1e500" #:inf+ "+inf")
"+inf"
> (jsexpr->string (string->jsexpr "1e500" #:inf+ "+inf"))
"\"+inf\""
> (string->jsexpr (jsexpr->string (string->jsexpr "1e500" #:inf+ "inf")))
"inf"
```

If Typed Racket use only `+inf.0` and `-inf.0` to represent infinity, then `(string->jsexpr (jsexpr->string (string->jsexpr "1e500")))` is illegal.

samth:
I don't think configuring the behavior of JSON parsing in this way is a good idea. Specifically for Typed Racket, allowing `#:inf+` to specify non-`jsexpr?` values is problematic.

sorawee:
I think the PR is really getting too large. Wouldn't it be better to separate the PR into two (or more parts), where one only adds `#:format?` and `#:indent`, and another deals with infinity, etc. They seem to be orthogonal and should be reviewed separately.

NoahStoryM:
It sounds good. I'll separate this pr into 3 parts later.

PR #4304, opened 2022-06-16
match: nonlinear pat refs in list-no-order

AlexKnauth:
Allows patterns inside `list-no-order` to refer to nonlinear patterns declared prior, such as in `(cons a (list-no-order a rst ...))`.

Includes nonlinear patterns declared under ellipses, such as in `(list (cons a (list-no-order a rst ...)) ...)`.

This also makes nonlinear pattern declarations syntax errors if the first use is in the scope of a `...` pattern that doesn't encompass all uses.

### DIFF
--- a/racket/collects/racket/match/compiler.rkt
+++ b/racket/collects/racket/match/compiler.rkt
@@ -560,10 +560,9 @@
      (lambda (id)
        (if (identifier? id)
            id
-           (begin
-             (log-error "non-linear pattern used in `match` with ... at ~a and ~a"
-                        (car id) (cadr id))
-             #f)))]
+           (raise-syntax-error #f
+             "non-linear pattern used in `match` with ..."
+             (car id) (cadr id))))]
     [else #f]))
 
 ;; generate-temporaries/seen :
--- a/racket/collects/racket/match/compiler.rkt
+++ b/racket/collects/racket/match/compiler.rkt
@@ -228,7 +228,7 @@
                    [else (make-Row ps
                                    #`(let ([#,v* #,x]) #,(Row-rhs row))
                                    (Row-unmatch row)
-                                   (cons (cons v* x) (Row-vars-seen row)))])]))])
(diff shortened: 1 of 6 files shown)


racket/collects/racket/match/compiler.rkt
```diff
@@ -551,5 +543,27 @@
...
+  (cond
+    [(for/or ([e (in-list seen)])
+       (let ([v* (car e)] [id (cdr e)])
+         (and (bound-identifier=? v v*) (or id (list v v*)))))
+     =>
+     (lambda (id)
+       (if (identifier? id)
+           id
+           (begin
+             (log-error "non-linear pattern used in `match` with ... at ~a and ~a"
```
samth, line 557: If this is supposed to work now this log should be removed.

AlexKnauth: I didn't want things like this to change for now:
```racket
(match '(1 2 3 4)
  [(list a ... a) a])
```
Order matters. When the `a` without ellipses comes first, that's when it works as a nonlinear pattern:
```racket
(match '(1 1 1 1)
  [(list a a ...) a])
```

AlexKnauth:
I have an idea for checking the problems that lead to warnings:

Track ellipsis depth, and when merging pattern-variables together into non-linear patterns, only accept them when the ellipsis-depth of the first one, the declaration, is less than or equal to the ellipsis-depth of the later ones, the references.

So `(list a a ...)` would be allowed because the declaration ellipsis-depth 0 is less than the reference ellipsis-depth 1, but `(list a ... a)` would be an error because the declaration ellipsis-depth 1 is greater than the reference ellipsis-depth 0.

racket/collects/racket/match/compiler.rkt
AlexKnauth: I've updated the PR, it now makes nonlinear pattern declarations syntax errors if the first use is in the scope of a `...` pattern that doesn't encompass all uses.

samth:
Closing since #5467 changes things in a different direction.

PR #4312, opened 2022-06-19
bit-vector: check for out of range error in bit-vector-set!

### LINKED ISSUE #4311: bit-vector: no out of bound check for bit-vector-set!
8.5 [cs]

**What program did you run?**

```
#lang racket
(require data/bit-vector)
(define b (bit-vector #f #f #f #f #f #f #f #f #f))
(bit-vector-set! b 15 #t)
(bit-vector->list b)
```

A contract error, indicating that 15 is out of bound

N/A

N/A

sorawee:
Fixes #4311


sorawee:
Oh wow, the tests locate in another repo... This doesn't seem ideal either, but I will follow that since it's already done.

samth:
It's basically the same situation as the `db` collection -- some parts of the `data` collection are used by core Racket, so that source code is here. It might be that we should merge that repository back here, but we haven't done that.

sorawee:
I'm surprised that the CI didn't catch this issue earlier. Anyhow, this is now fixed.

samth:
I think that a conflict like that is not an error (as in, does not cause something to exit with result 1).

PR #4315, opened 2022-06-21
ci: run arm linux on schedule

sorawee:
Due to low availability of the machine, we can't afford to run this workflow on every push. Instead, run it on schedule (every day).


sorawee:
Closing, since the workflow was deleted in #4390

PR #4319, opened 2022-06-25
doc: add more fast-seq

sorawee:


samth:
I assume the implementations of those already exist, right?

sorawee:
Yes. https://docs.racket-lang.org/reference/sequences.html#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._in-inclusive-range%29%29 and https://docs.racket-lang.org/reference/sequences.html#%28def._%28%28lib._racket%2Fprivate%2Fbase..rkt%29._in-mlist%29%29

Issue #4286, opened 2022-06-06
Zlib for `racket/draw` affected by CVE-2018-25032


LiberalArtist:
The following packages in https://github.com/racket/libs (which has issues disabled) include versions of Zlib affected by CVE-2018-25032:

- `draw-x86_64-linux-natipkg-2`
- `draw-x86_64-linux-natipkg-3`
- `draw-win32-x86_64-2`
- `draw-win32-x86_64-3`
- `draw-win32-i386-2`
- `draw-win32-i386-3`

samth:
This code from Chromium has something that could be the start of a test case: https://chromium-review.googlesource.com/c/chromium/src/+/3550971/2/third_party/zlib/contrib/tests/utils_unittest.cc

samth:
Just to be clear on what's at issue here:

1. It is very likely that this code can be reached from pure Racket code, or even from user input (in the form of compressed data) in a Racket application.
2. Almost certainly the `-2` versions of these libraries do not have an update for this bug. However, the Zlib interface has not changed much and so changing those to a more modern version is unlikely to break even very old versions of Racket.

For the `-3` versions, it seems like we should pack new versions now. For the `-2` versions, I would suggest doing the same.

mflatt:
I've updated the `-3` packages with zlib 1.2.12

PR #4327, opened 2022-06-28
chore: Included githubactions in the dependabot config

naveensrinivasan:
This should help with keeping the GitHub actions updated on new releases. This will also help with keeping it secure.

Dependabot helps in keeping the supply chain secure https://docs.github.com/en/code-security/dependabot

GitHub actions up to date https://docs.github.com/en/code-security/dependabot/working-with-dependabot/keeping-your-actions-up-to-date-with-dependabot

https://github.com/ossf/scorecard/blob/main/docs/checks.md#dependency-update-tool
Signed-off-by: naveen <172697+naveensrinivasan@users.noreply.github.com>

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "github-actions"
+    directory: "/"
+    schedule:
+      interval: "weekly"


LiberalArtist:
Hi! I don't know much about Dependabot, but it would be really great if it could help with checking for issues like https://github.com/racket/racket/issues/4286 with the shared libraries we distribute: would you happen to know if that would be possible?

naveensrinivasan:
> Hi! I don't know much about Dependabot, but it would be really great if it could help with checking for issues like #4286 with the shared libraries we distribute: would you happen to know if that would be possible?

Dependabot supports other lang ecosystems. TBH without some kind of SBOM it would be hard to identify vulns.

As long as you have a commit SHA and repo then it can be queried from https://osv.dev/

sorawee:
Thanks!

PR #4328, opened 2022-06-28
pkg: don't cache if use-cache? is false

sorawee:
When use-cache? is false, we also should not cache the file.
This significantly improves the performance since file/gzip is quite slow.


mflatt:
Thanks!

PR #4329, opened 2022-06-28
zuo-doc typos

casaca24:
opened another pull request, couldn't figure how or if it's possible to edit the first one.


casaca24:
don't know if that merge commit in between is an issue. please, let me know.
go racket go!

mflatt:
Thanks! We do try to avoid merge commits, so as long as it's ok with you, I'll squash and merge your two commits outside of GitHub.

PR #4339, opened 2022-07-03
Allow comment in the hash literal

### LINKED ISSUE #4333: reader: comment is not handled correctly in #hash literal
8.5 [cs]

**What program did you run?**

```
#lang racket
      
#hash((key . ;; comment
         ))
```

and

```
#lang racket
      
#hash((key . 1
           ;; comment
         ))
```

The first program should error. The second program should evaluate to `#hash((key . 1))`. In particular, the documentation states that by default, special-comment should be treated as whitespace.

> The default port read handler reads standard Racket expressions with Racket’s built-in parser (see [The Reader](https://docs.racket-lang.org/reference/reader.html)). It handles a special result from a custom input port (see make-custom-input-port) by treating it as a single expression, except that special-comment values (see [Special Comments](https://docs.racket-lang.org/reference/special-comments.html)) are treated as whitespace.

Instead, the first program results in:

```
'#hash((key . #<special-comment>))
```

and the second program errors with:

```
read-syntax: expected `)` after value within a hash
```

N/A

N/A

sorawee:
Fixes #4333


racket/src/expander/read/hash.rkt
```diff
@@ -168,29 +168,48 @@
...
-    
-    (unless (eqv? closer-ec closer)
-      (reader-error in (reading-at config closer-line closer-col closer-pos)
-                    #:due-to closer-c
-                    "expected ~a after value within a hash"
-                    (closer-name closer config)))
-    
-    (cons (coerce-key k elem-config) v)]))
+    (let step-read-key ()
+      (define k (read-one #f in (disable-wrapping elem-config)))
```
sorawee, line 172: It's unclear to me if I need to use `(keep-comment ...)` here.

mflatt:
Thanks for the tests and repair, but I think this is not quite the right repair. The `read-char/skip-whitespace-and-comments` function should already discard comments; the problem is that `read-unwrapped-sequence` has turned on comment preservation to handle them at the level of the parentheses for `#hash(....)`. I think the right fix is for the function returned by `make-read-one-key+value` to adjust the given `config` to discard comments, since it handles them more directly.

sorawee:
Note that the file originally (and also currently) has this

```
;; If it's a special or we have a readtable, we need to read ahead
;; to make sure that it's not a comment. For consistency, always
;; read ahead.
      (define v (read-one c in (keep-comment elem-config)))
      (cond
       [(special-comment? v)
;; Try again
```

If reading the opener can't result in a special comment, wouldn't this be a deadcode? Should it be removed then?

(This is the reason I structured the PR the way it currently is; I followed how it's done here)

sorawee:
The question is this, where should we expect `*` to work, in the following test:

```
(test #hash((apple . (red round))
            (banana . (yellow long)))
      values
      (parameterize ([current-readtable
                      (make-readtable #f
                                      #\! 'terminating-macro (lambda (ch port . args)
                                                               (read/recursive port))
                                      #\* 'terminating-macro (lambda args
                                                               (make-special-comment #f)))])
        (read (open-input-string
               "!#hash((apple . (red round)) * (banana . (yellow long)))"))))
```

It works in the above test because we are currently specifically guarding against that possibility. However, the following test would fail:

```
(test #hash((apple . (red round))
            (banana . (yellow long)))
      values
      (parameterize ([current-readtable
                      (make-readtable #f
                                      #\! 'terminating-macro (lambda (ch port . args)
                                                               (read/recursive port))
                                      #\* 'terminating-macro (lambda args
                                                               (make-special-comment #f)))])
        (read (open-input-string
               "!#hash((apple . (red round)) (banana * . (yellow long)))"))))
```

Do you want to allow this possibility?

PR #4344, opened 2022-07-05
expander: remove dead code

sorawee:
1aa1cbbdad7 adjusts so that comments are skipped,
so special comment handling here is now a dead code.


sorawee:
Oh wait, looks like this is somehow needed after all. I tested wrong code.

PR #4349, opened 2022-07-08
improve errors for yield outside generator

benknoble:
The spec requires that `yield` be variable arity, but the default
`yield` outside of `generator` forms accepts only a single argument,
leading to:

```
> ,r racket/generator
> (yield)
...acket/generator.rkt:47:3: arity mismatch;
  expected: 1
  given: 0
 [,bt for context]
> (yield 1) ;; this is what I expect for the others, as it's more clear about the problem
yield: must be called in the context of a generator [,bt for context]
> (yield 2 3)
...acket/generator.rkt:47:3: arity mismatch;
  expected: 1
  given: 2
 [,bt for context]
```

The expression `(yield 1)` gives the best error message; now, the others
should give the same, and the behavior is tested.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/generator.rktl
+++ b/pkgs/racket-test-core/tests/racket/generator.rktl
@@ -4,11 +4,24 @@
 (Section 'generator)
 
 (require racket/generator
+         (only-in racket/string string-prefix?)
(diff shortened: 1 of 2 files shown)


samth:
Thanks!

PR #4350, opened 2022-07-08
grapheme clusters and port position counting

mflatt:
Changes:

 * upgrades from Unicode 7.0 to Unicode 14.0; none of the old Racket tests fail after upgrading
 * adds support for grapheme-cluster parsing in strings via `string-grapheme-length` (which finds the length of a a grapheme at a starting index) and `string-grapheme-count` (which conceptually iterates `string-grapheme-length` to count the total number of graphemes)
 * adds support for streaming grapheme-cluster parsing via `char-grapheme-step`
 * changes built-in column and position counting via `port-count-lines!` to count grapheme clusters instead of characters
 * exposes `char-extended-pictographic?` and `char-extend?`, which are used internally for `char-grapheme-step` (which is used internally for `string-grapheme-length`)

This might be all the grapheme support that we want built into core Racket. Building this much in is necessary for port counting to use graphemes, to take advantage of the core infrastructure for character properties, and to take advantage of internal shortcuts for performance. So far, other grapheme-based operations seem just as well implemented in libraries.

Compatibility
-------------

Changing port counting from characters to graphemes is not backward compatible. The difference shows up, for example, when an input stream has a Unicode modifier like U+300 (maybe after `e`, instead of the single Unicode character U+E8 `è`) or a multi-character emoji (such as a pirate flag `🏴‍☠️`).

Concretely, this change is likely to make locations in syntax errors consistent with editors other than DrRacket. Meanwhile, the intent is to make further changes that bring DrRacket in line with grapheme counting and drawing.

An alternative to this change would be to add grapheme counting as a separate options, which would be straightforward internally. However, `make-input-port` and `make-output-port` would have to change to support the new option, and uses of those functions would have to be updated to support grapheme counting. If grapheme counting is what we practically always want, it seems better to just fix the sources (such as editors) and sink (built-in counting) than to make everything in between work in both modes. If a port implementation performs its own counting, then it would need to be updated to use `char-grapheme-step`, but port implementations typically rely on built-in counting.

Performance
-------------

See the commit messages for measurements. The short version is that the cost of grapheme counting is measurable but similar in scale to the current cost of counting. Maybe future improvements could reduce the cost, but it already seems fast enough for uses like parsing with locations in errors.

Next Steps
-----------

A likely next step is to make `draw-text` and `text-extent` in `dc<%>` support drawing by graphemes. Then, `text%` can be updated to work in terms of graphemes instead of characters. That's still not a completely right approach for editing, because it doesn't deal with ligatures and right-to-left text, but it gets a lot closer than the current editor. (The remaining gap is not about position counting, where grapheme clusters seem like the right choice, but about drawing and editor actions).

Shrubbery notation probably should define indentation in terms of graphemes instead of characters. In fact, the current shrubbery parser depends on port-based counting, so the set of commits here implies that change.


sorawee:
I'm worried about the performance. The current `port-count-lines!` is already too slow for some applications. See https://github.com/racket/racket/pull/3274#issuecomment-652682835 and the original mailing list thread https://www.mail-archive.com/racket-users@googlegroups.com/msg44574.html for how the current `port-count-lines!` makes `read-xml` slow. Can we have a benchmark on some test programs?

samth:
@sorawee there are some benchmarks in the commit messages; I think the difference is going to be small and line counting is unlikely to become desired in those applications whatever we do.

gus-massa:
I'm not sure if it was already discussed or if it's the standard name but I think that `string-grapheme-length` is confusing because it's too similar to `string-length`. Perhaps my wrong interpretation should be called `string-length/grapheme-aware-version` or something like that.

mflatt:
I've wondered about that `length`, too. Is `string-grapheme-span`a better name for the function currently called `string-grapheme-length`?

samth:
What about `string-grapheme-count`?

mflatt:
I think `string-grapheme-count` is better used as currently, which is the function that takes a string and tells you how many grapheme clusters it contains.

samth:
Ah, I understand better now. Maybe `-size`?

mflatt:
For text in Racket, "size" usually refers to the graphical extent when drawing. Meanwhile, "span" at least has a similar meaning in `srcloc-span` to the meaning needed here. So, I like "span" better here than "size". Otherwise, I haven't found an existing function that's similar enough to suggest a convention.

mflatt:
As reflected in racket/gui#277, the idea of switching editor items from characters to graphemes did not work out, but adding a grapheme counting layer there gives up a path forward. Overall, I'm convinced that this is a workable direction and a reasonable compromise, and I'm planning to merge this afternoon.

mflatt:
Merged

PR #4351, opened 2022-07-08
Provide `racket/symbol` from `racket`

otherjoel:
The function `symbol->immutable-string` was moved out of `racket/base` in 37ce9478cd, but was not provided from `racket` at that time, even though the doc edits in that commit indicate that the intent was to do so.

Obviously if the preferred fix is to amend the docs, please feel free to close this PR.

### DIFF
--- a/racket/collects/racket/main.rkt
+++ b/racket/collects/racket/main.rkt
@@ -15,6 +15,7 @@
          racket/list
          racket/vector
          racket/string
+         racket/symbol
          racket/bytes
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks for the PR, but I think I probably just edited the docs in the wrong way. As far as I can remember, the intent was to leave the function out of `racket` as well as `racket/base`, although I don't remember whether the conflict I saw was via `racket` or `racket/base`.

PR #4354, opened 2022-07-11
Mark generated files for github

samth:
This will cause them to be hidden by default in GitHub diff view.

Also, indicate that .zuo files should be highlighted like Racket.

Other generated files could be added here too.


sorawee:
This looks good to me. Is there anything else to be done?

PR #4358, opened 2022-07-14
doc: clarify the explanation for `-j` for `raco test`

capfredf:


capfredf:
@samth can you take a look and merge it if it looks good to you?

PR #4361, opened 2022-07-17
Parameterization for dependent function contracts

camoy:
For some time now, `impersonate-procedure` and `chaperone-procedure` have supported installing a continuation mark during the dynamic extent of a wrapped procedure. While this facility is available for custom contracts, none of Racket's existing contract combinators (as far as I know) provide access to this mechanism, even though it's a useful one. A few examples:

* `yield` can only be called during the dynamic extent of a generator
* A promise cannot be forced again while it's already being forced.
* `contract-random-generate-get-current-environment` can only be called during contract generation.
* `call-with-escape-continuation` can only be called during `call-with-escape-continuation`.
* [Authorization Contracts](https://dl.acm.org/doi/10.1145/2983990.2984021)

This PR adds a `#:param` option to `->i` that sets a parameter during the protected function's execution. I chose to implement this as an option to `->i` so that the parameter's value can depend on the arguments passed to the function. I can't think of another implementation strategy that could support this. Only contracts that use `#:param` will add any code relating to parameters, so there should be no adverse performance costs to existing code. This PR doesn't include documentation, but I'll be happy to write some after getting feedback.

Here is a larger example showing it in action on a nested generator where yielded values must satisfy some predicate. The contracts are straightforward and have reasonable error messages thanks to `#:pre/desc`:

```rkt
(module generator-lib racket
  (provide
   (contract-out
    [generator
     (->i ([p predicate/c]
           [f (p) (->i ()
                       #:param ()
                       currently-generating?
                       (cons p (currently-generating?))
                       [r any/c])])
          [r generator?])]

    [yield (->i ([x any/c])
                #:pre/desc ()
                (or (cons? (currently-generating?))
                    "yield called outside a generator")
                #:pre/desc (x)
                (or ((car (currently-generating?)) x)
                    "yielded value does not satisfy generator predicate")
                [r any/c])]))

  (define currently-generating? (make-parameter null))
  (define generator-prompt-tag (make-continuation-prompt-tag))

  (struct generator (pred proc) #:mutable
    #:property prop:procedure
    (λ (self)
      (call-with-continuation-prompt
       (λ ()
         ((generator-proc self)))
       generator-prompt-tag
       (λ (v kont)
         (set-generator-proc! self (λ () (kont (void))))
         v))))

  (define (yield v)
    (let/cc kont
      (abort-current-continuation generator-prompt-tag v kont))))

;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;

(require 'generator-lib)

;; A generator where the yielded value must satisfy `integer?`
(define g1
  (generator
   integer?
   (λ ()
     (for ([i 3])
       (yield i)))))

;; A generator where the yielded value must satisfy `string?`
(define g2
  (generator
   string?
   (λ ()
     (for ([_ 3])
       (yield (format "hello ~a" (g1)))))))

;; E.g.
(g2)
(g2)
(g2)
```

cc @rfindler @chrdimo @mfelleisen

### DIFF
--- a/racket/collects/racket/contract/private/arr-i.rkt
+++ b/racket/collects/racket/contract/private/arr-i.rkt
@@ -9,6 +9,9 @@
          "generate.rkt"
          syntax/location
          racket/private/performance-hint
+         (only-in '#%paramz
+                  extend-parameterization
+                  parameterization-key)
          (for-syntax racket/base
                      racket/stxparam-exptime
                      syntax/name
@@ -98,6 +101,7 @@
          blame swapped-blame ;; used by the #:pre and #:post checking
          (append blames
                  (->i-pre/post-procs ctc)
+                 (->i-param-procs ctc)
                  partial-doms
                  (->i-arg-dep-ctcs ctc)
                  partial-indy-doms
@@ -190,7 +194,7 @@
 (struct ->i (blame-info
              arg-ctcs arg-dep-ctcs indy-arg-ctcs
              rng-ctcs rng-dep-ctcs indy-rng-ctcs
-             pre/post-procs
+             pre/post-procs param-procs
              mandatory-args opt-args mandatory-kwds opt-kwds rest
              mtd? here mk-wrapper name-info)
         #:property prop:custom-write custom-write-property-proc)
@@ -238,8 +242,9 @@
                    [args-info (vector-ref name-info 0)]
                    [rest-info (vector-ref name-info 1)]
                    [pre-infos  (vector-ref name-info 2)]
-                   [rng-info  (vector-ref name-info 3)]
-                   [post-infos (vector-ref name-info 4)])
+                   [param-infos  (vector-ref name-info 3)]
+                   [rng-info  (vector-ref name-info 4)]
+                   [post-infos (vector-ref name-info 5)])
               `(->i ,(arg/ress->spec args-info
                                      (map ->i-arg-contract (->i-arg-ctcs ctc))
                                      (->i-arg-dep-ctcs ctc)
@@ -276,6 +281,13 @@
                             `(#:pre ,ids ,code)]
                            [(equal? name 'desc)
                             `(#:pre/desc ,ids ,code)])))
+                    ,@(apply
+                       append
+                       (for/list ([param-info param-infos])
+                         (define ids (list-ref param-info 0))
+                         (define pexp (list-ref param-info 1))
+                         (define vexp (list-ref param-info 2))
+                         `(#:param ,ids ,pexp ,vexp)))
                     ,(cond
                        [(not rng-info)
                         'any]
@@ -323,14 +335,14 @@
 (define (make-->i is-chaperone-contract? blame-info
                   arg-ctcs arg-dep-ctcs indy-arg-ctcs
                   rng-ctcs rng-dep-ctcs indy-rng-ctcs
-                  pre/post-procs
+                  pre/post-procs param-procs
                   mandatory-args opt-args mandatory-kwds opt-kwds rest
                   mtd? here mk-wrapper name-info)
   (define maker (if is-chaperone-contract? chaperone->i impersonator->i))
   (maker blame-info
          arg-ctcs arg-dep-ctcs indy-arg-ctcs
          rng-ctcs rng-dep-ctcs indy-rng-ctcs
-         pre/post-procs
+         pre/post-procs param-procs
          mandatory-args opt-args mandatory-kwds opt-kwds rest
          mtd? here mk-wrapper name-info))
 
@@ -358,18 +370,18 @@ evaluted left-to-right.)
 
 
   ;; set up some unreferred to variables for
-  ;; the pre/post conditions to base the graph on
-  ;; get-var : (or/c pre/post arg) -> identifier
+  ;; the pre/post conditions and params to base the graph on
+  ;; get-var : (or/c pre/post? param? arg) -> identifier
   ;; (unfortuntately we rely on `eq?` here)
-  (define pre/post-fake-vars (make-hasheq))
+  (define fake-vars (make-hasheq))
   (for ([arg (in-list args)]
-        #:when (pre/post? arg))
-    (hash-set! pre/post-fake-vars arg
+        #:when (or (pre/post? arg) (param? arg)))
+    (hash-set! fake-vars arg
                (car (generate-temporaries (list arg)))))
   (define (get-var arg)
     (if (arg/res? arg)
         (arg/res-var arg)
-        (hash-ref pre/post-fake-vars arg)))
+        (hash-ref fake-vars arg)))
 
   ;; track the indices into `args` for the nodes in the graph
   ;; and do the same thing but only for the subset that are actually args
@@ -381,13 +393,20 @@ evaluted left-to-right.)
     (hash-set! numbers arg i)
     (free-identifier-mapping-put! id->arg/res (get-var arg) arg))
 
-  ;; track the original order of the pre/post conditions
+  ;; track the original order of the pre/post conditions and params
   (define pre/post-numbers (make-hasheq))
-  (let ([i 0])
-    (for ([arg (in-list args)])
-      (when (pre/post? arg)
-        (hash-set! pre/post-numbers arg i)
-        (set! i (+ i 1)))))
+  (define param-numbers (make-hasheq))
+
+  (define (count! pred? ht)
+    (define i 0)
+    (for ([arg (in-list args)]
+          #:when (pred? arg))
+      (hash-set! ht arg i)
+      (set! i (+ i 1))))
+
+  (count! pre/post? pre/post-numbers)
+  (count! param? param-numbers)
+
   ;; build the graph, where `comes-before` are the backwards
   ;; edges and `comes-after` are the forwards edges
   ;; we use new temporary variables for the pre/posts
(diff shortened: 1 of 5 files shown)


racket/collects/racket/contract/private/arr-i.rkt
```diff
@@ -583,18 +613,12 @@ evaluted left-to-right.)
...
-       (apply values res-checker
-              (append (filter (λ (x) (not (eq? x the-unsupplied-arg))) args) rest-args))]
-      [(null? supplied-kwd-args)
-       (apply values (append (filter (λ (x) (not (eq? x the-unsupplied-arg))) args) rest-args))]
-      [res-checker
-       (apply values res-checker supplied-kwd-args
-              (append (filter (λ (x) (not (eq? x the-unsupplied-arg))) args) rest-args))]
-      [else
-       (apply values supplied-kwd-args
-              (append (filter (λ (x) (not (eq? x the-unsupplied-arg))) args) rest-args))])))
```
rfindler, line 597: I'm not sure why I didn't write the simpler thing using the append here, but I'm skeptical of removing it.

racket/collects/racket/contract/private/arr-i.rkt
```diff
@@ -545,6 +573,7 @@ evaluted left-to-right.)
          (for/list ([arg (in-list args)])
            (hash-ref wrapper-args arg)))
        #`(return/no-unsupplied #,(if (null? result-checkers) #f (car result-checkers))
+                               (list #,@param-mark)
                                #,(if rst
                                      #'rest-args
                                      #''())
```
rfindler, line 579: I don't know the performance implication of these additional lists (probably low) but this isn't no change to the generated code, right?

racket/collects/racket/contract/private/arr-i.rkt
```diff
@@ -1066,22 +1126,27 @@ evaluted left-to-right.)
                          [i (in-naturals)])
                 (string->symbol (format "post-proc~a" i)))
 
+            ;; the param procs
```
rfindler, line 1129: I'm not sure this comment is helping 😄 . The comments in the ones below seem like they provide some value to me, but this one really just repeats what's two lines below.

rfindler:
Can you say a little bit more about what this change is actually doing? In particular, from your description I expected to see `impersonator-prop:application-mark` in the diff, but I don't see it. Also, I don't see any explanation of how we go from marks to parameters. And just a few hints about the new syntax would be welcome.

camoy:
`impersonator-prop:application-mark` is one way of doing it. Instead, I'm using this feature of `impersonate-procedure`:

> any number of repetitions of 'mark key val (i.e., three values), where the call proc is wrapped to install a continuation mark key and val.

The advantage (well I'm not sure this is useful, but it's at least more flexible) is that the parameter itself isn't fixed. So, something silly like

```rkt
(define a (make-parameter #f))
(define b (make-parameter #f))

(define/contract (f x)
  (->i ([x boolean?])
       #:param (x) (if x a b) #t
       any)
  (list (a) (b)))
```

is possible, where the parameter that you want to change is itself dependent.

***

The crux of the mark to parameter code is in [`build-param-code`](https://github.com/racket/racket/pull/4361/files#diff-a1a913f821f46f019b1d8b7d56d3c1cbba19fef4f89a0a7e7b37151544ff3f06R699). This uses `extend-parameterization` from `#%paramz` to assign a new value for the parameter. That value is then installed with the internal `parameterization-key` continuation mark key [here](https://github.com/racket/racket/pull/4361/files#diff-a1a913f821f46f019b1d8b7d56d3c1cbba19fef4f89a0a7e7b37151544ff3f06R555). As far as I know, `racket/base` doesn't expose any way to directly manipulate parameters via continuation marks.

***

Sure, so the proposed syntax is

```
param-options ::=
  | #:param (id ...) param-expr val-expr param-options
```
where `param-options` goes right after the `#:pre` options.

racket/collects/racket/contract/private/arr-i.rkt
camoy: Ah yes, that's a good catch. An easy solution is to create a variant, `return/no-unsupplied*`, with the extra argument and then expand to that version when needed. I could probably abstract over both versions with a macro to avoid code duplication. Does that sound reasonable?

rfindler: I haven't been using macros for abstraction like that in this code so I'm unsure about that aspect.

As for the name, the words in the name are intended to say what case it is, so probably another slash with another word is better than a star suffix.

rfindler:
Thanks for the answers. This makes sense to me.

I understand why marks are more low-level than parameters -- is that the reason you're exposing a parameter-based api at the `->i` level instead of a mark-based one? Or is there another reason? Does it make sense to expose a mark-based API?

camoy:
Yes, that's exactly the reason. I figure most Racket programmers are familiar with parameters, but probably not continuation marks, so a parameter-based API makes sense to me. We could also support continuation marks, but that would clutter up the interface and I don't think is any more useful. (If one really wants continuation marks and not parameters, you can always write a custom contract.)

Thanks for the quick feedback. I've updated the PR based on your suggestions and also added documentation.

rfindler:
I think there needs to be a `parameter?` check somewhere (don't generate the code in the error case, just make it call a helper function that makes a good error message)

camoy:
You're right that the error message isn't great, but it's actually the same error that you get from `parameterize` itself (in terms of `extend-parameterization`):

```rkt
> (parameterize ([#t #t]) #t)
; extend-parameterization: contract violation
;   expected: parameter?
;   given: #t
```

The reason I didn't do this initially is I figured the redundant `parameter?` check per function call isn't worth the nicer error message. What do you think?

rfindler:
It is 100% worth it, IMO. When I use ->i I am not using
extend-parameterization!

Robby

camoy:
Ok, sounds good. [Here](https://github.com/racket/racket/pull/4361/files#diff-a1a913f821f46f019b1d8b7d56d3c1cbba19fef4f89a0a7e7b37151544ff3f06R760) is the `parameter?` check.

rfindler:
I think there is a way to include #:param in the message?

camoy:
Funny, `raise-argument-error` supports specifying a positional argument in the error message, but not a keyword argument. So, I've just followed what `dynamic->*` does and made the expected string "parameter? in the #:param argument."

LiberalArtist:
There are some notable differences between parameters and continuation marks, e.g. imperative mutation and the interaction with `thread`. I don't know if those differences matter here or not, I'm just generally wary of unforeseen implications of parameters. (I gave some examples [here](https://racket.discourse.group/t/running-a-procedure-from-delay-does-not-capture-modified-parameters/936/10?u=liberalartist).)

In particular, I'm not sure whether:

> The crux of the mark to parameter code is in [`build-param-code`](https://github.com/racket/racket/pull/4361/files#diff-a1a913f821f46f019b1d8b7d56d3c1cbba19fef4f89a0a7e7b37151544ff3f06R699). This uses `extend-parameterization` from `#%paramz` to assign a new value for the parameter. That value is then installed with the internal `parameterization-key` continuation mark key [here](https://github.com/racket/racket/pull/4361/files#diff-a1a913f821f46f019b1d8b7d56d3c1cbba19fef4f89a0a7e7b37151544ff3f06R555). As far as I know, `racket/base` doesn't expose any way to directly manipulate parameters via continuation marks.

means that this would change any assumptions that currently hold about the way parameters can be manipulated by safe/public Racket.

rfindler:
@LiberalArtist I think the code in this PR is doing the same thing as with-parameterization does. Here's the expansion of `(with-parameter ([x y]) z)`:

```
(with-continuation-mark
 parameterization-key
 (#%app
  extend-parameterization
  (#%app continuation-mark-set-first '#f parameterization-key)
  x
  y)
 z)
```

That said, someone thinking critically about whether that's true is welcome!

camoy:
> that this would change any assumptions that currently hold about the way parameters can be manipulated by safe/public Racket.

Indeed, as @rfindler said, this should not be the case. The only reason I've used these low-level facilities is because we want the procedure to be a chaperone/impersonator of the original value. The intent it to produce a wrapper like:

```rkt
(lambda args
  (parameterize ([p v] ...)
    (apply f args)))
```

So, the continuation mark manipulation just emulates that. The low-level API should not leak to clients in any way.

camoy:
@rfindler anything else that you want me to do before this PR can be merged?

rfindler:
Just looking over the code a little bit, I'm okay with this. If you were to do a sanity check to make sure that this expression doesn't change before/after the PR

```
#lang racket
(syntax->datum
 (expand-once
  #'(->i ([x integer?]) any)))
```

and also a few others that you know will trigger executing code paths that you touched, based on the changes, that would be great. I can merge after that and the documentation tweak I suggested.

rfindler:
Actually, I just realized that this might introduce some pain in the parser for future extensions. What do you think about making people write `#:para-value` each time they add one, instead of having it just take all expressions afterward?

camoy:
Maybe I'm not understanding you correctly, but I think it's already like that. You have a `#:param` for each one. Here's one of the [unit tests](https://github.com/racket/racket/blob/48005791b38dbe8390d2bcb5c4329f01b77edc18/pkgs/racket-test/tests/racket/contract/arrow-i.rkt#L1310).

rfindler:
Oh, I'm sorry. I was confused.

camoy:
So I did the sanity checks you asked and other than the additional (empty) argument to `make-->i` there's no changes to the generated code. In my last push I changed the wording of the documentation and also added a history note.

Would you prefer the keyword to be `#:param-value` or is just `#:param` alright?

rfindler:
I think `#:param` is better. Once John has created the release branches, let's push this to master (so we have a full release cycle to react to unanticipated problems).

camoy:
Sounds good, I've updated the version number accordingly. Thanks for your help!

rfindler:
Thanks you!

Issue #4357, opened 2022-07-13
Initializing `docindex.sqlite` from a read-only layer fails


LiberalArtist:
When initializing a layer that chains to a parent layer with documentation, `raco setup` fails with a string of errors like this:
```
raco setup: --- building documentation ---                         [20:58:52]
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/scribble.css
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/scribble-style.css
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/racket.css
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/manual-style.css
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/manual-racket.css
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/manual-racket.js
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/manual-fonts.css
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/scribble-common.js
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/doc-site.css
raco setup: installing: /gnu/store/yn0bhwrixxnhplvzd094m61681055xxl-racket-8.5+video0.2.3-1.3c69669/share/doc/racket/doc-site.js
raco setup: 4 running: <pkgs>/racket-index/scribblings/main/start.scrbl
raco setup: 2 running: <pkgs>/racket-index/scribblings/main/local-redirect.scrbl
raco setup: 3 running: <pkgs>/racket-index/scribblings/main/search.scrbl
[...]

PR #4364, opened 2022-07-18
racket-index: set write permission when copying `docindex.sqlite`

LiberalArtist:
Fixes https://github.com/racket/racket/issues/4357


mflatt:
Thanks!

PR #4374, opened 2022-07-26
Typo etc fixes to engine documentation

jasonhemann:
I noticed a couple of missing articles, and one place where the documentation incorrectly leaves out one kind of suspension condition; in the latter case I just stay generic, the suspension conditions are explained above.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/engine.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/engine.scrbl
@@ -13,7 +13,7 @@ inspired by the work of Haynes and Friedman @cite["Haynes84"].
 Engines log their behavior via a logger with the name 
 @racket['racket/engine]. The logger is created when the module
 is instantiated and uses the result of @racket[(current-logger)]
-as its parent. The library adds logs a @racket['debug] level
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/engine.scrbl
```diff
@@ -13,7 +13,7 @@ inspired by the work of Haynes and Friedman @cite["Haynes84"].
 Engines log their behavior via a logger with the name 
 @racket['racket/engine]. The logger is created when the module
 is instantiated and uses the result of @racket[(current-logger)]
-as its parent. The library adds logs a @racket['debug] level
+as its parent. The library adds to logs a @racket['debug] level
```
mflatt, line 16: I think just dropping `adds` is probably better: "The library logs a @racket['debug] level...".

Also, probably there should have been a hyphen between `@racket['debug]` and `level`.

Ok if I make those changes here?

jasonhemann: Please and thank you!

mflatt:
Thanks!

PR #4379, opened 2022-08-01
Zuo: show `CC_FOR_BUILD` in `--help` and infer from `--host`

LiberalArtist:
The first commit just tweaks the help output from Zuo's `configure.ac`. The second uses `AX_PROG_CC_FOR_BUILD` from the Autoconf archive to infer those values, so you can do `./configure --host=powerpc64le-linux-gnu` rather than `./configure --host=powerpc64le-linux-gnu CC_FOR_BUILD=gcc`. If the extra M4 file doesn't seem like too much, I think it might be even more useful elsewhere.


mflatt:
The first of the two commits looks like a clear improvement.

For the second commit, I'm still reluctant to pull in the extra support for `configure` to define `CC_FOR_BUILD`. Also, it doesn't seem like a good idea for `zuo` sources to depend on things inside the Racket source tree but outside the `zuo` directory. How bad is requiring `CC_FOR_BUILD` to be specified? Is it common that it's inferred automatically?

LiberalArtist:
I've dropped the second commit. Supplying `CC_FOR_BUILD' isn't causing any difficulty for me here, and I agree that keeping Zuo as self-contained as reasonably possible seems good.

mflatt:
Thanks!

Issue #4359, opened 2022-07-15
Chez test failure on `release` branch


LiberalArtist:
I've gotten a new Chez Scheme test failure while updating my Guix patch series to build fb526d1d31f5c26a606b62e2074566c7f5a13e54 from the `release` branch. The full build log is at [z5b0b77px8iscdmnma5q0zgvx5gfli-chez-scheme-for-racket-9.5.9.2.drv.gz](https://github.com/racket/racket/files/9117071/z5b0b77px8iscdmnma5q0zgvx5gfli-chez-scheme-for-racket-9.5.9.2.drv.gz); here's the tail:
```
Hunk #9 succeeded at 8055 (offset 23 lines).
Hunk #10 succeeded at 9331 (offset 23 lines).
Hunk #11 succeeded at 10855 (offset 23 lines).
running ta6le/bin/ta6le/scheme to build all with output to ta6le/mats/output-interpret-0-f-t-f-rmg2/make.out
patch ta6le/mats/experr-interpret-3-f-t-f ../ChezScheme/mats/patch-interpret-3-f-t-f
running ta6le/bin/ta6le/scheme to build all with output to ta6le/mats/output-interpret-3-f-t-f-rmg2/make.out
running ta6le/bin/ta6le/scheme to build all with output to ta6le/mats/output-compile-0-f-f-f-hci-cl/make.out
running ta6le/bin/ta6le/scheme to build all with output to ta6le/mats/output-interpret-3-f-f-f-hci-rmg/make.out
-------- o=0 eoc=f simple --------
-------- o=3 eoc=f simple --------
7d6
< misc.mo:Bug in mat compute-size-increments clause 17 at line 1102, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.9.2.drv-0/source/racket/src/ChezScheme/mats/misc.ms
-------- o=0 cp0=t cl=3 cl3 --------
-------- o=3 cp0=t cl=3 cl3 --------
7d6
< misc.mo:Bug in mat compute-size-increments clause 17 at line 1102, char 3 of /tmp/guix-build-chez-scheme-for-racket-9.5.9.2.drv-0/source/racket/src/ChezScheme/mats/misc.ms
-------- o=0 spi=t p=t rmg=2 spi-rmg --------
-------- o=3 spi=t p=t rmg=2 spi-rmg --------
7d6
[...]

PR #4380, opened 2022-08-02
Chez Scheme: Fix build on GNU/Hurd.

LiberalArtist:
Mostly GNU/Hurd should take the same options as GNU/Linux. One
difference is that the Hurd does not define macros such as `PATH_MAX` or
`NOFILE`, because it avoids imposing arbitrary limits on such resources.
This patch provides alternatives for localized uses of those constants,
but it accepts the pervasive use of `PATH_MAX` in finding bootfiles for
now. See https://www.gnu.org/software/hurd/hurd/porting/guidelines.html.

Tested on Debian GNU/Hurd 20220331 and cross-compilation via Guix for
i586-pc-gnu with the pb and tpb32l backends.

### DIFF
--- a/racket/src/ChezScheme/c/version.h
+++ b/racket/src/ChezScheme/c/version.h
@@ -80,7 +80,7 @@ FORCEINLINE void store_unaligned_uptr(uptr *addr, uptr val) {
 /*****************************************/
 /* Operating systems                     */
 
-#if defined(__linux__)
+#if defined(__linux__) || defined(__GNU__) /* Hurd */
 #define NOBLOCK O_NONBLOCK
 #define LOAD_SHARED_OBJECT
 #define USE_MMAP
@@ -91,7 +91,10 @@ FORCEINLINE void store_unaligned_uptr(uptr *addr, uptr val) {
 #define GETPAGESIZE() getpagesize()
 typedef char *memcpy_t;
 #define MAKE_NAN(x) { x = 0.0; x = x / x; }
-#define GETWD(x) getcwd((x),PATH_MAX)
+#ifndef __GNU__ /* Hurd: no PATH_MAX */
+/* n.b. don't test PATH_MAX directly: we have not yet included <limits.h>  */
+# define GETWD(x) getcwd((x),PATH_MAX)
+#endif
 typedef int tputsputcchar;
 #ifndef __ANDROID__
 # define LOCKF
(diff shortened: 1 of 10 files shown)


racket/src/ChezScheme/c/version.h
```diff
@@ -91,7 +91,9 @@ FORCEINLINE void store_unaligned_uptr(uptr *addr, uptr val) {
 #define GETPAGESIZE() getpagesize()
 typedef char *memcpy_t;
 #define MAKE_NAN(x) { x = 0.0; x = x / x; }
-#define GETWD(x) getcwd((x),PATH_MAX)
+#ifdef PATH_MAX
+# define GETWD(x) getcwd((x),PATH_MAX)
+#endif
 typedef int tputsputcchar;
```
LiberalArtist, line 97: This caused a problem that wasn't exposed until I tried to cross-compile Racket CS using Guix (this patch had fixed only Chez Scheme). Apparently we have not yet included `<limits.h>` at this point, so `PATH_MAX` is never defined here: all of the references on the right-hand sides of `#define` are unhygienic. The problem was masked by the fallback implementation of `current-directory` using `/bin/pwd`:
https://github.com/racket/racket/blob/2f84598893f3409fee9f2b0d5c952cf09be98916/racket/src/ChezScheme/s/6.ss#L37-L50
Guix revealed the issue because `/bin` doesn't exist in its build environment. It seems to have been just coincidental that the failure came up while cross-compiling specifically: apparently most build modes don't exercise `(current-directory)`, and I have the Chez test suite disabled again in Guix due to https://github.com/racket/racket/issues/4359#issuecomment-1202036929.

In some sense it seems like the "right thing" would be to include `<limits.h>` or test in `configure` or something, but I'm inclined to change the test to `#ifndef __GNU__`, instead: that's the style of the other conditionals here, and it should keep the change small and prevent any other surprises for non-Hurd systems.

It might be more robust to remove this use of `PATH_MAX` across the board along the lines of https://github.com/racket/racket/commit/b4d452cc71c1ca6f848d7404c09d75f14f51d170—in this case, we're dynamically allocating the storage anyway!—but that seems like a bigger change that should go in its own patch.

LiberalArtist: I've pushed a revised commit, which I'm now testing.

LiberalArtist:
This is working nicely with both Debian and Guix. I've also pushed some commits adding a native machine type for the Hurd, which was shockingly easy.

PR #4384, opened 2022-08-05
BC: Fix build on GNU/Hurd

LiberalArtist:


LiberalArtist:
While working on this, I noticed that cross-compiling CGC (and possibly 3M) for any platform seems to require `--enable-racket`, or the build scrips will try to try to run the cross-compiled Racket for `raco setup`. That's not a problem for me at all; I just thought I'd mention it in case it was supposed to build a Racket for the build platform automatically.

samth:
It looks like this change broke a bunch of tests on DrDr. Eg http://drdr.racket-lang.org/61304/pkgs/racket-test-core/tests/racket/test.rkt

racket/src/bc/configure.ac
```diff
@@ -557,7 +557,7 @@ case "$host_os" in
       X_PRE_LIBS=""
     fi
     ;;
-  linux*)
+  linux*,gnu*)
     LIBS="$LIBS -ldl -lm -lrt -rdynamic"
```
LiberalArtist, line 561: @samth This, and the corresponding generated code, should have used been `linux*|gnu*)`, rather than `,`. I don't know why it would have caused problems for DrDr but not me locally, but I'll open a PR a repair momentarily.

PR #4387, opened 2022-08-06
Fix #4386

### LINKED ISSUE #4386: Interface contract erroneously affects superclass
e.g., 8.6.0.1 [cs]

**What program did you run?**

```rkt
(define foo%
  (class* object% ()
    (super-new)
    (define/public (f x) x)))

;; Change `[f any/c]` to `f` and this will work.
(define interface<%> (interface () [f any/c]))

;; This class is thrown away, so it should have no effect.
(class* foo% (interface<%>))

(define obj (new foo%))
(send obj f 10)
```

No error.

```
; application: not a procedure;
;  expected a procedure that can be applied to arguments
;   given: '(#<procedure:f method in foo%> ((#<flat-contract: any/c> (interface interface<%>) (class ...home/camoy/tmp/a.rkt:10:0) #f)))
```

camoy:
A subclass can share values with its superclass when there are no
changes to the method table. Attaching an interface to a subclass
was not considered a method change, however when that interface
has contracts it will mutable the method table. Sharing in this
case causes the superclass to be messed up. The solution is to
consider interface attachment as a method change so that sharing
no longer occurs.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/object.rktl
+++ b/pkgs/racket-test-core/tests/racket/object.rktl
@@ -2400,5 +2400,29 @@
   (test 'outer values (send (new c2%) f)))
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; Check that interface contracts on a subclass work correctly.
+
+(let ()
+  (define sup%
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks!

PR #4390, opened 2022-08-12
Remove self-hosted runner workflows

pmatos:
For quite some time, I have been having issues keeping all my boards up. The days here are getting hotter every year and this year I lost a few boards. In addition to that energy is getting more expensive and I am finding it harder and harder to justify running many boards 24/7 on my cellar and buying new ones or new SD cards when the previous ones fail.

For that reason, I am removing the self-hosted running workflows. It will hopefully bring some needed stability to the CI.

An alternative is to find a host proving arm machines for rent. Hetzner has dedicated arm machines but they are rising their prices due to energy costs as well, and the dedicated machines are too expensive. It would be better to have something allocated as needed but I am not sure who provides that at this point.


samth (approved):

pmatos:
Thanks.

Issue #4393, opened 2022-08-15
[Feature Request] Document Y/n/a/c/ dependencies installation options


rdela:
In [1.4 Installing Packages](https://docs.racket-lang.org/pkg/getting-started.html#%28part._installing-packages%29), the `--auto` option is documented, but the meaning of `Y/n/a/c/` in:

```shell
Would you like to install these dependencies? [Y/n/a/c/?] 
```

…is not. `Y` is probably "yes" and `n` is probably "no." What do `a` and `c` mean? Why would programmers be expected to know this? This should be documented in [1.4 Installing Packages](https://docs.racket-lang.org/pkg/getting-started.html#%28part._installing-packages%29).

sorawee:
That's what `?` is for, though interestingly it says `Invalid answer: ?`. This should be fixed.

Anyway, here's the response:

```
 Answer nothing or `y' or `Y' for "yes", `n' or `N' for "no", or
 `a' or `A' for "yes for all", or `c' or `C' for "cancel".
```

rdela:
How would a programmer know that typing `?` was even supposed to explain the choices? I read that as a second question mark, not an option. I think an explanation of all of the options in 1.4 Installing Packages is in order, once `?` is fixed. Something like:

> If the package depends on other packages that you do not have installed already, then [raco pkg install](https://docs.racket-lang.org/pkg/cmdline.html#%28part._raco-pkg-install%29) will alert you and ask whether it should install them, too. Answer nothing or `y` or `Y` for "yes", `n` or `N` for "no", `a` or `A` for "yes for all", `c` or `C` for "cancel", or `?` for "help". Use `--auto` to skip the question and make dependencies installed automatically.

PR #4394, opened 2022-08-15
raco pkg: Don't show "Invalid answer" for "?"

sorawee:
Related to #4393

### DIFF
--- a/racket/collects/pkg/private/print.rkt
+++ b/racket/collects/pkg/private/print.rkt
@@ -67,7 +67,8 @@
          [""
           (if default-yes? 'yes 'no)]
          [x
-          (eprintf "Invalid answer: ~a\n" x)
+          (unless (equal? x "?")
+            (eprintf "Invalid answer: ~a\n" x))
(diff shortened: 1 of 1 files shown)


otherjoel:
(Could the response also be updated to use `'` instead of the annoying backticks-as-fake-curly-quotes?)

sorawee:
Done.

mflatt:
I would have either left this as

```
`x'
```

or changed to

```
`x`
```

but not

```
'x'
```

The first form was the convention for Racket messages for a long time (the first decade or two), where the intent was to make something that didn't look like an S-expression to help clarify that the content is meant as literal. Since Markdown became popular and literalness with backquotes is widely understood, we moved to the second backquote convention. I've been converting the old forms only gradually, though, because I always worry that something may be relying on exact text.

Any objection to changing to the backquote form here?

sorawee:
I can change it to backtick.

shhyou:
I wonder whether using actual quotes i.e. `‘’` and `“”` could be better. Backticks are also used for starting commands in many shells, so I'm worried that printing pairs of backticks may create security problems when the outputs accidentally ended up in the shell. That said, I don't know any immediate issues.

Issue #4375, opened 2022-07-28
[Feature Request] support errortrace in raco test


sorawee:
`raco test` should support `--errortrace`.

See also https://github.com/shhyou/raco-test-with-errortrace. The difficulty in the integration to `raco test` is: "It's just that... raco test also uses processes and places, so I'm not sure how to pass the configuration" (from @shhyou)

capfredf:
>  I'm not sure how to pass the configuration"

does `dynamic-require` work for that purpose?

PR #4396, opened 2022-08-17
Allows for loading errortrace in raco test (#4375)

shhyou:
This PR adds a hack-y way to support `--errortrace` for `raco test` (#4375).
@sorawee @capfredf let me know if you see any issue.

### DIFF
--- a/pkgs/compiler-lib/compiler/commands/test.rkt
+++ b/pkgs/compiler-lib/compiler/commands/test.rkt
@@ -1167,9 +1167,19 @@
    ;; take place; if errortrace (and racket/base) is loaded before make? sets
    ;; current-load/use-compiled, raco test will end up compiling everything
    ;; which is unlikely to be what the user wants.
-   (when (and load-errortrace?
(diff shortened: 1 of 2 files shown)


samth:
In what sense is this hacky? It seems like exactly how it ought to work. Is there some part that doesn't work right?

shhyou:
`--make` and `--errortrace` don't work well together, so I directly restrict their combination. If `errortrace` is loaded before the compilation/load handler is set and the user has already compiled Racket with `raco setup --mode errortrace -j 1` then it could actually work.

capfredf:
I've started using this new option. Cheers!

otherjoel:
> I see that this didn't include a history annotation. @shhyou can you make a PR to add one?

While in there, should the double-entry for `--output` be corrected?

https://github.com/racket/racket/blob/0f0148c7d96d62b9991a682907d3d6f07e1067d0/pkgs/racket-doc/scribblings/raco/test.scrbl#L224-L225

shhyou:
Sure to both.

PR #4401, opened 2022-08-21
racket/promise: add for{,*}/concurrent

Bogdanp:
I often find myself wanting a form like this when I write code that does concurrent IO over a set of things. Happy to change the name if anyone can think of a better one.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/promise.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/promise.scrbl
@@ -174,3 +174,49 @@ TODO: Say something on:
   demand.
 ;}
 }
+
+@(define promise-eval
+   (let ([eval (make-base-eval)])
+     (eval '(require racket/promise))
+     eval))
+
+@defform[
+  (for/list/concurrent maybe-group (for-clause ...)
+    body-or-break ... body)
+  #:grammar
+  [(maybe-group (code:line)
+                (code:line #:group thread-group-expr))]
+  #:contracts ([thread-group-expr thread-group?])]{
+
+  Iterates like @racket[for/list], but the bodies (following any
+  @racket[#:break] or @racket[#:final] clauses) are wrapped in
+  @racket[delay/thread].  Each @tech{promise} is forced before the
+  result list is returned.
+
+  Threads are created under @racket[thread-group-expr], which defaults
+  to @racket[(make-thread-group)].  An optional @racket[#:group]
+  clause may be provided, in which case the threads will be created
+  under that thread group.
+
+  This form does not support returning multiple values.
+
+  @examples[
+    #:eval promise-eval
+    (time
+     (for/list/concurrent ([i (in-range 5)])
+       (define duration (/ 1.0 (random 50 100)))
+       (sleep duration)
+       (printf "thread ~a slept for ~a milliseconds~n" i (truncate (* duration 1000)))
+       i))
+  ]
+
+  @history[#:added "8.6.0.4"]
+}
+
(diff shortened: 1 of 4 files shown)


racket/collects/racket/promise.rkt
```diff
@@ -223,3 +225,31 @@
...
+                   ([promises null] #:result (map force (reverse promises)))
+                   clauses
+                   pre-body ...
+                   (cons (delay/thread*
+                          #:group group
+                          (let () post-body ...))
+                         promises))))))]
+        [(_ clauses body ...)
+         ((make-transformer for/fold/derived-stx stx)
+          #'(fc #:group (make-thread-group) clauses body ...))]))
```
samth, line 252: Where is `fc` bound?

pkgs/racket-doc/scribblings/reference/promise.scrbl
```diff
@@ -174,3 +174,46 @@ TODO: Say something on:
...
+     eval))
+
+@defform[
+  (for/concurrent maybe-group (for-clause ...) body-or-break ... body)
+  #:grammar
+  [(maybe-group (code:line)
+                (code:line #:group thread-group-expr))]
+  #:contracts ([thread-group-expr thread-group?])]{
+
+  Iterates like @racket[for/list], but the bodies (following any
```
samth, line 190: I guess I'd expect the name `for/concurrent` to be like `for`, instead of like `for/list`.

pkgs/racket-doc/scribblings/reference/promise.scrbl
```diff
@@ -174,3 +174,46 @@ TODO: Say something on:
...
+  @racket[#:break] or @racket[#:final] clauses) are wrapped in
+  @racket[delay/thread].  Each @tech{promise} is forced before the
+  result list is returned.
+
+  Threads are created under @racket[thread-group-expr], which defaults
+  to @racket[(make-thread-group)].  An optional @racket[#:group]
+  clause may be provided, in which case the threads will be created
+  under that value.
```
samth, line 198:
```suggestion
  under that thread group.
```

samth (commented):
This seems like a good addition, although I'm not sure about the name. I think tests are also needed.

pkgs/racket-doc/scribblings/reference/promise.scrbl
Bogdanp: Yeah, that's part of why I don't like the name either. I think the fact that it returns a list is important, though (otherwise you'd just use a `for` with `thread`s in the body). So, it should definitely be like `for/list`, but with a better name.

racket/collects/racket/promise.rkt
Bogdanp: It's not bound anywhere because the macro ignores it. It just needs to be there to match the pattern (and this is how a few of the other `for` macros (like `for/vector`) do recursion).

pkgs/racket-doc/scribblings/reference/promise.scrbl
mflatt: `for/conscurrent`?

Ok, I don't have a good name, either. I assume that no one is suggesting something like `for/list/concurrent` because it's too long.

Bogdanp: > I assume that no one is suggesting something like for/list/concurrent because it's too long.

I was thinking it would be, but now I've updated some code that uses `for/concurrent` and it's not too bad. I'll update the PR later today to go with that name.

Bogdanp:
I've renamed the forms and squashed the commits.

PR #4403, opened 2022-08-22
racket/future: fix for{,*}/async #:break clauses

Bogdanp:
While working on `for/list/concurrent`, I noticed some issues with `for/async`:

* it doesn't support break clauses,
* ~~touching the futures isn't tail recursive, and~~
* the form returns a result even though it shouldn't.

I've fixed the first ~~two~~ one and added some tests. It would be nice to fix the 3rd item as well, but it's probably not worth making a backwards-incompatible change here.

EDIT: Then again, the documentation does state that the touch order isn't guaranteed so no one should be depending on the result value anyway. Maybe it wouldn't be so bad to change it to return `void`.

EDIT2: I took some measurements and the non-tail recursive version doesn't actually consume more memory than the tail-recursive one on my machine until I do 300k iterations or more, which is probably a lot more than anybody using this form would do anyway, so I've undone that change to preserve the original implementation.

EDIT3: I wrote a little tool to iterate over all packages in a catalog, `read` their `.rkt` sources and search for uses & applications of a given `id` (it does a naïve match by symbol, so it isn't aware of renames, bindings, etc). Running it on my most recent catalog snapshot, [these](https://gist.github.com/Bogdanp/ea71e27ad53339ffb7c0196554b88c88) are all the usages of the two forms that turn up. They all look like they already assume that `for/async` doesn't return a value, so I think the chances of anything breaking if we change the forms to return `void` are very small.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/futures.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/futures.scrbl
@@ -131,8 +131,8 @@ execute through a call to @racket[touch], however.
 }
 
 @deftogether[[
-@defform[(for/async (for-clause ...) body ...+)]
-@defform[(for*/async (for-clause ...) body ...+)]]]{
+@defform[(for/async (for-clause ...) body-or-break ... body)]
+@defform[(for*/async (for-clause ...) body-or-break ... body)]]]{
 
 Like @racket[for] and @racket[for*], but each iteration of the
 @racket[body] is executed in a separate @racket[future], and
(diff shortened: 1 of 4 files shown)


Bogdanp:
Based on "EDIT3", I've gone ahead and changed the forms to return `void`. I think the likelihood of that change breaking any code in the wild is very low, and it should make it more clear that this form is meant to be used for its effects.  I've left the order of touches intact since it doesn't seem to make much of a difference in terms of time or space either way you do it. I think this is good to go.

PR #4409, opened 2022-08-24
openssl updates (tls13, hostname validation, tls-exporter)

rmculpepper:
This PR combines several fixes and improvements:
- sets OpenSSL 1.0.2 (LTS, released 1/2015, EOL 12/2019) as the minimum required for full functionality
- add `'tls13`, fix protocol version handling for recent OpenSSL versions
- use OpenSSL to do hostname verification (replaces some Racket code, failure sends proper alert to peer)
- add `'tls-exporter` channel binding type, fix a memory bug in `'tls-server-end-point` (when used on server)

It also splits the related FFI bindings out into a separate private/ffi.rkt module and adds some version information.

### DIFF
--- a/pkgs/racket-doc/openssl/openssl.scrbl
+++ b/pkgs/racket-doc/openssl/openssl.scrbl
@@ -276,6 +276,14 @@ current platform for client connections.
 @history[#:added "6.1.1.3"]
 }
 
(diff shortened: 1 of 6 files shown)


mflatt:
LGTM

PR #4412, opened 2022-08-25
openssl: auto DHE, ECDHE

rmculpepper:
Enable DHE and ECDHE automatically for all server contexts.

In OpenSSL 1.1.0 and later, ECDHE is enabled by default with automatic selection of the group (curve) used for key-exchange, and DHE can be enabled with auto group selection. In 1.0.2, ECDHE can be manually enabled with auto group selection, and DHE can be enabled using a 2048-bit built-in group. That is, for DHE this PR replaces the external 4096-bit DH group with either auto-selection (recommended by OpenSSL developers) or with an internal 2048-bit group (only for 1.0.2, EOL since 12/2019).

This change simplifies the code and the interface, but it removes some customizability.


mflatt:
Looks good as far as I understand

sorawee:
Is there a reason to hold this back from merging (besides fixing the `@history` to update the version number).

PR #4414, opened 2022-08-25
Suggesting removing automatically installed packages after `raco pkg remove`

shhyou:
Suggesting removing automatically installed packages after `raco pkg remove`. This may provide a slightly better experience for <https://github.com/racket/drracket/issues/575#issuecomment-1226580105>.

Examples:

```
$ ./racket/bin/raco pkg remove --no-setup compiler-test
Removing compiler-test
The following packages were automatically installed and are no longer used:
  2d-lib at-exp-lib class-iop-lib compatibility-lib data-lib db-lib
  deinprogramm-signature draw-aarch64-macosx-3 draw-lib drracket-plugin-lib
  drracket-tool-lib drracket-tool-text-lib dynext-lib eli-tester gui-aarch64-macosx
  gui-lib htdp-lib html-lib icons images-gui-lib images-lib mzscheme-lib
  net-cookies-lib net-lib option-contract-lib parser-tools-lib pconvert-lib
  pict-lib plai-lib planet-lib r5rs-lib r6rs-lib racket-index rackunit-gui
  rackunit-typed sandbox-lib sasl-lib scribble-html-lib scribble-lib
  scribble-text-lib simple-tree-text-markup-lib slideshow-lib snip-lib
  srfi-lib srfi-lite-lib string-constants-lib syntax-color-lib tex-table
  typed-racket-lib typed-racket-more unix-socket-lib web-server-lib
  wxme-lib
Use 'raco pkg remove --auto' to remove them.
$ ./racket/bin/raco pkg remove --auto
Removing class-iop-lib
Moving class-iop-lib to trash: ...
```


samth:
This seems like an excellent idea, but it needs some docs and tests.

rfindler:
I agree this is a great idea!

shhyou:
Thanks! I didn't realize that `raco pkg` has a testing infrastructure.

Re docs update: given that there is no functionality change, what should the document say? The document on [pkg-remove](https://docs.racket-lang.org/pkg/lib.html#%28def._%28%28lib._pkg%2Flib..rkt%29._pkg-remove%29%29) already says that it may print messages if `from-command-line?` is true.

mflatt:
The documentation says that `from-command-line?` allows suggestions about command-line flags when there's an *error*. Printing extra things when there's no error can create trouble. With `git`, for example, I have to remember to include `-q` when calling it from a script. Along those lines, maybe `--batch` should disable the suggestion here, which probably means that `pkg-remove` gets a new argument.

Also, I advocate quoting suggestion with

```
` `
```

instead of

```
' '
```

See also #4394, and maybe a next step (for another commit/PR) is to update all suggestions to use backquoting.

shhyou:
@mflatt Thanks! I edited the docs to note that `raco pkg remove` may print additional instructions and also check the `quiet?` flag. Since `raco pkg remove` already indicates the list of removed packages in the standard output (e.g. the first line in the example), I think it's relatively safe to print more suggestions.

Other than `raco pkg remove`, the other package commands also don't support the `--quiet` flag despite having the `quiet?` argument. I think that warrants a separate PR.

mflatt:
Thanks!

PR #4415, opened 2022-08-26
Add `list-update*` and `list-set*`.

NoahStoryM:


racket/collects/racket/list.rkt
```diff
@@ -129,16 +131,61 @@
...
+  (let loop ([l l] [i i])
+    (cond
+      [(zero? i) (cons (f (car l)) (cdr l))]
+      [else (cons (car l) (loop (cdr l) (sub1 i)))])))
+
+(define (list-update* l . pairs)
+  (unless (list? l)
+    (apply raise-argument-error 'list-update* "list?" 0 l pairs))
+  (unless (even? (length pairs))
+    (error 'list-update* "expected an even number of association elements, but received an odd number: ~e" pairs))
```
sorawee, line 143: IIUC, we should avoid using `error`, since it does not follow the error message conventions.

Here's how this kind of pairing errors are reported in `racket/collects/racket/private/hash.rkt`'s `paired-fold`:

```
(raise-arguments-error
          who
          (format "expected ~a, but received ~a"
                  "an even number of association elements"
                  "an odd number of association elements")
          "association elements"
          pairs0)
```

racket/collects/racket/list.rkt
```diff
@@ -129,16 +131,61 @@
...
+  (let loop ([l l] [i i])
+    (cond
+      [(zero? i) (cons v (cdr l))]
+      [else (cons (car l) (loop (cdr l) (sub1 i)))])))
+
+(define (list-set* l . pairs)
+  (unless (list? l)
+    (apply raise-argument-error 'list-set* "list?" 0 l pairs))
+  (unless (even? (length pairs))
+    (error 'list-set* "expected an even number of association elements, but received an odd number: ~e" pairs))
```
sorawee, line 176: Ditto.

racket/collects/racket/list.rkt
```diff
@@ -129,16 +131,61 @@
...
+  (let check ([p pairs] [i 1])
+    (unless (null? p)
+      (unless (exact-nonnegative-integer? (car p))
+        (apply raise-argument-error 'list-set* "exact-nonnegative-integer?" i l pairs))
+      (check (cddr p) (+ 2 i))))
+  (define cache (apply hasheq pairs))
```
sorawee, line 182: I think there must be additional checks that all indices are in bound. Otherwise, you get:

```
> (list-set* '(1 2 3) 1000 1)
car: contract violation
  expected: pair?
  given: '()
```

Issue #3137, opened 2020-04-30
Preserve procedure-arity for composed functions


countvajhula:
**Observed behavior**: For composed functions, `procedure-arity` always(?) reports an arity of `at least 0`.

**Expected behavior**: the arity of a composed function should be the same as the arity of the leading function in the composition.

**Example**:
```
(define (str-append-3 x y z)
  (string-append x y z))
```
`(procedure-arity str-append-3) ; => 3`

```
(define f (compose string-length str-append-3))
(procedure-arity f) ; => (arity-at-least 0)
```
```
(f "hello" "there" "friend") ; => 16
(f "hello")
; str-append-3: arity mismatch;
;  the expected number of arguments does not match the given number
;   expected: 3
;   given: 1
; Context:
;  .../private/list.rkt:322:16 composed
;  .../racket/repl.rkt:11:26

```
**Additional details** (probably superfluous, but in case it is useful):
I'm using `procedure-arity` inside an experimental high-level function abstraction that can in principle stand in for normal functions but provide additional insight into their composition and semantics. It is also curried by default, like Haskell functions. `procedure-arity` is being used to report the arity of the composed function inside the wrapper struct, to decide whether the fresh arguments should be incorporated into the function call to return a curried function, or whether the results of invocation (including errors, such as arity errors) should simply be returned to the caller. Since the arity reported for the composed function doesn't match the leading function arity (leading to premature termination of currying) I'm instead checking the leading function specifically (which is retained and only composed at invocation time), which works correctly. But I'm not sure if this might lead to other issues down the line -- for instance, if the leading function is already (an elementary) composed one, it seems like we wouldn't be able to reliably know the arity here.

NoahStoryM:
The comment in the source (https://github.com/racket/racket/blob/master/racket/collects/racket/private/list.rkt#L336) says that the reason why `procedure-reduce-keyword-arity` is not used to preserve the information is because it adds a significant overhead. But if I understand correctly, `compose` runs `mk-simple-compose` at most twice, which means it runs `procedure-reduce-keyword-arity` at most twice too, why is there a significant overhead?

PR #4416, opened 2022-08-28
[compose]: Preserve the information of the 1st procedure.

NoahStoryM:
Fix https://github.com/racket/racket/issues/3137 .

It seems that `procedure-reduce-arity` and `procedure-reduce-keyword-arity` won't add a significant overhead since they are called at most twice?


mflatt:
I think the original concern was not the time to call `procedure-reduce-arity` or `procedure-reduce-keyword-arity`, but the extra time added to every call a procedure whose arity has been adjusted.

Did create any benchmarks to measure `compose` time and the times for calling composed functions before and after this change? Racket CS is faster at calling procedures, and the penalty for wrapping to adjust arity is probably lower in absolute terms, but I'm not sure what the times will look like.

NoahStoryM:
Code:
```racket
(define (f0 a b c x) '())
(define (f1 a b c x #:y [y 'y]) '())
(define (f2 a b c x #:y y) '())

(for ([f (in-list (list f0 f1))])
  (displayln (object-name f))
  (time
   (for ([i (in-range 1000000)])
     (for ([C (in-list (list compose compose1))])
       (C list list list list f))))
  (time
   (for ([i (in-range 1000000)])
     (for ([C (in-list (list compose compose1))])
       ((C list list list list f) 'a 'b 'c 'x))))
  (newline))

(displayln (object-name f2))
(time
 (for ([i (in-range 1000000)])
   (for ([C (in-list (list compose compose1))])
     (C list list list list f2))))
(time
 (for ([i (in-range 1000000)])
   (for ([C (in-list (list compose compose1))])
     ((C list list list list f2) 'a 'b 'c 'x #:y 'y))))
```


[cs] Before:
```
f0
cpu time: 2364 real time: 2368 gc time: 46
cpu time: 2542 real time: 2546 gc time: 13

f1
cpu time: 2038 real time: 2042 gc time: 11
cpu time: 2525 real time: 2529 gc time: 12

f2
cpu time: 2039 real time: 2043 gc time: 11
cpu time: 3365 real time: 3370 gc time: 15
```

[cs] After:
```
f0
cpu time: 2256 real time: 2268 gc time: 36
cpu time: 2539 real time: 2548 gc time: 13

f1
cpu time: 2911 real time: 2915 gc time: 15
cpu time: 3595 real time: 3601 gc time: 18

f2
cpu time: 14300 real time: 14326 gc time: 860
cpu time: 15931 real time: 15967 gc time: 862
```

NoahStoryM:
I'm surprised that `procedure-reduce-keyword-arity` has such a big difference in performance between `f1` and `f2`.

mflatt:
Let's lift the compose action out when measuring application time, so it's clearer what were measuring:

```
(define (f0 a b c x) '())
(define (f1 a b c x #:y [y 'y]) '())
(define (f2 a b c x #:y y) '())

(for ([f (in-list (list f0 f1))])
  (displayln (object-name f))
  (time
   (for ([i (in-range 1000000)])
     (for ([C (in-list (list compose compose1))])
       (C list list list list f))))
  (time
   (for ([C (in-list (list compose compose1))])
     (let ([f (C list list list list f)])
       (for ([i (in-range 1000000)])
         (f 'a 'b 'c 'x)))))
  (newline))

(displayln (object-name f2))
(time
 (for ([i (in-range 1000000)])
   (for ([C (in-list (list compose compose1))])
     (C list list list list f2))))
(time
 (for ([C (in-list (list compose compose1))])
   (let ([f (C list list list list f2)])
     (for ([i (in-range 1000000)])
       (f 'a 'b 'c 'x #:y 'y)))))
```

Here's a summary of results on my machine (M1 MacBook Pro), where `comp` is the current version and `comp'` is this PR, etc., with times in milliseconds from one particular set of runs (where the variation across runs was small):

```
          CS (arm64osx)           BC (a6osx)
    comp  comp'  app  app'   comp  comp'  app  app'
f0   852   817    86   90    1514  3483   359  479
f1   673  1017   184  186    1811  3664   568  857
f2   672  4947   416  423    1886  7555   594  771
```

The `a6osx` numbers should be considered approximate, since that's going through Rosetta2, but my experience has been that relative timings are usually consistent with x86_64 hardware.

Racket CS has pretty much eliminated the extra cost on the result of composition. Maybe we don't care that there's some overhead on applications for BC. As you noted, it's the composition itself in the case of keyword arguments that's especially expensive.

I think the big cost is a result of an implementation strategy that creates a new structure type to represent the new keyword arity or procedure name. I don't remember why things are implemented that way, but it might be merely because `prop:object-name` didn't exist when keyword arguments were first implemented. I'll look into changing the implementation to avoid creating a structure type.

mflatt:
By changing the implementation of `procedure-reduce-keyword-arity` for required-keyword procedures, I get to these numbers for `comp''` (where square brackets indicate measurements that aren't expect to change):

```
          CS (arm64osx)                    BC (a6osx)
    comp  comp' comp''  app  app'   comp comp' comp''  app  app'
f0   852   817  [818]    86   90    1514 3483 [3354]   359  479
f1   673  1017 [1084]   184  186    1811 3664 [3576]   568  857
f2   672  4947  1801    416  423    1886 7555  4584    594  771
```

Another possible improvement is to use `procedure-arity-mask` instead of `procedure-arity`. It turns out that the mask variant is more primitive in CS, and also still faster in BC. Here are all variants (including the original) with that change added:

```
          CS (arm64osx)                    BC (a6osx)
    comp  comp' comp''  app  app'   comp comp' comp''  app  app'
f0   284   380  [378]    85   90     821 1802 [1774]   334  439
f1   394   784  [774]   180  186    1156 3106 [3002]   561  744
f2   397  4561  1540    410  394    1181 6896  4047    540  766
```

Here's the overall diff, where `comp` is with arity instead of arity mask, and `comp''` here is everything:

```
          CS (arm64osx)            BC (a6osx)
    comp  comp'' app  app''  comp comp'' app  app''
f0   852   378    86   90    1514 3483   359  479
f1   673   774   184  186    1811 3664   568  857
f2   672  1540   416  394    1886 7555   594  771
```

One caution is that I haven't run tests carefully, yet, but here's the implementation I measured: https://github.com/mflatt/racket/tree/compose

Maybe this is cheating a bit, comparing a arity-mask new implementation against the non-mask old implementation, but it looks like it might be a good trade-off for getting precise arity information on composed functions. I'd be interested to hear what others think.

mflatt:
It looks like new tests should go in `pkgs/racket-test-core/tests/racket/function.rktl`.

rfindler:
Am I understanding the numbers right: if we get precise arity information on a composed function, we take a performance hit of about 2x when we're calling compose in the case that we pass a keyword function with mandatory arguments, but everything else (both calls to compose and calls to the resulting function) is either the same (or better because of a better implementation strategy)?

On one hand, this seems totally fine to merge from the numbers perspective (at least to me); on the other hand, the fact that mandatory keyword arguments slowed down and optional ones didn't seems weird. Is there something special in the implementation for optional keyword arguments that's making a difference here? Or did I just misread the experiment?

mflatt:
> Is there something special in the implementation for optional keyword arguments that's making a difference here?

Yes, functions that require keywords are implemented differently than functions with optional keywords. The different implementations balance trade-offs related to error reporting, arity reporting, and minimizing closures. I don't know whether the trade-offs are the best, though.

NoahStoryM:
Maybe this [commit (faster arity reduce and rename for keyword-requiring procedures )](https://github.com/mflatt/racket/commit/0f11de7be5be671d3fed3337cac3253f372f4a46) should be in a new PR.

mflatt:
> Maybe this commit should be in a new PR.

I've finished it and pushed, so you could rebase now.

mfelleisen:
> On Aug 28, 2022, at 10:18 PM, Matthew Flatt ***@***.***> wrote:
> 
> I'd be interested to hear what others think.

Parenthetical thought: 

[[ My Sw Dev project has run into one particularly bad benchmark, and I use compose 
and I used to use #:keywords in the innermost loop. I am not sure I’d like a 2x hit. 

But as I said, it’s a one-off bad benchmark and I doubt the compose/keyword would
have made a real difference. — Too much work to run the 10-min benchmark on this
branch :-) ]]

mflatt:
It looks like recent improvements recover some performance that had been lost relative to v8.6 (as part of earlier changes for `values`?).  Numbers on CS `arm64osx` as of the current PR, where this time I tried `(compose identity f)` in addition to `(compose list list list list f)` and added an `f` that takes a single argument (https://gist.github.com/mflatt/887034c62172c266584492279854ad33); the `<`s mark significant slowdowns:

```
    v8.6  new    v8.6  new
    comp  comp    app  app'

compose one `identity`
f     90    66     5    5
f0   104   103     24   25
f1   159 < 291     82   84
f2   155 < 632    177  183

compose four `list`s
f    367   246     43   43
f0   387   281     64   65
f1   442 < 490    124  126
f2   442 < 840    217  224
```

With more small adjustments, I can remove a lot of that remaining overhead, especially the still-large gap between optional and required keyword arguments:

```
compose one `identity`
f     90    66      5    5
f0   104   103     24   25
f1   159 < 255     82   84
f2   155 < 296    177  183

compose four `list`s
f    367   246     43   43
f0   387   281     64   65
f1   442   457    124  126
f2   442 < 496    217  224
```

That these times are in milliseconds for 2M composes/applications. While there's still a significant slowdown in `compose` with keyword arguments, it's on a scale that is consistent with using keyword arguments with higher-order functions, so I think this is ready to merge.

PR #4420, opened 2022-08-29
Add `const*`.

NoahStoryM:
```racket
(for ([const
       (in-list
        (list
         (λ (c)
           (define (const . _) c)
           (make-keyword-procedure const const))
         (case-lambda
           [()    (define (const . _) (values))
                  (make-keyword-procedure const const)]
           [(c)   (define (const . _) c)
                  (make-keyword-procedure const const)]
           [(c d) (define (const . _) (values c d))
                  (make-keyword-procedure const const)]
           [c*    (define (const . _) (apply values c*))
                  (make-keyword-procedure const const)])
         (case-lambda
           [()    (thunk* (values))]
           [(c)   (thunk* c)]
           [(c d) (thunk* (values c d))]
           [c*    (thunk* (apply values c*))])))])
  (time (for ([i (in-range 10e7)]) (const 'foo)))
  (define foo (const 'foo))
  (time (for ([i (in-range 10e7)]) (foo 1 2 3 4)))
  (newline))
```

racket [cs]:
```
cpu time: 3011 real time: 3018 gc time: 65
cpu time: 4351 real time: 4356 gc time: 0

cpu time: 2991 real time: 3003 gc time: 36
cpu time: 4412 real time: 4425 gc time: 0

cpu time: 3080 real time: 3095 gc time: 42
cpu time: 4402 real time: 4415 gc time: 0
```


samth:
Can you explain more what the program in the comment is supposed to indicate?

NoahStoryM:
Ah, I just wanted to show that the `thunk*` version performs as well as the original.

racket/collects/racket/function.rkt
```diff
@@ -9,9 +9,12 @@
 
 (define (identity x) x)
 
-(define (const c)
-  (define (const . _) c)
```
sorawee, line 13: I think the original intention is to make it so that `(object-ame (const 1))` is `const`. So instead of returning `thunk*` directly, you might want to bind it to a variable `const` first, and then return it.

NoahStoryM: `thunk*` seems to already be able to infer names:

```racket
Welcome to Racket v8.6 [cs].
> (object-name (const))
'const
> (object-name (const 1))
'const
> (object-name (const 1 1))
'const
```

sorawee: Oh nice. Nvm then. (I tested from the benchmark variant that you posted as a comment, and that variant doesn't work)

NoahStoryM:
Instead of modify `const`, it may be more appropriate to add `const*`:

Before:
```racket
Welcome to Racket v8.6 [cs].
> (curry const 9)
#<procedure:const>
```

After:
```racket
Welcome to Racket v8.6 [cs].
> (curry const 9)
#<procedure:curried:const>
```

NoahStoryM:
Maybe a little off topic, I wonder if Racket is willing to support `global`?
````racket
(define global (λ (c) (thunk c)))
````

`global` can be used to construct [global elements](https://en.wikipedia.org/wiki/Global_element), which treats any value as no argument function.

In Racket, some higher-order functions that accept procedures as their arguments may require `global` (I've met this case https://github.com/countvajhula/qi/pull/69).

And if Racket is willing to support `global`, do I need to open a new PR? Or just add it here?

**Comment by samth** [comment]
How is `global` different than `thunk`? Just that it can be used in argument position?

**Comment by NoahStoryM** [comment]
I think the difference between `global` and `thunk` is the same as the difference between `const` and `thunk*` .

**Comment by NoahStoryM** [comment]
Thinking about it more, I did met a case that only `thunk` was not enough when I used Qi.

 One advantage of using global elements is that the programmer only needs to focus on functions, aka point free style.  Racket cannot bind values ​​to a variable, so `thunk`s are usually used to store values.  `call-with-values`​​ accepts `thunk` as arguments.

 If we want to define a new data type based on `values`, we must first convert `values` ​​to `thunk`, and then store it in the new data type.  For example we want to define `covalues` ​​as `values` tagged with natural number:
```racket
 (struct covalues ​​(thk tag))
```

 And from a composition perspective, `global*` seems to be the only way to turn `values` ​​into `thunk`.
```racket
 (~> (1 2 3) (-< global* (const 0)) covalues)
```

capfredf:
I am confused. 

`(define global (λ (c) (thunk c)))` is a function that returns a thunked *value* that it is given, i.e. before being thunked, the value is an evaluation result of some argument. This function doesn't have the same behavior as the macro `thunk`. 

e.g.:

```
(global (begin (displayln
                  "42") 1))
```
prints 42 and returns a nullary function, whose application evaluates to 1.

```
(thunk
   (begin (displayln 10) 1)))
```

simply returns a nullary function.

NoahStoryM:
Yeah, I think that's an important difference between them. thunk is lazy, global isn't.

gus-massa:
Small side comment about the benchmark:

Defining the functions in a list and then using `in-list` prevents a lot of the optimizations. In many cases the functions can be inlined. So I made a new version that defines each function independently, and I added `collect-garbage`. The code is at the bottom, not that **all** the loops use 1e6 instead of 1e7.

    cpu time: 1875 real time: 1901 gc time: 96
    cpu time: 406 real time: 400 gc time: 0

    cpu time: 1906 real time: 1918 gc time: 78
    cpu time: 391 real time: 397 gc time: 0

    cpu time: 2547 real time: 2532 gc time: 0
    cpu time: 391 real time: 400 gc time: 0

    cpu time: 1875 real time: 1891 gc time: 46
    cpu time: 469 real time: 475 gc time: 0

    cpu time: 1875 real time: 1888 gc time: 31
    cpu time: 468 real time: 467 gc time: 0

    cpu time: 1859 real time: 1907 gc time: 31
    cpu time: 469 real time: 475 gc time: 0

So ..., whatever optimizations the compiler is applying, it's only important for the last definition.


    #lang racket

    (for ([const
           (in-list
            (list
             (λ (c)
               (define (const . _) c)
               (make-keyword-procedure const const))
             (case-lambda
               [()    (define (const . _) (values))
                      (make-keyword-procedure const const)]
               [(c)   (define (const . _) c)
                      (make-keyword-procedure const const)]
               [(c d) (define (const . _) (values c d))
                      (make-keyword-procedure const const)]
               [c*    (define (const . _) (apply values c*))
                      (make-keyword-procedure const const)])
             (case-lambda
               [()    (thunk* (values))]
               [(c)   (thunk* c)]
               [(c d) (thunk* (values c d))]
               [c*    (thunk* (apply values c*))])))])
      (collect-garbage)
      (time (for ([i (in-range 10e6)]) (const 'foo)))
      (define foo (const 'foo))
      (time (for ([i (in-range 10e6)]) (foo 1 2 3 4)))
      (newline))

    (define const1 (λ (c)
                     (define (const . _) c)
                     (make-keyword-procedure const const)))
    (define const2 (case-lambda
                     [()    (define (const . _) (values))
                            (make-keyword-procedure const const)]
                     [(c)   (define (const . _) c)
                            (make-keyword-procedure const const)]
                     [(c d) (define (const . _) (values c d))
                            (make-keyword-procedure const const)]
                     [c*    (define (const . _) (apply values c*))
                            (make-keyword-procedure const const)]))
    (define const3 (case-lambda
                     [()    (thunk* (values))]
                     [(c)   (thunk* c)]
                     [(c d) (thunk* (values c d))]
                     [c*    (thunk* (apply values c*))]))

    (let ()
      (collect-garbage)
      (time (for ([i (in-range 10e6)]) (const1 'foo)))
      (define foo (const1 'foo))
      (time (for ([i (in-range 10e6)]) (foo 1 2 3 4)))
      (newline))

    (let ()
      (collect-garbage)
      (time (for ([i (in-range 10e6)]) (const2 'foo)))
      (define foo (const2 'foo))
      (time (for ([i (in-range 10e6)]) (foo 1 2 3 4)))
      (newline))

    (let ()
      (collect-garbage)
      (time (for ([i (in-range 10e6)]) (const3 'foo)))
      (define foo (const3 'foo))
      (time (for ([i (in-range 10e6)]) (foo 1 2 3 4)))
      (newline))

samth:
I continue to be a little puzzled about why these functions are needed. Generalizing `const` to handle multiple arguments/return multiple values seems potentially useful, but why would we then keep the old version? And what is the reason for `global`, which is just like `const` but less useful, as far as I can tell.

NoahStoryM:
`global element` is a category theory concept that allows us to use arbitrary values as functions, so I guess for some higher-order functions `const` can not replace `global` (`relay*` is just an example).

And I feel that since `racket/function` provides `thunk` and `thunk*`, it already shows that in some cases `thunk` is irreplaceable (maybe better error messages or check the input). `global` and `const` are somewhat consistent with `thunk` and `thunk*`. So I think `global` will be useful in some scenarios other than `arity-checking`.

NoahStoryM:
Maybe need more time to consider whether supporting `global` is necessary, I've removed them from this PR.

mflatt:
Finally merged, with two small changes: an update to the `history` note and reverting a change to `thunk`.

The `thunk` change doesn't seem like it should be needed, and no tests failed after I changed it back, but let me know if I'm missing something.

Thanks!

NoahStoryM:
> The thunk change doesn't seem like it should be needed, and no tests failed after I changed it back, but let me know if I'm missing something.

Ah, I forgot to revert `thunk` when I removed `global`. The thunk change was used to ensure `global` has the correct name:

Initial:
```racket
Welcome to Racket v8.6 [cs].
> (define global (λ (c) (thunk c)))
> (global 123)
#<procedure>
```

With the change:
```racket
Welcome to Racket v8.6 [cs].
> (define global (λ (c) (thunk c)))
> (global 123)
#<procedure:global>
```

PR #4422, opened 2022-08-29
add support to get-module-code to use a list of compiled/ subdirectories

rfindler:
This change allows the `subpath` argument to `get-module-code` to be a list of subpaths. One thing it doesn't do is change the default argument to `(use-compiled-file-paths)`. The default is currently, effectively `(list (car (use-compiled-file-paths)))` which seems a bit strange but for now this is the conservative, backwards compatible change here so I could  ask what people thought or if anyone remembered more about the why. 

Looking back through the git history, I see that in 2009 (in d4e40e0b256ab) the `#:sub-path` argument was added with a default of just `"compiled"` and in 2018 (in 59ef2543182e0445, the monumental commit that brought the expander out of C into Racket) the default value changed from `"compiled"` to be the current default. So maybe there wasn't any particular reason this functionality wasn't added, just other stuff was going on. If that's the case, it may make sense to change the default to just be `(use-compiled-file-paths)`. I'm not sure of this history of `use-compiled-file-paths`, but my guess is that it is older than both of these changes.

Background: DrRacket wants to recur over the compiled files to build the module graph (using `get-module-code`) and with the current pile of code, we end up looking only in a place where there are no .zo files, making this a much slower process than it needs to be. This change enables DrRacket to pass `#:sub-path (use-compiled-file-paths)` to `get-module-code`, speeding things up considerably.


samth:
Here's the original change by @mflatt that changed the default from `"compiled"`: https://github.com/racket/racket7/commit/eb37d2c41d04be04342ff64e73a125466b35454a

pkgs/racket-doc/syntax/scribblings/modcode.scrbl
```diff
@@ -33,7 +39,9 @@ submodule.
 
 The @racket[compiled-subdir] argument defaults to @racket[(default-compiled-sub-path)];
 it specifies the sub-directory to search for a compiled version of the
-module. The @racket[roots] list specifies a compiled-file search path
+module. if @racket[compiled-subdir] is a list, then the first directory
```
samth, line 42:
```suggestion
module. If @racket[compiled-subdir] is a list, then the first directory
```

racket/collects/syntax/modcode.rkt
```diff
@@ -7,13 +7,17 @@
                      get-module-path
                      get-metadata-path))
 
+(define subpath/c
+  (or/c (and/c path-string? relative-path?)
+        (listof (and/c path-string? relative-path?))))
```
samth, line 12: I think this should require a non-empty list (the implementation takes `(car null)` if you give it an empty list).

rfindler: Thanks!

mflatt:
Based on a grep for uses of `get-module-code`, I think it's likely a good idea to make `(use-compiled-file-paths)` the default. I don't recall making and explicit choice to have it consult a single directory, and the commit to use the first of `(use-compiled-file-paths)` looks like a common pattern used to allow a CS+BC overlay, which was the concern in 2018.

racket/collects/syntax/private/modcode-noctc.rkt
```diff
@@ -91,6 +91,15 @@
            (and (file-exists? p) p))
          (apply build-path (reroot-path* base (car roots)) args))]))
 
+(define (get-metadata-path/multiple-sub-paths #:roots roots path0-base sub-paths . args)
+  (define candidates
+    (for/list ([sub-path (in-list (if (list? sub-paths) sub-paths (list sub-paths)))])
+      (apply get-metadata-path #:roots roots path0-base sub-path args)))
+  (or (for/first ([candidate (in-list candidates)]
```
mflatt, line 98: I'd fuse the two `for` iterations here into a `for*`, instead of computing candidates that may not be needed. More importantly, to be consistent with the compiled-flie load handler, the nesting should be the other way around: each compiled directory within a root, instead of each root within a compiled directory.

racket/collects/syntax/modcode.rkt
mflatt: I think it should accept the empty list and not fail, since `use-compiled-file-paths` is allowed to be empty.

rfindler:
Okay, I've made another attempt in order to check things in the right order. Now what I'm doing is iterating over the potential sub-paths in order and, as soon as any one of the candidates in that sub-path (ie zo, so, alt-zo, or alt-so) exists, we stop the iteration and ask `check` which one to prefer. If none of them exist, we use the ones that we got from the first sub-path. Is this the right strategy?

(Also I fixed the empty-list thing by treating the empty list the same as if they'd just passed `"compiled"` because that's what the code used to be doing when `use-compiled-file-paths` was the empty list.)

racket/collects/syntax/private/modcode-noctc.rkt
mflatt: `get-metadata-path/multiple-sub-paths` is no longer used and can be removed, right?

mflatt:
The search loop looks basically right.

To record a bit of our in-person discussion: the old definition of `try-alt?` looks wrong. Also, to handle the case where no sources are present and a compiled-file name uses the alt suffix and be consistent with the default load handler, then there needs to be something like `try-main?`:

```
  (define try-main? (or main-src-date (not alt-src-date)))
  (define try-alt? (and alt-src-file (or alt-src-date (not main-src-date))))
```

The intent here is that the presence of a ".rkt" or ".ss" constrains the search, but when neither is present, we look for both "_rkt.zo" and "_ss.zo" files.

> (Also I fixed the empty-list thing by treating the empty list the same as if they'd just passed "compiled" because that's what the code used to be doing when use-compiled-file-paths was the empty list.)

I think this behavior should probably change. The old behavior incorrectly used a pattern for picking a destination for compiled files, where I applied that pattern too blindly; here, we want to pick a source. So, when the list is empty, i think the right thing is to run 0 iterations of the search loop (which means always using source).

rfindler:
With @mflatt's help I've managed to extract the code that the default `current-load/use-compiled` uses to make a decision about which file it is going to use (zo/so/rkt/ss all that stuff; a bit more than 100 lines of fiddly choices) and am looking into replacing the logic in `get-module-path` (which `get-module-code` calls) with a call to that extracted function (the extracted function will be private somewhere). The goal is that that fiddly decision making is the same between the default `current-load/use-compiled` handler and `get-module-code`.

The two main impediments to just dropping the code in are:

- the `choose` argument to `get-module-code` (which doesn't make a lot of sense, given how the code in the default `current-load/use-compiled` works)
- the fact that the `compiled-subdir` argument to `get-module-code` is a single path but, as discussed above, we seem to be happy to change the default here; it seems more likely to fix bugs than introduce them.

So I've done an audit of the calls to `get-module-code` and `get-module-path` in the standard distribution to see what they are doing wrt to the `choose` argument. I'd welcome others' thoughts on these:

- [the call in `scribble/search`](https://github.com/racket/scribble/blob/master/scribble-lib/scribble/search.rkt#L131):
I think this code is probably not good. I think the reason it is passing an argument with #:choose is to avoid triggering a recompilation. But the current code can get stale values. I also think that it is swallowing a lot of errors which is also not good, but possibly a separate problem. 

  I think a better approach would be to call get-module-path (just the path-checking part, not the recompilation part) and see if a .zo file is found; if it is, try it, and if it isn't, then move on the way the code is currently moving on in the lines just below the link above. I'm not super confident about this one, tho.

- [the call in `compiler/embed`](https://github.com/racket/racket/blob/master/racket/collects/compiler/embed.rkt#L517): 
I think that the use choose here is the same as the default value, if we change `get-module-code` to use the implementation that the default `load/use-compiled` handler uses. That is, the only time that the code there returns something different than #f is when an so and a zo exist and the so is newer. In that case, however, I believe that the so will be preferred anyway, since so is chosen first by the default load handler. That said, even if it wasn't chosen first, then I think it shouldn't be chosen here, as we want the embedding executable to be the same as running the program anyway.... right?

- [the call in `compiler/cm`](https://github.com/racket/racket/blob/master/racket/collects/compiler/private/cm-minimal.rkt#L782):
This one it looks like is forcing things to always come from source. I think I can add an option to get-module-code that has that behavior pretty simply as it will translate into a boolean passed to the new helper function that's been pulled out and then change this code to supply that argument instead of what it is currently doing.

- [The call in `syntax/rerequire`](https://github.com/racket/racket/blob/master/racket/collects/racket/rerequire.rkt#L52):
  This one doesn't pass the troublesome arguments and I guess the change to its behavior that it would inherit by the changes suggested here would amount to bug fixes.

- [the calls in `rackunit/test-docs-complete`](https://github.com/racket/racket/blob/master/pkgs/racket-index/rackunit/docs-complete.rkt#L43-L45): Ditto `syntax/rerequire`

- [the call in the sandbox](https://github.com/racket/racket/blob/master/pkgs/sandbox-lib/racket/sandbox.rkt#L351): Ditto `syntax/rerequire`

- [the call that helps DrRacket's teachpack dialog detect conflicts](https://github.com/racket/htdp/blob/master/htdp-lib/lang/private/tp-dialog.rkt#L406): Ditto `syntax/rerequire`.

- There are calls in the typed racket test suite, r6rs, and xrepl that also seem likely to be improvements along the same lines.

- The calls in the macro stepper look like there's also passing the defaults and the new behavior/new defaults are likely to be improvements.

- The call in DrRacket's module browser is what triggered those whole thing; this is a big performance improvement for it.

Overall, my feeling is to simply deprecate the troublesome `choose` argument by just ignoring it, but I could also change the default value be some special value and keep the old code around, using the old code when the `choose` argument is not the default.

mflatt:
The suggested changes to existing `get-module-code` calls sound plausible to me.

rfindler:
> The suggested changes to existing `get-module-code` calls sound plausible to me.

Do you have opinions about how to change `get-module-code` itself? I think the pending is question is if we keep the old behavior when a non-default `choose` is given or if we just never keep the old behavior.

rfindler:
I've just updated the pull request with a version of the change that takes the behavior changes discussed above aren't backwards incompatibility but are bug fixes. If the audit of the main distribution is representative, this is probably reasonable.

rfindler:
Also, I have a question: should the version number change for this commit? (The `@history` might be misleading in the current state but I've done similarly approximate history updates in the past ... for better or worse.)

PS: I made two commits so if you want to have a look at the change, you can ignore one of them as it has only build products.

pkgs/racket-doc/syntax/scribblings/modcode.scrbl
```diff
@@ -52,10 +57,14 @@ The @racket[rkt-try-ss?] argument defaults to @racket[#t].  If it is not
 @racket[#f], then if @racket[path] ends in @filepath{.rkt}, then the
 corresponding file ending in @filepath{.ss} will be tried as well.
 
-The @racket[choose-proc] argument is a procedure that takes three
-paths: a source path, a @filepath{.zo} file path, and an extension path
-(for a non-@tt{_loader} extension). Some of the paths may not
-exist. The result should be either @racket['src], @racket['zo],
+If the @racket[allow-zo-so] argument is @racket[#f], then the code is
+always loaded from source, never from a @filepath{.zo } or @filepath{.so}
```
mflatt, line 61: Extra space after `.zo`. The "extension" suffix is not always `.so` (it's `.dylib` on Mac OS and `.dll` on Windows), so the word "extension" is probably more right there.

pkgs/racket-doc/syntax/scribblings/modcode.scrbl
```diff
@@ -7,23 +7,26 @@
 
 @defproc[(get-module-code [path path-string?]
                           [#:submodule-path submodule-path (listof symbol?) '()]
-                          [#:sub-path compiled-subdir0 (and/c path-string? relative-path?) (default-compiled-sub-path)]
-                          [compiled-subdir (and/c path-string? relative-path?) compiled-subdir0]
+                          [#:sub-path compiled-subdir0
+                           (or/c (and/c path-string? relative-path?)
+                                 (listof (and/c path-string? relative-path?)))
+                           (default-compiled-sub-path)]
```
mflatt, line 13: Is this meant to change to `(use-compiled-file-paths)`?

mflatt:
I think it's ok to not bump the version number here.

pkgs/racket-doc/syntax/scribblings/modcode.scrbl
rfindler: As in ".zo file or extension"?

pkgs/racket-doc/syntax/scribblings/modcode.scrbl
rfindler: Right! This also triggered a number of other smaller changes to the docs. I'll push after it builds and I look at it.

pkgs/racket-doc/syntax/scribblings/modcode.scrbl
mflatt: Yes. It looks like "dynamic extension" is the defined term, so `@tech{dynamic extension}` instead of just `extension`.

racket/collects/racket/private/choose-file-to-load.rkt
```diff
@@ -0,0 +1,177 @@
...
+             (let ([p (if try-main? path alt-path)])
+               (values (and expect-module
+                            (not try-main?)
+                            p) ;; current-module-declare-source
+                       'src ;; plain file
+                       (and (or (not (pair? expect-module))
+                                (file-exists? p)) ;; used to be `quiet-failure?`
+                            p)))]
+            [else
+             ;; here the load-handler should do nothing; this was a fallthru case in the other
```
mflatt, line 145: Nit: "in the other" makes sense at the moment, but won;t after the merge.

racket/collects/racket/private/choose-file-to-load.rkt
```diff
@@ -0,0 +1,177 @@
...
+(define-values (dll-suffix)
+  (system-type 'so-suffix))
+
+(define (is-compiled-file? p)
+  (and (file-exists? p)
+       (call-with-input-file* p
+         (λ (port)
+           (starts-with-compiled-object-bytes? port)))))
+
+(define (starts-with-compiled-object-bytes? p)
```
mflatt, line 157: Could this be the function "load-handler.rkt" and used from there, so there's not two (mostly) copies? There may be some obstacle to sharing that I'm not seeing.

mflatt (commented):
The `raco exe` test fails with these changes on the case that's meant to use "embed-me12.ss" (as opposed to "embed-me12.rkt", which doesn't exist). So, something about an alt name isn't getting back to `get-module-code`?

I also get an error trying to render the Macro Debugger documentation. Something goes wrong with the xexpr near the description of `check-requires`, but I think it's probably really about an unexpected error in an example.

My two other comments on the code are minor.

mflatt:
Forgot to mention: the `raco exe` test suite is "pkgs/compiler-test/tests/compiler/embed/test.rkt".

rfindler:
> The raco exe test fails with these changes on the case that's meant to use "embed-me12.ss" (as opposed to "embed-me12.rkt", which doesn't exist). So, something about an alt name isn't getting back to get-module-code?

This one's fixed with 3146888

rfindler:
The issue with the `check-requires` is that this change is required:

```
diff --git a/macro-debugger-text-lib/macro-debugger/analysis/private/util.rkt b/macro-debugger-text-lib/macro-debugger/analysis/private/util.rkt
index acaf203..be2c8d5 100644
--- a/macro-debugger-text-lib/macro-debugger/analysis/private/util.rkt
+++ b/macro-debugger-text-lib/macro-debugger/analysis/private/util.rkt
@@ -57,7 +57,7 @@
                   [path (values path null)])])
     (get-module-code path
                      #:submodule-path subs
-                     #:choose (lambda _ 'src)
+                     #:allow-zo-so? #f
                      #:compile (lambda (stx)
                                  (let-values ([(stx deriv) (trace/result stx expand #t)])
                                    (values (compile stx) deriv))))))
```

which actually points out that I didn't do the best job in my audit (as I missed that `#:choose` was being passed here).

Does this mean that I should figure out a way to support `choose`? I'm somewhat reluctant because it is adding a nontrivial knob on a pretty core and fiddly part of our codebase (that doesn't have a very clean specification). But backwards compatibility is serious so I'm not sure what is best.

racket/collects/racket/private/choose-file-to-load.rkt
rfindler: I tried to look into this a bit more and I'm not sure what's going on, actually. It looks like `linklet-directory-start` is exported from load-handler.rkt, but I can't find where it is used outside that file. If it isn't used, then I think it should be fine to move a bunch more code from load-handler.rkt to (the new) `racket/private/choose-file-to-load`. 

That said, I think the duplication is pretty small and the test case gives us above average coverage and confidence in `starts-with-compiled-object-bytes?`.

mflatt: I think it was used only by "handler.rkt". While it still seems better to me to restore the few lines in `racket/private/choose-file-to-load` to make it `linklet-bundle-or-directory-start` and share that — less about bugs and more about changing fewer places when the format inevitably changes — I'm ok with this organization.

mflatt:
Would it make sense to continue to use `choose`, but in a restricted way: by calling it once with three made-up paths to initialize the new `allow-zo-so?` argument if the result is just `'src`?

rfindler:
> Would it make sense to continue to use `choose`, but in a restricted way: by calling it once with three made-up paths to initialize the new `allow-zo-so?` argument if the result is just `'src`?

This seems to work out. I've pushed a commit that does that and the macro stepper's `check-requires` seems to work again.

racket/collects/racket/private/choose-file-to-load.rkt
rfindler: Okay, I see it now. It really isn't as much code as I thought it was going to be. (I'm imagining that these commits will be squashed into one before merging to master, just in case this is the last comment. I'm happy to do it, of course.)

mflatt:
LGTM

samth:
I see a number of the CI tests are failing. Is that expected? Also, can you squash these commits (and maybe write something explanatory for the commit message) before merging?

rfindler:
I've just noticed that the documentation for `get-module-path` points out that .so files cannot contain submodules (and thus it has a `submodule?` argument), but the code that was pulled out of the default value of `current-load/use-compiled` doesn't seem to depend on whether or not a submodule is being looked for. Is this comment obsolete?

PR #4428, opened 2022-09-06
[Document]: `>` can accept one argument.

NoahStoryM:


NoahStoryM:
> ... means it can accept one argument.

Yes
```racket
Welcome to Racket v8.6 [cs].
> (> 0)
#t
```

PR #4433, opened 2022-09-12
Use `~.a` to display error messages instead of `~e`.

NoahStoryM:
Use `~.a` to display error messages for better readability.

For example, `draft.rkt`:
```racket
#lang racket/base
(require rackunit)
(check-eq? (eq? 0) #t)
```

Before:
```
$ raco test --drdr draft.rkt
raco test: 0 "draft.rkt"
draft.rkt: raco test: non-empty stderr: #"--------------------\nERROR\nname:       check-eq?\nlocation:   draft.rkt:5:0\n\neq?: arity mismatch;\n the expected number of arguments does not match the given number\n  expected: 2\n  given: 1\n--------------------\n"
1 1 draft.rkt
1/1 test failures
```

After:
```
$ raco test --drdr draft.rkt
raco test: 0 "draft.rkt"
draft.rkt: raco test: non-empty stderr:
--------------------
ERROR
name:       check-eq?
location:   draft.rkt:5:0

eq?: arity mismatch;
  expected: 2
  given: 1
--------------------

1 1 draft.rkt
1/1 test failures
```

### DIFF
--- a/pkgs/compiler-lib/compiler/commands/test.rkt
+++ b/pkgs/compiler-lib/compiler/commands/test.rkt
@@ -306,8 +306,19 @@
                            ignore-stderr-patterns)
                     (and ignore-stderr
                          (regexp-match? ignore-stderr s)))
-          (parameterize ([error-print-width 16384])
-            (error test-exe-name "#<<non-empty stderr\n~.a\nnon-empty stderr\n" s))))
+          (define tag #"non-empty stderr")
+          (define n
+            (cond
+              [(and (equal? #"#<<" (subbytes s 0 3))
+                    (equal? tag (subbytes s 3 19))
+                    (for/first ([i (in-naturals 19)]
+                                [b (in-bytes (subbytes s 19))]
+                                #:when (eq? 10 b))
+                      i))
+               => (λ (i) (gensym (~a (subbytes s (+ 3 (bytes-length tag)) i))))]
+              [else '||]))
+          (parameterize ([error-print-width #x4000])
+            (error test-exe-name "#<<~a~a\n~.a\n~a~a" tag n s tag n))))
       (unless (zero? result-code)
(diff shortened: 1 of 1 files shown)


mflatt:
The motivation here is clear, but I went with `~s` originally to ensure that the error output is separated from `raco test`'s output. Would it be worth the effort to wrap the output as a here string? In other words: add `#<<`, a tag that doesn't appear in the output at the start of a line, and then a newline, and then add the same tag as a suffix.

(I also vaguely recall wanting to convert unprintable characters in error output to printable characters, and adding a here-string prefix and suffix won't do that. But that feature feels less important than delimiting output.)

NoahStoryM:
Here string version:

```
$ raco test --drdr draft.rkt
raco test: 0 "draft.rkt"
draft.rkt: raco test: #<<non empty-stderr
--------------------
ERROR
name:       check-eq?
location:   draft.rkt:3:0

eq?: arity mismatch;
  expected: 2
  given: 1
--------------------

non empty-stderr
1 1 draft.rkt
1/1 test failures
```

mflatt:
To make this reliably delimit the output text, I think the implementation needs to check whether the tag `non-empty stderr` exists after an empty line in the output to be printed, and if so, maybe add a random number to the end of the tag and try again.

rfindler:
I'm not sure I'm getting what you're saying, but isn't the concern that the text might contain the closing tag somewhere in the middle of it, thereby prematurely ending the here string? (Also that function crashes on short byte strings.)

NoahStoryM:
> the text might contain the closing tag somewhere in the middle of it, thereby prematurely ending the here string? (Also that function crashes on short byte strings.)

Ah, I was thinking too complicatedly.  The situation I thought before was to wrap a here string with another here string.  Just checking the suffix tag should work. I'll try it later.

PR #4435, opened 2022-09-19
Support required methods in racket/generic

### LINKED ISSUE #1554: error messages for implementing gen:equal+hash are opaque
Consider

```racket
(struct wuznub (x)
  #:transparent
  #:methods gen:equal+hash
  [(define (equal-proc a b _) (= (wuznub-x a) (wuznub-x b)))
   (define (hash-proc a _) (wuznub-x a))
   (define (hash-proc2 a _) (wuznub-x a))])
```

This produces the error message

```
guard-for-prop:gen:equal+hash: contract violation
  expected: 
   (vector/c (procedure-arity-includes/c 3)
             (procedure-arity-includes/c 2)
             (procedure-arity-includes/c 2))
  given: '#(#<procedure:equal-proc> #<procedure:hash-proc> #f)
  context...:
   /home/rntz/c/rkt/useless-error-message.rkt: [running body]
```

The most useful problem diagnosis is that one of the methods is named `hash-proc2` when it should be named `hash2-proc`. Neither of those names even occurs in the error message. The error message is very unhelpful in directing a user towards fixing their mistake.

sorawee:
This PR makes it possible to mark some methods in a generic interface as "required". Required methods must be implemented in struct's #:methods clauses. If they are not implemented, a compile time error is raised.

The PR also switches gen:equal+hash, gen:equal-mode+hash, gen:custom-write, and gen:stream to use this new protocol, mandating that methods must be implemented. This allows the error message to be much more better.

Fixes #1554

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/generic.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/generic.scrbl
@@ -22,7 +22,8 @@ a structure type are defined using the @racket[#:methods] keyword
                 (code:line #:fallbacks [fallback-impl ...])
                 (code:line #:defined-predicate defined-pred-id)
                 (code:line #:defined-table defined-table-id)
-                (code:line #:derive-property prop-expr prop-value-expr)]
+                (code:line #:derive-property prop-expr prop-value-expr)
+                (code:line #:requires [required-method-id ...])]
                [kw-formals* (arg* ...)
                             (arg* ...+ . rest-id)
                             rest-id]
@@ -109,6 +110,11 @@ automatically implement this structure type property using the provided values.
 When @racket[prop-value-expr] is executed, each @racket[method-id] is bound to
 its specific implementation for the @tech{structure type}.
 
+The @racket[#:requires] option may be provided at most once.
+When it is provided, any instance of the generic interface
+@emph{must} supply an implementation of the specified @racket[required-method-id]s.
+Otherwise, a compile-time error is raised.
+
 If a value @racket[v] satisfies @racket[id]@racketidfont{?}, then @racket[v] is
 a @deftech{generic instance} of @racketidfont{gen:}@racket[id].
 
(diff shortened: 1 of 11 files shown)


racket/collects/racket/private/generic-methods.rkt
```diff
@@ -172,28 +190,46 @@
...
                (letrec-syntaxes+values
-                ([(method) (make-unimplemented 'method)] ...)
-                ()
-                (syntax-parameterize ([generic-method-inner-context #'gen])
-                  def ...
-                  (combine (implementation method) ...)))))))]
+                   ([(method) (make-unimplemented
+                               'method
+                               req-method
+                               (list (quote-syntax def) ...))] ...)
```
mflatt, line 219: It looks like this copies all definitions into a `make-unimplemented` for each method. If that's right, I can't help worry about the O(N^2)-ish expansion, even if N is normally small in practice. Maybe share the list, something like the code below?
```
[(method ...) 
 (let ([defs (list (quote-syntax def) ...)])
    (values (make-unimplemented 'method req-method defs) ...))]
```

pkgs/racket-doc/scribblings/reference/generic.scrbl
```diff
@@ -109,6 +110,11 @@ automatically implement this structure type property using the provided values.
 When @racket[prop-value-expr] is executed, each @racket[method-id] is bound to
 its specific implementation for the @tech{structure type}.
 
+The @racket[#:requires] option may be provided at most once.
```
mflatt, line 113: Now that we're ready to merge, this could use a `history` note.

mflatt (approved):
Small requests, but looks generally ready to merge.

Should there be a `history` note in `gen:equal+hash` and `gen:equal-mode+hash` that omitting a definition is now a syntax error?

racket/collects/racket/private/generic-methods.rkt
sorawee: Just curious. Say, if `letrec-syntaxes+values` doesn't support multiple values, would wrapping it with `(let-syntax ([defs (list (quote-syntax def) ...)]) ...)` and then using `(syntax-local-value #'defs)` be a good sharing mechanism?

sorawee:
Done, and all checks passed.

racket/collects/racket/private/generic-methods.rkt
mflatt: Yes, I think that would work.

PR #4437, opened 2022-09-21
[and/c | or/c]: ignore `any/c` | `none/c`.

NoahStoryM:
Before:
```racket
Welcome to Racket v8.6 [cs].
> (or/c #t #f (or/c))
(or/c #t #f (or/c))
> (or/c #t #f (or/c) 'x)
(or/c #t #f (or/c) 'x)
> (first-or/c #t #f (first-or/c) 'x)
(first-or/c #t #f (first-or/c) 'x)
> (and/c any/c)
(and/c any/c)
> (and/c any/c any/c)
(and/c any/c any/c)
> (and/c number? (and/c) integer?)
(and/c number? any/c integer?)
```

After:
```racket
Welcome to Racket v8.6 [cs].
> (or/c #t #f (or/c))
boolean?
> (or/c #t #f (or/c) 'x)
(or/c #t #f 'x)
> (first-or/c #t #f (first-or/c) 'x)
(first-or/c #t #f 'x)
> (and/c any/c)
any/c
> (and/c any/c any/c)
any/c
> (and/c number? (and/c) integer?)
(and/c number? integer?)
```

And the path error in `contract/private/types.rkt` seems to have no effect on TR, does the current version of TR still need this file to set up information?

By the way, I didn't find `define/subexpression-pos-prop/name` in the [document](https://docs.racket-lang.org/search/index.html?q=define%2Fsubexpression-pos-prop).

### DIFF
--- a/racket/collects/racket/contract/private/and.rkt
+++ b/racket/collects/racket/contract/private/and.rkt
@@ -190,101 +190,104 @@
      (identifier? #'x)
      #'and/c]))
 
-(define/subexpression-pos-prop/name real-and/c-name (and/c . raw-fs)
-  (let ([contracts (coerce-contracts 'and/c raw-fs)])
-    (cond
-      [(null? contracts) any/c]
-      [(andmap flat-contract? contracts)
-       (define preds (map flat-contract-predicate contracts))
-       (cond
-         [(and (pair? (cdr preds))
-               (null? (cddr preds)))
-          (cond
-            [(eq? (car preds) real?)
-             (define second-pred (cadr preds))
-             (cond
-               [(eq? second-pred negative?)
-                (renamed-<-ctc 0 `(and/c real? negative?))]
-               [(eq? second-pred positive?)
-                (renamed->-ctc 0 `(and/c real? positive?))]
-               [else
-                (define second-contract (cadr contracts))
-                (cond
-                  [(equal? (contract-name second-contract) '(not/c positive?))
-                   (renamed-between/c -inf.0 0 `(and/c real? (not/c positive?)))]
-                  [(equal? (contract-name second-contract) '(not/c negative?))
-                   (renamed-between/c 0 +inf.0 `(and/c real? (not/c negative?)))]
-                  [else
-                   (make-first-order-and/c contracts preds)])])]
-            [(or (eq? (car preds) exact-nonnegative-integer?)
-                 (eq? (car preds) natural?)
-                 (eq? (cadr preds) exact-nonnegative-integer?)
-                 (eq? (cadr preds) natural?))
-             (define other (if (procedure? (car preds)) (cadr contracts) (car contracts)))
-             (cond
-               [(between/c-s? other)
-                (define other-low (between/c-s-low other))
-                (define other-high (between/c-s-high other))
-                (integer-in (exact-ceiling (max 0 (if (= other-low -inf.0) 0 other-low)))
-                            (if (= other-high +inf.0) #f (exact-floor other-high)))]
-               [else (make-first-order-and/c contracts preds)])]
-            [(or (eq? (car preds) exact-positive-integer?)
-                 (eq? (cadr preds) exact-positive-integer?))
-             (define other (if (procedure? (car preds)) (cadr contracts) (car contracts)))
-             (cond
-               [(between/c-s? other)
-                (define other-low (between/c-s-low other))
-                (define other-high (between/c-s-high other))
-                (integer-in (exact-ceiling (max 1 (if (= other-low -inf.0) 1 other-low)))
-                            (if (= other-high +inf.0) #f (exact-floor other-high)))]
-               [else (make-first-order-and/c contracts preds)])]
-            [(or (eq? (car preds) exact-integer?)
-                 (eq? (cadr preds) exact-integer?))
-             (define other (if (procedure? (car preds)) (cadr contracts) (car contracts)))
-             (cond
-               [(between/c-s? other)
-                (define other-low (between/c-s-low other))
-                (define other-high (between/c-s-high other))
-                (integer-in (if (= other-low -inf.0) #f (exact-ceiling other-low))
-                            (if (= other-high +inf.0) #f (exact-floor other-high)))]
-               [else (make-first-order-and/c contracts preds)])]
-            [else
-             (make-first-order-and/c contracts preds)])]
-         [(and (pair? (cdr preds))
-               (pair? (cddr preds))
-               (null? (cdddr preds)))
-          (cond
-            [(or (eq? (car preds) exact-integer?)
-                 (eq? (cadr preds) exact-integer?)
-                 (eq? (caddr preds) exact-integer?))
-             (define lb #f)
-             (define ub #f)
-             (for ([ctc (in-list contracts)])
-               (cond
-                 [(between/c-s? ctc)
-                  (define lo (between/c-s-low ctc))
-                  (define hi (between/c-s-high ctc))
-                  (cond
-                    [(and (= lo -inf.0) (integer? hi))
-                     (set! ub (inexact->exact hi))]
-                    [(and (= hi +inf.0) (integer? lo))
-                     (set! lb (inexact->exact lo))])]
-                 [(</>-ctc? ctc)
-                  (define x (</>-ctc-x ctc))
-                  (when (integer? x)
-                    (cond
-                      [(<-ctc? ctc) (set! ub (- (inexact->exact x) 1))]
-                      [(>-ctc? ctc) (set! lb (+ (inexact->exact x) 1))]))]))
-             (cond
-               [(and lb ub)
-                (integer-in lb ub)]
-               [else (make-first-order-and/c contracts preds)])]
-            [else (make-first-order-and/c contracts preds)])]
-         [else
-          (make-first-order-and/c contracts preds)])]
-      [(andmap chaperone-contract? contracts)
-       (make-chaperone-and/c contracts)]
-      [else (make-impersonator-and/c contracts)])))
+(define/subexpression-pos-prop/name and/c-name and/c
+  (case-lambda
+    [() any/c]
+    [raw-f*
+     (define raw-fs (remq* (list any/c) raw-f*))
+     (case (length raw-fs)
+       [(0) any/c]
+       [(1) (coerce-contract 'and/c (car raw-fs))]
+       [else
+        (define contracts (coerce-contracts 'and/c raw-fs))
+        (cond
+          [(andmap flat-contract? contracts)
+           (define preds (map flat-contract-predicate contracts))
+           (cond
+             [(and (pair? (cdr preds))
+                   (null? (cddr preds)))
+              (cond
+                [(eq? (car preds) real?)
+                 (define second-pred (cadr preds))
+                 (cond
+                   [(eq? second-pred negative?)
+                    (renamed-<-ctc 0 `(and/c real? negative?))]
+                   [(eq? second-pred positive?)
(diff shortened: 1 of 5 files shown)


rfindler:
Should `(or/c)` come out as `none/c`? (Then maybe some of these changes will amount to just looking for `none/c`?)

NoahStoryM:
> Should (or/c) come out as none/c? (Then maybe some of these changes will amount to just looking for none/c?)

Treating `(or/c)` and `(first-or/c)` as `none/c` has no effect since nones (`none/c`, `(or/c)`, `(first-or/c)`) are ignored by `remq*`. And I'm not sure if this change will work, maybe `(or/c)` and `(first-or/c)` are used somewhere.

rfindler:
I'm not following the comment about remq. And if `(or/c)` is used somewhere, then that seems fine; it accepts the same values as `none/c`, right?

NoahStoryM:
I mean maybe some places already use `'(or/c)` as contract-name.  And on the other hand racket provides `make-none/c` to make new none, so I guess making `(or/c)` evaluate to `none/c` may affect some code.

rfindler:
That's definitely true. We could do something similar to the way that `any/c` works (c.f. `named-any/c`), where there are many "names" for `none/c` but all of answer true to `none/c?` (or some other predicate) and then use that predicate when removing them from arguments?

NoahStoryM:
For `first-or/c` this change I think is feasible.

But if `or/c` does the same thing, I don't know how to make `(opt/c (or/c ...))` return the same name as `(or/c ...)`:

````racket
Welcome to Racket v8.6 [cs].
> (or/c 1 2 (make-none/c 'any/c))
(or/c 1 2)
> (opt/c (or/c 1 2 (make-none/c 'any/c)))
#<opt-chaperone-contract: (or/c 1 2 any/c)>

File: racket/collects/racket/contract/private/misc.rkt [hunk 1]
```diff
@@ -666,34 +665,21 @@
   (define env (contract-random-generate-get-current-environment))
   (λ () (random-any/c env fuel)))
 
-(define-struct any/c ()
+(define-struct any/c (name)
```
NoahStoryM, line 668: Before:
```racket
Welcome to Racket v8.6 [cs].
> (object-name (or/c any/c any/c any/c))
'named-any/c
```

After:
```racket
Welcome to Racket v8.6 [cs].
> (object-name (or/c any/c any/c any/c))
'any/c
```

NoahStoryM:
I wonder if `racket/contract` is willing to provide `make-any/c` (as the dual of `make-none/c`) ?
Thus `premissive/c` can be described as:
> If `(permissive-xexprs)` is `#t`, then equivalent to `(make-any/c 'permissive)` ~`any/c`~, otherwise equivalent to `(make-none/c 'permissive)`.

And also `prop:any/c?` and `prop:none/c?` ?

rfindler:
> I wonder if racket/contract is willing to provide make-any/c (as the dual of make-none/c) ?

I don't think that's necessary. 

My thinking is that `(or/c)` would actually return `none/c` but the name of *that* `none/c` would be still be `(or/c)`. So we'd have `(contract-name (or/c))` returning `'(or/c)` even though, internally, it is a `none/c`. 

This is the setup for the way `any/c` works for exactly the reasons that you're talking about above.

rfindler:
I think `none/c?` is an internal detail for now. But if there is some contract outside the contract library can benefit from it the way or/c is, then we could provide it. There is always a worry about adding extra names to widely-used libraries, tho.

racket/collects/racket/contract/private/misc.rkt
rfindler: I'm not really sure what's going on here (I can't tell from what I'm reading on github), but I see a diff where a `name` field was added to `any/c` and I don't think we want that. We should stick with using `named-any/c` if we want a name.

racket/collects/racket/contract/private/prop.rkt
```diff
@@ -620,7 +620,7 @@
   (make-struct-type-property 'prop:any/c))
 
 ;; this property's value isn't looked at; it is just a signal
-;; that the contract accepts any value
+;; that the contract accepts none value
```
rfindler, line 623: Maybe "no" instead of "none"?

racket/collects/racket/contract/private/and.rkt
```diff
@@ -190,101 +191,108 @@
...
+     (case (length raw-args)
+       [(0) any/c]
+       [(1) (coerce-contract 'and/c (car raw-args))]
+       [else
+        (define args (coerce-contracts 'and/c raw-args))
+        (define the-and/c
           (cond
-            [(eq? (car preds) real?)
-             (define second-pred (cadr preds))
+            [(andmap flat-contract? args)
```
rfindler, line 206: Does putting this guard on the code change the behavior of some of the following cases?

rfindler: Oh, this was here before; the diff confused me.

racket/collects/racket/contract/private/and.rkt
```diff
@@ -190,101 +191,108 @@
...
-       (define preds (map flat-contract-predicate contracts))
-       (cond
-         [(and (pair? (cdr preds))
-               (null? (cddr preds)))
+(define/subexpression-pos-prop/name and/c-name and/c
+  (case-lambda
+    [() any/c]
+    [raw-arg*
+     (define raw-args (remove-duplicates (filter-not prop:any/c? raw-arg*) eq?))
+     (case (length raw-args)
```
rfindler, line 199: Let's not do this. Instead do `(null? (cdr contracts))` as a new second case.

racket/collects/racket/contract/private/misc.rkt
```diff
@@ -666,34 +672,18 @@
   (define env (contract-random-generate-get-current-environment))
   (λ () (random-any/c env fuel)))
 
-(define-struct any/c ()
+(struct any/c ()
```
rfindler, line 675: Is removing `define-` here necessary for the rest of the things that are happening in this PR? If not, let's leave it as is. (I would like to keep the history very clean here as I have used it in the past and it's been helpful.)

racket/collects/racket/contract/private/misc.rkt
```diff
@@ -706,19 +696,20 @@
    (none/c-name ctc)
    val))
 
-(define-struct none/c (name)
+(struct none/c (name)
```
rfindler, line 699: Ditto for this change.

racket/collects/racket/contract/private/misc.rkt
```diff
@@ -706,19 +696,20 @@
...
   #:property prop:custom-write custom-write-property-proc
-  #:omit-define-syntaxes
+  #:property prop:none/c #f
   #:property prop:flat-contract
   (build-flat-contract-property
    #:trusted trust-me
    #:late-neg-projection none-curried-late-neg-proj
-   #:stronger (λ (this that) #t)
-   #:equivalent (λ (this that) (none/c? that))
+   #:stronger (λ (this that) (prop:none/c? that))
```
rfindler, line 707: I think this change is incorrect. If there aren't test cases that this change breaks, let's add some.

racket/collects/racket/contract/private/orc.rkt
```diff
@@ -14,65 +15,75 @@
...
-              (make-impersonator-single-or/c name pred flat-contracts (car ho-contracts)))]
-         [else
-          (define name (apply build-compound-type-name 'or/c args))
-          (if (andmap chaperone-contract? ho-contracts)
-              (make-chaperone-multi-or/c name flat-contracts ho-contracts)
-              (make-impersonator-multi-or/c name flat-contracts ho-contracts))]))
-     (cond
-       [(ormap prop:any/c? args) (named-any/c (contract-name the-or/c))]
-       [else the-or/c])]))
+  (let ([none? (make-none/c '(or/c))])
```
rfindler, line 18: I don't think adding this let is a good idea. We are already using `none?` as somethign elsewhere here and it is a different thing than this. Also, the let change the indentation, making the diff more painful to work through; in this case we can do this without an indentation change by using `define`.

racket/collects/racket/contract/private/orc.rkt
```diff
@@ -14,65 +15,75 @@
...
-    [(x) (coerce-contract 'first-or/c x)]
-    [raw-args
-     (define args (coerce-contracts 'first-or/c raw-args))
-     (cond
-       [(andmap flat-contract? args)
-        (make-flat-first-or/c (make-flat-predicate args) args)]
-       [(andmap chaperone-contract? args)
-        (make-chaperone-first-or/c args)]
-       [else (make-impersonator-first-or/c args)])]))
+  (let ([none? (make-none/c '(first-or/c))])
```
rfindler, line 67: Ditto here with the `let`.

racket/collects/racket/contract/private/orc.rkt
```diff
@@ -13,66 +14,76 @@
...
-                [else
-                 (loop (cons arg ho-contracts) flat-contracts (cdr args))]))])))
-     (define pred (make-flat-predicate flat-contracts))
-     (define the-or/c
-       (cond
-         [(null? ho-contracts)
+  (case-lambda
+    [() or/c:none/c]
+    [raw-arg*
+     (define raw-args (remq* (list none/c or/c:none/c (first-or/c)) raw-arg*))
```
rfindler, line 22: Why not use a `none/c?` predicate here?

NoahStoryM: Because `(or/c ...)` and `(opt/c (or/c ...))` should have the same name.
https://github.com/racket/racket/pull/4437#issuecomment-1255747763

rfindler: I'm not following the answer. Definitely it should be the case that `opt/c` is less important than the regular use, so we should improve `opt/c`. Can you say more about what's going on here?

NoahStoryM: The `contract-name` of `(opt/c (or/c ...))` is checked in the test file [name.rkt](https://github.com/NoahStoryM/racket/blob/andor/pkgs/racket-test/tests/racket/contract/name.rkt#L11).

So if we use `none/c?`, this test will fail:
```racket
> (contract-name (or/c (make-none/c 'none) #t #f))
'boolean?
> (contract-name (opt/c (or/c (make-none/c 'none) #t #f)))
'(or/c none #t #f)
> (test-name 'boolean? (or/c (make-none/c 'none) #t #f))
FAILED (#<procedure:contract-name> #<opt-chaperone-contract: (or/c none #t #f)>)
       got (or/c none #t #f)
  expected boolean?
```

NoahStoryM: Here is the definition of the name of [(opt/c (or/c ...))](https://github.com/NoahStoryM/racket/blob/0be5a4c8938c3cc5b0da6e55bb209f58dcdc3538/racket/collects/racket/contract/private/opters.rkt#L147) .

rfindler: The two options are to adjust the test case to avoid checking `opt/c` or to improve `opt/c` to recognize this case and drop the call to `make-none/c`. I'm happy either way. I think the code that was there, dynamically doing the removal, isn't the best choice, so probably better not to make that more complicated. (If it has to get more complicated, it should be moved into a helper function and a call to the helper function should be generated.)

NoahStoryM: I haven't figured out how to improve `opt/c`, so I just avoid checking it.

samth:
What's the current status of this PR?

PR #4443, opened 2022-09-29
Chez Scheme: replace another internal `#3%fx+`

gus-massa:
As a followup of https://github.com/racket/racket/commit/7027a001ee1b9bbb5e48a5bfcf582cb23c0a80c8 , I think this should be changed too.

(By the way, it's a very strange code. I can't believe it's better than comparing with `(eq? x 1)` `(eq? x 0)` `(eq? x -1)`.)


mflatt:
Ah, I missed a copy. LGTM, and thanks!

PR #4445, opened 2022-09-29
Improvements for pass-printing output.

samth:


racket/src/cs/linklet.sls
```diff
@@ -192,7 +192,13 @@
...
                              [else #f])]))
 
-  (define gensym-on? (getenv "PLT_LINKLET_SHOW_GENSYM"))
+  (define gensym-on? (cond [(getenv "PLT_LINKLET_SHOW_GENSYM")
+			    =>
+			    (lambda (v)
+			      (case (string->symbol v)
+				[(pretty pretty/suffix) (string->symbol v)]
+				[else #t]))]
+			   [else #f]))
```
mflatt, line 201: I didn't know about `'pretty/suffx`! That seems like a better default than `#f`, and I can't imagine a reason to ever use `#f`. So, I'd changed the default for `gensym-on?`, change the name from `gensym-on?` to `gensym-mode`, and then not use `or` in the new use of `print-gensym`.

samth:
I think this is ready.

PR #4447, opened 2022-10-03
Fix bootstrapping via Racket from top-level makefile

jryans:
The build docs mention `RACKET_FOR_BOOTFILES` as a top-level makefile argument for bootstrapping via existing Racket, but this seems to have been renamed to `BOOTFILE_RACKET` somewhere along the way. This fixes the build docs and top-level makefile to follow along.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -176,7 +176,7 @@ BUILD_VARS = MAKE=$(MAKE) \
              CS_CROSS_SUFFIX="$(CS_CROSS_SUFFIX)" \
              RACKET="$(RACKET)" \
              PLAIN_RACKET="$(PLAIN_RACKET)" \
-             RACKET_FOR_BOOTFILES="$(RACKET_FOR_BOOTFILES)" \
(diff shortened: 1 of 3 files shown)


mflatt:
Thanks!

If anyone was using `RACKET_FOR_BOOTFILES` previously, then I imagine it would be better to support both. I'm guessing that you weren't using it before, though. Is that right? Assuming so, and also guessing that no one else was using it, this seems like the right change.

jryans:
> If anyone was using `RACKET_FOR_BOOTFILES` previously, then I imagine it would be better to support both. I'm guessing that you weren't using it before, though. Is that right?

Your guess is correct: this is not something I was using myself. I am instead using e.g. `configure —enable-scheme=chez` for cases where bootstrapping is needed.

I was just studying the bootstrapping pathways and happened to notice this discrepancy.

Assuming no one else has mentioned this `RACKET_FOR_BOOTFILES` issue, I would think it’s okay to remove like this PR currently does.

Issue #4448, opened 2022-10-03
`BOOTFILE_SCHEME` in Makefile.in seems unused


jryans:
A few weeks ago in 3f853dc05a8832cf1a10751ee12f673b99218a5f, @mflatt added support for a "reboot" mode to produce boot files from an old Scheme (which seems nice to have).

As part of this, `BOOTFILE_SCHEME` was added to `racket/src/cs/c/Makefile.in`...

https://github.com/racket/racket/blob/d71f64484d5228d89c6a8adce1576e2344206ca1/racket/src/cs/c/Makefile.in#L47-L50

...but this makefile template variable is not currently given a value, and nothing seems to attempt reading from it if it did have one.

Is there some missing code that wires up `BOOTFILE_SCHEME`, or should it perhaps be removed...?

mflatt:
Instead of `BOOTFILE_SCHEME`, `SCHEME` is automatically used in a way that's analogous to `BOOTFILE_RACKET` when the selected `scheme` reports a version that doesn't match the one being built. If I remember correctly, I experimented with `BOOTFILE_SCHEME` first, and I must have forgotten to clean up that part of the experiment.

So, I think it should go away, and the comment around `SCHEME` should be clarified to reflect a more relaxed compatibility requirement.

PR #4449, opened 2022-10-03
Clarify status of Chez Scheme bootstrapping across versions

jryans:
3f853dc05a8832cf1a10751ee12f673b99218a5f and related commits have added support for "rebooting" the boot files, which allows older Chez Scheme versions to be used (in particular, they don't have to exactly match the version being built together with Racket).

This clarifies various build docs to highlight this new ability. It also removes the (unused) makefile template variable `BOOTFILE_SCHEME`, as its functionality is actually part of the existing `SCHEME` / `--enable-scheme` pathways.

Fixes https://github.com/racket/racket/issues/4448

### DIFF
--- a/racket/src/README.txt
+++ b/racket/src/README.txt
@@ -430,9 +430,11 @@ Cross-compilation requires at least two flags to `configure`:
 
 For Racket CS, an additional flag is required:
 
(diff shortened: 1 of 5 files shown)


mflatt:
Thanks!

PR #4450, opened 2022-10-07
pkg/strip: handle read-only input

LiberalArtist:
A package directory supplied to the functions from `pkg/strip` might have had all of its write permission bits unset. Since `copy-file` preserves the permissions of the source file, we may end up with a read-only file that we want to overwrite (e.g. an `info.rkt` file). Explicitly setting `user-write-bit` before writing avoids this problem. Conservatively, we restore the original permissions when we are done.

### DIFF
--- a/racket/collects/pkg/strip.rkt
+++ b/racket/collects/pkg/strip.rkt
@@ -306,9 +306,8 @@
            #t
            new-mod*-subs))))
   (unless (eq? mod new-mod)
-    (call-with-output-file*
+    (call-with-output-file/writable
      new-p
-     #:exists 'truncate/replace
      (lambda (out) (write new-mod out)))))
 
 (define (fixup-local-redirect-reference p js-path #:user [user-js-path js-path])
@@ -340,9 +339,8 @@
                                       (string->bytes/utf-8 user-js-path)
                                       (subbytes s (+ delta end2)))]
                        [else s]))))
-    (call-with-output-file*
+    (call-with-output-file/writable
      p
-     #:exists 'truncate/replace
      (lambda (out) (write-bytes new-bstr out)))))
 
 ;; Used in binary[-lib] mode:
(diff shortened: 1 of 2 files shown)


racket/collects/pkg/strip.rkt
```diff
@@ -503,3 +500,25 @@
...
+  (define (run)
+    (call-with-output-file* pth
+      #:exists 'truncate/replace
+      proc))
+  (cond
+    [(file-exists? pth)
+     (define old-mode
+       (file-or-directory-permissions pth 'bits))
+     (define new-mode
+       (bitwise-ior old-mode user-write-bit))
```
mflatt, line 516: I think this won't work on Windows, where the only thing you can control via `file-or-directory-permissions` is whether writing is allowed for everyone (i.e., no user vs. group vs. other distinction).

So, `(if (eq? (system-type) 'windows) (bitwise-ior user-write-bit group-write-bit other-write-bit) user-write-bit)`? 

Also, maybe it would be better to try to change permissions only if `user-write-bit` is not already set.

LiberalArtist: Oh, I bet this is also the reason why tests for https://github.com/racket/racket/pull/4126 failed on Windows. I had seen in the documentation that Windows doesn't have a user/group/other distinction, but I guess I must have thought the non-user bits were just ignored or managed automatically somehow: I didn't realize the user of `file-or-directory-permissions` was responsible for keeping the bits consistent.

From the linked issues there, it seems like I got it wrong in the same way in https://github.com/racket/racket/commit/55b6cbdca1f36a4f37bab1519c1b658717d3cad2: I'll try to fix that, too.

LiberalArtist:
I've pushed a revised commit with the fix, plus a commit fixing https://github.com/racket/racket/commit/55b6cbdca1f36a4f37bab1519c1b658717d3cad2 the same way.

LiberalArtist:
@mflatt could you take another look at this? I'm happy to make other changes if needed!

PR #4452, opened 2022-10-07
Regexp docs: Distinguish multiline and dotall modes

sschwarzer:
From the first commit message:
```
Distinguish multiline and dotall modes

Before this change, the documentation spoke of "multi mode", with
`(?m...)` and `(?s...)` switching between multi mode and non-multi mode.

However, PCRE defines `m` and `s` separately, as "multiline" and
"dotall" modes, respectively. [1] Multiline mode only changes the
behavior of the anchors `^` and `$` [2], while dotall mode only changes
the behavior of `.` [3].

The Python documentation explains the modes a bit more comprehensibly.
[4]

I reversed the order of the `(?m...)` and `(?s...)` descriptions to
order them alphabetically by mode letter.

[1] https://www.pcre.org/original/doc/html/pcrepattern.html#SEC13
[2] https://www.pcre.org/original/doc/html/pcrepattern.html#SEC6
[3] https://www.pcre.org/original/doc/html/pcrepattern.html#SEC7
[4] https://docs.python.org/3/library/re.html#re.MULTILINE

P.S.: I saw the comment at the top of the file. Sorry for making the
table wider. If you have suggestions for different wordings to keep the
table narrower, please let me know.
```

The first commit contains the "actual" change, the second commit is only for realigning the table. So you probably want to review the commits separately.

@mflatt It would be nice if this could go into Racket 8.7! :-)

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/rx.rkt
+++ b/pkgs/racket-doc/scribblings/reference/rx.rkt
@@ -7,143 +7,143 @@
 ;; the table is unlikely to see this request, but it's worth a try.)
 
 (define grammar @string-append{
-  Regexp   ::= Pces               Match Pces                                      #co
-            |  Regexp|Regexp      Match either Regexp, try left first             #co 1
-  Pces     ::=                    Match empty                                     #co
-            |  PcePces            Match Pce followed by Pces                      #co
-  Pce      ::= Repeat             Match Repeat, longest possible                  #co 3
-            |  Repeat?            Match Repeat, shortest possible                 #co 6
-            |  Atom               Match Atom exactly once                         #co
-  Repeat   ::= Atom*              Match Atom 0 or more times                      #co 3
-            |  Atom+              Match Atom 1 or more times                      #co 4
-            |  Atom?              Match Atom 0 or 1 times                         #co 5
-  Repeat   ::= ...                ...                                             #px
-            |  Atom{N}            Match Atom exactly N times                      #px 7
-            |  Atom{N,}           Match Atom N or more times                      #px 8
-            |  Atom{,M}           Match Atom between 0 and M times                #px 9
-            |  Atom{N,M}          Match Atom between N and M times                #px 10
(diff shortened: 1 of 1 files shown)


sschwarzer:
I see the ["Cloisters" section](https://docs.racket-lang.org/guide/regexp-clusters.html#%28part._regexp-cloister%29) in the Racket Guide also needs an update. I'm on it.

sschwarzer:
Wait a minute. Dotall mode seems to be active by default.
```
> (regexp-match #px"a.*b" "a\nb")
'("a\nb")
```
(In line with this, the corresponding match in Python needs the DOTALL flag.)

I suggest we delay this PR for now until the actual behavior of Racket is clear. I'll look into it.

sorawee:
I don't think this is correct. In Racket, `s` does not indicate DOTALL. Instead, the multi-mode has a bit of behavior of both DOTALL and MULTILINE.

```
> (regexp-match #px"(?:^a$)" "b\na\nb")
#f
> (regexp-match #px"(?-s:^a$)" "b\na\nb")
'("a")
```

So it appears that `-s` is just a synonym of `m`, and `s` is a synonym for `-m`, as documented.

My experience with regexp is from PCRE (via Python), so this was surprising to me when I noticed it too. Luckily, the fact that you can toggle on and off the mode means that you can nest modes, so you can still express the regexp with no problem. E.g. an equivalent of 

```
import re
re.search("(?ms:(^.$))", "aaaaa\n\n\nbbbbbb")
```

is

```
(regexp-match #px"(?m:^(?-m:.)$)" "aaaaa\n\n\nbbbbbb")
```

sschwarzer:
@sorawee Thanks for chiming in, that's indeed helpful to know. I thought the previous documentation was from a misunderstanding of PCRE. Seems that _I_ misunderstood Racket's regexp's instead. ;-)

So I'll revoke this PR. It hasn't been in vain though, because I learned something. :-)

PR #4460, opened 2022-10-11
Make `release` a descendant of `stable`

LiberalArtist:
@jbclements This PR does what I suggested in <https://racket.discourse.group/t/racket-v8-6-release-thread/1091/32>:
> It also tries to address a much less significant quirk I noticed, which is that the head of the stable branch tends to point to a merge commit like https://github.com/racket/racket/commit/424d167500a9c5cb83e3f23217028954f47a69b7 rather than the commit pointed to by the release tag.
> 1.  When starting a new release branch, either just before or just after the commit like https://github.com/racket/racket/commit/6ef43da80cc2b84c05f7409646745d971aa8c379 setting the alpha version number, use `git merge` with the `ours` strategy so that the new head of the `release` branch is also a descendant of the tag from the previous release/the head of the `stable` branch.

I happened to notice that this seems to have been the practice at some point in not-too-distant Racket history: https://github.com/racket/racket/commit/20916f71375f7e162de67da925c7b17f104355df seems to do the same thing as this PR.

The way GitHub is displaying this is quite confusing: it's more useful to think of this PR as adding 2 commits (not 1,694), and it's correct that there are 0 files changed. Here's what I did:

1. `git merge -s ours stable` -> 247806d25341f46ee9a607f209abe2a5c3bfc022
2. `git merge -s ours v8.6` -> 6d8f9694d50188a49e78c8d4557440ae6db27923

In both cases I used the auto-generated commit message.

In the future only `git merge -s ours stable` should be necessary. In this case, because of the mix-up, the (current/changed) `v8.6` is actually not on the `stable` branch.

This should let you merge into `stable` with `--ff-only` when the release is ready, so the head of the `stable` branch can point to the actual commit tagged `v8.7`.


LiberalArtist:
The "Scribble License Files" failure looks like a bug in the CI script, not in this PR: probably it needs to use `--skip-installed` or something.

LiberalArtist:
Closed via https://github.com/racket/racket/commit/4468b8a316b937abd09381749824915b4e41317f and https://github.com/racket/racket/commit/d291ef3e56de18db19ab5c476d1557a64e3f36e9—thanks @jbclements!

PR #4461, opened 2022-10-12
openssl: add ssl-set-keylog-callback!

Bogdanp:
I needed to debug some TLS connections today using Wireshark and couldn't find an existing way to do it with the openssl lib. This commit exposes a way to hook into OpenSSL's built-in [keylogging callback](https://www.openssl.org/docs/man1.1.1/man3/SSL_CTX_set_keylog_callback.html) from Racket.

cc @rmculpepper

### DIFF
--- a/pkgs/racket-doc/openssl/openssl.scrbl
+++ b/pkgs/racket-doc/openssl/openssl.scrbl
@@ -738,7 +738,7 @@ The client sends this information via the TLS
 extension, which was created to allow @hyperlink["http://en.wikipedia.org/wiki/Virtual_hosting"]{virtual hosting}
 for secure servers.
 
-The suggested use it to prepare the appropriate server contexts,
+The suggested use is to prepare the appropriate server contexts,
 define a single callback which can dispatch between them, and then
 apply it to all the contexts before sealing them. A minimal example:
 
--- a/pkgs/racket-doc/openssl/openssl.scrbl
+++ b/pkgs/racket-doc/openssl/openssl.scrbl
@@ -738,7 +738,7 @@ The client sends this information via the TLS
 extension, which was created to allow @hyperlink["http://en.wikipedia.org/wiki/Virtual_hosting"]{virtual hosting}
 for secure servers.
 
-The suggested use it to prepare the appropriate server contexts, 
+The suggested use it to prepare the appropriate server contexts,
 define a single callback which can dispatch between them, and then
 apply it to all the contexts before sealing them. A minimal example:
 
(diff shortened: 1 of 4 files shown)


samth:
Would it be worth making it easy to log this info to a specified logger/topic/level?

Bogdanp:
> Would it be worth making it easy to log this info to a specified logger/topic/level?

I don't think there would be much benefit. Logging this info implicitly (for example, based on an env var) seems like it would needlessly increase the attack surface area and I'd worry about stuff like sandboxed code being able to listen for these logs (I don't really know to what extent logging is isolated within sandboxes). Having an explicit procedure to turn logging on for a specific context doesn't seem like an improvement over just setting the callback directly since with a logger you'd have to also set up a separate log receiver somewhere (and you could always just make the callback send the data to a logger).

Bogdanp:
Bump

mflatt:
Callbacks from C are tricky. The `#:in-original-place? #t` annotation affects callouts, not callbacks, so I don't think it's doing what you had in mind. More importantly, this is an unsafe interface, because a callback needs to be atomic, can't raise exceptions, etc. For example, adding a `(sleep)` to the callback function crashes Racket CS.

One way to make this interface safe could be to use logging, since `log-message` works in atomic mode. Instead of providing a callback function, a client might provide a logger (and then the client can receive logged messages to send them wherever). I don't know whether that would fit the uses that you have in mind.

Meanwhile, when registering a callback function, another big problem is making sure that the Racket-side close remains reachable for as long as the callback is installed. The program below does not crash for me, but it does crash if I add a `(collect-garbage)` on line 1417 of "mzssl.rkt". That's in `wrap-ports` soon after `create-ssl` returns. The intent of the example is that the `ctx` object can be collected after the connection is created and before the callback is invoked. Maybe the callback needs to be retained in an `mzssl` record, but I'm not sure when the callback can be invoked (and therefore how long it needs to be retained).

```
#lang racket/base
(require openssl)

(define-values (i o)
  (let ()
    (define ctx (ssl-make-client-context 'auto))
    (ssl-set-keylog-callback! ctx (lambda (line)
                                    (log-error "~s" line)))
    (define c (make-custodian))
    (parameterize ([current-custodian c])
      (ssl-connect "www.racket-lang.org" 443 ctx))))

'DONE
(collect-garbage)
(display "oops\n" o)
(read-line i)
```

Bogdanp:
Thanks Matthew. I had entirely the wrong mental model for `#:in-original-place?`, thinking it would somehow set things up such that the cb would be safe to call. I also hadn't realized the sslctx wasn't retained along with its wrapped ports.

I think logging to an explicit logger could work, assuming someone who doesn't have access to that exact logger can't somehow intercept the messages (or maybe that would also be fine, as long as we document it as a potential danger), so I'll rework this in that direction. Re. retaining the callback, I think it would make sense to have the callback automatically disconnect as soon as the context is garbage-collected, so in your example the logging would silently turn off by the time the `read-line` is reached. Might be somewhat surprising, but I think it's better than complicating the mzssl code since this is just a debugging feature. Are there any caveats I should be aware of with using will executors at this level?

mflatt:
Finalizing could make sense, but instead of a will executor, I think you'll want to use `register-finalizer`.

Bogdanp:
I think the revised commit addresses the problems with the original implementation. Thanks again for the help!

mflatt:
How does the dynamic extent of the callback relate to to the lifetime of an SSL connection created with a context? If it's only until `SSL_new` returns, then this implementation seems fine, and I don't think the finalizer would be necessary. But it appears that the underlying context object is retained in a connection, which makes me think that the callback could be invoked later. 

Even so, this looks like it would mostly† work, but I think it might be surprising for logging to go away when a context object becomes in accessible while a connection is still being used. Insteda of finalization, how about adding `ctx` to the `mzssl` record? It seems like that would work to keep the callback alive.

† Currently, it looks like setting a callback and then sealing the context will lead to an error when the GC tries to reclaim the context, since `ssl-set-keylogger!` wants an unsealed context. Another good reason to avoid finalization.

pkgs/racket-doc/openssl/openssl.scrbl
```diff
@@ -738,7 +738,7 @@ The client sends this information via the TLS
 extension, which was created to allow @hyperlink["http://en.wikipedia.org/wiki/Virtual_hosting"]{virtual hosting}
 for secure servers.
 
-The suggested use it to prepare the appropriate server contexts, 
+The suggested use it to prepare the appropriate server contexts,
```
otherjoel, line 741: (“it” → “is” ?)

PR #4463, opened 2022-10-13
Chez Scheme: configure: Ensure portability by using `=` to compare strings

cypheon:
The operator `==` is not supported in all shells an can silently fail (evaluates to false even when the strings are equal).


mflatt:
Thanks!

PR #4472, opened 2022-10-28
doc: typo

sorawee:
"to protected" still sound weird, so let me know if there's any better alternative.


gus-massa:
What about `@tech{protected}`? I'm not sure if `@tech` only modifies how it looks or if it has some side effect.

PR #4473, opened 2022-10-28
docs: Add example for `hash-ref`

jessealama:
Suggestion from the RacketCon hackathon.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -354,11 +354,11 @@ result:
 
 @examples[
 #:eval the-eval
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -352,6 +352,15 @@ result:
 
 ]
 
+@examples[
+#:eval the-eval
+(hash-ref (hash) "hi")
```
sorawee, line 357: I think you need `(eval:error ...)` wrapping around expressions that raise errors.

jessealama: Ack, good catch -- done!

PR #4474, opened 2022-10-28
docs: Add example of contract-out

ZachOBrien:
Suggestion from the Racketcon hackathon

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1957,7 +1957,6 @@ the export.
           (recip 3)
           (eval:error (recip 1+2i))]
 
-
 @history[#:changed "7.3.0.3" @list{Added @racket[#:unprotected-submodule].}
          #:changed "7.7.0.9" @list{Started ignoring @racket[ignored-id].}]
 }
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/contracts.scrbl
```diff
@@ -1944,6 +1944,22 @@ is bound to vectors of two elements, the exported identifier and a
...
+
+ (require racket/contract)
+
+ (provide
+  (contract-out
+   [recip (-> (and/c real? non-zero?)
+              (between/c -1 1))]))
+
+ (define (recip x) (/ 1 x))
+ (define (non-zero? x) (not (= x 0)))
```
jessealama, line 1960: Perhaps `(zero? x)`?

jessealama: Or perhaps, using `not/c`:

````
(and/c real? (not/c zero?))
```

And then you can drop `(define (non-zero? x) ...)`.

ZachOBrien: Thanks, good suggestion

sorawee:
I'd suggest using `@examples`. See `recontract-out` for an example.

ZachOBrien:
Thanks, I'll look into examples.

I took this contract from a nearby example in the documentation for consistency's sake, but it turns out it's incorrect... Given a real number less than 1, the reciprocal will be outside range [-1, 1]. So there is a bug in the existing documentation here:

https://github.com/racket/racket/blob/0b8732d9666044be6786dcbf5ca651c41eaed7d8/pkgs/racket-doc/scribblings/reference/contracts.scrbl#L1965-L1971

PR #4481, opened 2022-10-31
docs: warn about creating nested vectors with `make-vector`

jestarray:


pkgs/racket-doc/scribblings/reference/vectors.scrbl
```diff
@@ -35,7 +35,9 @@ Returns @racket[#t] if @racket[v] is a vector, @racket[#f] otherwise.}
                       [v any/c 0]) vector?]{
 
 Returns a mutable vector with @racket[size] slots, where all slots are
-initialized to contain @racket[v].
+initialized to contain @racket[v]. Note that if @racket[v] is a @racket[vector],
```
sorawee, line 38: It's not just `vector`. `struct` with mutable fields, `box`, etc will have the same problem. I think adjusting the doc to warn sharing makes sense, but should be specific to only `vector`.

jestarray: I did not know that! Unsure how I would word that though.

sorawee: How about this?

> Note that @racket[v] is shared for all elements, so for mutable data, mutating an element will affect other elements.

```
@examples[
  (make-vector 3 2)
  (define v (make-vector 5 (box 3)))
  v
  (set-box! (vector-ref v 0) 7)
  v
]
```

jestarray: sounds good to me! updated

PR #4491, opened 2022-11-15
Fixes a bug in bc's unicode normalization

97jaz:
Addresses a missing Hangul decomposition case in
do_string_normalize_c(), which is caught by a more recent version of the NormalizationTest.txt file.


mflatt:
Thanks!

PR #4499, opened 2022-11-24
Adjust format-id to automatically unwrap syntax object

sorawee:


jackfirth:
Context: we were talking in the discord and realized we wanted Resyntax to [suggest replacing](https://github.com/jackfirth/resyntax/issues/169) `(format-id #'foo "~a/c" (syntax-e #'bar))` with `(format-id #'foo "~a/c" #'bar)`, but that suggestion is technically unsound for cases like `(format-id #'foo "~a/c" (syntax-e #'42))`.

Issue #4504, opened 2022-11-29
Some racket/file procedures have incorrectly documented keyword arguments


benknoble:
**What version of Racket are you using?**
8.6 [cs], but this is present in the latest documentation: https://github.com/racket/racket/blob/master/pkgs/racket-doc/scribblings/reference/filesystem.scrbl#L1178-L1272

Several procedures use `defproc` with `[#:keyword var default-value]` which should be `[#:keyword var contract default-value]`; in particular, optional keywords that default to true or false (so probably the contract is `any/c`).

The documentation suggests that they are mandatory and that they have contract `#t` or `#f`.

---

I can submit a PR at the end of the day.

sorawee:
`string-path?` should also be `path-string?` in `get-preference`.

PR #4505, opened 2022-11-29
docs: fix some optional keyword definitions

benknoble:
Also correct `string-path?` (should be `path-string?`).

Close https://github.com/racket/racket/issues/4504


mflatt:
Thanks!

PR #4506, opened 2022-11-30
net/http-client: fix header field value LWS regexes. fixes #4503

### LINKED ISSUE #4503: net/http-client: http-conn-send! and http-conn-recv! use incorrect regexes to parse headers
[RFC 2616](https://www.rfc-editor.org/rfc/rfc2616#section-4.2) states:

> Each header field consists of a name followed by a colon (":") and the field value. Field names are case-insensitive. The field value MAY be preceded by any amount of LWS, though a single SP is preferred.

However, [`net/http-client`](https://github.com/racket/racket/blob/10a5aec3df13b35311a33168bf1e0ab8b9f26a05/racket/collects/net/http-client.rkt) insists on the space after the colon in several places for parsing the headers it needs:

```
$ rg ':\) \+|: \+'
net/http-client.rkt
163:  (unless (regexp-member #rx"^(?i:Host:) +.+$" headers-bs)
168:  (unless (regexp-member #rx"^(?i:User-Agent:) +.+$" headers-bs)
172:              (regexp-member #rx"^(?i:Accept-Encoding:) +.+$" headers-bs))
180:              (not (regexp-member #rx"^(?i:Content-Length:) +.+$" headers-bs)))
183:    (unless (regexp-member #rx"^(?i:Connection:) +.+$" headers-bs)
348:        (regexp-member #rx#"^(?i:Connection: +close)$" headers)))
355:      [(regexp-member #rx#"^(?i:Transfer-Encoding: +chunked)$" headers)
359:                (match (regexp-match #rx#"^(?i:Content-Length:) +(.+)$" h)
397:              (regexp-member #rx#"^(?i:Content-Encoding: +gzip)$" headers)
401:              (regexp-member #rx#"^(?i:Content-Encoding: +deflate)$" headers)
```

Proposed fix is simply to replace `<space>+` with `<space>*` in these regexes. I'll try to have a patch in shortly with accompanying tests.

For context, this was discovered while using `net/http-easy` to make calls against [Chromedriver](https://chromedriver.chromium.org/home), which does not use spaces in its response headers. The client would hang during `copy-port`, polling a socket for data after `Content-length` bytes had already been read.

mxork:


mflatt:
Thanks!

PR #4510, opened 2022-12-02
search: ranking results based on package status

### LINKED ISSUE #4083: [Feature Request] prioritize main distribution in search results
When there are multiple results, results from the main distribution should be prioritized first. Following shows undesirable orderings.

Screen Shot 2021-12-05 at 4 10 56 PM

Screen Shot 2021-12-05 at 4 11 31 PM

One possible fix would be to sort each sub-array in `matches` according to the package information before flattening them, at https://github.com/racket/racket/blob/master/pkgs/racket-index/scribblings/main/private/search.js#L578

A question, perhaps for @mflatt: is there a reliable way to fetch all package names in the main distribution? I could manually scrape the information, but then we need to keep this information in-sync with the actual main distribution somehow.

sorawee:
This is a first cut for search result ranking.
Within a group of search results of the same score, search results associated with the `base` package will come first, followed by those in `main-distribution`, and then everything else.

Search results also have an additional indicator on the right border, which can be hovered to read the description.

Unfortunately, it slows down `raco setup scribblings/main` by about 10 seconds. The command is invoked quite often, as it is run as a part of `raco pkg install` and `raco pkg remove`.
It is unclear if the performance hit is a deal breaker.

Fixes #4083 and issues raised in
https://racket.discourse.group/t/whats-official-racket-and-whats-not/1447

### DIFF
--- a/pkgs/racket-index/scribblings/main/private/make-search.rkt
+++ b/pkgs/racket-index/scribblings/main/private/make-search.rkt
@@ -22,7 +22,8 @@
          (for-syntax racket/base)
          (for-syntax racket/runtime-path)
          (for-syntax compiler/cm-accomplice)
-         "index-scope.rkt")
+         "index-scope.rkt"
+         "pkg.rkt")
 
 (provide make-search)
 
@@ -177,10 +178,12 @@
              [else
               "\"module\""])]
           [else "false"]))
+      (define pre-pkg-name (tag->pkg tag))
+      (define pkg-name (if pre-pkg-name (quote-string pre-pkg-name) "false"))
       (and href
            (string-append "[" (quote-string text) ","
                           (quote-string href) ","
-                          html "," from-libs "]"))))
+                          html "," from-libs "," pkg-name "]"))))
   (define l (filter values l-all))
 
   (define user (if user-dir? "user_" ""))
@@ -223,7 +226,13 @@
                                  (string-append (quote-string (car x)) ": "
                                                 (number->string (cdr x))))
                                ms)])
-                 (add-between ms ",\n  "))};
+                 (add-between ms ",\n  "))
+          };
+          @||
+          // an array of (transitive) dependencies of main-distribution
+          var plt_main_dist_pkgs = [
+            @,@(add-between (map quote-string (get-main-dist-pkgs)) ",\n  ")
+          ];
           @||})))
 
   (for ([src (append (list search-script search-context-page)
--- a/pkgs/racket-index/scribblings/main/private/make-search.rkt
+++ b/pkgs/racket-index/scribblings/main/private/make-search.rkt
@@ -40,9 +40,11 @@
 ;; ideally we could just inform scribble/raco that they need
 ;; installing, and they would just do that when appropriate.
 (begin-for-syntax
+  (define-runtime-path search-style "search.css")
   (define-runtime-path search-script "search.js")
   (define-runtime-path search-merge-script "search-merge.js")
   (define-runtime-path search-context-page "search-context.html")
+  (register-external-file search-style)
   (register-external-file search-script)
   (register-external-file search-merge-script)
   (register-external-file search-context-page))
--- a/pkgs/racket-index/scribblings/main/private/pkg.rkt
+++ b/pkgs/racket-index/scribblings/main/private/pkg.rkt
@@ -56,18 +56,28 @@
 ;; tag->pkg
 ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
+(define cache-for-module-path->path (make-hash))
+(define cache-for-->pkg (make-hash))
 (define pkg-cache-for-path->pkg (make-hash))
 
 (define (module-path->path module-path)
-  (with-handlers ([exn:missing-module? (lambda (exn) #f)])
-    (resolved-module-path-name
-     (module-path-index-resolve
-      (module-path-index-join module-path #f)))))
+  (hash-ref!
+   cache-for-module-path->path
+   module-path
+   (λ ()
+     (with-handlers ([exn:missing-module? (lambda (exn) #f)])
+       (resolved-module-path-name
+        (module-path-index-resolve
+         (module-path-index-join module-path #f)))))))
 
 (define (->pkg path)
-  (cond
-    [(path->pkg path #:cache pkg-cache-for-path->pkg)]
-    [else (and (path->main-collects-relative path) "base")]))
+  (hash-ref!
+   cache-for-->pkg
+   path
+   (λ ()
+     (cond
+       [(path->pkg path #:cache pkg-cache-for-path->pkg)]
+       [else (and (path->main-collects-relative path) "base")]))))
 
 (define (->path tag)
   (match tag
--- a/pkgs/racket-index/scribblings/main/private/pkg.rkt
+++ b/pkgs/racket-index/scribblings/main/private/pkg.rkt
@@ -76,7 +76,6 @@
          (list 'tech (list path-string _ ...)))
 
      (match path-string
-       ;; remove /xxx.scrbl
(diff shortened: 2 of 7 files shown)


sorawee:
Current search:

Screenshot 2022-12-02 at 11 33 32 PM

This PR:

Screenshot 2022-12-02 at 11 31 41 PM

benknoble:
If you haven't already, you might want to check the red/green being used for color-blind-friendly contrast.

sorawee:
I'm happy to adjust the colors. Do you have any recommendation for the color hex codes or color names?

benknoble:
A quick search turned up https://davidmathlogic.com/colorblind/#%23D81B60-%231E88E5-%23FFC107-%23004D40 which I liked

sorawee:
The blue-yellow ones look pretty decent across four (or five) columns. I pushed a change to use them. Thanks for the feedback!

sorawee:
Alright, adding even more caching fixes the performance issue. Now it runs in 2 seconds.

sorawee:
Here's one complication:

For bindings, the associated package is accurate.

For sections and tech terms, the associated package is the doc package. So we have a weird result where `syntax/parse` is in `base` (correct), but "Syntax: Meta-Programming Helpers" is in main-distribution, even though "Syntax: Meta-Programming Helpers" contains `syntax/parse`.

Should this be a concern? And if so, is there a solution?

mflatt:
I'm in favor of the goal here. I have some minor quibbles if we stick to this implementation:

 * the distinctions to make in sorting should be configurable, probably via "config.rktd", instead of being hardwired to "base", "main-distribution", and everything else;
 * red means "warning" or "error" in many contexts (although it is a perfectly good color in many contexts <insert University of Utah "U" here>), so I'd suggest something like gold and blue as the colors.

But I'm more concerned with the way the categories are currently derived. The package inference here works well, and I have no better idea given the documentation as-is. But it seems fragile and the long way around for information that should be declared. Given that the immediate goal is to treat main-distribution packages specially, it's within our reach to modify main-distribution documentation to declare the right information. Maybe documentation should start with an explicit declaration of the package that it's describing, for example, and have that propagated to index entries.

The issue here seems related to the problem of reporting bindings with syntactic categories as needed for Rhombus. Maybe it's worth looking at extending index information for these goals (and, of course, trying to imagine future needs) at the same time.

sorawee:
Gold and blue don't work well, because highlighted entries are colored light yellow, which doesn't contrast well with gold. What about blue and brown?

Indeed, this is related to a couple of my pending PRs on the "kind information". In the PRs, I created a new `exported-index-desc*` which extends `exported-index-desc`, and then deprecate other struct types that extends `exported-index-desc` (like `procedure-index-desc`, `form-index-desc`, etc.).

However, the change that you suggest here would be far more involved. I think the problem stems from how `struct` is not easily extensible (without creating new struct types, which is not a viable approach). It would be better if an index entry description is a hash table. But changing the representation means we deprecate `scribble/manual-struct` entirely. Is that OK with you?

mflatt:
Brown and blue sounds like it could work.

Yes, I'm ok with deprecating the old `index-desc` structure types and switching to one that's based on a hash table.

sorawee:
Got it. I am working on a different implementation, so I will submit further PRs to amend this PR.

shhyou:
@sorawee I really appreciate this feature!! It has been confusing to find identically-named bindings from other packages rather than the ones from main distribution.

While searching for `namespace`, I found this result. Do you know why `@deftech{namespace}` does not come before others?

(I don't know what is `qcr`. I saw it in the current search result so I installed it locally to test the ranking algorithm.)

With search ranking (HEAD):
Searching for namespace, with ranking

Without search ranking (stable):
Searching for namespace, without ranking

sorawee:
This is because "only bindings can be used for rexact matches"

https://github.com/racket/racket/blob/ba8803583c21796256f0aed7aee9c77c401051a6/pkgs/racket-index/scribblings/main/private/search.js#L464

which is introduced in https://github.com/racket/racket/commit/ecc43ff104ea02bd6fbba7a7a88fb54a6ac30731

I do not know the justification for this, but it seems to be working as intended.

shhyou:
> This is because "only bindings can be used for rexact matches" (where rexact is a special rank that comes before exact)
> 
> https://github.com/racket/racket/blob/ba8803583c21796256f0aed7aee9c77c401051a6/pkgs/racket-index/scribblings/main/private/search.js#L464
> 
> which is introduced in [ecc43ff](https://github.com/racket/racket/commit/ecc43ff104ea02bd6fbba7a7a88fb54a6ac30731)
> 
> I do not know the justification for this, but it seems to be working as intended.

@sorawee okay, thanks for the pointer! I'll play with it to see if it's possible to also include `deftech` for r-exact matches. In its current form, the sorting result of `namespace` is a bit unsatisfactory as those from `guide` and `reference` would be more preferable than the ones from other packages.

samth:
This PR should probably be closed as merged.

Unfortunately, I think it resulted in some regressions in package installation. See eg https://plt.cs.northwestern.edu/pkg-build/server/built/fail/sketching.txt which fails with:

```

standard-module-name-resolver: contract violation
  expected: module-path?
  given: #f
  context...:
   /home/root/racket/share/pkgs/racket-index/scribblings/main/private/pkg.rkt:68:5
   /home/root/racket/collects/racket/private/more-scheme.rkt:377:2: hash-ref!
   /home/root/racket/share/pkgs/racket-index/scribblings/main/private/pkg.rkt:109:0: tag->pkg
   /home/root/racket/share/pkgs/racket-index/scribblings/main/private/make-search.rkt:67:0: make-script
   /home/root/racket/share/pkgs/scribble-lib/scribble/base-render.rkt:974:4: render-content method in render%
   /home/root/racket/share/pkgs/scribble-lib/scribble/html-render.rkt:1561:4: render-plain-content method in render-mixin
   /home/root/racket/share/pkgs/scribble-lib/scribble/base-render.rkt:974:4: render-content method in render%
   /home/root/racket/share/pkgs/scribble-lib/scribble/html-render.rkt:1218:4: do-render-paragraph method in render-mixin
   /home/root/racket/share/pkgs/scribble-lib/scribble/html-render.rkt:1246:4: render-intrapara-block method in render-mixin
   /home/root/racket/share/pkgs/scribble-lib/scribble/base-render.rkt:921:20: loop
   /home/root/racket/share/pkgs/scribble-lib/scribble/base-render.rkt:920:4: render-compound-paragraph method in render%
   /home/root/racket/share/pkgs/scribble-lib/scribble/html-render.rkt:1802:4: render-compound-paragraph method in render-mixin
   /home/root/racket/share/pkgs/scribble-lib/scribble/html-render.rkt:1205:6: loop
   [repeats 2 more times]
   /home/root/racket/share/pkgs/scribble-lib/scribble/html-render.rkt:1118:4: render-part-content method in render-mixin
   /home/root/racket/share/pkgs/scribble-lib/scribble/html-render.rkt:821:4: render-one-part method in render-mixin
   ...
```

sorawee:
Will take a look, thanks! If it's difficult to fix, let's just revert it. Though if it's easy to fix, I'd prefer to make a quick fix. And in any case, let's close this PR.

samth:
For example, https://plt.cs.northwestern.edu/pkg-build/server/built/fail/dropbox.txt

sorawee:
It's a closely related problem, though not quite the same. My planned fix for Sketching does fix the problem with dropbox.

I still want to better understand how these tags are generated before submitting a PR to fix the problem.

In any case, it would also be nice to fix the [dropbox documentation](https://docs.racket-lang.org/dropbox/), which has many failed links, the underlying reason for the failure.

CC: @stchang

sorawee:
@shhyou: I agree with you though that this ranking could be unsatisfactory. The "hash code" tech term, for example, is the _last_ entry (on my local machine), even though it's the exact match.

So my plan, now that the PR is reverted, is to also revisit the ranking algorithm. Perhaps @elibarzilay could help us understand the motivation for "only bindings can be used for rexact matches".

PR #4519, opened 2022-12-12
Fix the documentation of `resolve-module-path`

### LINKED ISSUE #4498: `resolve-module-path` doesn't load the module

v8.7

**What program did you run?**

`resolve-module-path` doesn't load the module.

```racket
(require syntax/modresolve)
(resolve-module-path 'errortrace)
(module-declared? 'errortrace)
;=> #f
```

The [doc](https://docs.racket-lang.org/syntax/module-helpers.html#%28def._%28%28lib._syntax%2Fmodresolve..rkt%29._resolve-module-path%29%29) says:

> When `module-path-v` refers to a module using a collection-based path, resolution invokes the current [module name resolver](https://docs.racket-lang.org/reference/Module_Names_and_Loading.html#%28tech._module._name._resolver%29) and loads the module if it is not declared.

In other words `(module-declared? 'errortrace)` should be `#t`. However, the implementation of `resolve-module-path` doesn't ask `module-path-index-resolve` to load the module.

https://github.com/racket/racket/blob/848d2a5db2f7523ea6d9d7c4d3a3a18485dd59a7/racket/collects/syntax/private/modresolve-noctc.rkt#L60-L72

---

Should I fix the code or should I fix the doc?

shhyou:
The function `resolve-module-path` does not load new modules when resolving the given module path. This commit fixes the documentation.

Closes #4498.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/module-reflect.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/module-reflect.scrbl
@@ -665,7 +665,7 @@ more than the namespace's @tech{base phase}.}
          boolean?]{
 
 Returns @racket[#t] if the module indicated by @racket[mod] is
-declared (but not necessarily @tech{instantiate}d or @tech{visit}ed)
+@tech{declare}d (but not necessarily @tech{instantiate}d or @tech{visit}ed)
 in the current namespace, @racket[#f] otherwise.
 
 If @racket[load?] is @racket[#t] and @racket[mod] is not a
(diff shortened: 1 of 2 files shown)


shhyou:
I documented the doc change in the history note. Let me know if there's a better place for this.

mflatt:
I don't think it's necessary to include a `history` note for documentation corrections. I view the `history` notes as mostly being useful for "my program needs to work on Racket version X and later, can I use this?" and not a summary of all changes.

mflatt:
Thanks for the adjustment!

I should have thought of this before, but probably the documentation should be explicit that the module name resolver is called *without* loading the module if it is not already declared. Maybe the change is `and loads` -> `, but without loading`.

shhyou:
Done. Thanks for the suggestion!

mflatt:
Thanks!

Issue #4456, opened 2022-10-09
[Feature Request] How to build Racket for ALT Linux


Ancieg:
**Is your feature request related to a problem? Please describe.**
We build packages in ALT Linux via `hasher`, the program which creates isolated environment during building (chroot). That building environment are strictly disconnected from the Internet (for some reasons, like a security, more clean result and so on).
We use rpm as a package management tool.

I need to describe our build process briefly:
1. Hasher creates isolated environment and copy source files (tar archive) into `/usr/src/RPM/SOURCES` in chroot (spec file is copied to `/usr/src/RPM/SPECS`).
2. Hasher installs all needed packages to build (they are taken from BuildRequires). The following steps are all in the chroot.
3. RPM starts building using spec from `/usr/src/RPM/SPECS`.
3.1. **(%prep)** `/usr/src/RPM/SOURCES/<name>-<version>.tar` is extracted to `/usr/src/RPM/BUILD/<name>-<version>/`
3.2. **(%build)** The program is built in-place (or into subdirectory).
3.3. **(%install)** Move building results into `%buildroot` (in our case it is `/usr/src/tmp/<name>-buildroot/`). In `%buildroot` we create paths like they are related to `/`: for example, `/usr/bin/<program-name>`. But full file path is
`/usr/src/tmp/<program-name>-buildroot/usr/bin/<program-name>`.
3.4. **(%files)** We specify files to package relative to `%buildroot`, `/usr/bin/<name>` by example (it will be moved to RPM-package from `%buildroot/usr/bin/<name>`).
4. Creates RPM package with files taken from **%files** section.
5. We install the package into the system (without `%buildroot` prefix in files paths).

The problem is that Racket has hard-coded paths inside binaries and libraries, that is `/usr/src/tmp/<program>-buildroot/` prefixes all paths, but it doesn't exist in user's system (in non-chroot).

**Describe the solution you'd like**
I want to know how to build and install Racket via way described above so that the paths do not contain %buildroot as a prefix.

As an alternative, I would like to see generation more flexible Makefiles for the Racket source code repository.

**Do you want to contribute to this feature**
Yes, I want. But I can't because I novice to Racket and learn "How to design programs" book. But I can't deep into Racket learning, because we don't have Racket in our repository (Sisyphus) for now (previous maintainer can't maintain for now, so I want to build the package for our repository).

---

(I apologize for my poor English language skills)

LiberalArtist:
[...]

PR #4521, opened 2022-12-12
Update config.guess and config.sub from upstream

### LINKED ISSUE #4520: Invalid configuration `riscv64-suse-linux-gnu': machine `riscv64-suse' not recognized
https://build.opensuse.org/package/live_build_log/openSUSE:Factory:RISCV/racket/standard/riscv64

./configure --host=riscv64-suse-linux-gnu --build=riscv64-suse-linux-gnu --program-prefix= --disable-dependency-tracking --prefix=/usr --exec-prefix=/usr --bindir=/usr/bin --sbindir=/usr/sbin --sysconfdir=/etc --datadir=/usr/share --includedir=/usr/include --libdir=/usr/lib64 --libexecdir=/usr/libexec --localstatedir=/var --sharedstatedir=/var/lib --mandir=/usr/share/man --infodir=/usr/share/info --prefix=/usr --exec-prefix=/usr --libdir=/usr/lib64 --docdir=/usr/share/doc/packages/racket --enable-shared --disable-static --disable-strip --enable-places --enable-lt=/usr/bin/libtool --enable-libz --enable-liblz4 --enable-pthread
=== Racket CS enabled
checking build system type... Invalid configuration `riscv64-suse-linux-gnu': machine `riscv64-suse' not recognized
configure: error: /usr/bin/bash ../../lt/config.sub riscv64-suse-linux-gnu failed

andreas-schwab:
Fixes #4520

### DIFF
--- a/racket/src/lt/config.guess
+++ b/racket/src/lt/config.guess
@@ -1,12 +1,14 @@
 #! /bin/sh
 # Attempt to guess a canonical system name.
-#   Copyright 1992-2017 Free Software Foundation, Inc.
+#   Copyright 1992-2022 Free Software Foundation, Inc.
 
-timestamp='2020-12-03'
+# shellcheck disable=SC2006,SC2268 # see below for rationale
+
+timestamp='2022-09-17'
 
 # This file is free software; you can redistribute it and/or modify it
 # under the terms of the GNU General Public License as published by
-# the Free Software Foundation; either version 3 of the License, or
+# the Free Software Foundation, either version 3 of the License, or
 # (at your option) any later version.
 #
 # This program is distributed in the hope that it will be useful, but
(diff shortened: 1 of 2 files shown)


LiberalArtist:
Alt Linux (another RPM-based distro) also seems to have started adding the distro name between the architecture and `-linux-gnu` in the platform tuple: https://github.com/racket/racket/issues/4456#issuecomment-1280739982

If more distro are going to be doing that (IDK why), it does seem like we should update our `config.guess` and `config.sub` to versions that support them.

I don't know any reason why this change wouldn't be fine, but it seems like it might be better to run `autoupdate`/`autoreconf` and update all the generated files. Debugging Autotools issues is tricky enough without mixing and matching versions. But if we were to do that, we sound update to an Autoconf version that's convenient for @mflatt.

mflatt:
Updating this way seems to be as good as any approach, so I'll merge. Thanks!

PR #4522, opened 2022-12-14
match: compile to case on simple cases

sorawee:
From the code:

  ;; Attempt to use `case` when it is possible.
  ;;
  ;; (match id [case-pattern rhs] ... [other-pattern rhs*] ...)
  ;;
  ;; is compiled to:
  ;;
  ;; (let ([next (λ () (match id [other-pattern rhs*] ...))])
  ;;   (case id
  ;;     [(case-pattern)
  ;;      (syntax-parameterize ([fail (make-rename-transformer #'next)])
  ;;        rhs)] ...
  ;;     [else (next)]))
  ;;
  ;; The syntax-parameterization is done to support (failure-cont).
  ;; In match, failure-cont would yield to the next clause.
  ;; In our scheme, however, all case-pattern clauses will be disjoint,
  ;; so we can skip to the else clause (next) right away.
  ;;
  ;; Disjointness across clauses is therefore very important:
  ;;
  ;; (match 1 [1 (failure-cont)] [1 (displayln 'hello)])
  ;;
  ;; should print "hello". A compilation that doesn't take disjointness into
  ;; account, such as:
  ;;
  ;; (let ([next (λ () (match id))])
  ;;   (case id
  ;;     [(1)
  ;;      (syntax-parameterize ([fail (make-rename-transformer #'next)])
  ;;        (fail))]
  ;;     [(1)
  ;;      (syntax-parameterize ([fail (make-rename-transformer #'next)])
  ;;        (displayln 'hello))]
  ;;     [else (next)]))
  ;;
  ;; would be incorrect.
  ;;
  ;; (=> exit-id) is handled regularly, but with the target being the else clause
  ;; due to the same reason as above.

Co-authored by sam(th) and sam(ph)

### DIFF
--- a/racket/collects/racket/match/compiler.rkt
+++ b/racket/collects/racket/match/compiler.rkt
@@ -58,16 +58,17 @@
                            (predicate-stx x)
                            (quasisyntax/loc predicate-stx
                              (#,predicate-stx #,x)))]
-                  [rhs (compile* (cons x xs)
-                                 (map (lambda (row)
-                                        (define-values (p ps)
-                                          (Row-split-pats row))
-                                        (make-Row (cons (make-Dummy #f) ps)
-                                                  (Row-rhs row)
-                                                  (Row-unmatch row)
-                                                  (Row-vars-seen row)))
-                                      rows)
-                                 esc)])
+                  [rhs (compile**
+                        (cons x xs)
+                        (map (lambda (row)
+                               (define-values (p ps)
+                                 (Row-split-pats row))
+                               (make-Row (cons (make-Dummy #f) ps)
+                                         (Row-rhs row)
+                                         (Row-unmatch row)
+                                         (Row-vars-seen row)))
+                             rows)
+                        esc)])
       #'[lhs rhs]))
   (define (compile-con-pat accs pred pat-acc)
     ;; eliminate accessors for columns where every pat is a Dummy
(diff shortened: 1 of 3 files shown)


racket/collects/racket/match/compiler.rkt
```diff
@@ -551,5 +562,167 @@
...
+               ;;
+               ;; In this case,
+               ;; - 1 can be case-dispatched.
+               ;; - (or 1 2) can't because of duplicate 1, so it must be matched
+               ;; - (or 2 3) also can't because we just moved (or 1 2)
+               ;;   to the else clause. Since matching on 2 should reach
+               ;;   (or 1 2), not (or 2 3), (or 2 3) must be moved to the else
+               ;;   clause too.
+
+               (for ([literal (in-list literal-list)])
```
samth, line 705: Can we combine these two for loops into one?

sorawee: Not straightforwardly. Consider:

```
(match ...
  [(or 1 2 2) ...]
  [3 ...])
```

If we query and mark in the same loop, the the optimization would be inhibited because of duplicate `2`: when we see the second `2`, `2` is already marked.

Separating into two loops fixes the problem. The querying loop would say the `(or 1 2 2)` is OK. And only then, the marking loop would mark `1` and `2`.

So it adds more optimization opportunities.

samth: Makes sense. Can you make sure there's a test that ensures this case is optimized?

sorawee: It's already there. See test-case "Duplicate in a clause".

PR #4532, opened 2022-12-27
match: fix hash-table pattern and its doc, add more optimizations

sorawee:
https://github.com/racket/racket/commit/8377fa842fb3b29650b0fd19b6ef13a555b843e4 introduces an optimization for hash-table patterns whose keys are all literals. Unfortunately, it incorrectly allows partial matching, such as:

(match (hash 1 2 3 4)
  [(hash-table (1 x)) 'matched])

This issue was noticed and there's an attempt to fix it in https://github.com/racket/racket/commit/393582492203608a579a062b3fcece0ad4e18e88 Yet, it only fixes a very specific case, and the problem in general remains.

This PR fixes the issue properly, by requiring matching against the full hash table. However, this buggy behavior in fact is useful in many situations, so the PR also adds more optimization opportunity for this case:

(match (hash 1 2 3 4)
  [(hash-table (1 x) _ ...) 'matched])

Lastly, the PR documents the id pattern for matching against a key-value pair with the list representation. The feature was intentionally supported but undocumented since 2008
(https://github.com/racket/racket/commit/ae4acf1d517beeb7673edb12cff77add32ebf89c).

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match-grammar.rkt
+++ b/pkgs/racket-doc/scribblings/reference/match-grammar.rkt
@@ -15,8 +15,8 @@ pat     ::= id                                @match anything, bind identifier
          |  (LIST-NO-ORDER pat ...)           @match pats in any order
          |  (LIST-NO-ORDER pat ... lvp)       @match pats in any order
          |  (VECTOR lvp ...)                  @match vector of pats
-         |  (HASH-TABLE (pat pat) ...)        @match hash table
-         |  (HASH-TABLE (pat pat) ...+ ooo)   @match hash table
+         |  (HASH-TABLE kvp ...)              @match hash table
+         |  (HASH-TABLE kvp ...+ ooo)         @match hash table
          |  (CONS pat pat)                    @match pair of pats
          |  (MCONS pat pat)                   @match mutable pair of pats
          |  (BOX pat)                         @match boxed pat
@@ -42,6 +42,9 @@ literal ::= #t                                @match true
          |  keyword                           @match equal% keyword
          |  regexp literal                    @match equal% regexp literal
          |  pregexp literal                   @match equal% pregexp literal
+kvp     ::= _                                 @match any hash table key-value pair
+         |  id                                @match hash table key-value pair (as a list)
+         |  (pat pat)                         @match hash table key-value pair
 lvp     ::= (code:line pat ooo)               @greedily match pat instances
          |  pat                               @match pat
 qp      ::= literal                           @match literal
--- a/pkgs/racket-doc/scribblings/reference/match-parse.rkt
+++ b/pkgs/racket-doc/scribblings/reference/match-parse.rkt
@@ -29,6 +29,8 @@
     => (fixup s (fixup-sexp 'qp))]
    [(regexp-match-positions #rx"lvp" s)
     => (fixup s (fixup-sexp 'lvp))]
+   [(regexp-match-positions #rx"kvp" s)
+    => (fixup s (fixup-sexp 'kvp))]
    [(regexp-match-positions #rx"struct-id" s)
     => (fixup s (fixup-sexp 'struct-id))]
    [(regexp-match-positions #rx"pred-expr" s)
@@ -64,7 +66,7 @@
            (cdr (map fixup-sexp (vector->list (struct->vector s)))))]
    [(symbol? s)
     (case s
-      [(lvp pat qp literal ooo datum struct-id
+      [(lvp kvp pat qp literal ooo datum struct-id
             string bytes number character expr id
             rx-expr px-expr pred-expr
             derived-pattern)
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -223,7 +223,7 @@ In more detail, patterns match as follows:
          [(vector 1 (list a) ..3 5) a])
        ]}
 
- @item{@racket[(#,(match-kw "hash-table") (_pat _pat) ...)] ---
+ @item{@racket[(#,(match-kw "hash-table") _kvp ...)] ---
        similar to @racketidfont{list-no-order}, but matching against
        hash table's key--value pairs.
 
@@ -231,14 +231,21 @@ In more detail, patterns match as follows:
        #:eval match-eval
        (match #hash(("a" . 1) ("b" . 2))
          [(hash-table ("b" b) ("a" a)) (list b a)])
+       (match #hash(("a" . 1) ("b" . 2) ("c" . 3))
+         [(hash-table ("c" _) ("b" _) x) x])
+       (match #hash(("a" . 1) ("b" . 2))
+         [(hash-table ("b" b)) "not matched"]
+         [(hash-table ("b" b) _) "matched"])
        ]}
 
- @item{@racket[(#,(racketidfont "hash-table") (_pat _pat) ...+ _ooo)]
+ @item{@racket[(#,(racketidfont "hash-table") _kvp ...+ _ooo)]
        --- Generalizes @racketidfont{hash-table} to support a final
        repeating pattern.
 
        @examples[
        #:eval match-eval
+       (match #hash(("a" . 1) ("b" . 2))
+         [(hash-table ("b" b) _ ...) b])
        (match #hash(("a" . 1) ("b" . 2))
          [(hash-table (key val) ...) key])
        ]}
--- a/pkgs/racket-test/tests/match/match-tests.rkt
+++ b/pkgs/racket-test/tests/match/match-tests.rkt
@@ -51,7 +51,173 @@
                        [_ "non-empty"])
                      "empty"))
      ))
-  
+
+  (define hash-table-no-rep-tests
+    (test-suite "hash-table patterns (no ..k)"
+      (test-case "literal keys"
+        (check-equal? (match (hash 1 2 3 4)
+                        [(hash-table (1 a) (3 b))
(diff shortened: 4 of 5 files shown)


shhyou:
This bug fix breaks existing programs like [racket-langserver](https://github.com/jeapostrophe/racket-langserver) that depends on partial matches. While the fix is correct, I'm worry that there may be other programs that assumes the old, buggy semantics.

shhyou:
(Only slightly related: the match expander `struct*` also expects partial matches so the buggy behavior it's not entirely unreasonable.)

sorawee:
I'll let @samth decide what to do with this. But I agree that I should have added `@history`.

In any case, I will make a case here that the semantics of `match` prior this PR is _totally_ unreasonable.

1) Prior this PR, the following programs return `'does-not-match`

```
(match (hash 1 'a 2 'b 3 'c)
  [(hash-table ((== 1) x) ((== 2) y)) 'matches]
  [_ 'does-not-match])

(match (hash 1 'a 2 'b 3 'c)
  [(hash-table ((? (lambda (z) (= z 1))) x) ((? (lambda (z) (= z 2))) y)) 'matches]
  [_ 'does-not-match])
```

while the following program returns `'matches`.

```
(match (hash 1 'a 2 'b 3 'c)
  [(hash-table (1 x) (2 y)) 'matches]
  [_ 'does-not-match])
```

even though `(hash-table (1 x) (2 y))` have the same shape as `(hash-table ((== 1) x) ((== 2) y))` and `(hash-table ((? (lambda (z) (= z 1))) x) ((? (lambda (z) (= z 2))) y))`, so the same reasoning that makes `(hash-table ((== 1) x) ((== 2) y))` and friends not matched should be applied to `(hash-table (1 x) (2 y))` too.

2) Prior this PR, the following program returns `'does-not-match`

```
(match (hash 1 'a 2 'b 3 'c)
  [(hash-table) 'matches]
  [_ 'does-not-match])
```

while the following programs return `'matches`.

```
(match (hash 1 'a 2 'b 3 'c)
  [(hash-table (1 x)) 'matches]
  [_ 'does-not-match])

(match (hash 1 'a 2 'b 3 'c)
  [(hash-table (1 x) (2 y)) 'matches]
  [_ 'does-not-match])

(match (hash 1 'a 2 'b 3 'c)
  [(hash-table (1 x) (2 y) (3 z)) 'matches]
  [_ 'does-not-match])
```

even though `(hash-table)` is a continuation of the `(hash-table (1 x) (2 y) (3 z))`, `(hash-table (1 x) (2 y))`, `(hash-table (1 x))` pattern downward, so the same reasoning that makes `(hash-table (1 x) (2 y) (3 z))` and friends matched should be applied to `(hash-table)` too.

-----------------------

We can consider `(hash-table (lit-pattern val-pattern) ....)` to be a special case, but doing so means we need to admit not just one, but two independent inconsistencies. The second inconsistency in particular occurred because the commit 39358249220360 recognizes the first inconsistency, but did not properly fix the problem. Instead, the commit causes further inconsistencies.

sorawee:
Dang it, just discovered a bug in this PR... :( 

My bottom line is still the same, which is that the behavior prior this PR doesn't make sense, and should be eventually fixed. But I'd be OK with reverting this PR for the 8.8 release. 

Really sorry for the mess.

----------------------------

One more example to show why the behavior prior this PR AND the behavior of this PR is wrong.

```
(match (hash 1 'a 2 'b 3 'c)
  [(hash-table (1 x) (2 y) (2 z) _ ...) 'matches]
  [_ 'does-not-match])

(match (hash 1 'a 2 'b 3 'c)
  [(hash-table (1 x) (2 y) (2 z)) 'matches]
  [_ 'does-not-match])

(match (hash 1 'a 2 'b 3 'c)
  [(hash-table ((== 1) x) (2 y) (2 z) _ ...) 'matches]
  [_ 'does-not-match])

(match (hash 1 'a 2 'b 3 'c)
  [(hash-table ((== 1)  x) (2 y) (2 z)) 'matches]
  [_ 'does-not-match])
```

produces:

```
'does-not-match
'matches
'does-not-match
'does-not-match
```

prior this PR, and:

```
'matches
'does-not-match
'does-not-match
'does-not-match
```

in this PR. They should all be `'does-not-match`.

mfelleisen:
@sorawee "1." does make sense if you accept that the occurrence of literal constants in a program is different from computed equality with the same constants. There is precedent in the literature/history of PL. -- But yes, the spirit of Racket suggests that this should not happen.

samth:
I definitely agree that the prior behavior was wrong, and we can't keep it. But given that the current state is wrong, we should revert this PR for 8.8.

mfelleisen:
Yes. -- My comment was intended to alert you to the "big" picture of PL since pretty soon you might be teaching a course on it.

rfindler:
> I don't know what's the better option here. Could it be using a different syntax for matching hash tables and document hash-table as deprecated?

That sounds like a nice way forward to me. The current syntax isn't the most beautiful option!

gus-massa:
Somewhat related https://github.com/racket/rhombus-prototype/pull/241 "_Add Map.empty a binding as well as an expression_"

> An empty map. Differs from `{}` and `Map()` because while those as binding forms match any maps including ones with extra keys, `Map.empty` as a binding only matches empty maps, with no keys.

sorawee:
@shhyou I think that's a brilliant idea. 

If we make a new pattern, we have freedom to pick new semantics too. What should it be? And what should be the name?

Re name: `hash` is the straightforward choice, but it's unfortunate that the `(hash ...)` constructor creates immutable hash table, whereas the pattern should be able destruct mutable hash table... I suspect that's the reason why the current hash table pattern was named `hash-table`.

Re semantics: I'm thinking about `(pattern-name [k-expr v-pat] ... tail)` where `k-expr` is an arbitrary expression (and NOT a pattern). All `k-expr`s should be distinct. `tail` follows the grammar:

```
tail ::= 
       | #:tail tail-pat
```

- When `tail` is not given, the whole pattern fully matches against the value.
- When `tail` is given, `tail-pat` matches against `(hash-remove* h k-expr1 k-expr2 ...)`.

This is less expressive than the current `hash-table`, but I think covers common use cases of matching against a hash table.

What do you think?

CC: @shhyou, @samth, @AlexKnauth, @jackfirth

samth:
The name `hash-table` dates back to Bruce Haumann's implementation in ~2002; at that time all the functions used the prefix `hash-table` (see https://docs.racket-lang.org/mzscheme/Old_Functions.html#%28def._%28%28lib._mzscheme%2Fmain..rkt%29._make-hash-table%29%29; also the types in Typed Racket). 

I think we should use the name `hash`, and the syntax should follow the grammar of the `hash` function (ie, without extra `[]`). I think we should start minimally without `...` and tail support, and then build that in based on looking at what people actually do.

jackfirth:
> The name `hash-table` dates back to Bruce Haumann's implementation in ~2002; at that time all the functions used the prefix `hash-table` (see https://docs.racket-lang.org/mzscheme/Old_Functions.html#%28def._%28%28lib._mzscheme%2Fmain..rkt%29._make-hash-table%29%29; also the types in Typed Racket).
> 
> I think we should use the name `hash`, and the syntax should follow the grammar of the `hash` function (ie, without extra `[]`). I think we should start minimally without `...` and tail support, and then build that in based on looking at what people actually do.

I'm in favor of this. But I have concerns about the behavior with respect to extra keys. People are going to do this:

```
(match (get-json-from-some-web-api ...)
  [(hash 'user user-id 'posts post-list)
   ...])
```
If the pattern fails to match hash tables that have more keys than just `'user` and `'posts`, then this code will break as soon as whatever web API is being called decides to add some extra data to their response. Not only that, it won't even be obvious that that's the cause of the failure - the error message will simply say "no matching clause" and then include a (possibly massive!) hash table in the error message. Given the prevalence of JSON APIs in today's ecosystem, and the accepted practice of using "must ignore" semantics for unrecognized fields in those APIs, this is a huge footgun. I think the `(hash ...)` pattern should accept hash tables with extra keys to avoid this problem.

sorawee:
I strongly advocate for some forms of key-value grouping, like bracket. It make it much easier to read and understand what's going on.

My proposal includes `#:tail` precisely to allow users to choose full vs partial match behavior. I am not in favor of using `...` for the hash table, since that would require a hash to list conversion. Furthermore, "the rest" would be more useful in the hash table form rather than the list form.

AlexKnauth:
It should still be possible to specify "full" match behavior, as in no-extras, with something like Rhombus's `Map.empty` as the tail pattern, leaving the default as "partial" match behavior, as in extras allowed.

jackfirth:
> It should still be possible to specify "full" match behavior, as in no-extras, with something like Rhombus's `Map.empty` as the tail pattern, leaving the default as "partial" match behavior, as in extras allowed.

Why? I'm not sure what the use case is.

jackfirth:
> I strongly advocate for some forms of key-value grouping, like bracket. It make it much easier to read and understand what's going on.
> 
> My proposal includes `#:tail` precisely to allow users to choose full vs partial match behavior. I am not in favor of using `...` for the hash table, since that would require a hash to list conversion. Furthermore, "the rest" would be more useful in the hash table form rather than the list form.

Agreed on exposing the rest as a hash rather than a list. Personally I'd be fine with using ellipses _and_ making the ellipses-bound variable contain a hash instead of a list, but that might be too wacky to some.

jbclements:
From the Release Regressions issue, it appears that the decision is to revert this on both master and release. I'm just following @sorawee 's orders here :).

samth:
Yes that's correct.

PR #4536, opened 2023-01-06
file/zip: fix inconsistent general purpose bits

cloudrac3r:
Racket previously wrote general purpose bits to the local file header only. With this commit, it correctly writes the bits to the central directory entry too.

Previously, some zip tools like Linux '7z x' and 'file-roller' would state a headers error when attempting to open an archive with this problem.

This issue only appears when writing to an output port that is not seekable, since that causes Racket (and other applications) to write placeholders for CRC and size at first, then to write an "EXT signature" ("data descriptor" in the zip spec) with the real values. Only in this situation are the general purpose bits not zero. Output ports that are seekable would not exhibit this problem.

Test case:
```
racket -e '(require file/zip) (zip->output \'("/some/file")) | cat > test-not-seekable.zip
7z t test-not-seekable.zip # output will say ok or error
```

### DIFF
--- a/racket/collects/file/zip.rkt
+++ b/racket/collects/file/zip.rkt
@@ -16,8 +16,8 @@
   (define-struct metadata
     (path name directory? time date compression attributes))
 
-  ;; header : metadata * exact-integer * nat * nat * nat
-  (define-struct header (metadata crc compressed uncompressed size))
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #4541, opened 2023-01-08
Improvements to the stencil vector documentation

MartianMoorGrove:


MartianMoorGrove:
I pushed my real name to be the commits' author name. I don't think anyone wants to see "Martian MoorGrove" in their git log.

mflatt:
Thanks!

PR #4543, opened 2023-01-15
Refactor some things in racket/unit

lexi-lambda:
This PR refactors some things in `racket/unit`. There should be **no behavioral changes** aside from some minor alterations to error messages, and the new error messages should be better and more consistent (since they are now generated by `syntax/parse`).

The bulk of the changes are new implementations of `define-signature` and `compound-unit`, which I’ve rewritten to use `syntax/parse`. There are also a couple minor tweaks: I’ve eliminated the `error-syntax` parameter from `racket/private/unit-syntax` in favor of using `current-syntax-context` from `racket/syntax`, and I’ve changed the type of the internal `unit-deps` field to be more consistent with the other unit fields.

One might wonder why I would go to the trouble of making these changes out of the blue, and the answer is that I wanted to make a minor improvement to `define-signature` but found the implementation largely inscrutable. So you can consider this PR mostly a side-effect of me trying to understand the way everything works, and I hope to submit a followup PR with my behavioral changes.

### DIFF
--- a/pkgs/racket-test/tests/units/test-unit-contracts.rkt
+++ b/pkgs/racket-test/tests/units/test-unit-contracts.rkt
@@ -87,10 +87,10 @@
  "misuse of define-signature keyword"
  contracted)
 (test-syntax-error
- "expected a list of [id contract]"
+ "expected [id contract] pair"
   (define-signature x ((contracted x y))))
 (test-syntax-error
- "expected a list of [id contract]"
+ "expected identifier"
   (define-signature x ((contracted [(-> number? number?) x]))))
 
 (test-syntax-error
--- a/pkgs/racket-test/tests/units/test-unit.rkt
+++ b/pkgs/racket-test/tests/units/test-unit.rkt
@@ -331,31 +331,31 @@
 
 ;; compound-unit syntax errors (without sub-signatures)
 (test-syntax-error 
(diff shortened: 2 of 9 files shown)


mflatt:
Thanks!

As I minor note, I checked the memory use and load time of `racket/unit` (starting from `racket/base`), and that led me to remove the `syntax/id-set` import (which is expensive and was unused, unless I'm missing something).

lexi-lambda:
Thanks, good catch—I was using the import in an earlier version of this patch, but it is indeed unused now.

rfindler:
Could this change be related to the failures in [this drdr run](http://drdr.racket-lang.org/62661/), which seem to happen consistently or [this drdr run](http://drdr.racket-lang.org/62662/), which happened only once?

samth:
I think the consistent ones are definitely caused by this change. The transient failure I don't think is related, unless it's indicating a Racket bug that got hit that one time.

lexi-lambda:
It looks like the failures are in Typed Racket, which makes sense—I’d forgotten that TR supports units, and it depends on precise details of their expansion. It would perhaps be nice if TR could be made to depend somewhat less on those details, but it’s probably not worth the effort. I can take a look at what TR is expecting and see which place makes the most sense to change.

lexi-lambda:
I just sat down to look into this, only to fine that Matthew already fixed it for me in https://github.com/racket/typed-racket/commit/526b9c7ec85b90206e1a30c34903d69d16866606. Thanks Matthew!

PR #4545, opened 2023-01-16
Document exception raised for malformed inputs

jessealama:
Also, add example of repaired JSON (the example before this one is bad, but the new example shows that it can be repaired).

### DIFF
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -123,9 +123,10 @@ the @rfc for more information about JSON.
   remains. Like @racket[read], the function leaves all remaining
   characters in the port so that a second call can retrieve the
   remaining JSON input(s). If the JSON inputs aren't delimited per se
-  (true, false, null), they  must be separated by whitespace from the
-  following JSON input. 
-  
+  (true, false, null), they must be separated by whitespace from the
+  following JSON input. Raises @racket[exn:fail:read] if @racket[in] is not
+  at EOF and starts with malformed JSON (that is, no initial sequence of bytes
+  in @racket[in] can be parsed as JSON); see below for examples.
 
 @examples[#:eval ev
   (with-input-from-string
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/json/json.scrbl
```diff
@@ -123,9 +123,10 @@ the @rfc for more information about JSON.
   remains. Like @racket[read], the function leaves all remaining
   characters in the port so that a second call can retrieve the
   remaining JSON input(s). If the JSON inputs aren't delimited per se
-  (true, false, null), they  must be separated by whitespace from the
-  following JSON input. 
-  
+  (true, false, null), they must be separated by whitespace from the
+  following JSON input. Raises @racket[exn:fail:read?] if @racket[in] is not
```
mflatt, line 127: The documentation normally refers to exceptions by the name without `?`.

pkgs/racket-doc/json/json.scrbl
```diff
@@ -157,10 +158,20 @@ the @rfc for more information about JSON.
       "sandwich sandwich" (code:comment "invalid JSON")
       (λ () (read-json))))
 
+  (eval:error
+    (with-input-from-string
+      "false sandwich" (code:comment "valid JSON prefix, invalid remainder is not (immediately) problematic")
+      (λ () (read-json))))
+
```
mflatt, line 165: Did you mean to remove `eval:error` here, since there's no error?

mflatt (changes requested):

pkgs/racket-doc/json/json.scrbl
jessealama: Ah, OK, thanks. Fixed.

pkgs/racket-doc/json/json.scrbl
jessealama: Oops -- quite right. Fixed.

mflatt:
Thanks!

Issue #3526, opened 2020-11-29
[Feature Request] Streamlining `gen:equal+hash`/`prop:equal+hash`


LiberalArtist:
Most uses of `gen:equal+hash`/`prop:equal+hash` conceptually amount to folding over a list of fields to be considered using the provided recursive equality/hashing function. Indeed, a close relationship between the functions is necessitated by the [requirement](https://docs.racket-lang.org/reference/Equality.html#(def._((lib._racket%2Fprivate%2Fbase..rkt)._gen~3aequal%2Bhash))) that “`hash-proc` and `hash2-proc` should produce the same value for any two structures for which `equal-proc` produces a true value.”

I would like a linguistic mechanism to write such `gen:equal+hash`/`prop:equal+hash` implementations without boilerplate or error-prone duplication of the list of fields. Presumably a macro needs to be involved to facilitate short-circuiting for `equal-proc`.

On a related note, recent commits (e.g. https://github.com/racket/racket/commit/5ca183a7107efe2b7a69ed738b484a723f546658, https://github.com/racket/racket/commit/f34ff31aef5bb95f22c90afd6bcdee7d52cf7673, https://github.com/racket/racket/commit/b5a9fbb03d481e013d4a3898faaa9660f426b714, https://github.com/racket/racket/commit/56560736379fba7b4d72d97f29f46fa0cbf15b85, and https://github.com/racket/racket/commit/dc82685ce08a3d1f4b72958e1803c8fe3a14d443) have highlighted the importance of good hash mixing. Currently, implementers of `gen:equal+hash`/`prop:equal+hash` are left on their own to combine recursive hash results.

One model to consider is Swift Evolution proposal [SE-0206 “Hashable Enhancements”](https://github.com/apple/swift-evolution/blob/master/proposals/0206-hashable-enhancements.md), which “introduces a new `Hasher` type representing the standard library's universal hash function, and it extends the `Hashable` protocol with a new `hash(into:)` requirement that expresses hashing in terms of `Hasher`. This new requirement is intended to replace the old `hashValue` property, which is deprecated.”

LiberalArtist:
Apparently [Rust’s `std::hash`](https://doc.rust-lang.org/std/hash/) has a similar API that was considered in the design of Swift’s `Hasher`.

jackfirth:
[...]

PR #4546, opened 2023-01-18
Add hash-code operations

AlexKnauth:
Adds a `racket/hash-code` module, providing `hash-code-combine` and `hash-code-combine-unordered`.

 - [x] documentation for `hash-code-combine` and `hash-code-combine-unordered`
 - [x] update documentation for `hash-code-combine` to say something about mixing with one argument

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/equality.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/equality.scrbl
@@ -1,6 +1,7 @@
 #lang scribble/manual
 @(require (only-in scribblings/style/shared compare0)
-          "mz.rkt")
+          "mz.rkt"
+          (for-label racket/hash-code))
 
 
 @title{Equality}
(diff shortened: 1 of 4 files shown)


racket/collects/racket/hash-code.rkt
```diff
@@ -0,0 +1,46 @@
...
+(define (um2) (arithmetic-shift (most-positive-fixnum) 1))
+
+;; fixnum shift-left logical with wraparound
+(define fxsll/wraparound fxlshift/wraparound)
+
+;; fixnum logical exclusive or
+(define fxlogxor fxxor)
+
+;; fixnum shift-right logical
+;; shifting negative fixnums can produce large positive fixnums
```
mflatt, line 17: I don't see how that can be the case. It seems like an arithmetic shift should always preserve the sign of a negative fixnum. What am I missing?

mflatt: Oh, I'm missing that the "l" means logical!

racket/collects/racket/hash-code.rkt
```diff
@@ -0,0 +1,46 @@
...
+(define (mix-hash-code hc)
+  (fxmix-hash-code (->fx hc)))
+
+(define (fxmix-hash-code hc)
+  (let ([hc2 (fx+/wraparound hc (fxsll/wraparound (fx+/wraparound hc 1) 10))])
+    (fxlogxor hc2 (fxsrl hc2 6))))
+
+(define (hash-code-combine hc . rst)
+  (for/fold ([acc (->fx hc)]) ([v (in-list rst)])
+    (fx+/wraparound (fxmix-hash-code acc) (->fx v))))
```
AlexKnauth, line 42: Problem: if `(hash-code-combine A B)` = `(hash-code-combine-unordered (mix-hash-code A) B)`, then `(hash-code-combine-unordered (hash-code-combine A B) C)` = `(hash-code-combine-unordered (hash-code-combine A C) B)`, which might cause hash collisions where users wouldn't expect, for example `{[A, B], C}` might collide with `{[A, C], B}`.

If instead `(hash-code-combine A B)` is made to equal `(mix-hash-code (hash-code-combine-unordered (mix-hash-code A) B)`, then `{[A, B], C}` and `{[A, C], B}` implemented this way would be different.

racket/collects/racket/hash-code.rkt
```diff
@@ -0,0 +1,54 @@
...
+;; which is adapted from an `update` function in ChezScheme/s/newhash.ss
+;; and adapted from One-at-a-Time Hash in "A Hash Function for Hash Table Lookup"
+;; Bob Jenkins's expansion of his September 1997 Dr Dobbs article on Hash Functions
+;; http://www.burtleburtle.net/bob/hash/doobs.html#one
+(define (fxmix-hash-code hc)
+  (let ([hc2 (fx+/wraparound hc (fxsll/wraparound (fx+/wraparound hc 1) 10))])
+    (fxlogxor hc2 (fxsrl hc2 6))))
+
+;; NOTE: this `hash-code-combine` function is NOT the same as the
+;; `hash-code-combine` from racket/src/cs/rumble/hash-code.ss
```
sorawee, line 43: This is no longer accurate, since you also change `racket/src/cs/rumble/hash-code.ss`, correct?

sorawee:
@AlexKnauth I think there's a still a hole?

`(hash-code-combine 1 (hash-code-combine 2))` =  `(hash-code-combine 2 (hash-code-combine 1))`

So if I implement `hash([xs, ...]) = hash-code-combine(hash(xs), ...)`, then

`hash([1, [2]])` = `hash([2, [1]])`

rocketnia:
To avoid collisions between similarly shaped data of different types, I would expect a type tag to be incorporated into the hash somehow, like `hash([x, ...]) = hash-code-combine(hash-of-type-tag, hash(x), ...)`. (I'm not sure I understand the collisions under discussion or whether this factors into them.)

Something else I notice is that as of the latest code, `hash-code-combine` is a varargs function that allocates a cons list for its arguments and then iterates over it and throws it away. If that's the case, is`(hash-code-combine (recur a) (recur b) (recur c))` actually any more efficient than `(recur (list a b c))`? Is the idea to turn `hash-code-combine` into a macro or a `case-lambda` at some point?

AlexKnauth:
In what I've tested so far, type tags don't seem to solve the problem. At least, not how I've applied them:
```racket
(combine TAG A (combine TAG B))
 =
(mix (unordered (mix (unordered (mix TAG) A))) (mix (unordered (mix TAG) B))))
 =
(mix (unordered (mix (unordered (mix TAG) B))) (mix (unordered (mix TAG) A))))
 =
(combine TAG B (combine TAG A))
```

However, I do have another idea, inspired by encoding `(A, B)` as `{{A}, {A, B}}`:
```racket
(define (combine hc . rst)
  (for/fold ([acc (mix hc)]) ([v (in-list rst)])
    (mix (unordered (mix acc) (mix (unordered acc v))))))
```
which would make
```racket
(combine A) = (mix A)
(combine A B) = (mix (unordered (mix (mix A)) (mix (unordered (mix A) B))))
(combine A (mix B)) = (mix (unordered (mix (mix A)) (mix (unordered (mix A) (mix B)))))
(combine B (mix A)) = (mix (unordered (mix (mix B)) (mix (unordered (mix A) (mix B)))))
```

pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -632,3 +633,170 @@ and only access mutable data when the mode is true:
...
+  @examples[
+    (require racket/hash-code)
+    (struct ordered-triple (fst snd thd)
+      #:methods gen:equal+hash
+      [(define (equal-proc self other rec)
+         (and (rec (ordered-triple-fst self) (ordered-triple-fst other))
+              (rec (ordered-triple-snd self) (ordered-triple-snd other))
+              (rec (ordered-triple-thd self) (ordered-triple-thd other))))
+       (define (hash-proc self rec)
+         (hash-code-combine (rec (ordered-triple-fst self))
```
sorawee, line 656: Since this demonstrates how `hash-code-combine` can be used for `hash-proc`, I think it would be nice to show the best practice, which probably should include the type tag info.

AlexKnauth: Okay, I can add `(eq-hash-code struct:name)` to these examples.

AlexKnauth: I didn't add `(eq-hash-code struct:wrap)` to the `(hash-code-combine (rec (wrap-value self)))` examples because the point of those was to show `hash-code-combine` with only one argument.

sorawee: Does it make sense to add `(code:comment ...)` saying that "in practice, it is advisable to pass `(eq-hash-code struct:wrap)` into `hash-code-combine`"?

pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -747,6 +751,7 @@ and only access mutable data when the mode is true:
          (define p (rec (rotate-triple-paper self)))
          (define s (rec (rotate-triple-scissors self)))
          (hash-code-combine-unordered
+          (eq-hash-code struct:rotate-triple)
```
sorawee, line 754: This looks "wrong" to me, FWIW.

AlexKnauth: Just a feeling? Or is there a hash collision that this allows, that could be avoided by doing something else?

AlexKnauth: After some searching, I found this hash collision, involving `rotate-triple` and a `rotate-triple2` structure that's just an exact copy with only the name changed, to make
```racket
> (equal-hash-code (set (rotate-triple 'A 'B 'C) (rotate-triple2 'D 'E 'F)))
548402909030490366
> (equal-hash-code (set (rotate-triple2 'A 'B 'C) (rotate-triple 'D 'E 'F)))
548402909030490366
```

However... I personally don't think it's it's that important to fix this just for a documentation example.

pkgs/racket-doc/scribblings/reference/equality.scrbl
AlexKnauth: I added a comment saying this:
```racket
; demonstrates `hash-code-combine` with only one argument
; but it's good to combine `(eq-hash-code struct:wrap)` too
```

sorawee: Thanks!

pkgs/racket-doc/scribblings/reference/equality.scrbl
sorawee: I write many programs by adapting examples in the documentation, so it would indeed be nice to fix the example. I can submit a separate PR to fix it though if you don't want to do it yourself.

```
     (hash-code-combine
      (eq-hash-code struct:rotate-triple)
      (hash-code-combine-unordered
       (hash-code-combine r p s)
       (hash-code-combine p s r)
       (hash-code-combine s r p)))
```

should do the job, right?

AlexKnauth: I made the change... because after seeing your comment I also realized `rotate-triple` could get away with only
```racket
(hash-code-combine
 (eq-hash-code struct:rotate-triple)
 (hash-code-combine-unordered
  (hash-code-combine r p)
  (hash-code-combine p s)
  (hash-code-combine s r)))
```
and that simplification makes up for it

sorawee:
I think @rocketnia's comment on turning `hash-code-combine` and friends into `case-lambda` is the only outstanding issue. Though that could be a separate improvement too if performance issue is out-of-scope. Otherwise, the PR looks good to me.

LiberalArtist:
I definitely support relieving every `gen:equal+hash` implementer of the burden of reinventing hash mixing!

I'd thought about this a bit (but then did nothing) a few years ago as part of https://github.com/racket/racket/issues/3526. With the caveat that I have no expertise on hashing, there's one consideration from the comparison with Swift (and IIRC Rust?) that gives me pause here.

Swift Evolution proposal [SE-0206 “Hashable Enhancements”](https://github.com/apple/swift-evolution/blob/master/proposals/0206-hashable-enhancements.md) deprecated explicit custom implementations of `hashValue` (like our `equal-hash-code`) in favor of having programmers implement a method `func hash(into hasher: inout Hasher)`, where the `Hasher` type encapsulates the state of the universal hash algorithm. This is [their rationalle](https://github.com/apple/swift-evolution/blob/main/proposals/0206-hashable-enhancements.md#the-hashinto-requirement), which I find reasonably persuasive:

> At first glance, it may not be obvious why we need to replace
> `hashValue`. After all, `Hasher` can be used to take the guesswork out
> of its implementation:
> 
> ```swift
> extension GridPoint: Hashable {
>   var hashValue: Int { 
>     var hasher = Hasher()
>     hasher.combine(x)
>     hasher.combine(y)
>     return hasher.finalize()
>   }
> }
> ```
> 
> What's wrong with this? What makes `hash(into:)` so much better that's
> worth the cost of a change to a basic protocol?
> 
> * **Better Discoverability** -- With `hashValue`, you need to know
>   about `Hasher` to make use of it: the API does not direct you to do
>   the right thing. Worse, you have to do extra busywork by manually
>   initializing and finalizing a `Hasher` instance directly in your
>   `hashValue` implementation. Compare the code above to the
>   `hash(into:)` implementation below:
>   
>   ```swift
>   extension GridPoint: Hashable {
>     func hash(into hasher: inout Hasher) {
>       hasher.combine(x)
>       hasher.combine(y)
>     }
>   }
>   ```
>   
>   This is nice and easy, with minimal boilerplate. `Hasher` is part of
>   the function signature; people who need to implement `hash(into:)`
>   are naturally guided to learn about it.
>   
> * **Guaranteed Dispersion Quality** -- Keeping the existing
>   `hashValue` interface would mean that there was no way for `Set` and
>   `Dictionary` to guarantee that `Hashable` types produce hash values
>   with good enough dispersion. Therefore, these collections would need
>   to keep postprocessing hash values. We'd like to eliminate
>   postprocessing overhead for types that upgrade to `Hasher`.
>   
> * **Hasher Customizability** -- `hash(into:)` moves the initialization
>   of `Hasher` out of `Hashable` types, and into hashing
>   collections. This allows us to customize `Hasher` to the needs of
>   each hashing data structure. For example, the stdlib could start
>   using a different seed value for every new `Set` and `Dictionary`
>   instance; this somewhat improves reliability by making hash values
>   even less predictable, but (probably more importantly), it
>   drastically improves the performance of some relatively common
>   operations involving [copying data between `Set`/`Dictionary`
>   instances][quadratic-copy].
>   
> * **Better Performance** -- Similarly, `hash(into:)` moves the
>   finalization step out of `Hashable`. Finalization is a relatively
>   expensive operation; for example, in SipHash-1-3, it costs three
>   times as much as a single 64-bit `combine`. Repeating it for every
>   single component of a composite type would make hashing unreasonably
>   slow.
>   
>   &hellip;
>   
> Switching to `hash(into:)` gets us more robust hash values faster, and
> with cleaner, simpler code.
> 

[quadratic-copy]: https://bugs.swift.org/browse/SR-3268

I don't remember the details, but I'd noticed there was some short-circuiting in the implementation of `equal-hash-code`: https://github.com/racket/racket/blob/b5a9fbb03d481e013d4a3898faaa9660f426b714/racket/src/cs/rumble/hash-code.ss#L117-L133

This PR seems like a very good solution for the currently-existing world. I guess my question is really whether there's any appetite for moving to an API more like `hash-into`.

racket/collects/racket/hash-code.rkt
mflatt: Still, the result is a fixnum, so why is special treatment needed? (Sorry - intended to ask long ago, but apparently left the comment as "pending".)

AlexKnauth: Is there a better way to do a fixnum shift-right-logical that doesn't require converting a negative fixnum to a large positive big-integer that can be shifted-right back into a fixnum?

AlexKnauth: I've updated the `fxsrl` function to use a different method of calculating the logical right-shift when `a` is negative:
```racket
     (fx-/wraparound (fxrshift (bitwise-and a (most-positive-fixnum)) b)
                     (fxrshift (most-negative-fixnum) b))
```
This should do the same thing without needing to convert it to a large positive big-integer and back.

mflatt: The question I should have asked is "why use a logical shift instead of an arithmetic shift?". The answer might be "because hash-code.ss uses a logical shift". But should hash-code.ss use an arithmetic shift, instead? Does it matter either way?

AlexKnauth: A logical right-shift can preserve a distinction between negative numbers that would otherwise all get shifted to the same number, `-1`, by an arithmetic right-shift.

mflatt: Sorry that I'm being so dense here, but I still don't get it. What two negative numbers when shifted by 10 would produce the same result with sra but have different results for srl?

AlexKnauth: For them to be different they need to be shifted by different amounts. If one is shifted by 10 and the other by 9, they'll be different after a logical shift but the same `-1` after an arithmetic shift.

mflatt: So, it doesn't lose information in a hash-code calculation, since the function is only ever called with 10 as a second argument? It still seems like the readily available fixnum operation would work here.

mflatt:
@AlexKnauth Thanks for this improvement!

I think there needs to be argument checking in the functions in `racket/hash-code`, but I'm happy to add that after merging, if you prefer. One of the main checks is currently in `->fx`, which could take a "who" argument and complain as that function, instead of as `->fx`. (I would continue to avoid a dependency on the contract library here.)

I tried running some hashing benchmarks and programs, and as far as I can tell, the change to the internal implementation does work at least as well as the existing hashing. As intended, it works better for cases that have both `(list x (list y))` and `(list y (list x))`.

The only issue I see for merging is sorting out whether we really need a logical version of `fxrshift`.

jackfirth:
> This PR seems like a very good solution for the currently-existing world. I guess my question is really whether there's any appetite for moving to an API more like `hash-into`.

I'm in support of this for equality in Rhombus. For Racket, I'm not sure it's worth the trouble.

sorawee:
Random thoughts about Swift's approach.

1. I wonder why `finalize` is _that_ expensive?
2. How does Swift deal with unorderable fields?
3. Compared to our approach, Swift needs to allocate/construct the hasher. How is the performance affected?

PR #4547, opened 2023-01-21
Add dynamic-instantiate procedure to racket/class

lexi-lambda:
This PR adds a `dynamic-instantiate` procedure to `racket/class`, by analogy to `dynamic-send` and `dynamic-get-field`, which makes it possible to dynamically compute a set of by-name initialization variables. It seems like a fairly natural addition, and it’s quite easy to implement.


mflatt:
Thanks!

PR #4548, opened 2023-01-23
racket/unit: refactor many forms to use syntax/parse

lexi-lambda:
This PR significantly overhauls some of the key internal logic of `racket/unit`. As with its predecessor, #4543, there should be **no behavioral changes** aside from some minor alterations to error messages. The changes can be summarized as follows:

  1. The code that handles processing of unit imports and exports has been restructured to represent parsed imports and exports as compile-time `signature` structures rather than as 5-element lists. This makes code that manipulates these values significantly easier to read, as it is no longer filled with uses of `cadddr`.

  2. Parsing and processing of imports, exports, and init dependencies is handled by syntax classes that provide a variety of convenience attributes, which reduces the need for procedural syntax manipulation in many places.

  3. The `unit`, `unit/new-import-export`, `invoke-unit`, `invoke-unit/infer`, `define-values/invoke-unit`, `build-unit-from-context`, `provide-signature-elements`, and `open` forms have been updated to use `syntax/parse`. (Some other unit forms continue to use `syntax-case`, but many of them also share the bulk of their implementation with the forms mentioned above.)

The first of the above points was really my main goal, and the other two were just necessary improvements along that path. I think the more structured representation is significantly easier to understand and should also be much easier to change.

### DIFF
--- a/pkgs/racket-test/tests/units/test-unit-contracts.rkt
+++ b/pkgs/racket-test/tests/units/test-unit-contracts.rkt
@@ -634,7 +634,7 @@
   (unit/c (import)
           (export (sig1 [x string?] [x number?]))))
 (test-syntax-error
- "unit/c: unknown signature"
+ "unit/c: expected identifier bound to a signature"
   (unit/c (import faux^) (export)))
 
 (test-contract-error "(definition bad-export@)" "bad-export@" "unit must export signature sig1"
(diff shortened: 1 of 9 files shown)


samth:
Potentially related to the compiled code size, this change increased the overall DrDr run time by about 1 minute.

PR #4553, opened 2023-01-26
Fix bitwise-and and bitwise-ior return types

### LINKED ISSUE #4552: fixnumness of bitwise-and is wrong
8.8 [cs]

**What program did you run?**

```
#lang racket
(require racket/fixnum)
(define (f a)
  (fixnum? (bitwise-and a (add1 (most-positive-fixnum)))))
(f (- (random 1) 1)) ;=> #t
```

It should return `#f`, as `(bitwise-and ......)` = `1152921504606846976`, which exceeds `(most-positive-fixnum)` = `1152921504606846975`.

N/A

`(random 1)` is intentionally used to prevent CP0 from inlining and reducing the whole function call to `#f`.

After CP0, the above program is transformed to:

```
  (letrec ([f.440 (lambda (a_1.441)
                    (#2%bitwise-and 1152921504606846976 a_1.441)
                    #t)])
    (variable-set!/define f5.439 f.440 'consistent)
    (call-with-module-prompt
      (lambda ()
        (#%call-with-values
          (lambda ()
            (#2%bitwise-and 1152921504606846976 (#2%- (random 1) 1))
            #t)
          print-values4.438)))
    (#2%void)))
```

We can see that an optimization pass somewhere reduces `(fixnum? (bitwise-and ....))` to `#t`, though this is unsound.

Originally discovered by @AlexKnauth

AlexKnauth:
Fixes #4552 by setting the return types of `bitwise-and` and `bitwise-ior` to fixnum only when the constant that would constrain them is a fixnum.


gus-massa:
I agree with the test at the Chez Scheme level, but the test at the racket level should be near [optimize.rktl#L3074-L3095](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/optimize.rktl#L3074-L3095) I prefer a version that uses `test-comp` to check them comparing the "bytecode".

(Note that the optimization for `bitwise-ior` is not implemented in BC, so one of the test is disabled. The other test for `bitwise-ior` run in BC and CS, just to ensure that in both version the compiler is not doing something silly.)

mflatt:
More test suggestions:

* I'd add this as the next-to last mat in `fx.ms` (to complement the cptypes tests — generally good to check the end result, like at the Racket level):
```
(mat fixnum?
  (fixnum? (bitwise-and (- (random 1) 1) (most-positive-fixnum)))
  (not (fixnum? (bitwise-and (- (random 1) 1) (add1 (most-positive-fixnum)))))
  (fixnum? (bitwise-ior (most-negative-fixnum) (random 1)))
  (not (fixnum? (bitwise-ior (sub1 (most-negative-fixnum)) (random 1))))
)
```

* Instead of "tests/racket/bitwise.rkt", I'd much prefer this near the end of "tests/racket/fixnum.rktl":

```
;; Regression tests related to `fixnum?` optimization

(test #t
      (lambda (a) (fixnum? (bitwise-and a (most-positive-fixnum))))
      (- (random 1) 1))
(test #f
      (lambda (a) (fixnum? (bitwise-and a (add1 (most-positive-fixnum)))))
      (- (random 1) 1))
(test #t
      (lambda (a) (fixnum? (bitwise-ior (most-negative-fixnum) a)))
      (random 1))
(test #f
      (lambda (a) (fixnum? (bitwise-ior (sub1 (most-negative-fixnum)) a)))
      (random 1))

```

gus-massa:
LGTM, just squash everything before merging.

Is it too late to include this in 8.8?

mflatt:
Thanks for the repair and extensive tests!

I'll suggest merging for v8.8 to release management.

PR #4554, opened 2023-01-27
Support multiple values in stream-cons + use stream-cons in stream operations

sorawee:
This PR supports multiple values in `stream-cons`. As a result, `stream`, `stream*`, and `for/stream` also support multiple values.

`stream` and `stream*` patterns are also adjusted to allow matching against multiple-valued elements.

We then switch stream operations which need to handle multiple values to use `stream-cons`, both simplifying the implementation and making these operations properly memoizing at the same time. The switch also fixes the non-constant space problem raised in https://racket.discourse.group/t/stream-filter-not-in-constant-space/1643.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/stream.rktl
+++ b/pkgs/racket-test-core/tests/racket/stream.rktl
@@ -344,16 +344,16 @@
   (let ()
     (define-values (xs cpu real gc) (time-apply (λ () e) '()))
     (when (run-unreliable-tests? 'timing)
-      (test #t < real 500))
+      (test #t < real 100))
     (apply values xs)))
 
 (let ()
   (define st
-    (for/stream ([i 10000001])
-      (modulo i 10000000)))
+    (for/stream ([i (in-naturals)])
+      (modulo i 2000000)))
   (define st* (terminate-quickly (stream-filter zero? st))) ; should be fast
   (terminate-quickly (stream-rest st*)) ; should be fast
-  (test 0 stream-first (stream-rest st*)) ; should take time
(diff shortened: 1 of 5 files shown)


pkgs/racket-test-core/tests/racket/stream.rktl
```diff
@@ -220,10 +220,175 @@
...
+  (let ()
+    (define-values (xs cpu real gc) (time-apply (λ () e) '()))
+    (when (run-unreliable-tests? 'timing)
+      (test #t < real 500))
+    (apply values xs)))
+
+(let ()
+  (define st
+    (for/stream ([i 10000001])
+      (modulo i 10000000)))
```
mflatt, line 353: This test takes a while — as intended, but it's slower than I would normally include in this test suite. Would the test still work about as well with one less zero?

sorawee: On my machine (M1), it takes 2.2s.

What about changing the number to `2000000`, which would reduce the execution time to ~0.5s? And we then adjust `terminate-quickly` to check that `real` is less than `100`.

mflatt: > On my machine (M1), it takes 2.2s.

I bet it takes 21s on your machine if you run BC. Those times really are slower than I would normally include in this test suite — especially when the result is only used when opting into unreliable timing tests (so the work usually has no benefit).

If the test needs to be so large to be effective, then maybe it should be in "racket-test/tests/racket"? (It doesn't seem so large/slow that is need to be in the "stress" subdirectory, though.)

PR #4556, opened 2023-01-28
Fix contracts for generate-temporaries

sorawee:
I avoid using the name stx-pair, since many non syntax pairs, like a regular list or an empty syntax list, are also acceptable.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/stx-ops.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/stx-ops.scrbl
@@ -376,8 +376,7 @@ If @racket[shift] is @racket[0], then the result is @racket[stx].}
 
 Returns a list of identifiers that are distinct from all other
 identifiers. The list contains as many identifiers as
-@racket[v] contains elements. The @racket[v] argument
-must be a value that can be flattened into a list (satisfying @racket[stx-list?]).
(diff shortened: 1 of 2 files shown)


pkgs/racket-doc/scribblings/reference/stx-ops.scrbl
```diff
@@ -370,14 +371,14 @@ at @tech{phase level} 0 are shifted to the @tech{label phase level}.
...
+@defproc[(generate-temporaries [v stx-list?])
          (listof identifier?)]{
 
 Returns a list of identifiers that are distinct from all other
 identifiers. The list contains as many identifiers as
-@racket[stx-pair] contains elements. The @racket[stx-pair] argument
-must be a syntax pair that can be flattened into a list. The elements
-of @racket[stx-pair] can be anything, but string, symbol, keyword
+@racket[v] contains elements. The @racket[v] argument
+must be a value that can be flattened into a list (satisfying @racket[stx-list?]).
```
mflatt, line 380: Since the corrected contract already says that `v` must be `stx-list?`, should this sentence be dropped?

sorawee: That makes sense. I dropped the sentence.

PR #4557, opened 2023-01-29
Reorganize the implementation of racket/unit

lexi-lambda:
This is yet another PR that refactors the implementation of `racket/unit`. As with its predecessors, #4543 and #4548, there should be **no behavioral changes** aside from some minor alterations to error messages. However, this PR is somewhat different from the previous two, as it overwhelmingly consists of just moving code around.

I have reorganized the implementation as follows:

  * Most of the implementation has been moved into `racket/private/unit/`. Modules intended to be imported `for-syntax` live in `racket/private/util/exptime/`.

  * The implementation of signatures has been broken into three modules: `exptime/signature.rkt` contains the core structure definitions, `exptime/import-export.rkt` handles import and export processing, and `signature.rkt` implements `define-signature` and other signature forms.

  * The implementation of the unit forms has been broken into `unit-core.rkt` and `unit-infer.rkt` modules, the latter of which defines inferred-linking forms separately.

I have also deleted several unused definitions and attempted to remove some duplication of logic where possible. I adjusted implementation of a few forms very slightly, but nothing has been significantly changed.


mflatt:
Looks great - thanks!

PR #4560, opened 2023-01-29
Add fxrshift/logical

AlexKnauth:
Adds `fxrshift/logical` to `racket/fixnum`, with the same behavior as Chez Scheme's `fxsrl`.

With the sign bit treated as just another bit, a logical right-shift of a
negative-signed fixnum can produce a large positive fixnum.

Can be useful for hash-code operations, see #4546


mflatt:
I've avoided adding this operation, because it seems stranger than most fixnum operations (which have a generalization to non-fixnums).

Can you explain more how it's useful for hashing? For #4564, it still seems like plain `fxrshift` would work.

AlexKnauth:
The way `mix-hash-code` uses a logical right-shift, it always shifts by 6: `(fxrshift/logical x 6)`. But if there are multiple mixes involved, or multiple shifts involved, `((fxrshift/logical x 6) 6)` is equivalent to `(fxrshift/logical x 12)`.

If `x` is a negative number, shifting twice like that produces a different answer from shifting once, with a logical shift. But with an arithmetic shift, both might be funneled to the same `-1`. For example: with an arithmetic shift `(fxrshift -50 6) = -1` and `(fxrshift -50 12) = -1`, but with a logical shift `(fxrshift/logical -50 6) = 2^{55} - 1` and `(fxrshift/logical -50 12) = 2^{49} - 1`.

The `mix-hash-code` does more stuff than just a right-shift by 6, so `(mix-hash-code (mix-hash-code x))` and `(mix-hash-code x)` are still different in every case I've checked. But funnels where different numbers of mix operations converge on the same answers seem like they're more likely to happen when one of the operations in `mix-hash-code` tends to funnel things towards `-1`, and less likely to happen when it can shift through `-1` into the set of numbers one less than a power of 2.

mflatt:
For every pair of negative-number shifts (one at 6 and one at 12, say) that land at the same value with an arithmetic shift, isn't there a negative and a positive number that land at the same value with logical shifts but would be different for arithmetic shifts?

For hashing, it seems like there's the same amount of information preservation in both an arithmetic shift and a logical shift. In either case, after a shift by N, N bits of information has been lost. Whether you make the top bits the same as the formerly top bit or not wouldn't seem to matter.

AlexKnauth:
After some searching, I found a collision that happens with `rshift` but doesn't happen with `rshift/logical`.

`mix/a` uses an arithmetic shift `fxrshift`, while `mix/l` uses a logical shift `fxrshift/logical`.

On zeros, Both `(mix/a 0)` and `(mix/l 0)` produce `#b10000010000`, or `1040` in decimal. But on a number with a pattern of alternating runs of 10 ones and 10 zeros, some of those can cancel each other out and behave like `0`:
```
(mix/a #b111111111100000000001111111111000000000011111111101111111111)
= #b10000010000 = 1040
(mix/a 0)
= #b10000010000 = 1040
```
This is a case where the the top bits mostly cancel out to make the intermediate `hc2` = `#b-10000000001`, or `-1025` in decimal.

Then `mix/a` arithmetic-shifts `-1025` into only `-17` and produces `(fxxor -1025 -17) = 1040`.

Where `mix/l` logical-shifts `-1025` into `36028797018963951`, and produces `(fxxor -1025 36028797018963951) = -36028797018962928`, avoiding the collision.

AlexKnauth:
The example I show above with `0` appears to be generalizable to any fixnum:
```racket
(define (10runs- n)
  (fx-/wraparound #b111111111100000000001111111111000000000011111111101111111111 n))
```
For all fixnums `n` that I've tried, `n` and `(10runs- n)` collide with each other under `mix/a`, but `mix/l` avoids those collisions.

mflatt:
Ok! I still don't really understand why arithmetic shift preserves more here — an interaction with addition? — but it occurred to me that it's easy to just check all fixnums for a collision on a 32-bit build, and I do find one collision for `mix/a` for each of some random integers, and I don't find any for `mix/l`.

AlexKnauth:
I'm curious, do the collisions for those integers on a 32-bit build follow a pattern like the `10runs-` one?

As a guess, do they follow a function like this?
```racket
(define (10runs-/32 n)
  (fx-/wraparound #b11111111101111111111 n))
```

mflatt:
Yes, they do.

Do the further changes here look ok? https://github.com/mflatt/racket/tree/fxrshift (I imaging squashing into your commit.)

Goals of those changes:

* Add an unsafe variant of `fxrshift/logical`.

* Fix BC procedure flags to be consistent with the JIT (i.e., no inlining).

* Adjust documentation to not have markdown.

* Adjust examples to not include an example that would fail on a 32-bit platform or where rendering depends on the host platform type. (The platform probably leaks through in some other places, but I try to avoid that.)

* Reduce the set of examples to ones that seem more illustrative — but it's possible that I didn't understand what some of the examples were meant to show.

* Bumped version again, since I had meanwhile made other changes that bumped the version.

AlexKnauth:
Yes, after the repair for BC `unsafe-fxrshift/logical`, and I've pulled them on to the PR branch here.

gus-massa:
I finally understood why is important to use the "logical" shift.

The difference is that `mix/l` is a 1-to-1 transformation in `fixnum? -> fixnum?`, but `mix/a` is not. The result of `mix/a` is always a positive number, and each postive number has exactly two preimages.

If I cut&pasted correctly, the  relevant code is

```
#lang racket

(require racket/fixnum)

(define fxsll/wraparound fxlshift/wraparound)

(define fxlogxor fxxor)

(define fxsr/A fxrshift)

(define (fxsr/L a b)
  (cond
    [(and (fixnum? a) (negative? a) (positive? b))
     (fx-/wraparound (fxrshift (bitwise-and a (most-positive-fixnum)) b)
                     (fxrshift (most-negative-fixnum) b))]
    [else (fxrshift a b)]))

(define (fxmix/L hc)
  (define hc0 hc)
  (define hc1 (fx+/wraparound hc0
                              (fxsll/wraparound hc0 10)))
  (define hc2 (fx+/wraparound hc1 1024))
  (define hc3 (fxlogxor hc2
                        (fxsr/L hc2 6)))
  hc3)

(define (fxmix/A hc)
  (define hc0 hc)
  (define hc1 (fx+/wraparound hc0
                              (fxsll/wraparound hc0 10)))
  (define hc2 (fx+/wraparound hc1 1024))
  (define hc3 (fxlogxor hc2
                        (fxsr/A hc2 6))) ; <- the difference is here
  hc3)

(fxmix/L 0) ; ==> 1040
(fxmix/A 0) ; ==> 1040

; #b111111111100000000001111111111000000000011111111101111111111
(fxmix/L 1151796703138937855) ; ==> -36028797018962928
(fxmix/A 1151796703138937855) ; ==> 1040
```

Note that I made the code look much worse to make my explanation easier.

The first transformation `hc0->hc1` in `fxmix` is 1-to-1. To get the inverse, the trick is to realize that the last `10` bits of `hc0<<10` are `0`, so if you know `hc1` you get the last `10` bits for free. With the `0-9` bits, you can get the bits `10-19` combining the bits `0-9` you discovered and the bits `10-19` of `hc1`, but now there is a carry bit.  Now you can get the bits `20-29` with a similar method, but remember to use the carry bit and store the new carry bit. And so on, ... from `hc1` it's possible to discover the bits of `hc0` in chunks of `10` bits.

An alternative is to use that the definition of hc1 is equivalent to 

    (define hc1 (fx*/wraparound hc0 1025)

and it's invertible because `1025` is coprime with `(expt 2 62)`.

The second transformation `hc1->hc2` is trivially 1-to-1.

The third transformation `hc2->hc3` in `fxmix/L` is also 1-to-1. It's similar to the firs one, but in the inverse order. To get the inverse, the trick is to realize that the sign bit and the first `5` bits of `hc2>L>6` are `0`. So if you know `hc3` you get the sign bit and the bits `57-61`, or if you put all of them together to the bits `57-62`. With the `57-62` bits, you can get the bits `51-56` combining the bits `57-62` you discovered and the bits `51-62` of `hc3`, and luckily there is no carry bit.  Now you can get the bits `45-50` with a similar method. And so on, ..., everything is possible and deterministic. 

(I didn't find a nice algebraic trick for this, but the lack of carry makes everything more easy to see.)

The difference is that the third transformation `hc2->hc3` in `fxmix/A` is not 1-to-1. The trick is to realize that the sign bit of `hc2>A>6` is equal to the sign bit of `hc2` so when you `xor` them you always get `0`, so there are no negative results. Also, if the sign bit of `hc3` is `0` you have `2` options for the sign of if `hc2`. 

* If you pick `0` for the sign of `hc2`, the bits `57-61` of `hc2>A>6` are also `0`, and the rest is exactly like the "logical" case. You get exactly one result after this choice.

* If you pick `1` for the sign of `hc2`, bits `57-61` of `hc2>A>6` are also `1`, so the bits `57-61` of `hc2` are flipped, but after that point you can continue with the method of the "logical" case. You get exactly one result after this choice.

So, in the third transformation `hc2->hc3` in `fxmix/A` you get no preimages for the negatives and exactly `2` preimages for the non negatives.

PR #4561, opened 2023-01-29
added unicode normalization examples to racket string docs

dr-neptune:
Added a few examples for unicode string normalization and some examples for string->immutable-string

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/strings.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/strings.scrbl
@@ -55,7 +55,13 @@ whose positions are initialized with the given @racket[char]s.
 @defproc[(string->immutable-string [str string?]) (and/c string? immutable?)]{
 Returns an immutable string with the same content as
  @racket[str], returning @racket[str] itself if @racket[str] is
- immutable.}
+ immutable.
+
+@mz-examples[
+(immutable? (string #\H #\e #\l #\l #\o #\R #\a #\c #\k #\e #\t))
+(immutable? (string->immutable-string (string #\H #\e #\l #\l #\o #\R #\a #\c #\k #\e #\t)))
+]
+}
 
 
 @defproc[(string-length [str string?]) exact-nonnegative-integer?]{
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

After merging, I realized that the `string->immutable-string` example overflows the right-hand margin. Ok if I shorten the example string to just `(string #\H #\e #\l #\l #\o)`?

rfindler:
It looks like this broke the PDF build; specifically scribble doesnt' know how to generate latex for the character `ℌ` (that's the first error anyway).

rfindler:
Oh, whoops: changing the example doesn't help; the printer seems to be turning the string with the unicode code point number into the actual ℌ in the string and then it appears in the latex, regardless. So we probably do want to change scribble or change the example to be a string with a different character in it.

rfindler:
I've added the missing characters but things still don't work properly because the code that's identifying combining characters isn't working right for the example in `string-normalize-nfkc`. In particular, according to the comments in the file, [the call to `string-normalize-nfc`](https://github.com/racket/scribble/blob/master/scribble-lib/scribble/latex-render.rkt#L1053-L1056) is intended to merge characters together so that a character with a combining characters next to is is combined into a single thing. Then, [slightly lower down in the code, is a case to detect the combinations](https://github.com/racket/scribble/blob/master/scribble-lib/scribble/latex-render.rkt#L1053-L1056) by calling `string-normalize-nfd` and then getting the pieces out to latex.

This dance is tricked by the character in `ℋ̧` (aka `"\u210b\u327"`) since the call to `string-normalize-nfkc` doesn't combine it.

One "obvious" solution is just to have another case for situations like this (where the combination doesn't combine) but it occurred to me to ask @dr-neptune who seems to know a lot more about this normalization than I do(!) if the entire loop should be organized differently.

sorawee:
I tried Xe(La)Tex before (for the Thai language support). It works really well, but my impressed is that it's somewhat in a deprecated state. My understanding is that it may be better to target Lua(La)TeX, which is supposedly solving the same kind of problems that Xe(La)Tex does, and it's chosen as a successor for pdf(La)TeX.

Though to be fair to Xe(La)Tex, last time that I tried to typeset a Thai document, Xe(La)Tex works perfectly, while I have trouble getting Lua(La)TeX to work. But that's several years ago. I don't know what it's like right now.

samth:
My impression is that xelatex continues to work better and more out-of-the-box, but that lualatex has more active development.

otherjoel:
It looks like Scribble [already has a `--xelatex` option][1] (and `--lualatex` for that matter). Where is “the PDF build” happening and is there a reason it can’t be set up to use one of these instead?

[1]: https://docs.racket-lang.org/scribble/running.html

otherjoel:
> I tried Xe(La)Tex before (for the Thai language support). It works really well, but my impressed is that it's somewhat in a deprecated state.

That may very well be, I have never looked. Indeed all of LaTeX could be said to be “somewhat in a deprecated state” 😆 All I know is, I switched to xelatex eight years ago and my life got nothing but easier.

samth:
> It looks like Scribble [already has a `--xelatex` option](https://docs.racket-lang.org/scribble/running.html) (and `--lualatex` for that matter). Where is “the PDF build” happening and is there a reason it can’t be set up to use one of these instead?

The pdf build happens by passing the `#:pdf-doc?` option here: https://docs.racket-lang.org/distro-build/distro-build-language.html which uses the `--doc-pdf` option here: https://docs.racket-lang.org/raco/running.html neither of which has a way to use `xelatex`.

rfindler:
> Where is “the PDF build” happening and is there a reason it can’t be set up to use one of these instead?

It is running on a machine I administer that sits in a machien room at Northwestern and whose output eventually ends up [here](https://plt.cs.northwestern.edu/snapshots/) (well it produces other things too but that's the idea). To do so, it passes `#:pdf-doc? #t` as part of a [distro build](https://docs.racket-lang.org/distro-build/). I think this ends up doing `raco setup --doc-pdf` and it looks like that [ends up here](https://github.com/racket/racket/blob/master/racket/collects/setup/setup-core.rkt#L1469-L1494). 

I think it can be changed, as long as the change doesn't break stuff, but there's probably real work involved in figuring that part out!

rfindler:
For the time being, so the build starts succeeding again, I've pushed something that just piles on more complexity to the already complex code with [this commit](https://github.com/racket/scribble/commit/b1e268abfeb5e2837e613af93d128a7609ccf0b1). Maybe there's a better approach, tho.

PR #4563, opened 2023-01-31
ci: try fixing scanbuild

sorawee:


sorawee:
That doesn't seem to fix the issue, unfortunately. Sigh...

PR #4565, opened 2023-02-02
match: make duplicate hash keys always fail

sorawee:
In the hash-table pattern, the ground truth is the generic case, which compiles the pattern to list-no-order. Hence, patterns with duplicate keys will always fail.

The PR #4532 has a mistake because it allows duplicate keys with ..0 to match successfully. This PR amends #4532 to always fail such cases.

Also added a history note that should have been added in #4532


samth:
If that pr has a mistake then does it need to be omitted from the release? cc @jbclements

sorawee:
The `hash-table` pattern is doomed. As @shhyou suggested, we should abandon any change to it and create a new pattern instead. 

To be done: revert #4532 from master.

Issue #4550, opened 2023-01-24
RFC: Inter-signature dependencies in `racket/unit`


lexi-lambda:
Though it does not seem to get much use these days, I have always had an appreciation for `racket/unit`. First-class modules are useful. Unfortunately, just about every time I try to use it, I find I run into the same problem: it is difficult to bring bindings into scope that I want to use in my signatures’ contracts.

## Motivation

To illustrate the issue, suppose I have a signature that provides an interface to an abstract cell type:

```racket
(define-signature cell-type^
  [(contracted
    [cell? predicate/c]
    [make-cell (-> any/c cell?)]
    [cell-ref (-> cell? any/c)]
    [cell-set! (-> cell? any/c void?)])])
```

Now suppose I define another signature that provides some convenience operations on cells:

```racket
(define-signature cell-ops^
  [(contracted
    [cell-swap! (-> cell? any/c any/c)]
    [cell-copy (-> cell? cell?)])])
```

The problem with this signature is that the `cell?` predicate is not in scope. This is annoying, because the whole point of having a separate `cell-ops^` signature is to parameterize my `cell-ops@` unit over the particular `cell-type^` implementation. Currently, there are two available workarounds that I can see:

  1. Make `cell-ops^` extend or `open` the `cell-type^` signature (or just explicitly make `cell-ops^` also export `cell?`).

     This brings all the `cell-type^` bindings into scope, but it clutters the implementation of `cell-ops@`. Unlike with `require` and `provide`, a unit cannot directly re-export one of its imports, so `cell-ops@` must contain a `define` for each of the members of the `cell-type^` signature:

     ```racket
     (define-unit cell-ops@
       (import (prefix type: cell-type^))
       (export cell-ops^)
       (init-depend cell-type^)

       (define cell? type:cell?)
       (define make-cell type:make-cell)
       (define cell-ref type:cell-ref)
       (define cell-set! type:cell-set!)

       ....)
     ```

     This boilerplate can be partially mitigated using a macro, but it also complicates unit linking: inferred linking fails if multiple units export the same signature. That can be further worked around using signature tags, but tags introduce their own complications.

  2. Define each of the contracts of `cell-ops^` as separate signature elements, like this:

     ```racket
     (define-signature cell-ops^
       [cell-swap!/c
        cell-copy/c
        (contracted
         [cell-swap! cell-swap!/c]
         [cell-copy cell-copy/c])])
     ```

[...]

PR #4567, opened 2023-02-03
racket/unit: Improve and extend unit and signature contracts

lexi-lambda:
This commit makes several substantial changes to signature and especially unit contracts for the reasons described in #4550. Along with those changes, it also makes a handful of internal improvements to support their implementation. Here is the high level overview:

  * `unit/c` has been extended to be fully dependent. Subcontracts can refer to bindings from any listed signature.

  * The new version of `unit/c` implements `->i`-style “indy” blame semantics, so contracts are enforced within the contract expressions themselves, and violations within the contract expressions blame the contract.

  * The protocol for linking contracted signature bindings has been completely overhauled. Contract expressions in `contracted` forms are now only compiled and evaluated once per unit that exports the signature (rather than additionally in every unit that imports the signature). The new linking protocol is described in `Note [Linking for contracted imports and exports]` in `racket/private/unit/runtime`.

  * The implementations of `unit/new-import-export` and `unit/c` have been rewritten to (effectively) expand uses of `unit` and `define-values/invoke-unit` rather than to operate on import/export tables directly.

  * To support the previous bullet, `define-values/invoke-unit` has been extended to accept an optional `(values ....)` clause to bind any results returned by the unit body.

  * Also to support the previous bullet, `unit-from-context` has been rewritten to be lazier in certain circumstances, which allows reusing more machinery without causing evaluation to happen to early. `Note [Preserve import laziness]` in `racket/private/unit/runtime` describes the new strategy.

Unlike my other recent unit PRs, this PR is very much *not* just an internal refactor: it includes real behavioral changes. Most of the changes are extensions to existing functionality, but there are a few changes that pose a threat (at least in theory) to backwards compatibility, each of which I list below.

## Potential backwards incompatibilities

The issues in this section are listed in order of potential severity, from most severe to least severe.

### Scoping of signature bindings in `unit/c` subcontracts

Since `unit/c` has been made fully dependent, all signature bindings are now visible in all subcontract expressions. Since this changes what bindings are in scope for these expressions, it can change the semantics of existing programs in one of two ways:

  1. Signature bindings may capture free variables that previously referred to bindings in the context of the `unit/c` expression.

  2. It is now a syntax error to list two signatures in a `unit/c` contract that export a binding with the same name.

I suspect that the second of these issues is far more likely to be a problem than the first. Concretely, it means that the following program that previously compiled no longer compiles:

```racket
(define-signature a^ [x])
(define-signature b^ [x])
(unit/c (import) (export a^ b^))
```

To allow such contracts to be expressed, the grammar of `unit/c` has been extended to allow arbitrary signature `import` or `export` specs. The above program can therefore be fixed by modifying the contract expression as follows:

```racket
(unit/c (import) (export a^ (prefix b: b^)))
```

I have not yet found any programs where this causes an issue, but I am essentially only compiling the main distribution. There are two ways the potential damage of this change could be reduced:

  * For uses of `unit/c` that include no subcontract expressions (such as the example above), there is no need to refer to any signature bindings at all, so the need to do renaming is a bit silly. Therefore, `unit/c` could be changed to handle that case specially, which I suspect would eliminate the large majority of potential problems. However, it would make the semantics a little more confusing.

  * The compatibility problem of conflicting bindings could be completely eliminated by identifying conflicting bindings and making them only accessible within their containing signature (without explicit renaming). This would make `unit/c` completely backwards compatible in this respect except for potential free identifier capture. However, I am reluctant to do this because I think it could be very confusing.

### Internal enforcement of `unit/c` contracts

The existing implementation of `unit/c` does not enforce contracts on bindings that are used within the contract expressions themselves. In this way, the existing implementation of `unit/c` is like `->d`. My new implementation *does* enforce these contracts internally, and it blames the `unit/c` contract itself for any internal violations, just like `->i`.

This change comes with two potential compatibility problems:

  1. Obviously, contracts are now enforced where previously they were not, so the semantics is straightforwardly different. I strongly suspect this is actually a non-issue. I doubt any programs exist that rely on this.

  2. It has a performance cost. `->i` includes explicit dependency specifications to avoid doing extra work as much as possible, but `unit/c` does not. This means *every* subcontract projection must now be applied twice (once with the normal blame and once with the “indy” blame) where previously it only needed to be applied once.

The second of these issues could be mitigated to some extent by adding explicit dependency specifications between signatures, but since `unit/c` was already somewhat dependent and lacked any such dependency specifications, the gains would be fairly minimal.

A different option would be to leave `unit/c` unmodified and implement a new contract combinator that supports full dependency and uses the indy blame semantics. I am reluctant to do this simply because it would be more code to maintain, and I don’t think it’s worth the compatibility burden.

### Changes to initialization

The new implementations of `unit/new-import-export` and `unit/c` expand to a use of `unit` where previously they directly operated on import and export tables.

For `unit/new-import-export`, there should be no changes in functionality, but there could be changes in performance. However, whether the changes are better or worse is not straightforward when contracts are involved, due to the way contracts are now evaluated fewer times.

The story is mostly the same for `unit/c`, but `unit/c` includes subexpressions that are evaluated at unit invocation time. The new implementation can sometimes evaluate these expressions in a different order from the old implementation. This applies to both evaluation of the contracts themselves and the application of their projections. This is theoretically observable, but I suspect any program depending on this order was brittle in practice, anyway. The new order is based on the order in which the programmer specifies the contract clauses, so it gives the programmer more control and should be more predictable.


lexi-lambda:
I’ve unfortunately just realized one way in which the initialization order changes in this PR can cause some trouble, and I don’t have time to fix the issue this evening. I’ve marked the PR as a draft for now, but I should be able to fix it at some point over the next couple days.

mflatt:
I see that installing the package "lwc2016" makes DrRacket not start up, and it looks like an initialization-order problem. The same one?

FWIW, below is a list of non-main-distritribution packages registered at pkgs.racket-lang.org that seem to use units (to the best of my ability to grep and then filter by hand).

mm
pl
lwc2016
typed-minikanren
handin
smart-completion
minikanren
syndicate-classic
swindle
get-bonus
DrRacketTheme
tightlight
cover-codecov
rkt-tree-widget
mischief
fmt
version-case
dracula
ricoeur-tei-utils
text-markdown
mischief-dev
rosetta
cover-lib
combinator-parser
pop-pl
sirmail
aws

sorawee:
fmt formats `unit` forms, but doesn't use `unit` by itself. Please feel free to cross it out.

lexi-lambda:
> I see that installing the package "lwc2016" makes DrRacket not start up, and it looks like an initialization-order problem. The same one?

I’m not certain, but I did observe DrRacket not starting up due to an initialization order problem, so that seems plausible. I’ve just pushed a fix that reworks some things to be lazier while preserving sharing. `Note [Preserve import laziness]` in `racket/private/unit/runtime` describes the new approach.

Thanks for the package list—I’ll do my best to see if I can verify that none of them are broken by the new version of this patch. I do happen to already know that Typed Racket needs updating to cope with some changes to the expansion, so I’ll look into that as well.

lexi-lambda:
It looks like the TR issue is due to changes to the `unit-from-context` form. Without getting into the details: it would be possible to update TR to handle the new expansion, but the new expansion is cleverer, which also means it’s less regular and more obnoxious to parse. What’s more, once I started reading the relevant bits of the TR implementation, I quickly spotted a soundness hole in the existing implementation, even without my patch (racket/typed-racket#1305).

This whole business of groveling around in the expansion of unit forms and praying that TR finds the right things makes me uneasy. It would be both much nicer and significantly less fragile if `unit` could somehow cooperate with Typed Racket, but I’m not entirely sure what that might look like. Ideally, `unit` could attach syntax properties that TR could consume, but I’ve found that itself can be tricky business: scope adjustments do not automatically recur into syntax property values, so syntax-valued properties can rapidly go out of date.

@samth, is there some strategy that you might suggest here? Is there some approach that TR has used for other forms that could be of help? I have not seriously touched the TR codebase in quite some time, so I am not sure what the latest and greatest TR technology is like.

lexi-lambda:
I have verified that the test suites of the following packages pass after this PR:

* mm
* pl
* typed-minikanren
* smart-completion
* minikanren
* swindle
* DrRacketTheme
* tightlight
* cover-codecov
* rkt-tree-widget
* version-case
* text-markdown
* cover-lib
* combinator-parser
* pop-pl
* aws

The following packages produced some failures, but the failures were equivalent before and after this patch:

* lwc2016
* syndicate-classic (I could not even get this to build)
* get-bonus
* mischief
* ricoeur-tei-utils

I did not figure out how to run all the tests for handin and sirmail, as they seem to require some context setup for some of the tests. But the tests I did run did not seem to have any issues.

lexi-lambda:
I believe the only outstanding issue on this PR is a strategy for TR compatibility. That is an important concern, so it should not be merged until that is worked out, but I have marked it as ready for review anyway, as it’s a big patch, and I don’t expect TR compatibility to require significant changes. (I have an idea for the TR issue, which I will try today.)

mflatt:
I'm unclear on the status with respect to TR. It sounds like there's an old problem, but maybe no new problems from this PR. Or do the changes here still need a TR change other than racket/typed-racket#1303 to preserve enough existing behavior?

lexi-lambda:
I have pushed an additional commit to this PR that includes adjustments to the expansion of `unit-from-context` to assist Typed Racket. I’ve also opened racket/typed-racket#1306, which uses the new syntax properties, and I’ve confirmed that the full TR units test suite passes with that change.

pkgs/racket-doc/scribblings/reference/units.scrbl
```diff
@@ -314,7 +319,17 @@ treated as a kind of import into the local definition context. That
...
-@racket[define-values/invoke-unit] form.}
+@racket[define-values/invoke-unit] form.
+
+If no @racket[maybe-results-clause] is provided, the unit body may return
+any number of values, all of which are ignored. Otherwise, the values
+returned from the unit body are bound to the given @racket[result-id]s,
+in order. If no @racket[rest-results-id] is provided, the body must return
+exactly as many values as there are @racket[result-id]s, but if it is
+provided, the body may return arbitrarily many more, and
+@racket[rest-result-id] is bound to a list containing the extra results.
```
mflatt, line 330: Small typo: missing `s` in `rest-results-id`.

pkgs/racket-doc/scribblings/reference/units.scrbl
```diff
@@ -314,7 +319,17 @@ treated as a kind of import into the local definition context. That
...
+
+@history[#:changed "8.8.0.2" @elem{Added @racket[maybe-results-clause].}]}
```
mflatt, line 332: We're up to 8.8.0.6. Bumping to 8.8.0.7 (by changing `racket/src/version/racket_version.h` and `pkg/base/info.rkt`) would make sense, especially in combination with the TR change.

pkgs/racket-doc/scribblings/reference/units.scrbl
```diff
@@ -547,11 +562,17 @@ information of the signatures for the unit's imports (i.e., the lexical
...
-@defform/subs[
-#:literals (export link)
-(define-values/invoke-unit/infer maybe-exports unit-spec)
-[(maybe-exports code:blank (export tagged-sig-spec ...))
- (unit-spec unit-id (link link-unit-id ...))]]{
+@defform[
+#:literals (export link values)
+(define-values/invoke-unit/infer unit-spec
+  maybe-exports
+  maybe-results-clause)
```
mflatt, line 569: The old order should be listed here, too, since it's still supported. I'd use `defform*`.

pkgs/racket-doc/scribblings/reference/units.scrbl
```diff
@@ -759,11 +791,21 @@ superset of the export signatures listed in the unit contract. Additionally,
...
+@history[
+ #:changed "8.8.0.2" @elem{Changed @racket[sig-spec-block] to allow arbitrary
+                           @racket[tagged-sig-spec]s instead of only allowing
+                           @racket[tagged-sig-id]s.}
+ #:changed "8.8.0.2" @elem{Made bindings from @emph{all} signatures visible in
+                           the scope of each @racket[contract] expression
+                           instead of only the bindings from the same signature.
+                           Additionally, contracts on signature bindings are
+                           enforced for uses that appear in the @racket[contract]
+                           expressions themselves.}]}
```
mflatt, line 809: I'd normally put both changes in the same `#:changed`, so doing that is probably more consistent with other documentation, but I don't feel strongly that it has to be that way.

mflatt (changes requested):
Small documentation changes requested.

pkgs/racket-doc/scribblings/reference/units.scrbl
lexi-lambda: I certainly don’t have a strong opinion, so I’ve merged them together into the same use of `#:changed`.

pkgs/racket-doc/scribblings/reference/units.scrbl
lexi-lambda: I’ve done this, and I’ve added a paragraph about it in the body, so I’ve also pared down the wording in the `@history` annotation. Let me know if the updated version looks good.

lexi-lambda:
Thanks for your comments—I’ve pushed a commit that I think fixes all of them. I’ve also pushed a separate commit with a small fix (and test case) for a minor but mostly unrelated units bug that I happened to notice today.

PR #4568, opened 2023-02-04
Improve error messages in when/unless

sarna:
When you forget any of the parts of `when`/`unless`, the error you'll get is `"bad syntax"` with no additional info. It can be frustrating, especially if you have misplaced parens. This PR makes error messages more useful in the common cases when there's either test expression or body missing, and keeps `"bad syntax"` for all other cases.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/syntax.rktl
+++ b/pkgs/racket-test-core/tests/racket/syntax.rktl
@@ -234,9 +234,9 @@
 (test 0 'when (when (< 1 2) (cons 1 2) 0))
 (test-values '(0 10) (lambda () (when (< 1 2) (values 0 10))))
 (syntax-test #'when)
-(syntax-test #'(when))
+(syntax-test #'(when) #rx"missing test expression and body")
 (syntax-test #'(when . 1))
-(syntax-test #'(when 1))
+(syntax-test #'(when 1) #rx"missing body")
 (syntax-test #'(when 1 . 2))
 (error-test #'(when (values 1 2) 0) arity?)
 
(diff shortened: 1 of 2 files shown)


samth (commented):
I can't think of a way to hit it, but the fact that the (null? l) case isn't handled makes me a little nervous.

rfindler:
Is it worth doing `(null? (cdr l))` and `(null? (cddr l))` instead of the `(= (length ...` checks? Also, I think that it isn't possible to get an empty list bound to `l`, but it might not be a list at all, in a case like this: `(+ when 2)` so the `#f` error message could be improved with something like "expected an open paren immediately before `when`" or similar.

rfindler:
Also, this looks like [the spot](https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/syntax.rktl#L231-L241) for test cases for when. I think you can add lines like:

```
(syntax-test #'when #rx"something better than bad syntax")
```

and then run the tests with `racket -r syntax.rktl` to try them out.

mflatt:
Thanks for working on this!

I have some quibbles with the messages, though:

* The "not allowed in expression context" message seems confusing, because `when` forms are allowed in expression contexts. Also, this error might be triggered in, say, a definition context if `when` is used without parentheses there. Getting that message for `(when . 1)` seems particularly confusing to me. Something about using `when` not in a parenthesized sequence might be sensible — but pretty much all other forms just say "bad syntax", and I'm not sure it's a good idea to specialize the message for just `when` and `unless`. So, even though it seems contrary to your goal, I think "bad syntax" is the more consistent choice (at least unless we try to change messages more generally). That may be what you had originally.

* The message "missing test-expr" makes sense with respect to the documentation, but normally we don't rely on documentation abbreviations. I suggest "missing test expression and body".

mflatt:
Thanks!

PR #4569, opened 2023-02-05
list: support equal-always? as same? in check-duplicates and remove-d…

sorawee:
…uplicates


AlexKnauth:
It would be good to have tests that `check-duplicates` and `remove-duplicates` on `equal-always?` always produces the same result as they do with `(λ (x y) (equal-always? x y))`. For `remove-duplicates` it would be good to test that for lists both shorter and longer than 40 elements.

AlexKnauth (approved):

PR #4571, opened 2023-02-05
Update Honest Custom Equality docs

AlexKnauth:
Don't use "recur" on counts, don't use type-specific equality on mutable types. Type-specific equality on immutable types, such as `=` and `symbol=?`, are fine.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/equality.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/equality.scrbl
@@ -663,31 +663,16 @@ Type-specific equality functions on immutable types, such as
 
 @compare0[#:left "fine" #:right "bad"
 @racketblock0[
-  (struct thing (name)
+  (define (equal-proc self other rec)
     (code:comment "symbols are immutable: no problem")
-    #:methods gen:equal+hash
-    [(define (equal-proc self other rec)
-       (symbol=? (thing-name self) (thing-name other)))
-     (define (hash-proc self rec)
-       (+ (eq-hash-code struct:thing)
-          (rec (thing-name self))))
-     (define (hash2-proc self rec)
-       (+ (eq-hash-code struct:thing)
-          (rec (thing-name self))))])
+    (symbol=? (thing-name self) (thing-name other)))
 ]
 
 @racketblock0[
-  (struct thing (name)
-    (code:comment "strings can be mutable: wrongly accesses mutable data")
-    #:methods gen:equal+hash
-    [(define (equal-proc self other rec)
-       (string=? (thing-name self) (thing-name other)))
-     (define (hash-proc self rec mode)
-       (+ (eq-hash-code struct:mcell)
-          (rec (thing-name self))))
-     (define (hash2-proc self rec)
-       (+ (eq-hash-code struct:thing)
-          (rec (thing-name self))))])
+  (define (equal-proc self other rec)
+    (code:comment "strings can be mutable: accesses mutable data")
+    (code:comment "wrong for an struct that's not declared mutable")
+    (string=? (thing-name self) (thing-name other)))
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -634,6 +655,42 @@ and only access mutable data when the mode is true:
...
+
+@compare0[#:left "fine" #:right "bad"
+@racketblock0[
+  (struct thing (name)
+    (code:comment "symbols are immutable: no problem")
+    #:methods gen:equal+hash
+    [(define (equal-proc self other rec)
+       (symbol=? (thing-name self) (thing-name other)))
+     (define (hash-proc self rec)
+       (+ (eq-hash-code struct:thing)
```
sorawee, line 672: Should this use a hash code mixing function?

AlexKnauth: Mm. Maybe, but it doesn't feel like that's the relevant thing here. Maybe the example would be clearer without a `hash-proc` at all. Just the definition of `equal-proc` is enough, as long as it's clear that the struct it's for is immutable.

pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -531,6 +531,27 @@ and enables some cycle detection.
 ]
 ]
 
+Don't use the third argument to ``recur'' on counts of
+elements.
+When a data structure cares about discrete numbers, it can
+use @racket[=] on those, not @racket[equal?] or ``recur''.
```
jackfirth, line 537: This doesn't explain _why_ this is what you should do. The prose needs to motivate this advice by explaining how not following it can cause problems with `check-within`.

pkgs/racket-doc/scribblings/reference/equality.scrbl
```diff
@@ -634,6 +655,27 @@ and only access mutable data when the mode is true:
 ]
 ]
 
+Certain type-specific equality functions on mutable types,
+such as @racket[string=?], should also be avoided unless the
+struct is mutable or the mode is true.
+Type-specific equality functions on immutable types, such as
+@racket[symbol=?], are fine.
```
jackfirth, line 662: Why is it fine for immutable types? The advice needs more explanation about how we're trying to ensure `chaperone-of?` doesn't break.

AlexKnauth: A previous paragraph in the documentation attempts to explain it like this:

> Mutable structs will only use the custom equality for `equal?` and `impersonator-of?`, so that `equal-always?` and `chaperone-of?` don't change on mutation. Structs that represent mutable data should either be declared mutable, or use `equal-mode-proc` from `gen:equal-mode+hash` instead of `equal-proc` from `gen:equal+hash`, and only access mutable data when the mode is true

But that explanation isn't good enough to make it obvious that it applies to `string=?` being wrong here, like 3 codeblock-comparison examples down.

The first-pass guideline could be something like "don't access potentially-mutable data", and then exceptions like, "it's okay in `equal-proc` if the struct is declared mutable or okay in `equal-mode-proc` if the mode is true", can be separated later.

If I reorder them so the `string=?` comes first with the "don't access potentially-mutable data" guideline above it, that could be clearer.

pkgs/racket-doc/scribblings/reference/equality.scrbl
AlexKnauth: I added this sentence after:

> Using "recur" on counts is bad when a "recur" argument from `equal?/recur` is too tolerant on numbers within some range of each other.

PR #4576, opened 2023-02-11
add indexes for uppercase, upper-case, etc.

jbclements:
Add index terms so that searches for "uppercase" and "upper-case" succeed. Um... if I'm understanding the use of `index`, that is.


mflatt:
LGTM. I was uncertain about having both "uppercase" and "upper-case", but I see that this way covers both for searching, and neither one does by itself.

Issue #4580, opened 2023-02-15
Optimizer fails to eliminate branch on repeated `eq?` comparison


lexi-lambda:
This program exhibits a missed optimization:

```racket
(let ([special (gensym 'special)])
  (lambda (x)
    (if (eq? x special)
        (if (eq? x special)
            0
            1)
        2)))
```

Compiling with `PLT_LINKLET_SHOW_CP0=1` reveals the following optimized expression:

```scheme
(let ([special_1.12 (gensym 'special)])
  (lambda (x_2.13)
    (if (#2%eq? x_2.13 special_1.12)
        (if (#2%eq? x_2.13 special_1.12) 0 1)
        2)))
```

I would expect the inner `if` to be eliminated and replaced with the constant `0`.

(This issue is similar to, but distinct from, #4578.)

gus-massa:
Oops! https://github.com/racket/racket/pull/4582  is unrelated.

Issue #4578, opened 2023-02-14
Optimizer fails to remove redundant branch


lexi-lambda:
I am surprised by the optimized code produced by this program:

```racket
#lang racket/base
(#%plain-lambda (x1)
  (let* ([x2 (if (eq? x1 'special) 'special x1)]
         [x3 (if (eq? x2 'special) #f x2)])
    x3))
```

Running racket with `PLT_LINKLET_SHOW_CP0=1` yields essentially the following output:

```scheme
(lambda (x1_1.8)
  (let ([x2_2.9 (if (#2%eq? x1_1.8 'special) 'special x1_1.8)])
    (if (#2%eq? x2_2.9 'special) #f x2_2.9)))
```

I would expect the first `if` to be deleted.

I don’t know if missed optimizations rise to the level of being considered bugs, especially since Racket is very vague about what optimizations are guaranteed. But I think you can make the case that this is a “Macro Writer’s Bill of Rights” violation.

lexi-lambda:
Here’s a simpler version of this example that I think illustrates the general principle more directly:

```racket
#lang racket/base
(#%plain-lambda (x y)
  (if (eq? x y) y x))
```

I would like this to be optimized to `(#%plain-lambda (x y) x)`, since I think that optimization is sound in general.

gus-massa:
It's not a bug, it's a wish list addition request :) 

The first one should be working, so I'm very confused. Because cptype can "constant propagate" symbols form an `eq?` comparison. So it must work even in more general cases. So probably there is a bug. 

---

The second one is very difficult in the general case, for example

    (if (and (eq? x y) (eq? z w) (eq? x z)) y w)

The first problem is to keep the list of transitive equivalences in O(N) instead of O(N^2). According to my daughter and Wikipedia, it's possible, but I must check that the auxiliary data of the algorithm can be stored in the trees that are used inside cptypes.

Also, note that you can replace the last `y` by `w`, but not in the other direction, so only calculating the transitive equivalences is not enough.

Anyway, if the simple pattern with two variables appears in too many cases of idiomatic code, it is possible to add it as an special case. I prefer the cases that can be generalized, but in some occasions a special case is very useful. For example, `cp0` has a special case to detect the expansion of `or`, and in `BC` there is a special case for `(if (null? x) '() x)` that I never copied to CS but I probably should.

gus-massa:
In the fist one, I made a mistake and I read

[...]

PR #4582, opened 2023-02-15
Chez Scheme; Add more reductions for if and eq? in cp0

gus-massa:
In the first commit, I extend the one of the current reductions in cp0 to include 

    (if (eq? x y) x y) => (begin x y)

For now, `x` and `y` must be `record-equal?`[1] and `simple?`. Both are super restrictive, but I hope in the future to extend `record-equal?` and replace `simple?` with a weaker condition [2]. 

Anyway, I think this is general enough to include all the cases in https://github.com/racket/racket/issues/4578 and https://github.com/racket/racket/issues/4580  (Did I miss any of them?)

As explained in those issues, this is useful to reduce the code generated by functions with optional arguments in Racket.

[1] `record-equal?` is a very confusing name

[2] for example to reduce `(if (eq? (unbox x) 7) (unbox x) 7) => (begin (unbox x) 7)`

--

In the second commit, I add a reduction `(eq? x #f) => (if x #f #t)`.
Currently, `(not x)` is reduced to `(if x #f #t)` so I think is not harmful.
Expanding it makes it easier to trigger any of the other reductions, so I think it's better to expand it in this way instead of adding a special case for `(if (eq? x #f) a b)`. 

--

In the future, I'd like to add some support for `(if (null? x) '() x)`. I think it's better to add this extension in cptypes, and my idea is to just expand `(null? x) => (eq? x '())` and let the other reductions compose nicely.

### DIFF
--- a/racket/src/ChezScheme/s/cp0.ss
+++ b/racket/src/ChezScheme/s/cp0.ss
@@ -2659,7 +2659,17 @@
         [args #f])
 
       (define-inline 2 (eq? eqv? equal?)
-        [(arg1 arg2) (handle-equality ctxt arg1 (list arg2))])
+        [(arg1 arg2) (or (handle-equality ctxt arg1 (list arg2))
+                         (let ([val1 (value-visit-operand! arg1)]
+                               [val2 (value-visit-operand! arg2)])
+                           (cond
+                             [(cp0-constant? not (result-exp val1))
+                              (residualize-seq (list arg2) (list arg1) ctxt)
+                              (make-if ctxt sc val2 false-rec true-rec)]
+                             [(cp0-constant? not (result-exp val2))
+                              (residualize-seq (list arg1) (list arg2) ctxt)
+                              (make-if ctxt sc val1 false-rec true-rec)]
+                             [else #f])))])
 
       (define-inline 3 (bytevector=? enum-set=? bound-identifier=? free-identifier=? ftype-pointer=? literal-identifier=? time=?)
         [(arg1 arg2) (handle-equality ctxt arg1 (list arg2))])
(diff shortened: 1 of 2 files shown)


racket/src/ChezScheme/s/cp0.ss
```diff
@@ -906,6 +906,16 @@
                                 (make-if ctxt sc e11 (non-result-exp e12 e2) (non-result-exp e13 e3)))])))
                   #f)]
              [else #f])]
+          [(nanopass-case (Lsrc Expr) (result-exp e1)
+             [(call ,preinfo ,e10 ,e11, e12)
```
mflatt, line 910: Typo: space should be before last `,` instead of after

mflatt:
LGTM

It's interesting that the first commit's change fires rarely for the main distribution, but it fires several times for "cptypes.ss" where inlining plus other simplifications lead to `(if (eq? x 'bottom) x 'bottom)`.

racket/src/ChezScheme/s/cp0.ss
gus-massa: fixed

gus-massa:
In the second one, should I add `(bump sc 1)`?

mflatt:
> In the second one, should I add `(bump sc 1)`?

My best guess is that it's better to not add it, but I'm unsure.

gus-massa:
I don't understand if it's counting the size of the expression of the amount of work done. My guess is `(bump sc 2)` because  the reduction ads ìf``, `#f`and `#t` but there is a `#f` that was there before the reduction and is going to be eliminated because it's ignored, but the discarded parts don't reduce the counter.

Anyway, I think it's better to not add the `bump`.

PR #4586, opened 2023-02-23
make BOOTFILE_RACKET=racket stop pb fetch

benknoble:
The docs (build.md:249; section 1.6) already mention that setting `BOOTFILE_RACKET` should bootstrap the pb boot files instead of fetching them.

---

@mflatt I think I understood your Discord comment correctly, but I may have (read: probably) missed other spots where this is relevant. ~~I will test it, though.~~ Edit: building with `BOOTFILE_RACKET` from a very clean clone worked. I have not yet run any further tests.


mflatt:
This looks right - thanks!

PR #4588, opened 2023-02-27
setup: detect and report invalid package versions

Bogdanp:
Currently, `raco setup` fails with an exception when a package contains an invalid `version` spec (but only when said package is depended on by another package) ~~and also when a dependency spec contains an invalid version~~ (actually, this is already checked in `extract-pkg-dependencies`, but I don't think the additional check hurts here). I noticed this because of a recent issue introduced in `futures-visualizer`, but any package could run into this problem. This change makes it so that setup instead reports these as package problems. 

Before: 

```
raco setup: ...

version<?: contract violation
  expected: valid-version?
  given: '("1.1")
  argument position: 1st
  other arguments...:
   "1.1"
  context...:
   /Users/bogdan/sandbox/racket/racket/collects/version/utils.rkt:20:0: check-version-inputs
   /Users/bogdan/sandbox/racket/racket/collects/version/utils.rkt:17:7: version<?
   /Users/bogdan/sandbox/racket/racket/collects/setup/private/pkg-deps.rkt:28:0: check-package-dependencies
   /Users/bogdan/sandbox/racket/racket/collects/setup/setup-core.rkt:2117:2: do-check-package-dependencies
   body of "/Users/bogdan/sandbox/racket/racket/collects/setup/main.rkt"
   body of "/Users/bogdan/sandbox/racket/racket/collects/raco/main.rkt"
```

After:

```
raco setup: ...
raco setup: package depends on package whose version is invalid:
raco setup:   package: "racket-doc"
raco setup:   depends on package: "future-visualizer"
raco setup:   "future-visualizer" version: ("1.1")
raco setup: --- summary of package problems ---                    [12:40:20]
raco setup: package has invalid version:
raco setup:   package: "future-visualizer"
raco setup:   version: ("1.1")
raco setup: package has invalid version:
raco setup:   package: "libargon2-x86_64-macosx"
raco setup:   version: "20210625"
raco setup: package has invalid version:
raco setup:   package: "libargon2"
raco setup:   version: "20210625"
raco setup: package has invalid version:
raco setup:   package: "memoize-lib"
raco setup:   version: "3"
raco setup: package has invalid version:
raco setup:   package: "country"
raco setup:   version: "2022-12-11"
raco setup: package has invalid version:
raco setup:   package: "box-extra"
raco setup:   version: ("1.0")
```

One difference from before is that bad package versions are now reported regardless of whether those packages are depended on by other packages. While this might be initially a little noisy, I think this is probably a good thing, because it will make authors aware that versions need to follow a certain format. I hadn't been aware of that when I wrote the `libargon2` and `country` packages, for example.

### DIFF
--- a/racket/collects/setup/private/pkg-deps.rkt
+++ b/racket/collects/setup/private/pkg-deps.rkt
@@ -47,6 +47,7 @@
   (define dup-mods (make-hash)) ; modules that are provided by multiple packages
   (define pkg-version-deps (make-hash)) ; save version dependencies
   (define pkg-versions (make-hash)) ; save declared versions
+  (define bad-pkg-versions (make-hash))
   (define path-cache (make-hash))
   (define metadata-ns (make-base-namespace))
   (define pkg-dir-cache (make-hash))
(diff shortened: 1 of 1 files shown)


mflatt:
Thanks!

PR #4589, opened 2023-02-27
Chez Scheme: add more support for vectors in cptypes

gus-massa:
### Chez Scheme: add flvectors to equivalent-expansion? in tests

Add support for `flvector`s in `equivalent-expansion?` that is used to compare the expanded code in `cpo.ms` and `cptypes.ms` and other mats. I put some test in `cptypes.ms` because I lost a lot of time because the checks of the expansions of `flvector` in `cptypes.ms` failed mysteriously, and I'm worried that syncing upstream may delete this changes to `equivalent-expansion?`. Anyway, it makes more sense to move the checks to `5_6.ms`.

The first checks to ensure it's working, the other is about `NAN` and the other about `+/-0.0`  because I almost used `fl=` instead of `eqv?` to compare the numbers.

### Chez Scheme: add more support for vectors in cptypes

This is the main part. It includes `vectors`, `string`, `bytevectos`, `fxvectors`, `flvectors `and `boxes`. Currently the associated types are stored as a `symbol` in the `normalptr` slot and each union/intersection is a special case. After this change the null versions are stores in the `singleton` slot and the not-null versions are stored in the `multiplet` slot. Thee slots are bitfields, so the calculation must be slightly faster and there are a few additional reductions already programed for them.

I didn't make a difference between not-null-mutable-vectors and not-null-immutable-vectors, because  I don't expect that to be very useful. It's not hard to add it, but there are some problems because Chez Scheme vectors can be made immutable after they are constructed. 

In `primdata.ms` I named the not-null versions of the vectors as `vector*`. I't like `C*` that is the complex plane without the `0.0`, but I'm not sure if it's clear enough.

I was hopping to add something smarter like `(vector-of-lenght-at-least 7)`, but I want to finish the smarter code for `exact-integers` before that, but the code for `exact-integers` has too many special cases and will take forever. So I guess this is a good compromise for now.

### Chez Scheme: add x->immutable-x to cp0
 
This is a followup of https://github.com/racket/racket/pull/4220 but the problem in `rktboot `is fixed now. I could add a similar code in `cptypes` that would catch a few additional case but I think too few to worth the duplication.

I also added to `primvars.ms` the mutable and immutable version of vectors and similar stuff. Mostly so `cptypes` understand the predicates and can make some reductions like

    (lambda (x) (when (pair? x) (mutable-vector? x)))  ==> (lambda (x) (when (pair? x) #f))

I added the immutable version to a few primitives that use `immutable-vectors`, but I didn't add the mutable version because there are just too many and I don't expect the difference to be very useful. Perhaps add this in the future with a shorter name like `(vector, ivector, mvector)` (default = any) or `(xvector, ivector, vector)` (default = mutable).

### experr
Again, I have problem with the experr. Now after `git clean -dfx` they are almost fine, but I got a few errors in `thread.ms` that I blame to cosmic rays.


mflatt:
It looks like only `root-experr-compile-0-f-f-f` needs updating to sync. I added that change here.

Instead of a `*` to mean "non-empty", I suggest a `nonempty-` prefix, such as `nonempty-string`.

Otherwise, LGTM.

gus-massa:
I think `notnull-` is more consistent with the internal names in Chez Scheme, but `notempty-` is more clear. Which one do you prefer?

PR #4590, opened 2023-02-28
Recommend 100 columns instead of 102 columns

jackfirth:
The current recommended line width in the Racket style guide is [102 columns](https://docs.racket-lang.org/style/Textual_Matters.html#%28part._.Line_.Width%29). This is a rather bizarre number; more traditional line widths would be 80, 100, or 120 columns. I can't find any information about where the extra 2 columns come from or what their purpose is. This pull request changes the recommended line width to 100 columns to make it less bizarre.

If there's a reason it's 102 columns instead of 100 and we want to keep it at 102 columns, let me know so I can instead open a pull request to update the docs with the rationale. It's a strange enough choice that I think some explanation would be helpful.

**Changing this line width to 80 or 120 columns is explicitly out of scope for this pull request and discussion.** I'm just trying to tweak the line width to something normal, not start a giant bikeshedding session.


jackfirth:
I don't use emacs, so this may be a silly question, but couldn't people type ctrl-U 97 just as quickly as ctrl-U 99?

mfelleisen:
(When I say Emacs, I mean DrRacket with Emacs keybindings, those you indicated.) 

9-7 would mean two _distinct_ keys. For bad keyboarders, like myself, this raises the probability of hitting the wrong one.

jackfirth:
So the 102 limit only offers an advantage to people who use emacs keybindings?

mfelleisen:
It's been too long for me to recall vi bindings, and frankly, I'd rather not admit that I can also edit in vi and that even write code in VSC and IntelliJ.  [[ I will say that the line addition is also something I saw others use, like Gregor Kiczales then at Xerox now at UBC. ]] 

You asked for the rationale. This is it. And I will say this compromise made all people equally unhappy, which made me very happy because it meant it was kind of acceptable to write code this way. People who wanted 256 columns grumbled along. People who wanted less just wrote narrower code.

benknoble:
The Vi equivalent would be something like (assuming a blank line) `i;; <esc>99a-<esc>` (where the second insertion command could also be `a-<esc>99.`).

otherjoel:
I looked at the 102 number as being sort of in the spirit of [Hettinger’s “90ish”](https://youtu.be/wf-BqAjZb8M?t=326)[^1] preferred line length limit in Python. I.e., “you probably want to break it at 100, but it’s ok to go a *little* over”. I don’t know what the original reason was for Racket, but I like that it is a weird number because it might keep people from taking the limit too seriously.

[^1]: Timestamp link, should start at 5:26

PR #4594, opened 2023-03-01
Fix hash-hash-code 

### LINKED ISSUE #4593: CS immutable-hash-table hash-code ignores some values inside

Racket version 8.8.0.8--2023-03-01 [cs]

**What program did you run?**

```racket
> (equal-hash-code (hash 1 4 2 0 3 0))
608705005784926398
> (equal-hash-code (hash 1 4 2 5 3 6))
608705005784926398
```

The hash codes should be different because the values for keys `2` and `3` are different.

No error message, just some hash collisions.

The hash codes for mutable hash-tables are fine, and the code for those seems like it could be adapted to immutable ones without much trouble. Is there a reason it isn't that way already?

It seems to depend on only the keys and one of the values, ignoring the rest of the values.

AlexKnauth:
Fixes #4593 and some related issues found along the way.

### DIFF
--- a/pkgs/racket-test/tests/racket/hash-code.rkt
+++ b/pkgs/racket-test/tests/racket/hash-code.rkt
@@ -9,15 +9,47 @@
   ;; [A, B] != [B, A]
   (check-not-equal? (equal-hash-code (hash 0 1))
                     (equal-hash-code (hash 1 0)))
+  (check-not-equal? (equal-hash-code (make-hash '((0 . 1))))
+                    (equal-hash-code (make-hash '((1 . 0)))))
   ;; {[A, B], C} != {[A, C], B}
   ;; {{A: B}, {C: D}} != {{A: D}, {C: B}}
   (check-not-equal? (equal-hash-code (hash (hash 0 2) #true (hash 1 3) #true))
                     (equal-hash-code (hash (hash 0 3) #true (hash 1 2) #true)))
   (check-not-equal? (equal-hash-code (set (hash 0 2) (hash 1 3)))
                     (equal-hash-code (set (hash 0 3) (hash 1 2))))
+  (check-not-equal? (equal-hash-code (make-hash
+                                      (list (cons (make-hash '((0 . 2))) #true)
+                                            (cons (make-hash '((1 . 3))) #true))))
+                    (equal-hash-code (make-hash
+                                      (list (cons (make-hash '((0 . 3))) #true)
+                                            (cons (make-hash '((1 . 2))) #true)))))
+  (check-not-equal? (equal-hash-code (mutable-set (make-hash '((0 . 2)))
+                                                  (make-hash '((1 . 3)))))
+                    (equal-hash-code (mutable-set (make-hash '((0 . 3)))
+                                                  (make-hash '((1 . 2))))))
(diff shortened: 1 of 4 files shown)


mflatt:
It looks like the problem with `bnode-hash-code` is that it doesn't increment `offset` as it increments `i` in the loop at line 1158. Incrementing `offset` in the loop that corrects the immediate problem of ignoring some values.

Using the generic hash-table traversal on an immutable hash table seems to take 2-3 times as long as the specialized traversal, so keeping the specialized traversal seems worthwhile.

AlexKnauth:
Do you mean like changing
```racket
            (let ([offset (fx+ HAMT-STATIC-FIELD-COUNT child-count key-count)])
              (let loop ([i 0] [hc hc])
                (cond
                  [(fx< i val-count)
                   (loop (fx1+ i)
                         (hash-code-combine hc (hash (#%$stencil-vector-ref n offset))))]
                  [else hc])))
```
to
```racket
            (let ([offset (fx+ HAMT-STATIC-FIELD-COUNT child-count key-count)])
              (let loop ([i 0] [hc hc])
                (cond
                  [(fx< i val-count)
                   (loop (fx1+ i)
                         (hash-code-combine hc (hash (#%$stencil-vector-ref n (fx+ offset i)))))]
                  [else hc])))
```
Or some other incrementing like
```racket
            (let loop ([i 0]
                       [offset (fx+ HAMT-STATIC-FIELD-COUNT child-count key-count)]
                       [hc hc])
              (cond
                [(fx< i val-count)
                 (loop (fx1+ i)
                       (fx1+ offset)
                       (hash-code-combine hc (hash (#%$stencil-vector-ref n offset))))]
                [else hc]))
```
Or a variation that does something more complicated than `fx1+` there?

mflatt:
Yes, either of those work, unless I'm missing something (and adding `i` to `offset` for the ref seems better).

LiberalArtist:
I happened to be looking at some of these files this morning, and one thing that surprised me was that `eqv-hash-code` checks for `number?` and `char?` before calling `eq-hash-code`:

https://github.com/racket/racket/blob/34b48a05fb0f010120516f89d474a5a1f8c47af0/racket/src/cs/rumble/hash-code.ss#L94-L98

But `eq-hash-code` goes on to do exactly the same thing again as soon as it's finished checking `(and (symbol? x) (not (gensym x)))`:

https://github.com/racket/racket/blob/34b48a05fb0f010120516f89d474a5a1f8c47af0/racket/src/cs/rumble/hash-code.ss#L41-L42

AlexKnauth:
I broke up my changes to CS hash-hash-code into smaller commits, hoping to find out which one of those changes caused the [buildtest-win (CS) CI failure on `set/c32+opt/c`](https://github.com/racket/racket/actions/runs/4307851876/jobs/7513418898#step:9:43). But that test failure didn't pop up a 2nd time, even though its use of sets based on hash-codes seemed relevant to the changes here.

shhyou:
This commit https://github.com/racket/racket/commit/7a7f1756a62c62d9faca2fa8d63f29c5f53f9cda has similar test failures but the changes are unrelated to this PR, so the test failures might be irrelevant.

mflatt:
Thanks!

PR #4603, opened 2023-03-10
Support bootstrapping via Chez Scheme from top-level makefile

jryans:
It has been possible to supply a Chez Scheme for bootstrapping via configure args (`--enable-scheme`) for some time, but it was not clear how to do so if using the top-level makefile. This wires up that support in a similar way to the existing `RACKET`-related makefile options.


mflatt:
Thanks!

PR #4605, opened 2023-03-14
syntax/parse interpreter

rmculpepper:
This PR changes the implementation of syntax/parse to use an interpreter. Specifically:
- the environment of attribute matches is represented as a data structure and only reflected as attribute bindings around user expressions
- patterns are compiled to higher-order functions that use a backtracking monad
- especially simple patterns are represented by a run-time AST and interpreted

The benefits are reduced compiled code size and, hopefully, reduced memory use. (The implementation is arguably easier to understand, too.) The cost is that syntax-parse is somewhat slower. It would be useful to know if there are cases where the slowdown is a problem.


samth:
In the past, profiling Typed Racket has indicated that it spends a lot of time in syntax-parse. So I would expect a slowdown here to be bad for (the already problematic) TR compile times.

mfelleisen:
Perhaps you can get @bennn to measure the time it takes to get the GTP benchmarks through.

bennn:
For typechecking, the full GTP Benchmarks are overkill. I'd just use the typed code in `gtp-checkup` (`./benchmarks/*/typed/main.rkt`):
https://github.com/bennn/gtp-checkup

There's also the `tr-performance` benchmarks by @pnwamk. (I think you'll want to comment-out the plotting code and avoid the `schml` benchmarks to save time.)
https://github.com/pnwamk/tr-performance

mflatt:
Here are some measurements that maybe help put a scale on the performance changes.

I tried removing all ".zo" files in the "math" collection and then ran `racket -W "debug@GC:major error" -l math` before and after this patch. As expected, the patch makes loading (mostly expansion and typechecking) take longer, but slightly less memory is used, probably because there's less code implementing the matches:

```
          CPU Time   Peak Memory
          (seconds)     (MB)
before       67         409
after        74         406
```

Rhombus also uses `syntax-parse` extensively, and the Rhombus documentation involves a lot of examples run in a sandbox. A priori, just loading the Rhombus document seems like it could go either way, since a lighter expansion of Rhombus's implementation makes the sandbox setup faster, but then the example has to be syntax-parsed. This table also shows the total ".zo" size of Rhombus's core implementation. It ends up being slightly faster.

```
rhombus.scrbl    CPU Time   Peak Memory    rhombus/private/compiled/*.zo
                (seconds)     (MB)               (MB)
before            35.6        395                9.2
after             34.6        395                7.8
```

Here are some relevant lines from the output of`dump-memory-stats` after starting DrRacket (including code still in marshaled bytevector) to show a little more about memory use:

``` 
before
 bytevector                34,572   24,832,080  |     35,603   24,895,824
 code                     149,272  131,200,464  |    149,278  131,201,744
 total                  7,228,719  501,474,116  |  7,401,631  510,486,404
after
 bytevector                34,349   22,854,528  |     35,380   22,918,272
 code                     151,348  132,071,520  |    151,354  132,072,800
 total                  7,236,443  500,788,604  |  7,409,355  509,800,892
```

Possibly not so relevant, but are some timings using `racket -cl`, which conflates the compilation of syntax-parse itself, compilation of uses of `syntax-parse,` and running some uses of `syntax-parse`. These are faster and use much less memory, which probably reflects that there are lots of uses of `syntax-parse`, and only a few are used to expand the target program.

For `racket -cl math`:

```
          CPU Time   Peak Memory
          (seconds)     (MB)
before      122         523
after       120         481
```

For `racket -ct demo.rhm` (a Rhombus demo):

```
          CPU Time   Peak Memory
          (seconds)     (MB)
before       40.5       403
after        33.1       361
```

samth:
Thanks @mflatt. My view is that a 10% slowdown for Typed Racket is fairly substantial, but I like the idea of generating less code. Are there ways that we could use this change in some places but not others, perhaps limiting the runtime cost?

mfelleisen:
@samth  Do you think you think @mflatt’s 10% is a definitive conclusion? Wouldn't it make sense to check what @bennn proposed? Type checking performance may depend on particular uses. (My conjecture is that it varies a lot, and we may see much more and much less than 10%.)

mflatt:
I agree that more checking is needed.

@rmculpepper Did you already experiment with making small and simple patterns compiled while larger or more complex patterns are interpreted (which is what the `syntax-case` implementation does)?

mflatt:
Also, because I forgot to say so before: I'm enthusiastic about this change, assuming we can find a good balance of performance. Reducing code size has been a path to steady improvement in Racket CS. Every individual change contributes only a little, but it seems to add up, because reducing code size affects things like load time, GC time, and compile time.

mfelleisen:
I don't see any `++` in the "outcome" column, and my "back of the envelope" calculation (in my head) says up to 5% slowdown (on large-ish programs, like "acquire" or "schml-".  Is 5% a good trade off for space @mflatt ?

mfelleisen:
Hah! That's good. Is your change on a branch that @bennn could use for a 'rerun'?

rmculpepper:
Thanks Matthew and Ben for the benchmarks and measurements.

I've tried out a few changes that lead to minor improvements in the tr-performance times, but I haven't pushed them yet. They don't close the gap, in any case, and the gap is larger than 5% on most of the tr-performance benchmarks. I'd like to look at how Typed Racket is using syntax/parse in more detail, but that won't happen in time for the April release. I have reservations about keeping two implementations around in the long term, but it sounds like Matthew's heuristic is a good solution for now.

rfindler:
Having two implantations can be a great help in testing, at least. :)

Robby

mflatt:
@rmculpepper We're almost halfway through the v8.10 cycle, and it would be helpful to get this change in. Should I merge with the compile-syntax-classes heuristic for now?

PR #4609, opened 2023-03-18
Linklet: Add logging option to show all passes

jryans:
Chez Scheme's tracer will log the output of all passes if `#t` is supplied as a pass name to log. This wires that up to Racket's linklet-level `PLT_LINKLET_SHOW_PASSES` env var via the special pass name `all`.

While this does log a semi-overwhelming amount of info since there are so many passes, I am planning wire it up to a pass-by-pass view in Compiler Explorer, which should be more comprehensible.

I attached additional commits to fix a few other minor issues in this area that I happened to notice while working on this.


mflatt:
Thanks!

PR #4610, opened 2023-03-20
Chez Scheme: improve support for lists in cptypes

gus-massa:
Move list form the `normalptr` slot to the `multiplet` slot. So it's calculated automatically instead of having a special case in `normalptr`. Also, this allows the reduction of

    (lambda (x f) (unless (list-assuming-immutable? x) (f) (list? x)))

`pair`s are split in `list-pairs` and `nonlist-pairs`. The new problem is that a  `pair` may go from one classification to the other, so it's necessary a special case to reduce

    (lambda (x f) (when (list-assuming-immutable? x) (f x) (list? x)))

but not reduce incorrectly

    (lambda (x f) (when (list? x) (f x) (list? x)))

---

The first commit is the easy super safe version. The bounding of the `list` type is too wide, so it's necessary to add a special case for `list?` in `fold-predicate`.

---

The second commit is a more general version. The idea is to have 4 approximations to the real type (in order, from the smaller to the biggest):

* forever / non-extended
* now / non-extended
* [the real type that perhaps is not expressible]
* now / extended
* forever / extended

The names make sense in `fold-predicate`. The two in the middle are used to reduce `(pred? x)` to `#t` or `#f`. The two in the extremes are used to assign a type to the variable `x` in the branches of `(if (pred? x) T F)`.

They are also used in `fold-primref/next` to realize that the expression will raise an error or make it unsafe or recover the type for a variable for the following code. (My idea is to clean up the code in that function.)

But in `fold-primref/next` the names don't make too much sense. Any better suggestion?  Anyway,, my idea is to remove "forever" from the name, so the default version is safer and the "now" version is not used unintentionally.  

---

The third third commit adds similar cases for `sub-list` an a few variants used in primdata. 

---

My idea is to squash all of them before merging, but I'm still not 100% happy with the second part.

### DIFF
--- a/racket/src/ChezScheme/mats/cptypes.ms
+++ b/racket/src/ChezScheme/mats/cptypes.ms
@@ -230,18 +230,6 @@
   (cptypes/once-equivalent-expansion?
     '(lambda (x) (when (fixnum? x) (zero? x) 7))
     '(lambda (x) (when (fixnum? x) 7)))
-  (cptypes-equivalent-expansion?
-    '(lambda (x f) (when (list-assuming-immutable? x) (f x) (list-assuming-immutable? x)))
-    '(lambda (x f) (when (list-assuming-immutable? x) (f x) #t)))
-  (not (cptypes-equivalent-expansion?
-         '(lambda (x f) (when (list? x) (f x) (unless (list? x) 1)))
-         '(lambda (x f) (when (list? x) (f x) (unless (list? x) 2)))))
-  (cptypes-equivalent-expansion?
-    '(lambda (f) (define x '(1 2 3)) (f x) (list-assuming-immutable? x))
-    '(lambda (f) (define x '(1 2 3)) (f x) #t))
-  (cptypes-equivalent-expansion?
-    '(lambda () (define x '(1 2 3)) (pair? x))
-    '(lambda () (define x '(1 2 3)) #t))
 )
 
 (mat cptypes-type-if
@@ -662,7 +650,6 @@
   (test-chain* '(record? #3%$record?))
   (test-chain* '((lambda (x) (eq? x car)) procedure?))
   (test-chain* '(record-type-descriptor? #3%$record?))
-  (test-chain* '(null? list-assuming-immutable? list? (lambda (x) (or (null? x) (pair? x)))))
   (test-disjoint '(pair? box? #3%$record? number?
                    vector? string? bytevector? fxvector? symbol?
                    char? boolean? null? (lambda (x) (eq? x (void)))
@@ -676,11 +663,6 @@
   (test-disjoint '(integer? ratnum?))
   (test-disjoint '((lambda (x) (eq? x 'banana)) (lambda (x) (eq? x 'apple))))
   (test-disjoint* '(list? record? vector?))
-  (not (test-disjoint* '(list? null?)))
-  (not (test-disjoint* '(list? pair?)))
-  (not (test-disjoint* '(list-assuming-immutable? null?)))
-  (not (test-disjoint* '(list-assuming-immutable? pair?)))
-  (not (test-disjoint* '(list-assuming-immutable? list?)))
 )
 
 ; use a gensym to make expansions equivalent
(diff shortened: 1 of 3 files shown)


mflatt:
While `(list-assuming-immutable? x)` should imply `(list-assuming-immutable? x)` for the same `x`, I'm not sure it should apply `(list? x)`. That is, I view the "assuming" part of `list-assuming-immutable?` to be about calls to `list-assuming-immutable?` and nothing else. If I understand this commit, though, the idea is that `list-assuming-immutable?` implies a more general claim that `x` will not be mutated.

For Racket programs, I would expect that `list?` never shows up, anyway. Does `list-assuming-immutable?` need to imply a more general to make some other optimizations work? Or have I misunderstood the direction?

gus-massa:
The optimization for

```
(lambda (x) (when (list-assuming-immutable? x) (list? x))))
```

has been working for a long time. The new part are the optimization for 

```
(lambda (x) (unless (list-assuming-immutable? x) (list? x))))
```

and to remove `(list->vector x))` in

```
(lambda (x) (when (list-assuming-immutable? x) (list->vector x) 7)))
```

My idea is that once `(list-assuming-immutable? x)` is call, the "listicity" of `x` will not change. I'm not assuming that the content of the list nor the length are constant.

mflatt:
Ok, I guess I missed the point where `list-assuming-immutable?` started implying `list?` (or I somehow thought about it differently at the time), and I can imagine that it was a long time ago.

But that wasn't the intent of `list-assuming-immutable?`. The `list-assuming-immutable?` wasn't meant to declare an (unchecked) promise on listness — only that the implementation of `list-assuming-immutable?` itself could cache a result.

Maybe we could refine the meaning of `list-assuming-immutable?`, but I'm having trouble deciding whether that makes sense:

* On the one hand, a `list?` by itself turns out to be useless as a safety guard, since pairs in the one-time list can change at any time, so probably an unchecked assumption does not lead to crashes in safe mode. 

* On the other hand, if `list-assuming-immutable?` has a dynamic effect on `list?`, then to reason completely about the meaning of `list?` checks (on lists that you completely control, say), then every function needs to say whether it uses `list-assuming-immutable?` on input and output values. Otherwise, a programmer might accidentally call a function that uses `list-assuming-immutable?`, then mutate a `cdr`, and then expect `list?` to produce `#f`. That seems like an unlikely combination in practice, and maybe from the perspective of using Chez Scheme for Racket, we can just treat all functions as potentially using `list-assuming-immutable?`. But from the perspective of Chez Scheme as a standalone language, it seems like an awkward semantics choice.

Are the optimizations enabled by a different definition of `list-assuming-immutable?` worthwhile? Or would it be worthwhile to add functions along the lines of `list-assuming-immutable->vector` and specify them in a way that makes dropping them ok if listness changes (e.g, a truncated vector instead of an error).

gus-massa:
Calling two functions that use `list-assuming-immutable?` and mutating the list in the middle may still be dangerous, but it need a combination of some bad luck, too much inlining and a very smart optimizer. So the functions should proclaim if they use `list-assuming-immutable?`. Anyway, I really doubt it's possible to get a problem with something like this in a realistic scenario.

My long term objective is to remove the `normalptr` slot in `or-pred`, and move all the types there to bitfields. I don't expect too much improvements caused by `list-assuming-immutable?` => `list?`, or the other reductions, so I think it's fine to remove them.

I expect some huge wins in realistic scenarios with `memq` but it's too specialized, so I left it outside this PR.

Perhaps `list-asuming-immutalisticity` can `cons-asuming-immutalisticity` may be useful, because many list are created just to be an argument to a function, but it's a different idea., perhaps add them after cptypes has more support for the data flow in loops. 

Back to the current PR..., I'll make a new version (hopefully today or tomorrow.).  Something like the 1st commit without the special case for `list?`, most of the tests of the 2nd, and an adaptation of the 3rd, and perhaps uncomment some test in optimize.rktd.

mflatt:
If I understand, this commit improves things without any new pieces that assume or depend on `list-assuming-immutable?` ⇒ `list?`. Sounds good to me, and the changes look good as far as I can see.

PR #4616, opened 2023-03-26
Clarify that `predicate/c` is obsolete.

jackfirth:
This change strengthens the wording a bit. See prior conversation in 09f74002eb346fd22996fcc70f5d0c2bc011bc0d and https://github.com/jackfirth/resyntax/issues/191.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1665,18 +1665,12 @@ be blamed using the above contract:
 
 @defthing[predicate/c contract?]{
 
- Use this contract to indicate that some function is a
- predicate.
-
-  This contract also includes an optimization so that functions returning
-  @racket[#t] from @racket[struct-predicate-procedure?] are just returned directly, without
-  being wrapped. This contract is used by @racket[provide/contract]'s
-  @racket[struct] sub-form so that struct predicates end up not being wrapped.
-
- This contract is semantically equivalent to
- @racket[(-> any/c boolean?)] and
- @racket[(-> any/c boolean?)] also has the same optimization.
-
+ Equivalent to @racket[(-> any/c boolean?)]. This contract includes an optimization so that functions returning
+ @racket[#t] from @racket[struct-predicate-procedure?] are just returned directly, without
+ being wrapped. This contract is used by @racket[provide/contract]'s
+ @racket[struct] sub-form so that struct predicates end up not being wrapped. In previous versions of Racket,
+ this contract was necessary for that optimization. Now however, @racket[(-> any/c boolean?)] performs the same
+ optimization. This contract is still provided for backwards compatibility.
 }
 
 @defthing[the-unsupplied-arg unsupplied-arg?]{
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1668,7 +1668,7 @@ be blamed using the above contract:
  was necessary as it included an additional optimization that was not
  included in @racket[->]. Now however, @racket[->] performs the same
  optimization, so the contract should no longer be used. The contract
- is still provided for backwards compatibility.
+ is still provided for backward compatibility.
 }
 
 @defthing[the-unsupplied-arg unsupplied-arg?]{
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1664,13 +1664,11 @@ be blamed using the above contract:
 ]}
 
 @defthing[predicate/c contract?]{
-
- Equivalent to @racket[(-> any/c boolean?)]. This contract includes an optimization so that functions returning
- @racket[#t] from @racket[struct-predicate-procedure?] are just returned directly, without
- being wrapped. This contract is used by @racket[provide/contract]'s
- @racket[struct] sub-form so that struct predicates end up not being wrapped. In previous versions of Racket,
- this contract was necessary for that optimization. Now however, @racket[(-> any/c boolean?)] performs the same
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/contracts.scrbl
```diff
@@ -1665,18 +1665,12 @@ be blamed using the above contract:
...
-  This contract also includes an optimization so that functions returning
-  @racket[#t] from @racket[struct-predicate-procedure?] are just returned directly, without
-  being wrapped. This contract is used by @racket[provide/contract]'s
-  @racket[struct] sub-form so that struct predicates end up not being wrapped.
-
- This contract is semantically equivalent to
- @racket[(-> any/c boolean?)] and
- @racket[(-> any/c boolean?)] also has the same optimization.
-
+ Equivalent to @racket[(-> any/c boolean?)]. This contract includes an optimization so that functions returning
```
sorawee, line 1668: Personally, I would simply say:

> Equivalent to @racket[(-> any/c boolean?)]. Previously, this contract was necessary as it includes an additional optimization that was not included in `->`. Now however, @racket[->] performs the same
 optimization, so the contract should no longer be used. The contract is still provided for backwards compatibility.

I'll leave the decision whether to use `@deprecated` to you.

jackfirth: Sounds good to me. I'll leave the `@deprecated` bit out since that seems unnecessarily strong.

sorawee: Is it "backwards compatibility" or "backward compatibility"?

jackfirth: It's "backward" but I always get it wrong.

LiberalArtist:
Is the optimization also performed by `(-> any/c any/c)`? It might be worth giving a caveat that you often want to use that rather than `(-> any/c boolean?)` to avoid needlessly rejecting functions that return a non-`#false` value other than `#true`.

jackfirth:
@LiberalArtist I think that's a separate conversation. `predicate/c` is equivalent to `(-> any/c boolean?)`, so I think its docs should discuss which of those two equivalent options is preferable.

rfindler:
The specific optimization is that `(-> any/c boolean?)` checks to see if the function is a struct predicate function (the annotations that say that some functions are struct predicate functions are trusted because the runtime system puts them in) and then avoids creating the chaperone wrapper that checks that the result is just one value and the value is a boolean. The contract `(-> any/c any)` has an optimization -- it is actually a flat contract that never creates a wrapper. In principle, the `(-> any/c boolean?)` optimization could be added to `(-> any/c any/c)` but the original optimization was added because it mattered for benchmarks for TR. Do you know of a program for which `(-> any/c any/c)` on struct predicates happens in a measurable way?

LiberalArtist:
> Do you know of a program for which `(-> any/c any/c)` on struct predicates happens in a measurable way?

No, I agree with @jackfirth that there were two separate issues in my post, because I misread the diff: with the revised text, the details of the optimization don't matter.

> @LiberalArtist I think that's a separate conversation. `predicate/c` is equivalent to `(-> any/c boolean?)`, so I think its docs should discuss which of those two equivalent options is preferable.

Here is the scenario I was imagining:

----
I am writing a contract for `filter`, and I'm trying to figure out what to put in the hole in:
```racket
(parametric->/c [α] (-> 🕳 (listof α) (listof α)))
```
I look at the table of contents, and `predicate/c` sounds just right: I click on it and read more. The documentation now tells me that `predicate/c` is equivalent to `(-> any/c boolean?)`, so I think I have the answer to my question&mdash;but actually, `(-> any/c any/c)` would have been a better choice.

(I wouldn't have wanted `(-> any/c any)` because I don't want to be blamed for result arity errors.)

----

I think the misunderstanding in that scenario is fairly common, and I think people reading the documentation for `predicate/c` *ought* to be asking themselves whether they want `(-> any/c boolean?)` or `(-> any/c any/c)`.

What about something like this?
```racket
@defthing[predicate/c contract?]{
 Equivalent to @racket[(-> any/c boolean?)].

 Previously, this contract
 was necessary, as it included an additional optimization that was not
 included in @racket[->]. Now, however, @racket[->] performs the same
 optimization, so @racket[predicate/c] should no longer be used. It
 is still provided for backward compatibility.

 Often, functions returning non-false values other than @racket[#t]
 should also be accepted as predicates: to do so, use
 @racket[(-> any/c any/c)] rather than @racket[(-> any/c boolean?)]
 or @racket[predicate/c].
}
```

jackfirth:
I'm actually not convinced that's always sound advice to follow. Truthiness makes sense to me when it's combined with control flow, but outside that domain I think it makes it easier for bugs to slip through. I've never liked the `(filter values xs)` pattern, for example.

LiberalArtist:
What about `(filter (curryr member xs) ys)`?

jackfirth:
I think the real problem in that example is that `member` returns data instead of just being a boolean-returning function called `member?`. If you use `set-member?` (taking advantage of the (possibly dubious) generic set implementation for lists) the problem goes away.

LiberalArtist:
I don't entirely disagree with you. Sometimes it is important to enforce `boolean?`, especially to avoid needlessly keeping data alive.

But, as with `member`, it's a common idiom to program with a sort of `Maybe` type where `#f` represents `Nothing` and the `Just` value is unwrapped. Like most things, it has both pros and cons, but it's quite common in currently-existing Racket code  (and in Scheme more broadly). Likewise, it's rare to encounter a library function that requires the caller to write `(and 🕳 #t)` around arguments. (Even though I personally *like* `#t`, enough to have put [`any->boolean`](https://docs.racket-lang.org/adjutor/Stable.html#(def._((lib._adjutor%2Fmain..rkt)._any-~3eboolean))) in a package.)

jackfirth:
Oh interesting. Both those functions are _documented_ as requiring booleans, but they actually accept anything.

PR #4618, opened 2023-03-27
doc: `seconds->date` uses seconds since the epoch

LiberalArtist:
It does not use a platform-specific time.

Use `@tech{the epoch}` for `current-inexact-milliseconds`, too.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/time.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/time.scrbl
@@ -46,7 +46,7 @@ are not comparable.
          date*?]{
 
 Takes @racket[secs-n], a time in seconds since @tech{the epoch} (like the value of
-@racket[(current-seconds)], @racket[(file-or-directory-modify-seconds)],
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/time.scrbl
```diff
@@ -45,9 +45,9 @@ are not comparable.
                         [local-time? any/c #t])
          date*?]{
 
-Takes @racket[secs-n], a platform-specific time in seconds returned by
-@racket[current-seconds], @racket[file-or-directory-modify-seconds],
-or 1/1000th of @racket[current-inexact-milliseconds], and returns an
+Takes @racket[secs-n], a time in seconds since @tech{the epoch} (like the value of
+@racket[(current-seconds)], @racket[(file-or-directory-modify-seconds)],
```
LiberalArtist, line 49: Oh, `(file-or-directory-modify-seconds)` doesn't work &hellip;

PR #4621, opened 2023-03-31
net-lib: add net/server module with start-server and run-server

Bogdanp:
Opening as draft in case folks have comments.  If `net-lib` seems like a bad place for this, I can just make a new package, but it would be nice to have something like this in the main distribution.

This is a port of `run-server` from `mzlib/thread`, with some improvements carried over from changes I had made to that implementation when I ported it to `web-server-lib`. Inspired by [a recent discussion](https://discord.com/channels/571040468092321801/907499351432106024/1089222735374319706) on Discord.

This implementation should be compatible with the one in `mzlib/thread` so a follow-up change could be to rebase that implementation on top of this one, if merged, since `compatibility-lib` already depends on `net-lib`. It should also be possible to make `web-server-lib` use this implementation.

cc @LiberalArtist @samth


pkgs/net-doc/net/scribblings/server.scrbl
```diff
@@ -0,0 +1,176 @@
...
+  synchronization, will cause the connection's custodian to be shut
+  down and, consequently, the handling thread to be killed if it is
+  still running by that time.
+
+  The server keeps track of the number of active connections and
+  pauses accepting new connections once that number reaches
+  @racket[max-concurrent], resuming once the number goes down again.
+
+  The @racket[reuse?] and @racket[max-allow-wait] arguments have the
+  same meaning as in @racket[tcp-listen].  The @racket[listen],
```
LiberalArtist, line 76: I'm of two minds about these: handling them as arguments to `start-server` is obviously much more ergonomic for TCP, but I wonder how well they generalize to other kinds of connections.

LiberalArtist: Actually, as I look more closely, it seems `port`, `max-allow-wait`, `reuse?`, and `host` are all used only by passing them to `listen` exactly once, before starting threads or looping or any of the other complexity:
```racket
  (define listener
    (listen port max-allow-wait reuse? host))
```
What would you think of dropping all of them and having `start-server` take the `listener` as an argument, instead?

pkgs/net-doc/net/scribblings/server.scrbl
```diff
@@ -0,0 +1,176 @@
...
+      listener)
+    (define port-ch
+      (make-channel))
+    (define stop
+      (start-server
+        #:listen-proc (make-listen-proc port-ch)
+        "127.0.0.1" 0 void))
+    (channel-get port-ch)
+    (stop)
+  ]
```
LiberalArtist, line 111: Having `run-server` take `listener` as an argument would simplify this considerably.

pkgs/net-lib/net/server.rkt
```diff
@@ -0,0 +1,100 @@
...
+                          (add1 in-progress)))))))))))
+         (lambda ()
+           (close listener))))))
+  (lambda ()
+    (break-thread server-thd)
+    (thread-wait server-thd)))
+
+(define run-server
+  (make-keyword-procedure
+   (lambda (kws kw-args . args)
```
LiberalArtist, line 95: It would be nice to get `object-name`, `procedure-arity`, and `procedure-keywords` right here.

LiberalArtist (commented):
So glad you're working on this! Here are a few (well, more than a few &hellip;) initial thoughts.

pkgs/net-doc/net/scribblings/server.scrbl
samth: This was also my major reservation here. The make-listen-proc pattern here seems like it could usefully be a part of the abstraction.

pkgs/net-doc/net/scribblings/server.scrbl
Bogdanp: Thanks for all the feedback so far!

> What would you think of dropping all of them and having start-server take the listener as an argument, instead? 

I'm reluctant to make that change, because it feels wrong for the listener to be created externally to `start-server` and  for `start-server` to then take ownership of the listener and be in charge of closing it. My preference would be to reduce the set of arguments that `start-server` takes and rely on the caller to provide a customized `listen-proc`.

pkgs/net-doc/net/scribblings/server.scrbl
Bogdanp: My argument here would be that this is a special case. Most users of these procedures are likely not going to be using ephemeral ports and I wouldn't have used them here if I could know for sure that the documentation would build cleanly with specific ports used in the examples. Probably, I should use a log-based evaluator instead for the simple examples and mark this example as a special case.

pkgs/net-lib/net/server.rkt
Bogdanp: I can fix the procedure name and arity right here, but is there a way to fix `procedure-keywords` without spelling out all the keywords?

pkgs/net-doc/net/scribblings/server.scrbl
samth: That sounds like a sensible plan; I think we want the examples to be what people would actually do.

pkgs/net-lib/net/server.rkt
samth: You can use procedure-reduce-keyword-arity.

pkgs/net-doc/net/scribblings/server.scrbl
```diff
@@ -0,0 +1,176 @@
...
+  still running by that time.
+
+  The server keeps track of the number of active connections and
+  pauses accepting new connections once that number reaches
+  @racket[max-concurrent], resuming once the number goes down again.
+
+  The @racket[reuse?] and @racket[max-allow-wait] arguments have the
+  same meaning as in @racket[tcp-listen].  The @racket[listen],
+  @racket[accept] and @racket[close] arguments may be used to run the
+  server on an alternate protocol (such as SSL).  The
```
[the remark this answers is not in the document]
samth: Would it be worth it to build SSL handling in directly to this abstraction? Or is that adding too much complexity here?

pkgs/net-lib/net/server.rkt
Bogdanp: Thanks!

pkgs/net-doc/net/scribblings/server.scrbl
Bogdanp: OK, I've made a separate section for examples and added TCP, TCP+SSL, Unix socket and port examples and dropped the ephemeral port example entirely.

pkgs/net-doc/net/scribblings/server.scrbl
Bogdanp: After working on the new set of examples, I think it's probably worth keeping the keywords as-is. Partly because TCP is likely to be the common use case, but also because of the implementation of `tcp-listen`: the `host` argument is the final argument, so we'd have to provide our own defaults to `listen-proc` anyway (which is what `mzlib`'s `run-server` does, but seems worse). Having the `host` argument be a required arg to `start-server` seems better than relying on `tcp-listen`'s default of listening on all interfaces, too.

pkgs/net-doc/net/scribblings/server.scrbl
```diff
@@ -0,0 +1,176 @@
...
+@defmodule[net/server]{The @racketmodname[net/server] library provides
+support for running general-purpose networked servers.}
+
+@(define ev (make-base-eval))
+
+@defproc[(start-server [host string?]
+                       [port (integer-in 0 65535)]
+                       [handle (-> input-port? output-port? any)]
+                       [#:reuse? reuse? any/c #t]
+                       [#:max-allow-wait max-allow-wait exact-nonnegative-integer? 511]
```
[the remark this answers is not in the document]
Bogdanp: Yes, I remember being baffled a year or two ago when I was profiling something and the `backlog` argument wasn't working as I expected. Finally, I opened the manual for `listen` and saw:

> BUGS
>     The backlog is currently limited (silently) to 128.

I think I'll just change it so that it matches the default value of `tcp-listen`, namely 4.

pkgs/net-doc/net/scribblings/server.scrbl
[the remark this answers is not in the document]
Bogdanp: I've added an example of using `start-server` with `ports->ssl-ports` and dropped the "(such as SSL)" bit from the documentation. I think in light of how simple `ports->ssl-ports` makes this, it's not worth adding explicit support here.

pkgs/net-lib/net/server.rkt
```diff
@@ -91,12 +91,17 @@
...
-    (lambda (kws kw-args host port handle)
-      (parameterize-break #f
-        (define stop (keyword-apply start-server kws kw-args host port handle null))
-        (with-handlers ([exn:break? void])
-          (sync/enable-break never-evt))
-        (stop))))
-   'run-server))
+  (let-values ([(required-kws optional-kws) (procedure-keywords start-server)])
+    (procedure-rename
+     (procedure-reduce-keyword-arity
```
LiberalArtist, line 96: Since you're using `procedure-reduce-keyword-arity` anyway, you don't need `procedure-rename`: you can just pass the name to `procedure-reduce-keyword-arity`. Also, it might be slightly more efficient to use `procedure-reduce-keyword-arity-mask` and `procedure-arity-mask`.

pkgs/net-lib/net/server.rkt
```diff
@@ -0,0 +1,106 @@
...
+(define (server-proc/c res/c)
+  (->* ((or/c #f string?)
+        listen-port-number?
+        (-> input-port? output-port? any))
+       (#:reuse? any/c
+        #:max-allow-wait exact-nonnegative-integer?
+        #:max-concurrent (or/c +inf.0 exact-positive-integer?)
+        #:listen-proc (-> listen-port-number? exact-nonnegative-integer? any/c (or/c #f string?) evt?)
+        #:accept-proc (-> any/c (values input-port? output-port?))
+        #:close-proc (-> any/c void?)
```
LiberalArtist, line 20: I think we probably don't want to enforce that `close-proc` specifically return `void?` rather than `any/c`. Actually, it seems like `(-> any/c any)` would work fine.

pkgs/net-doc/net/scribblings/server.scrbl
```diff
@@ -0,0 +1,274 @@
...
+@defmodule[net/server]{The @racketmodname[net/server] library provides
+support for running general-purpose networked servers.}
+
+@defproc[(start-server [host (or/c #f string?)]
+                       [port listen-port-number?]
+                       [handle (-> input-port? output-port? any)]
+                       [#:reuse? reuse? any/c #t]
+                       [#:max-allow-wait max-allow-wait exact-nonnegative-integer? 4]
+                       [#:max-concurrent max-concurrent
+                                         (or/c +inf.0 exact-positive-integer?)
```
LiberalArtist, line 23: I wonder if we should accept `(or/c +inf.0 natural-number/c)`. It would admittedly be a bit silly to start a server that would never accept any connections, but maybe it would be useful for testing or something.

pkgs/net-lib/net/server.rkt
```diff
@@ -0,0 +1,106 @@
...
+       res/c))
+
+(define (start-server host port handle
+                      #:reuse? [reuse? #t]
+                      #:max-allow-wait [max-allow-wait 4]
+                      #:max-concurrent [max-concurrent +inf.0]
+                      #:listen-proc [listen tcp-listen]
+                      #:accept-proc [accept tcp-accept]
+                      #:close-proc [close tcp-close]
+                      #:timeout-evt-proc [make-timeout-evt (λ (_thd _in _out _break-sent?) never-evt)])
```
LiberalArtist, line 31: I think I'd rather the keywords be `#:listen`, `#:accept`, and `#:close`. I'd also prefer `#:make-timeout-evt` to `#:timeout-evt-proc`, but maybe there's something even better.

pkgs/net-lib/net/server.rkt
```diff
@@ -0,0 +1,106 @@
...
+        #:close-proc (-> any/c void?)
+        #:timeout-evt-proc (-> thread? input-port? output-port? boolean? evt?))
+       res/c))
+
+(define (start-server host port handle
+                      #:reuse? [reuse? #t]
+                      #:max-allow-wait [max-allow-wait 4]
+                      #:max-concurrent [max-concurrent +inf.0]
+                      #:listen-proc [listen tcp-listen]
+                      #:accept-proc [accept tcp-accept]
```
LiberalArtist, line 29: Do we need an argument like `tcp-accept/enable-break`? I just saw that the old `run-server` had both: I haven't refreshed my memory as to why yet.

Bogdanp: Not with this implementation, since the job of allowing breaks within the listening thread is handled by the `sync/enable-break` call instead.

LiberalArtist:
Moving down here to reply to a few threads at once.

Re https://github.com/racket/racket/pull/4621#discussion_r1155136495:
> After working on the new set of examples, I think it's probably worth keeping the keywords as-is. Partly because TCP is likely to be the common use case, but also because of the implementation of `tcp-listen`: the `host` argument is the final argument, so we'd have to provide our own defaults to `listen-proc` anyway (which is what `mzlib`'s `run-server` does, but seems worse). Having the `host` argument be a required arg to `start-server` seems better than relying on `tcp-listen`'s default of listening on all interfaces, too.

From my perspective, I think the essence of this abstraction is in managing the subtleties of threads (listening thread, connection-handling threads, and supervisor threads), custodians, exceptions, parameterizations, breaks, and graceful (or forcible) shutdown: many of the ideas presented in [*More: Systems Programming with Racket*](https://docs.racket-lang.org/more/index.html). (I think it would be worth adding a link to `net/server` from its conclusion!)

We definitely want to provide an excellent experience with TCP, since I indeed expect that will be the most common use case. Even in the relatively near term, though, I could imagine QUIC supplanting many uses for TCP+TLS if we had it implemented in Racket, and similarly I expect at least some of TCP's current popularity in Racket is because we don't have much support for named pipes (without resorting to extra-linguistic mechanisms) and [`make-unix-socket-tcp@`](https://docs.racket-lang.org/unix-socket/index.html#%28def._%28%28lib._racket%2Funix-socket-tcp-unit..rkt%29._make-unix-socket-tcp~40%29%29) is easy to overlook. (This reminds me that I've meant to look at https://github.com/racket/unix-socket/issues/6 some day.)

I think we agree that one of the reasons the old [`run-server`](https://docs.racket-lang.org/mzlib/mzlib_thread.html#%28def._%28%28lib._mzlib%2Fthread..rkt%29._run-server%29%29) deserved to be deprecated is that it baked a few two many assumptions in at the wrong level (I particularly remember the kill timeout being twice the break timeout), such that the web server's use of it involved passing dummy values for certain arguments and working around some of its behavior. I think a goal for `net/server` should be not to require such hijinks for protocols that might be considerably different than TCP.

Re https://github.com/racket/racket/pull/4621#discussion_r1155077820:
> 
> > What would you think of dropping all of them and having `start-server` take the listener as an argument, instead?
> 
> I'm reluctant to make that change, because it feels wrong for the listener to be created externally to `start-server` and for `start-server` to then take ownership of the listener and be in charge of closing it. My preference would be to reduce the set of arguments that `start-server` takes and rely on the caller to provide a customized `listen-proc`.

I'd see this as similar to [`reencode-input-port`](https://docs.racket-lang.org/reference/port-lib.html#%28def._%28%28lib._racket%2Fport..rkt%29._reencode-input-port%29%29), [`port->string`](https://docs.racket-lang.org/reference/port-lib.html#%28def._%28%28lib._racket%2Fport..rkt%29._port-~3estring%29%29), and many other functions from `racket/port` which which accept a `close?` argument. In particular, passing `#:close void` to `start-server` would work like `#:close? #f` in those cases.

I do agree that supporting some arguments tied to `tcp-listen` while also relying on some defaults seems like a particularly unpleasant option, but I'm increasingly interested in a signature like this:
```
(define (start-server listener handle
                      #:max-concurrent [max-concurrent +inf.0]
                      #:accept [accept tcp-accept]
                      #:close [close tcp-close]
                      #:make-timeout-evt
                      [make-timeout-evt (λ (_thd _in _out _break-sent?)
                                          never-evt)])
  ...)
```

It also has the benefit of clearly communicating which arguments control the functionality of `start-server` and distinguishing them from the general options for configuring a TCP listener.

Re https://github.com/racket/racket/pull/4621#discussion_r1155078329:
> My argument here would be that this is a special case. Most users of these procedures are likely not going to be using ephemeral ports and I wouldn't have used them here if I could know for sure that the documentation would build cleanly with specific ports used in the examples. Probably, I should use a log-based evaluator instead for the simple examples and mark this example as a special case.

Actually, I thought this was a great example! It took me a long time to figure out how to learn what ephemeral port a listener had been assigned (I somehow missed `#:confirmation-channel`, or actually I guess [`serve/launch/wait`](https://docs.racket-lang.org/web-server-internal/web-server.html#%28def._%28%28lib._web-server%2Fservlet-dispatch..rkt%29._serve%2Flaunch%2Fwait%29%29) doesn't support that option), and I wrote several programs with hard-coded ports when I really would have preferred ephemeral ones as a result.

Moreover, it prompted me to think more (and I'm continuing to do so) about what kinds of additional communication protocols other than TCP might want. Maybe QUIC Connection IDs? Or maybe not.

pkgs/net-lib/net/server.rkt
```diff
@@ -0,0 +1,106 @@
...
+         (lambda ()
+           (with-handlers ([exn:break? void])
+             (let loop ([in-progress 0])
+               (loop
+                (with-handlers ([exn:fail:network?
+                                 (λ (e)
+                                   (begin0 in-progress
+                                     ((error-display-handler)
+                                      (format "Connection error: ~a" (exn-message e))
+                                      e)))])
```
LiberalArtist, line 52: This seems like a place where it would be nice to allow interposition, though I'm not sure exactly how.

Bogdanp:
I don't think I have any new thoughts regarding the external listener approach. I think passing in a listener breaks encapsulation in a way that makes me uncomfortable, but I don't have any better arguments against it. That approach does strictly limit what `start-server` can do (eg. it can't wrap the listener in its own custodian should we want to add a mode later on that guarantees it frees all resources on stop), but that's not a very strong argument. I think ergonomics of passing in a listener are also worse for the general case, but I've already said that.

I think a QUIC listener is likely to also take similar arguments, though maybe not in the same order, so probably a use would look like:

```racket
(define stop
  (start-server 
   #:listen-proc (lambda (port backlog reuse? host)
                   (quic-listen ssl-ctx host port backlog))
   #:accept-proc quic-accept
   #:close-proc quic-close
   host port handle))
```

While I agree that it's not ideal to have to shift arguments around and to outright ignore one of them, I still feel that's better, and probably less error prone, than:

```racket
(define listener (quic-listen ssl-ctx host port backlog))
(define stop
  (start-server 
   #:accept-proc quic-accept
   #:close-proc quic-close
   listener handle))
```

Because there's more of a disconnect between things in the 2nd example.  I'm more likely to only pass in a listener and forget to pass in accept and close procs in the 2nd case. Of course, I could instead write

```racket
(define stop
  (start-server 
   #:accept-proc quic-accept
   #:close-proc quic-close
   (quic-listen ssl-ctx host port backlog) 
   handle))
```

and maybe that makes my last point moot, but then again that might also mean I could write

```racket
(define stop
  (start-server (tcp-listen port) handle))
```

and not realize that my server is now binding to all interfaces. What's worse, now I have to write

```racket
(define stop
  (start-server (tcp-listen 8000 4 #f "127.0.0.1") handle))
```

just to get it to listen only on localhost. Versus:

```racket
(define stop
  (start-server "127.0.0.1" 8000 handle))
```

pkgs/net-lib/net/server.rkt
Bogdanp: I prefer to enforce since typically these kinds of destructive procedures return void. We can relax it later if necessary, but I think it's more likely to catch issues as is.

LiberalArtist:
> because of the implementation of `tcp-listen`: the `host` argument is the final argument, so we'd have to provide our own defaults to `listen-proc` anyway (which is what `mzlib`'s `run-server` does, but seems worse). Having the `host` argument be a required arg to `start-server` seems better than relying on `tcp-listen`'s default of listening on all interfaces, too.


> but then again that might also mean I could write
> 
> ```racket
> (define stop
>   (start-server (tcp-listen port) handle))
> ```
> 
> and not realize that my server is now binding to all interfaces. What's worse, now I have to write
> 
> ```racket
> (define stop
>   (start-server (tcp-listen 8000 4 #f "127.0.0.1") handle))
> ```
> 
> just to get it to listen only on localhost. Versus:
> 
> ```racket
> (define stop
>   (start-server "127.0.0.1" 8000 handle))
> ```

(Very belated reply &hellip;)

To me, these seem like good arguments for a (keyword?) alternative to `tcp-listen`, rather than entangling `tcp-listen` with `run-server`.

Bogdanp:
Thanks @joergen7 for taking over and implementing @LiberalArtist's suggested API changes. @LiberalArtist, mind taking another look? The new tests still need to be adjusted for signature change, but I think this should be good to go afterwards.

LiberalArtist:
Thanks @joergen7 and @Bogdanp for staying on top of this!

Overall, I think this is great!

I've pushed a few commits to https://github.com/LiberalArtist/racket/tree/pr-4621-run-server with my most concrete suggestions (as separate commits so you can pick only ones you like):

1. 3ea3ce44cd930c6eb81c3faa13d9e85ecccfb82c bikesheds the keyword argument names to `#:accept`, `#:close`, and `#:make-timeout-evt`.
2. 32d4ce052531ba7e68eef75f5c0c1a53f7388cbd relaxes the `#:close` result contract to `any` instead of requiring `void?`.
3. 8217575ff2563b6ca906619cc17bc16a8a7a80ae has a bunch of doc edits, particularly to the way of documenting `listener`/`listen-evt`: there's more explanation in that commit message.
4. 4f36b41293e2736d5eae9bb60b68e41bb00494e6 changes the connection supervisor release resources *before* reporting to the server thread that it has finished, instead of after.

There are a few other things I think we might want to consider before merging:

1. The implementation goes to a lot of effort to protect itself from `handle` in terms of parameterizations and separate custodians, but `make-timeout-evt` is also a callback, and currently it runs in the dynamic environment of the connection supervisor thread with no isolation. Should we do more here? Or maybe this is ok?
2. I don't love the names `run-server` and `start-server`: in particular, I don't find it obvious which one does what. On the other hand, we definitely *don't* need to get bogged down in bikeshedding names. My best thought so far, based on names from the web server, is `serve` and `serve/wait`. Do those seem better enough to be worth changing? Does anyone have better ideas?
3. The equivalent code in [the web server](https://github.com/racket/web-server/blob/87a1e58a9b92c621f7d96c76de1da39587f04140/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt#L52-L55) selects between `sync`+`tcp-accept` vs. `sync/enable-break`+`tcp-accept/enable-break`, and [`mzlib/thread`](https://github.com/racket/compatibility/blob/a393782feeb140dceda4bdcf3747659d9cfa82a6/compatibility-lib/mzlib/thread.rkt#L94-L101), which does not do any kind of `sync` on a listener, selects between `tcp-accept` and `tcp-accept/enable-break`. In this PR, we unconditionally use `sync/enable-break`+`tcp-accept` (not `/enable-break`). Is this correct? I definitely don't have the subtleties of breaks paged in at the moment. Assuming it is correct, it might be worth a comment for posterity.

On the other hand, there are some things I thought about but **don't** think need any action, which I'll mention for the record:

- [`mzlib/thread`](https://github.com/racket/compatibility/blob/a393782feeb140dceda4bdcf3747659d9cfa82a6/compatibility-lib/mzlib/thread.rkt#L114) has a `(current-custodian c)` that this PR doesn't. I'm pretty convinced it's superfluous. [The web server](https://github.com/racket/web-server/blob/87a1e58a9b92c621f7d96c76de1da39587f04140/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt#L89-L94) doesn't have it, either.
- I still think allowing control over handling `exn:fail:network?` would be nice (see https://github.com/racket/racket/pull/4621#discussion_r1155326780), but that could definitely be added later.
- I wondered if this PR was enough to support tracking when all outstanding connections have finished. I think it is, since `make-timeout-evt` gives you a way to get each connection handler thread.
- Similarly, I wondered about the fact that, while this PR lets you stop the server gracefully, I don't think it gives you a way to notice if the server has somehow stopped abnormally. But it seems like that could also be added later, and it looks like the web server doesn't provide a way to do that, either, so maybe it's not much in demand.

Bogdanp:
Thanks @LiberalArtist!

> 1. [3ea3ce4](https://github.com/racket/racket/commit/3ea3ce44cd930c6eb81c3faa13d9e85ecccfb82c) bikesheds the keyword argument names to `#:accept`, `#:close`, and `#:make-timeout-evt`.

I think I still prefer to keep the old arg names. 

> 2. [32d4ce0](https://github.com/racket/racket/commit/32d4ce052531ba7e68eef75f5c0c1a53f7388cbd) relaxes the `#:close` result contract to `any` instead of requiring `void?`.

Using `void?` as a contract in side-effectul procedures has uncovered bugs for me in the past so I prefer to keep the return type as is for now.

> 3. [8217575](https://github.com/racket/racket/commit/8217575ff2563b6ca906619cc17bc16a8a7a80ae) has a bunch of doc edits, particularly to the way of documenting `listener`/`listen-evt`: there's more explanation in that commit message.

Merged.

> 4. [4f36b41](https://github.com/racket/racket/commit/4f36b41293e2736d5eae9bb60b68e41bb00494e6) changes the connection supervisor release resources _before_ reporting to the server thread that it has finished, instead of after.

Merged.

> 1. The implementation goes to a lot of effort to protect itself from `handle` in terms of parameterizations and separate custodians, but `make-timeout-evt` is also a callback, and currently it runs in the dynamic environment of the connection supervisor thread with no isolation. Should we do more here? Or maybe this is ok?

I thought about that as well when I first implemented this change. The way I saw it, `handle` had more chances to be called by end-user code (a web application handler) as opposed to library code (the web-server library), so there are more reasons to protect it. Something like the `make-timeout-evt` would instead be used more at the library level, where further protections can be added if necessary (I realize you could make the same argument for `handle`). That said, a side-effect of `4f36b41` is that the call to `make-timeout-proc` is now owned by the supervisor custodian, so at least its resources will be cleaned up. But, it can still kill the supervising thread if it so chooses. I think this is something we can adjust later if it turns out to be a problem.

> 2. I don't love the names `run-server` and `start-server`: in particular, I don't find it obvious which one does what. On the other hand, we definitely _don't_ need to get bogged down in bikeshedding names. My best thought so far, based on names from the web server, is `serve` and `serve/wait`. Do those seem better enough to be worth changing? Does anyone have better ideas?

Since this is a port with improvements of `mzlib`'s `run-server`, I think we should keep the name. I think `start-server` makes sense because it returns a `stop` procedure and `run-server` just runs forever (unless `exn:break`ed).

> In this PR, we unconditionally use `sync/enable-break`+`tcp-accept` (not `/enable-break`). Is this correct? I definitely don't have the subtleties of breaks paged in at the moment. Assuming it is correct, it might be worth a comment for posterity.

We use `sync/enable-break` to ensure that the server can be stopped before a new connection is ready to be accepted. We use `accept` without `enable-break` because the accept is guaranteed to not block within the event handler as long as the listener is not being used concurrently (this goes back to the earlier discussion about whether the server should create and own the listener or whether it should be passed in like it is now).

> On the other hand, there are some things I thought about but **don't** think need any action, which I'll mention for the record:
> 
> * [`mzlib/thread`](https://github.com/racket/compatibility/blob/a393782feeb140dceda4bdcf3747659d9cfa82a6/compatibility-lib/mzlib/thread.rkt#L114) has a `(current-custodian c)` that this PR doesn't. I'm pretty convinced it's superfluous. [The web server](https://github.com/racket/web-server/blob/87a1e58a9b92c621f7d96c76de1da39587f04140/web-server-lib/web-server/private/dispatch-server-with-connect-unit.rkt#L89-L94) doesn't have it, either.

I think that line is superfluous so I didn't include it when I made the web-server change.

> * I still think allowing control over handling `exn:fail:network?` would be nice (see [net-lib: add net/server module with start-server and run-server #4621 (comment)](https://github.com/racket/racket/pull/4621#discussion_r1155326780)), but that could definitely be added later.

Agreed that this can be added later if desired.

> * I wondered if this PR was enough to support tracking when all outstanding connections have finished. I think it is, since `make-timeout-evt` gives you a way to get each connection handler thread.

Yep, that's the reason `make-timeout-evt` takes the handler thread as an argument.

> * Similarly, I wondered about the fact that, while this PR lets you stop the server gracefully, I don't think it gives you a way to notice if the server has somehow stopped abnormally. But it seems like that could also be added later, and it looks like the web server doesn't provide a way to do that, either, so maybe it's not much in demand.

Barring OOM or bugs in Racket, I think the server should not stop abnormally.

I'm going to take another look at this in a few days, squash it and then it'll be ready for review.

LiberalArtist:
Thanks!

> > 1. [3ea3ce4](https://github.com/racket/racket/commit/3ea3ce44cd930c6eb81c3faa13d9e85ecccfb82c) bikesheds the keyword argument names to `#:accept`, `#:close`, and `#:make-timeout-evt`.
> 
> I think I still prefer to keep the old arg names.
> 

I couldn't think of existing keywords suffixed with `-proc` in the style of `#:accept-proc`. I'd forgotten about it, but apparently this was my opinion back in https://github.com/racket/racket/pull/4621#discussion_r1155301068, too. For a more objective metric:
```
philip@bastet:/tmp/racket-8.15.900$ rg '#:\S*-proc\s'
src/native-libs/build.rkt
388:                #:fixup-proc [fixup-proc #f]
498:             #:fixup-proc (and win?

share/pkgs/deinprogramm-signature/deinprogramm/signature/signature.rkt
69:                    #:<=?-proc (<=?-proc
72:                    #:=?-proc (=?-proc

share/pkgs/racket-doc/scribblings/reference/surrogate.scrbl
18:              ([use-wrapper-proc #:use-wrapper-proc (code:line)]
```

----

> > 2. [32d4ce0](https://github.com/racket/racket/commit/32d4ce052531ba7e68eef75f5c0c1a53f7388cbd) relaxes the `#:close` result contract to `any` instead of requiring `void?`.
> 
> Using `void?` as a contract in side-effectul procedures has uncovered bugs for me in the past so I prefer to keep the return type as is for now.
> 

I looked a bit for discussion I vaguely remembered about `any` instead of `void?` (on arguments: *promising* to return void has different ergonomics) and found https://github.com/racket/web-server/pull/17, in which @lexi-lambda wrote:

> This restriction is both annoying and actually confusing to people. It has come up on the mailing list before, and recently it came up again on the IRC channel. Furthermore, one of [Matt Might’s blog posts](http://matt.might.net/articles/low-level-web-in-racket/) actually depends on this permissive [due to a bug] behavior of `response`, which is especially awkward given that it will fail if you change `response` to `response/output`.
> 
> This simply weakens the contract to what I think it probably should be, `(output-port? . -> . any)`, removing the possibility for confusion and being generally more permissive without any loss of safety.

I did likewise in https://github.com/racket/web-server/pull/28.

----

> 
> > 1. The implementation goes to a lot of effort to protect itself from `handle` in terms of parameterizations and separate custodians, but `make-timeout-evt` is also a callback, and currently it runs in the dynamic environment of the connection supervisor thread with no isolation. Should we do more here? Or maybe this is ok?
> 
> I thought about that as well when I first implemented this change. The way I saw it, `handle` had more chances to be called by end-user code (a web application handler) as opposed to library code (the web-server library), so there are more reasons to protect it. Something like the `make-timeout-evt` would instead be used more at the library level, where further protections can be added if necessary (I realize you could make the same argument for `handle`). That said, a side-effect of `4f36b41` is that the call to `make-timeout-proc` is now owned by the supervisor custodian, so at least its resources will be cleaned up. But, it can still kill the supervising thread if it so chooses. I think this is something we can adjust later if it turns out to be a problem.

This seems reasonable to me. From a slightly different perspective, you could say that `start-server` isn't protecting itself from `handle`, but rather helping `handle` to protect itself from badly-behaving network clients that might otherwise consume excessive resources. In contrast, `make-timeout-proc` doesn't deal with all of the flakiness of networks, and its role is part of the cleaning-up of resources, not intensively consuming resources.

----

> 
> > 2. I don't love the names `run-server` and `start-server`: in particular, I don't find it obvious which one does what. On the other hand, we definitely _don't_ need to get bogged down in bikeshedding names. My best thought so far, based on names from the web server, is `serve` and `serve/wait`. Do those seem better enough to be worth changing? Does anyone have better ideas?
> 
> Since this is a port with improvements of `mzlib`'s `run-server`, I think we should keep the name. I think `start-server` makes sense because it returns a `stop` procedure and `run-server` just runs forever (unless `exn:break`ed).

Since `mzlib/thread` has been deprecated for so long, I don't think it's choice of name should get significant weight. On the other hand, the explanation that `start-server` is the one that returns a `stop` makes sense to me, and I don't feel that other names I've thought of are significantly better!

----

> 
> > In this PR, we unconditionally use `sync/enable-break`+`tcp-accept` (not `/enable-break`). Is this correct? &hellip;
> 

Apparently I had the same question and you had the same (correct) answer almost two years ago: https://github.com/racket/racket/pull/4621#discussion_r1155313619

This indirectly reminded me of https://github.com/racket/racket/pull/4621#discussion_r1154899135 and inspired me to push a few more docs commits:

1. https://github.com/LiberalArtist/racket/commit/476557ede2dc1044445f5239878fd90274e86848 general docs edit
2. https://github.com/LiberalArtist/racket/commit/e8cba88057e959711caf145ab7edf4f300b98426 rename "Echo Server over P̶o̶r̶t̶s̶Racket Pipes"
3. https://github.com/LiberalArtist/racket/commit/3ab83dd5dcb4a1cb243897b13abd253d4af1973c add timeout example

While working on https://github.com/LiberalArtist/racket/commit/3ab83dd5dcb4a1cb243897b13abd253d4af1973c, I found that I wanted to write the break timeout before the kill timeout, i.e. in the order they are sent, but the boolean if first `#f` and then `#t`. In other words,  you end up with:
```racket
(if break-sent?
    'second
    'first)
```
What would you think of flipping the meaning of the boolean?

----

> 
> I'm going to take another look at this in a few days, squash it and then it'll be ready for review.

Yes! This is a really great addition that has been a long time coming, and it definitely should not be bogged down by any of these details.

LiberalArtist:
> We use `accept` without `enable-break` because the accept is guaranteed to not block within the event handler as long as the listener is not being used concurrently (this goes back to the earlier discussion about whether the server should create and own the listener or whether it should be passed in like it is now).

Radical idea: what if we change to use procedures like `tcp-accept-evt` instead of `tcp-accept`? It would eliminate the distinction between `listen-evt` and `listener`, and it seems like it would eliminate a lot of the caveats and unchecked invariants about `tcp-accept`-like functions, re https://github.com/racket/racket/pull/4621#discussion_r1154899135. 
The one complication I see is that, for the web server, `tcp^` doesn't have a `tcp-accept-evt`, but I think we could solve that with a `tcp*^`, similarly to what we did with `web-config*^`. (This also brings to mind https://github.com/racket/compatibility/issues/8.)

It feels a lot nicer in general to have an event that does some action atomically, rather than an event that signals some action would not block, but then leaves.

It also reminded me that @mflatt had once [suggested](https://groups.google.com/g/racket-users/c/dwK14kdG4R4/m/4bLImleUAgAJ) making TCP listeners `place-message-allowed?`. If they were, the atomicity of `tcp-accept-evt` seems like it might be especially useful.

Here's an initial sketch of what I have in mind:

```patch
From 0cb9ef11441e84ff689d9663053f4622861a2b77 Mon Sep 17 00:00:00 2001
From: Philip McGrath <philip@philipmcgrath.com>
Date: Sun, 19 Jan 2025 23:04:47 -0500
Subject: [PATCH] accept-evt

---
 pkgs/net-lib/net/server.rkt | 46 ++++++++++++++++++++++---------------
 1 file changed, 27 insertions(+), 19 deletions(-)

diff --git a/pkgs/net-lib/net/server.rkt b/pkgs/net-lib/net/server.rkt
index 826dadef6e..5b97301e8b 100644
--- a/pkgs/net-lib/net/server.rkt
+++ b/pkgs/net-lib/net/server.rkt
@@ -9,18 +9,18 @@
   [run-server (server-proc/c void?)]))
 
 (define (server-proc/c res/c)
-  (->* [evt? (-> input-port? output-port? any)]
+  (->* [any/c (-> input-port? output-port? any)]
        [#:max-concurrent   (or/c +inf.0 natural-number/c)
-        #:timeout-evt-proc (-> thread? input-port? output-port? boolean? evt?)
-        #:accept-proc      (-> any/c (values input-port? output-port?))
-        #:close-proc       (-> any/c void?)]
+        #:make-timeout-evt (-> thread? input-port? output-port? boolean? evt?)
+        #:accept-evt       (-> any/c (evt/c (list/c input-port? output-port?)))
+        #:close            (-> any/c void?)]
        res/c))
 
-(define (start-server listen-evt handle
+(define (start-server listener handle
                       #:max-concurrent   [max-concurrent +inf.0]
-                      #:timeout-evt-proc [make-timeout-evt (λ (_thd _in _out _break-sent?) never-evt)]
-                      #:accept-proc      [accept tcp-accept]
-                      #:close-proc       [close tcp-close])
+                      #:make-timeout-evt [make-timeout-evt (λ (_thd _in _out _break-sent?) never-evt)]
+                      #:accept-evt       [accept-evt tcp-accept-evt]
+                      #:close            [close tcp-close])
   (define can-break?
     (break-enabled))
   (define paramz
@@ -32,8 +32,11 @@
          void
          (lambda ()
            (with-handlers ([exn:break? void])
-             (let loop ([in-progress 0])
-               (loop
+             (define server-cust (current-custodian))
+             (define supervisor-cust (make-custodian))
+             (let loop ([in-progress 0]
+                        [supervisor-cust supervisor-cust]
+                        [client-cust (make-custodian supervisor-cust)])
                 (with-handlers ([exn:fail:network?
                                  (λ (e)
                                    (begin0 in-progress
@@ -47,16 +50,17 @@
                       (let drain-loop ([in-progress in-progress])
                         (if (thread-try-receive)
                             (drain-loop (sub1 in-progress))
-                            in-progress))))
+                            (loop in-progress supervisor-cust client-cust)))))
                    (handle-evt
-                    (if (< in-progress max-concurrent) listen-evt never-evt)
-                    (lambda (listener)
-                      (define supervisor-cust (make-custodian))
-                      (define client-cust (make-custodian supervisor-cust))
+                    (if (< in-progress max-concurrent)
+                        (parameterize ([current-custodian client-cust])
+                          (accept-evt listener))
+                        never-evt)
+                    (lambda (ports)
                       (parameterize ([current-custodian client-cust])
                         (parameterize-break #f
-                          (define-values (in out)
-                            (accept listener))
+                          (define in (car ports))
+                          (define out (cadr ports))
                           (define client-thd
                             (thread
                              (lambda ()
@@ -82,9 +86,13 @@
                                  (custodian-shutdown-all client-cust)
                                  (thread-send server-thd 'done void)
                                  (custodian-shutdown-all supervisor-cust)))))
-                          (add1 in-progress)))))))))))
+                          (parameterize ([current-custodian server-cust])
+                            (define next-supervisor-cust (make-custodian))
+                            (loop (add1 in-progress)
+                                  next-supervisor-cust
+                                  (make-custodian next-supervisor-cust))))))))))))
          (lambda ()
-           (close listen-evt))))))
+           (close listener))))))
   (lambda ()
     (break-thread server-thd)
     (thread-wait server-thd)))
-- 
2.45.2
```

PR #4623, opened 2023-04-09
fix resolve-path-spec docs

### LINKED ISSUE #4622: resolve-path-spec documented incorrectly
https://github.com/racket/racket/blob/35b84b86e4a76670c0c7e4b2fcfe3aaf41c173bf/pkgs/racket-doc/syntax/scribblings/path-spec.scrbl#L11

It does not accept a 4th argument.

benknoble:
Close #4622


mflatt:
Thanks!

PR #4632, opened 2023-04-19
match: add support for hash and hash* patterns

sorawee:
As discussed in
https://racket.discourse.group/t/rfc-hash-table-pattern-matching/1686 and https://github.com/racket/racket/pull/4532
the currrent hash-table is buggy, so we would like to deprecate it and replace with a new set of patterns that are well-behaved.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -345,7 +345,7 @@ In more detail, patterns match as follows:
        ]}
 
  @item{@racket[(#,(match-kw "hash-table") (_pat _pat) ...)] ---
-       @bold{This pattern is deprecated because it has a bug.}
+       @bold{This pattern is deprecated because it can be incorrect.}
        However, many programs rely on the incorrect behavior,
        so we still provide this pattern for backward compatibility reasons.
 
@@ -359,7 +359,7 @@ In more detail, patterns match as follows:
        ]}
 
  @item{@racket[(#,(racketidfont "hash-table") (_pat _pat) ...+ _ooo)] ---
-       @bold{This pattern is deprecated because it has a bug.}
+       @bold{This pattern is deprecated because it can be incorrect.}
        However, many programs rely on the incorrect behavior,
        so we still provide this pattern for backward compatibility reasons.
 
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -274,7 +274,7 @@ In more detail, patterns match as follows:
            [_ 'not-matched]))
        ]
 
-       The behavior of unmatched key-value entries depends on @racket[_ht-opt].
+       The behavior of residue key-value entries in the hash table value depends on @racket[_ht-opt].
 
        When @racket[_ht-opt] is not provided or when it is @racket[#:closed],
        all of the keys in the hash table value must be matched.
@@ -320,11 +320,17 @@ In more detail, patterns match as follows:
        ]}
 
  @item{@racket[(#,(match-kw "hash*") [_expr _pat _kv-opt] ... _ht-opt)] ---
-       similar to @racketidfont{hash}, but (1) key-value-pattern must be grouped;
-       (2) if @racket[_ht-opt] is not specified, it behaves like @racket[#:open];
-       (3) if @racket[_kv-opt] is specified, and the key does not exist in
-       the hash table value, the default value will be matched against the
-       value pattern instead.
+       similar to @racketidfont{hash}, but with the following differences:
+
+       @itemlist[
+         @item{The key-value pattern must be grouped syntactically.}
+         @item{If @racket[_ht-opt] is not specified, it behaves like @racket[#:open]
+               (as opposed to @racket[#:closed]).}
+         @item{If @racket[_kv-opt] is specified with @racket[#:default _def-expr],
+               and the key does not exist in the hash table value, then the default value
+               from @racket[_def-expr] will be matched against the value pattern,
+               instead of immediately failing to match.}
+       ]
 
        @examples[
        #:eval match-eval
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -249,36 +249,21 @@ In more detail, patterns match as follows:
  @item{@racket[(#,(match-kw "hash") _expr _pat ... ... _ht-opt)] ---
        matches against a hash table where @racket[_expr] matches
        a key and @racket[_pat] matches a corresponding value.
-       The key matchings use the key comparator of the matching hash table.
-       Many @racket[_expr]s could evaluate to the same value.
-       The behavior of unmatched key-value entries depend on @racket[_ht-opt].
-
-       When @racket[_ht-opt] is not provided or when it is @racket[#:closed],
-       the matching is closed to extension.
-       I.e., there must not be any unmatched key-value entries.
-
-       When @racket[_ht-opt] is @racket[#:open], the matching is open to extension.
-       I.e., there can be unmatched key-value entries.
-
-       When @racket[_ht-opt] is @racket[#:rest _pat], @racket[_pat] is further
-       matched against the residue hash table.
-       If the matching hash table is immutable, this residue matching is efficient.
-       Otherwise, the matching hash table will be copied, which could be expensive.
-
 
        @examples[
        #:eval match-eval
-       (match #hash(("aa" . 1) ("b" . 2))
-         [(hash "b" b (string-append "a" "a") a) (list b a)])
-       (match #hash(("a" . 1) ("b" . 2))
-         [(hash "b" b) 'matched]
+       (match (hash "aa" 1 "b" 2)
+         [(hash "b" b (string-append "a" "a") a)
+          (list b a)])
+       (match (hash "aa" 1 "b" 2)
+         [(hash "b" _ "c" _) 'matched]
          [_ 'not-matched])
-       (match #hash(("a" . 1) ("b" . 2))
-         [(hash "b" _ #:open) 'matched]
-         [_ 'not-matched])
-       (match #hash(("a" . 1) ("b" . 2))
-         [(hash "b" _ #:rest (hash "a" a)) a]
-         [_ #f])
+       ]
+
+       The key matchings use the key comparator of the matching hash table.
+
+       @examples[
+       #:eval match-eval
        (let ([k (string-append "a" "b")])
          (match (hasheq "ab" 1)
            [(hash k v) 'matched]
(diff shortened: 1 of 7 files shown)


racket/collects/racket/match/parse.rkt
```diff
@@ -128,6 +257,30 @@
...
+    ;; Hash table patterns
+    [(hash* kvp ... #:rest rest-pat)
+     (eq? (syntax-e #'rest-pat) '_)
+     (do-hash stx #'(kvp ...) #f)]
+    [(hash* kvp ... #:rest rest-pat) (do-hash stx #'(kvp ...) #'rest-pat)]
+    [(hash* kvp ... #:closed) (do-hash stx #'(kvp ...) #t)]
+    [(hash* kvp ... #:open) (do-hash stx #'(kvp ...) #f)]
+    [(hash* kvp ...) (do-hash stx #'(kvp ...) #f)]
+
+    [(hash es ... #:rest rest-pat)
```
samth, line 270: It's better to call `do-hash` directly here rather than going through `parse` again.

racket/collects/racket/match/parse.rkt
```diff
@@ -52,6 +55,131 @@
...
+  (define k-ids (generate-temporaries k-exprs))
+  (define ref-ids (generate-temporaries k-exprs))
+
+  (with-syntax ([(k-id ...) k-ids]
+                [(ref-id ...) ref-ids]
+                [(k-expr ...) k-exprs]
+                [(v-pat ...) v-pats]
+                [(def-expr ...) def-exprs]
+                [(def-id ...) def-ids])
+    (parse
```
samth, line 116: Is it necessary to construct a pattern and then parse it? I would prefer to go directly to the relevant AST (or build new kinds of AST if that's needed).

racket/collects/racket/match/parse.rkt
```diff
@@ -52,6 +55,131 @@
...
+       (? hash?
+          ;; we use let explicitly to prevent macro expander from nesting too much
+          (app (λ (e)
+                 ;; initially assign k-ids and ref-ids to #f, so that
+                 ;; if section 1 short-circuits, there's no need to
+                 ;; evaluate all k-exprs and hash-refs
+                 (let ([k-id #f] ... [ref-id #f] ...)
+                   (values
+                    ;; SECTION 1: predicate
+                    (or (undef? (begin
```
samth, line 127: Why do we need `set!` to accomplish this?

racket/collects/racket/match/runtime.rkt
```diff
@@ -69,3 +76,21 @@
...
                 (syntax-position stx)
                 (syntax-span stx))))
+
+(define undef (gensym))
+(define user-def (gensym))
+
+(define (undef? v)
+  (eq? undef v))
+
+(define (user-def? v)
```
samth, line 86: I don't understand the role of `user-def` in this.

racket/collects/racket/match/parse.rkt
sorawee: I agree for `#:closed` and `#:open`, but for `#:rest`, I think it makes sense to go through `parse` again because it handles `_` specially, and I'd rather not duplicate that logic here.

racket/collects/racket/match/runtime.rkt
sorawee: When you use `(hash* [1 a #:default 3])`, if the hash table value doesn't have key 1, `a` is matched against `3`.

Now, consider `(hash* [1 a #:default (begin (println 'foo) 3)] [2 3])`. What should happen regarding `(println 'foo)` in the following cases?

- The hash table has key `1`
- The hash table does not have have key `1` and `2`.

The strategy that I currently use is to evaluate `(println 'foo)` only when the pattern is really matched and the corresponding key really doesn't exist. `user-def` is a dummy value to facilitate this delayed evaluation.

Of course, other strategies are possible too. I.e., always evaluate `(println 'foo)` whether or not key exists, and whether or not the clause is matched. That would be simpler to implement, but I personally don't like that, primarily because the default value might be expensive to compute. But you might have different opinion.

More generally, this is related to the order of evaluation among expressions in the pattern. Right now I try to be as lazy as possible, where keys are first evaluated, and then default values. But another possibility is to interleave key and default value evaluations.

racket/collects/racket/match/parse.rkt
sorawee: It is simply easier to write the pattern than to construct an AST explicitly. I would prefer if we can keep the pattern, since this happens at compile-time, and it doesn't change the time complexity anyway. But if you strongly think that we should switch to constructing an AST explicitly, I can do that too.

samth: My experience is that constructing and then parsing new syntax often leads to sub-optimal error behavior, and thus I'm always in favor of not doing it.

racket/collects/racket/match/runtime.rkt
gus-massa: > *The strategy that I currently use is to evaluate `(println 'foo)` only when the pattern is really matched and the corresponding key really doesn't exist.*

That's strange. What do other parts of `match` do with expressions with side effects? 

Another stranger case:

    (hash* [1 a #:default (begin (println 'foo) (random 7))] [2 3])

samth:
Generally, I think it would be helpful to look at the generated code here for a few simple examples, like:
```
(match-lambda [(hash 1 x #:closed) x])
(match-lambda [(hash 1 x #:open) x])
(match-lambda [(hash x y #:open) (list x y)])
```

sorawee:
> Generally, I think it would be helpful to look at the generated code here for a few simple examples, like:
> 
> ```
> (match-lambda [(hash 1 x #:closed) x])
> (match-lambda [(hash 1 x #:open) x])
> (match-lambda [(hash x y #:open) (list x y)])
> ```

Can you elaborate this, @samth? What generated code? Are you talking about the actual code generation, or examples in the doc? 

Note that your last example is not valid, unless `x` is already bound.

racket/collects/racket/match/runtime.rkt
sorawee: @gus-massa this issue is trivial in other patterns, because there are at most one expression in other patterns.

`match` has many short-circuiting opportunities, so generally expressions are only evaluated when needed. E.g.

```
(define (f)
  (println "foobar")
  (λ (v) #t))

(match 1
  [(and 2 (app (f))) 'matched]
  [_ 'not-matched])
```

doesn't print `"foobar"`.

---------------------

Your example is a good one. Should `random` be called once or multiple times per site? This is relevant when there's a `match` backtracking that tries the same pattern multiple times. But this already occurs in the current `match`. E.g.

```
#lang racket

(define (f)
  (println "foobar")
  (random 2))

(match (list (list 0) (list 1))
  [(list-no-order
    (list (== (f)))
    (list (== (f)))) 'matched]
  [_ 'not-matched])
```

can print `"foobar"` more than twice. The `hash` pattern has the same behavior.

I'm inclined to say that there's no guarantee regarding evaluation order when backtracking occurs. People rarely use it. And when they do, I think it's reasonable that they must ensure the program doesn't have anything "weird."

samth:
I mean, what does your implemention generate, and I'm suggesting you paste it here in this discussion.

sorawee:
The generated code won't be easy to read, so let's use the expanded pattern instead.

```
(hash* [(+ 1 1) x-id] [(+ 2 2) y-id #:default (+ 3 3)] #:open)
```

expands to:

```
(? hash?
   (app (λ (e)
          (let ([k-id1 #f]
                [k-id2 #f]
                [ref-id1 #f]
                [ref-id2 #f])
            (values
             ;; SECTION 1: predicate
             (or (undef? (begin
                           (set! k-id1 (+ 1 1))
                           (set! ref-id1 (hash-ref e k-id1 undef))
                           ref-id1))
                 (undef? (begin
                           (set! k-id2 (+ 2 2))
                           (set! ref-id2 (hash-ref e k-id2 user-def))
                           ref-id2)))

             ;; SECTION 3: values
             ref-id1
             ref-id2)))

        ;; SECTION 1
        #f

        ;; SECTION 3
        (app (λ (ref-id1)
               (if (user-def? ref-id1)
                   undef
                   ref-id1))
             x-id)
        (app (λ (ref-id2)
               (if (user-def? ref-id2)
                   (+ 3 3)
                   ref-id2))
             y-id)))
```

samth:
I guess I'd like to see the actual code generated by match, and particularly for the simple patterns I gave. For the second one, for example, it should be possible to compile it after optimization to just a hash-ref and perhaps a test.

sorawee:
That is unlikely to happen due to all code in `main-fun`. I can specialize the simple cases that you mentioned to make it generate simpler code. E.g., if it's `#:open`, I think there's no need to use `set!`. But I'm not sure if it will be worth the effort.

Anyhow, here's the CP0'ed code for `(match-lambda [(hash 1 x #:open) x])`.

```
(lambda (g1_1.1491)
  (if (hash? g1_1.1491)
      (#2%void)
      (let ([srclocs_1411.1492 ($app/value
                                syntax-srclocs3.1486
                                ($app/value
                                 .get-syntax-literal!1.1483
                                 0))])
        ($app/no-return
         raise
         (let ([app_13.1493 ((#3%$top-level-value
                              '1/format.rkt-io.sls-1/format-0)
                             "~a: no matching clause for ~e"
                             'match-lambda
                             g1_1.1491)])
           ((if (#2%procedure? make-exn:misc:match8.1485)
                make-exn:misc:match8.1485
                ($app/no-inline
                 (#3%$top-level-value
                  'slow-extract-procedure.rkt-rumble.sls-slow-extract-procedure-0)
                 make-exn:misc:match8.1485
                 4))
            app_13.1493
            (current-continuation-marks
             (#3%$top-level-value
              'the-default-continuation-prompt-tag.rkt-rumble.sls-the-default-continuation-prompt-tag-0))
            g1_1.1491
            srclocs_1411.1492)))))
  (#2%call-with-values
   (lambda ()
     (let ([temp4_8.1494 #f])
       (let ([temp5_9.1495 #f])
         (let ([app_15.1496 (#2%eq?
                             undef5.1488
                             (begin
                               (set! temp4_8.1494 1)
                               (set! temp5_9.1495
                                     (let ([k.1497 temp4_8.1494])
                                       (let ([v.1498 ((#3%$top-level-value
                                                       'hash-ref/none.rkt-rumble.sls-hash-ref/none-0)
                                                      g1_1.1491
                                                      k.1497)])
                                         (if (#3%eq?
                                              v.1498
                                              'none.kwcju864gpycc2h151s9atbmo-1)
                                             ((#3%$top-level-value
                                               'fail-hash-ref.rkt-rumble.sls-fail-hash-ref-0)
                                              'hash-ref
                                              undef5.1488)
                                             v.1498))))
                               temp5_9.1495))])
           (#2%values app_15.1496 temp5_9.1495)))))
   (case-lambda
     [(temp11_5.1499 temp12_6.1500)
      (if temp11_5.1499
          (let ([srclocs_1411.1501 ($app/value
                                    syntax-srclocs3.1486
                                    ($app/value
                                     .get-syntax-literal!1.1483
                                     0))])
            ($app/no-return
             raise
             (let ([app_13.1502 ((#3%$top-level-value
                                  '1/format.rkt-io.sls-1/format-0)
                                 "~a: no matching clause for ~e"
                                 'match-lambda
                                 g1_1.1491)])
               ((if (#2%procedure? make-exn:misc:match8.1485)
                    make-exn:misc:match8.1485
                    ($app/no-inline
                     (#3%$top-level-value
                      'slow-extract-procedure.rkt-rumble.sls-slow-extract-procedure-0)
                     make-exn:misc:match8.1485
                     4))
                app_13.1502
                (current-continuation-marks
                 (#3%$top-level-value
                  'the-default-continuation-prompt-tag.rkt-rumble.sls-the-default-continuation-prompt-tag-0))
                g1_1.1491
                srclocs_1411.1501))))
          (#2%void))
      temp12_6.1500]
     [args.1503
      ($app/no-return
       2
       args.1503)])))
```

jbclements:
Bump, @samth : is this ready to be merged, and if so, should it go in 8.10?

jbclements:
@samth, I'm thinking this will not go in the v8.10 release, but I'd like to get it merged now, so it can simmer for a full 3 months before the next release. Does this sound reasonable to you?

sorawee:
Let's definitely wait for @samth on this one. He doesn't sound to be totally on board with it yet.

sorawee:
This is ready for another look, @samth.

Here's an updated expansion:

## Closed (general case)

```
(match-lambda [(hash 1 x #:closed) x])
```

is expanded to:

```
(if (#%app hash? g12)
  (let-values ()
    (let-values (((temp9)
                  (#%app (lambda (ht) (#%app hash-state2 ht '() '())) g12)))
      (let-values (((temp11 temp12 temp13)
                    (#%app
                     (lambda (state)
                       (let-values (((ht) (#%app hash-state-ht state)))
                         (let-values (((acc-keys)
                                       (#%app hash-state-keys state)))
                           (let-values (((acc-vals)
                                         (#%app hash-state-vals state)))
                             (let-values (((key) '1))
                               (let-values (((val)
                                             (#%app hash-ref ht key undef)))
                                 (#%app
                                  values
                                  (#%app undef? val)
                                  (if (#%app user-def? val)
                                    (let-values () undef)
                                    (let-values () val))
                                  (#%app
                                   hash-state2
                                   ht
                                   (#%app cons key acc-keys)
                                   (#%app cons val acc-vals)))))))))
                     temp9)))
        (if (#%app equal? temp11 '#f)
          (let-values ()
            (if (#%app hash-state-closed? temp13)
              (let-values ()
                (let-values ()
                  (let-values ()
                    (let-values (((x) temp12)) (let-values () x)))))
              (let-values () (#%app fail3))))
          (let-values () (#%app fail3))))))
  (let-values () (#%app fail3)))
```

## Open (simple case)

```
(match-lambda [(hash 1 x #:open) x])
```

is expanded to:

```
(if (#%app hash? g12)
  (let-values ()
    (let-values (((temp9 temp10)
                  (#%app
                   (lambda (ht)
                     (let-values (((val)
                                   (#%app hash-ref ht '1 (lambda () undef))))
                       (#%app values (#%app undef? val) val)))
                   g12)))
      (if (#%app equal? temp9 '#f)
        (let-values ()
          (let-values ()
            (let-values () (let-values (((x) temp10)) (let-values () x)))))
        (let-values () (#%app fail3)))))
  (let-values () (#%app fail3)))
```

racket/collects/racket/match/parse.rkt
sorawee: `set!` is no longer needed.

pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -256,8 +358,12 @@ In more detail, patterns match as follows:
          [(hash-table ("b" b) ("a" a)) (list b a)])
        ]}
 
- @item{@racket[(#,(racketidfont "hash-table") (_pat _pat) ...+ _ooo)]
-       --- Generalizes @racketidfont{hash-table} to support a final
+ @item{@racket[(#,(racketidfont "hash-table") (_pat _pat) ...+ _ooo)] ---
+       @bold{This pattern is deprecated because it has a bug.}
```
samth, line 362: I would prefer "because it can be incorrect".

sorawee: Changed!

samth:
Can you post a couple examples of the new expansion? And also, does TR build and more generally raco setup of the whole distribution?

sorawee:
I already provided some expansions in the above comment. Is there anything else you would like to see in particular?

I expect that this PR will not affect TR or any other code, since it's a new pattern. We are not modifying anything. But I can recheck.

samth:
Oh sorry, I missed that comment. The expansion looks good. The only thing I see that could be improved is avoiding the immediately applied lambda, which might help TR in the future. But we can address that down the road.

samth (approved):

samth:
Yay! 🎉

PR #4637, opened 2023-04-22
Respect `#:decimal-sep` with exponential notation (Fixes #4635)

### LINKED ISSUE #4635: `~r` ignores `#:decimal-sep` for `#:notation 'exponential`
8.8 [cs]

**What program did you run?**

```
$ racket
Welcome to Racket v8.8 [cs]
> (~r 12345678.123456 #:notation 'exponential #:decimal-sep ",")
"1.234568e+07"
```

"Positional" notation for comparison:
```
> (~r 12345678.123456 #:notation 'positional #:decimal-sep ",")
"12345678,123456"
```

The output should be "1,234568e+07" (decimal _comma_ because it was specified with `#:decimal-sep ","`.

n.a.

Fedora Linux 37, Racket from Racket download page

If the behavior is intended, it should be explained in the documentation. However, I can't imagine why the behavior should be intentional. :-)

otherjoel:
Currently when you give `~r` a non-zero number with `#:notation 'exponential` it uses `#\.` as the hard-coded decimal separator, ignoring the `#:decimal-sep` argument.

I tried to think of a clever way to keep the original speed hack in place but I’m not that smart. Of course feel free to edit if there’s a better way.

### DIFF
--- a/pkgs/racket-test/tests/racket/format.rkt
+++ b/pkgs/racket-test/tests/racket/format.rkt
@@ -320,6 +320,12 @@
     "0e+00")
 (tc (~r #:notation 'exponential 0 #:precision '(= 4))
     "0.0000e+00")
+(tc (~r #:notation 'exponential 0 #:precision '(= 4) #:decimal-sep ",")
+    "0,0000e+00")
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks!

PR #4638, opened 2023-04-25
Support .rhm as a module suffix.

samth:
Right now, opening require paths doesn't find `.rhm` files in DrRacket or racket-mode. This should fix that, although I'm not sure if it's the best way.


rfindler:
I see them when I do "File|Open Require Path..." without this change (and I believe I also see .scrbl files, which aren't in that list).

mflatt:
The "info.rkt" file for the "rhombus" collection includes `(define module-suffixes '(#"rhm"))`, so that's how ".rhm" is supposed to work.

PR #4640, opened 2023-04-26
BC: shared mutable empty string

AlexKnauth:
Fixes `unsafe-string->immutable-string!` on BC so that it doesn't mutate any empty strings, since mutable empty strings may be shared through `zero_length_char_string` (now `scheme_zero_length_char_string`). Instead of mutating empty strings, it returns `scheme_zero_length_char_immutable_string`.


mflatt:
The `static` declaration on`zero_length_char_immutable_string` means that it's instantiated separately for every `.c` file. Change `static` with `extern` and keep (a copy of) the declarations in `string.c` without `extern` — but also add `scheme_` to the front of each name (because all non-`static` names should have that prefix). Also drop `READ_ONLY` in `schpriv.h`, since the convention is to put it only on the defining declaration.

mflatt:
Thanks!

PR #4641, opened 2023-04-26
Correct docs for `write-resources`

otherjoel:
In testing, I have not been able to get `write-resource` to return `#f` when writing to the  registry; it always falls back to writing to 
mred.ini and returns #t. This edit brings the docs in line with that reality. 

Arguably, there should be a way to force various kinds of exceptions if the registry write attempt fails (due to permissions, or to non-existent keys with `create-key?` set to `#f`). Also, there don’t appear to be any tests for the functions in `file/resource`.


mflatt:
Thanks!

PR #4644, opened 2023-05-04
Fxvector and flvector in fasl

Metaxal:
This PR is for supporting fxvector and flvector in fasl.

Added code and tests.

I did not update the docs --- it doesn't seem that there's anything to update.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/fasl.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/fasl.scrbl
@@ -1,5 +1,5 @@
 #lang scribble/doc
-@(require "mz.rkt" (for-label racket/fasl racket/serialize))
+@(require "mz.rkt" (for-label racket/fasl racket/serialize racket/fixnum racket/flonum))
 
 @(define fasl-eval (make-base-eval))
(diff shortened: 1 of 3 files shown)


mflatt:
Some quick initial thoughts:

Since flvector and fxvector literals can be in compiled code in BC, it does seem like the documentation says this should work for BC already. So, it's a bug fix in that sense. CS doesn't support flvector and fxvector literals, but that's because CS uses `racket/fasl`... So, fixing `racket/fasl` will fix CS, if I'm remembering all this correctly. Still, I think it probably makes sense to add a `history` note in the documentation for this change.

It would be better to specialize the output format to avoid a redundant tag on each fxvector or flvector element. Each element of `fxvector` could be written directly with `write-fasl-integer`, and each element of `flvector` can be written using the `write-bytes` that's currently in the `flonum?` case (and should be lifted out to a function). That specialization might be usefully more compact and faster — and since the fasl format is intended to be forward-compatible, it seems worth investing effort now.

Metaxal:
Good points. The new commits should address them. Happy to revise in any case.

PR #4646, opened 2023-05-10
Add syntax-property for srcloc of module+ spliced source modules

greghendershott:
As discussed in <https://racket.discourse.group/t/finding-srcloc-for-modules-in-fully-expanded-code/1932>.

The motivation is to make it possible for check-syntax(-like) tools to build from fully expanded syntax a data structure supporting a conversion of a source position to a module name. 

This would support various things:

- a "run/enter module at point" command
- building smarter (more limited) lists of completion candidates, based on imports visible in the module around point
- possibly other things

I'm not attached to the name of the syntax property, or its value using a `srcloc` `struct` (as opposed to say a `vector` of srcloc values).

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/syntax.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax.scrbl
@@ -369,7 +369,7 @@ A submodule must not be defined using @racket[module+] @emph{and}
 of @racket[module+] pieces, then it must be made @emph{only} of
 @racket[module+] pieces. }
 
-@history[#:added "8.9.0.1"
+@history[#:changed "8.9.0.1"
          @elem{Added @racket['origin-form-srcloc] syntax property.}]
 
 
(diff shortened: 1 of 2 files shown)


mflatt:
This approach seems ok to me.

I think it makes sense to document this addition as part of the `module+` form's documentation.

The functionality here doesn't seem inherently tied to `module+`. I can imagine that the `'origin` property might be insufficient for some other form. Maybe the property name should be more generic, so we can use the same name if and when we implement the same idea for some other form? If that seems like a good idea, maybe `'origin-form-srcloc`, or something that similarly suggests an improvement over `'origin`?  (Making a property like this automatic `'origin` sounds expensive for something that isn't usually needed, though.)

greghendershott:
Thank you for reviewing and the feedback!

I pushed some commits to implement your suggestions.

greghendershott:
p.s.

> (Making a property like this automatic 'origin sounds expensive for something that isn't usually needed, though.)

The idea it could be automatic hadn't occurred to me. Automatic does sound appealing: My needs tend toward "srcloc all the things!". :)

If something automatic were too expensive, I wonder if the expander could do so only when parameterized, so things like check-sytnax could enable it, but it would be off by default?

---

Having said all that, I'm not immediately aware of more examples like `module+`.

The only examples on my to-do-list (that I'll open distinct issues or PRs for) are things like `prefix-in` -- where missing or weird srcloc is due IIRC to the require transformer system discarding syntax objects for symbols. (And so any better/automatic handling of syntax object srcloc in certain scenarios would probably be N/A.)

mflatt:
Thanks!

PR #4647, opened 2023-05-12
add index entry for configure-expand

jbclements:
It looks like you cannot currently search for `configure-expand` in the documentation. I believe this pull request repairs this.


mflatt:
Thanks! I've updated to use `as-index` on the existing instance of the name, and to do the same for `configure-runtime`.

PR #4649, opened 2023-05-15
stx prop on prefixed identifiers, plus changing export-sym to export-id

greghendershott:
**Not intended to be merged**. This does not compile. For example we can't actually import/use format-id from racket/syntax here.

The point of this PR is to outline a change where the old `out-sym` field of `export` is now `out-id-or-sym`, and, although an identifier is preferred, a symbol is still supported for backward compatibility.

Although this seems feasible, it's likely I'm failing to understand some basic things about how provide transformers work.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/module-provide.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/module-provide.scrbl
@@ -104,5 +104,12 @@ omit.
 Like @racket[provide-spec], but adding @racket[prefix-id] to the
 beginning of the external name for each exported binding.
 
+A @tech{syntax property} with the key
+@indexed-racket['import-or-export-prefix-ranges] is added to the
+resulting identifier.
+
+@history[#:changed "8.9.0.5"
+         @elem{Added @racket['import-or-export-prefix-ranges] syntax property.}]
+
 }
 
--- a/pkgs/racket-doc/scribblings/guide/module-provide.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/module-provide.scrbl
@@ -104,12 +104,5 @@ omit.
 Like @racket[provide-spec], but adding @racket[prefix-id] to the
 beginning of the external name for each exported binding.
 
-A @tech{syntax property} with the key
-@indexed-racket['import-or-export-prefix-ranges] is added to the
-resulting identifier.
-
-@history[#:changed "8.9.0.5"
-         @elem{Added @racket['import-or-export-prefix-ranges] syntax property.}]
-
 }
 
--- a/pkgs/racket-doc/scribblings/guide/module-require.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/module-require.scrbl
@@ -97,13 +97,6 @@ identifiers from @racket[require-spec] that are not mentioned as an
 This is a shorthand for renaming, where @racket[prefix-id] is added to
 the front of each identifier specified by @racket[require-spec].
 
-A @tech{syntax property} with the key
-@indexed-racket['import-or-export-prefix-ranges] is added to the
-resulting identifier.
-
-@history[#:changed "8.9.0.5"
-         @elem{Added @racket['import-or-export-prefix-ranges] syntax property.}]
-
 }
 
 The @racket[only-in], @racket[except-in], @racket[rename-in], and
--- a/pkgs/racket-doc/scribblings/guide/module-require.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/module-require.scrbl
@@ -97,6 +97,13 @@ identifiers from @racket[require-spec] that are not mentioned as an
 This is a shorthand for renaming, where @racket[prefix-id] is added to
 the front of each identifier specified by @racket[require-spec].
 
+A @tech{syntax property} with the key
+@indexed-racket['import-or-export-prefix-ranges] is added to the
+resulting identifier.
+
+@history[#:changed "8.9.0.5"
+         @elem{Added @racket['import-or-export-prefix-ranges] syntax property.}]
+
 }
 
 The @racket[only-in], @racket[except-in], @racket[rename-in], and
--- a/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
@@ -1674,7 +1674,7 @@ Returns @racket[#t] if @racket[v] has the
 
 
 @defstruct[export ([local-id identifier?]
-                   [out-id identifier??]
+                   [out-id identifier?]
                    [mode phase+space?]
                    [protect? any/c]
                    [orig-stx syntax?])]{
--- a/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/stx-trans.scrbl
@@ -1674,7 +1674,7 @@ Returns @racket[#t] if @racket[v] has the
 
 
 @defstruct[export ([local-id identifier?]
-                   [out-sym symbol?]
+                   [out-id identifier??]
                    [mode phase+space?]
                    [protect? any/c]
                    [orig-stx syntax?])]{
@@ -1686,7 +1686,7 @@ A structure representing a single exported identifier:
  @item{@racket[local-id] --- the identifier that is bound within the
        exporting module.}
 
- @item{@racket[out-sym] --- the external name of the binding.}
+ @item{@racket[out-id] --- the external name of the binding.}
 
  @item{@racket[mode] --- the @tech{phase level} and @tech{binding
        space} of the export (which affects how it is imported).}
@@ -1699,8 +1699,15 @@ A structure representing a single exported identifier:
 
 ]
 
-@history[#:changed "8.2.0.3" @elem{Generalized @racket[mode] to phase--space combinations.}]}
+@history[#:changed "8.2.0.3" @elem{Generalized @racket[mode] to phase--space combinations.}]
 
+@history[#:changed "8.9.0.5" @elem{Changed the @racket[out-sym] field
+        to @racket[out-id]. For backward compatibility, the
+        @racket[make-export] constructor also accepts a symbol, and a
+        @racket[export-out-sym] function returns the @racket[syntax-e]
+        value of the @racket[out-id].}]
+
+}
 
 @defproc[(syntax-local-provide-certifier)
          ((syntax?) (or/c #f (syntax? . -> . syntax?))
--- a/pkgs/racket-doc/scribblings/reference/syntax.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax.scrbl
@@ -1133,11 +1133,7 @@ as follows.
      (define num-eggs 2))
    (require 'nest)
    chicken:num-eggs
- ]
-
- @history[#:changed "8.9.0.4" @elem{Each export name has a
-          @indexed-racket['sub-range-binders] @tech{syntax
-          property}.}]}
+ ]}
 
  @defsubform[(struct-out id)]{Exports the bindings associated with a
  structure type @racket[id]. Typically, @racket[id] is bound with
--- a/pkgs/racket-doc/scribblings/reference/syntax.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax.scrbl
@@ -608,7 +608,11 @@ bindings of each @racket[require-spec] are visible for expanding later
     (require (prefix-in tcp: racket/tcp))
     tcp:tcp-accept
     tcp:tcp-listen
-  ]}
+  ]
+
+  @history[#:changed "8.9.0.5" @elem{Add a @tech{syntax property} with
+           the key @indexed-racket['import-or-export-prefix-ranges] to
+           the resulting local identifier.}]}
 
  @defsubform[(rename-in require-spec [orig-id bind-id] ...)]{
   Like @racket[require-spec], but replacing the identifier to
@@ -1133,7 +1137,11 @@ as follows.
      (define num-eggs 2))
    (require 'nest)
    chicken:num-eggs
- ]}
+ ]
+
+ @history[#:changed "8.9.0.5" @elem{Add a @tech{syntax property} with
+          the key @indexed-racket['import-or-export-prefix-ranges] to
+          the resulting exported identifier.}]}
 
  @defsubform[(struct-out id)]{Exports the bindings associated with a
  structure type @racket[id]. Typically, @racket[id] is bound with
--- a/pkgs/racket-doc/scribblings/reference/syntax.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax.scrbl
@@ -1133,7 +1133,11 @@ as follows.
      (define num-eggs 2))
    (require 'nest)
    chicken:num-eggs
- ]}
+ ]
+
+ @history[#:changed "8.9.0.4" @elem{Each export name has a
+          @indexed-racket['sub-range-binders] @tech{syntax
+          property}.}]}
 
  @defsubform[(struct-out id)]{Exports the bindings associated with a
  structure type @racket[id]. Typically, @racket[id] is bound with
--- a/racket/collects/racket/private/reqprov.rkt
+++ b/racket/collects/racket/private/reqprov.rkt
@@ -315,6 +315,7 @@
                        [(submod . s)
                         (check-lib-form in)
                         (list (mode-wrap base-mode (xlate-path in)))]
+                       ;; See `prefix-in` function comment.
                        [(prefix-in pfx path)
                         (simple-path? #'path)
                         (list (mode-wrap
@@ -729,7 +730,16 @@
                   (map car new+olds)
                   leftover-imports)
                  sources))))]))))
-  
+
+  ;; In simple cases (require (prefix-in)) expands to a single
+  ;; (#%require [prefix pre mod]) clause. Such simple cases bypass
+  ;; this `prefix-in` transformer. That's fine because the prefix and
+  ;; suffix identifiers are distinct in the prefix clause. Analyzers
+  ;; like drracket/check-syntax already see/use these.
+  ;;
+  ;; Whereas this handles the general cases, including nested
+  ;; prefix-in. These expand to (#%require [rename new old]). Here we
+  ;; do need a sub-range-binders property on `new`.
   (define-syntax prefix-in
     (make-require-transformer
      (lambda (stx)
@@ -746,14 +756,7 @@
             (values
              (map (lambda (import)
                     (let ([id (import-local-id import)])
-                      (make-import (datum->syntax
-                                    id
-                                    (string->symbol
-                                     (format "~a~a" 
-                                             (syntax-e pfx)
-                                             (syntax-e id)))
-                                    id
-                                    id)
+                      (make-import (append-ids id pfx id)
                                    (import-src-sym import)
                                    (import-src-mod-path import)
                                    (import-mode import)
@@ -1249,37 +1252,12 @@
             (check-prefix stx #'pre-id)
             (let ([exports (expand-export #'out modes)])
               (map (lambda (e)
-                     ;; Although we can't import format-id from
-                     ;; racket/syntax here, we do the equivalent for
-                     ;; this special case, similar to the
-                     ;; hyphentated-id example in the Check Syntax
-                     ;; docs for sub-range-binders.
-                     (let* ([pre-str (symbol->string (syntax->datum #'pre-id))]
-                            [pre-len (string-length pre-str)]
-                            [out-id  (export-out-id e)]
-                            [out-str (symbol->string (syntax->datum out-id))]
-                            [out-len (string-length out-str)]
-                            [out-srb (syntax-property out-id 'sub-range-binders)]
-                            [new-sym (string->symbol (string-append pre-str out-str))]
-                            [new-id  (datum->syntax #'pre-id new-sym #'pre-id #'pre-id)]
-                            [new-srb (list
-                                      (vector-immutable (syntax-local-introduce new-id)
-                                                        0 pre-len 0.5 0.5
-                                                        (syntax-local-introduce #'pre-id)
-                                                        0 pre-len 0.5 0.5)
-                                      (vector-immutable (syntax-local-introduce new-id)
-                                                        pre-len out-len 0.5 0
-                                                        (syntax-local-introduce out-id)
-                                                        0 out-len 0.5 1))]
-                            [srb     (if out-srb
-                                         (cons new-srb out-srb)
-                                         new-srb)])
-                       (make-export
-                        (export-local-id e)
-                        (syntax-property new-id 'sub-range-binders srb)
-                        (export-mode e)
-                        (export-protect? e)
-                        (export-orig-stx e))))
+                     (make-export
+                      (export-local-id e)
+                      (append-ids #'pre-id #'pre-id (export-out-id e))
+                      (export-mode e)
+                      (export-protect? e)
+                      (export-orig-stx e)))
                    exports))]))
        (lambda (stx modes)
          (syntax-case stx ()
@@ -1287,7 +1265,6 @@
             (check-prefix stx #'pfx)
             (with-syntax ([out (pre-expand-export #'out modes)])
               (syntax/loc stx (prefix-out pfx out)))])))))
-           
 
   ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
@@ -1365,4 +1342,69 @@
                                (values (make-rename-transformer (quote-syntax lifted)) ...)))))])))
 
   ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+
+  ;; Although we can't import format-id from racket/syntax here, we
+  ;; implement the equivalent for this special case of appending two
+  ;; identifiers. (Similar to the hyphentated-id example in the Check
+  ;; Syntax docs for sub-range-binders.)
+  ;;
+  ;; When there already exist sub-range-binders on the suffix-id -- as
+  ;; is the case with nested prefix-{in out} -- update them: Replace
+  ;; their "fuller" syntax with new-id, and shift the offset by
+  ;; prefix-len. In other words, the srbs for new-id all share the
+  ;; same "fuller" syntax in the first element of the vector; it's
+  ;; "flat" and does not reflect the nesting. (By contrast if we just
+  ;; cons them together, such that the suffix represents
+  ;; "sub-sub-binders", I think it would confuse consumers like
+  ;; drracket/check-syntax. Anyway it's easier and more reliable for
+  ;; us to "flatten" here.)
+  (define-for-syntax (append-ids lctx prefix-id suffix-id)
+    (let* (;; CAVEAT: Using sub-range-binders for nested prefix-in
+           ;; forms seems to freak out check-syntax. After looking at
+           ;; traversals.rkt, I'm not sure why, except maybe it
+           ;; assumes the positions are relative to something other
+           ;; than the supplied prefix and suffix?? For now I'm using
+           ;; another syntax-property key, but otherwise this is
+           ;; intended to match the docs for sub-range-binders.
+           ;;
+           ;; [If it turns out a new/distinct property is appropriate,
+           ;; then I would simplify its compared to sub-range-binders.
+           ;; Maybe just (vector full-sym full-ofs full-span sub-sym
+           ;; sub-ofs). No syntax object, because no free id table
+           ;; needed; just symbols and syntax positions and spans.
+           ;; "pdb" needs this only to refine the end points of
+           ;; _import_ arrows, never lexical arrows. Symbol and
+           ;; position suffices, combined with identifier-binding
+           ;; values.]
+           [prop-key    'prefix-sub-range-binders]
+           [prefix-str  (symbol->string (syntax->datum prefix-id))]
+           [prefix-len  (string-length prefix-str)]
+           [suffix-str  (symbol->string (syntax->datum suffix-id))]
+           [suffix-len  (string-length suffix-str)]
+           [new-sym     (string->symbol (string-append prefix-str suffix-str))]
+           [new-id      (datum->syntax lctx new-sym prefix-id #f)]
+           [_           (syntax-track-origin new-id prefix-id #'append-ids)]
+           [prefix-vec  (vector (syntax-local-introduce new-id)
+                                0 prefix-len 0.5 0.5
+                                (syntax-local-introduce prefix-id)
+                                0 prefix-len 0.5 0.5)]
+           [suffix-srb  (syntax-property suffix-id prop-key)]
+           [suffix-vecs (cond
+                          [suffix-srb
+                           (unless (list? suffix-srb)
+                             (raise-syntax-error #f "expected list of sub-range-binders at " suffix-id suffix-srb))
+                           (map (lambda (vec)
+                                  (vector-set! vec 0 new-id)
+                                  (vector-set! vec 1 (+ prefix-len (vector-ref vec 1)))
+                                  vec)
+                                suffix-srb)]
+                          [else
+                           (list
+                            (vector (syntax-local-introduce new-id)
+                                    prefix-len suffix-len 0.5 0
+                                    (syntax-local-introduce suffix-id)
+                                    0 suffix-len 0.5 1))])]
+           [srb         (cons prefix-vec suffix-vecs)])
+      (syntax-property new-id prop-key srb)
+      ))
   )
--- a/racket/collects/racket/private/reqprov.rkt
+++ b/racket/collects/racket/private/reqprov.rkt
@@ -1353,25 +1353,33 @@
   ;; the case with nested prefix-{in out} -- we don't just cons them.
   ;; Instead we "flatten": Replace the old widest identifier with the
   ;; new id, and shift the old offsets by the new prefix length. As a
-  ;; result there is a flat simple list of one or more prefixes.
+  ;; result there is a flat simple list of one or more prefixes and
+  ;; final suffix part.
   ;;
-  ;; The property value is a list of vectors, each of which says that
-  ;; a sub-range of the full identifier originated from another
-  ;; identifier whose syntax-e and syntax-position is given (the span
-  ;; is the same).
+  ;;   (vector/c
+  ;;    identifier?  ;the full identifier ^1 composed from others...
+  ;;    (listof (vector/c natural?       ;from this *offset* in the full id
+  ;;                      natural?       ;and up to this *span* from the offset
+  ;;                      identifier?))) ;originates from this identifier ^2
   ;;
-  ;;   (vector offset ;into the full identifier: non-neg integer
-  ;;           span   ;from offset: pos integer
-  ;;           sub-syntax-e         ;for convenience: symbol
-  ;;           sub-syntax-position) ;pos integer
+  ;; ^1: The full identifier syntax will have no useful srcloc because
+  ;; it does not appear in the original source.
   ;;
-  ;; Example: The value for "foobarbang" formed from "foo" at
-  ;; syntax-position 42, "bar" at syntax-position 99, and "bang" at
+  ;; ^2: Generally the only meaningful values to take from this sub
+  ;; identifier syntax are syntax-e and syntax-position (syntax-span
+  ;; is the same as the other span value). However note that
+  ;; syntax-position can be #f when the identifier is not original in
+  ;; the source.
+  ;;
+  ;; Example: The value for #'foobarbang formed from #'foo at
+  ;; syntax-position 42, #'bar at syntax-position 99, and #'bang at
   ;; syntax-position 777, would be:
   ;;
-  ;;   (list (vector 0 3 'foo   42)
-  ;;         (vector 3 3 'bar   99)
-  ;;         (vector 6 4 'bang 777))
+  ;;   (vector
+  ;;    #'foobarbang
+  ;;    (list (vector 0 3 #'foo)    ;syntax-position  42
+  ;;          (vector 3 3 #'bar)    ;                 99
+  ;;          (vector 6 4 #'bang))) ;                777
   ;;
   ;; Note: Although this may sound similar to the sub-range-binders
   ;; property, that is intended as a directive to drracket
@@ -1384,46 +1392,50 @@
                 (unless (identifier? stx)
                   (raise-syntax-error #f "expected identifier" stx)))
               (list lctx prefix-id suffix-id))
-    (let ([new-id (datum->syntax lctx
-                                 (string->symbol
-                                  (format "~a~a"
-                                          (syntax-e prefix-id)
-                                          (syntax-e suffix-id)))
-                                 #f ;not in original file
-                                 ;; REVIEW: Need to combine props here?
-                                 #f)])
+    (let ([full-id (datum->syntax lctx
+                                  (string->symbol
+                                   (format "~a~a"
+                                           (syntax-e prefix-id)
+                                           (syntax-e suffix-id)))
+                                  #f ;not in original file
+                                  ;; REVIEW: Need to combine props here?
+                                  #f)])
       (cond
         [(and (syntax-position prefix-id)
               (syntax-span prefix-id))
-         (let* ([prop-key    'import-or-export-prefix-ranges]
-                [prefix-vec  (vector 0
-                                     (syntax-span prefix-id)
-                                     (syntax-e prefix-id)
-                                     (syntax-position prefix-id))]
-                [suffix-vecs (cond
-                               [(syntax-property suffix-id prop-key)
-                                => (lambda (vs)
+         (let* ([prop-key     'import-or-export-prefix-ranges]
+                [prefix-vec   (vector 0
+                                      (syntax-span prefix-id)
+                                      (syntax-local-introduce prefix-id))]
+                [suffix-vecs  (cond
+                                ;; When suffix-id has the prop, discard
+                                ;; its full id. Retain its list of sub
+                                ;; id vectors, just shifting their
+                                ;; offets by the new prefix len.
+                                [(syntax-property suffix-id prop-key)
+                                 =>
+                                 (lambda (old-prop-val)
+                                   (let ([old-prop-vecs (vector-ref old-prop-val 1)])
                                      (map (lambda (v)
                                             (vector-set! v 0
                                                          (+ (syntax-span prefix-id)
                                                             (vector-ref v 0)))
                                             v)
-                                          vs))]
-                               [else
-                                ;; Note that suffix-id might lack
-                                ;; srcloc, e.g. when resulting from
-                                ;; all-from-xxx. In that case we still
-                                ;; include it; it is up to the user of
-                                ;; the prop to handle this as it sees
-                                ;; fit.
-                                (list
-                                 (vector (syntax-span prefix-id)
-                                         (or (syntax-span suffix-id)
-                                             (string-length
-                                              (symbol->string
-                                               (syntax-e suffix-id))))
-                                         (syntax-e suffix-id)
-                                         (syntax-position suffix-id)))])]
-                [prop-val    (cons prefix-vec suffix-vecs)])
-           (syntax-property new-id prop-key prop-val))]
-        [else new-id]))))
+                                          old-prop-vecs)))]
+                                ;; Base case. We include suffix-id even
+                                ;; when it lacks srcloc (as with e.g.
+                                ;; all-from-xxx). The offset/span into
+                                ;; the full id are still useful. Up to
+                                ;; client to check for #f srcloc.
+                                [else
+                                 (list
+                                  (vector (syntax-span prefix-id)
+                                          (string-length
+                                           (symbol->string
+                                            (syntax-e suffix-id)))
+                                          (syntax-local-introduce suffix-id)))])]
+                [new-vecs     (cons prefix-vec suffix-vecs)]
+                [new-prop-val (vector (syntax-local-introduce full-id)
+                                      new-vecs)])
+           (syntax-property full-id prop-key new-prop-val))]
+        [else full-id]))))
--- a/racket/collects/racket/private/reqprov.rkt
+++ b/racket/collects/racket/private/reqprov.rkt
@@ -1245,8 +1245,8 @@
       (make-provide-transformer
        (lambda (stx modes)
          (syntax-case stx ()
-           [(_ pfx out)
-            (check-prefix stx #'pfx)
+           [(_ pfx-id out)
+            (check-prefix stx #'pfx-id)
             (let ([exports (expand-export #'out modes)])
               (map (lambda (e)
                      ;; NOTE: IIUC we can't actually import format-id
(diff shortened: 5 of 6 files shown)


mflatt:
I think this direction is probably fine.

The only way I can think of it going wrong is if some code uses `(struct-copy export .... [out-sym ....] ....)`, but I didn't find any instances of `struct-copy export`, so it seems unlikely to be a problem. It's also possible that keeping a syntax object instead of a symbol preserves some scopes that could get discarded early, that but that also seems unlikely, since the `orig-stx` field will tend to have the syntax object or an enclosing one.

It might make sense to have `export-out-id` check `orig-stx` and return it if it's `syntax-e` component is the same as `out-sym-or-id` — like the first chunk of code replaced here in `reqprov.rkt`.

greghendershott:
Thanks for the feedback!

> It might make sense to have export-out-id check orig-stx and return it if it's syntax-e component is the same as out-sym-or-id — like the first chunk of code replaced here in reqprov.rkt.

OK I'll review to confirm, but I believe I myself added that chunk in a PR a year or two ago, and it's no longer necessary as of this PR.

---

Longer version:

I had made a few commits trying to preserve srcloc in both require and provide flavors of rename transformers. Just tried to work around the transformer system "downgrading" the new name identifier to a symbol, by adding back some srcloc in the final trampoline step.

But now, when I realized I also needed a syntax property (for sub-range-binders), it seemed like I had to tackle the root issue -- have the provide transformer's `export` struct be able to preserve the full identifiers.

As a result I believe that chunk is no longer necessary. My tests in `pdb` -- some of which depend on srcloc in those other cases addressed by the earlier PRs -- still pass with this PR. 

But I'll give it a few more looks and thinks.

Also I should probably update `prefix-out` docs to mention `sub-range-binders`.

greghendershott:
The latest commit d0664ec adds sub-range-binders for `prefix-in` -- but tentatively _not_ using the syntax property key `'sub-range-binders`.

Why not? When I do, drracket/check-syntax draws confusing arrows and ovals. I can't figure out why. I think I'll need some help from @rfindler, if I'm to use `'sub-range-binders` as the property key.

(Otherwise, for my own selfish purposes, a distinct syntax property key would work fine in the `pdb` project. So I need to figure out the way to go, that's acceptable to you all.

(Maybe it's even possible that a distinct property is best for Dr Racket, too. Keep in mind that these properties for `prefix-out` and `prefix-in` have a limited use. They are only used to refine the ends of import arrows -- never lexical arrows. Maybe Dr Racket already handles simple `prefix-in` differently for a good reason; maybe these needn't get thrown into the same free identifier table as other identifiers. Or maybe they should. I don't really understand enough yet to say either way. I'm just pointing out the limited use case, if that's helpful to know.)

---

To back up a bit:

Simple uses of `prefix-in` like `(require (prefix-in PRE "foo.rkt"))` expand to a special `#%require` clause `prefix`. And check-syntax looks for and handles that, already, though in a different way than sub-range-binders generally (maybe relevant?).

More complicated uses of `prefix-in`, including but not limited to nested `prefix-in`s, expand instead to a `#%require` `rename` clause for each imported item. That's what I wanted to handle here, by attaching a sub-range-binders prop to the `local-id`.

That seems to take different path through `check-syntax`, as far as I can follow the flow in traversals.rkt.

My test cases (in `pdb`) use a `prefix-define.rkt` file:

```
#lang racket/base
(define a 42)
(define b 42)
(provide (prefix-out A: a)
         (prefix-out ALL: (all-defined-out)))
(define c 42)
(provide (prefix-out NESTED: (prefix-out PREFIXES: (prefix-out FUN: c))))
(define d 42)
(provide d)
```

and `prefix-require.rkt`:

```
#lang racket/base
(require "prefix-define.rkt"
         (prefix-in IN: "prefix-define.rkt"))
A:a
ALL:a
ALL:b
IN:A:a
IN:ALL:a
IN:ALL:b
NESTED:PREFIXES:FUN:c
(require (prefix-in OUTER: (prefix-in INNER: "prefix-define.rkt")))
OUTER:INNER:d
OUTER:INNER:NESTED:PREFIXES:FUN:c
(let ([OUTER:INNER:d 42]) ;red herring
  OUTER:INNER:d)
```

Only the last 5 lines matter here, wrt `prefix-in`.

**Without** my PR, Dr Racket sees `OUTER:INNER:d` as one opaque identifier. Hovering over any portion of `OUTER:INNER:d` gives one arrow from it to the `"prefix-define.rkt"` source.

image

**With** my PR and a `sub-range-binders` property, Dr Racket sees the components. But... there are some weird green ovals suggesting relations to unrelated items. Some ovals appear two lines tall (?!).

image

image

image

---

Here is the expansion, and the property value for `OUTER:INNER:d`:

image

rfindler:
I tried to have a look but I'm not getting the same behavior. I'm on 

```
commit d0664ec4536f97c461edd762b1906b0454cc8ca1 (HEAD -> pr-4649, origin/pr/4649)
Author: Greg Hendershott <git@greghendershott.com>
Date:   Tue May 23 11:11:45 2023 -0400

    Add prefix-sub-range-binders in rename clauses from prefix-in
```

but I don't see the same thing as you in the macro stepper and the arrows I see are much simpler. 

Screenshot 2023-05-24 at 5 06 59 PM


Are there other changes I need besides the ones in this PR? (I have multiple versions of racket on my machine and have, many many many times, far too many to count, though I'm running on when I'm running the other so I've checked and rechecked that I'm running the version with the code from this PR upteen times but it is still possible I'm confusing myself..... so let me know if this doesn't make sense to you too?)

rfindler:
Trying to look at sub-range-binders property in the screenshot, it looks like `OUTER:INNER:d` has an arrow that starts at the range 176-182 and then goes to that same range (so check syntax will ignore it). The second arrow there seems to go from the range 176-182 (the "OUTER:" on line 11) to the range 194-200 (the "INNER:" on line 11), which doesn't seem right. The third arrow also starts at that same range and then goes to 201-202 (the open quote on line 11).

Also, I see that the identifiers are chopped up too. The idea of sub-range-binders was that you'd use the original actual identifier (ie somethat looks like the first identifier in that property) but then put different ranges into the other positions in the vector so indicate that this was an arrow from just a part of the identifier. You don't need (want?) to also make up a new identifier that has a smaller span.

greghendershott:
As for the property value, I'll double-check.

But mostly I'm following the hyphenate example from the docs.

The only wrinkle is that, when the suffix already has the prop, I flatten rather than just cons up "sub-sub-ranges". (Initially I did that thinking the "nesting" was what was confusing Dr Racket. Although that wasn't _the_ main problem, I thought it could be _a_ problem; anyway I kept it because it seemed saner for the consumer; e.g. it simplified what "pdb" does to read these.)

```racket
  ;; Although we can't import format-id from racket/syntax here, we
  ;; implement the equivalent for this special case of appending two
  ;; identifiers. (Similar to the hyphentated-id example in the Check
  ;; Syntax docs for sub-range-binders.)
  ;;
  ;; When there already exist sub-range-binders on the suffix-id -- as
  ;; is the case with nested prefix-{in out} -- update them: Replace
  ;; their "fuller" syntax with new-id, and shift the offset by
  ;; prefix-len. In other words, the srbs for new-id all share the
  ;; same "fuller" syntax in the first element of the vector; it's
  ;; "flat" and does not reflect the nesting. (By contrast if we just
  ;; cons them together, such that the suffix represents
  ;; "sub-sub-binders", I think it would confuse consumers like
  ;; drracket/check-syntax. Anyway it's easier and more reliable for
  ;; us to "flatten" here.)
  (define-for-syntax (append-ids lctx prefix-id suffix-id)
    (let* (;; CAVEAT: Using sub-range-binders for nested prefix-in
           ;; forms seems to freak out check-syntax. After looking at
           ;; traversals.rkt, I'm not sure why, except maybe it
           ;; assumes the positions are relative to something other
           ;; than the supplied prefix and suffix?? For now I'm using
           ;; another syntax-property key, but otherwise this is
           ;; intended to match the docs for sub-range-binders.
           ;;
           ;; [If it turns out a new/distinct property is appropriate,
           ;; then I would simplify its compared to sub-range-binders.
           ;; Maybe just (vector full-sym full-ofs full-span sub-sym
           ;; sub-ofs). No syntax object, because no free id table
           ;; needed; just symbols and syntax positions and spans.
           ;; "pdb" needs this only to refine the end points of
           ;; _import_ arrows, never lexical arrows. Symbol and
           ;; position suffices, combined with identifier-binding
           ;; values.]
           [prop-key    'sub-range-binders] ;; <=== which property key to use
           [prefix-str  (symbol->string (syntax->datum prefix-id))]
           [prefix-len  (string-length prefix-str)]
           [suffix-str  (symbol->string (syntax->datum suffix-id))]
           [suffix-len  (string-length suffix-str)]
           [new-sym     (string->symbol (string-append prefix-str suffix-str))]
           [new-id      (datum->syntax lctx new-sym prefix-id #f)]
           [_           (syntax-track-origin new-id prefix-id #'append-ids)]
           [prefix-vec  (vector (syntax-local-introduce new-id)
                                0 prefix-len 0.5 0.5
                                (syntax-local-introduce prefix-id)
                                0 prefix-len 0.5 0.5)]
           [suffix-srb  (syntax-property suffix-id prop-key)]
           [suffix-vecs (cond
                          [suffix-srb
                           (unless (list? suffix-srb)
                             (raise-syntax-error #f "expected list of sub-range-binders at " suffix-id suffix-srb))
                           (map (lambda (vec)
                                  (vector-set! vec 0 new-id)
                                  (vector-set! vec 1 (+ prefix-len (vector-ref vec 1)))
                                  vec)
                                suffix-srb)]
                          [else
                           (list
                            (vector (syntax-local-introduce new-id)
                                    prefix-len suffix-len 0.5 0
                                    (syntax-local-introduce suffix-id)
                                    0 suffix-len 0.5 1))])]
           [srb         (cons prefix-vec suffix-vecs)])
      (syntax-property new-id prop-key srb)
      ))
```

rfindler:
> If [here](https://github.com/racket/racket/pull/4649/commits/d0664ec4536f97c461edd762b1906b0454cc8ca1#diff-0e8c0ed6ac37514f0a7386d1cd13bd53dbe447a75ed87e64ef0008bd2c292dcdR1379) you change `prop-key` from `'prefix-sub-range-binders` to `'sub-range-binders` then you should see the puzzling arrows/ovals I described?

Ah, thanks! I do see it now. I'll look more carefully into it.

> The only wrinkle is that, when the suffix already has the prop, I flatten rather than just cons up "sub-sub-ranges". (Initially I did that thinking the "nesting" was what was confusing Dr Racket. Although that wasn't _the_ main problem, I thought it could be _a_ problem; anyway I kept it because it seemed saner for the consumer; e.g. it simplified what "pdb" does to read these.)

FWIW, the tree of conses isn't because of a choice at the level of DrRacket; it's because of the way the expander combines properties in some cases that forces tree of conses on us (syntax property values will always be trees of conses, in fact). But since it has to be that way, I figured might as well let anyone do it that way, in case they find it to be more convenient. Just in case, you wanted to check that part of things, [here's where it is consumed](https://github.com/racket/drracket/blob/master/drracket-tool-text-lib/drracket/private/syncheck/traversals.rkt#L644-L654) by check syntax.

greghendershott:
> FWIW, the tree of conses isn't because of a choice at the level of DrRacket; it's because of the way the expander combines properties in some cases that forces tree of conses on us (syntax property values will always be trees of conses, in fact). But since it has to be that way, I figured might as well let anyone do it that way, in case they find it to be more convenient. Just in case, you wanted to check that part of things, [here's where it is consumed](https://github.com/racket/drracket/blob/master/drracket-tool-text-lib/drracket/private/syncheck/traversals.rkt#L644-L654) by check syntax.

I understand and I've seen that in traversals.rkt. I definitely wasn't trying to say it's "too complicated" in the general case. It makes sense.

In this special case:

- require and provide transformers use their own small expander system (for inside-out combination IIUC), which AFAICT doesn't combine syntax props automatically. Since we need to, we can choose how.

- In `prefix-{in out}` if we just cons them up, then the srbs are "recursive". First you're told that `OUTER:INNER:d` is made from 2 pieces `OUTER:` and `INNER:d`. Then you're told, `INNER:D` is made from 2 pieces `INNER:` and `d`. And so on. I had code to deal with such values in `pdb`; it's doable. It's just messy, compared to `prefix-in` directly presenting `OUTER:INNER:d` as an identifier made from 3 pieces (without including a nesting history lesson :smile:).

- Although I don't think you're saying it's wrong to simplify it, if you do, I'm happy to change it back.

rfindler:
> * In `prefix-{in out}` if we just cons them up, then the srbs are "recursive". First you're told that `OUTER:INNER:d` is made from 2 pieces `OUTER:` and `INNER:d`. Then you're told, `INNER:D` is made from 2 pieces `INNER:` and `d`. And so on. I had code to deal with such values in `pdb`; it's doable. It's just messy, compared to `prefix-in` directly presenting `OUTER:INNER:d` as an identifier made from 3 pieces (without including a nesting history lesson 😄).

Ohhhh...... I see now! Thank you, that makes a ton of sense. For the `'sub-range-binders` part, all that is supposed to be resolved already and you are supposed to see the source and destination of one particular arrow (so if there was nested structure it would have had to have been resolved already too).

rfindler:
I think the goals with the addition of these properties make a lot of sense. As you observed, there isn't the right information in the fully expanded form to track all the prefixing that's happening so I'm definitely on board with these kinds of changes and I also am more than happy to change the traversals code to recognize a different format for the arrows (or other parts of the codebase).

greghendershott:
> I think the goals with the addition of these properties make a lot of sense. As you observed, there isn't the right information in the fully expanded form to track all the prefixing that's happening so I'm definitely on board with these kinds of changes and I also am more than happy to change the traversals code to recognize a different format for the arrows (or other parts of the codebase).

Thanks for the positive feedback and confirmation on the direction!

How to proceed and coordinate?  I see two paths:

- We keep the _new_ property key, `'prefix-sub-range-binders`.
  - Open to better names. Suggesting "sub-ranges for import/export identifiers created by prefix-in or prefix-out".
  - I review and clean up (e.g. delete or rewrite some comments to explain this intentional distinction; add some docs for the new syntax property; etc.).
  - This PR can be merged sooner, and used by "pdb". It can handle these edge cases for multi-file renaming (the last known holes on my to-do list).
  - You can update traversals.rkt and/or gui.rkt as/when desired. Because of the distinct key, you could handle this differently, much like you already handle `(#%require [prefix pre mod])` differently, from sub-range-binders generally.
  - Meanwhile Dr Racket doesn't improve in this (tiny?) respect... but it also doesn't break.

- We reuse the property key `'sub-range-binders`.
  - It's not possible for you to handle these differently than sub-range-binders generally (if that turns out to be helpful).
  - This can't be merged, probably, until you update Dr Racket. Otherwise, you're on the hook make changes before the next release. Meanwhile, Dr Racket can act weird. Admittedly, this is relatively rare. But the sometimes bizarre "matching identifier ovals" are confusing when they do occur.
  - Until this PR is merged, "pdb" is kind of stuck, at least wrt reliable multi-file renaming.

Are there other paths?

(If just these two, then of course I prefer the first. But want to make sure I understand your perspective/wisdom/needs/schedule!)

rfindler:
> [...] How to proceed and coordinate? [...]

I am not sure. For now, I don't understand what the interpretation of property values is. I guess my feeling is that if the property values are match for the way that `'sub-range-binders` is interpreted, we should use that and I agree that I should fix any bugs in DrRacket before we release with this change but if you wanted to merge before I fixed the bugs then it should probably be okay if we're not strictly in lock step (although I would prioritize these fixes so hopefully I wouldn't slow things down too much if you chose to wait). If, however, the way that `'sub-range-binders` is interpreted is not a match for the way that this property is interpreted, then a different name makes sense. In that case, however, my thoughts about the syncronization timing on the other are the same -- I'll prioritize it and get something released ASAP!

rfindler:
> First, I didn't mean to create a fire-drill for you.

Not at all! I just meant to convey my excitement to work on it. I get a lot of satisfaction out of seeing new arrows popup in my drracket (am I a wierdo? .... maaaybbbeeee..... :)

I tried to look at a specific example and didn't understand how the information in the properties was connected to arrows that might get drawn. I can share the details here if that seems helpful?

greghendershott:
> I tried to look at a specific example and didn't understand how the information in the properties was connected to arrows that might get drawn. I can share the details here if that seems helpful?

Yes that would be great.

greghendershott:
Meanwhile, one idea.

The sub-range-binders are on the `local-id` in the `#require` `rename` clause. You discover them here within `phaseless-spec->require-context`:

```racket
    [(rename raw-module-path local-id exported-id)
     (found-local-id #'local-id)
     (require-context (list #'local-id) #t #f (binder+mods #'raw-module-path mods))]
```

1. Note that `local-id` -- e.g. "OUTER:INNER:d" -- is _not_ original in the user source `require`. For every other `rename` example I can think of, e.g. `rename-in` or `only-in`, the new name id _would_ be there in the original `require`.

2. Also the srcloc for "OUTER:INNER:d" is the id for the outermost `prefix-in`, "OUTER:" (because what else could it be).

3. Finally, there will be a `#%require` `rename` clause for every imported item -- and sub-range-binders on its local-id -- even if local-id isn't used by the program. (Presumably that's why the special `prefix` clause exists, to optimize simpler cases, avoid hundreds or thousands of `rename` clauses (think `(prefix-in pre: racket)` vs. `(prefix-in a: (prefix-in b: racket))`).)

Does any of this violate some assumption(s) in traversals.rkt -- in how the sub-range-binders directives are used to:

  - draw arrows
  - draw "matching id" ellipses
    - how many
    - size

rfindler:
If I'm understanding point 3, I think this does conflict with the idea of sub-range-binders, because each vector inside a sub-range-binders is supposed to correspond to an arrow that's actually going to be drawn in a specific place in the buffer. It sounds like that's not how you're thinking about things from the multi-file rename perspective? 

As a concrete example, with this in "tmp.rkt" (a trimmed down version of the file you posted higher up):

```
#lang racket/base
(define c 42)
(provide (prefix-out P: c))
```

and this code in "tmp2.rkt":

```
#lang racket/base
(require "tmp.rkt"
         (prefix-in OUTER: (prefix-in INNER: "tmp.rkt")))

P:c
```

I see these three vectors 

```
#(#<syntax:tmp2.rkt:3:20 OUTER:INNER:P:c> 0 6 0.5 0.5 #<syntax:tmp2.rkt:3:20 OUTER:> 0 6 0.5 0.5)
#(#<syntax:tmp2.rkt:3:20 OUTER:INNER:P:c> 6 6 0.5 0.5 #<syntax:tmp2.rkt:3:38 INNER:> 0 6 0.5 0.5)
#(#<syntax:tmp2.rkt:3:20 OUTER:INNER:P:c> 12 3 0.5 0 #<syntax:tmp2.rkt:3:45 P:c> 0 3 0.5 1)
```

where the first syntax object in each of the vector's syntax-position is 58, and the syntax-position of the second one in the first vector is 58, the second one in the second vector is 76, and the second one in the third vector is 83.

The interpretation of this is three arrows, one that starts at the range 58:64 and also ends there, the second that starts at 58:64 and ends at 76:82 and a third that starts at 60:63 and ends at 80:86 (except that I think check syntax drops some because the offsets are larger than the spans in some cases). 

Arrows at these source locations, however, doesn't make sense from the user viewing the program text, however. I guess that things along these lines are why DrRacket seems to go crazy with the larger program that you tried out.

I guess my question is what information is contained in these six vectors that tells you something about this program and helps renaming? I'm still more than happy to adapt -- if these make sense in some other way, we'd probably have to use a different key for the syntax property symbol, tho (not a big deal!).

greghendershott:
You'd sent me some private email which I noticed and caught up with, earlier today.

But to reply to your comment here, and summarize:

I (mis?)understood the `sub-range-binders` syntax property to mean generally "this identifier was made from other identifiers; here are those pieces".

However, it seems like you intended it to mean, more specifically, "tell check-syntax to draw more arrows".

So if that causes confusing arrows (or worse, confusing "matching identifier" ellipses) to be drawn, then it's not the right property to use here. Instead it should be some new/other property (as in e.g. my commit, the key is `'prefix-sub-range-binders`).

Having said all that, you were going to ping Matthew offline to get his ideas.

Thanks again for your help reviewing this!

rfindler:
Thanks, @greghendershott ! Does the behavior in the example I wrote above (the message starting "If I'm understanding point 3,...") a bug in the code in this PR? Or is that the expected result for that example? As you wrote in the offline conversation, that result should match up to three parts of an identifier whose name is "OUTER:INNER:P:c" but there isn't such an identifier in that example. I'm not sure, however, if the goal is to put a general piece of information about the structure of the requires/provides or if it is supposed to be some identifier-specific information.

greghendershott:
> Thanks, @greghendershott ! Is the behavior in the example I wrote above (the message starting "If I'm understanding point 3,...") a bug in the code in this PR? Or is that the expected result for that example? As you wrote in the offline conversation, that result should match up to three parts of an identifier whose name is "OUTER:INNER:P:c" but there isn't such an identifier in that example.

Even if the user program has no references to it, that identifier always exists in the fully-expanded code -- within a `rename` clause of `%#require`. In your example, something like:

```racket
(#%require (just-meta 0 (just-space #f (rename "tmp.rkt" OUTER:INNER:P:c P:c))))
```

In fact, that `OUTER:INNER:P:c` identifier is the syntax to which the property is attached.

But yeah that identifier syntax is not an original/visible part of the user program, so trying to draw an arrow there will definitely be strange.

Is there a way to flag not to draw an arrow? Would you testing `syntax-original?` or something like that help?? (Or I could set the srcloc to `#f`, but IIRC you need at least some info from that, like the position and/or span??)

---

> I'm not sure, however, if the goal is to put a general piece of information about the structure of the requires/provides or if it is supposed to be some identifier-specific information.

General information: _If_ the composed imported identifier is referenced anywhere, how to unpack its pieces. I need the moral equivalent of an interval map: "Given some offset into the composed id, instead treat that as this other offset into this other sub id."

[How could the `require` `prefix-in` transformer know the zero or more later references -- which don't yet exist -- to the identifier it's creating?]

greghendershott:
I just edited my previous comment twice. I don't recall if GitHub emails edits. So to avoid confusion, maybe please check the latest version live on GitHub https://github.com/racket/racket/pull/4649#issuecomment-1567271205.

rfindler:
It sounds like these property values are not something that match how DrRacket interprets the `'sub-range-binders` property. The values in that property are, as you say, meant to each correspond to an arrow. But, interpreted as arrows, these property values would point to parts of the program in the source that aren't actually there. Based on your comments above, I've made an attempt to revise the documentation for `'sub-range-binders` ([Diff here](https://github.com/racket/drracket/commit/73188bd370906a9edf8f37637bdc08e87d9bd962). It really was written in quite an obtuse manner before!)

I see the occurrence of `OUTER:INNER:P:c` in the fully expanded form of the `tmp2` module in my example above, also as you say. It's source location is bogus, however, looking like it is pointing the open double quote in the string `"tmp.rkt"` in the `require`. It looks like Check Syntax ignores this identifier because `syntax-original?` returns `#f` for it. 


> > I'm not sure, however, if the goal is to put a general piece of information about the structure of the requires/provides or if it is supposed to be some identifier-specific information.
> 
> I don't know how to make this a "draw arrow opcode". How could the `require` `prefix-in` transformer know the zero or more later references -- which don't yet exist -- to the identifier it's creating?
> 
> I don't need check-syntax to draw arrows for these; "pdb" already can figure out what it needs from the status quo import arrow, plus this info. (Although I'm happy to work with you to draw end user arrows in Dr Racket, if desired/practical.)

Maybe the right thing here is that we can reuse the smarts that pdb has to figure out what's going on with the identifiers and when I (it is taking longer than I'd hoped) finish refactoring the GUI part of check syntax to have a pdb-style interface to the information it is drawing, then putting the real pdb in there will cause DrRacket to have these arrows. Does that seem like the best path forward?

Another idea would be to have the code in traversals.rkt track these syntax properties and report them via a new method in `annotations<%>` that the tools can do with as they wish. Then we can work out who/what should figure out the arrows at a later date. I do think it would be really great to have arrows for situations like what this code is addressing, eg when there are two `prefix-in`s.

------

Should the name of the property should be something that doesn't have "binding" in the name but has "prefix-in" or "require" in the name to help indicate the purpose of the property is to record that some prefixing on imported names happened? 

I did also think briefly about the idea that the property should not have one vector per identifier but instead one vector per use of `prefix-in` but I concluded that would be a bad idea as it would mean that the property would have to be in a specific place in the expanded program (ie on the syntax object enclosing the rename that the prefix-in expanded to) and that might be trouble for reusing this property if other languages want to leave behind information about prefixing on import.

Along those lines, I'm wondering if the documentation for the property belongs in the tools documentation so that others who do add other prefixing import macros (eg rhombus?) can generate the property too and have stuff "just work".

----

Also, I feel like it has taken me a lot longer than it should to get my head around what you're doing here. Thanks for your patience explaining it all!

greghendershott:
OK here's the current status:

- The syntax property key is `'import-or-export-prefix-ranges`. (IIUC @rfindler suggestion.)

   On the one hand this feels like a property with more general use -- whenever you might want to reveal the parts of an identifier, without also the "draw arrows!" directive. OTOH I don't now know of other uses cases, so maybe it's best to keep the name so specific.

- The syntax property value is much simpler than sub-range-binders (see below).

Here is the full comment for the `prefix-identifier` function used by both `prefix-in` and `prefix-out`.

```
  ;; Combine prefix and suffix identifiers into a new identifier.
  ;;
  ;; When prefix-id and suffix-id have sufficient srcloc, add a syntax
  ;; property to reveal them. (It's possible they may have no srcloc,
  ;; as in the module.rktl tests of prefix-out.)
  ;;
  ;; When there already exists a property on the suffix-id -- as is
  ;; the case with nested prefix-{in out} -- we don't just cons them.
  ;; Instead we "flatten": Replace the old widest identifier with the
  ;; new id, and shift the old offsets by the new prefix length. As a
  ;; result there is a flat simple list of one or more prefixes.
  ;;
  ;; The property value is a list of vectors, each of which says that
  ;; a sub-range of the full identifier originated from another
  ;; identifier whose syntax-e and syntax-position is given (the span
  ;; is the same).
  ;;
  ;;   (vector offset ;into the full identifier: non-neg integer
  ;;           span   ;from offset: pos integer
  ;;           sub-syntax-e         ;for convenience: symbol
  ;;           sub-syntax-position) ;pos integer
  ;;
  ;; Example: The value for "foobarbang" formed from "foo" at
  ;; syntax-position 42, "bar" at syntax-position 99, and "bang" at
  ;; syntax-position 777, would be:
  ;;
  ;;   (list (vector 0 3 'foo   42)
  ;;         (vector 3 3 'bar   99)
  ;;         (vector 6 4 'bang 777))
  ;;
  ;; Note: Although this may sound similar to the sub-range-binders
  ;; property, that is intended as a directive to drracket
  ;; check-syntax to draw arrows between identifiers present in the
  ;; original user program. Which isn't necessarily possible here; the
  ;; new id might be present only in the fully-expanded `rename`
  ;; clause and have no references in the original.
```

---

How does this seem, generally?

If OK, I have a couple known to-do items:

- [ ] Write docs about the stx prop for the drracket tools page.
- [ ] Review how stx props (other than this) are combined from the identifiers (currently they are discarded; OK???).

After which I might be ready to remove the "DRAFT/WIP" flag from this.

rfindler:
I'm having a little trouble understanding how this information is supposed to be used. Is the idea that this property is on an identifier to indicate the provenance of that identifier? So the context of the identifier matters?  If I'm getting that right, what happens if the property is put on an identifier that happens to also be an (identifier) macro which is then subject to further expansion? Can this cause the property to be "wrong" (for some value of that word)?

Also: is it the case that if this property appears on a binding occurrence of the identifier then, if there are any bound uses of that identifier in the fully expanded program, we should know that it applies to the bound uses too?  

I'm also unsure of how macros more generally interact with this property and how the information can be useful if the identifier does not correspond to one that actually appears in the original program. (Maybe these are all the same question and I'm just missing some larger point about how this works?)

greghendershott:
Below when I say "rename sites" imagine a command that renames some identifier -- or prefix, or name introduced by a surface `require`'s `rename-in` or `only-in` -- across multiple files.

How would that work.

It needs to follow a chain:

- of same-named sites in the user sources;

- across multiple files;

- in either direction: If the user gives a use site, then it needs to follow the chain to find the definition/introduction site... then "fan back out" the other way to find all the use sites;

- the chain must be granular, following all export/import steps (what `identifier-binding` refers to as the _nominal_ source of the identifier);

A huge head start on this, _within_ each file, comes from check-syntax arrows, both lexical and import arrows. But that's not quite enough. So I'm doing more, to support multi-file renaming.

The use cases for the syntax property are:

Use case 1: Following a chain of rename sites involving **exports** created by `prefix-out`, possibly nested, that expand to `#%provide` `rename` clauses where the `export-id` is composed and does not appear in the user source.

Use case 2: Following a chain of rename sites involving **imports** created in ways that don't expand to a simple `#%require` `prefix` clause (and therefore check-syntax doesn't supply sufficient arrows) but instead to a `rename` clause whose local-id is composed and does not appear in the user source. This happens with any sufficiently non-simple use of `prefix-in`, including nested.

In both cases the stx prop lets me recover the pieces of the composed identifier such that I can understand which piece is involved. Is it a prefix (and if so which prefix, from where in the user source) or not.

[_Without_ this stx prop, such composed identifiers are opaque blobs. They don't even have useful srcloc for the entire composed id (it doesn't appear in the source), much less the pieces.]

greghendershott:
p.s. I had written:

> On the one hand this feels like a property with more general use -- whenever you might want to reveal the parts of an identifier, without also the "draw arrows!" directive. OTOH I don't now know of other uses cases, so maybe it's best to keep the name so specific.

Maybe let's forget I said that? It's above my pay grade to speculate whether something like "`sub-range-binders' without the draw-arrow! directive" could be generally useful, and what would be involved to make it work properly in all cases. I have no idea what I'm talking about. That isn't the motivation of this PR.

Instead:

The syntax property key in the commit is named `'import-or-export-prefix-ranges`. It's not general, it's super specifically about `prefix-in` and `prefix-out`. It's only used "at the boundary" of files, with imports and exports, where e.g. check-syntax and identifier-binding say identifier involved is imported or exported -- never lexical. All I need is a way to recover the source location of its pieces.

That's my proposal, period.

rfindler:
Thanks for the helpful explanation and taking the time so I get what's going on! As for the hope/idea that the property could be more useful, I think that's great and am happy to see those kinds of thoughts if for no other reason than to help me understand where you're coming from (but possibly much more)!

As for the existing explanation, I am getting that the big picture idea that you're closing the gap between what check syntax's arrows have and what you need to find the occurrences (going both directions -- that makes total sense). I tried to work through an example to see if I understand this [*]. I'm not sure this is a representative example, so let me know if doesn't exhibit some interesting complexity.

tmp.rkt:
```
#lang racket/base
(define c 42)
(provide (prefix-out P: c))
```

tmp2.rkt
```
#lang racket/base
(require "tmp.rkt"
         (prefix-in OUTER: (prefix-in INNER: "tmp.rkt")))
OUTER:INNER:P:c
```

In this program, there seems to be two interesting multi-file rename options, namely the `P:` and `c` in tmp.rkt and one reference to each, namely in the `OUTER:INNER:P:c` in tmp2.rkt

- `c`: for this one we can see that the reference `OUTER:INNER:P:c` has an identifier binding with `c` as the "from-sym" and `P:c` as the "nominal from sym". If we make the assumption that the only way that identifiers are constructed is by adding prefixes, then it seems like we can find the `c` at the reference to do the renaming.

- `P:`: using the same information and assumption, it seems like we can also find the `P:` inside the identifier.

But the assumption that the only thing we do is prefixing seems somewhat suspicious. So maybe the key information content here is basically saying "the identifier that's imported here, namely `OUTER:INNER:P:c` was, in fact, constructed by concatenating those prefixes onto it", which would let us just avoid renaming when some other transformer did something else different that happens to end up with the same letters somehow? Or maybe I'm drawing some incorrect conclusion based on the `identifier-binding` information? Or maybe something else wrong?

----

[*] -- and I'm doing all this just to make sure that things are the best choices for what we're doing here. It isn't that I doubt what you've done (far from it!) it is just that I don't quite have all the pieces clear in my head yet so I appreciate your patience helping me get things straight. I definitely have already come a fair way in my understanding of the technical details thanks to your explanations but I'm not quite there yet. I do feel like we're going to be locked in to whatever gets decided here so I hope you don't mind me trying to think it all through carefully. In all honesty, I'm totally fine with just merging this, so if you think I'm wasting our time, then we can merge it. That said, however, hopefully the discussion here will help us at least get a good set of docs for the new property!

greghendershott:
That's a good example.

Keep in mind that the prefixes do not necessarily end up as distinct identifiers in the fully expanded code.

They never do for `prefix-out` because it always expands to a `(#%provide [rename local-id exported-id])`, like `[rename c P:c]`. There's no `P:` in the expanded code, and, no way to find its original srcloc (AFAICT).

`prefix-in` might expand to the special `(#%require [prefix prefix-id raw-module-path])` clause, and check-syntax handles that already (which is great). But less simple examples expand to `(#%require [rename raw-module-path local-id imported-id])`, with the same problem.

Furthermore, both those `#%require` and `#%provide` low-level `rename` clauses can be the result of expanding various surface syntax, including `rename-in`, `rename-out`, `only-in`, etc. Sometimes a rename is just rename, said someone like Freud or Turing.

---

In terms of usage, I only look for this proposed new syntax property only on:

- the `exported-id` in a `(#%provide [rename local-id exported-id])`
- the `local-id` in a `(#%require [rename raw-module-path local-id imported-id])`

Those are the identifiers that could be created by one or more prefixes.

greghendershott:
p.s. The only other idea I have, would be to change the grammar of fully-expanded syntax, to generalize the `#%require` `prefix` clause:

- Add a `prefix` clause for `#%provide`.
- Specify that both it and the one for `#%require` take one _or more_ prefixes.

That would have a similar benefit as the existing limited `prefix` clause: Avoiding expanding into a `rename` clause for every import or export (which could be dozens or hundreds).

Maybe it would also be better for langs like rhombus: they can expand to this if they want to support prefixed imports and exports. (Because my changes to the `provide` and `require` transformers do nothing IIUC for those langs.)

On the other hand changing the grammar probably has a bunch of downsides/costs/risks.

rfindler:
Okay, thanks; I've now gotten another piece that I was missing! (thank you!) -- the fact that there is nothing in the original program with `P:` as its source location so you can't even tell that `P:` is something that one might try to rename in the first place!

I think I'm now getting the point, but let me ask to confirm: is this piece of information meant to answer questions of the form "Given an identifier: is this identifier made from a prefixing operation and, if so, what are the precise prefixes?" So if something were the collect the values of all of the syntax properties for a particular module, then one would be able to answer this question? That is, this property isn't bridging a gap in the same way that the arrows connect things: we know where the identifier comes from all along (via `identifier-binding`)?

I guess the idea is that then the property could be stored on any occurrence of an identifier that appears in the module and the tool would then know that any identifier that is `free-identifier=?` to the one that has the property was also constructed with this prefixing? If that's the case, then I think this is (finally!) all making sense to me.

I would have a suggestion: instead of relying on the placement of the property (i.e. requiring it to be in a particular place in the fully expanded code), let it appear on any identifier and then the property would be interpreted as "any identifier that's free-identifier=? to this one was constructed by these prefixes". Or we could go even slightly further and put the identifier itself into the property and then the location of the property would be completely irrelevant. I'm suggesting this because some other language might have some other set of macros that do similar, but slightly different things, where this specific piece of information would cause the renaming to do the right thing for that language but, for some reason, it is difficult to get the syntax property into this particular place. 

Along those lines, I would also suggest the "tree of cons" thing as well, also to slightly future proof things (with zero conses also being fine if there's only one property). It is something that's pretty pervasive with syntax properties so probably won't be any particular burden on anyway. (After all, the [code to deal with it (same as linked before so you don't really need to look)](https://github.com/racket/drracket/blob/master/drracket-tool-text-lib/drracket/private/syncheck/traversals.rkt#L649-L654) isn't too bad.)

---

One issue that occurs to me: identifiers that are renamed versions of other identifiers are going to all be `free-identifier=?` to each other even if they are spelled in dramatically different ways and so this might lead to some strange codepaths inside pdb. I can't think of any way that the information that's going in to the syntax objects in this PR would be insufficient, but it does seem like someone could do something like this on the exporting side:


```
#lang racket/base
(define c 42)
(provide (prefix-out P: c)
         (prefix-out OUTER:INNER: (prefix-out P: c)))
```

and then something like this on the importing side:

```
#lang racket/base
(require "tmp.rkt"
         (prefix-in OUTER: (prefix-in INNER: "tmp.rkt")))
OUTER:INNER:P:c
```

and the `OUTER:INNER:P:c` reference that appears effectively has two different prefixing operations that led to it and we can't tell which one is the "real" one (or, maybe more accurately, both are "the" real one). Not sure how the tool should respond in this case, but in the slightly different situation where the exporting side actually has two prefixes, then probably the tool should rename both prefixes automatically somehow?



----

Also, is this perhaps more general than prefixing? Is this really a "concatenation" identifier construction property? Is it always telling us "this identifier was constructed by concatenating these pieces" so, if some strange language had a suffixing operation for export, then this property could just cause pdb work? Or is there some other place that the prefix part of it matters?

----

Is pdb going to let people rename the prefixes themselves? Like if I'm exporting an identifier with a prefix and I used `prefix-out`, is pdb going to offer the ability to rename the prefix? If so, how does it find the source location of the original prefix itself?

rfindler:
> The only other idea I have, would be to change the grammar of fully-expanded syntax, [ ... ]

Yeah, I agree that's unlikely to fly. :(

greghendershott:
- I don't know how to respond to the ideas about changing to what syntax the property is attached. To me it seems natural to attach it to the identifier it is describing.

   (In that sense you can almost think of it as "adding enhanced multi `prefix` clauses to the fully-expanded grammar... by other means". If you encounter a `rename` clause, you check for this property on that id, to understand what kind of renaming took place. If the new name has srloc and lacks this property, it was a surface rename. Otherwise the prop describes the prefixing that had no choice but to expand to a `rename` clause.)

- I'm not sure how to respond to changing the value to be more like `sub-range-binders`. I guess that's fine because that's how I started. But I distilled it down to just the necessary subset of information. The offset in the full identifier, the span, and the syntax position where the sub piece comes from.  (Including it's symbolic value is not necessary but it's been helpful debugging both the implementation and use; if space is a concern (??) could delete.)

   Also when you "just `cons` everything up", it's pretty messy to the user. Navigating the tree of `cons`s is easy. Instead it's a mess needing to recursively discover prefixes of prefixes (indefinitely). [Like we discussed up thread](https://github.com/racket/racket/pull/4649#issuecomment-1563035377); you seemed to have the opposite preference?

- I'm digesting your example (which is a great example; thanks!!):

> ```
> #lang racket/base
> (define c 42)
> (provide (prefix-out P: c)
>          (prefix-out OUTER:INNER: (prefix-out P: c)))
> ```
> 
> and then something like this on the importing side:
> 
> ```
> #lang racket/base
> (require "tmp.rkt"
>          (prefix-in OUTER: (prefix-in INNER: "tmp.rkt")))
> OUTER:INNER:P:c
> ```
> 
> and the `OUTER:INNER:P:c` reference that appears effectively has two different prefixing operations that led to it and we can't tell which one is the "real" one (or, maybe more accurately, both are "the" real one). Not sure how the tool should respond in this case, but in the slightly different situation where the exporting side actually has two prefixes, then probably the tool should rename both prefixes automatically somehow?

So today, `pdb` relies on check-syntax arrows as the starting point.

check-syntax says that a require arrow exists, from the second `"tmp.rkt` on line 3, to `OUTER:INNER:P:c`. (I double-checked this on an older DrRacket w/o my commit; same in both.) `identifier-binding` says the nominal source of `OUTER:INNER:P:c` here is `P:c` from `tmp.rkt`.

Therefore pdb interprets it as the identifier created by the local `(prefix-in OUTER: (prefix-in INNER: "tmp.rkt"))` in the importing file.

Given that interpretation, for all of `OUTER:`, `INNER:`, `P:` and `c` it gives all the expected sites (and no unexpected sites).

rfindler:
**re: syntax property placement.** I agree that it is natural to place the property on the identifier and I agree that there does not seem to be any immediate use case beyond putting it in that one spot in the fully expanded form. I am just thinking that pdb doesn't really *need* it to be there and it may be the case that putting it in another place may be more convenient 

**re: syntax property's actual value.** That is not the tree interpretation of the value that I have in mind. I agree that somehow a recursive tree indicating the prefixes that was comingled with multiple properties would be a big mess.

I'm saying that if the value attached to the property is a tree of cons's whose leaves match the contract: 
```
(vector/c
 identifier?
 (listof (vector natural?
                 natural?
                 symbol?
                 srcloc?)))
```
then the property is self-contained and moving it to another place won't matter (ie the placement of the property would not play a role in the interpretation of the property) and that this might matter for some future refactoring of the provide/require macros or some other language that doesn't for some reason, expand into a fully expanded require/provide but still wants to provide this information. Additionally, if this is making sense, we would probably also want to add a method to `syncheck-annotations<%>` to share this information with tools instead of making someone else write yet another "loop over the fully expanded form". Naturally, I'm happy to make this change (and it won't take me long, I don't think).

Of course, as you are probably thinking, there isn't a definite usecase for this now. So we can always do it later, if one appears. I'm merely saying that setting up the APIs in this way seems to have worked well and it will also provide some consistency.

greghendershott:
OK I understand what you're proposing" Including the full identifier in the value, means that someday the property could be attached to _any_ piece of syntax, and the value would still work.

Having said that, down in `reqprov.rkt` I'd like to attach it directly to the `local-id` in `rename` clauses. And document that as the only place (for now) that tools should expect to find it. Is that OK?

greghendershott:
p.s. I think the contract will need to allow `srcloc` to be `#f` -- or at least for all its members except `source` to be false (no line/col/pos/span).

The reason why is edge cases like `(prefix-out PRE: (all-from-out racket/base))`. None of the "suffixes" have original srcloc.

I think it's best still to include them in the list of "parts", with the false srcloc, because it's still useful to know the offset/span into the full identifier. The client (e.g. pdb) can then deal with the suffix in the same way it deals with `all-from-out` generally; chasing the re-provide transitively.

rfindler:
**re only putting the property in only one place:** okay by me. Documenting it appears only there with a plan to possibly allow it in other places at first worried me because of backwards compatibility but I think that was a wrong worry. We can update the tools and only tools will still not crash if they don't find it, which seems good enough. (That is, the tools need to not crash currently when they don't find it, as someone can write a macro that removes the information or does strange things to it.)

**re the `srcloc?` contract:** while `srcloc?` does not accept `#f`, it does accept `(srcloc #f #f #f #f #f)`. In general, I think these tools don't get to assume that srclocs have meaningful information, even when they are srclocs. Like they could be totally bizarre numbers in there when someone writes a macro that munges the source locations, and we'd want the tool to just ignore such things instead of crashing.

rfindler:
I guess my knee-jerk reaction to the property change is that identifiers can have a lot of stuff in them and so I would ask what rule that stuff is going to play here. More concretely, are the identifiers in the vector ever going to be passed to `free-identifier=?` or something like that?

But as for streamlining the value, it does occur to me that there is some redundancy there because a srcloc has a span in it. Maybe the overall contract should be:

```
(vector/c
 identifier?  ;the full identifier composed from one or more others...
 (listof (cons/c string?     ; the characters in this subpiece
                 srcloc?)))  ; the source location of this subpiece
```

Or, actually, now that I'm really thinking (yes, you can probably see the smoke behind my ears right now), what is the point of the source location at all? Is the outer list (that orders the pieces) and the characters in the string sufficient? Could it just be:

```
(vector/c identifier? (listof string?))
```

Also, btw, it may make sense to make it *actually* be:

```
(vector/c identifier? (listof (and/c string? immutable?)) #:immutable #t)
```

That is, we probably want to add `#:immutable #t` to the end of any `vector/c`s that end up in the final answer. Two reasons come to mind: that contract is a `flat-contract?`, so it can be used as a predicate in the implementation that fishes out the properties, which is handy, and we don't really want anyone to mutate these things or tools to have to guard themselves against that sort of malfeasance. (The immutable on the string only benefits from one of the reasons, FWIW.)

greghendershott:
> Or, actually, now that I'm really thinking (yes, you can probably see the smoke behind my ears right now), what is the point of the source location at all? Is the outer list (that orders the pieces) and the characters in the string sufficient? Could it just be:
>
>  (vector/c identifier? (listof string?))

No it couldn't. I need the position (if any), as I explained [before](https://github.com/racket/racket/pull/4649#issuecomment-1573967934). Knowing the source position is (for me) the entire point of this property for purposes of doing multi-file renaming. It's part of knowing the sites to be renamed.

---

For each segment of the identifier composed from other identifiers, the minimum information needed (by me) is:

1. The offset into the full id.
2. The span from that offset.

i.e. what segment of the full identifier are we talking about.

3. The syntax-position (possibly `#f`) of the identifier that contributed this segment

Although not necessary, I found it helpful to include

4. The syntax-e of of the identifier that contributed this segment

And I had distilled it to those 4 values in my previous version:

```
  ;; The property value is a list of vectors, each of which says that
  ;; a sub-range of the full identifier originated from another
  ;; identifier whose syntax-e and syntax-position is given (the span
  ;; is the same).
  ;;
  ;;   (vector offset ;into the full identifier: non-neg integer
  ;;           span   ;from offset: pos integer
  ;;           sub-syntax-e         ;for convenience: symbol
  ;;           sub-syntax-position) ;pos integer
  ;;
  ;; Example: The value for "foobarbang" formed from "foo" at
  ;; syntax-position 42, "bar" at syntax-position 99, and "bang" at
  ;; syntax-position 777, would be:
  ;;
  ;;   (list (vector 0 3 'foo   42)
  ;;         (vector 3 3 'bar   99)
  ;;         (vector 6 4 'bang 777))
```

(I understand/agree we're now also including the full identifier prop value. I'm quoting from an earlier version; only talking about the list of segment vectors, here. How to rep the minimum needed info for each.)

I really don't care strongly whether the 3rd and 4th values are given individually, or conveyed via a single syntax object. I see pros and cons, each way. Please decide either way, I'll do it.

Thanks.

greghendershott:
> I really don't care strongly whether the 3rd and 4th values are given individually, or conveyed via a single syntax object. I see pros and cons, each way. Please decide either way, I'll do it.

To be clear, the other way would be something like this:

```diff
@@ -1357,19 +1357,14 @@
   ;; final suffix part.
   ;;
   ;;   (vector/c
-  ;;    identifier?  ;the full identifier ^1 composed from others...
-  ;;    (listof (vector/c natural?       ;from this *offset* in the full id
-  ;;                      natural?       ;and up to this *span* from the offset
-  ;;                      identifier?))) ;originates from this identifier ^2
-  ;;
-  ;; ^1: The full identifier syntax will have no useful srcloc because
-  ;; it does not appear in the original source.
-  ;;
-  ;; ^2: Generally the only meaningful values to take from this sub
-  ;; identifier syntax are syntax-e and syntax-position (syntax-span
-  ;; is the same as the other span value). However note that
-  ;; syntax-position can be #f when the identifier is not original in
-  ;; the source.
+  ;;    identifier?  ;An identifier composed from these other identifiers:
+  ;;    (listof
+  ;;     (vector/c ;; A segment of the full idenitifer
+  ;;               natural? ;from this offset in the full id
+  ;;               natural? ;and up to this span from the offset
+  ;;               ;; Originates from another identifier with
+  ;;               symbol?               ;this syntax-e
+  ;;               (or/c natural? #f)))) ;this syntax-position
   ;;
   ;; Example: The value for #'foobarbang formed from #'foo at
   ;; syntax-position 42, #'bar at syntax-position 99, and #'bang at
@@ -1377,9 +1372,9 @@
   ;;
   ;;   (vector
   ;;    #'foobarbang
-  ;;    (list (vector 0 3 #'foo)    ;syntax-position  42
-  ;;          (vector 3 3 #'bar)    ;                 99
-  ;;          (vector 6 4 #'bang))) ;                777
+  ;;    (list (vector 0 3 'foo 42)
+  ;;          (vector 3 3 'bar 99)
+  ;;          (vector 6 4 'bang 777)))
   ;;
   ;; Note: Although this may sound similar to the sub-range-binders
   ;; property, that is intended as a directive to drracket
@@ -1406,7 +1401,8 @@
          (let* ([prop-key     'import-or-export-prefix-ranges]
                 [prefix-vec   (vector 0
                                       (syntax-span prefix-id)
-                                      (syntax-local-introduce prefix-id))]
+                                      (syntax-e prefix-id)
+                                      (syntax-position prefix-id))]
                 [suffix-vecs  (cond
                                 ;; When suffix-id has the prop, discard
                                 ;; its full id. Retain its list of sub
@@ -1422,18 +1418,20 @@
                                                             (vector-ref v 0)))
                                             v)
                                           old-prop-vecs)))]
-                                ;; Base case. We include suffix-id even
-                                ;; when it lacks srcloc (as with e.g.
-                                ;; all-from-xxx). The offset/span into
-                                ;; the full id are still useful. Up to
-                                ;; client to check for #f srcloc.
+                                ;; Base case. We include suffix-id
+                                ;; even when its syntax-position is #f
+                                ;; (as with e.g. all-from-xxx). The
+                                ;; offset/span into the full id are
+                                ;; still useful. Up to client to check
+                                ;; for #f srcloc.
                                 [else
                                  (list
                                   (vector (syntax-span prefix-id)
                                           (string-length
                                            (symbol->string
                                             (syntax-e suffix-id)))
-                                          (syntax-local-introduce suffix-id)))])]
+                                          (syntax-e suffix-id)
+                                          (syntax-position suffix-id)))])]
                 [new-vecs     (cons prefix-vec suffix-vecs)]
                 [new-prop-val (vector (syntax-local-introduce full-id)
                                       new-vecs)])
```

rfindler:
Looking at the diff, if I'm getting things right, I think you're right that including the identifier is best, as the identifier is actually there in the original source so it might end up being useful in some other way that it's present purpose. 

And sorry for asking again about the prefix identifier. Duh. 😄

greghendershott:
Thanks for confirming. I'll work on updating some Scribble docs for the syntax property as well as the change to the `export` struct. Then I'll probably remove the "draft" flag from this PR.

mflatt:
@greghendershott: It looks like that last commit removes a `history` note from the reference instead of removing them from the guide. Did you mean to change `guide/module-require.scrbl`, instead?

Also, I notice a small typo: `identifier??` in `stx-trans.scrbl`.

greghendershott:
> @greghendershott: It looks like that last commit removes a `history` note from the reference instead of removing them from the guide. Did you mean to change `guide/module-require.scrbl`, instead?

I'll double-check. Thanks for pointing out!

> Also, I notice a small typo: `identifier??` in `stx-trans.scrbl`.

I'll fix.

(At least I didn't do "indentifer", which seems to be my most frequent typing-too-fast mistake lately. The previous winner was probably "exmaple".)

PR #4650, opened 2023-05-16
add a let to let check syntax know about local member names

rfindler:
This change seems to make this program have the expected arrows:

```
#lang racket

(define-local-member-name the-method-name)
(class object%
  (define/public (the-method-name x) x)
  (super-new))
(λ (x)
  (send x the-method-name 1))
```

But I'm not at all sure about the `#f`s in the first argument to `datum->syntax`.

Unfortunately, it runs afoul of typed racket:

```
<pkgs>/typed-racket-lib/typed-racket/base-env/class-prims.rkt:375:7: let-values: expected the identifier `#%plain-app'
  at: let-values
  in: (let-values (((get-plot-bounds dc->plot plot->dc plane-vector get-plot-metrics-functions) (#%app values (lambda () (quote (declare-this-escapes)) (let-values () (#%expression (lambda () (let-values (((get-plot-bounds) (quote #f))) get-plot-bounds))) (#%...
  compilation context...:
   /Users/robby/git/exp/plt/racket/share/pkgs/plot-lib/plot/private/common/plotmetrics.rkt
  location...:
   /Users/robby/git/exp/plt/racket/collects/racket/private/qq-and-or.rkt:193:51
  context...:
   /Users/robby/git/exp/plt/racket/collects/syntax/parse/private/runtime-report.rkt:725:0: error/report
   /Users/robby/git/exp/plt/racket/collects/syntax/parse/private/runtime-report.rkt:24:0: call-current-failure-handler
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt:332:0: do-check
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/check-class-unit.rkt:287:0: check-class
   .../private/runtime.rkt:89:23: fail-handler114
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt:112:0: tc-expr/check
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt:106:0: tc-lambda-body
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt:521:0: check-mono-lambda/expected
   .../match/compiler.rkt:559:40: f714
   .../match/compiler.rkt:559:40: f748
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-lambda-unit.rkt:784:0: tc/lambda
   .../private/runtime.rkt:89:23: fail-handler114
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt:112:0: tc-expr/check
   .../private/parse.rkt:1018:13: dots-loop
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-toplevel.rkt:546:21
   /Users/robby/git/exp/plt/racket/share/pkgs/typed-racket-lib/typed-racket/typecheck/tc-expr-unit.rkt:416:6: loop
   ...
```

Wondering if maybe @samth might be interested to have a look?


samth:
That's not code that I know well, unfortunately. What's going on, I think, is that it's looking for the remembered type information and finding something that it doesn't expect. But it's not obvious from the error what the relevant syntax class or pattern is that would need to change.

Is it really necessary to add all that additional dead code, though? Why doesn't something like `'disappeared-use` do the right thing here?

rfindler:
`'disappeared-use` might work. I am running a build using lifting that seems to be working. I'll let it finish and then update.

rfindler:
With the new commit, `raco setup` gets through the zo-building phase without errors. In addition to lifting, now it just always uses the name `x`, as the previous version looks like it was running into problems with import prefixing.

rfindler:
Ah-- one problem with this commit is that imported bindings defined with `define-local-member-name` don't end up counting, eg if you have this in tmp2.rkt:

```
#lang racket
(define-local-member-name the-method-name)
```

then this doesn't get an arrow from the require

```
#lang racket
(require "tmp2.rkt")

(class object%
  (define/public (the-method-name x) x)
  (super-new))
(λ (x)
  (send x the-method-name 1))
```

mflatt:
I'm worried about the extra code from lifting, too. It's worse than just generating dead code, because lifting creates a definition to hold the lifted expression, so it can't be discarded. Changing the lifted expression to `(begin (lambda ...) (void))` would help, in that the `lambda` can get dropped, but there will still be useless definitions slowing things down a little.

rfindler:
Changing the `binding` function to just this:

```
(define (binding from to stx)
  (syntax-property
   stx
   'disappeared-use
   (syntax-local-introduce to)))
```

actually fixes the import problem for uses of `define-local-member-name`s when they are in `send`s. But the ones in method definitions do not work (at all) with this change. I'm not entirely sure, but by the time check syntax gets ahold of the identifiers, it does see a reference at the method definition site, but it doesn't see the `define-local-member-name` as its binder, it sees it as something like what you might get from a macro like this one:

```
(define-syntax-rule (m x) (let ([x 1]) x))
(m the-id)
```

where `the-id` appears to bind itself. That is, the binding occurrence of the method name itself seems to be getting in the way of the connection between the method name and the `define-local-member-name` binder.

More concretely, in this program:

```
1: #lang racket
2: (define-local-member-name the-method-name)
3: (class object%
4:  (define/public (the-method-name x) x)
5:  (super-new))
```

I do see the occurrence on line 4 showing up as a reference, but its only binding occurrences are also only line 4, not on line 2 (I'm using a `free-id-table` where they keys are the identifiers in binding positions and the values are lists of the actual binders that were found in the program; I do see separate entries for both the line 2 and line 4 occurrence of `the-method-name`.).

When I look at the fully expanded form, it looks like the occurrence for line 4 that's being inserted by calling `binder` ends up outside the binding occurrence of `the-method-name` in the fully expanded form for the method procedure itself, so I'm puzzled on what's going on.

mflatt:
The extra binding comes from the fact that
```
(define/public (the-method-name x) x)
```
is first local-expanded in an internal definition context. Expansion exposes `the-method-name`, which is discovered to be a local name. But, then `the-method-name` is marked as bound in the internal-definition context — which is a side effect that changes the binding of the exposed `the-method-name`. This side effect is why the class implementation has a `localized-map` table to record the pre-effect renaming for use by `lookup-localize`.

One solution is to remove the internal-definition context's scopes on the identifier that is registered with `'disappeared-use`. The commit https://github.com/mflatt/racket/commit/29a7590b62ac70fa0408a4ab58439a0795a71a4f illustrates one way to thread that through, but I didn't test it beyond trying it on your example.

rfindler:
This seems to be passing in CI; also the `racket-test-core/tests/racket/object.rktl` tests pass; and `raco setup` completes and drracket starts up fine. It also seems to do the right things on the small examples I was playing around with (which is now in a test case [here](https://github.com/racket/drracket/pull/615)). This one doesn't seem to have issues with module imports so I didn't make a test case where the `define-local-member-name` was in a different file but I could if that was felt to be important (the test suite doesn't seem to be set up particularly well for that but it is doable).

rfindler:
Ah, thought of another example that, as it turns out, doesn't yet work.

rfindler:
Here's the example:

```
#lang racket
(define-local-member-name the-method-name)
(define c%
  (class object%
    (define/public (the-method-name n)
      (printf "n ~a\n" n)
      (unless (zero? n)
        (if (even? n)
            (send this the-method-name (- n 1))
            (the-method-name (- n 1)))))
    (super-new)))

(send (new c%) the-method-name 4)
```

The two calls to `the-method-name` in the body of `the-method-name` don't get arrows and, it appears, also don't get calls to `binding`.

mflatt:
The direction in the latest commit seems ok, but I do think that pulling apart the syntax object would be better than using `syntax-local-lift-expression`. If I'm following correctly, `do-localize` would be pulling apart an expression that it was responsible for constructing, (in the previous `cond` clause), so it seems reasonable to recognize that pattern.

rfindler:
I wrote, in an earlier WIP commit:
> There are still problems with this commit, however:

I think that all of these are now addressed, as follows.

> - I'm really not sure of the overall strategy of going back up through
>   the binding of the method to the define-local-member-name
>   binding. It seems a bit baroque.

Well, not this one. There might be a better strategy.

> - the `syntax-local-lift-expression` that puts the syntax property on
>   `(void)` is undesirable, as discussed earlier in the pull request
>   (because it creates extra definitions) but that case in `do-localize`
>   does not have easy access to an expression (unlike the case above
>   it). Is it even reasonable to think about pulling the syntax object
>   apart and sticking another property inside?

This is changed and with a note in the code why it is not unacceptable.

> - we don't have a direct arrow to the regular use of the method in the
>   body (the else branch of the `if` expression)

These arrows are now there.

>   maybe this is okay as we have a "transitive" arrow from the
>   `define-local-member-name` binder to the header of the method and
>   then from there to the invocation, but renaming doesn't currently
>   follow all those arrows so if we wanted to keep this, we'd probably
>   want to change the graph traversal part of the renaming algorithm

This arrow is now there---no need to change the renaming algorithm.

> - (minor) the definition of `method-name-without-def-ctx` duplicates
>   some of the definition of `method-name`

I think this was okay, after all.

> - if this approach is a good one, then all of the `#f`'s being passed
>   to `mk-set!-trans` need fixing

These are all fixed now or else are not places where we can have a `define-local-member-name`-introduced binding.

rfindler:
Okay, I've cleaned up the history. I'll wait and see what CI thinks and if anyone has any comments but I currently don't have any of my own ideas on what more to do here.

samth:
I wonder if this code suggests a different way to communicate to check syntax. Could we put the internal-definition scope into the syntax property itself (or in a different syntax property), and then remove it in the consuming code in check syntax, which would reduce the need to plumb everything through the code?

mfelleisen:
Michael’s `syntax-spec` stores away such information at the top-level and provides support for automatically doing this. [I had Michael present the basic ideas to Robby. Perhaps they can be useful after all.]

rfindler:
re @mfelleisen 's comment, if I'm getting the reference to @michaelballantyne 's work correctly, then yes, I think that would be great. It might even allow us to factor out the check-syntax issues entirely. It would seem to be a reimplementation or at least a non-trivial porting effort, however. But perhaps that would be a good thing as we could use really use class chaperones that don't copy method vectors. And it sure seems like it would be a significant validation of the ideas, if it works well.

-----

re @samth 's comment, my initial question is: how would check syntax know which definition context scopes to connect to which disappeared uses? 

Trying to go in a bit more depth, the way that the macro is working is that `define-local-member-name` defines a private piece of information and that information is used to synthesize an identifier (symbol?) that no one else has which is used to name the method. The trouble is that the private identifier is what is actually bound by `define/public` so by the time we get to a reference to the method, there is no need for the main implementation to consult the information stashed in the `define-local-member-name` definition, so it doesn't. Check syntax, however, wants to draw an arrow from each of those references back up to the `define-local-member-name` so it has to get at that information somehow.

That said, my guess is that someone who understands the structure of the class system implementation could probably have made a change (possibly with a larger diff) that would have resulted in clearer code overall. And then maybe the plumbing I introduced just wouldn't be there in that version of the code.

As for the larger question of a reusable pattern inspired by this that might warrant support some additional support from check syntax, I don't quite see that yet but would welcome your thoughts about how it might go.

The `define-local-member-name` introduces a binder for the name but so does the method definition and so we end up with a fairly funky bound->binding graph, e.g. in this program:
```
#lang racket
(define-local-member-name m)
(define c%
  (class object%
    (define/public (m x)
      (unless (zero? x)
        (m (- x 1))))))
```

we get a dag where there is an edge from each `m` every other `m` that is further down (the joining together in the dag shows up at the last `m`, which has two incoming edges). Redex also gets some crazy arrows because the non-terminal names themselves get arrows but so do local binders. But it doesn't seem too common to have this kind of thing happening and Redex, IIRC uses a different strategy, without disappeared-use (but instead making up little lets in dead code) but it has been a while since I looked at that code.

rfindler:
This example also works!

```
#lang racket

(define o
  (let ()
    (define-local-member-name the-method-name)
    (define %
      (class* object% ()
        (define/public (the-method-name x)
          x)
        (define/public (another-method)
          (class object%
            (define/public (the-method-name x)
              x)
            (super-new)))
        (super-new)))
    (send (new %) another-method)))

(send o the-method-name)
```


As does a similar one where the inner class is inside the outer class's method named `the-method-name` (although in that case, I think one could argue that lots of different behavior is correct).

rfindler:
Ah, rats. This is the interesting test case, actually. It is the calls that were troublesome before, too, not the method headers. And this one, alas, doesn't pass.

```
(module m racket
  (define-local-member-name the-method-name)
  (define %
    (class* object% ()
      (define/public (the-method-name x)
        (class object%
          (define/public (the-method-name x)
            (the-method-name x)
            (send this the-method-name x))
          (super-new)))
      (define/public (another-method)
        (class object%
          (define/public (the-method-name x)
            (the-method-name x)
            (send this the-method-name x))
          (super-new)))
      (super-new))))
```

rfindler:
> Ah, rats. [ ... ]

That wasn't too bad to fix, actually! Just have to loop back up in `build-disappeared-use-of-local-member-name`.

Issue #4664, opened 2023-06-06
equal? in bit-vector fails


Metaxal:
**What version of Racket are you using?**
Welcome to DrRacket, version 8.9.0.4--2023-05-09(a8eb566337/a) [cs].

There's a bug in the `equal?` comparison for bit-vectors. See the penultimate line:
```
#lang racket
(require data/bit-vector)

(print-boolean-long-form #t)
(define size 10)

(define bv (make-bit-vector size))
(define bv2 (bit-vector-copy bv))

(equal? bv bv2) ; true
(equal? (bit-vector->list bv) (bit-vector->list bv2)) ; true
 
(bit-vector-set! bv2 4 #true)
(equal? bv bv2) ; true (wrong!)
(equal? (bit-vector->list bv) (bit-vector->list bv2)) ; false
```

The issue is due to an additional `sub1` that shouldn't be there:
https://github.com/racket/racket/blob/c1efe916e992f3be071f50d0ef782cb1fa677984/racket/collects/data/bit-vector.rkt#L222

While this is easy to fix, I have 2 other concerns:
* Minor: I don't understand the rationale for this line:
  https://github.com/racket/racket/blob/c1efe916e992f3be071f50d0ef782cb1fa677984/racket/collects/data/bit-vector.rkt#L226
  Why isn't it  instead:
```racket
(let ([used-bits (- nx (* 8 (sub1 (bytes-length vx))))]) ...)
```

* Major: Equality as currently implemented is quite slow. This is an issue when using bit vector with a best-first search for example, where fast comparisons are critical. It would be much faster if one could use directly `equal?` from `bytes` instead. I believe this should be possible if some care is taken in `make-bit-vector` to always fill the unused bits with 0s. Can you confirm this?

(cc @soegaard )

soegaard:
I can confirm there is a bug with `equal?`.

I believe your plan of zeroing out the last bits of the byte vector will improve the speed - and at the same time make the code for `equal?` simpler.

As for `(min 8 (remainder nx 256)` does look wrong -- I can't remember where it came from.  
Can you remember it @rmculpepper ?

Anyways, feel free to change both it and the way `equal?` works.

Metaxal:
Where are the tests for `bit-vector`? I can't find any in the main distribution

soegaard:
I am afraid there are none.

soegaard:
Good news. I decided to check rather than rely on memory.
The tests are in the `data` repo:

[...]

PR #4666, opened 2023-06-08
Simplify bit-vector

Metaxal:
equal and hash-equal are deferred to the underlying bytes, which should be more efficient and less error-prone. This PR also fixes a bug in the original code and
another potential bug, see #4664.

Tests are in racket/data (to come).

Documentation does not change.


mflatt:
Thanks!

soegaard:
Elegant solution!

PR #4668, opened 2023-06-10
Fix typos

sschwarzer:
### Checklist

- [x] documentation

### Description of change

I removed a few duplicated words in the documentation.

In the second commit, I removed a word "and" which doesn't seem to belong there. I hope the resulting sentence is correct. If not, please change it as appropriate. I can remove the last commit and force-push if necessary.


mflatt:
Thanks!

PR #4672, opened 2023-06-15
Update contracts.scrbl: blame-name -> blame-value

bennn:
@rfindler is this right?


rfindler:
Yes, this looks right to me.

Issue #4674, opened 2023-06-19
Poor match error message


sorawee:
**What version of Racket are you using?**
8.9 [cs]

**What program did you run?**

```
(match 1
  [1 (=> exit)])
```

**What should have happened?**

An error saying that there should be a _match clause_ body.

**If you got an error message, please include it here.**

```
let: bad syntax (missing binding pairs or body)
  in: (let ())
```

**Please include any other relevant details**
N/A

sorawee:
Fixed by #4676

PR #4676, opened 2023-06-19
match: add #:do

### LINKED ISSUE #4591: [Feature Request] add a #:do clause for each match clause
This is similar to `#:do` in `for`. It allows us to write something like:

```
   #:do [(define options (or options-one options-any))]
   #:when options
   (define mode (if options-one 'oneOf 'anyOf))
```

Otherwise, we need to write:

```
   #:when (or options-one options-any)
   (define options (or options-one options-any))
   (define mode (if options-one 'oneOf 'anyOf))
```

which repeats the `(or options-one options-any)` evaluation twice.

(Or use mutation, etc)

### LINKED ISSUE #4675: match: brittle =>
8.9 [cs]

**What program did you run?**

```
(define (=> . xs)
  xs)

(match 1
  [1 (=> 1 2)])

(match 1
  [1 (=>)])
```

Compare this against:

```
(define (=> . xs)
  xs)

(match 1
  [1 (=> 1)])
```

I believe all of them should result in a compile-time error. Right now, the first program works while the second program does not. This seems weirdly inconsistent.

N/A

N/A

sorawee:
This PR adds a support of `#:do` in match clauses and generalizes the existing match clause options.

Currently, match clause option is either:

1. None
2. `(=> id)`
3. `#:when cond-expr`

With this PR, the option becomes: `option=> option ...` where `option=>` is either:

1. None
2. `(=> id)`

and `option` is either:

1. `#:when cond-expr`
2. `#:do [do-body ...]` 

In particular, it allows `#:when` and `#:do` to appear multiple times.

All match programs prior this PR continue to work under this PR.

## Examples

```
> (define (m x)
    (match x
      [(list a b c)
       #:do [(define sum (+ a b c))]
       #:when (> sum 6)
       (format "the sum, which is ~a, is greater than 6" sum)]
      [(list a b c) 'sum-is-not-greater-than-six]))
> (m '(1 2 3))
'sum-is-not-greater-than-six
> (m '(2 3 4))
"the sum, which is 9, is greater than 6"
```

Resolves #4591 and #4674. But it does not resolve #4675, as it's unclear what should be done.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -55,7 +55,7 @@ a lower-level mechanism achieving the same ends.
 (m '(2 3 4))
 ]
 
-An optional @racket[#:do [do-body ...]] simply executes @racket[do-body] forms.
+An optional @racket[#:do [do-body ...]] executes @racket[do-body] forms.
 In particular, the forms may introduce definitions that are visible in the remaining
 options and the main clause body.
 Both @racket[#:when] and @racket[#:do] options may appear multiple times
@@ -74,7 +74,7 @@ Both @racket[#:when] and @racket[#:do] options may appear multiple times
 (m '(2 3 4))
 ]
 
-An optional @racket[(=> id)], which must appear right after @racket[pat],
+An optional @racket[(=> id)], which must appear immediately after @racket[pat],
 is bound to a @deftech{failure procedure} of zero
 arguments.
 @racket[id] is visible in all clause options and the clause body.
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -17,9 +17,11 @@ on regular-expression matching on strings, bytes, and streams.
 @note-lib[racket/match #:use-sources (racket/match)]
 
 @defform/subs[(match val-expr clause ...)
-              ([clause [pat body ...+]
-                       [pat (=> id) body ...+]
-                       [pat #:when cond-expr body ...+]])]{
+              ([clause [pat option=> option ... body ...+]]
+               [option=> (code:line)
+                         (=> id)]
+               [option (code:line #:when cond-expr)
+                       (code:line #:do [do-body ...])])]{
 
 Finds the first @racket[pat] that matches the result of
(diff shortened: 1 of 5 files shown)


pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -53,9 +55,30 @@ a lower-level mechanism achieving the same ends.
 (m '(2 3 4))
 ]
 
-An optional @racket[(=> id)] between a @racket[pat] and the
-@racket[body]s is bound to a @deftech{failure procedure} of zero
-arguments.  If this procedure is invoked, it escapes back to the
+An optional @racket[#:do [do-body ...]] simply executes @racket[do-body] forms.
```
samth, line 58: I don't think "simply" is useful here.

pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -53,9 +55,30 @@ a lower-level mechanism achieving the same ends.
...
+     #:do [(define sum (+ a b c))]
+     #:when (> sum 6)
+     (format "the sum, which is ~a, is greater than 6" sum)]
+    [(list a b c) 'sum-is-not-greater-than-six]))
+
+(m '(1 2 3))
+(m '(2 3 4))
+]
+
+An optional @racket[(=> id)], which must appear right after @racket[pat],
```
samth, line 77: "Immediately" is better than "right".

samth (approved):

sorawee:
Can this be merged for 8.10?

jbclements:
This PR is now merged to the main branch. Should we pick it for the v8.10 release, or let people try it on HEAD until the next release, when it will get picked up automatically? My generic preference would be to make sure that all changes like this have a few months, so we can see if they cause problems. @sorawee @samth

sorawee:
> This PR is now merged to the main branch.

Are you sure? I don't see it in https://github.com/racket/racket/commits/master.

Regardless, I am OK with waiting for the next release.

jbclements:
Sorry, I believe you're right and I'm wrong. Based on what appears to me to be @samth approval above, I'm going to merge this to the main branch, but not pick it for v8.10. Thanks!

sorawee:
Thank you!

PR #4680, opened 2023-06-23
fully overwrite build-time `coLLECTs dIRECTORy`

LiberalArtist:
When rewriting the embedded `coLLECTs dIRECTORy`, `coNFIg dIRECTORy`, or `dLl dIRECTORy` in an executable, always null out any unused space so as to reliably overwrite the entire old value.

Previously, we would simply write the new value and null terminators, ignoring the contents of the rest of the buffer, which might contain a suffix of the supposedly-overwritten value. For example, the bytes following `coNFIg dIRECTORy` might end up like this:

```racket
#":../collects\0\0ld-racket-vm-cs-8.8.drv-0/source/racket/collects\0\0*"
```

In particular, preserving build paths this way was one of the problems Debian, ALT, and other distributions faced in building Racket reproducibly. The results @Ancieg shared in https://github.com/racket/racket/issues/4456#issuecomment-1281969897 helped me to find this issue.

This commit centralizes executable rewriting in a new `(submod compiler/private/collects-path set-executable-tag)`. It consistently checks that we do not write past the end of the reserved space, which only sometimes was checked previously.

### DIFF
--- a/racket/collects/compiler/private/collects-path.rkt
+++ b/racket/collects/compiler/private/collects-path.rkt
@@ -4,6 +4,36 @@
 	   check-collects-path
 	   set-collects-path)
 
+  (module set-executable-tag racket/base
+    (provide max-dir-len
+             set-executable-tag)
+    (define max-dir-len 1024) ; also ok for DLL path: 2 * sizeof(wchar_t)
+    (define (set-executable-tag who tag desc file ignore-non-executable? given bs)
+      (unless ((bytes-length bs) . <= . max-dir-len)
+        (raise-arguments-error who (string-append desc " too long")
+                               "given" given))
+      (define-values (i o)
+        (open-input-output-file file #:exists 'update))
+      (dynamic-wind
+       void
+       (lambda ()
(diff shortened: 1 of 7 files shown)


mflatt:
This looks great - thanks!

Do the unchecked boxes in your comment mean that you're still working on the patch? FWIW, I think the `raco exe` tests provide most of the existing testing for these path updates. Those run on DrDr for Linux, and they run on Windows and macOS as part of the snapshot build process.

LiberalArtist:
> Do the unchecked boxes in your comment mean that you're still working on the patch? FWIW, I think the `raco exe` tests provide most of the existing testing for these path updates. Those run on DrDr for Linux, and they run on Windows and macOS as part of the snapshot build process.

No, they were just there in the PR template :) I agree that the main test outcome is not breaking any of the existing usages. I have tested the patch with Guix in addition to CI, and I've inspected the binaries to confirm that there are no more residual path fragments.

mflatt:
Ok, good!

PR #4682, opened 2023-06-25
Fix `->*` where `#:pre` and `#:rest` are present

camoy:
The following program

```rkt
(module m racket
  (provide
   (contract-out
    [f (->* () #:rest (listof integer?) #:pre #t any)]))
  (define (f . args) args))

(require 'm)

(f 1 12)
```

has a cryptic internal error:

```
...ct/private/list.rkt:215:14: arity mismatch;
  expected: 2
  given: 0
  context...:
   /usr/share/racket/collects/racket/contract/private/arrow-higher-order.rkt:83:0: check-pre-cond
   /usr/share/racket/collects/racket/contract/private/arrow-val-first.rkt:486:18
```

I tracked this down to the inconsistent ordering of `#:pre` and `#:rest` components in various parts of the arrow contract implementation. In the above code, the thunk for the `#:pre` is being used instead of the late neg projection of the `#:rest`. I didn't bother to try and make everything consistent, only to flip the order in the one place that mattered to fix this bug.

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/arrow-star.rkt
+++ b/pkgs/racket-test/tests/racket/contract/arrow-star.rkt
@@ -712,6 +712,29 @@
                  'pos 'neg)
      2)))
   
+  (test/neg-blame
+   '->*-pre/post-20
+   '((contract (->* () #:rest list? #:pre #f integer?)
+               (λ args 1)
+               'pos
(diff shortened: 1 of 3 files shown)


camoy:
@rfindler would you be able to review this for 8.10?

PR #4686, opened 2023-06-30
Add `#:indent` to `write-json` [reprise]

LiberalArtist:
This is my attempt to finish up the excellent https://github.com/racket/racket/pull/4300, which @NoahStoryM opened over a year ago, now. It should also close @cwebber's five-year-old (!) https://github.com/racket/racket/pull/2100.

Almost the only thing I've added to https://github.com/racket/racket/pull/4300 is a bunch of tests in e1488ab7c5a6d9c238820b1695effbe520b294b3 that our indentation is consistent with what Node and Python do. These exposed one small issue with the indentation of objects, which I fixed in 701dc4d4f8ccb6918641f1772f83b3fdb6b24a4d.

See also discussion at https://racket.discourse.group/t/json-beautifier/1541/5.


jbclements:
Yikes, reviewing this is ... scary.

My inclination here would be to merge this now but not cherry-pick it to 8.10, so that it gets 3 months of potential use on git head before it goes in a release?

Thoughts?

LiberalArtist:
> My inclination here would be to merge this now but not cherry-pick it to 8.10, so that it gets 3 months of potential use on git head before it goes in a release?

That sounds reasonable to me.

Issue #3948, opened 2021-07-31
rktboot error on aarch64-linux


LiberalArtist:
[According to Guix CI](https://ci.guix.gnu.org/build/847137/details), bootstrapping Raket's Chez Scheme fork fails an error from `string-append`. Here's the tail of the [build log](https://ci.guix.gnu.org/build/847137/log/raw):
```
starting phase `build'
Assuming current directory has Chez Scheme sources
string-append: contract violation
  expected: string?
  given: #f
  argument position: 1st
  other arguments...:
   ".def"
  context...:
   /tmp/guix-build-racket-bootstrap-chez-bootfiles-8.2.drv-0/racket-minimal-8.2-checkout/racket/src/ChezScheme/rktboot/machine-def.rkt:6:0: open-file-with-machine.def-redirect
command "/gnu/store/1m10cjmip18isdxf0bljq733brq2gg9p-racket-minimal-bc-3m-8.2/bin/racket" "rktboot/main.rkt" "--dest" "/gnu/store/xlsnnphcqn1xxp2yqr519jh9sv0ai9xw-racket-bootstrap-chez-bootfiles-8.2" failed with status 1
builder for `/gnu/store/xg8sa6qssr46n5n887bzb5l3cl78fy52-racket-bootstrap-chez-bootfiles-8.2.drv' failed with exit code 1
@ build-failed /gnu/store/xg8sa6qssr46n5n887bzb5l3cl78fy52-racket-bootstrap-chez-bootfiles-8.2.drv - 1 builder for `/gnu/store/xg8sa6qssr46n5n887bzb5l3cl78fy52-racket-bootstrap-chez-bootfiles-8.2.drv' failed with exit code 1
```
The offending `string-append` seems to be here: https://github.com/racket/racket/blob/4f0e1703979d704c4f91b009945c324450acec7a/racket/src/ChezScheme/rktboot/machine-def.rkt#L10

I don't know if `target-machine` being `#f` is indicative of a broader problem inferring the target machine or not. A further complication is that I don't have an aarch64-linux setup myself, though I could try to get something working via QEMU.

I hope this problem isn't specific to Guix, but the build is driven by this Guile code, which I can (hopefully) explain if necessary: https://git.savannah.gnu.org/cgit/guix.git/tree/gnu/packages/racket.scm

garrgravarr:
[...]

PR #4702, opened 2023-07-19
rktboot: improve machene type inference

LiberalArtist:
Related to https://issues.guix.gnu.org/62231
Related to https://github.com/racket/racket/issues/3948

I've just done the obvious thing here and parsed `(system-library-subpath #f)`, but I still feel like I've read and written this code too many times.

In particular, I noticed that RISC-V machine types are defined for the BSDs at the Chez Scheme level:

https://github.com/racket/racket/blob/373bf233ecddee44a14821d272ec7f265a1b8845/racket/src/ChezScheme/s/cmacros.ss#L376-L418

But Rumble only handles RISC-V on Linux:

https://github.com/racket/racket/blob/373bf233ecddee44a14821d272ec7f265a1b8845/racket/src/cs/rumble/system.ss#L54-L65

https://github.com/racket/racket/blob/373bf233ecddee44a14821d272ec7f265a1b8845/racket/src/cs/rumble/system.ss#L112-L113

And `raco cross` doesn't handle RISC-V at all:

https://github.com/racket/raco-cross/blob/01a58ff14ee2ee126cf65897761556501a7197ae/raco-cross-lib/private/cross/platform.rkt#L41-L82

I think it might make sense to implement more of this translation logic in Chez itself, but I haven't been completely satisfied with any of the concrete ideas I've come up with so far.

### DIFF
--- a/racket/src/rktboot/config.rkt
+++ b/racket/src/rktboot/config.rkt
@@ -15,20 +15,69 @@
                                (path->complete-path scheme-dir))))))
 (hash-set! ht 'make-boot-scheme-dir scheme-dir)
 
+(define (infer-target-machine)
+  ;; Compute a native or pbarch machine string for the current platform.
+  ;; Some caveats:
+  ;;  1. A pbarch Racket will always infer a pbarch machine,
+  ;;     even if a native machine exists. Hopefully this is an
+  ;;     unlikely scenario: if you're running Racket CS, you've
+  ;;     bootstrapped Chez somehow, so you could use `re.boot`.
+  ;;  2. A `tpb` or `pb` Racket on a 32-bit platform would still return
+  ;;     64 from `(system-type 'word)`, but, in addition to the above,
+  ;;     it is not currently possible or desired to build Racket as
+  ;;     `tpb` or `pb` (as opposed to pbarch variants):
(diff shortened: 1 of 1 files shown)


racket/src/rktboot/config.rkt
```diff
@@ -15,20 +15,69 @@
...
+      [("arm") "arm32"]
+      [("ppc") "ppc32"]
+      [("riscv64") "rv64"]
+      [("unknown") #f] ; pb machine types
+      [else #f]))
+  (define os
+    (case (cadr arch+os)
+      [("macosx" "darwin" "ios") "osx"]
+      [("win32" "cygwin") "nt"]
+      [("linux" "android") "le"]
```
LiberalArtist, line 61: The even-more-specific results for the `'os*` part are returned only on BC. Given that, I'm not sure about the relative benefits of specificity vs. consistency.

mflatt:
I also merged #4703, which is doesn't seem to be covered here — but i didn't look or think closely, so it's possible that something here means that code is not reached. And along those lines of hopefully making things work better without necessarily making them better implemented, I added `riscv64` <-> `rv64` conversion to `raco-cross`.

PR #4704, opened 2023-07-21
using category symbols not listed behaviour

spdegabrielle:
Documenting behaviour when library uses a category symbol that is not listed.

### DIFF
--- a/pkgs/racket-doc/scribblings/raco/setup.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/setup.scrbl
@@ -651,7 +651,7 @@ Optional @filepath{info.rkt} fields trigger additional actions by
 
    @history[#:changed "6.4" @elem{Allow a category to be a string
                                  instead of a symbol.}
-            #:changed "8.9.x.x" @elem{Add 'drracket-plugin category 
+            #:changed "8.9.0.6" @elem{Add 'drracket-plugin category 
                                       symbol.}]}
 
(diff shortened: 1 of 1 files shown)


sorawee:
Could you modify `@history` below to say that `'drracket-plugin` was added on 8.9.x.x?

spdegabrielle:
> Could you modify `@history` below to say that `'drracket-plugin` was added on 8.9.x.x? (this should have been a part of #4687 -- I just realized it)

PR created: https://github.com/racket/racket/pull/4705

sorawee:
Sorry, I didn't mean that. In the file `pkgs/racket-doc/scribblings/raco/setup.scrbl` on line 652, there was:

```
@history[#:changed "6.4" @elem{Allow a category to be a string
                               instead of a symbol.}]}
```

I propose that we changed that to:

```
@history[#:changed "6.4" @elem{Allow a category to be a string
                               instead of a symbol.}
         #:changed "8.9.0.6" @elem{Add a new category @racket['drracket-plugin].}]}
```

I haven't tested this though.

spdegabrielle:
thanks done now

spdegabrielle:
changed 8.9.x.x to 8.9.0.6

sorawee:
FWIW, I just saw that the documentation for `'library` already mentions this information:

```
@item{@racket['library] : Documentation for libraries; this
           category is the default and used for unrecognized category
symbols.}
```

I still think the placement of the information in this PR makes more sense. It's very easy to miss the information when it's in the middle of a list!

PR #4707, opened 2023-07-22
[ ci ] Run at most one snapshot action at a time

shhyou:
### Description of change

I think it makes sense to generate snapshots only for the latest commit to `master`.
Therefore, this PR cancels any in-progress snapshot actions when a new one is triggered.
However, there are some drawbacks.

- Up sides: no need to wait for earlier snapshot actions to finish when they are going to be superseded by the newer commit
- Down sides: the canceled snapshot actions will send an email. They are also marked as "**✗**" in commit history.

### DIFF
--- a/.github/workflows/ci-snapshot.yml
+++ b/.github/workflows/ci-snapshot.yml
@@ -5,6 +5,13 @@ on:
     branches:
       - master
 
+# Only run at most one snapshot action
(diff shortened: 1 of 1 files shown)


shhyou:
There's currently no way in GHA to avoid the cancellation notification: <https://github.com/orgs/community/discussions/13015>

Issue #4710, opened 2023-07-30
Multiple evaluation of regexp pattern


sorawee:
**What version of Racket are you using?**
8.10 [cs]

**What program did you run?**

```
(match "a"
  [(pregexp (begin (println 'a) "a")) 1])
```

**What should have happened?**

`'a` is printed only once, but it is printed twice.

**If you got an error message, please include it here.**

N/A

**Please include any other relevant details**

N/A

sorawee:
Fixed by #4711

PR #4711, opened 2023-07-30
match: fix pregexp pattern

### LINKED ISSUE #4281: Match documentation and implementation for pregexp are not in agreement
I'm trying to use match and realized that I needed pregexps to get the regexp features I needed.

Example:
```
> (match "12" [(regexp "\\d+") 'matched])
string:1:0
match: no matching clause for "12" [,bt for context]
> (match "12" [(pregexp "\\d+") 'matched])
'matched
```

But during my attempts to get the right code I wrote the following:

```
> (match "12" [(regexp #rx"\\d+") 'matched])
string:1:0
match: no matching clause for "12" [,bt for context]
> (match "12" [(pregexp #rx"\\d+") 'matched])
pregexp: contract violation
  expected: string?
  given: #rx"\\d+"
 [,bt for context]
```

That error message seem strange to me, since I wasn't calling pregexp directly. And so I tried to read the documentation on match.

https://docs.racket-lang.org/reference/match.html

This has examples of `regexp` working with `#rx` literals, and then documents `pregexp` as:

> like the regexp patterns, but if rx-expr produces a string, it is converted to a pattern using pregexp.

But the implementation checks if the expression produces a pregexp, and if not converts it into a pregexp

https://github.com/racket/racket/blob/448b77a6629c68659e1360fbe9f9e1ecea078f9c/racket/collects/racket/match/parse.rkt#L92-L97

My guess is the desired behavior is that  `pregexp? => #f` but `regexp? => #t` values should error out when used with the `pregexp` matcher, but probably with a better error and the documentation should be reworded to not imply that regular regexps would be accepted.

There is also interesting behavior with regards to byte-regexps and how `regexp` accepts them but `pregexp` doesn't even for `byte-pregexp? => #t` values.
```
> (match #"12" [(regexp #rx#"\\d+") 'matched])
string:1:0
match: no matching clause for #"12" [,bt for context]
> (match #"12" [(regexp #px#"\\d+") 'matched])
'matched
> (match #"12" [(pregexp #px#"\\d+") 'matched])
pregexp: contract violation
  expected: string?
  given: #px#"\\d+"
 [,bt for context]
```

sorawee:
This commit fixes two issues with the pregexp pattern.

1. Avoids unintentional expression duplication, which results in multiple evaluations (#4710)
2. Provides better error handling when a regexp value is given to pregexp pattern (#4281), along with better documentation.

Fixes #4281, #4710

### DIFF
--- a/racket/collects/racket/match/parse.rkt
+++ b/racket/collects/racket/match/parse.rkt
@@ -6,7 +6,8 @@
          "patterns.rkt"
          "parse-helper.rkt"
          "parse-quasi.rkt"
-         (for-template (only-in "runtime.rkt" matchable? mlist? mlist->list)
+         (only-in "stxtime.rkt" current-form-name)
+         (for-template (only-in "runtime.rkt" matchable? pregexp-matcher mlist? mlist->list)
                        (only-in racket/unsafe/ops unsafe-vector-ref)
                        racket/base))
 
@@ -91,15 +92,11 @@
      (trans-match #'matchable? #'(lambda (e) (regexp-match r e)) (parse #'p))]
     [(pregexp r)
      (trans-match #'matchable?
-                  (rearm
-                   #'(lambda (e)
-                       (regexp-match (if (pregexp? r) r (pregexp r)) e)))
+                  #`(pregexp-matcher r '#,(current-form-name))
                   (Pred #'values))]
     [(pregexp r p)
      (trans-match #'matchable?
-                  (rearm 
-                   #'(lambda (e)
-                       (regexp-match (if (pregexp? r) r (pregexp r)) e)))
+                  #`(pregexp-matcher r '#,(current-form-name))
                   (rearm+parse #'p))]
     [(box e) (Box (parse #'e))]
     [(vector es ...)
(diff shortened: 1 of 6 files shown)


racket/collects/racket/match/parse.rkt
```diff
@@ -52,6 +53,16 @@
...
           (if (eq? ddk-size #t) 0 ddk-size)))
 
+(define (make-pregexp-matcher r)
+  #`(lambda (e)
+      (let ([r #,r])
+        (regexp-match
+         (cond
+           [(pregexp? r) r]
+           [(string? r) (pregexp r)]
+           [else (raise-argument-error '#,(current-form-name) "(or/c pregexp? string?)" r)])
```
sorawee, line 63: Not sure if `raise-argument-error` is the right thing to use here.

mflatt:
@samth - does this look ok to you?

racket/collects/racket/match/parse.rkt
samth: This seems ok to me.

racket/collects/racket/match/parse.rkt
```diff
@@ -52,6 +53,16 @@
           (rest suffix)
           (if (eq? ddk-size #t) 0 ddk-size)))
 
+(define (make-pregexp-matcher r)
+  #`(lambda (e)
```
samth, line 57: It seems like this function could be defined once and then referenced in the generated code.

sorawee: Fixed.

samth (approved):

PR #4712, opened 2023-08-01
stencil vector doc: correct xref and code highlight

sorawee:
EDITED: jk, it looks like one `syntax-local-introduce` suffices to correct xref. No need for hacks.

Before:
Screenshot 2023-07-31 at 11 12 58 PM

After:

Screenshot 2023-08-01 at 3 08 04 AM

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/stencil-vectors.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/stencil-vectors.scrbl
@@ -13,7 +13,7 @@
                                   (with-syntax ([str str]
                                                 [e (read (open-input-string str))])
                                     #'(eval:alts #,(code str) e)))])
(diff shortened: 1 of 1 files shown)


rfindler:
I think similar kinds of hacks are used to render things elsewhere (maybe like `#true` vs `#t`?) and these could be put into the core document rendering forms. 

It looks like there is no overlap in printed sizes with binary and the regular formatting. This isn't the case with hexidecimal, tho, as `10000000` is `#x989680` and they have the same length. Anyway, that makes it seem safe to me to support this kind of thing only for binary and regular formatting (but not hex).

sorawee:
Well, no need for any hacks! It just needs a `syntax-local-introduce` to get xref working.

PR #4714, opened 2023-08-11
Fix docs for traits

gus-massa:
It looks like an error in the order of the arguments.

Reported by [matteo-daddio](https://racket.discourse.group/u/matteo-daddio) in https://racket.discourse.group/t/traits-documentation-bug/2199


mflatt:
LGTM

PR #4718, opened 2023-08-14
doc: correct the contract for the `new-scope?` argument to `syntax-local-lift-require`

rocketnia:
##### Checklist

- [x] Documentation update

### Description of change

The `new-scope?` argument to `syntax-local-lift-require` currently shows up in the docs as a required argument with a contract of `#t`. This update fixes the docs to show that it's an optional argument with a contract of `any/c` and a default value of `#t`.


sorawee:
Thanks!

PR #4724, opened 2023-08-17
fix the interaction of `#:unprotected-submodule` and `struct`

### LINKED ISSUE #4661: #:unprotected-submodule in contract-out doesn't export the struct type transformer binding
v8.9 [cs]

**What program did you run?**
foo.rkt:
```racket
#lang racket/base

(require racket/contract)

(provide
 (contract-out
  #:unprotected-submodule unchecked
  [struct foo ((field number?))]))

(struct foo (field))
```
bar.rkt:
```racket
#lang racket/base

(require
  racket/match
  (submod "foo.rkt" unchecked))

(define instance (foo 2))

(match instance
  [(foo num)
   (printf "foo with ~a\n" num)])
```
I ran `racket bar.rkt`.

`foo with 2` should be printed.

```
bar.rkt:10:3: match: syntax error in pattern
  in: (foo num)
  location...:
   bar.rkt:10:3
  context...:
   /Applications/Racket v8.9/collects/racket/match/gen-match.rkt:55:11: mk
   /Applications/Racket v8.9/collects/racket/match/gen-match.rkt:24:0: go
   /Applications/Racket v8.9/collects/syntax/wrap-modbeg.rkt:46:4
```

When I replace `(submod "foo.rkt" unchecked)` in bar.rkt with just `"foo.rkt"`, it runs fine.

```zsh
$ uname -a
Darwin sarnas-MacBook-Air.local 22.4.0 Darwin Kernel Version 22.4.0: Mon Mar  6 21:00:41 PST 2023; root:xnu-8796.101.5~3/RELEASE_ARM64_T8103 arm64
```

usaoc:
##### Checklist
- [x] Bugfix
- [x] tests included

### Description of change
Make the `#:unprotected-submodule` submodule export structs as `struct-out` does:

- `tl-code-for-one-id/new-name` is adjusted to accept `upe-id`, which is used as the exported id in the submodule when present;

- The original struct-type name and descriptor are exported in the submodule.

`#:omit-constructor` has no effect on the original struct-type name.  This is intended and documented accordingly.

Fix #4661.

### DIFF
--- a/pkgs/racket-test/tests/racket/contract/contract-out.rkt
+++ b/pkgs/racket-test/tests/racket/contract/contract-out.rkt
@@ -1958,4 +1958,39 @@
                #`'#,(struct-field-info-list (syntax-local-value #'foo))))
       (eval '(extract-field-names)))
    (list 'x))
+
+  (test/spec-passed/result
+   'provide/contract-upe-struct-transformer
+   '(begin
+      (eval '(module provide/contract-upe-struct-transformer-def racket/base
+               (require racket/contract/base)
+               (provide (contract-out
+                         #:unprotected-submodule no-contract
+                         (struct s ([x integer?]))))
+               (struct s (x) #:constructor-name make-s)))
+      (eval '(module provide/contract-upe-struct-transformer-ans racket/base
+               (require racket/match
+                        (submod 'provide/contract-upe-struct-transformer-def no-contract))
+               (provide answer)
+               (define answer (match (make-s 1) [(s x) x]))))
+      (dynamic-require ''provide/contract-upe-struct-transformer-ans 'answer))
+   1)
+
+  (test/spec-passed/result
+   'provide/contract-upe-struct-type-descriptor
+   '(begin
+      (eval '(module provide/contract-upe-struct-type-descriptor-def racket/base
+               (require racket/contract/base)
+               (provide (contract-out
+                         #:unprotected-submodule no-contract
+                         (struct s ([x integer?]))))
+               (struct s (x))))
+      (eval '(module provide/contract-upe-struct-type-descriptor-ans racket/base
+               (require racket/match
+                        (submod 'provide/contract-upe-struct-type-descriptor-def no-contract))
+               (provide answer)
+               (struct t s (y))
+               (define answer (match (t 1 2) [(t x y) (list x y)]))))
+      (dynamic-require ''provide/contract-upe-struct-type-descriptor-ans 'answer))
+   (list 1 2))
   )
--- a/racket/collects/racket/contract/private/provide.rkt
+++ b/racket/collects/racket/contract/private/provide.rkt
@@ -283,6 +283,15 @@
 
 (define-for-syntax current-unprotected-submodule-name (make-parameter #f))
 
+;; make-unprotected-submodule-code
+;; : (-> (-> syntax?) (listof syntax?))
+;; Calls `thunk` if `current-unprotected-submodule-name` is non-#f,
+;; incorporating the resulting syntax object into a singleton list of
+;; `module+` form; otherwise returns empty list.
+(define-for-syntax (make-unprotected-submodule-code thunk)
+  (define upe-submod (current-unprotected-submodule-name))
+  (if upe-submod (list #`(module+ #,upe-submod #,(thunk))) null))
+
 ;; tl-code-for-one-id/new-name : syntax syntax syntax (union syntax #f) -> (values syntax syntax)
 ;; given the syntax for an identifier and a contract,
 ;; builds a begin expression for the entire contract and provide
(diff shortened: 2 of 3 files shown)


usaoc:
Seems like this doesn’t handle `#:constructor-name` properly.  Relatedly, when `#:omit-constructor` is present, the transformer binding is omitted.  I’ll need to use a different approach than changing `tl-code-for-one-id/new-name`.

usaoc:
Update: This PR is ready for review.  The above mentioned issue has been fixed, but do let me know if there’s any further issue.

pkgs/racket-test/tests/racket/contract/contract-out.rkt
```diff
@@ -1958,4 +1958,33 @@
...
+               (require racket/contract/base)
+               (provide (contract-out
+                         #:unprotected-submodule no-contract
+                         (struct s ([x integer?]))))
+               (struct s (x) #:constructor-name make-s)))
+      (eval '(require racket/match
+                      (submod 'provide/contract-upe-struct-transformer no-contract)))
+      (eval '(match (make-s 1) [(s x) x])))
+   1)
+
```
rfindler, line 1975: It is probably better to put this into another module to avoid having stuff defined at the top-level that could potentially mess with other tests.

pkgs/racket-test/tests/racket/contract/contract-out.rkt
```diff
@@ -1958,4 +1958,33 @@
...
+               (require racket/contract/base)
+               (provide (contract-out
+                         #:unprotected-submodule no-contract
+                         (struct s ([x integer?]))))
+               (struct s (x))))
+      (eval '(require racket/match
+                      (submod 'provide/contract-upe-struct-type-descriptor no-contract)))
+      (eval '(struct t s (y)))
+      (eval '(match (t 1 2) [(t x y) (list x y)])))
+   (list 1 2))
```
rfindler, line 1989: Ditto here. To test a result, you can have the module export a definition with a name that has a prefix of `provide/contract-upe-struct-type-descriptor`.

racket/collects/racket/contract/private/provide.rkt
```diff
@@ -291,20 +300,20 @@
...
                                    id-rename-without-source))
+  (define ext-name (or user-rename-id id))
   (with-syntax ([ctrct (syntax-property 
                         (syntax-property
                          ctrct/no-prop
                          'racket/contract:contract-on-boundary
                          (gensym 'provide/contract-boundary))
-                        'inferred-name ex-id)]
-                [external-name (or user-rename-id id)])
+                        'inferred-name ex-id)])
```
rfindler, line 316: Why did you remove `external-name` in favor of `ext-name`? I also see three arrows comes out of that binding in DrRacket but only two of them are changed in the below.

racket/collects/racket/contract/private/provide.rkt
```diff
@@ -317,21 +326,19 @@
...
-                                                (provide (rename-out [#,id external-name]))))
-                                      (list)))
-                             #,@(if provide?
-                                    (list #`(provide (rename-out [#,id-rename external-name])))
+                             #,@(if upe-id
+                                    (append
+                                     (make-unprotected-submodule-code
+                                      (lambda ()
+                                        #`(provide (rename-out [#,upe-id #,ext-name]))))
+                                     (list #`(provide (rename-out [#,id-rename #,ext-name]))))
```
rfindler, line 341: I can't recall exactly what was going on here before, but why isn't there something conditional on `upe-id` here anymore? (That would be the naive change, anyway.)

rfindler:
I've made some comments. Overall, the change seems reasonable. Did you try running the contract test suites? You can run them with `racket -l tests/racket/contract/all`. I'm surprised that the `external-name` didn't trigger an error somewhere. So that's another thing to consider doing (adding a test case that fails because of that). Obviously that's beyond the purview of this issue, so it isn't necessary before merging this.

racket/collects/racket/contract/private/provide.rkt
usaoc: The original logic was totally unconditional: as long as the unprotected submodule exists, the `provide` *in the unprotected submodule* will be generated, even when `provide?` is false.  I’m not sure what the intent was—it’s difficult to tell because the logic was very entangled.  To simplify the matter, I just pun `upe-id` to serve double-duty as both an indicator of whether the `provide`s *in both modules* will be generated, as well as the uncontracted identifier to be exported in the unprotected submodule.  This is why both `provide`s are under the consequent.  On the other hand, `make-unprotected-submodule-code` already accounts for the case when the unprotected submodule is missing (which is probably what you really have in mind).

usaoc:
I’ve pushed an update to address the comments.  Regarding `external-name`, I see that there *are* test cases for `'provide/contract-original-contract`, but those curiously only test the second fields, i.e., syntax objects for the contracts.  Adding more test cases shouldn’t be difficult.

racket/collects/racket/contract/private/provide.rkt
rfindler: Okay, I think that makes sense. Thanks!

rfindler:
Thanks!

PR #4725, opened 2023-08-18
doc: hash doc improvement

sorawee:
- [x] documentation

This PR (intentionally) consists of two commits. 

- The first commit replaces the name `hash` with `hsh`, since naming an argument `hash` makes linking to the function `(hash)` failed. 
- The second commit fixes a contract and strengthen a contract to be consistent with other functions.

Before:

Screenshot 2023-08-18 at 9 31 55 AM

After:

Screenshot 2023-08-18 at 10 13 18 AM

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -813,12 +813,12 @@ the result is @racket[bad-index-v] if provided, otherwise the
 @deftogether[(
 @defproc[(hash-iterate-value [ht hash?]
                              [pos exact-nonnegative-integer?])
-         any]
+         any/c]
 @defproc[#:link-target? #f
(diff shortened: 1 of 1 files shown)


usaoc:
I think `ht` is a more commonly used name for hash tables, at least in the Racket code base.

AlexKnauth (approved):
Looks good to me, but worth noting that other places in the documentation, such as `hash-union` etc. further down in the same page, use `h`, `h0` etc. as argument names.

sorawee:
I like the name `ht` better than `h`, so I changed everything to `ht` (including "further down in the same page" that @AlexKnauth mentioned).

usaoc:
The commit message still refers to the original new name (`hsh`).

sorawee:
Thanks for the feedback! @usaoc and @AlexKnauth

Issue #4737, opened 2023-09-01
make install is not incremental for non-in-place builds


riastradh:
**What version of Racket are you using?**
8.10

**What program did you run?**
```
./configure ...
gmake
gmake install
gmake install
```

**What should have happened?**
Since I didn't rerun configure or change the content of any of the file inputs, `gmake install` should complete almost instantly without doing anything extra.  (I tweaked some things outside Racket, which is what caused an automated process to rerun `gmake install`, but that's all.)

Instead, it seems to spend a long time recompiling all the Scheme code that was installed, even though it had already been compiled and I see no reason why it should have changed.

As a consequence, this makes incremental work on fixing build issues very slow and tedious.

**Please include any other relevant details**
NetBSD 9

samth:
Calling `raco setup` (which is the primary expensive thing `make install` does) is in fact incremental; it detects already compiled files and doesn't recompile them, and similar for documentation. That incrementality is not as fast as one might like, which is why it still takes a while, but it is very much there.

samth:
That does seem like it's recompiling but it shouldn't happen if you already re-ran `raco setup`. Could something be off about the timestamps of content of that file and it's corresponding compiled version?

mflatt:
While `raco setup` is incremental, `make install` for a non-in-place install isn't incremental. That is, the title of the issue is correct).

I forget the details, but I think the "copying" step is implemented by removing things in the target directory first, instead of detecting when files it wants to copy are already in place and/or when ".zo" files in the destination should remain (as opposed to being removed because the corresponding source is no longer there, say).

juhp:
(Just adding that incremental wouldn't help (debugging) rpm builds, since rpmbuild removes all previously staged files in its %install phase anyway, so yeah (I expect it is a non-trivial raco change but) I would also like to see what @riastradh suggests)

mflatt:
There's a long history behind the current `make` versus `make install` split, but this is worth revisiting. Last year's overhaul of the build system (shifting to Zuo) should make a change much easier than previous times I considered making a change.

PR #4743, opened 2023-09-05
move work for `make install` to `make`

mflatt:
This commit is aimed at doing most `raco setup` work in `make` instead of `make install`, at least when configuring with `--prefix`.

The `make` step here performs the ".zo"- and documentation-build work (i.e., `raco setup` step) that `make install` currently performs — but with output confined to the build directory. The built files are then merged into place by `make install`. That way, `make install` can be relatively fast, and the ".zo"- and documentation-build step gets the advantage of being incremental like the rest of `make`. Related to #4737.

This change only applies to `make` as run within a "src" build directory, so it does not affect `make` in the repo's root directory. And since it only applies when `--prefix` is used, it doesn't affect a workflow that uses an in-place build (as typical with a repo checkout).

A new `info-domain-root` configuration option is needed to keep the info-domain step of `raco setup` confined to the build directory.

### DIFF
--- a/racket/collects/setup/unixstyle-install.rkt
+++ b/racket/collects/setup/unixstyle-install.rkt
@@ -44,6 +44,8 @@
 (define (get-arg)
   (when (null? args) (error "insufficient arguments"))
   (begin0 (car args) (set! args (cdr args))))
+(define (maybe-get-arg)
+  (and (pair? args) (get-arg)))
 
 (define op (string->symbol (get-arg)))
 (define adjust-mode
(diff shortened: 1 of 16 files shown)


racket/collects/setup/unixstyle-install.rkt
```diff
@@ -175,7 +177,7 @@
             [(link-exists? src)
              (make-file-or-directory-link (resolve-path src) dst)]
             [(directory-exists? src)
-             (make-directory dst)
+             (make-directory* dst) ; `*` support merging insteda of replacing
```
benknoble, line 180: Typo

mflatt: Fixed - thanks!

PR #4745, opened 2023-09-07
doc: correct in-directory examples

sorawee:
1. switch from collection-path to collection-file-path, since the former is deprecated.
2. remove one example that is incorrect. It is not possible to use in-directory to produce a list of unprefixed paths
3. fix the ordering of the last example. The old ordering was from before 980a9fbd (and possibly with a combination with Racket BC whose directory-list in the kernel doesn't sort the paths). But after the commit, the paths should be sorted.


mflatt:
LGTM

PR #4758, opened 2023-09-22
match: handle byte string in the pregexp pattern

### LINKED ISSUE #4281: Match documentation and implementation for pregexp are not in agreement

sorawee:
Fixes #4281

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -322,8 +322,10 @@ In more detail, patterns match as follows:
 
  @item{@racket[(#,(match-kw "regexp") _rx-expr)] --- matches a
        string that matches the regexp pattern produced by @racket[_rx-expr],
-       where @racket[_rx-expr] can be a @racket[regexp] or @racket[pregexp] or a string.
-       A string value is converted to a pattern using @racket[regexp].
+       where @racket[_rx-expr] can be either a @racket[regexp], a @racket[pregexp],
+       a @racket[byte-regexp], a @racket[byte-pregexp], a string, or a byte string.
+       A string and byte string value is converted to a pattern using
+       @racket[regexp] and @racket[byte-regexp] respectively.
        See @secref["regexp"] for more information about regexps.
 
        @examples[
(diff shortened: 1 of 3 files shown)


mflatt:
LGTM

PR #4759, opened 2023-09-22
guide: more #lang margin notes

sorawee:
This continues #4273

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/class.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/class.scrbl
@@ -13,8 +13,7 @@
 @title[#:tag "classes"]{Classes and Objects}
 
 @margin-note{This chapter is based on a paper @cite["Flatt06"].}
-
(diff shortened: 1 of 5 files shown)


shhyou:
Maybe for [contracts](https://github.com/racket/racket/blob/463cf39dc66ae0701cc33d5f0a365147a6e590fd/pkgs/racket-doc/scribblings/guide/contracts.scrbl) as well?

sorawee:
The contract section already has an explicit instruction in https://docs.racket-lang.org/guide/contract-boundaries.html

shhyou:
The last bit (except the macros which need `(for-syntax racket/base)`) seems to be `(require racket/port)` for `port->string` in I/O patterns. <https://github.com/racket/racket/blob/463cf39dc66ae0701cc33d5f0a365147a6e590fd/pkgs/racket-doc/scribblings/guide/io.scrbl#L402-L411>

sorawee:
Thanks!

PR #4760, opened 2023-09-23
doc: rephrase the explanation of make-do-sequence

sorawee:
I couldn't understand make-do-sequence for the longest time because it is very dense. Furthermore, the doc refers to values by position, which is not ideal since there are up to 7 values under consideration. Moreover, these positions can be shifted depending on how many values are returned, resulting in the phrase "The X (or X-1) result" throughout the doc, which makes it even more confusing.

This PR rephrases the doc, mostly by referring to results with names instead.


mflatt:
I guess my comments in #4761 are really about this PR.

sorawee:
Thanks for the feedback! Fixed.

sorawee:
I also made an additional change, replacing `->*` with `->`, which I find simpler to read.

mflatt:
LGTM

PR #4777, opened 2023-10-02
guide: explicitly mention variadic function

sorawee:
This helps people who are familiar with the term "variadic function" from other PLs.


mflatt:
Should other ways of creating variadic functions (i.e., optional arguments and `case-lambda`) also mention this term?

sorawee:
From a quick search, the way people use variadic arguments seems to imply unboundedness as well. And while it's true that `case-lambda` can create a function that consumes unbounded number of arguments, that's a feature of "rest arguments" inside `case-lambda` itself. Maybe we can add links from `case-lambda` to this explanation if we can find a good way to frame it. I don't know how to do that though. Meanwhile, I'd go ahead and merge this.

PR #4781, opened 2023-10-05
add more test cases for `'provide/contract-original-contract`

usaoc:
##### Checklist
- [x] tests included

### Description of change
It was found during the review of #4724 that the existing test cases only test the second fields of the property values.  Add test cases also for the first fields and the whole values.


rfindler:
Thanks!

PR #4784, opened 2023-10-06
static FFI call generation for CS

mflatt:
*Not to be merged until after the v8.11 branch.*

The `ffi/unsafe` API is an especialy dynamic interface, which made sense for BC, where all code generation is dynamic. For direct-to-machine-code Chez Scheme, it's not such a good fit. The `ffi/unsafe` library works in general by calling Chez Scheme's `eval` to build callout and callback code on demand; that pushes work to run time, and isn't compatible with a possible future bundling mode that omits the Chez Scheme compiler from a built executable.

Instead of changing the `ffi/unsafe` API, this commit approaches static generation as an optimization problem. The optimization is sufficient for all foreign callouts and callbacks in minimal Racket and in the "draw-lib" package, except for `net/macosx-ssl` (which needs support for struct C types).

The optimization is a cooperative effort in four parts:

 * The Rumble layer provides a new form `ffi-static-call-and-callback-core`, which generates the core piece of a foreign callout or callback at compile time based on Chez Scheme foreign representations. The Rumble layer also provides `ffi-maybe-call-and-callback-core`, which doesn't do anything at run time, but it serves a hint to the schemify layer. Finally, the Rumble layer provides `assert-ctype-representation`, which is also intended as a hint, and it checks its arguments only when the Rumble layer is compiled in safe mode.

 * The `unsafe/ffi` layer expands `_fun` forms to use `ffi-maybe-call-and-callback-core` as a hint to the schemify layer. It also takes care to define C types in a way that is visible to schemify's constant/copy propagation, including using `assert-ctype-representation` as needed. A few functions exported from `unsafe/ffi` have turned into macros that add an aseertion wrapper.

 * The schemify layer recognizes the primitive C type definitions, and it maps them to Chez Scheme representations, propagates representation information through definitions and across module boundaries, and detects when a `ffi-maybe-call-and-callback-core` can be converted to `ffi-static-call-and-callback-core` based on that propagated information.

 * The cross-compilation protocol adds support compile-time queries about the result of `system-type` and `foreign-sizeof`. The schemify layer uses this new interface to enable compile-time propagation of C type representations (which might depend, for example, on whether the platform is 32-bit or 64-bit).

The optimization path does not yet handle struct and union arguments or results --- as opposed to a pointer to a struct or union, which is handled.

The new `ffi/unsafe/static` library exports the same binindings as `ffi/unsafe`, except that it replaces `_fun` with one that will trigger a compile-time error if a `ffi-maybe-call-and-callback-core` cannot be converted into `ffi-static-call-and-callback-core`. The intent is that libraries use `ffi/unsafe/static` to make sure that dynamic code generation will not be needed by the module. There's a danger that some future schemify revision loses constant-propagation power and fails to compile a module using `ffi/unsafe/static`, but the idea is that such a regression should be treated as a bug, and hopefully `ffi/unsafe/static` will help catch it.

For BC, `ffi-maybe-call-and-callback-core` simply returns its arguments as in CS, and there is no conversion to static code generation.

The performance difference bwteen static and dynamic code generation tends to be very small, because on-demand compilation of foreign callouts and callbacks is already lazy, cached, and fast. Also the core of a callout or callback is always wrapped further to check arguments cooperate with the Racket thread scheduler. The immediate goal is not performance, but to have a path to compilation-free run time; further improvements may one day produce performance benefits. Meanwhile, improvements that were needed at the schemify level might provide small benefits unrelated to foreign calls.


samth:
How close is this to supporting the needs of the IO linklet/the `rktio` library on Racket CS?

mflatt:
The "io" layer in Racket CS already uses static bindings to access rktio. That layer doesn't go through `unsafe/ffi`, but instead uses macros implemented in "racket/src/cs/io.sls".

samth:
Right, I am wondering if this is sufficient to mean that the io layer could use `unsafe/ffi`, and if not what further additions would be needed.

mflatt:
I think `unsafe/ffi` probably could be used for the io layer, if it made sense, although probably a little more work on removing conversion-wrapper overheads would be in order. I am not certain, though.

PR #4786, opened 2023-10-08
for: clean up some outer checks in expansion

usaoc:
See racket/typed-racket#1338.

### DIFF
--- a/racket/collects/racket/private/for.rkt
+++ b/racket/collects/racket/private/for.rkt
@@ -722,16 +722,20 @@
   (define (:list-gen l)
     (values car cdr values l pair? #f #f))
 
+  (define (check-mlist l)
+    (unless (or (mpair? l) (null? l)) (raise-argument-error 'in-mlist "(or/c mpair? null?)" l)))
   (define (in-mlist l)
-    (unless (or (null? l) (mpair? l)) (raise-argument-error 'in-mlist "(or/c mpair? null)" l))
+    (check-mlist l)
     (make-do-sequence (lambda () (:mlist-gen l))))
 
   (define (:mlist-gen l)
     (values mcar #f mcdr l mpair? #f #f))
 
-  (define (in-input-port-bytes p)
+  (define (check-in-input-port-bytes p)
     (unless (input-port? p)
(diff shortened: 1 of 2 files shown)


sorawee:
Have you tested this against Typed Racket? I have been avoided touching this file because changing it will usually break Typed Racket.

usaoc:
I see—yes, three type annotations need to be added to Typed Racket under this change.  I can open a PR there.

racket/collects/racket/private/for.rkt
```diff
@@ -722,26 +722,32 @@
...
   (define (in-mlist l)
-    (unless (or (null? l) (mpair? l)) (raise-argument-error 'in-mlist "(or/c mpair? null)" l))
+    (check-mlist l)
     (make-do-sequence (lambda () (:mlist-gen l))))
 
   (define (:mlist-gen l)
     (values mcar #f mcdr l mpair? #f #f))
 
-  (define (in-input-port-bytes p)
+  (define (check-in-input-port-bytes p)
```
sorawee, line 734: If you will also change Typed Racket anyway, I think it might be a good opportunity to make this a general `check-input-port` that consumes a `who` argument and `p`. The name `'in-input-port-bytes` or `'in-input-port-chars` can then be supplied with the `who` argument.

usaoc: I have some concern about this because I don’t want to duplicate the `who` argument, which then must be supplied both in the expansion and the dynamic sequence constructor.

samth:
Should some of these checks be omitted in unsafe mode?

usaoc:
I have no strong opinion on this, but I do notice that those with `unless-unsafe` outer checks also use unsafe operations.  I suppose it isn’t too useful to omit the tests unless you also use unsafe operations.

usaoc:
I’ve pushed an update that does *a bit* more than the previous commit.  It now wraps more checks in `unless-unsafe`, unifies the style of always-succeeding tests to `(void)`, and changes one place that uses `error` to `raise-arguments-error`.

mflatt:
Do I understand correctly that Typed Racket is updated so that it's ok to merge this PR?

samth:
Yes

PR #4789, opened 2023-10-11
Correct asm-mul/ovfl of riscv backend.

maoif:
##### Checklist

- [x] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation

### Description of change

The old code fails to detect edge case like `(* -4503599627370496 512)` as overflow and will fail to call `big_mul()` in the runtime.

This fix detects multiplication overflow by testing whether `hi(x*y)` is equal to `lo(x*y) >> 63`, where `hi(n)` is the upper 64-bit of `n` and `lo(n)` is the lower 64-bit, assuming the product is 128-bit wide.
 If not equal, we have overflow.

### DIFF
--- a/racket/src/ChezScheme/s/riscv64.ss
+++ b/racket/src/ChezScheme/s/riscv64.ss
@@ -1008,20 +1008,18 @@
       (Trivit (dest src0 src1)
               (emit mul dest src0 src1 code*))))
 
+  ;; overflow if hi64(src0*src1) != (lo64(src0*src1) >> 63)
   (define asm-mul/ovfl
     (lambda (code* dest src0 src1)
       (Trivit (dest src0 src1)
-        (emit xor %scratch1 src0 src1 ; 1 high bit => expect negative
-              (emit mulh %scratch0 src0 src1
+              (emit mulh %scratch1 src0 src1
                     (emit mul dest src0 src1
-                          ;; overflow if %scratch0 doesn't hold 0 for an expected
-                          ;; positive result or -1 for an expected negative result;
-                          ;; also overflow if dest doesn't match expected sign
-                          (emit srai %scratch1 %scratch1 63 ; -1 => expected negative; 0 => expected positive
-                                (emit srli %cond dest 63 ; 1 => negative in `dest`
-                                      (emit or %scratch0 %scratch0 %cond ; combine negativity of results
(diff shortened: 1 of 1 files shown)


sorawee:
Does it make sense to add a test?

mflatt:
I'm not sure which test in `fl.ms` should be relevant, but it's not failing for me. Adding `(eqv? (* -4503599627370496 512) -2305843009213693952)` to `5_3.ms` seems best, since it clearly fails and is fixed by the commit.

@maoif I've pushed a revised commit at https://github.com/mflatt/racket/tree/rvmul. I can force push to your repo and then merge the PR, if that's ok — but it would be a force-push to your repo's `master` branch.

maoif:
> I'm not sure which test in `fl.ms` should be relevant, but it's not failing for me. Adding `(eqv? (* -4503599627370496 512) -2305843009213693952)` to `5_3.ms` seems best, since it clearly fails and is fixed by the commit.
> 
> @maoif I've pushed a revised commit at https://github.com/mflatt/racket/tree/rvmul. I can force push to your repo and then merge the PR, if that's ok — but it would be a force-push to your repo's `master` branch.

The following code fails to report expected error in `fl.ms` in my old cisco/ChezScheme:

```
(error? (flonum->fixnum
              (* (inexact (most-negative-fixnum)) 2.0)))
```

The bug showed up only when cp0 is enabled. The code for folding `flonum->fixnum` in cp0 is:

```
(fold (flonum->fixnum flonum?) target-fixnum? (lambda (x) (#2%truncate (#2%exact x))))
``` 
Then we have `#2%exact` -> `convert-to-exact` -> `decode-float`, which calculates the mantissa of `(* (inexact (most-negative-fixnum)) 2.0)` to be `-4503599627370496`, and the exponent to be `9`...

The addition of `(eqv? (* -4503599627370496 512) -2305843009213693952)` would make it more explicit anyway.
So please push the commit to my repo.

mflatt:
Aha — I should have thought to try `cp0=t`.

Meanwhile, I see that I already accidentally pushed your repair as c3ac7926aa60dce3ddddc865c9a7ba61edb66a1e. Thanks for the closing the PR, and I'm sorry that I merged in a way that didn't record this PR as merged! I'll make sure this repair gets copied over to cisco/ChezScheme when v9.9.9 is merged there.

PR #4800, opened 2023-10-20
match: fix box pattern

### LINKED ISSUE #4799: Contracts break pattern matching
I used every version from `8.3 [cs]` to `8.6 [cs]`.

**What program did you run?**
I ran the following program.

``` racket
#lang racket/base

(require racket/contract)
(require racket/match)

(define (explain obj)
  (match obj
    ((box int) (format "just ~a" int))
    (#f "nothing")
    (_ "something else")))

(define/contract (explain/contract obj)
  (-> (or/c #f (box/c exact-integer?)) string?)
  (explain obj))

(displayln (explain (box 3)))
(displayln (explain #f))
(newline)
(displayln (explain/contract (box 3)))
(displayln (explain/contract #f))
```

The program should have printed

```
just 3
nothing

just 3
nothing
```

but it actually printed

```
just 3
nothing

just #&3
nothing
```

violating the expected invariant `(not (equal? obj int))`
in the first clause of `match` in the body of `explain`.

sorawee:
A box could be impersonated, so we can't use unsafe-unbox*.

Fixes #4799

##### Checklist

- [x] Bugfix
- [x] tests included

### DIFF
--- a/racket/collects/racket/match/compiler.rkt
+++ b/racket/collects/racket/match/compiler.rkt
@@ -100,7 +100,7 @@
         #`[question (let ([tmps (accs #,x)] ...) body)])))
   (cond
     [(eq? 'box k)
-     (compile-con-pat (list #'unsafe-unbox*) #'box? (compose list Box-p))]
+     (compile-con-pat (list #'unsafe-unbox) #'box? (compose list Box-p))]
     [(eq? 'pair k)
      (compile-con-pat (list #'unsafe-car #'unsafe-cdr) #'pair?
                       (lambda (p) (list (Pair-a p) (Pair-d p))))]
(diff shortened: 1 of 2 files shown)


racket/collects/racket/match/compiler.rkt
```diff
@@ -100,7 +100,7 @@
         #`[question (let ([tmps (accs #,x)] ...) body)])))
   (cond
     [(eq? 'box k)
-     (compile-con-pat (list #'unsafe-unbox*) #'box? (compose list Box-p))]
+     (compile-con-pat (list #'unsafe-unbox) #'box? (compose list Box-p))]
```
sorawee, line 103: I won't change this PR any further (besides for addressing PR feedback). But it looks like this `compose` should have been `compose1`?

sorawee: Oh, but it's at compile-time anyway. Wouldn't matter much.

samth (approved):

PR #4807, opened 2023-10-30
doc: show value of defthing via #:auto-value

sorawee:
Not sure if this is the best approach, so let me know if there's a better way.

Before:

Screenshot 2023-10-30 at 3 48 43 AM

After:

Screenshot 2023-10-30 at 3 48 58 AM

### DIFF
--- a/pkgs/racket-doc/xml/xml.scrbl
+++ b/pkgs/racket-doc/xml/xml.scrbl
@@ -416,7 +416,7 @@ or otherwise escaping. Results from the leaves are combined with
 
 @section{Parameters}
 
-@defparam[current-unescaped-tags tags (listof symbol?) #:value null]{
+@defparam[current-unescaped-tags tags (listof symbol?) #:auto-value]{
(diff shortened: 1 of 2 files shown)


pkgs/racket-doc/xml/xml.scrbl
```diff
@@ -14,6 +16,10 @@
 @interaction-eval[#:eval xml-eval (require racket/list)]
 @interaction-eval[#:eval plist-eval (require xml/plist)]
 
+@(define-syntax-parse-rule (defthing/literal-value name:id contract #:from mod:id xs ...)
```
LiberalArtist, line 19: I think you can avoid the redundant `#:from mod:id` by using the third result of `(identifier-binding #'name #f)`. (In the general case you might have to do something more to handle renaming, too &hellip;)

sorawee:
I move the main mechanism to Scribble. See https://github.com/racket/scribble/pull/391.

mflatt:
LGTM

PR #4808, opened 2023-10-31
docs: update old-fashioned contracts

usaoc:
##### Checklist
- [x] documentation

### Description of change
This commit mainly updates `false/c` to `#f` and `one-of/c` to `or/c`. A few contracts are also simplified/corrected as I happen to have noticed them.

### DIFF
--- a/pkgs/racket-doc/scribblings/foreign/pointers.scrbl
+++ b/pkgs/racket-doc/scribblings/foreign/pointers.scrbl
@@ -92,7 +92,7 @@ in a pointer.  The same operation could be performed using
             any]
            [(ptr-ref [cptr cpointer?]
                      [type ctype?]
-                     [abs-tag (one-of/c 'abs)]
+                     [abs-tag 'abs]
                      [offset exact-nonnegative-integer?])
             any]
            [(ptr-set! [cptr cpointer?]
@@ -106,7 +106,7 @@ in a pointer.  The same operation could be performed using
             void?]
            [(ptr-set! [cptr cpointer?]
                       [type ctype?]
-                      [abs-tag (one-of/c 'abs)]
+                      [abs-tag 'abs]
                       [offset exact-nonnegative-integer?]
                       [val any/c])
             void?])]{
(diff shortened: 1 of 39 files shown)


pkgs/racket-doc/scribblings/foreign/pointers.scrbl
```diff
@@ -212,11 +212,11 @@ see @|InsideRacket|.
...
                  [cptr cpointer? @#,elem{absent}]
-                 [mode (one-of/c 'raw 'atomic 'nonatomic 'tagged
-                                 'atomic-interior 'interior
-                                 'stubborn 'uncollectable 'eternal)
+                 [mode (or/c 'raw 'atomic 'nonatomic 'tagged
+                             'atomic-interior 'interior
+                             'stubborn 'uncollectable 'eternal)
                        @#,elem{absent}]
-                 [fail-mode (one-of/c 'failok) @#,elem{absent}])
+                 [fail-mode 'failok @#,elem{absent}])
```
sorawee, line 219: Not related to your change, but I wonder why these arguments don't have the contract `(or/c #f ...)` with the default value of `#f`.

pkgs/racket-doc/scribblings/raco/api.scrbl
```diff
@@ -284,7 +284,7 @@ the files that it compiles and produces. The default is @racket[#f].}
 A @racket[#t] value for the parameter causes the compiler to print
 verbose messages about its operations. The default is @racket[#f].}
 
-@defparam[compile-subcollections cols (one-of/c #t #f)]{
```
sorawee, line 287: Two things:

1. It would be nice to mention explicitly in the doc below that the argument is expected to be a "true" value or `#f`.
2. I think `(compile-subcollections)` should ideally always return a boolean, similar to parameters like `eval-jit-enabled` (obviously this change is out of scope of your current PR).

usaoc: Actually it seems like *none* of the parameters in `compiler/option` do the Boolean conversion, so the `boolean?` contracts are technically wrong.  I can change them to the correct contracts (and descriptions) *now*, and another PR can address the Boolean conversion part.

usaoc: On a second thought, I can also submit that “another PR” first, so that this can be merged after that :P

pkgs/racket-doc/scribblings/raco/exe-api.scrbl
```diff
@@ -394,17 +394,17 @@ have been applied as needed to refer to the existing file).
...
+                                                       (list/c (or/c symbol? #f #t)
                                                                (or/c module-path? path?)
                                                                (listof symbol?))))]
                                [literal-files (listof path-string?)]
                                [literal-sexp any/c]
                                [cmdline (listof string?)]
                                [aux (listof (cons/c symbol? any/c)) null]
                                [launcher? any/c #f]
-                               [variant (one-of/c 'cgc '3m'cs) (system-type 'gc)]
+                               [variant (or/c 'cgc '3m'cs) (system-type 'gc)]
```
sorawee, line 407: Could you add a space between `'3m` and `'cs`?

sorawee:
I'm marking this PR as not ready to be merged, since it depends on #4810.

PR #4809, opened 2023-10-31
docs: minor corrections to Networking Libraries reference

usaoc:
##### Checklist
- [x] documentation

### DIFF
--- a/pkgs/net-doc/net/scribblings/http-client.scrbl
+++ b/pkgs/net-doc/net/scribblings/http-client.scrbl
@@ -21,7 +21,7 @@ Identifies an HTTP connection.
 @defproc[(http-conn-live? [x any/c])
          boolean?]{
 
-Identifies an HTTP connection that is "live", i.e. one that is still
+Identifies an HTTP connection that is ``live'', i.e., one that is still
 connected to the server.
 
 }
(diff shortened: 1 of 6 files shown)


pkgs/net-doc/net/scribblings/http-client.scrbl
```diff
@@ -204,13 +204,18 @@ response, which is why there is no @racket[#:closed?] argument like
...
  @itemize[
-   @item{If @racket[ssl?] was @racket[#f] then @racket[#f]. Otherwise
-     an @racket[ssl-client-context?] that has been negotiated with the
-     target.
-
-     If @racket[ssl?] was a protocol symbol, then a new
-     @racket[ssl-client-context?] is created, otherwise the current
-     value of @racket[ssl?] is used.}
+   @item{
+     If @racket[ssl?] is @racket[#f], returns @racket[#f], otherwise
```
sorawee, line 208: It's kinda weird to say "returns" here. I like the old text better. The first bullet point is about the first value of the four values. There's no function that returns this single value specifically.

usaoc: The original phrasing read like an `if` expression, not an *English* sentence.  It took me some puzzling to understand what it was talking about.  A better phrasing is welcome, but it should read like an Enflish sentence.

sorawee: Perhaps replace "returns ..." with "the (first) value is ..."

usaoc: Rephrased this part (actually, the whole list).

PR #4810, opened 2023-10-31
compiler/option: supply parameter guards and names

usaoc:
Raised by Sorawee during the review of #4808.


usaoc:
Re. the second point: I don’t think that’s a good idea.  What I *do* think is a good idea is more general use of `who`, but it’s probably too late for us to do this in Racket.  Maybe something we can consider for Rhombus?

PR #4821, opened 2023-11-14
CS: allow more error adjustment of primitives

usaoc:
Add some missing calls to `error-contract->adjusted-string`.  Also fix the name of `raise-arguments-error*` (when *it* raises).


mflatt:
Thanks!

PR #4824, opened 2023-11-15
`make pb-fetch` for Building `racket/src`

abhillman:
### Description of change

Adds a reference to `make pb-fetch` in `racket/src/readme.txt`, which is required to successfully build `racket/src`.

### DIFF
--- a/racket/src/README.txt
+++ b/racket/src/README.txt
@@ -78,6 +78,7 @@ Quick instructions:
  From this directory (where the `configure` file is), run the following
  commands:
 
+   make --directory=../../ pb-fetch
    mkdir build
    cd build
    ../configure


abhillman:
Looks great! With the exception of two comments I made on https://github.com/mflatt/racket/commit/f4d3c130ee4e677859a6cad4bbf1ff8059f9055e, I think it looks great; as, indeed, when one downloads sources from https://download.racket-lang.org/ `src/ChezScheme/boot/pb` is pre-populated.

abhillman:
Of course out of scope here, but do you know of any documentation regarding the "minimal"-ness of the source-based distribution? i.e. how one would augment it such that it is no longer "minimal"? Indeed, one can merely use the make file at the repository root, but clarifying how to get to a full-fledged distribution (besides reading build sources 😉) could be helpful in understanding how things "click together."

abhillman:
Might be good to update https://github.com/racket/racket/blob/7ed81eba162ba6af41e7522d54231df09b7ff186/build.md?plain=1#L99 with `make pb-fetch` as well.

mflatt:
Thanks for the comments. I pushed an update as bac99596547d5ef5fcd45d3c9362cf159a3f825b.

However, I misremembered what you suggested for "build.md", and so that change is not yet in place. I had added a note about installing "racket-lib" there, instead.

mflatt:
I think the latest commit covers the need changes, so I'll close the PR. Please open a new PR or issue if there are still/new problems. Thanks!

PR #4827, opened 2023-11-17
Handle get-info errors when getting pkg content.

samth:
Important to recover properly from eg https://pkgs.racket-lang.org/package/formica

##### Checklist

- [x] Bugfix

### DIFF
--- a/pkgs/racket-doc/pkg/scribblings/lib.scrbl
+++ b/pkgs/racket-doc/pkg/scribblings/lib.scrbl
@@ -705,7 +705,7 @@ and status reporting.
                                    @racket[#:quiet?] arguments.}]}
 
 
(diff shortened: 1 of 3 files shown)


mflatt:
Seems ok to me.

sorawee:
Not related to this PR per se, but would it be a good idea to relax `extract-pkg-dependencies`'s contract to allow `#f`?

mflatt:
As the first argument to `extract-pkg-dependencies`? No objection, although I'm not sure where to draw the line on allowing arguments that represent failure for an earlier step.

samth:
I think the specified contract should change; the implementation of `extract-pkg-dependencies` already handles `#f` which is what this PR relies on.

sorawee:
It may also be a good idea to add an option to `get-info/full` to not raise an error but instead return `#f`? But again, that's out of scope of this PR.

The PR looks good to me too.

PR #4830, opened 2023-11-21
schemify: more systematic unwrapping of vacuous `let-values`

usaoc:
There are several places in `schemify` that appears to unwrap *vacuous* `let-values` (i.e., without bindings) and friends.  This commit adds an `unwrap-let` function and attempts to systematically use it.  This way, more optimization can be done in `schemify`.  The most apparent benefit is that cross-module inlining can work better in the presence of vacuous `let-values`, possibly introduced by macros.


mflatt:
This looks great, and I'll merge.

If possible, while the change is fresh in mind, it would be good to add a test to https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/optimize.rktl that passes only after this improvement (to make sure it doesn't get lost in the future). It's not always so easy to construct an example, though.

PR #4832, opened 2023-11-22
remove vacuous `let-values` from `splicing-syntax-parameterize`

usaoc:
In the `define-values` case, the expansion can end up with a vacuous `let-values` even though no definition context is needed.  Remove this `let-values` and clean up the expansion code.

An example in the Guide wasn’t supposed to work, but accidentally worked due to this.

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/class.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/class.scrbl
@@ -989,47 +989,45 @@ thus can only be used, for methods where no Beta-style augmentation has taken
 place. The following example shows this difference:
 
 @racketblock[
+(define/contract glutton%
+  (class/c (override [eat (->m edible/c void?)]))
+  (class animal%
+    (super-new)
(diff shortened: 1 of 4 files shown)


usaoc:
Uh, and this also revealed that there was a broken example in the Guide.  I pushed a fix.

samth:
Aside from the fixed example, is the gain here less work for the new transformation you added to schemify?

usaoc:
I suppose yes.  For the context, this caught my attention due to some inlining issue in racket/rhombus-prototype#395; I tried `splicing-syntax-parameterize`, and surprisingly found that even trivial functions weren’t inlined!  Now that we have the transformation, `schemify` can recognize trivial functions hidden inside `let-values` in this fashion, but I think it’s still worth it to fix the expansion of `splicing-syntax-parameterize` itself.  (The fixed example also suggests that the `let-values` can lead to a wrong semantics, namely that it introduces a definition context.)

PR #4841, opened 2023-11-24
CS: fix behavior of `regexp-max-lookbehind`

### LINKED ISSUE #4840: Incorrect behavior of CS `regexp-max-lookbehind`
I’m not sure `regexp-max-lookbehind` is correct.  For example, the docs says that “`e(?<=a..)d` consults two bytes before a matching `ed`”, but shouldn’t this consult *zero* bytes given that the lookbehind is *after* the starting position?  CS and BC don’t seem to agree on this, either.  `(regexp-max-lookbehind #rx#"e(?<=a..)d")` evaluates to `2` (indeed like the docs says, but well …) on BC, but `3` on CS.

usaoc:
##### Checklist
- [x] Bugfix
- [x] tests included
- [x] documentation

### Description of change
For a lookbehind somewhere in the middle, the maximum lookbehind should take the minimum number of bytes after the starting position into account.  This commit fixes this behavior.

Also fix the contract string and add test cases.

Fix #4840.


mflatt:
Thanks!

PR #4842, opened 2023-11-27
Add compiler trace logs to linklet and expander layers

jryans:
This adds additional compiler trace logs (all off by default, guarded by various environment variables) that report names for each module and linklet as it flows by during compilation.

The new trace logs (and some existing ones as well) write to standard error instead of output, which testing has shown fares better. Logging to standard error preserves the expected ordering of events from different layers. Chez's nanopass tracer has also been changed to write to standard error as well. (I will submit this part upstream in a separate PR. I have only included it here to clarify the story for review.)

Along with being generally useful logs to have when working on compiler internals, these also provide enough info to add a Racket optimisation pipeline view to Compiler Explorer (see sneak peek below or [on Discord](https://discord.com/channels/571040468092321801/618895179343986688/1177684975278100480)).

Depends on https://github.com/cisco/ChezScheme/pull/771


sorawee:
Here's the screenshot in case anyone doesn't have a Discord account.

image

mflatt:
After a first quick look, I worry about the two implementations of logging, and also about calling `getenv` frequently.

Is the expander-level logging meant to connect to a subsequent `compile-linklet` call, or does it provide more information than that? Put another way, would adding an information argument to `compile-linklet` achieve the same goal (assuming that `compile-linklet` uses the argument for logging), or would that be different/insufficient?

jryans:
This is now ready for another look. It now feels quite a bit simpler, thanks to @mflatt's suggestions. 

I have removed the expander-layer logging by passing the additional expander data via an `info` hash to `compile-linklet`. The logging also now reuses the existing `PLT_LINKLET_SHOW` / `show-on?` state rather than checking an additional environment variable, which seemed like the natural thing to do once they expander-layer logging went away.

jryans:
Ah, looks like I actually need to tweak BC slightly, I'll investigate that now.

mflatt:
Thanks for this work and this update!

How difficult would it be to move `info` to the end of the argument list for `compile-linklet` and `recompile-linklet`? I think those are the public functions (https://docs.racket-lang.org/reference/linklets.html#%28def._%28%28lib._racket%2Flinklet..rkt%29._compile-linklet%29%29), in which case backward compatibility is better.

sorawee:
Are these functions allowed to use keyword argument? Functions with a lot of positional, optional arguments are really difficult to use (actually, even the non-optional ones too).

I guess an argument against keyword arguments here would be the consistency with the rest of the API, since none of them has it.

jryans:
> How difficult would it be to move `info` to the end of the argument list for `compile-linklet` and `recompile-linklet`? I think those are the public functions ([docs.racket-lang.org/reference/linklets.html#%28def._%28%28lib._racket%2Flinklet..rkt%29._compile-linklet%29%29](https://docs.racket-lang.org/reference/linklets.html#%28def._%28%28lib._racket%2Flinklet..rkt%29._compile-linklet%29%29)), in which case backward compatibility is better.

Ah yeah, just noticed they are public, I'll update the docs to match whatever API shape we end up with.

As for moving `info`, I'd ideally like the content of the previously existing `name` argument to be included within the `info` hash (rather than leaving `name` where it is and also adding `info` as well). To preserve compatibility, my current thinking is to change `compile-linklet` to accept either a symbol or hash in that argument position.

jryans:
> Are these functions allowed to use keyword argument? Functions with a lot of positional, optional arguments are really difficult to use (actually, even the non-optional ones too).

No, I don't believe the core `compile-linklet` function can use keywords arguments, as it's implemented in Scheme and not in Racket.

mflatt:
Ah, I missed that `name` was replaced by `info`. That plan makes sense.

samth:
> No, I don't believe the core compile-linklet function can use keyword arguments, as it's implemented in Scheme and not in Racket.

It would be possible to use keywords by wrapping these functions with Racket-level definitions, as is done for `open-input-file` et al, but that might be more trouble than it's worth.

samth:
If we wanted to make this all use Racket logging, I think we'd need to do the following:

- call the relevant logging functions from the Scheme `(io)` library in `linklet.sls`.
- add another parameter with a function to call instead of `printf` in `tracer` in `s/cpnanopass.ss` so that logging could be inserted.
- do the same for `#%assembly-output`, or allow that to be a function to call instead of a port to print to.

I think that could be done backwards-compatibly with this, so I'm not saying anything negative here; just wanted to note the steps we might want to take in the future.

pkgs/racket-doc/scribblings/reference/linklet.scrbl
```diff
@@ -142,8 +141,13 @@ As long as @racket['serializable] included in @racket[options], the
...
 part of a @tech{linklet bundle} (possibly in a @tech{linklet directory}).
 
-The optional @racket[name] is associated to the linklet for debugging
-purposes and as the default name of the linklet's instance.
+The optional @racket[info] hash provides various debugging details
+about the linklet, such as the module name the linklet is part of,
+the linklet name, and the phase for body linklets. If a @racket['name]
+value is present in the hash, it is associated to the linklet for
+debugging purposes and as the default name of the linklet's instance.
+If @racket[info] is not a hash, it assumed to be a name value directly
```
sorawee, line 149: "it is assumed"?

pkgs/racket-doc/scribblings/reference/compiler.scrbl
```diff
@@ -136,4 +139,6 @@ compilation and evaluation times on exit. When the
 information about macro-expansion time is printed on exit.
 
 @history[#:changed "8.8.0.10" @elem{Added special pass name @tt{all}
-                                    to @envvar{PLT_LINKLET_SHOW_PASSES}}]
+                                    to @envvar{PLT_LINKLET_SHOW_PASSES}}
+         #:changed "8.11.1.2" @elem{Added module and linklet info
+                                    to output}]
```
mflatt, line 144: A nit, but if you're updating docs anyway for a typo, each of theses `elems` (the old one and new one) should end with a period.

mflatt:
Looks just about ready to me, so I'll just wait for small doc updates.

I'm inclined to squash the commits when merging, but let me know if you think it's important to keep them separate.

jryans:
Thanks all for the reviews! I have addressed the docs issues, so hopefully this is now ready to merge.

> I'm inclined to squash the commits when merging, but let me know if you think it's important to keep them separate.

I'm okay with either... I do try to keep my commits as independent units of work, but if squashing is a better fit, then of course go with that (as long as the end result has reasonable commit message). 🙂

Issue #4277, opened 2022-06-02
Consider using win-iconv for iconv on Windows


samth:
Currently, the "base" package depends on the ["racket-win32-x86_64"](https://github.com/racket/libs/tree/master/racket-win32-x86_64-3) package, which includes libiconv. libiconv is under the LGPL v2.1, and is the only LGPL dependency of minimal Racket (the others are from OpenSSL and libedit on various platforms). 

The win-iconv project (see https://github.com/burgerrg/win-iconv) has a Windows-only iconv replacement that is already mentioned by the Chez documentation. It is available under a non-restrictive license. It's also much smaller (1 MB vs 30kB).

Aeva:
What license is win-iconv under?  I don't see it in the git repository.

samth:
The readme says "public domain"[^1] (see https://github.com/burgerrg/win-iconv/blob/master/readme.txt#L3) 

[^1]: Yes I know this isn't really something you can do in all relevant legal systems.

otherjoel:
If it is in the public domain, I would assume any person or organization in a country that recognizes that status could simply fork it and issue it under a new license. But now you're the maintainer.

LiberalArtist:
@samth wrote:
> @Aeva wrote:
> > We should try to reach out to the author and see if they're willing to apply CC0 to it, so it can be used internationally. Otherwise, this is exchanging one set of licensing problems for another.
>
> I'd be fine with contacting the author, although there haven't been any commits since 2016. But I think that while the public domain declaration has various legal issues, it's almost certainly not a problem in practice and would improve the licensing situation.

I know @burgerrg is involved in Chez Scheme, but it looks to me from https://github.com/burgerrg/win-iconv/blob/db14b52d58f98b7700863bb38ae69d7d5121f1fa/readme.txt#L20 like win-iconv is forked from an upstream project with several contributors: https://github.com/win-iconv/win-iconv/graphs/contributors I think to be really robust we would need all of them to agree to CC-0 (unless maybe some have made only trivial commits). (Maybe this is a situation where we should seek guidance from the SFC?)

I agree that the informal public-domain declaration is not very likely to be a problem in practice. However, when *not* building with `--embed-dlls` or for a platform (like a game console) that's hostile to relinking, I would find the LGPL more confidence-inspiring than the ad-hoc public-domain statement.

Aeva:
[...]

Issue #4542, opened 2023-01-13
Does Racket 8.x actually work properly on Windows 7 ?


rebb:
Does Racket 8.x actually run on Windows 7 as the Download page suggests ?
I'm having problems installing herbie on it, and DrRacket.exe also does not start, instead there are multiple DLL related errors.

**What version of Racket are you using?**
8.7

**What program did you run?**
raco pkg install --auto herbie

**What should have happened?**
Herbie should have installed properly.

**If you got an error message, please include it here.**
```
Resolving "herbie" via https://download.racket-lang.org/releases/8.7/catalog/
Resolving "herbie" via https://pkgs.racket-lang.org
Downloading repository https://github.com/uwplse/herbie.git?path=src#v1.6 commit 600fd02dd9057ff758c4c73f806e323470f72a02
make-file-or-directory-link: cannot make link
  path: e:\Temp\UserTemp\SYSTEM\16736245871673624587460-uwplse_herbie_git_v1_6\www\doc\latest
  system error: The parameter is incorrect.; win_err=87
  context...:
   G:\Progs\Racket\collects\net\git-checkout.rkt:776:3
   [repeats 2 more times]
   G:\Progs\Racket\collects\net\git-checkout.rkt:167:11
   G:\Progs\Racket\collects\net\git-checkout.rkt:55:2: retry-loop
   G:\Progs\Racket\collects\pkg\private\download.rkt:102:2: download!
   G:\Progs\Racket\collects\file\cache.rkt:63:2: fetch-and-continue
   G:\Progs\Racket\collects\racket\contract\private\arrow-val-first.rkt:555:3
   G:\Progs\Racket\collects\pkg\private\download.rkt:94:0: download-repo!
   G:\Progs\Racket\collects\pkg\private\stage.rkt:319:11
   G:\Progs\Racket\collects\pkg\private\stage.rkt:115:0: stage-package/info
   [repeats 1 more time]
   G:\Progs\Racket\collects\pkg\private\install.rkt:142:0: install-packages
   G:\Progs\Racket\collects\pkg\private\install.rkt:953:4
   G:\Progs\Racket\collects\racket\contract\private\arrow-val-first.rkt:555:3
   G:\Progs\Racket\collects\racket\file.rkt:763:8
   G:\Progs\Racket\collects\racket\file.rkt:752:0: call-with-file-lock
   ...
```

**Please include any other relevant details**
Windows 7

shhyou:
Try this:
[...]

PR #4844, opened 2023-12-03
rktio: support ICU as iconv alternative, e.g. for Windows

LiberalArtist:
This PR aims to support ICU as an alternative to `iconv` in rktio to implement [byte converters](https://docs.racket-lang.org/reference/bytestrings.html#%28def._%28%28quote._~23~25kernel%29._bytes-open-converter%29%29).

Windows does not have iconv, so Racket currently distributes `libiconv-2.dll` in platform-specific packages like `racket-win32-x86_64-3` that are part of even the minimal Racket distribution. Building and distributing third-party C libraries is no fun. However, while looking into https://github.com/racket/racket/issues/4277, I discovered that [Windows has provided ICU](https://learn.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-) as a supported system DLL since the Windows 10 Creators Update (1703), released in 2017. This appears to mean that all Windows versions in "mainstream support" from Microsoft include ICU.[^1]

[^1]: As explained in a comment in the code, in this PR I'm actually targeting Windows 10 version 1903, which avoids some per-thread initialization needed in earlier versions. The only earlier version of Windows still in mainstream support appears to be version 1809 (Enterprise LTSC 2019), for which [mainstream support ends in January](https://learn.microsoft.com/en-us/lifecycle/products/windows-10-enterprise-ltsc-2019).

My hope is that by using the OS-provided ICU, we can stop distributing `libiconv-2.dll` for minimal Racket on Windows.

The functionality may be useful in other contexts, too: for example, it seems Android has either no iconv or a very rudimentary iconv (vid. e.g. https://github.com/racket/racket/commit/8a087040128fad352a3b86cc94073a2e9585921c), but [Android provides ICU](https://developer.android.com/guide/topics/resources/internationalization#icu4c) as a system library.

I intend to contribute a similar patch to Chez Scheme, but I've found it easier to start with rktio.

# Open Design Questions

 1. I have not added any support to `configure` yet. I think the behavior should be:

     - On Windows, automatically enable ICU if `AC_CHECK_HEADER([icu.h])` finds it. To support building in older environments, don't build ICU support if the header is not found. To support running on older Windows versions, dynamically load `icu.dll` with `LoadLibraryW`/`GetProcAddress`. This means no linking flags are needed.
     - On other platforms, enable ICU only if specifically requested. If it is requested, find the appropriate flags either with [`AX_CHECK_ICU`](https://www.gnu.org/software/autoconf-archive/ax_check_icu.html) or by using `pkg-config` directly.

    But native Windows builds do not use Autoconf, and I'm not sure how to do the equivalent of `AC_CHECK_HEADER([icu.h])`. (I cannot overstate the extent to which I am not a Windows person&mdash;perhaps ironically, I have only tested this so far on GNU/Linux.) The simplest thing to do would probably be to omit the check and require an explicit `--disable-icu` (`/disableICU`?) to build in old Windows environments.

 2. If both iconv and ICU are enabled and found, currently ICU is used. That's certainly debatable. On non-Windows platforms, it seems to make sense because enabling ICU would be an explicit choice. On Windows, one consideration is that an iconv DLL might (sometimes?) end up being installed as a dependency of `racket/draw` (see https://github.com/racket/libs/commit/8ba7a8704b8052d297299be7f875e6a8b12364e1) or for some other reason, and having that change the behavior of `bytes-open-converter` seems worth avoiding.

 3. Currently, this change is entirely contained within the C code of rktio, with no change to the interface exposed to Racket. I could imagine ways of exposing more. The possibility I'd most strongly consider is adding a function to report whether iconv or ICU is being used. If there were a reason to want such a thing,  it would be possible to expose the new `rktio_iconv_converter_open` and `rktio_icu_converter_open` so they could both be used if both are available. If there were a desire to support `dlopen`ing ICU on non-Windows platforms, that might be a reason to drive more from the Racket side.

 4. There might be a better `ICU_BUF_SIZE`.

 5. If we want to stop distributing `libiconv-2.dll` for minimal Racket, there would need to be changes to the native library packages, which I have not attempted. That can (and probably should!) be done as a second step, though.

    As a compatibility consideration, I looked into the encoding names recognized by ICU compared to GNU libiconv. (The documentation is already clear that the supported encodings are system-dependent, so this is an extra level of scrutiny.) Both ICU and GNU libiconv recognize multiple aliases for some encodings. For most of the encodings supported by GNU libiconv (with `--enable-extra-encodings`), at least one alias is also listed by ICU. There are [some encodings](https://github.com/LiberalArtist/icuiconv/blob/8c5acb287bdd2966f356bed9bd67e5f9fa067734/encodings/missing-encodings.ss) for which that is not the case, though. I suspect these encodings are also supported by ICU under different names, and a sufficiently motivated person could fill in a translation table. Some of them might even be recognized by the matching functions in the ICU API: I only checked for exact matches.

### DIFF
--- a/racket/src/rktio/rktio_convert.c
+++ b/racket/src/rktio/rktio_convert.c
@@ -618,6 +618,11 @@ static void rktio_iconv_converter_close(rktio_t *rktio, rktio_iconv_converter_t 
   free(cvt);
 }
 
+static void rktio_iconv_convert_reset(rktio_t *rktio, rktio_iconv_converter_t *cvt)
+{
+  (void)iconv(cvt->cd, NULL, NULL, NULL, NULL);
+}
+
 static intptr_t rktio_iconv_convert(rktio_t *rktio,
                                     rktio_iconv_converter_t *cvt,
                                     char **in, intptr_t *in_left,
@@ -700,6 +705,14 @@ static void rktio_icu_converter_close(rktio_t *rktio, rktio_icu_converter_t *cvt
 #endif
 }
 
+static void rktio_icu_convert_reset(rktio_t *rktio, rktio_icu_converter_t *cvt)
+{
+  ucnv_reset(cvt->sourceCnv);
+  ucnv_reset(cvt->targetCnv);
+  cvt->pivotSource = &cvt->buf[0];
+  cvt->pivotTarget = &cvt->buf[0];
+}
+
 static intptr_t rktio_icu_convert(rktio_t *rktio,
                                   rktio_icu_converter_t *cvt,
                                   char **in, intptr_t *in_left,
@@ -713,10 +726,7 @@ static intptr_t rktio_icu_convert(rktio_t *rktio,
   if ((NULL == in) || (NULL == *in)) {
     if ((NULL == out) || (NULL == *out)) {
       /* Set cvt's conversion state to the initial state. */
-      ucnv_reset(cvt->sourceCnv);
-      ucnv_reset(cvt->targetCnv);
-      cvt->pivotSource = &cvt->buf[0];
-      cvt->pivotTarget = &cvt->buf[0];
+      rktio_icu_convert_reset(rktio, cvt);
       return 0;
     } else {
       /* out is not NULL and *out is not NULL */
@@ -748,9 +758,9 @@ static intptr_t rktio_icu_convert(rktio_t *rktio,
       if (U_SUCCESS(errorCode)) {
         return 0;
       } else {
-        ICUICONV_SET_ERRNO(U_BUFFER_OVERFLOW_ERROR == errorCode
-                           ? E2BIG
-                           : EBADMSG); /* ? */
+        set_racket_error((U_BUFFER_OVERFLOW_ERROR == errorCode)
+                         ? RKTIO_ERROR_CONVERT_NOT_ENOUGH_SPACE
+                         : RKTIO_ERROR_CONVERT_OTHER); /* ? */
         return RKTIO_CONVERT_ERROR;
       };
     };
@@ -806,7 +816,7 @@ rktio_converter_t *rktio_converter_open(rktio_t *rktio,
 
 void rktio_converter_close(rktio_t *rktio, rktio_converter_t *cvt)
 {
-  if (cvt->tag.is_iconv)
+  if (cvt->is_iconv)
     rktio_iconv_converter_close(rktio, (rktio_iconv_converter_t *) cvt);
   else
     rktio_icu_converter_close(rktio, (rktio_icu_converter_t *) cvt);
@@ -817,7 +827,7 @@ intptr_t rktio_convert(rktio_t *rktio,
                        char **in, intptr_t *in_left,
                        char **out, intptr_t *out_left)
 {
-  if (cvt->tag.is_iconv)
+  if (cvt->is_iconv)
     return rktio_iconv_convert(rktio, (rktio_iconv_converter_t *) cvt, in, in_left, out, out_left);
   else
     return rktio_icu_convert(rktio, (rktio_icu_converter_t *) cvt, in, in_left, out, out_left);
@@ -848,7 +858,10 @@ rktio_convert_result_t *rktio_convert_in(rktio_t *rktio,
 
 void rktio_convert_reset(rktio_t *rktio, rktio_converter_t *cvt)
 {
-  (void)iconv(cvt->cd, NULL, NULL, NULL, NULL);
+  if (cvt->is_iconv)
+    rktio_iconv_convert_reset(rktio, (rktio_iconv_converter_t *) cvt);
+  else
+    rktio_icu_convert_reset(rktio, (rktio_icu_converter_t *) cvt);
 }
 
 /*============================================================*/
--- a/racket/src/rktio/rktio_convert.c
+++ b/racket/src/rktio/rktio_convert.c
@@ -20,6 +20,15 @@
 # include <CoreFoundation/CFLocale.h>
 #endif
 
+#ifndef RKTIO_NO_ICU
+# ifdef RKTIO_SYSTEM_WINDOWS
+#  include <icu.h> /* see comment on init_icu() for why not <icucommon.h> */
+# else
+#  include "unicode/utypes.h" /* Basic ICU data types  */
+#  include "unicode/ucnv.h"   /* C   Converter API     */
+# endif
+#endif
+
 void rktio_convert_init(rktio_t *rktio) {
 #ifdef RKTIO_USE_XLOCALE
   rktio->locale = LC_GLOBAL_LOCALE;
@@ -151,7 +160,7 @@ static void init_iconv()
     }
   }
   
-  if (iconv) {
+  if (iconv) { /* can this be uninitialized? */
     if (hook_handle)
       iconv_errno = (errno_proc_t)rktio_get_proc_address(m, "_errno");
     else
@@ -220,6 +229,89 @@ static void init_iconv() { }
 void rktio_set_dll_path(rktio_char16_t *p) { }
 rktio_char16_t *rktio_get_dll_path(rktio_char16_t *s) { return NULL; }
 
+#endif
+
+/*============================================================*/
+/* ICU as an iconv alternative                                */
+/*============================================================*/
+
+typedef enum {
+  RKTIO_ICU_UNINITIALIZED,
+  RKTIO_ICU_READY,
+  RKTIO_ICU_UNAVAILABLE
+} icu_init_status_t;
+
+#ifdef RKTIO_NO_ICU
+# define icu_init_status RKTIO_ICU_UNAVAILABLE
+static void init_icu() { }
+#else
+# ifndef RKTIO_ICU_DLL
+#  define icu_init_status ICU_READY
+static void init_icu() { }
+# else
+typedef UConverter (*ucnv_open_proc_t)(const char *converterName, UErrorCode *err);
+typedef void (*ucnv_close_proc_t)(UConverter *converter);
+typedef void (*ucnv_reset_proc_t)(UConverter *converter);
+typedef void (*ucnv_convertEx_proc_t)(UConverter *targetCnv, UConverter *sourceCnv,
+                                      char **target, const char *targetLimit,
+                                      const char **source, const char *sourceLimit,
+                                      UChar *pivotStart,
+                                      UChar **pivotSource, UChar **pivotTarget,
+                                      const UChar *pivotLimit,
+                                      UBool reset, UBool flush, UErrorCode *pErrorCode);
+static ucnv_open_proc_t ucnv_open;
+static ucnv_close_proc_t ucnv_close;
+static ucnv_reset_proc_t ucnv_reset;
+static ucnv_convertEx_proc_t ucnv_convertEx;
+
+static icu_init_status_t icu_init_status = RKTIO_ICU_UNINITIALIZED;
+
+static void init_icu()
+{
+#  ifndef RKTIO_SYSTEM_WINDOWS
+#   error "RKTIO_ICU_DLL currently requires RKTIO_SYSTEM_WINDOWS"
+  /* would be something like dlopen("libicuuc.so", ...) */
+#  endif  
+  HMODULE m = NULL;
+  wchar_t *p;
+
+  /* Since Windows 10 version 1903, icu.dll is provided as a system DLL.
+     We could push support back to 1703 by using icuuc.dll (and icucommon.h instead of icu.h),
+     but we would need to arrange to call CoInitializeEx from each thread before usinc ICU,
+     which is not needed with icu.dll.
+     https://learn.microsoft.com/en-us/windows/win32/intl/international-components-for-unicode--icu-
+
+     Is this applicable? https://unicode-org.atlassian.net/browse/ICU-21380
+   */
+
+  p = rktio_get_dll_path(L"icu.dll");
+  if (p) {
+    m = LoadLibraryW(p);
+    free(p);
+  } else 
+    m = NULL;
+
+  if (!m)
+    m = LoadLibraryW(L"icu.dll");
+
+  if (m) {
+    ucnv_open = (ucnv_open_proc_t)GetProcAddress(m, "ucnv_open");
+    ucnv_close = (ucnv_close_proc_t)GetProcAddress(m, "ucnv_close");
+    ucnv_reset = (ucnv_reset_proc_t)GetProcAddress(m, "ucnv_reset");
+    ucnv_convertEx = (ucnv_convertEx_proc_t)GetProcAddress(m, "ucnv_convertEx");
+  }
+  
+  if (!ucnv_open || !ucnv_close || !ucnv_reset || !ucnv_convertEx) {
+    ucnv_open = NULL;
+    ucnv_close = NULL;
+    ucnv_reset = NULL;
+    ucnv_convertEx = NULL;
+    icu_init_status = RKTIO_ICU_UNAVAILABLE;
+  }
+
+  icu_init_status = RKTIO_ICU_READY;
+}
+# endif
 #endif
 
 /*============================================================*/
@@ -476,13 +568,34 @@ char *rktio_system_language_country(rktio_t *rktio)
 /*============================================================*/
 
 struct rktio_converter_t {
-  iconv_t cd;
+  rktio_bool_t is_iconv;
 };
 
-rktio_converter_t *rktio_converter_open(rktio_t *rktio, const char *to_enc, const char *from_enc)
+typedef struct rktio_iconv_converter_t {
+  rktio_converter_t tag;
+  iconv_t cd;
+} rktio_iconv_converter_t;
+
+#ifdef RKTIO_NO_ICU
+typedef rktio_char16_t UChar;
+typedef intptr_t UConverter;
+#endif
+#define ICU_BUF_SIZE 1024
+typedef struct rktio_icu_converter_t {
+  rktio_converter_t tag;
+  UConverter *sourceCnv;
+  UConverter *targetCnv;
+  UChar *pivotSource; /* pointer info buf */
+  UChar *pivotTarget; /* pointer info buf */
+  UChar buf[ICU_BUF_SIZE];
+} rktio_icu_converter_t;
+
+static rktio_iconv_converter_t *rktio_iconv_converter_open(rktio_t *rktio,
+                                                           const char *to_enc,
+                                                           const char *from_enc)
 {
   iconv_t cd;
-  rktio_converter_t *cvt;
+  rktio_iconv_converter_t *cvt;
 
   if (!iconv_ready) init_iconv();
 
@@ -493,21 +606,22 @@ rktio_converter_t *rktio_converter_open(rktio_t *rktio, const char *to_enc, cons
     return NULL;
   }
 
-  cvt = malloc(sizeof(rktio_converter_t));
+  cvt = malloc(sizeof(rktio_iconv_converter_t));
+  cvt->tag.is_iconv = 1;
   cvt->cd = cd;
   return cvt;
 }
 
-void rktio_converter_close(rktio_t *rktio, rktio_converter_t *cvt)
+static void rktio_iconv_converter_close(rktio_t *rktio, rktio_iconv_converter_t *cvt)
 {
   iconv_close(cvt->cd);
   free(cvt);
 }
 
-intptr_t rktio_convert(rktio_t *rktio,
-                       rktio_converter_t *cvt,
-                       char **in, intptr_t *in_left,
-                       char **out, intptr_t *out_left)
+static intptr_t rktio_iconv_convert(rktio_t *rktio,
+                                    rktio_iconv_converter_t *cvt,
+                                    char **in, intptr_t *in_left,
+                                    char **out, intptr_t *out_left)
 {
   size_t il = *in_left, ol = *out_left, r;
   int icerr;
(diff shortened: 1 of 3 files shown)


otherjoel:
> * On Windows, automatically enable ICU if `AC_CHECK_HEADER([icu.h])` finds it. To support building in older environments, don't build ICU support if the header is not found. To support running on older Windows versions, dynamically load `icu.dll` with `LoadLibraryW`/`GetProcAddress`. This means no linking flags are needed.

Just to be clear, it sounds from this that the minimum version of Windows required to run the official distributions of Racket available from `download.racket-lang.org` would **not** increase? (Currently that minimum version is [given][1] as Windows 7, although for 8.8 and 8.9 it was given as Windows 10. There was some discussion in #4542 about what versions of Windows are actually supported.)

[1]: https://download.racket-lang.org/releases/8.11.1/

mflatt:
I don't think we need to support compile-time configuration of ICU for Windows. It should just be always on in DLL mode and used as a fallback when iconv isn't available. (It looks like things are in place where we could avoid `#include`ing `icu.h` on Windows even if it's available, but I'm not sure I'm reading things right.) So, `--enable-icu-dll` could go away and `winfig.bat` wouldn't need to change. Maybe one day we'll switch the default from iconv to ICU.

An `--enable-icu` for other platforms makes sense, and the various `configure` improvements there look good.

LiberalArtist:
> I don't think we need to support compile-time configuration of ICU for Windows. &hellip; So, `--enable-icu-dll` could go away and `winfig.bat` wouldn't need to change. Maybe one day we'll switch the default from iconv to ICU.

That would certainly make things easier.

> (It looks like things are in place where we could avoid `#include`ing `icu.h` on Windows even if it's available, but I'm not sure I'm reading things right.)

I initially thought this wouldn't work, but then I ended up getting halfway there anyway, and now I think it probably would work, so I will try filling in the rest and see how it goes.

I think the biggest thing that had seemed problematic to me originally were comments pointing to https://unicode-org.atlassian.net/browse/ICU-12420 like the one deprecating `U_STANDARD_ERROR_LIMIT` because `The numeric value may change over time`. Now, though, I think that issue is actually specific to constants that were meant to mark the end of open-ended lists. The numeric values of the actual error codes we need to test, including for `U_SUCCESS` and `U_FAILURE`, do seem to be covered by ICU's ABI stability.

I had also seen that there is some complicated logic allowing for alternate definitions of `UChar`. I haven't fully understood it, but it seems like it is for niche situations that wouldn't apply here.

mflatt:
@LiberalArtist Did you decide that this wasn't worthwhile, or just lose interest, or just run out of time?

I remain interested in this improvement for both Racket and Chez Scheme — though admittedly not enough to try anything myself, so far.

LiberalArtist:
I just ran out of time and haven't gotten back to it. I still think it's a viable approach. It is on my to-do stack to get back to, but that won't be for the next month or so, at least. If someone else wants to fill things in before then, so much the better!

LiberalArtist:
I started looking at this again and found I had actually done but not committed much of the work.

Looking at the CI errors <https://github.com/racket/racket/actions/runs/16229453902/job/45828973905?pr=4844#step:8:79>:
```
Errors were [4]:
((portlib) ((#"\357\277\275\357\277\275") Error (port->bytes (reencode-input-port (open-input-bytes #"\377\377") "utf-8"))))
((portlib) ((not-an-error) Error (let () (port->bytes (reencode-input-port (open-input-bytes #"\377\377") "utf-8")) (quote not-an-error))))
((portlib) ((#<void>) Error (let ((o (reencode-output-port (open-output-bytes) "utf-8"))) (display #"\377\377" o) (flush-output o))))
((portlib) ((not-an-error) Error (let () (let ((o (reencode-output-port (open-output-bytes) "utf-8"))) (display #"\377\377" o) (flush-output o)) (quote not-an-error))))
```
Are these cases that were expected to be errors but now work? I'm having trouble locating the tests.

samth:
Those tests are here https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/portlib.rktl

LiberalArtist:
I found the file, I just haven't found the specific tests. For example, grepping that file for `not-an-error` doesn't find any results.

samth:
That in particular is from the expansion of err/rt-test. My guess is the relevant lines are around here: https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/portlib.rktl#L964

LiberalArtist:
Some notes on my investigation so far:

As @samth suggested, the failing tests are specifically: https://github.com/racket/racket/blob/52e33046eb195311c142bab45473883e8f5103ed/pkgs/racket-test-core/tests/racket/portlib.rktl#L973-L981

The relevant parts boil down to this expected behavior, i.e. catching invalid encoding:
```racket
> (bytes-convert (bytes-open-converter "UTF-8" "UTF-8") #"\xFF\xFF")
#""
0
'error
```

(Note that `#"\377\377"` is the way the printer writes `#"\xFF\xFF"`.)

One surprise is that this goes through the iconv/ICU path at all: `(bytes-open-converter "UTF-8" "UTF-8")` is one of the guaranteed combinations, even when no iconv/ICU library is available, so there is built-in support for this somewhere. Does the implementation switch to use always use the foreign library if present, even for guaranteed combinations? Something to investigate further.

Nonetheless, I think we do want to understand what's going on in these test failures before merging.

I noticed that:
```racket
> (bytes-convert (bytes-open-converter "UTF-8-permissive" "UTF-8") #"\xFF\xFF")
#"\357\277\275\357\277\275"
2
'complete
```
matching the unexpected result in the first test case.

So it seems to me that we are currently calling ICU in permissive mode, and need to figure out how to not do that.

LiberalArtist:
Apparently ICU's default behavior is to emit a substitution character for an invalid sequence, but setting callbacks can override this. I've pushed a first attempt.

LiberalArtist:
That took care of half of the test failures. What remains:
```
Errors were [2]:
((portlib) ((#<void>) Error (let ((o (reencode-output-port (open-output-bytes) "utf-8"))) (display #"\377\377" o) (flush-output o))))
((portlib) ((not-an-error) Error (let () (let ((o (reencode-output-port (open-output-bytes) "utf-8"))) (display #"\377\377" o) (flush-output o)) (quote not-an-error))))
```

Also, we should double-check that I resolved the merge conflicts from the new parallel threads correctly.

LiberalArtist:
> One surprise is that this goes through the iconv/ICU path at all: `(bytes-open-converter "UTF-8" "UTF-8")` is one of the guaranteed combinations, even when no iconv/ICU library is available, so there is built-in support for this somewhere. Does the implementation switch to use always use the foreign library if present, even for guaranteed combinations? Something to investigate further.

Apparently the matching for the guaranteed combinations is case-sensitive, which is presumably why the tests use `"utf-8"`:

https://github.com/racket/racket/blob/52e33046eb195311c142bab45473883e8f5103ed/racket/src/io/converter/main.rkt#L35-L68

LiberalArtist:
While doing a build on Debian for debugging, I found that there are some uses of encodings called `UCS-4LE` or `UCS-4BE`. It looks like they are defined in the same place as the implementation of [`_string/ucs-4`](https://docs.racket-lang.org/foreign/String_Types.html#%28def._%28%28quote._~23~25foreign%29.__string%2Fucs-4%29%29), but not actually used by it, at least on CS. The specific places are the ones changed in https://github.com/LiberalArtist/racket/commit/bbd48c7ca4f1f09e21e039063d10c51a02119ab5.

At some point in the process of building Racket, something tries to use `bytes-open-converter` to convert between UTF-8 and `UCS-4{L,B}E` (the one corresponding to the native endianness). This is a problem, because ICU doesn't recognize either of those as encoding names.

In more detail, these are the related encodings recognized by ICU vs. GNU libiconv built with `--enable-extra-encodings`. Names listed on the same line are aliases, and matching is case-insensitive (and flexible in other details I don't recall off-hand):
```text
;; ICU
UTF-32 ISO-10646-UCS-4 ibm-1236 ibm-1237 csUCS4 ucs-4 
UTF-32BE UTF32_BigEndian ibm-1232 ibm-1233 ibm-9424 
UTF-32LE UTF32_LittleEndian ibm-1234 ibm-1235 
UTF32_PlatformEndian 
UTF32_OppositeEndian
;; GNU libiconv w/ --enable-extra-encodings
ISO-10646-UCS-4 UCS-4 CSUCS4
UCS-4BE
UCS-4LE
UTF-32
UTF-32BE
UTF-32LE
UCS-4-INTERNAL
UCS-4-SWAPPED
```
[*The Unicode Standard,* Version 17.0.0, Appendix C&sect;2, &ldquo;Encoding Forms in ISO/IEC 10646&rdquo;](https://www.unicode.org/versions/Unicode17.0.0/core-spec/appendix-c/#G8861) says:
> ISO/IEC 10646:2011 has significantly revised its discussion of encoding forms, compared to earlier editions of that standard. The terminology for encoding forms (and encoding schemes) in 10646 now matches exactly the terminology used in the Unicode Standard. Furthermore, 10646 is now described in terms of a codespace U+0000..U+10FFFF, instead of a 31-bit codespace, as in earlier editions. This convergence in codespace description has eliminated any discrepancies in possible interpretation of the numeric values greater than 0x10FFFF. As a result, this section now merely notes a few items of mostly historic interest regarding encoding forms and terminology.
>
> **UCS-4.** UCS-4 stands for “Universal Character Set coded in 4 octets.” It is now treated simply as a synonym for UTF-32, and is considered the canonical form for representation of characters in 10646.

Furthermore, neither the [IANA Character Sets Registry](https://www.iana.org/assignments/character-sets/character-sets.xhtml) nor the [WHATWG Encoding Standard](https://encoding.spec.whatwg.org/#names-and-labels) recognizes endianness-specific variants of UCS-4. (WHATWG doesn't recognize any variant of UCS-4 as a name.) In brief searching, I can see a few, perhaps older, pieces of software that mention the endianness-specific variants, but only a few, and (while I haven't read it) I'm guessing they may not actually have been defined by ISO 10646.

The only reason I can guess for GNU libiconv to treat `UTF-32` and `UCS-4` as distinct encodings, rather than aliases, is that it might support, or once have supported, older versions of ISO 10646 in which only a subset of UTF-32 was valid UCS-4, and it could have raised errors if anything out of that subset was used.

**My conclusion is that we ought to just change to `UTF-32LE` and `UTF-32BE`.** That's what I've done in https://github.com/LiberalArtist/racket/commit/bbd48c7ca4f1f09e21e039063d10c51a02119ab5 (part of this PR, likely to be squashed eventually). I haven't done it yet, but I also think we should update the docs for `_string/ucs-4` and `_string/utf-16`, neither of which say anything about endianness or the lack of a BOM.

The existence of this code reinforces my thinking from https://github.com/LiberalArtist/racket/commit/bbd48c7ca4f1f09e21e039063d10c51a02119ab5 that we ought to support all of the Unicode variants without relying on an iconv implementation. But I'm not intending to look at that until after ICU support is done, because I don't want to mask any errors.

LiberalArtist:
These equivalent expressions:
```
(bytes-convert (bytes-open-converter "utf-8" "UTF-8") #"\xFF\xFF")
(bytes-convert (bytes-open-converter "UTF-8" "utf-8") #"\xFF\xFF")
```
with ICU both evaluate to:
```
#""
1
'error
```
but with iconv evaluate to:
```
#""
0
'error
```

I used DrRacket's step debugger in an `--enable-icu` build on the failing test:
```
(let ([o (reencode-output-port (open-output-bytes) "utf-8")]) 
  (display #"\xFF\xFF" o)
  (flush-output o))
```

and found that it got to this code:
https://github.com/racket/racket/blob/20b2596d695216e08d22cdaeb3ff97b18c86c106/racket/collects/racket/port.rkt#L2014-L2034
with:
```
got-c => 0
used-c => 1
status => error
```
so, at line 2022, `(zero? used-c)` is `#f`, and the block labeled `No progress before an encoding error.` never executes.

The test reaches line 2022 twice, then gets to this block of code:
https://github.com/racket/racket/blob/20b2596d695216e08d22cdaeb3ff97b18c86c106/racket/collects/racket/port.rkt#L1758-L1771
with `(= start end)` and `(buffer-flushed?)` both returning `#t`, so it returns 0 at line 1770.

----

There doesn't seem to be any path in the code that leads to an exception when there *persistently is* `progress before an encoding error`.

I'm not sure why the ICU and iconv backends are returning different second results from `(bytes-convert (bytes-open-converter "utf-8" "UTF-8") #"\xFF\xFF")`. It's very possible that there could be something wrong in my C code, but I can also imagine that the iconv API may not make a portable guarantee about this. I will investigate further, though I'd be glad for any suggestions.

Note that using capitalized `"UTF-8"` in both cases *does* in fact get Racket's built-in implementation, so this happens even *with* `--enable-icu`:
``
> (bytes-convert (bytes-open-converter "UTF-8" "UTF-8") #"\xFF\xFF")
#""
0
'error
```
After ICU support is in place, this further suggests to me that we should support all of the Unicode variants without relying on a foreign library.

**Comment by LiberalArtist** [comment]
Re parallel threads, as long as this still holds:
https://github.com/racket/racket/blob/20b2596d695216e08d22cdaeb3ff97b18c86c106/racket/src/rktio/rktio.h#L72-L76
it looks like it should still be safe to initialize static variables in the first `rktio_init()`.

**Comment by LiberalArtist** [comment]
The docs for [`ucnv_convertEx()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#a8c2852929b99ca983ccd1f33a203cc2a) say that, on output:

> `*source` points to the first unit after the last `char` read

That is less than clear about whether "the last `char` read" is a `char` read from the erroneous character, or if it's the last `char` *successfully* read. However, [an article by JeanHeyd Meneide](https://thephd.dev/the-c-c++-rust-string-text-encoding-api-landscape#updating-the-pointer-value-makes-sure-you-know-how-much-input-wa) (the editor of the ISO C standard) says:

> An even better version of this updates the pointer only if the operation was successful, rather than just doing it willy-nilly. This is the way ICU works, and it is incredibly helpful when `*pErrorCode` is not set to `U_SUCCESS`. It allows you to insert replacement characters and/or skip over the problematic input if needed, allowing for greater control over how errors are handled.

But my understanding is that the byte `#"\xFF"` cannot appear *anywhere* in UTF-8, so I don't see how it can possibly have been successfully read.

**Comment by LiberalArtist** [comment]
For another case, consider `(bytes-convert (bytes-open-converter "UTF-8" "utf-8") #"abc\xF0\x9F\xFF")`.  That byte string comes from:
```racket
(bytes-append #"abc"
              (let ([bs (string->bytes/utf-8 "🎜")])
                (subbytes bs 0 (/ (bytes-length bs) 2)))
              #"\xFF")
```
i.e., chopping off the second half of `#\🎜` in UTF-8 and replacing it with the invalid `#"\xFF"`.
The results we get are:
```
;; ICU
#"abc"
5
'error
;; iconv
#"abc"
3
'error
```
The iconv backend counts only the three bytes that were successfully converted, whereas the ICU backend counts the 5 bytes that *could* be part of valid UTF-8, but not, in this case, the bad `#"\xFF"` (why?).

It looks like the lower-level functions [`ucnv_fromUnicode()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#aa820d3bc3942522eb31bdb5b8ae73727) and [`ucnv_toUnicode()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#a9451f05be7b1b75832d5ec55b4e6d67f) provide enough granular offset information that this should be implementable, but I really thought `ucnv_convertEx` was supposed to have this capability.

LiberalArtist:
I've put together a little C program to verbosely explore the various input/output pointers and results: https://gist.github.com/LiberalArtist/3f32e8c0b4d9a7db2a23a0cafe444d12

LiberalArtist:
I found some comments about the state during error callbacks and where the various pointers are pointing:
https://github.com/unicode-org/icu/blob/4974e9b262366fe5b16eb10293c35172d55ce2a9/icu4c/source/common/unicode/ucnv_cb.h#L29-L59
I think our problem will be solvable by supplying custom callbacks.

LiberalArtist:
We now have a different test failure:
```racket
((portlib) ((EXN #(struct:exn:fail:contract "user port peek: contract violation\n  expected: \n   (or/c exact-nonnegative-integer? eof-object? evt? pipe-input-port? #f procedure?)\n  received: #<void>" #<continuation-mark-set>)) #f (#<procedure:regexp-match?> #px#"temple" #<input-port:string>)))
```
The test in question is here:
https://github.com/racket/racket/blob/20b2596d695216e08d22cdaeb3ff97b18c86c106/pkgs/racket-test-core/tests/racket/portlib.rktl#L983-L1004
But the contract violation is from an internal function not directly related to byte conversion. I'm not sure how anything I changed could have affected that. Also, the previous test failures I've been able to reproduce in a build on Debian with `--enable-icu`, but this test passes on my machine.

LiberalArtist:
# Status Update

I put some time into this last month, inspired by RacketCon, with the thought that 9.0 would have been an auspicious time for the switch, but there is more to do. Aside from tracking down the test failure in https://github.com/racket/racket/pull/4844#issuecomment-3408102873, there is a bigger-picture issue I've been thinking about.

ICU's converters are stateful. In particular, they use a pivot buffer containing an internal representation (usually UTF-16 for historical reasons, but IIUC there are fast paths for UTF-8) when converting from a source encoding to a target encoding. [This part of the documentation](https://unicode-org.github.io/icu/userguide/conversion/converters.html#conversion) is relevant:

> The converters always consume the source buffer as far as possible, and advance the source pointer.
>
> The converters write to the target all converted output as far as possible, and then write any remaining output to the internal services buffer. When the conversion routines are called again, the internal buffer is flushed out and written to the target buffer before proceeding with any further conversion.

Thus, from ICU's perspective, there may exist bytes that have been consumed from the source but not yet written to the target.

Initially, I thought this would be trivial: after all, in the general case, correctly using the byte converter API requires using [`bytes-convert-end`](https://docs.racket-lang.org/reference/bytestrings.html#%28def._%28%28quote._~23~25kernel%29._bytes-convert-end%29%29) to potentially get more output after all input has been consumed. 

On reflection, though, the qualifier &ldquo;in the general case&rdquo; seems significant. The `bytes-convert-end` docs say:

> instead of converting bytes, this procedure generates an ending sequence for the conversion (sometimes called a “shift sequence”), if any. Few encodings use shift sequences, so this function will succeed with no output for most encodings. In any case, successful output of a (possibly empty) shift sequence resets the converter to its initial state.

By implication, a user who knows that the *specific* encodings in some application do not use shift sequences would seem justified in never calling `bytes-convert-end`. In other words, `bytes-convert-end` is not allowed to emit anything other than a (possibly empty) shift sequence.

On further investigation, the requirement seems to follow from [the POSIX `iconv` specification](https://pubs.opengroup.org/onlinepubs/9799919799/functions/iconv.html), which says:

> For state-dependent encodings, the conversion descriptor `cd` is placed into its initial shift state by a call for which `inbuf` is a null pointer, or for which `inbuf` points to a null pointer. When `iconv()` is called in this way, and if `outbuf` is not a null pointer or a pointer to a null pointer, and `outbytesleft` points to a positive value, `iconv()` shall place, into the output buffer, the byte sequence to change the output buffer to its initial shift state.

This part of the POSIX spec explicitly requires that input be counted as consumed only when corresponding output has been produced, and it seems to ban any state other than shift sequences in conversion descriptors:

> If the output buffer is not large enough to hold the entire converted input, conversion shall stop just prior to the input bytes that would cause the output buffer to overflow. The variable pointed to by `inbuf` shall be updated to point to the byte following the last byte successfully used in the conversion. The value pointed to by `inbytesleft` shall be decremented to reflect the number of bytes still not converted in the input buffer. &hellip; For state-dependent encodings, the conversion descriptor shall be updated to reflect the shift state in effect at the end of the last successfully converted byte sequence.

Reluctantly, I am coming to the conclusion that we shouldn't use `bytes-convert-end` to flush an ICU converter's pivot buffer. While hypothetically we *could* change the specification of the Racket API, and existing uses that correctly handle the general case would continue to be correct, the whole point of this exercise in supporting niche non-Unicode encodings is backward compatibility.

It **definitely is** still possible to use ICU to implement the `iconv` API.

In the worst-case scenario, we could use [preflighting mode](https://unicode-org.github.io/icu/userguide/conversion/converters.html#4-pre-flighting), or possibly [character mode](https://unicode-org.github.io/icu/userguide/conversion/converters.html#2-character).

Ideally, though, we would want to stay within [buffered or streamed mode](https://unicode-org.github.io/icu/userguide/conversion/converters.html#3-buffered-or-streamed). I'm still trying to figure out how to do that.

[This mailing list thread](https://groups.google.com/u/2/a/unicode.org/g/icu-support/c/K0eT_iCX0ss/) is one part I want to better understand. The documentation for [`ucnv_fromUCountPending()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#adb065b7305e22375f6616a7f01e107be) and [`ucnv_toUCountPending()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#ad07fce62cb530974aab903210dbb62b1) says they are &ldquo;useful for mapping semantics of ICU's converter interface to those of `iconv`.&rdquo;

A lingering concern, though, is that some of the suggestions seem to work like this: Imagine an identity conversion, `UTF-8` to `UTF-8`, with `abcdefg` as the initial input, an empty two-byte target buffer, and an empty two-byte pivot buffer. After the first call, `ab` will be in the target buffer, `cd` will be in the pivot buffer, and, from ICU's perspective, `efg` will not have been consumed. An adapter layer could report instead, per `iconv` semantics, that only `ab` were consumed. Then, the client would make a **second call** with `cdefg` as the input and, let's say, another empty two-byte target buffer. The adapter layer would offset the input by the content of the pivot buffer, calling ICU with just `efg` as input and `cd` in the pivot. The result would have `cd` in the output buffer, `ef` in the pivot buffer, and, from ICU's perspective, just `g` not having been consumed.

The problem I see with that approach is that, in the **second call**, there is a requirement that the input *must* start with `cd` to match the content of the pivot buffer. Most of the time that will obviously be true, especially in reasonable, practical use-cases. But I don't think the `iconv` semantics *require* that to be true. Instead, a client could supply `vwxyz` for the **second call** and end up with `cd` wrongfully emitted, `vw` discarded completely, and  `xy` in the pivot buffer.

Another approach could use [`ucnv_getMaxCharSize()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#af43eaa49c8f0e9bd5c63ace95b014d8b) and [`ucnv_getMinCharSize()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#a5cca1b29e06b6423cd29e052f63f3c68) to calculate a safe buffer size for batch processing, perhaps in combination with [`ucnv_safeClone()`](https://unicode-org.github.io/icu-docs/apidoc/dev/icu4c/ucnv_8h.html#a11a66c267dce12a78dac2690d983d69e) to do limited preflighting or character-by-character conversion only when needed. This might have better performance than using preflighting mode or character mode all the time. The downside would be a potentially more complex implementation in `rktio`.

racket/src/rktio/rktio_convert.c
```diff
@@ -220,14 +241,158 @@ void rktio_set_dll_path(rktio_char16_t *p)
...
+static ucnv_setFromUCallBack_proc_t ucnv_setFromUCallBack = NULL;
+static UConverterFromUCallback UCNV_FROM_U_CALLBACK_STOP = NULL;
+#endif
+
+static void init_icu()
+{
+  /* Called from rktio_convert_init(rktio), which is called from rktio_init().
+     The ordering requirement on the first call to rktio_init() guarantees
+     exclusive access to static variables here.
+  */
```
LiberalArtist: @mflatt Is this still true after the changes for parallel threads?

PR #4845, opened 2023-12-03
repair `for/fold` binding and `#:break`

mflatt:
*[Draft until Typed Racket can deal with this expansion, either through changes here or in Typed Racket.]*

Besides fixing `#:break` when an accumulator variable is shadowed by a `for/fold` clause or local definition, this commit adjusts the way that accumulator variables are bound so that they work more sanely with shadowing. It also evaluates accumulator initialization expressions before the first set of iteration clauses.

An unusual aspect of `for/fold` is that the lifetimes of variables do not line with with textual nesting, but scope does --- or, at least, does after this commit. For example, in

 (for*/fold ([a ....])
            ([i ....]
             [j ....])
    ....)

you would expect that `a`, `i`, and `j` are bound for everything to the right of the binding. There's a new `i` and `j` for each iteration, but each `i` lives as long as all the `j`s` inside. But the liftime of `a` is inside the liftimes of `i` and `j`, insead of outside.

Before this commit, `a` wasn't exactly bound this way, and shadowing throught a `for`-clause binding acted weirdly: the value of the shadowed variable taken from the accumulator on the first iteration of a loop and the shadowing binding for subsequent loops. This weirdness interacted with the implementation of `#:break`, so that it didn't work right.

The new implementation binds each accumulator variable as a macro that refers to a suitable binding for the current iteration, essentially like using a syntax parameter (but implemented for this limited context in a more direct way).

A remaining weirdness is that an accumulator variable is not bound of the right-hande size of clauses for the ones that determine an outermost iteration. Unfortunately, change that breaks too much code, and in a way that can be particularly difficult to track down. Add `#:when? #t` as the first clause to avoid that binding weirdness.

##### Checklist

- [ ] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation


racket/collects/racket/private/for.rkt
```diff
@@ -1801,43 +1828,54 @@
...
-                         (values* fold-var ...)
-                         . rest)))
-                   (if result-expr
-                       (quasisyntax/loc #'orig-stx
-                         (let-values ([(fold-var ...) #,loop-stx])
-                           #,result-expr))
-                       loop-stx)])))]
+               (with-syntax ([(int-var ...)
+                              (map (lambda (fold-var)
+                                     (datum->syntax fold-var (gensym (syntax-e fold-var))))
```
samth, line 1833: I think the problem is just that we need to copy the syntax properties from `fold-var` onto `int-var` (and that fixes my initial testing). Is that generally reasonable?

mflatt: Can you send your change that made things work? I tried
```
(datum->syntax fold-var (gensym (syntax-e fold-var)) fold-var fold-var)
```
in place of
```
(datum->syntax fold-var (gensym (syntax-e fold-var)))
```
but was not able to build the `math` collection after that change.

PR #4847, opened 2023-12-03
Add build stamp to `--version` output

jryans:
This adds the build stamp value to Racket's `banner` and `--version` output when the value is non-empty. This is helpful as a quick and easy way to distinguish snapshot builds and verify the build date. This matches the behaviour of other compilers which often indicate a detailed version with a date or commit hash for development / snapshot builds.

### DIFF
--- a/racket/src/cs/main.sps
+++ b/racket/src/cs/main.sps
@@ -793,7 +793,10 @@
        (set-collects-dir! init-collects-dir)])
      (set-config-dir! init-config-dir)
      (let* ([config (read-installation-configuration-table)]
-            [name (get-installation-name config)])
+            [name (get-installation-name config)]
+            [build-stamp (or (hash-ref config 'build-stamp #f) "")])
+       (unless (equal? build-stamp "")
+         (set-build-stamp! build-stamp))
        (unless (eq? init-collects-dir 'disable)
          (current-library-collection-links
           (find-library-collection-links config name))
@@ -929,14 +932,15 @@
          (when (and n (exact-nonnegative-integer? n))
            (set-schedule-quantum! n)))))
 
-   (when version?
-     (display (banner)))
    (call/cc ; Chez Scheme's `call/cc`, used here to escape from the Racket-thread engine loop
     (lambda (entry-point-k)
       (call-in-main-thread
        (lambda ()
          (initialize-exit-handler!)
          (initialize-place!)
+
+         (when version?
+           (display (banner)))
 
          (when (and make? (not (null? compiled-file-paths)))
            (|#%app|
--- a/pkgs/racket-doc/scribblings/raco/config.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/config.scrbl
@@ -233,7 +233,11 @@ directory}:
  @item{@indexed-racket['build-stamp] --- a string that identifies a build,
        which can be used to augment the Racket version number to more
        specifically identify the build. An empty string is normally
-       appropriate for a release build.}
+       appropriate for a release build. The default @racket{banner}
+       also shows the build stamp when non-empty.}
+
+       @history[#:changed "8.11.1.7" @elem{Added build stamp to
+                                           @racket{banner}.}]
 
  @item{@indexed-racket['absolute-installation?] --- a boolean that is
        @racket[#t] if the installation uses absolute path names,
--- a/racket/src/bc/cmdline.inc
+++ b/racket/src/bc/cmdline.inc
@@ -1227,19 +1227,6 @@ static int run_from_cmd_line(int argc, char *_argv[],
 
   if (no_init_ns)
     init_ns = 0;
-
-  if (show_vers) {
-#ifndef RACKET_CMD_LINE  
-    if (!use_repl
-#ifdef CMDLINE_STDIO_FLAG
-	|| alternate_rep
-#endif
-	)
-#endif
-      PRINTF("%s", BANNER);
-
-   CMDLINE_FFLUSH(stdout);
-  }
 #endif /* DONT_PARSE_COMMAND_LINE */
 
   if (!syslog_level) {
@@ -1406,6 +1393,34 @@ static int run_from_cmd_line(int argc, char *_argv[],
   if (!skip_coll_dirs)
     scheme_init_collection_paths_post_config(global_env, collects_paths_l, collects_paths_r, config_table);
   scheme_init_compiled_roots_config(global_env, compiled_paths, config_table);
+
+  {
+    Scheme_Object *build_stamp;
+    build_stamp = scheme_hash_tree_get((Scheme_Hash_Tree *)config_table, scheme_intern_symbol("build-stamp"));
+    if (build_stamp) {
+      char *s;
+      intptr_t s_len;
+      s = scheme_display_to_string(build_stamp, &s_len);
+      if (s_len) {
+        scheme_set_build_stamp(s);
+      }
+    }
+  }
+
+#ifndef DONT_PARSE_COMMAND_LINE
+  if (show_vers) {
+#ifndef RACKET_CMD_LINE
+    if (!use_repl
+#ifdef CMDLINE_STDIO_FLAG
+        || alternate_rep
+#endif
+        )
+#endif
+      PRINTF("%s", BANNER);
+
+    CMDLINE_FFLUSH(stdout);
+  }
+#endif
 
   scheme_seal_parameters();
 
--- a/racket/src/bc/include/scheme.h
+++ b/racket/src/bc/include/scheme.h
@@ -1941,6 +1941,7 @@ MZ_EXTERN void scheme_set_stdio_makers(Scheme_Stdio_Maker_Proc in,
 
 
 MZ_EXTERN void scheme_set_banner(char *s);
+MZ_EXTERN void scheme_set_build_stamp(char *s);
 MZ_EXTERN Scheme_Object *scheme_set_exec_cmd(char *s);
 MZ_EXTERN Scheme_Object *scheme_set_run_cmd(char *s);
 MZ_EXTERN void scheme_set_collects_path(Scheme_Object *p);
--- a/racket/src/bc/src/string.c
+++ b/racket/src/bc/src/string.c
@@ -197,6 +197,8 @@ READ_ONLY Scheme_Object *scheme_zero_length_char_immutable_string;
 READ_ONLY Scheme_Object *scheme_zero_length_byte_string;
 
 SHARED_OK static char *embedding_banner;
+SHARED_OK static char *build_stamp;
+SHARED_OK static char build_stamp_banner[128];
 SHARED_OK static Scheme_Object *vers_str;
 SHARED_OK static Scheme_Object *banner_str;
 
@@ -222,6 +224,8 @@ THREAD_LOCAL_DECL(static void *current_locale_name_ptr);
 static void reset_locale(void);
 
 #define current_locale_name ((const mzchar *)current_locale_name_ptr)
+
+static void update_banner(void);
 
 static const mzchar empty_char_string[1] = { 0 };
 static const mzchar xes_char_string[2] = { 0x78787878, 0 };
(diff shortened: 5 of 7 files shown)


jryans:
> This needs a change for BC as well, right?

Ah yeah, seems good to add over there as well, will do!

jryans:
Updated as suggested. BC now has this behaviour as well.

LiberalArtist:
When `(get-build-stamp)` returns `"20231203-4f7eb34066"`, DrRacket shows:
```
Welcome to DrRacket, version 8.11.1.3--2023-12-03(4f7eb34066/a) [cs].
```
It seems like it would be better to keep the formats consistent.

It looks like this is where the formatting happens for DrRacket, though I haven't understood all the design considerations: https://github.com/racket/drracket/blob/edc80d6206c1457b26ae3525a103202310ac205e/drracket/repo-time-stamp/stamp.rkt#L22

jryans:
> When `(get-build-stamp)` returns `"20231203-4f7eb34066"`, DrRacket shows:
> 
> ```
> Welcome to DrRacket, version 8.11.1.3--2023-12-03(4f7eb34066/a) [cs].
> ```
> 
> It seems like it would be better to keep the formats consistent.
> 
> It looks like this is where the formatting happens for DrRacket, though I haven't understood all the design considerations: [racket/drracket@`edc80d6`/drracket/repo-time-stamp/stamp.rkt#L22](https://github.com/racket/drracket/blob/edc80d6206c1457b26ae3525a103202310ac205e/drracket/repo-time-stamp/stamp.rkt#L22)

Ah curious, I wasn't aware that DrRacket was doing its own additional processing.

I can definitely investigate it, perhaps attempting to format the build stamp value in the same way without the additional Git branch detection...? It's slightly more annoying to do the fancier Git detection in the layers this PR is operating at than it is for that existing DrRacket code.

If others have opinions on this, please do let me know. 😄

LiberalArtist:
> It's slightly more annoying to do the fancier Git detection in the layers this PR is operating at than it is for that existing DrRacket code.

I wasn't aware that DrRacket was doing all that extra stuff until commenting on this PR: I just knew snapshots displayed the build stamp somehow. Looking at it slightly more, I'm a little suspicious that it's really doing The Right Thing.

Poking at `git blame` a little, the file was last touched in https://github.com/racket/drracket/commit/058dce9d39dbbcbbd0ac2c265b04669258740392 a decade ago, when it was moved (with no changes) as part of splitting packages into their own Git repositories. I'm not even confident that it would find the right `.git` directory any more: at a minimum, doing so would would be dependent on very particular assumptions about how developers organize their local checkouts.

At a minimum, the [doc.txt](https://github.com/racket/drracket/blob/edc80d6206c1457b26ae3525a103202310ac205e/drracket/repo-time-stamp/doc.txt) file, which says:

> This collection is used only for the repository.  It is not included in normal releases.  (It's existence indicates a checked out tree.)

has been wrong since https://github.com/racket/racket/commit/2e067ab28c2b331713557feb880d8adf026a1aec (a.k.a. https://github.com/racket/drracket/commit/1451bb889cc673051266bd1eeb9d2ac11efb7673), which added the build stamp configuration in the first place.

The strategy for determining the stamp in that file has changed significantly over time. For years, from before the migration from CVS to SVN (see https://github.com/racket/racket/commit/1118b8d59b921d62a02844afd45f0054e8694a9a) through https://github.com/racket/racket/commit/d78928f18d7fc48c2c206bac90018cb5d9373e07 on April 17, 2010, there was a daily commit updating the file. Commit https://github.com/racket/racket/commit/aceeb391adc96d55cee9b8121f7cffbe0f547133 started using the Git attributes [`ident`](https://git-scm.com/docs/gitattributes#_ident) and [`export-subst`](https://git-scm.com/docs/gitattributes#_export_subst).

I don't really have a particular view about what information should be presented or how, but I do think it's probably more useful if the formats are consistent. Maybe we could simplify the behavior used by DrRacket, too? I certainly didn't know what the character between the `/` and the `)` meant until reading the comments in the source, nor why it was there.

It looks like @mflatt made the last substantive change to `stamp.rkt`, and according to https://github.com/racket/racket/commit/1fae7942d583352306e6acb302f98b4639e7eca8 (a.k.a. https://github.com/racket/drracket/commit/61fa510254c3a5c8451b7beade8df3f4af482f41) @samth once had an opinion about this.

LiberalArtist:
> When `(get-build-stamp)` returns `"20231203-4f7eb34066"`, DrRacket shows:
> 
> ```
> Welcome to DrRacket, version 8.11.1.3--2023-12-03(4f7eb34066/a) [cs].
> ```

Well, I do have one opinion: `"20231203-4f7eb34066"` is less readable than `"2023-12-03-4f7eb34066"`.

rfindler:
I doubt anyone relies on parsing out the text that's in the interactions window in drracket, so if changing the format of it makes it easier to make racket and drracket match (say with specific fields that drr has but that racket doesn't) then that'd be welcome!

mflatt:
In a release, I think the build stamp is "". The `--version` output in the PR so far guards against a missing build stamp, but not an empty one (if I'm reading the code right). Probably it should also avoid printing when the stamp is empty.

LiberalArtist:
> > Maybe we could simplify the behavior used by DrRacket, too?
> 
> This works for me too!

With a blank slate, I'd be inclined to give the configured build stamp the last word on what the stamp should be, and to move responsibility for beautifying it (whether as `"2023-12-03-4f7eb34066"`, `"2023-12-03(4f7eb34066)"`, or something else) to the code that sets the stamp. Then, the formatting could be simplified to something like:

```racket
(let ([stamp (or (get-build-stamp) "")])
  (if (equal? "" stamp)
      (version)
      (string-append (version) "-" stamp)))
```

If there's a desire for the kind of additional information DrRacket currently attempts to report when building Racket from Git, it seems like that responsibility could also be moved to the build scripts, which ought to be in a better position to report how Racket is built than a Racket package developed in a separate Git repository would be (especially now that we have a more capable language for build scripts).

jryans:
Thanks all for the feedback! @LiberalArtist’s suggestion above to show the build stamp unmodified as part of the version makes sense to me. I’ll work on adapting the PR to do this.

jryans:
This is ready for another look. 

- The build stamp is now integrated into `banner` and appears just after the `version` portion, with a hyphen between them as @LiberalArtist suggested
- The build stamp is tested for empty string values and only appears when both present and non-empty

As for DrRacket's special build stamp handling, I agree that separate work should remove its customisation and perhaps integrate VCS stamp generation into the build system.

mflatt:
I'm puzzled that the `banner` function moved to the expander. (Was it always that way in this PR?) I thought maybe that was to share the implementation between BC and CS, but there still seems to be a separate BC implementation of `banner`. Either way, the expander doesn't seem to be the right place for the function.

jryans:
Ah yeah, no it was not in the expander before… I moved it there because expander implements the config table reading code. Maybe there’s some way to access that back in the rumble layer where banner used to be…?

jryans:
Ready for another look! I've updated this to leave `banner` back in the Rumble layer and then set the build stamp from `main.sps` as @mflatt suggested. Seems like the right direction to take, and as a nice additional benefit, the BC and CS implementations are now quite similar.

jryans:
Hmm, the Windows-only CS error is a bit puzzling... It seems like we're unable to build some path while cataloging `compiler-lib`... I'm having trouble imagining how the changes here could cause that, maybe it's an intermittent CI issue...?

racket/src/cs/main.sps
```diff
@@ -929,6 +929,10 @@
          (when (and n (exact-nonnegative-integer? n))
            (set-schedule-quantum! n)))))
 
+   (let* ([config (read-installation-configuration-table)]
```
mflatt, line 932: I think this turns out not to be a valid place to call `read-installation-configuration-table`. When I build and run `racketcs` as created in `racket/build/cs/c`, for example, a `build-path` error gets printed as a result of a bad caching of the config directory. The error doesn't happen for the installed `racket`, but that's probably because caching and finding the config directory is complicated.

Similar to the reordering that happens in BC, maybe the best thing here in CS is to have `initialize-place!` return the config table that it reads, and then use that table to extract `'build-stamp` (a few lines below where the banner printout currently happens, inside the `call-on-main-thread` callback). That has the additional benefit of not reading the configuration file twice. I don't think it will be a problem to delay banner output that way.

mflatt (changes requested):

racket/src/cs/main.sps
jryans: Ah okay, I wasn't expecting `racket/src/build/cs/c/racketcs` and `racket/bin/racket` to have different behaviour here, but indeed they do. Thanks for the tip, I'm slowly understanding how all these startup paths work together. 🙂 

I've moved the config table portion inside `initialize-place!`, but I decided to set the build stamp there (instead of returning the config table) so that all places would have the build stamp value. It's probably quite unlikely that someone will call `banner` from an additional place, but if they do, it seems best for it to have the same value in all places.

mflatt: Hm... All places share the same instance of Rumble (and any other layer implemented by a ".sls"), so initializing the banner stamp in `initialize-place!` does not sound right.

But I see that some other things in `initialize-place!` also set shared items: `set-host-collects-dir!`, `set-host-config-dir!`, `set-collects-dir!`, and `set-config-dir!`. In fact, probably that was the source of the trouble calling `read-installation-configuration-table` before `initialize-place!`.

Maybe the right thing is to merge your change, because resetting the variables will be harmless, but then I can look at cleaning up these initialization steps.

jryans: Ah okay! I indeed assumed that initialisation of shared items was needed for each place based on those other `set...!` calls. 😅 

Merging this as-is (matching the current redundant initialisation of place variables) and then separately cleaning up initialisation seems like a good strategy to me. 🙂

mflatt:
Thanks!

rfindler:
Hi -- I'm wondering what the consensus was for changing the DrRacket version string. Should I simplify it to match what this change does in src/cs/rumble/version.ss?  Ie:

```
(define (banner)
  (if build-stamp
      (string-append-immutable "Welcome to Racket v" (version) "-" build-stamp " [cs].\n")
      (string-append-immutable "Welcome to Racket v" (version) " [cs].\n")))
```

This is [the current code](https://github.com/racket/drracket/tree/master/drracket/repo-time-stamp). I don't recall the history but I think that maybe @elibarzilay wrote the code?

jryans:
> Should I simplify it to match what this change does in src/cs/rumble/version.ss?

Yes, that seems reasonable to me, as that would give the same presentation of the build stamp for both Racket and DrRacket. If someone does want to recreate the other functions of that code (Git commit detection etc.), they could rebuild such things in the build system, and then it should magically flow both Racket and DrRacket via the build stamp value.

mflatt:
That makes sense to me, too.

LiberalArtist:
I concur.

> This is [the current code](https://github.com/racket/drracket/tree/master/drracket/repo-time-stamp).

I think a minimal change might be to:

  - Delete `doc.txt`, which has been wrong since  since https://github.com/racket/racket/commit/2e067ab28c2b331713557feb880d8adf026a1aec (a.k.a. https://github.com/racket/drracket/commit/1451bb889cc673051266bd1eeb9d2ac11efb7673) in 2013. 
  - Change `stamp.rkt` to just:
    ```racket
    (define stamp
      (let ([stamp (get-build-stamp)])
        (and (not (equal? "" stamp)) stamp)))
    ````
  - In `time-stamp.rkt`, change the call to `version:add-spec` to supply `'-` instead of `'--`.

I expect there's further simplification that could be done, but I think that would be enough to match the change in this PR.

> Yes, that seems reasonable to me, as that would give the same presentation of the build stamp for both Racket and DrRacket. If someone does want to recreate the other functions of that code (Git commit detection etc.), they could rebuild such things in the build system, and then it should magically flow to both Racket and DrRacket via the build stamp value.

I'd be interested in formatting  `"20231203-4f7eb34066"` as either `"2023-12-03-4f7eb34066"` or `"2023-12-03(4f7eb34066)"`. I'll try to look into it soon unless someone else does first.

rfindler:
Thanks all! And thanks for the explicit instructions @LiberalArtist .

rfindler:
I think it would be nice if a git build had a banner that looked different in some way than a regular release (more than just the version number of having extra decimals in it), but I say this without the energy to try to figure out how to do it myself! So if there's no other interest, the status quo's also fine with me 😄 .

LiberalArtist:
I've drafted changes to setting the stamp in https://github.com/racket/racket/pull/4871.

PR #4854, opened 2023-12-07
Add `get-failure-result` to racket/function

lexi-lambda:
This adds a very simple function named `get-failure-result` to `racket/function`. It is essentially just

```racket
(define (get-failure-result v)
  (if (procedure? v) (v) v))
```

except that the version in this PR also includes an explicit arity check.

Since this function is so simple, the convenience it affords is obviously extremely minor. Still, I’ve written it so many times that it feels worth putting somewhere, especially since we already have `failure-result/c`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/procedures.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/procedures.scrbl
@@ -904,5 +904,20 @@ arguments that procedures with arity @racket[b] accept.
 
 }
 
+@defproc[(get-failure-result [v failure-result/c]) any]{
+
+If @racket[v] is a procedure, it is applied in @tech{tail position} to zero
+arguments to produce a result. Otherwise, @racket[v] is returned.
+
+@racket[get-failure-result] can be used to extract the value of failure result
+arguments passed to procedures such as @racket[hash-ref].
+
+@mz-examples[#:eval fun-eval
+(eval:check (get-failure-result 1) 1)
+(eval:check (get-failure-result (lambda () 2)) 2)
+(get-failure-result (lambda () (values 3 4)))]
+
+@history[#:added "8.11.1.4"]}
+
 
 @close-eval[fun-eval]
--- a/racket/collects/racket/function.rkt
+++ b/racket/collects/racket/function.rkt
@@ -9,6 +9,7 @@
          curry   curryr
          conjoin disjoin
          negate
+         get-failure-result
          (all-from-out racket/private/arity))
 
 (define (identity x) x)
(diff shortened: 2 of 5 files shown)


racket/collects/racket/function.rkt
```diff
@@ -49,6 +50,15 @@
       [(2) (lambda (x y) (not (f x y)))]
       [else (compose1 not f)]))) ; keyworded or more args => just compose
 
+(define (get-failure-result v)
+  (if (procedure? v)
+      (if (procedure-arity-includes? v 0)
+          (v)
+          (raise-argument-error 'get-failure-result
+                                "a procedure that accepts 0 non-keyword arguments"
```
usaoc, line 58: This should use `(procedure-arity-includes/c 0)` as the contract string, I think.

mflatt: The intent here may have been to match the exception raised by `hash-ref`. It needs to different, though, to do that. The `hash-ref` function conceptually always applies a procedure failure result to 0 arguments, but it improves the `exn:fail:contract:arity` error message.

pkgs/racket-doc/scribblings/reference/procedures.scrbl
```diff
@@ -904,5 +904,20 @@ arguments that procedures with arity @racket[b] accept.
 
 }
 
+@defproc[(get-failure-result [v failure-result/c]) any]{
+
+If @racket[v] is a procedure, it is applied in @tech{tail position} to zero
+arguments to produce a result. Otherwise, @racket[v] is returned.
+
+@racket[get-failure-result] can be used to extract the value of failure result
```
mflatt, line 912: Nit: should be "The `get-failure-result` function" to avoid starting a sentence with a lowercase identifier (per the style guide).

LiberalArtist:
> the version in this PR also includes an explicit arity check

I have also written this function multiple times, but I think I have rarely if ever implemented an explicit arity check. That seems quite useful for cases when the argument isn't already protected by `failure-result/c`. It might be worth adding an optional `#:who` argument for even better error reporting in such cases.

racket/collects/racket/function.rkt
lexi-lambda: The error message is based on the one produced by `failure-result/c`, which is really just the error message produced by `(-> any)`. I’m happy to change it, if desired, but consistency with `failure-result/c` seemed logical.

pkgs/racket-doc/scribblings/reference/procedures.scrbl
lexi-lambda: I’ve reworded as suggested.

PR #4856, opened 2023-12-08
Add `equal-always?` variant of `case`

samth:
Currently undocumented; requires `else` clause, does not specially handle trivial situations. Intended only for use in Rhombus to fix:

racket/rhombus-prototype#417.

Could be changed to be public with a bit more code duplication and/or abstraction.

### DIFF
--- a/pkgs/base/info.rkt
+++ b/pkgs/base/info.rkt
@@ -14,7 +14,7 @@
 
 ;; In the Racket source repo, this version should change exactly when
 ;; "racket_version.h" changes:
-(define version "8.11.1.7")
+(define version "8.11.1.8")
 
 (define deps `("racket-lib"
                ["racket" #:version ,version]))
--- a/pkgs/racket-doc/scribblings/reference/syntax.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax.scrbl
@@ -14,6 +14,7 @@
                      racket/provide
                      racket/package
                      racket/splicing
+                     racket/case
                      racket/runtime-path
                      racket/lazy-require
                      (only-in compiler/cm-accomplice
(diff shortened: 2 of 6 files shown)


samth:
This code broke the `#f` behavior in this test:

```
(let ()
  (define (f x)
    (case x
      [(low) 0]
      [(one) 1]
      [(middle) 2]
      [(upper #t) 3]
      [(high big up-there more) 4]
      [(extreme massive huge #f gigantic) 5]))
  (test 0 f 'low)
  (test 1 f 'one)
  (test 2 f 'middle)
  (test 3 f 'upper)
  (test 3 f #t)
  (test 4 f 'high)
  (test 4 f 'big)
  (test 4 f 'up-there)
  (test 4 f 'more)
  (test 5 f 'extreme)
  (test 5 f 'massive)
  (test 5 f 'huge)
  (test 5 f #f)
  (test 5 f 'gigantic)
  (test (void) f 'gigante)
  (test (void) f 0))
```

usaoc:
Not related to the change per se: Can you do something to the link so that this PR won’t close the linked issue?

sorawee:
No longer linked.

samth:
This works now, so it's worth discussing the best choice overall here. Options:

1. Add tests but no docs, leave this as-is and use in Rhombus.
2. Make this an officially-supported feature of `case` in `racket/base`.
3. Write a wrapper macro that disallows the equality specifier, provide that as `case` in `racket/base`, provide the version I implemented from `racket/private/case` for Rhombus or from a `racket/case` module that is not included in `racket`. 

I think @mflatt originally suggested (3) but I'm open to any of those or some other option.

sorawee:
I think `racket/private/case` would be weird and go against the spirit of language-oriented programming: Rhombus shouldn't need a "specialized support" from Racket (that is not available for other langs).

mflatt:
If we go with 3, I suggest a slightly different implementation:

* Change `(define-syntax (case stx) ....)` to `(define-for-syntax (make-case type)  (lambda (stx) ....))` and pass `type` along to the helper macros
* Add `(define-syntax case (make-case #'#:equal?))`
* Provide `make-case` in a submodule for use by a module that defines and exports `(define-syntax case/always (make-case #'#:equal-always?))`, or something like that.

The intent here is to shrink the expansion path and pattern duplication a little.

AlexKnauth:
Would it be helpful for the various scheme-compatibility languages, if a version of this supported `eqv?` as well as `equal-always?`

samth:
@mflatt That was my initial implementation plan, but the way the implementation is structured it seemed like I would have to use that approach only for `case` and not eg `case/sequential`, so I went with the more uniform implementation. Is there a way you see to define the collection of forms together in a parameterized way (without define-syntax-set or something like it)?

mflatt:
@samth Yes, just `case` and not `case/sequential`, etc., is what I had in mind. My thought was that `case` and similar are entry points that need more syntax checking, and dealing with an optional keyword is a lot more work for error cases. In fact, looking again, I think the "Error cases" case of the `case` might not be right when `#:equal?` is provided but a clause has a bad shape.

@AlexKnauth I agree that `case/eqv` could be handy alongside to `case/always` (or `case/equal-always`, I guess, to be consistent).

samth:
This is now updated along the lines suggested by @mflatt (and supports all of the four equality operators in Racket).

samth:
It didn't seem obviously worth the additional identifiers lying around/documentation/etc, but I'm open to it.

mflatt:
I agree with @sorawee that Rhombus should use public Racket things, so that's the motivation for something outside of `racket/private`. I'm not sure that `racket/case` is the best idea, but it seems more straightforward than making `make-case` public.

mflatt:
I added a commit to propose `racket/case` that exports `case/equal` (equivalent to `case`), `case/equal-always`, `case/eqv`, and `case/eq`.

samth:
This looks good to me now. Any other thoughts?

Issue #4836, opened 2023-11-23
Weirdly slow subprocess-wait?


tonyg:
(Copied across from [discourse](https://racket.discourse.group/t/weirdly-slow-subprocess-wait/2512), just in case the relevant audience isn't there, but here...)

tonyg | 2023-11-17 14:44:27 UTC

Hi all,

Some time in the past, oh, year?, some inscrutable subset of my various machines' Racket installations have consistently started taking between 250ms and 300ms to execute a subprocess, even one as simple as `/usr/bin/true`, to completion.

On some configurations, running `racket -e '(time (system* "/usr/bin/true"))'` reports a few milliseconds realtime, as expected. On others, it is much slower. I can't figure out what the differences might be! Maybe someone here has some ideas about how to find out what is happening?

Try various combinations of these commands:

```
racket --version
racket -e '(time (system* "/usr/bin/true"))'
docker run -it --rm racket/racket racket --version
docker run -it --rm racket/racket racket -e '(time (system* "/usr/bin/true"))'
```

Here are results from a few of my available configurations, all running Linux 6.5.0 on Debian trixie/sid (except the docker configurations, which I think are on Debian 12):

| Machine    | Racket                                                          | Fast or slow? |
|------------|-----------------------------------------------------------------|---------------|
| float      | docker, v8.11 cs                                                | slow          |
| float      | prebuilt from download.racket-lang.org, v8.10 cs                | fast          |
| leap       | docker, v8.11 cs                                                | slow          |
| leap       | from git 38a7d658439d0817d03da797122cd2ff3329df2e, v8.8.0.6 cs  | fast          |
| leap       | from debian package, v8.10 cs                                   | fast          |
| a fresh VM | from debian package, v8.10 cs                                   | fast          |
| a fresh VM | docker, v8.11 cs                                                | slowish       |
| zip        | prebuilt from download.racket-lang.org, v8.11 cs                | slow          |
| zip        | from git b38d895640e45611142f59583e747a4f33a710de, v8.11.0.3 cs | slow          |
| zip        | from debian package, v8.10 cs                                   | slow          |

where slow is >250ms, slowish is weirdly about 150ms, and fast is less, usually much less, than 50ms.

All the time is during the `subprocess-wait` call in `system*`'s implementation. Running strace shows the time appearing on a line `read(4, "!", 10) = 1`, which is part of the self-pipe trick for handling (presumably) SIGCHLD.

Any ideas?

mflatt:
[...]

PR #4860, opened 2023-12-10
set FD_CLOEXEC on all file descriptors and change `subprocess` handling

mflatt:
Intended to address #4836, this is a relatively small change to the implementation, and I think it will work right for nearly all uses of Racket. Still, changing the rules here might create a file-descriptor leak in a program that uses both foreign libraries or embeds Racket and that also uses `subprocess`. Setting `current-subprocess-keep-file-descriptors` to `null` (which is an existing part of the Racket API) should restore the old behavior in that case.

The root of the problem is that a `fork`+`exec` on Unix keeps all file descriptors from the original process open in the new process. Racket has long used a traditional correction of closing all file descriptors in a forked process, specifically by trying `close` on every file descriptor up to the maximum value. That worked fine when the file-descriptor limit was something like 256 or 1024. These days, it can be 1M, and 1M erroring `close` system calls can take a while. Many OSes offer an enumeration of open file descriptors through a "/proc/self/fd" or /dev/fd" device that acts like a directory, but accessing that information is not completely reliable (due to the possibility of `chroot`, for example), and it's not so easy in the interstital phase between `fork` and `exec` that permits only async-signal-safe functions.

A modern alternative is to set the FD_CLOEXEC flag on descriptors so that they are automatically closed on the `exec` step, and leave FD_CLOEXEC off on the file descriptors that you intend to be comunicated to the subprocess. This strategy works as long as everyone plays along, because FD_CLOEXEC is not the default. Also, if you have `fork` calls that might happen concurrently with file-descriptor creation, the two-step process of creating a file descriptor and then setting its FD_CLOEXEC flag is another problem. Linux helps to avoid the race by having a variant of each descriptor-creating system call with an extra flag to set FD_CLOEXEC, but that's not portable.

The change here adapts Racket at the rktio level to set FD_CLOEXEC on all created file descriptors, but mostly through portable APIs (i.e., not Linux-specific) plus an extra lock to prevent a `subprocess`-based `fork` concurrent to a file descriptor's creation. The extra lock works because rktio controls the call to `fork`. Foreign libraries are expected to play along as well as they can, and libraries like glib, sqlite3, and Gtk+ do seem to work that way (i.e., they set FD_CLOEXEC, and they tend to use the atomic interface where available).

Meanwhile, the old behavior for `subprocess` is still available by setting the `current-subprocess-keep-file-descriptors` parameter to `null` --- an API that was already in place, because Racket's behavior on Windows was already closer to the FD_CLOEXEC world. Also, if FD_CLOEXEC is not available from the operating system (as determined at compile time), everything works as before.

There's no Racket-level or even rktio-level way here to create file descriptors without FD_CLOEXEC when FD_CLOEXEC is supported, except for pipes at the rktio level. Adding a way to do that would make sense, if it one day seems useful. Even better, `subprocess` could take a list of file descriptors to share, which seems more declarative; the `null` value of
`current-subprocess-keep-file-descriptors` was meant to support an extension in that direction.

The Windows implementation of `subprocess` was closer to right before, mainly because non-sharing of file handles is the default there. The way `subprocess` created inherited stdin, stdout, and stderr pipes created a race among calls to `subprocess` in parallel places, however, and that's fixed here.


mflatt:
Some pointers to related discussion (usually long ago) for other languages and libraries:

* https://github.com/rust-lang/rust/issues/12148, conclusion was same as this PR, but with more follow-up effort to use atomic operations on file-descriptor creation

* https://github.com/ocaml/ocaml/issues/5256, closing commit effectively exposes FD_CLOEXEC and asks OCaml programmers to use it

* https://mail.gnome.org/archives/gtk-devel-list/2015-March/msg00038.html, but I suspect that this wasn't the last word

* https://mail.haskell.org/pipermail/haskell-cafe/2015-July/120523.html

* https://github.com/python/cpython/blob/main/Modules/_posixsubprocess.c, especially `_close_open_fds_safe`; this is what it looks like to try and stick to the traditional solution

sorawee:
Line 154 of `pkgs/racket-doc/scribblings/reference/subprocess.scrbl` contains non-text bytes.

PR #4861, opened 2023-12-11
docs: add a simple note on the `for-syntax` export in `#lang racket`

usaoc:
##### Checklist
- [x] documentation

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/reference.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/reference.scrbl
@@ -57,7 +57,9 @@ languages.}
 the @racketmodname[racket] library and will typically load faster.
 
 The @racketmodname[racket] library combines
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/reference.scrbl
```diff
@@ -57,7 +57,9 @@ languages.}
 the @racketmodname[racket] library and will typically load faster.
 
 The @racketmodname[racket] library combines
-@racketmodname[racket/base]@racket-extra-libs[].}
+@racketmodname[racket/base]@racket-extra-libs[].
+In addition, it reexports @racket[for-syntax] everything from
```
sorawee, line 61: Nit: I would prefer "re-exports"

sorawee:
Thanks!

PR #4862, opened 2023-12-11
doc: `module->realm` produces the realm, not the exports, and…

rocketnia:
…`module->indirect-exports` produces specifically the indirect exports.

##### Checklist

- [x] documentation

### Description of change

The documentation for `module->realm` said it "produces the exports," which seems likely to be text incorrectly copied from `module->exports`. I changed it to "produces the realm". I used a terminology link for "realm" since the one other instance of "realm" on the page had such a link.

The same text was used for `module->indirect-exports`, so I assumed that one to be a similar typo and clarified it to "produces the indirect exports". I didn't go with a terminology link for this one. I think this may be the first instance of the term "indirect exports" in the documentation aside from in variable names, but I think it's likely people will infer the meaning of this term from context since a prominent link to `module-compiled-indirect-exports` is already the reference point this sentence is phrasing its description in terms of.


mflatt:
Thanks!

PR #4863, opened 2023-12-11
doc: The `'ptr_ptr->save` protocol has 2 arguments, not 3.

rocketnia:
##### Checklist

- [x] documentation

### Description of change

The `'ptr_ptr->save` protocol of `unsafe-add-collect-callbacks` was described in the documentation as corresponding to the signature `void* (*)(void*, void*, void*)`. I've corrected this to `void* (*)(void*, void*)`, which is evidently the signature in use by [relevant code in BC's thread.c](https://github.com/rocketnia/racket/blob/d4b4626ad1b4b82e468467a18508893c7476b256/racket/src/bc/src/thread.c#L9152C1-L9160C29) and [relevant code in CS's memory.ss](https://github.com/rocketnia/racket/blob/d4b4626ad1b4b82e468467a18508893c7476b256/racket/src/cs/rumble/memory.ss#L604C1-L605C71).


mflatt:
Thanks!

PR #4865, opened 2023-12-11
add `CFLAGS+=...` handling to `configure`

mflatt:
I've often wanted a way to add to `CFLAGS` without replacing whatever it would otherwise use or default to. There doesn't seem to be a common way to do this on the command line with `configure` or with more modern build systems. (Pointers welcome if I've overlooked something.) But `CFLAGS+=....` seems like a natural and easy-to-remember way of doing this,  andit worked out well for a use case in Chez Scheme. I think it makes sense to add here, even though it means hacking the output of `autoconf` a little more than we already do.

### DIFF
--- a/racket/src/README.txt
+++ b/racket/src/README.txt
@@ -183,8 +183,10 @@ Detailed instructions:
 
          env CC=cc [here]configure
 
-    To add an include path, be sure to use CPPFLAGS="-I..." instead of
-    CFLAGS="-I...". The CPPFLAGS variable controls C pre-processing,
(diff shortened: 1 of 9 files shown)


LiberalArtist:
> (Pointers welcome if I've overlooked something.)

The closest thing to a "common" way I've seen is an `EXTRA_CFLAGS` variable for the user's flags, with an empty default&mdash;but some projects seem to use `EXTRA_CFLAGS` in the opposite way, for never-to-be-overridden flags, so `+=` seems like a good solution to me.

PR #4866, opened 2023-12-12
schemify: fix some issues in recent “unwrapping” change

usaoc:
This is a follow-up to 5e39a36.  Fix two oversights:

- Some bug in conditional causes three-valued and five-valued (specifically!) cases to be skipped;

- Unwrapping of `values` without checking for single-valuedness can lead to unsound optimization.


mflatt:
Thanks!

PR #4871, opened 2023-12-17
adjust build stamp for development and snapshots

LiberalArtist:
This is a first cut at adding enhancements to build stamps like those previously done by DrRacket, as discussed in https://github.com/racket/racket/pull/4847. It seems to work, but I have not tested with `distro-build` yet, and I'd also like people to have a chance to weigh in on the content of the build stamps. 

The change for snapshots is fairly simple: instead of e.g. `20231203-4f7eb34066`, these are now formatted as `2023-12-03-4f7eb34066`. I also considered `2023-12-03(4f7eb34066)`, which would be closer to what DrRacket used to do, but the stamps end up in urls like https://users.cs.utah.edu/plt/snapshots/20231215-396fa983c4/ and https://plt.cs.northwestern.edu/snapshots/20231214-396fa983c4/, so it seemed better to stick with characters that should never need to be escaped.

For development builds, the stamp is set only under the same circumstances as when `installation-name` is set to `development`. I noticed that `make base` is **not** one of those circumstances. This behavior is documented in the sense that:

https://github.com/racket/racket/blob/5b3f530e690f2268d607477760a5dbdcdb6b82e8/pkgs/racket-build-guide/build.scrbl#L183-L191

only specifies the behavior for `make` and `make in-place`, but I don't think there's anything documenting that these things don't happen with `make base`, and it's a difference from `make in-place PKGS=`. It surprised me, though it may make sense.

I did arrange to update an old automatically-generated the stamp even when previously the `config.rktd` file would have been unchanged, e.g. with `make as-is`.

For the content of the stamp, in development mode:

 1. If `git` is unavailable or the source is not a Git working tree (e.g. if it was created with `git archive`), the stamp is `2023-12-16-development`, using the build date.
 2. With Git, the basic format is something like `2023-12-16-example-119156e17d`, which consists of:
      - The date of the current commit, or the build date if the commit date somehow can't be found;
      - A tag or branch name, if available; and
      - The abbreviated commit hash.
 3. If there are uncommitted changes to tracked files, the stamp is further extended with `changed` and the build date, so e.g. `2023-12-16-example-359ec98455-changed-2024-01-01`.

I'm concerned that the last case is a bit too long. On the other hand, I find all of those pieces of information useful, and I find it more self-explanatory than the cryptic letter between `/` and `)` previously used by DrRacket.

Currently `make as-is` fails if the `config.rktd` file does not exist, which doesn't seem ideal.

I also don't love the current overloading of the `pkgs-config.rkt` script in general.

### DIFF
--- a/main.zuo
+++ b/main.zuo
@@ -355,6 +355,10 @@
   (define (as-is token vm)
     (check-mode "in-place")
     (base token vm)
+    (racket (find-racket vm)
+            (list "-U" "-G" (path-only build-config.rktd)
+                  (at-source "racket/src/pkgs-config.rkt")
+                  "--maybe-update-stamp"))
     (setup vm))
 
   (define (in-place token vm [also? #f] [setup-extra-args ""])
@@ -392,6 +396,7 @@
     (racket vars
             (list "-U" "-G" (path-only build-config.rktd)
                   (at-source "racket/src/pkgs-config.rkt")
+                  "--pkgs-catalog"
                   default-src-catalog
                   (lookup 'SRC_CATALOG)))
     (racket vars
@@ -456,9 +461,8 @@
   (define (server-from-base token [options-in (hash)])
     (check-mode "server")
     (make-build/site.rkt)
-    (update-stamp.txt)
-
     (define options (add-server-compile-machine options-in))
+    (update-stamp.txt options)
 
     ;; Create a copy of `SRC_CATALOG', so that we snapshot checksums, and
     ;; start building from it. The packages are installed in user scope,
@@ -524,26 +528,25 @@
                            "(machine)")
                        build/site.rkt)))
   
-  (define (update-stamp.txt)
+  (define (update-stamp.txt options)
     (define given-stamp (lookup 'BUILD_STAMP))
     (define stamp
       (cond
         [(not (equal? given-stamp ""))
          given-stamp]
         [else
-         (define (shell-result . cmds)
-           (define p (shell cmds (hash 'stdout 'pipe)))
-           (define r (fd-read (hash-ref p 'stdout) eof))
-           (process-wait (hash-ref p 'process))
-           (unless (= 0 (process-status (hash-ref p 'process)))
-             (error "failed" cmds))
-           (car (string-split r "\n")))
-         (~a
-          (shell-result "date" (string->shell "+%Y%m%d"))
-          (if (directory-exists? (at-source ".git"))
-              (~a "-"
-                  (shell-result "git log -1 --pretty=format:%h"))
-              ""))]))
+         (define cmds
+           (list (find-built-racket options)
+                 (at-source "racket/src/pkgs-config.rkt")
+                 "--display-auto-stamp"))
+         (define p
+           (apply racket/process (append cmds (list (hash 'stdout 'pipe)))))
+         (define r
+           (fd-read (hash-ref p 'stdout) eof))
+         (process-wait (hash-ref p 'process))
+         (unless (= 0 (process-status (hash-ref p 'process)))
+           (error "failed" cmds))
+         (car (string-split r "\n"))]))
     (display-to-file (~a stamp "\n") (at-source "build/stamp.txt") :truncate))
 
   (define (server-cache-config options)
--- a/racket/src/pkgs-config.rkt
+++ b/racket/src/pkgs-config.rkt
@@ -1,13 +1,42 @@
 #lang racket/base
 (require racket/cmdline
          racket/format
-         racket/path
+         racket/pretty
+         racket/port
+         racket/string
+         racket/date
+         racket/system
          racket/file)
 
-;; Adjust the configuration to consult a catalog that is
-;; expected to map some packages to directory links.
-
-;; Used by the top-level Makefile in the main Racket repository.
+(module+ main
+  (command-line
+   #:usage-help
+   "Used by the top-level Makefile in the main Racket repository."
+   "If no <option> is provided, does nothing."
+   #:once-any
+   [("--pkgs-catalog")
+    default-src-catalog src-catalog
+    ("Configure a development build to consult first a local catalog,"
+     "which is expected to map packages in this repository to directory links,"
+     "next <src-catalog>, and then <default-src-catalog> (if different) before"
+     "falling back to the built-in default catalog."
+     "If a config.rktd file is created, also set its installation-name,"
+     "build-stamp, default-scope, interactive-file, and"
+     "gui-interactive-file to values suitable for development from Git."
+     "If a config.rktd file already exists, check its catalog configuration"
+     "and potentially update its build stamp as with `--maybe-update-stamp`.")
+    (pkgs-catalog default-src-catalog src-catalog)]
+   [("--maybe-update-stamp")
+    ("Update the build stamp if and only if a config.rktd file already"
+     "exists and indicates that its build stamp was previously generated"
+     "by this script.")
+    (maybe-update-stamp)]
+   [("--display-auto-stamp")
+    ("Display to standard output an automatic build stamp (in the same style"
+     "as above) followed by a newline. Used by `distro-build`.")
+    (displayln (auto-stamp))]
+   #:args ()
+   (void)))
 
 (define config-dir-path (build-path "racket" "etc"))
 (define config-file-path (build-path config-dir-path "config.rktd"))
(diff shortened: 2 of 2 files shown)


mflatt:
This all looks fine to me.

Seeing the read-again-then-write addition to `pkgs-config.rkt` made me think a little more about atomicity. Maybe that script should use `call-with-atomic-output-file` (in the old place and the new one), just in case someone interrupts their build with Ctl-C.

FWIW, seeing the build stamp when I start Racket is noisier than I prefer. But I can appreciate why others would want to see it, and it's obvious how to turn it off locally.

jryans:
> 2. With Git, the basic format is something like `2023-12-16-build-stamp-119156e17d`, which consists of:
>    
>    * The date of the current commit, or the build date if the commit date somehow can't be found;
>    * A tag or branch name, if available; and
>    * The abbreviated commit hash.

To clarify, I assume the literal text “build-stamp” does not appear in the end result and is replaced by the branch name?

> 3. If there are uncommitted changes to tracked files, the stamp is further extended with `changed` and the build date, so e.g. `2023-12-16-build-stamp-359ec98455-changed-2024-01-01`.

This last part does seem a bit excessive to me. It would be a bit unfortunate for the banner to become so long that it wraps to the next line for typical terminal widths.

LiberalArtist:
> Seeing the read-again-then-write addition to `pkgs-config.rkt` made me think a little more about atomicity. Maybe that script should use `call-with-atomic-output-file` (in the old place and the new one), just in case someone interrupts their build with Ctl-C.

That does seem better.



> > 2\. With Git, the basic format is something like `2023-12-16-build-stamp-119156e17d`, which consists of:
> >     
> >    * The date of the current commit, or the build date if the commit date somehow can't be found;
> >    * A tag or branch name, if available; and
> >    * The abbreviated commit hash.
> 
> To clarify, I assume the literal text “build-stamp” does not appear in the end result and is replaced by the branch name?

Yes, that was a confusing artifact of the fact that my branch in this PR is named `build-stamp`. I've edited the post to use `example` as the branch name, instead.

> > 3\. If there are uncommitted changes to tracked files, the stamp is further extended with `changed` and the build date, so e.g. `2023-12-16-build-stamp-359ec98455-changed-2024-01-01`.
> 
> This last part does seem a bit excessive to me. It would be a bit unfortunate for the banner to become so long that it wraps to the next line for typical terminal widths.

I agree that shorter would be better. Of course, people can choose arbitrarily long branch names, but it would be good to avoid wrapping for typical cases.

Some ideas instead of `2023-12-16-example-359ec98455-changed-2024-01-01`:

- We could use only the build date when there are uncommitted changes, instead of also showing the commit date, which would get `2024-01-01-example-359ec98455-changed`.
- We could further abbreviate `-changed`, maybe as `2024-01-01-example-359ec98455*`. I also considered `~`, but that has a different meaning with `git rev-parse`.

Thoughts?

benknoble:
Perhaps consider `git describe` (with whatever flags you prefer) when using Git for a build stamp?

jryans:
> Some ideas instead of `2023-12-16-example-359ec98455-changed-2024-01-01`:
> 
>  * We could use only the build date when there are uncommitted changes, instead of also showing the commit date, which would get `2024-01-01-example-359ec98455-changed`.

Two dates does seem excessive, so some path to get down to one seems good. Perhaps just always use the build date? (The commit date can be obtained via the hash when needed.)

>  * We could further abbreviate `-changed`, maybe as `2024-01-01-example-359ec98455*`. I also considered `~`, but that has a different meaning with `git rev-parse`.

Yes, some short symbol for changed (or just ignoring this case) seems better to me.

After some reflection, a date and commit hash (where available) feels like the most I would need here (but that's just me, perhaps others have uses for more). For example, I can get the branch name separately from Git and other context if I really need it.

rfindler:
My $0.02: it is nice to have something more complex (and scary looking :)
than simply a version number when I build it myself and a commit sha seems
about right to help me notice that the one I have in directory X is
different than the one in directory Y. That said, adding a date doesn't
seem like too much information.

It does seem backwards that the version string is longer when I get a
snapshot build than when I build it myself (of course, this is what you're
changing here: great!).

Robby

LiberalArtist:
> Perhaps consider `git describe` (with whatever flags you prefer) when using Git for a build stamp?

This was the first thing I tried, but it turned out not to be very informative. In the default mode, you get e.g. `v5.0.1-28205-g1c465a948f`, always starting with literally `v5.0.1-`, because the `master` branch does not contain any more recent release tags (see https://racket.discourse.group/t/proposed-stable-branch-discipline/1453). With something like `git describe --all --always --long`, you get e.g. `heads/my-branch-0-g1c465a948f`: the counter in this case is not very useful, and post-processing to remove `heads/`, at least, would still be needed to get a result as concise as DrRacket before https://github.com/racket/drracket/commit/841d5b3. At that point it seemed simpler to just choose exactly what we want.

> > Some ideas instead of `2023-12-16-example-359ec98455-changed-2024-01-01`:
> > 
> > * We could use only the build date when there are uncommitted changes, instead of also showing the commit date, which would get `2024-01-01-example-359ec98455-changed`.
> 
> Two dates does seem excessive, so some path to get down to one seems good. Perhaps just always use the build date? (The commit date can be obtained via the hash when needed.)

I'm persuaded to just always use the commit date, especially because the support for [`SOURCE_DATE_EPOCH`](https://reproducible-builds.org/docs/source-date-epoch/) gives the user a way to override the date: for example, `SOURCE_DATE_EPOCH=$(git log -1 --pretty=%ct)` would use the commit date.

> > * We could further abbreviate `-changed`, maybe as `2024-01-01-example-359ec98455*`. I also considered `~`, but that has a different meaning with `git rev-parse`.
> 
> Yes, some short symbol for changed (or just ignoring this case) seems better to me.

I've found it useful in other contexts to distinguish "I built exactly this commit" from "I build some messy changes on top of this commit", but I think `*` is enough, unless someone has a better idea. (I'm inclined to leave the `-broken` case as-is because it is a rare scenario that seems worth making noise about.)

> After some reflection, a date and commit hash (where available) feels like the most I would need here (but that's just me, perhaps others have uses for more). For example, I can get the branch name separately from Git and other context if I really need it.

> a commit sha seems about right to help me notice that the one I have in directory X is different than the one in directory Y. That said, adding a date doesn't seem like too much information.

I don't have a strong opinion about whether to include the branch name or not. For the sake of illustration, this would be a choice between:

 1. `2024-01-01-example-359ec98455*`; or
 2. `2024-01-01-359ec98455*`.

The biggest advantage I see to including the branch is the human-readable hint about why I built this Racket. The disadvantages I see are length and nondeterminism (i.e. many different branches may have the same commit).

According to https://github.com/racket/racket/commit/1fae7942d583352306e6acb302f98b4639e7eca8, @samth once requested that the branch name be included.

I guess, unless @samth or someone else affirmatively would like to have the branch name, I'll plan to change to the branch-less `2024-01-01-359ec98455*` style.

samth:
I am fine with leaving the branch name out.

LiberalArtist:
I've pushed a revised commit that uses the style we discussed:

> I'll plan to change to the branch-less `2024-01-01-359ec98455*` style.

In the process of revising the code, I noticed that we had basically recreated the format used for snapshot stamps by `distro-build`. I decided to go all the way and actually use the same code to generate the stamp in both cases. A few implications of that change:

- For a non-Git build, the stamp is just the date, e.g. `2024-01-01`, as was already the case with `distro-build`. The previous version of this PR used `2024-01-01-development`.
- Previously, `distro-build` would fail hard if a `.git` directory (not a regular file, so this didn't work with worktrees) was present but the `git` command failed or was unavailable. I have changed this behavior so that failure to get a stamp from Git never breaks the build.
- The code to detect uncommitted changes still runs when generating a stamp for a snapshot, though there should not be uncommitted changes in that case.

I've changed the script to always use `call-with-atomic-output-file`, and I've also organized the script into functions to avoid some of the overloading I disliked in the last version now.

I haven't worked out the best way to test this with `distro-build`, but otherwise I hope it's ready to go.

mflatt:
You can check the effect on `distro-build` with `make installers PKGS=""` in a fresh checkout (where the `PKGS=""` part just makes it finish a lot faster). It seems to have worked as intended when I try on macOS.

LiberalArtist:
Thanks! `make installers PKGS=""` also worked for me on Debian.

I pushed a typo fix to a help string just now.

rfindler:
It looks like [this file](https://github.com/racket/distro-build/blob/master/distro-build-server/private/find-desired-snapshots.rkt) needs some adjustments (below is the tail of the snapshot failure at northwestern from this morning). 

I guess the right thing to change the regexp to allow hyphens in the middle of the dates? I can make that change, but just wanted to double check that that makes sense with the folks here.


```
========================================= snapshot log:
  installers/page-not-found.html
  installers/.htaccess
  pdf-doc/robots.txt
  pdf-doc/favicon.ico
  pdf-doc/page-not-found.html
  pdf-doc/.htaccess
  download-page/robots.txt
  download-page/favicon.ico
  download-page/page-not-found.html
  download-page/.htaccess
Rendering done.
racket/bin/racket -MCR /Users/robby/git/snapshot/racket/build/zo: -G build/user/config -X racket/collects -A build/user -l- distro-build/manage-snapshots ../plt-admin/service-conf/nwu-snapshot-pkg-build/penghu-site.rkt default racket
Removing old "current" link
manage-snapshots.rkt: couldn't parse snapshot date: "2024-01-02-b3ee33b256"
  context...:
   body of "/Users/robby/git/snapshot/racket/build/user/[8.11.1.10/pkgs/distro-build-server/manage-snapshots.rkt](http://8.11.1.10/pkgs/distro-build-server/manage-snapshots.rkt)"
failed
 in build-one
 in loop
make: *** [snapshot-site] Error 1
```

rfindler:
Actually, that seems unlikely to be wrong and a new build is about to start so I've pushed the change. Happy to hear folks thoughts about it while the next build runs, tho!

LiberalArtist:
I'm not sure how general that file is supposed to be, but maybe we should consider ignoring the trailing `-` after the day, too. For example, a snapshot built without `git` available would a stamp like `"2024-01-03"` (i.e. just the date). (Though `git` may be required for other reasons.)

rfindler:
Oh, okay! Thank you.

PR #4876, opened 2023-12-18
doc: improve explanation of unsafety for atomic mode

### LINKED ISSUE #4858: Atomic mode is still allowing code to yield and exit atomic mode silently, including if it just takes a long time
8.11.1 [cs]

Also tested 8.7 [cs] where bug was present.

**What program did you run?**

```
#lang racket
(require ffi/unsafe/atomic)

(define (fibo n)
  (if (> n 2)
      (+ (fibo (sub1 n))
         (fibo (- n 2)))
      1))

(define (make-bad-worker)
  (thread
   (thunk
    (call-as-atomic
     (thunk
      (fibo 3) ;; sleep is not necessary if you make this big like 42
      (sleep 1))))))
(thread-wait (make-bad-worker))
```

It should raise some error on the worker thread and/or crash Racket.
Instead, it seems to just turn off atomic mode when sleep is called.

On version 8.1, it gives the correct error (I'm not sure exactly when this problem was introduced, but it seems old and even followed the current behavior in 8.7):

```
internal error: attempt to deschedule the current thread in atomic mode
  context...:
   C:\Program Files\Racket\collects\ffi\unsafe\atomic.rkt:73:13
```

```
internal error: not in atomic mode to end
  context...:
   body of "C:\gdrive\code\racket\fibo-time-ex.rkt"
```

Yielding also happens if you make the call to fibo big, like (fibo 42). That seems extremely bad since execution time is out of control the programmer.

There is also a way to sort of do this sort of thing and not totally nuke the scheduler.

```
#lang racket
(require ffi/unsafe/atomic)
(require ffi/unsafe/try-atomic)

(define (fibo n)
  (if (> n 2)
      (+ (fibo (sub1 n))
         (fibo (- n 2)))
      1))

(define (make-bad-worker)
  (thread
   (thunk
    (call-as-nonatomic-retry-point
     (thunk
      (try-atomic
       (thunk
        (fibo 3)
        (sleep 1))
       #f))))))
```

mflatt:
Closes #4858

### DIFF
--- a/pkgs/racket-doc/scribblings/foreign/atomic.scrbl
+++ b/pkgs/racket-doc/scribblings/foreign/atomic.scrbl
@@ -1,5 +1,7 @@
 #lang scribble/doc
-@(require "utils.rkt" (for-label ffi/unsafe/atomic))
+@(require "utils.rkt"
+          (for-label ffi/unsafe/atomic
+                     ffi/unsafe/try-atomic))
 
 @title{Atomic Execution}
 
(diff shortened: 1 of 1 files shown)


LiberalArtist:
I don't recall many specific questions offhand, but I know I have sometimes felt unclear about the distinctions (if any) between atomic mode and some other sets of restrictions (most of which include atomic mode) related to `ffi/unsafe`:
  - `#:async-apply` arguments to `_cprocedure` or `_fun`
  - callbacks created with `#:atomic #t`&mdash;are these additional restrictions, or just further examples of the restrictions of atomic mode?
    > In addition, the Racket code must not perform any potentially blocking operation (such as I/O), it must not raise an uncaught exception unless called through a [callout](https://docs.racket-lang.org/foreign/foreign_procedures.html#%28tech._callout%29) that supports exception (with `#:callback-exns? #t`), it must not perform any escaping continuation jumps, and (at least for the [BC](https://docs.racket-lang.org/guide/performance.html#%28tech._bc%29) implementation) its non-tail recursion must be minimal to avoid C-level stack overflow; otherwise, the process may crash or misbehave.
  - Functions called in an unspecified Racket thread, e.g. the callback for `register-custodian-shutdown`
  - Finalizers for `register-finalizer`
  - Functions used with `call-in-os-thread`

mflatt:
I agree that more is needed here. For example, there's some duplication in the description of `#:atomic`, but there are also additional restrictions about escaping and C-level overflow in that context.

LiberalArtist:
There's now also *uninterruptible mode*, which comes with a ban on &ldquo;any action that blocks the future or requires synchronization with coroutine threads&rdquo;.

I think (but am not sure) that there exist &ldquo;action[s] that block[ a] future&rdquo; which *are* allowed in atomic mode. Additionally, `start-atomic` &ldquo;cannot be used in uninterruptible mode in a future or parallel thread&rdquo;, so I think neither atomic mode nor uninterruptible mode is a subset of the other.

mflatt:
Merged with additional revisions as 94ab89c967ef5bfda9ffcd4d613938cb672fb3b8.

PR #4878, opened 2023-12-22
Use terminal control to print over existing lines when compiling zos and building docs in raco setup

rfindler:
Also, add in some logic to determine when one of the parallel workers is not doing anything anymore to include "idle" lines.

I'm not especially confident in the code that decides that something is idle. I tried watching top and looking at the printfs but this is fraught because sometimes there are some internal locks (eg, two separate workers try to compile two files x and y, but both x and y depend on z and one of them gets to z (and all its dependencies) first) that mean that a worker that has a task isn't actually *doing* anything. At least, that's how I think it works. And sometimes I saw what looked like N workers but a CPU load > N*100% (even > (N+1)*100%) but it didn't seem to last long when I saw it so maybe that's just some averaging effect.

Also, in [this case](https://github.com/rfindler/racket/blob/terminal-for-raco-setup/racket/collects/setup/parallel-build.rkt#L191-L198) there wasn't a printf but it looks like it is doing some compiling. But maybe the `#f` being passed to `build-job` might mean that it isn't actually doing work, I'm not sure.

### DIFF
--- a/pkgs/racket-index/setup/scribble.rkt
+++ b/pkgs/racket-index/setup/scribble.rkt
@@ -676,7 +676,7 @@
                                                      i)))
                                             infos)
                                 info<?)])
-          (define (say-rendering i workerid idle?)
+          (define (say-rendering i workerid idle? %age)
             (setup-printf (string-append
                            (if workerid (format "~a" workerid) "")
                            (if idle?
@@ -686,6 +686,7 @@
                                 (if (info-rendered? i) "re-" "")
                                 "rendering")))
                           #:only-if-terminal? idle?
+                          #:%age %age
                           #:n workerid
                           "~a"
                           (if idle? "idle" (path->relative-string/setup (doc-src-file (info-doc i))))))
@@ -704,7 +705,7 @@
                  (set-info-done-time! info (current-inexact-milliseconds)))]))
           (if ((min worker-count (length need-rerun)) . < . 2)
               (for ([i (in-list need-rerun)])
-                (say-rendering i #f #f)
+                (say-rendering i #f #f #f)
                 (prep-info! i)
                 (update-info! i (build-again! latex-dest i with-record-error
                                               no-lock (if gc-after-each-sequential? gc-point void)
@@ -717,8 +718,8 @@
                  (list workerid (verbose) latex-dest lock-ch main-doc-exists?))
                (list-queue
                 need-rerun
-                (lambda (i workerid) 
-                  (say-rendering i workerid #f)
+                (lambda (i workerid remaining-work original-amount-of-work)
+                  (say-rendering i workerid #f (- 1 (/ remaining-work original-amount-of-work)))
                   (prep-info! i)
                   (s-exp->fasl (serialize (list (info-doc i)
                                                 ;; Other content of `info' can be re-read from
@@ -730,12 +731,12 @@
                                                 (info-deps->rel-doc-src-file i)
                                                 (info-need-in-write? i)))))
                 (lambda (i r outstr errstr workerid)
-                  (say-rendering i workerid #t)
+                  (say-rendering i workerid #t #f)
                   (printf "~a" outstr) 
                   (printf "~a" errstr)
                   (update-info! i (deserialize (fasl->s-exp r))))
                 (lambda (i errmsg outstr errstr workerid)
-                  (say-rendering i workerid #t)
+                  (say-rendering i workerid #t #f)
                   (parallel-do-error-handler with-record-error (info-doc i) errmsg outstr errstr)))
                (define-worker (build-again!-worker2 workerid verbosev latex-dest lock-ch
                                                     main-doc-exists?)
--- a/pkgs/racket-index/setup/scribble.rkt
+++ b/pkgs/racket-index/setup/scribble.rkt
@@ -383,14 +383,14 @@
                  (list-queue
                   (list-tail docs num-sequential)
                   (lambda (x workerid) (s-exp->fasl (serialize x)))
-                  (lambda (work r outstr errstr) 
+                  (lambda (work r outstr errstr workerid)
                     (printf "~a" outstr)
                     (printf "~a" errstr)
                     (deserialize (fasl->s-exp r)))
-                  (lambda (work errmsg outstr errstr) 
+                  (lambda (work errmsg outstr errstr workerid)
                     (parallel-do-error-handler with-record-error work errmsg outstr errstr))
                   (lambda (args)
-                    (apply setup-printf args)))
+                    (keyword-apply setup-printf (list-ref args 0) (list-ref args 1) (list-ref args 2))))
                  (define-worker (get-doc-info-worker workerid program-name verbosev only-dirs latex-dest 
                                                      avoid-main? auto-main? auto-user? main-doc-exists?
                                                      force-out-of-date? lock-ch)
@@ -399,8 +399,10 @@
                                                 force-out-of-date? lock
                                                 send/report) 
                             doc)
-                     (define (setup-printf . args)
-                       (send/report args))
+                     (define setup-printf
+                       (make-keyword-procedure
+                        (λ (kwds kwd-args . args)
+                          (send/report (list kwds kwd-args args)))))
                      (define (with-record-error cc go fail-k)
                        (with-handlers ([exn:fail?
                                         (lambda (exn)
@@ -674,12 +676,19 @@
                                                      i)))
                                             infos)
                                 info<?)])
-          (define (say-rendering i workerid)
+          (define (say-rendering i workerid idle?)
             (setup-printf (string-append
-                           (if workerid (format "~a " workerid) "")
-                           (if (info-rendered? i) "re-rendering" "rendering") )
+                           (if workerid (format "~a" workerid) "")
+                           (if idle?
+                               ""
+                               (string-append
+                                (if workerid " " "")
+                                (if (info-rendered? i) "re-" "")
+                                "rendering")))
+                          #:only-if-terminal? idle?
+                          #:n workerid
                           "~a"
-                          (path->relative-string/setup (doc-src-file (info-doc i)))))
+                          (if idle? "idle" (path->relative-string/setup (doc-src-file (info-doc i))))))
           (define (prep-info! i)
             (set-info-start-time! i (current-inexact-milliseconds)))
           (define (update-info! info response)
@@ -695,7 +704,7 @@
                  (set-info-done-time! info (current-inexact-milliseconds)))]))
           (if ((min worker-count (length need-rerun)) . < . 2)
               (for ([i (in-list need-rerun)])
-                (say-rendering i #f)
+                (say-rendering i #f #f)
                 (prep-info! i)
                 (update-info! i (build-again! latex-dest i with-record-error
                                               no-lock (if gc-after-each-sequential? gc-point void)
@@ -709,7 +718,7 @@
                (list-queue
                 need-rerun
                 (lambda (i workerid) 
-                  (say-rendering i workerid)
+                  (say-rendering i workerid #f)
                   (prep-info! i)
                   (s-exp->fasl (serialize (list (info-doc i)
                                                 ;; Other content of `info' can be re-read from
@@ -720,11 +729,13 @@
                                                 ;; in this list:
                                                 (info-deps->rel-doc-src-file i)
                                                 (info-need-in-write? i)))))
-                (lambda (i r outstr errstr) 
+                (lambda (i r outstr errstr workerid)
+                  (say-rendering i workerid #t)
                   (printf "~a" outstr) 
                   (printf "~a" errstr)
                   (update-info! i (deserialize (fasl->s-exp r))))
-                (lambda (i errmsg outstr errstr) 
+                (lambda (i errmsg outstr errstr workerid)
+                  (say-rendering i workerid #t)
                   (parallel-do-error-handler with-record-error (info-doc i) errmsg outstr errstr)))
                (define-worker (build-again!-worker2 workerid verbosev latex-dest lock-ch
                                                     main-doc-exists?)
@@ -1122,8 +1133,10 @@
                                  (or (memq 'depends-all (doc-flags doc))
                                      (memq 'depends-all-user (doc-flags doc))))))])
 
-    (when (or (and (not up-to-date?) (not only-fast?))
-              (verbose))
+    (define print-worker-status?
+      (or (and (not up-to-date?) (not only-fast?))
+          (verbose)))
+    (when print-worker-status?
       (when out-of-date
         (verbose/log "Need run (~a) ~a" out-of-date (doc-name doc)))
       (setup-printf
@@ -1135,6 +1148,7 @@
                        "checking"
                        "running")]
          [else "skipping"]))
+       #:n workerid
        "~a"
        (path->relative-string/setup (doc-src-file doc))))
 
@@ -1143,6 +1157,7 @@
         (when (file-exists? p)
           (delete-file p))))
 
+    (define result
     (if up-to-date?
         ;; Load previously calculated info:
         (render-time
@@ -1270,7 +1285,14 @@
                      (db-shutdown)
                      info))))
              (lambda () #f))
-            #f))))
+            #f)))
+    (when (and print-worker-status? workerid)
+      (setup-printf
+       (format "~a" workerid)
+       #:n workerid
+       #:only-if-terminal? #t
+       "idle"))
+    result))
 
 (define (check-shared-files dir root? main? done setup-printf)
   (define dest-dir (simplify-path (if root?
--- a/racket/collects/setup/parallel-build.rkt
+++ b/racket/collects/setup/parallel-build.rkt
@@ -152,6 +152,7 @@
         (define (say-making id x)
           (unless (null? x)
             (printer (current-output-port) 
+                     #:n id
                      (format "~a making"  id)
                      "~a" 
                      (cc-name (car (car x))))))
--- a/racket/collects/setup/parallel-build.rkt
+++ b/racket/collects/setup/parallel-build.rkt
@@ -33,7 +33,7 @@
 
 (define lock-manager% 
   (class object%
-    (init-field worker-count)
+    (init-field worker-count [blocked-notify void] [unblocked-notify void])
     (field (locks (make-hash)))
     (define depends (make-hash))
     (define currently-idle 0)
@@ -57,6 +57,7 @@
                [else
                 (idle! +1)
                 (hash-set! depends wrkr (cons w fn))
+                (blocked-notify wrkr fn)
                 (list w (append waitlst (list wrkr)))]))]
            [_
             (wrkr/send wrkr (list 'locked))
@@ -68,6 +69,7 @@
          (for ([x (second (hash-ref locks fn))])
             (idle! -1)
             (hash-remove! depends x)
+            (unblocked-notify x)
             (wrkr/send x (list 'compiled)))
           (hash-set! locks fn 'done)]))
     (define/private (check-cycles w seen fns)
@@ -97,8 +99,25 @@
   (class* object% (work-queue<%>) 
     (init-field cclst printer append-error options)
     (init worker-count)
-    (field (lock-mgr (new lock-manager% [worker-count worker-count])))
-    (field (hash (make-hash)))
+    (field (lock-mgr (new lock-manager%
+                          [worker-count worker-count]
+                          [blocked-notify
+                           (λ (wrkr fn)
+                             (define id (send wrkr get-id))
+                             (printer
+                              (current-output-port)
+                              #:n id
+                              #:only-if-terminal? #t
+                              (format "~a waiting for" id)
+                              "~a" fn))]
+                          [unblocked-notify
+                           (λ (wrkr)
+                             (say-making
+                              (send wrkr get-id)
+                              #:only-if-terminal? #t))])))
+
+    ;; assigned-ccs : (hash natural?[workerid] -o> (μX. (or/c '() (cons (list cc? (listof file?) X) X))))
+    (define assigned-ccs (make-hash))
     (inspect #f)
 
     (define/public (work-done work wrkr msg)
@@ -109,6 +128,7 @@
             (match result-type
               [(list 'ERROR long-msg short-msg)
                 (append-error cc "making" (cons long-msg short-msg) out err "error")
+                (say-idle id)
                 #t]
               [(list 'LOCK fn) (lm/lock lock-mgr fn wrkr) #f]
               [(list 'UNLOCK fn) (lm/unlock lock-mgr fn) #f]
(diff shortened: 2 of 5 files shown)


sorawee:
I'm excited to see this!

One issue: in GitHub Actions, many lines appear to be duplicated with this change:

Screenshot 2023-12-21 at 5 06 24 PM

This doesn't happen prior this change, if I'm not mistaken.

rfindler:
<a href="http://www.cs.northwestern.edu/~robby/tmp/raco-setup.html">ObVideo</a>

There is an error during the documentation building shortly after the 4 minute mark, in case you want to see that.

One situation I don't know how to deal with is when someone types something at the terminal; currently the code doesn't notice that and so there are various off-by problems in the printing since the cursor isn't where the code left it before. It is possible to block all input using `terminal-raw-mode` but that brings a lot of other problems and the current state seems better than the best I could do while using raw mode.

rfindler:
> One issue: in GitHub Actions, many lines appear to be duplicated with this change

Ah, thanks! @mflatt had suggested that [that case](https://github.com/rfindler/racket/blob/terminal-for-raco-setup/racket/collects/setup/parallel-build.rkt#L191-L194) was for the situation where a worker tries to compile a file that's already compiled; probably I should have just believed him (I'm not sure where I got evidence that it wasn't!). Anyway I've made that be an "idle" case and the extra lines seem to go away now.

sorawee:
The video demo looks really great! :)

My wishlist:

- When the exception is printed, could you explicitly show which job/source/collection/package causes the error before the exception? I can infer that it's probably `raco setup: 5 running: <pkgs>/redex-doc/redex/redex.scrbl` that causes the the error in the demo, since it's pretty straightforward to make the connection in this case, but other cases might not be as straightforward.
- Maybe we can do more "clean up"? For example:
   - We can totally erase all idle lines at the end after all jobs finished -- this is at timestamp 3:09 and 7:48.
   - After an exception occur, since we are going to print the jobs again anyway, we can erase the previous jobs -- this is at timestamp 4:05.

rfindler:
> The video demo looks really great! :)

Thank you! (Working on this is my chrismas present to myself -- not to say I wouldn't welcome commits from you, of course!)

> My wishlist:
> 
>     * When the exception is printed, could you explicitly show which job/source/collection/package causes the error before the exception? I can infer that it's probably `raco setup: 5 running: <pkgs>/redex-doc/redex/redex.scrbl` that causes the the error in the demo, since it's pretty straightforward to make the connection in this case, but other cases might not be as straightforward.

That could be done, but it would probably best be done in a way orthogonally from this change. Whatever is printing the exception out would just have to print out that information. I'm not sure if the information is near each other at the moment, tho. 

>     * Maybe we can do more "clean up"? For example:
>       
>       * We can totally erase all idle lines at the end after all jobs finished -- this is at timestamp 3:09 and 7:48.

That seems doable; there is logic in `setup-printf` that knows what the state of the screen is, so we'd probably just need to have another call into that from the place in the guts of setup that knows when each given phase is done.

>       * After an exception occur, since we are going to print the jobs again anyway, we can erase the previous jobs -- this is at timestamp 4:05.

This one seems hard. The problem is that the code that printed out the exception information is not connected to the code that is printing the lines. There is just [the `output-didnt-move?` boolean](https://github.com/rfindler/racket/blob/terminal-for-raco-setup/racket/collects/setup/private/setup-fprintf.rkt) that notices that something has been printed since last time it printed a line, so it rebuilds the state below whatever it was that was printed. I don't think there is a way to know what is on the screen (at least I don't know how to to do with `#%terminal` but maybe there is something) so this is just a kind of best effort situation.

More generally, there are tons of place where setup prints stuff out. When life is going well, that doesn't happen, but there about a gazillion reasons why something might get printed out and so instead of trying to track all those down (some of them are even calls to `log-error` and not printing IIRC) it seemed safer to just try to react to them in a hopefully-not-horrible way.

rfindler:
note to self: check to see if the [guard here](https://github.com/rfindler/racket/blob/terminal-for-raco-setup/racket/collects/setup/private/setup-fprintf.rkt#L58-L60) needs to explicitly check stderr as well.

rfindler:
Another note to self: make sure to run taco setup in the way that uses separate processes to see if that code path works.

samth:
This is great! One request while you're looking at this code -- I would love to have a log message when a task starts and when it finishes, with the number of the worker that did it.

rfindler:
> This is great! One request while you're looking at this code -- I would love to have a log message when a task starts and when it finishes, with the number of the worker that did it.

thanks! I'm not sure what you mean with the request tho. Can you elaborate a little bit?

shhyou:
This is fantastic!

I have some questions and some requests about the command-line user interface.

- Will there be an option to explicitly enable and disable this behavior? Like `--XYZ` and `--no-XYZ`
- How does this interact with `/usr/bin/screen` (or tmux) when switching between windows?
- By default, `-j 1` prints more information during `raco setup`. Has that changed?

Also, I think it'd be better to detect the output type to decide if this printing behavior is going to be enabled. One necessary condition I have in mind is that the output port has to be a file stream port[^1] and `file-stream-buffer-mode`[^2] needs to be `'none` or `'line`, to avoid affecting piped/redirected outputs, interfering with other kinds of output ports or affecting other tools.

```racket
●[Dec22 09:36:42] $ racket -e '(file-stream-buffer-mode (current-output-port))'
'line
●[Dec22 09:37:11] $ racket -e '(file-stream-buffer-mode (current-output-port))' | cat
'block
●[Dec22 09:37:14] $ racket -e '(file-stream-buffer-mode (current-output-port))' > bufmode.txt
●[Dec22 09:37:23] $ cat bufmode.txt 
'block
```

[^1]: https://docs.racket-lang.org/reference/file-ports.html#%28tech._file._stream._port%29
[^2]: https://docs.racket-lang.org/reference/port-buffers.html#%28def._%28%28quote._~23~25kernel%29._file-stream-buffer-mode%29%29

rfindler:
> This is fantastic!

Thank you! :) 

> I have some questions and some requests about the command-line user interface.
> 
>     * Will there be an option to explicitly enable and disable this behavior? Like `--XYZ` and `--no-XYZ`

I hadn't planned on adding extra command-line flags, as you can do `raco setup | cat` to effectively disable it.

>     * How does this interact with `/usr/bin/screen` (or tmux) when switching between windows?

I just tried using `screen`, detaching and reattaching and things seem to work. 

I don't know how those work internally, but I guess if they interpose between the process that's running and the terminal itself, nothing special should be required on the part of raco setup.

>     * By default, `-j 1` prints more information during `raco setup`. Has that changed?

That output shouldn't have changed. What I did was hijack the code that knows the natural number, eg the `2` in a line like this:

```
raco setup: 2 making: <pkgs>/data-doc/data/scribblings
```

If there is such a number available then it is now passed via the new `#:n` keyword argument to `setup-fprintf`, which then causes `setup-fprintf` to go into a more complex printing setup. All of the logic of "going back and overwriting a previous line" is based on that argument and in the new file setup-fprintf.rkt. If there isn't such a number available, then `setup-fprintf` bails out of the new complex printing mode and prints normally. In the "raco setup -j 1" case, we never enter into this mode as nothing will pass the `#:n` argument. Which, of course, means that "raco setup -j 1" doesn't get any improvements.

> 
> Also, I think it'd be better to detect the output type to decide if this printing behavior is going to be enabled. One necessary condition I have in mind is that the output port has to be a file stream port[1](#user-content-fn-1-971826b9fb38fc603022902f9f92d7ba) and `file-stream-buffer-mode`[2](#user-content-fn-2-971826b9fb38fc603022902f9f92d7ba) needs to be `'none` or `'line`, to avoid affecting piped/redirected outputs, interfering with other kinds of output ports or affecting other tools.
> 
> ```racket
> ●[Dec22 09:36:42] $ racket -e '(file-stream-buffer-mode (current-output-port))'
> 'line
> ●[Dec22 09:37:11] $ racket -e '(file-stream-buffer-mode (current-output-port))' | cat
> 'block
> ●[Dec22 09:37:14] $ racket -e '(file-stream-buffer-mode (current-output-port))' > bufmode.txt
> ●[Dec22 09:37:23] $ cat bufmode.txt 
> 'block
> ```

I think that these checks are triggered by [this code](https://github.com/rfindler/racket/blob/terminal-for-raco-setup/racket/collects/setup/private/setup-fprintf.rkt#L24-L28), thanks to me being able to reuse the `#%terminal` library that expeditor also uses. (Which reminds me-- it would be really great to test this under windows before merging!)

rfindler:
> > ```
> > * Maybe we can do more "clean up"? For example:
> >   
> >   * We can totally erase all idle lines at the end after all jobs finished -- this is at timestamp 3:09 and 7:48.
> > ```
> 
> That seems doable; there is logic in `setup-printf` that knows what the state of the screen is, so we'd probably just need to have another call into that from the place in the guts of setup that knows when each given phase is done.

It looks like the way the terminal controls work and how they interact with regular printing won't work out for this one, I'm sorry to say. Consider this program:

```
#lang racket
(require '#%terminal ffi/unsafe/port)
(define ip (current-input-port))
(define op (current-output-port))

(define terminal-available?
  (let ()
    (define in-fd (unsafe-port->file-descriptor ip))
    (define out-fd (unsafe-port->file-descriptor op))
    (and in-fd
         out-fd
         (terminal-init in-fd out-fd))))
(unless terminal-available? (error 'no-terminal))

(for ([i (in-range 10)])
  (printf "~a\n" i))
(void (terminal-move-cursor 'up 5))
(void (terminal-move-cursor 'left 1))
(void (terminal-write-char #\A))
(printf "B")
(flush-output)
```

When it runs, it leaves the cursor halfway back up in the lines with the digits on them (where the "A" goes), but the printf of "B" still puts its output down below all that, just where it would have put its output without the terminal-move-cursor calls. So, what this means is that, while I can erase the lines, I don't seem to be able to make the subsequent output go on top of the erased lines, so there would still be a bunch of blank lines there.

Unless someone sees something I'm missing?

rfindler:
Well, I'm starting to be a lot more skeptical of the "idle" messages in the compiling phase. After [this change](https://github.com/racket/racket/commit/bcdf5610a8876f63af3d89cb36fa0cc530c32600) I was watching the IO and I am seeing a lot of “idle” stuff, probably too much (like I even see everything idle sometimes).

So now I’m back to trying to understand what is happening in this case:

  https://github.com/racket/racket/blob/master/racket/collects/setup/parallel-build.rkt#L183-L185

I can’t seem to figure out how this extra false is influencing anything, however. After much tracing through the code, it sure looks like it ends up back as the first argument to `work-done` in `collects-queue%` (which makes total sense) and the pattern matching that's happening there seems to confirm that but, unfortunately, the place where the false goes is ignored. I think I must have missed something along the way. Like the code that actually *does* the work! It seems like that code must be the `parallel-build` function but I can't seem to find a place where there is a three element list whose third element is a boolean factors into anything. Probably there's more digging around I to do!

But now I'm starting to wonder if maybe this whole “show idle messages” is a fool’s game anyway. Even when something is ostensibly compiling a collection, it might not actually be doing any work, as it might be blocked, waiting for another worker to finish some file that it needs to proceed. So maybe it is pointless to try to put in idle messages for the compilation phase of raco setup?

samth:
I was thinking something like:
- a log-info message when a place starts working on something, which is when you initially write a line with a number and some task. The message should include basically the same information as the line you're printing out. 
- a log message with that same information when you switch to "idle". 

Does that clarify things?

rfindler:
> I was thinking something like:
> 
>     * a log-info message when a place starts working on something, which is when you initially write a line with a number and some task. The message should include basically the same information as the line you're printing out.
> 
>     * a log message with that same information when you switch to "idle".
> 
> 
> Does that clarify things?

Yes, that seems pretty easy to support -- just adding a log message while also updating the display. I do worry that the "idle" messages are wrong, tho, and it may not really be possible to give some kind of meaningful information about when something is "done" or when a core is idle, at least not without some work on the code that is actually doing the compilation. It does seem like something would need to somehow interpose at the level that it sees the requires in order to truly know what files are actually begin compiled by any particular worker.

The documentation running phase seems like it could have similar problems but things seem a lot less connected in practice, so maybe it could be ignored. And probably the render and re-rendering phases are all totally separate, so those would make sense.

What use do you have in mind for these messages? So far, I've been thinking about this only as a UI -- something to watch go by as you're waiting for raco setup to finish. So it is more about how things look visually rather than what actual information they are conveying. Like you're getting *some* information: things are changing means nothing froze; if you see larger letters in the alphabet you know you're closer to done (kind of). But not much more than that!

samth:
I'm hoping to use it to visualize the use of parallel processors, along the lines of the future visualizer. And I agree that there's lots more that could be done to improve the information, which I hope to donate some point but just the information you have easily would be a good start.

rfindler:
Okay, @samth, I think the current state actually is showing accurate information (assuming we don't spot any more bugs), so it may make sense to add some logging now (more details in [this commit message](https://github.com/racket/racket/pull/4878/commits/a74d0eec1f6012f29a49804a622f22620f77bbe5)). There's still more to do here before I really want to push to master, but I just wanted to let you know (I also tried to tidy up the set of commits to make things easier to follow if someone ever needs to do git blame at some point in the future).

rfindler:
I ran some timing tests; not sure how important this difference is. This is the average of two runs, using "time" in the shell and adding up "user" and "sys" categories. And "raco setup" had been run multiple times before so it wasn't doing any actual compilation or documentation building work (just running over everything to confirm that there wasn't anything to do.)

|Command|Seconds|
|--|--|
|time raco setup > /dev/null | 25.345|
|time raco setup \| cat | 25.793|
|time raco setup | 26.378 |

The difference between the second and third lines is the cost of the fancy output.

rfindler:
I think this is in pretty good shape now and could be pushed to master, but let's do it after the jan release branch is created.

@samth I didn't add those log messages but they should go exactly where `setup-fprintf` calls happen with `#:n` as the argument (or possibly inside `setup-fprintf`) so if someone wants to make a tool, hopefully it will be clear what to do for the data collection.

samth:
This is great. I'm happy to add the logging calls myself in a subsequent PR.

rfindler:
> This is great. I'm happy to add the logging calls myself in a subsequent PR.

Okay, great. I think that probably it makes sense to have the use of them in hand when adding the calls, so that'll probably work out better. 

Let me know if I can help!

samth:
Having now tried this, I have a few thoughts:

1. It's awesome! I've wanted something like this forever (but haven't ever done anything about it). Thank you, Robby, for making it happen.
2. I find a extra blank line after the progress bar surprising, which I see is also in the video Robby posted. Could that be eliminated?
3. Is there enough information when `awaiting ...` is printed to say which collection or file is waiting on that zo file?
4. When an error is printed the results are not great. Is that inevitable? Do other build systems manage this in a different way?

Having this also suggests lots of other exciting improvements, like using terminal control with other parts of the `raco setup` output, or using fancier characters and color, or applying this to `raco make -j`, or having finer-grained work in `raco setup` itself so the beginning doesn't spend quite so much time serialized waiting for Typed Racket and plot and math to compile. 

I'm looking forward to having this!

rfindler:
> Having now tried this, I have a few thoughts:
> 
>     1. It's awesome! I've wanted something like this forever (but haven't ever done anything about it). Thank you, Robby, for making it happen.
> 
> 
>     3. Is there enough information when `awaiting ...` is printed to say which collection or file is waiting on that zo file?

Yes, there should be. The way the code is structured, I know which id is working on which collection and the awaiting lines are written over that and then, when the awaiting is done, the previous line is replaced. The answer should always be one of the other "making" lines, of course.

All that said, there is a bug where I see 7 awaitings and one idle and I think I've even seen 8 awaitings (on an 8 core machine) which is clearly wrong. I don't know where the bug is, however.

>     2. I find a extra blank line after the progress bar surprising, which I see is also in the video Robby posted. Could that be eliminated?
>
>     4. When an error is printed the results are not great. Is that inevitable? Do other build systems manage this in a different way?

This particular change is definitely in the "don't make a big change to anything" category. Raco setup has lots of random code in it that just decides to print stuff out. If we wanted to do something more sophisticated, we could, and a good first step towards that would be to refactor all of the printing so that it all goes through some slightly higher-level interface that we could implement both with terminal control as well as just with regular printing (for the situation when terminal control isn't available). Going through and refactoring all that would reveal what the full set of possible prints are and when they happen and then we could start to think harder about how to make them look nice.

So, how the current situation came to be (the code on this branch) is that I just found everywhere that prints lines of the form "raco setup: <n> <action>: <information>" and it was relatively easy because there is a function "setup-printf" that already is the only place that puts that "raco setup" prefix on the output. Then, I changed the implementation of that function go up and overwrite the previous line with the same value of "n" in it. (Of course, along the way I ended up going into the code that was actually generating the prints in order to print more useful stuff, which is how we got the awaiting lines and there are a few other differences in terminal-printing mode that maybe aren't worth going into right now.)

This means that the code that still is doing whatever random printing it wants (eg when an error is found in compiling a file or when a cross reference is messed up in the documentation or probably another dozen things) will print and I won't know it is coming. But the print should still look reasonable. So the approach I took was to just be ready for that to happen at any point and recover, if it does. This "allow it to print" and recover doesn't just handle "abnormal" printing for errors, but it is also triggered when we finish compiling and move on to the next stage of raco setup. 

I can elaborate more on the details if anyone's curious but the basic idea of this change is very much tied up in the goal to avoid doing that big audit/refactor that I mentioned above. The blank line is part of this -- we need to leave space for whatever printout might happen and I take over that bottom line, then the printout will overwrite it, but we'll see some of the leftovers after the end of the line that is doing the regular printout.

Anyway, I hope this makes _some_ sense!

rfindler:
@mflatt @samth Is the easter egg a bad idea? It makes the progress bar be red & green on christmas....

mflatt:
> Is the easter egg a bad idea?

That looks very manageable, so no objection from me.

mfelleisen:
12/25 huh?

samth:
Seems fine to me. That code seems like it can't go wrong and is relatively easy to test.

shhyou:
1. Can the Easter egg of the progress bar be less flashy (in terms of frequency)? If that's difficult to do with `random`, what about changing colors deterministically?
2. The progress bar does not show up when `raco` is running the `.scrbl` files. It does show up when building the `.zo`s and when rendering the `.scrbl` files.

rfindler:
> * Can the Easter egg of the progress bar be less flashy (in terms of frequency)? If that's difficult to do with `random`, what about changing colors deterministically?

How about this? Does it look better to you? (It seems like an improvement to me but I'm not sure if it is enough.)

```
--- a/racket/collects/setup/private/setup-fprintf.rkt
+++ b/racket/collects/setup/private/setup-fprintf.rkt
@@ -105,7 +105,7 @@
              [(= %age 1) 7]
              [(and (= (date-day today) 25)
                    (= (date-month today) 12))
-              (+ (random 2) 1)]
+              (if (< (random) %age) 1 2)]
              [else 7]))
          (terminal-set-color fg-color #f)
          (for ([i (in-range w)])
```

> * The progress bar does not show up when `raco` is running the `.scrbl` files. It does show up when building the `.zo`s and when rendering the `.scrbl` files.

Yeah, that phase doesn't seem to know how many there are to run; it discovers it as it goes, currently. There may be some way to count but, unlike the other phases, the code isn't already there to do the counting.

rfindler:
I think it is hard to predict how often it is going to be called; some scenarios it gets called a lot and others not so much so something based on real time is probably better than counts.

shhyou:
Yes! I think based on the actual time is the best option. Counting is just to show that using `random` is going to change the color too quickly.

rfindler:
@shhyou I have tried a different approach to the colors with the latest commit. What do you think of that?

rfindler:
Once we're happy with the easter egg, then I think it is okay to merge this but without the last commit (I probably shouldn't have even pushed it here, actually!). So I'll remove that commit before merging.

shhyou:
Looks great to me!

samth:
I agree, looks great. As to the last commit, I think we should first build the (very large) graph of all the dependencies and then figure out from there how to parallelize better.

rfindler:
Okay, I got rid of the last commit and I think we're ready to merge here.

sorawee:
Some feedback:

Screenshot 2024-01-07 at 2 28 47 PM

- Progress bar usually comes with estimated time. Is it easy to add one?
- The "awaiting" lines don't seem informative, especially when there are multiple of them. Maybe `making: <X> (awaiting <Y>)` will be better?
- Perhaps there should be a blank line before the progress bar?

samth:
I think estimated time would be very hard. The progress bar as it is does much more work in the early percentiles because it compiles dependencies (like Typed Racket and math) which doesn't increase the progress percentage.

rfindler:
> * Progress bar usually comes with estimated time. Is it easy to add one?

I also think it is hard to get a true time. The way the progress bar works is file-based. That is, the first thing raco setup seems to do is to collect a list of all of the files that it will compile and then it organizes them into "cc"s, which are roughly collections (and subcollections). Then, each time it finishes a file, it updates the progress (but it prints only the name of the collection it is working on instead of every single file). 

You can see this yourself if you want. To do so, run "raco setup -c" to delete all .zo files, and then run "raco setup" until it gets to, say, 60%. Then type control-c and run "raco setup" again. In the second run, it will very quickly get up to that same spot where you typed control-c and then it will slow down once it passes that point.

So, even beyond the issue that we're running arbitrary racket code to compile (which is hard to predict timing for), there's also the issue that we don't know, ahead of time, which files are going to require recompilation and which aren't.

> * The "awaiting" lines don't seem informative, especially when there are multiple of them. Maybe `making: <X> (awaiting <Y>)` will be better?

I think the awaiting lines require you to understand what's going on under the hood and I'm comfortable with that. In the suggested change, the "X" and "Y" is actually redundant and will have a lot of overlap ("X" will usually be a substring of "Y" I think).

As for the situation where there are many awaiting of the same file that's actually very informative, I'd say. It is telling you that those N parallel workers all need that one file (which is a bad situation!)

That said, there is a bug in the awaiting lines that I could not figure out. Specifically, it is possible to see one idle line and 7 waiting lines on my 8 core machine and I think I even saw 8 awaiting. If you have the energy to dig into parallel-compile.rkt and parallel-go.rkt, maybe you'll be able to figure out what the bug is! (I feel somewhat confident that I didn't introduce it with my changes but that's purely based on me reading and rereading (and rereading) my changes -- not any kind of principled argument!)

> * Perhaps there should be a blank line before the progress bar?

Personally, I prefer not to require another line but I could see that reasonable people might differ on this point.

---

I wonder if there is any way to query the terminal to ask if it is in white-on-black or black-on-white mode ......... (sigh). At least it looks kind of reasonable in your screenshot!

sorawee:
> I wonder if there is any way to query the terminal to ask if it is in white-on-black or black-on-white mode ......... (sigh). At least it looks kind of reasonable in your screenshot!

https://stackoverflow.com/a/30540928/718349 seems to work for me (iTerm2, Mac M2)

benknoble:
> I wonder if there is any way to query the terminal to ask if it is in white-on-black or black-on-white mode ......... (sigh). At least it looks kind of reasonable in your screenshot!

Vim has struggled for a long time with "background detection" (light or dark). I think it eventually gave up on doing so reliably, but you could poke around there.

PR #4879, opened 2023-12-22
add `vector-copy`, `vector-append` and `vector-set/copy` as primitives

mflatt:
These functions can be implemented with `make-vector` and vector-set!`, but adding them as primitives can avoid redundant initial and write barriers, which provides better performance for implementing persistent data structures such as RRB trees.

The `racket/vector` library already provides `vector-copy` and `vector-append`, so those are not new, except that they're now in `#%kernel`. The `vector-update` function is new, and so there's some danger of a collision, but I haven't found one, yet.


benknoble:
I would have expected vector-update to compute the new element from the old one (like hash-update) or be called vector-set (like hash-set).

mflatt:
Good point about `vector-update`. Using `vector-set` seems unhelpfully close to `vector-set!` — and unilke `hash-set` versus `hash-set!`, `vector-set` would apply to the same values as `vector-set!`. Maybe `vector-copy-set`?

AlexKnauth:
or `vector-set/copy`

mflatt:
The new version has `vector-set/copy` instead of `vector-update`.

PR #4883, opened 2023-12-28
fix FFI identifier convention for camelCase

### LINKED ISSUE #1971: ffi-definer convention for camelCase is incorrect
see: [conventions.rkt](https://github.com/racket/racket/blob/2f6c42f3c1a14a1bb3c8780b84f3b23ea7229526/racket/collects/ffi/unsafe/define/conventions.rkt)
Existing ffi convention: `convention:hyphen->camelcase` returns PascalCase instead of camelCase, ie: 
`(convention:hyphen->camelcase #'tEst-me-To-see-what-haPpens)` returns TestMeToSeeWhatHappens

suggested solution:
- rename `convention:hyphen->camelcase` to `convention:hyphen->pascalcase`
- add new `convention:hyphen->camelcase`:
```
(define (convention:hyphen->camelcase id)
  (define str (symbol->string (syntax-e id)))
  (define str-lst (string-split str "-"))
  (format-id id
             (apply string-append (string-downcase (car str-lst))
                    (map string-titlecase (cdr str-lst)))))
```
the new camelcase result:
`(convention:hyphen->camelcase #'tEst-me-To-see-what-haPpens)` returns testMeToSeeWhatHappens

usaoc:
##### Checklist
- [x] Bugfix
- [x] tests included
- [x] documentation

### Description of change
Previously, `convention:hyphen->camelcase` had the wrong behavior, that is, that of the newly added `convention:hyphen->PascalCase`.  Deprecate `convention:hyphen->camelcase` and add `convention:hyphen->{camelCase,PascalCase}`.

Fix #1971.

### DIFF
--- a/pkgs/racket-doc/scribblings/foreign/define.scrbl
+++ b/pkgs/racket-doc/scribblings/foreign/define.scrbl
@@ -1,5 +1,8 @@
 #lang scribble/doc
-@(require "utils.rkt" (for-label ffi/unsafe/define ffi/unsafe/alloc))
+@(require "utils.rkt"
+          (for-label ffi/unsafe/define
+                     ffi/unsafe/alloc
+                     ffi/unsafe/define/conventions))
 
 @title{Defining Bindings}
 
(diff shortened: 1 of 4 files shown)


mflatt:
Does this break existing code that uses `convention:hyphen->camelcase`?

usaoc:
Yes, if there is any; given the *wrong* behavior, I quite doubt that (someone please correct me), but it’s easy to migrate (`convention:hyphen->pascalcase` has the old behavior).

mflatt:
I think we'll need to check existing packages. If some package registered at `pkgs.racket-lang.org` would be broken by a backward-incompatible changes, then there's a much higher bar to make the change (i.e., we just don't, unless we go out of our way to agree that it's an exceptional case).

mflatt:
There appear to be uses of `convention:hyphen->camelcase` in the `keyring-keychain-lib`, `fontconfig`, and `fontland` packages.

Maybe the right change is to add `convention:hyphen->camelCase` and `convention:hyphen->PascalCase` and document `convention:hyphen->camelcase` as a deprecated name for `convention:hyphen->PascalCase`, or something like that.

usaoc:
I incorporated @mflatt’s suggestion.

PR #4884, opened 2023-12-28
Reduce memory footprint of net/http-client

Bogdanp:
This change reduces some of the imports in `net-lib` and makes it so that `win32-ssl.rkt` doesn't have to be loaded on non-Windows platforms and vice-versa for `osx-ssl.rkt` and non-macOS platforms, improving the memory footprint of `net/http-client` by about 5MB.

Before this change:

```
$ racket -l racket/base -e '(define s (current-memory-use)) (require net/http-client) (- (current-memory-use) s)' | awk '{print $1/1024/1024}'
30.8346
```

After:

```
$ racket -l racket/base -e '(define s (current-memory-use)) (require net/http-client) (- (current-memory-use) s)' | awk '{print $1/1024/1024}'
25.5956
```

### DIFF
--- a/racket/collects/ffi/unsafe/define.rkt
+++ b/racket/collects/ffi/unsafe/define.rkt
@@ -1,6 +1,6 @@
 #lang racket/base
 
-(require (for-syntax syntax/parse
+(require (for-syntax syntax/parse/pre
                      racket/base)
          ffi/unsafe)
 
--- a/racket/collects/net/http-client.rkt
+++ b/racket/collects/net/http-client.rkt
@@ -1,16 +1,16 @@
 #lang racket/base
-(require racket/contract/base
-         racket/match
+
+(require file/gunzip
+         openssl
+         racket/contract/base
          racket/list
-         racket/string
+         racket/match
          racket/port
+         racket/string
          (rename-in racket/tcp
                     [tcp-connect plain-tcp-connect]
                     [tcp-abandon-port plain-tcp-abandon-port])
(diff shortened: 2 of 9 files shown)


sorawee:
Have you tried `lazy-require`? How is that different from the approach in `platform-ssl.rkt`?

Bogdanp:
> Have you tried lazy-require? How is that different from the approach in platform-ssl.rkt?

I've thought about it, but I don't think it would help since `http-client.rkt` calls `osx-old-openssl?` and `win32-ssl-available?`, so it would end up loading the modules regardless.

mflatt:
Is it important to get this into v8.12? If not, this might be a good one to merge after the release branch to get extra checking before merging, just in case.

Bogdanp:
> Is it important to get this into v8.12? If not, this might be a good one to merge after the release branch to get extra checking before merging, just in case.

Not at all! I'm happy to wait until after the release.

samth:
@Bogdanp is this ready to go?

Bogdanp:
@samth yep.

mflatt:
I overlooked that this implementation expands a macro differently depending on the value of `(cross-system-type)`. Making a macro expansion platform-dependent like that does not work right for distributing code in "built" form, like the source + built distribution.

It looks like expanding to use `(system-type)` at run time (@rfindler's suggestion) would achieve the same effect with just a little more generated code, and without making macro expansion depend on the platform. Does that sound right?

PR #4885, opened 2023-12-28
add pregexp-quote

Bogdanp:
Adds `pregexp-quote`, an analog of `regexp-quote` for `pregexp` values that works like Perl's `quotemeta`.

### DIFF
--- a/pkgs/racket-test/tests/racket/pregexp.rkt
+++ b/pkgs/racket-test/tests/racket/pregexp.rkt
@@ -1,0 +1,18 @@
+#lang racket/base
+
+(require rackunit)
+
+(check-equal?
+ (regexp-quote "ab-cd")
+ "ab\\-cd")
+
+(let ([re (pregexp (pregexp-quote "-"))])
+  (check-true (regexp-match-exact? re "-"))
+  (check-false (regexp-match-exact? re "a")))
+
+(let ([re (pregexp (format "[~a]" (pregexp-quote ")-;")))])
+  (for ([s (in-list '(")" "-" ";"))])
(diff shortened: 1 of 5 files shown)


sorawee:
The CI error seems really weird:

```
Section(cm)
ERROR: compile-zo: date for newly created .zo file (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/plt-cm-test/compiled/COLLECTS-racket-file.rkt-586-20_17037866041703786604393 @ 2023-12-28 18:3:24) is before source-file date (/var/folders/24/8k48jl6d249_n_qfxwsl6xvm0000gn/T/plt-cm-test/e.rkt @ 2023-12-28 18:3:27), which appears to be in the future
```

I will rerun the job in case it's an intermittent failure.

pkgs/racket-test/tests/racket/pregexp.rkt
```diff
@@ -0,0 +1,18 @@
+#lang racket/base
+
+(require rackunit)
+
+(check-equal?
+ (regexp-quote "ab-cd")
```
sorawee, line 6: Shouldn't this be `pregexp-quote`? Is this test file running in the CI?

mflatt: Yes, I overlooked that, and I'll push a repair. Probably only tests are `racket-test-core` run in GitHub CI.

samth: https://github.com/racket/racket/blob/master/.github/workflows/ci-pr.yml#L23 is what runs in CI. Probably more files should be added there.

PR #4887, opened 2023-12-31
Chez Scheme: add add1, sub1 and abs to cptypes

gus-massa:
##### Checklist

- [x] Feature
- [x] Tests included

### Description of change
Add special cases for `add1`, `sub1` and `abs` to cptypes. In particular, that they are closed for flonums, reals and exact-integers.


samth:
Does cptypes have a type for numbers that might be returned by length or vector-length (bigger than a byte, smaller than a fixnum)?

mflatt:
Thanks, @gus-massa! But now that Racket's Chez Scheme is the same as https://github.com/cisco/ChezScheme, probably this pull request should start on that repo.

gus-massa:
@mflatt: OK

@samth: There are no "small fixnums" type. I isolated exact integer to do something like that, and even allow arbitrary intervals, but it's has too many corner cases. Perhaps add only `lengh`s isa good idea, but I have to take look becuse each thing can have a different maximal size. Do you think it would be useful to add `byte`s too?

samth:
I found a smaller fixnum type, corresponding to lengths, to be necessary for getting useful optimization results for loops in Typed Racket. Typed Racket also has a byte type but I'm less sure that's useful.

samth:
A question about these types: is the `sub-length` or `sub-index` type used in `primdata.ss` the kind of type we're discussing?

gus-massa:
Reposted in https://github.com/cisco/ChezScheme/pull/790

gus-massa:
@samth: I don't remember the diference between `length` and `index`. Some types are essentialy duplicated.

The idea would be to define `lenght` as a positive 28bit fixnum, where 28 is a made up number. I should read the implementation details. IIRC the problem were long lists,, whose length can be almost a greatest fixnum in 32 bits if they use *all* the memory of the computer.

`sub-length`  is an internal type of primdata to avoid testing the primitive, so it can say that the index in `vector-ref` is a `length` but it doen not try to use the standard value of `length` in the standard example of `vector` because it may be too short and raise an error.

In cptypes, `sub-length` is bounded by `'bottom` and `'fixmun`. The idea woub be to use a better upper bound. 

(I'm not planing to implement this soon,  many details in my comment may be wrong.)

PR #4890, opened 2024-01-02
adjust inexact->exact error message when given, eg, +inf.0

rfindler:
Before this, we have this:

```
> (inexact->exact +inf.0)
exact: no exact representation for +inf.0 [,bt for context]
```

this commit fixes it to this:

```
> (inexact->exact +inf.0)
inexact->exact: no exact representation for +inf.0 [,bt for context]
```

but perhaps even better would be:

```
> (inexact->exact +inf.0)
inexact->exact: contract violation
  expected: rational?
  given: +inf.0
 [,bt for context]
```

I see that we get something better here (but `number?` appears to be too loose):

```
> (inexact->exact "wrong")
inexact->exact: contract violation
  expected: number?
  given: "wrong"
 [,bt for context]
```

Any opinions on whether to try for the improved error message or if `rational?` is the right predicate?


rocketnia:
I think `rational?` is the wrong contract for this since `inexact->exact` works on complex numbers with `rational?` excludes anything with a non-exact-zero real component.

Also, `rational?` sounds potentially confusing even when it is the right predicate. I think it points to the bug being that we've somehow passed an irrational number in, not that we've somehow passed in an infinity or NaN.

Maybe in the spirit of the original error message, there could be a `number-with-exact-equivalent?` predicate. Something like `(complex/c (not/c (or/c nan? infinite?)))`, where `(complex/c c)` checks the flat contract `c` on each rectangular component, might be nice too.

rfindler:
Ah, right! Complex numbers work too. If we had a complex/c, then maybe:

```
(or/c rational?
      (complex/c rational?
                 rational?))
```

would be right?

And I agree that the name `rational?` is a confusing name!

mfelleisen:
[[ Introduce ω and have `(inexact->exact +inf.0)` yield it :-)  ]]

Seriously, `finite?` doesnt seem to apply to “wrong”.

rfindler:
Maybe just merging this is best for the time being?

samth:
Yes, I agree.

PR #4893, opened 2024-01-07
doc: clarify some `raco setup` APIs

LiberalArtist:
In particular:
  - Explain how `find-relevant-directories` treats package-level `info.rkt` files.
  - Note that a normalized result of `path->module-path` more specifically satisfies `normalized-lib-module-path?`.
  - Mark many `#:cache` arguments as optional.


samth:
This all looks right to me but maybe @mflatt should review it too.

mflatt:
Thanks!

Issue #4864, opened 2023-12-11
`raco setup` is slow with little work to do


samth:
Currently, running `raco setup` is quite slow even if it does very little work. For example, `raco setup doesntexist` takes about 2.9 seconds on my machine, and `raco setup racket/private` takes 12 seconds (with no actual work to do). 

This issue is for recording my investigation/improvement to this, but help is always welcome.

samth:
With the changes from #4896, 70% of the remaining time for `raco setup notexist` is spent in the `omitted-paths` function from `setup/private/omitted-paths`.

PR #4896, opened 2024-01-12
Avoid calling path->pkg repeatedly when it is unlikely to be necessary.

samth:
This saves about 5 seconds on my machine when building with no documents (eg `raco setup racket/private`).

I did not try to memoize the resulting `path->pkg` computation. Storing a promise in a serializable struct won't work, and making the struct mutable seemed like a bad idea.

### DIFF
--- a/pkgs/racket-index/setup/scribble.rkt
+++ b/pkgs/racket-index/setup/scribble.rkt
@@ -27,8 +27,8 @@
          scribble/xref
          syntax/modcollapse
          racket/place
-         pkg/lib
-         pkg/strip
+         (only-in pkg/lib path->pkg)
+         (only-in pkg/strip fixup-local-redirect-reference)
          openssl/sha1
          compiler/compilation-path
          (prefix-in u: net/url)
@@ -59,7 +59,6 @@
                                  flags
                                  under-main?
                                  via-search?
-                                 pkg?
                                  category
                                  out-count
                                  name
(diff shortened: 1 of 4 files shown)


samth:
Related to #4864.

pkgs/racket-index/setup/scribble.rkt
```diff
@@ -1717,3 +1718,7 @@
     (build-path base root)]
    [else
     (reroot-path base root)]))
+
+
+(define (doc-pkg? doc)
+  (and (path->pkg (doc-src-file doc)) #t))
```
sorawee, line 1724: There's already a `#:cache` argument in `path->pkg` for memoization, right?

rfindler: FWIW that `cache` argument just need to be an empty hash to start (and then not used for anything else). The docs aren't super clear on how to use that argument but almost every call to `path->pkg` (and other similar functions) should be using that argument.

samth: Ok, I updated that to use `#:cache`.

pkgs/racket-index/setup/scribble.rkt
```diff
@@ -184,48 +183,50 @@
...
+				 (list* (car spec)
+					(car d)
+					(if (eq? 'planet (car spec))
+					    (list (append (cdr spec)
+							  (list (directory-record-maj rec)
+								(list '= (directory-record-min rec)))))
+					    (cdr spec))))
+			       src
+			       dest
+			       flags under-main? via-search?
```
sorawee, line 219: I'm confused. So you define `doc-pkg?` below, which appears to be unused, and the `(and (path->pkg src) #t)` argument here is gone. Is this correct?

sorawee: Ah, so that `doc-pkg?` shadows the existing `doc-pkg?`, which is a struct field ref. 

Shouldn't the existing struct field ref be fast already though?

samth: The field no longer exists in the changed code. It's computing the value of the field that is slow.

sorawee: Ah, got it!

samth:
The added commit speed up `raco setup notexist` by about 17% (3 sec -> 2.5 sec).

mflatt:
LGTM, as far as I can tell.

Issue #4898, opened 2024-01-13
Release Regressions for v8.12


jbclements:
Current Release vs Next Release
========================================
>>> fetching https://pkg-build.racket-lang.org/summary.rktd 
>>> fetching https://plt.cs.northwestern.edu/release-pkg-build/summary.rktd

### Build Failures
- [x] redex-etc https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/redex-etc.txt

#### Network failure 
- [x] cbor https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/cbor.txt
- [x] r16 https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/r16.txt
- [x] racket-glossary https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/racket-glossary.txt 
- [x] raco-exe-multitarget https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/raco-exe-multitarget.txt
- [x] sudoku-solver https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/sudoku-solver.txt
- [x] todo-txt https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/todo-txt.txt
- [x] unmozlz4 https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/unmozlz4.txt
- [x] zstd https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/zstd.txt

### Test Failures
- [x] rascas https://plt.cs.northwestern.edu/release-pkg-build/server/built/test-fail/rascas.txt
- [x] crypto-test https://plt.cs.northwestern.edu/release-pkg-build/server/built/test-fail/crypto-test.txt
- [ ] process-queue https://plt.cs.northwestern.edu/release-pkg-build/server/built/test-fail/process-queue.txt
- [x] timable https://plt.cs.northwestern.edu/release-pkg-build/server/built/test-fail/timable.txt

### Other Failures
- [ ] video-v0-0

rfindler:
Assuming I found [the right repo](https://github.com/Metaxal/rascas/tree/HEAD), it hasn't changed in 10 months but I'm confused because the checksum in the build failure transcript is `9df641129d04f750ae9f511ff8125d9aca7e2279` but that's not the latest commit nor the checksum on pkgs.racket-lang.org (which are both `f575e5a63b02358e80bfde53c7cb3eaae396e41a`)

[...]

PR #4899, opened 2024-01-16
fix `for/fold` binding in the presence of `#:result`

usaoc:
##### Checklist
- [x] Bugfix
- [x] tests included

### Description of change
This bug was introduced in 3b5b0a7.  The `fold-var`s incorrectly scoped over the whole `for/fold` expansion, while the “remaining weirdness” behavior was that outermost iteration clauses should not see the `fold-var`s.

Also revert the `for/foldr` change that appears to be incorrect.  Unlike `for/fold`, `for/foldr` does not need to use this approach, because `fold-var`s are only visible to the body.

Related to #4898.


mflatt:
Yes, this looks good — thanks!

sorawee:
We want to include this to the release branch of 8.12, right? (cc: @jbclements)

mflatt:
Yes. I'm not sure what phase we're in, but I'm starting the release-management request now, in case it's needed.

PR #4906, opened 2024-01-22
fix some paths that raise “required keyword not supplied”

usaoc:
##### Checklist
- [x] Bugfix
- [x] tests included


mflatt:
Thanks!

PR #4916, opened 2024-01-30
added make-hash and hash-map examples

dr-neptune:
##### Checklist

- [X] documentation

### Description of change

This change adds examples to the documentation for `make-hash` with assocs and `hash-map`

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -212,6 +212,13 @@ mappings can hide earlier mappings.
 
 See also @racket[make-custom-hash].
 
+@examples[
+#:eval the-eval
+(make-hash)
+(make-hash '((0 1) (1 2) (2 3)))
+(make-hash '((0 1) (0 2)))
+]
+
 @history[#:changed "8.5.0.3" @elem{Added @racket[make-hashalw].}]}
 
 @deftogether[(
@@ -630,6 +637,12 @@ with the following order (earlier bullets before later):
 
 @history[#:changed "6.3" @elem{Added the @racket[try-order?] argument.}
          #:changed "7.1.0.7" @elem{Added guarantees for @racket[try-order?].}]}
+
+@examples[
+#:eval the-eval
+(hash-map (make-hash '((0 1) (1 2) (2 3))) (λ (k v) k))
+(hash-map (make-hash '((0 1) (1 2) (2 3))) (λ (k v) v))
+]
 
 @defproc[(hash-map/copy
           [hash hash?]
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -215,11 +215,11 @@ See also @racket[make-custom-hash].
 @examples[
 #:eval the-eval
 (make-hash)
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -212,6 +212,13 @@ mappings can hide earlier mappings.
 
 See also @racket[make-custom-hash].
 
+@examples[
+#:eval the-eval
+(make-hash)
+(make-hash '((0 1) (1 2) (2 3)))
```
sorawee, line 218: Do you really intend to have `'(1)`, `'(2)`, and `'(3)` as the values here?

dr-neptune: No, the reason I made this change was because I couldn't find the assocs written anywhere. Is there a way to make the values just 1, 2, 3?

sorawee: `'([0 . 1] [1 . 2] [2 . 3])` would be a way to do it. Alternatively, you could also do `(list (cons 0 1) (cons 1 2) (cons 2 3))`, which is more general, since the key/values don't need to be `quote`-able.

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -212,6 +212,13 @@ mappings can hide earlier mappings.
...
+(make-hash '((0 1) (0 2)))
```
samth, line 219: Could we add an example with elements that aren't just numbers?

pkgs/racket-doc/scribblings/reference/hashes.scrbl
dr-neptune: Thanks, I appreciate the examples. I have updated the docs to show various ways to instantiate initial values

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -212,6 +212,16 @@ mappings can hide earlier mappings.
...
 See also @racket[make-custom-hash].
 
+@examples[
+#:eval the-eval
+(make-hash)
+(make-hash '([0 . 1] [1 . 2] [2 . 3]))
+(make-hash '([0 . 1] [1 . 2] [0 . 3]))
+(make-hash (list (cons 0 1) (cons 1 2) (cons 2 3)))
+(make-hash '((0 1) (1 2) (2 3)))
+(make-hash (list (cons "hello" "world") (cons 42 "meaning of life") (cons (list 1 2 3) (list #\a #\b #\c)) (cons #t #f) (cons 'apple 'orange) (cons (list 'a 1 "apple") (vector #t 'b 2 "banana")) (cons + -)))
```
sorawee, line 222: I think this is a bit too much, and it wouldn't be pretty in the rendered doc.

Screenshot 2024-02-02 at 7 25 13 AM

sorawee (changes requested):

pkgs/racket-doc/scribblings/reference/hashes.scrbl
dr-neptune: haha, yes that is _very_ reasonable. I will trim it back

sorawee (approved):

PR #4917, opened 2024-01-30
Key filter val filter

dr-neptune:
##### Checklist

- [x] Feature
- [x] tests included
- [x] documentation

### Description of change

This change adds the following 2 functions to `racket/hash`:
- `hash-filter-keys`: given a hash table and a predicate, return a hash table where the predicate is true for the input hash tables keys
- `hash-filter-values`: the same as for keys, but applies the predicate to the input hash tables values

This is my first contribution to `racket/hash`, please let me know if I am missing anything or if there are broader guidelines for contributing that are applicable here

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -1005,4 +1005,34 @@ one for the result.
 
 @history[#:added "7.9.0.1"]}
 
+
+@defproc[(hash-filter-values [ht hash?] [pred procedure?])
+         hash?]{
+
+Filters a hash table by applying a predicate to it's values.
+
+@examples[
+#:eval the-eval
+(hash-filter-values (for/hash ([num '(1 2 3 4 5)]) (values num (add1 num))) (λ (v) (< v 3)))
+(hash-filter-values (make-hash) (λ (v) (< v 3)))
+]
+
+@history[#:added "8.11.1"]
+}
+
+
+@defproc[(hash-filter-keys [ht hash?] [pred procedure?])
+         hash?]{
+
+Filters a hash table by applying a predicate to it's keys.
+
+@examples[
+#:eval the-eval
+(hash-filter-keys (for/hash ([num '(1 2 3 4 5)]) (values num 0)) (λ (k) (< k 3)))
+(hash-filter-keys (make-hash) (λ (k) (< k 3)))
+]
+
+@history[#:added "8.11.1"]
+}
+
 @(close-eval the-eval)
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -1040,36 +1040,80 @@ depends on both the key and value meeting the criteria defined by @racket[pred].
                (λ (k v) (and (not (number? k)) (number? v) (> v 1))))
 ]
 
-@history[#:added "8.12.0.6"]
+@history[#:added "8.12.0.9"]
 }
-
-@defproc[(hash-filter-values [ht hash?] [pred procedure?])
-         hash?]{
-
-Filters a hash table by applying a predicate to it's values.
-
-@examples[
-#:eval the-eval
-(hash-filter-values (for/hash ([num '(1 2 3 4 5)]) (values num (add1 num))) (λ (v) (< v 3)))
-(hash-filter-values (make-hash) (λ (v) (< v 3)))
-]
-
-@history[#:added "8.11.1"]
-}
-
 
 @defproc[(hash-filter-keys [ht hash?] [pred procedure?])
          hash?]{
 
-Filters a hash table by applying a predicate to it's keys.
+Filters the @racket[hash?] @racket[ht] based on a predicate @racket[pred] applied to its keys.
+This function constructs a new hash table that includes only those key-value pairs
+from the input @racket[ht] for which the predicate @racket[pred] returns true when
+applied to the keys. Similar to @racket[hash-filter-values], the output hash table
+maintains the mutability and key comparator of the input hash table, ensuring that
+the structural and operational properties of the original hash are retained.
 
 @examples[
-#:eval the-eval
-(hash-filter-keys (for/hash ([num '(1 2 3 4 5)]) (values num 0)) (λ (k) (< k 3)))
-(hash-filter-keys (make-hash) (λ (k) (< k 3)))
+  #:eval the-eval
+  ;; Filtering keys less than 3 from a hash table
+  (hash-filter-keys (for/hash ([num '(1 2 3 4 5)]) (values num 0)) (λ (k) (< k 3)))
+
+  ;; Filtering keys from an empty hash table
+  (hash-filter-keys (make-hash) (λ (k) (< k 3)))
+
+  ;; Filtering with eq? hash table
+  (hash-filter-keys (make-hasheq '([#f . "false"] [#t . "true"])) (λ (k) (eq? k #t)))
+
+  ;; Filtering lists as keys
+  (hash-filter-keys (hash (list 1 2) 'pair (vector 3 4) 'vector) list?)
+
+  ;; Filtering keys of mixed types: numbers and strings
+  (hash-filter-keys (hash "one" 1 2 "two" "three" 3) (lambda (k) (number? k)))
+
+  ;; Filtering keys that are symbols
+  (hash-filter-keys (hash 'apple "fruit" 'carrot "vegetable" "banana" "fruit")
+                    (lambda (k) (symbol? k)))
 ]
 
-@history[#:added "8.11.1"]
+@history[#:added "8.12.0.9"]
+}
+
+
+@defproc[(hash-filter-values [ht hash?] [pred procedure?])
+         hash?]{
+
+Filters the @racket[hash?] @racket[ht] based on a predicate @racket[pred] applied to its values.
+This function returns a new hash table containing only the key-value pairs for which
+the predicate @racket[pred] returns true when applied to the values of @racket[ht].
+The resulting hash table retains the mutability and the key comparison predicate
+(e.g., @racket[eq?], @racket[eqv?], @racket[equal-always?], @racket[equal?]) of the input
+hash table @racket[ht]. This ensures that the characteristics of the input hash are preserved
+in the output, allowing for a seamless transition between different types of hash tables.
+
+@examples[
+   #:eval the-eval
+   ;; Filtering values less than 3
+   (hash-filter-values (for/hash ([num '(1 2 3 4 5)]) (values num num)) (λ (v) (< v 3)))
+
+   ;; Filtering values from an empty hash table
+   (hash-filter-values (make-hash) (λ (v) (< v 3)))
+
+   ;; Filtering with eqv? hash table
+   (hash-filter-values (make-hasheqv '([1 . "one"] [2 . "two"])) (λ (v) (eqv? v "two")))
+
+   ;; Filtering values of mixed types: strings and numbers
+   (hash-filter-values (hash 'one "1" 'two 2 'three "3") (lambda (v) (string? v)))
+
+   ;; Filtering values to include only vectors
+   (hash-filter-values (hash 'list (list 1 2 3) 'vector #(4 5 6) 'string "hello")
+                       (lambda (v) (vector? v)))
+
+   ;; Filtering based on complex values (hash tables and lists)
+   (hash-filter-values (hash 'nested-hash (hash 'a 1 'b 2) 'nested-list (list 'x 'y 'z))
+                       (lambda (v) (hash? v)))
+ ]
+
+@history[#:added "8.12.0.9"]
 }
 
 @(close-eval the-eval)
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -1005,6 +1005,43 @@ one for the result.
 
 @history[#:added "7.9.0.1"]}
 
+@defproc[(hash-filter [ht hash?] [pred (-> any/c any/c boolean?)])
+         hash?]{
+
+Filters the @racket[hash?] @racket[ht] based on a predicate @racket[pred] applied to both its keys and values.
+This function constructs a new hash table that includes only those key-value pairs
+from the input @racket[ht] for which the predicate @racket[pred] returns true when
+applied simultaneously to the keys and values of @racket[ht]. The output hash table
+retains the mutability and the key comparison predicate (e.g., @racket[eqv?],
+@racket[equal-always?], @racket[equal?]) of the input hash table @racket[ht], ensuring that
+the structural and operational properties of the original hash are preserved
+in the output. This allows for nuanced filtering where the inclusion of a key-value pair
+depends on both the key and value meeting the criteria defined by @racket[pred].
+
+@examples[
+  #:eval the-eval
+  ;; Filtering key-value pairs where the key is less than 3 and value is even
+  (hash-filter (for/hash ([num '(1 2 3 4 5)]) (values num (* num 2)))
+               (λ (k v) (and (< k 3) (even? v))))
+
+  ;; Filtering key-value pairs from an empty hash table
+  (hash-filter (make-hash) (λ (k v) (< k 3)))
+
+  ;; Filtering with eq? hash table based on specific key-value conditions
+  (hash-filter (make-hasheq '([#f . "false"] [#t . "true"]))
+               (λ (k v) (and (eq? k #t) (string=? v "true"))))
+
+  ;; Filtering key-value pairs where the key is a list and the value is a symbol
+  (hash-filter (hash (list 1 2) 'pair (vector 3 4) 'vector)
+               (λ (k v) (and (list? k) (symbol? v))))
+
+  ;; Filtering key-value pairs of mixed types based on custom logic
+  (hash-filter (hash "one" 1 2 "two" "three" 3)
+               (λ (k v) (and (not (number? k)) (number? v) (> v 1))))
+]
+
+@history[#:added "8.12.0.6"]
+}
 
 @defproc[(hash-filter-values [ht hash?] [pred procedure?])
          hash?]{
--- a/pkgs/racket-doc/scribblings/reference/hashes.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/hashes.scrbl
@@ -1008,15 +1008,16 @@ one for the result.
 @defproc[(hash-filter [ht hash?] [pred (-> any/c any/c boolean?)])
          hash?]{
 
-Filters the @racket[hash?] @racket[ht] based on a predicate @racket[pred] applied to both its keys and values.
-This function constructs a new hash table that includes only those key-value pairs
-from the input @racket[ht] for which the predicate @racket[pred] returns true when
-applied simultaneously to the keys and values of @racket[ht]. The output hash table
-retains the mutability and the key comparison predicate (e.g., @racket[eqv?],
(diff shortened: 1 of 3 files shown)


pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -212,6 +212,13 @@ mappings can hide earlier mappings.
 
 See also @racket[make-custom-hash].
 
+@examples[
```
sorawee, line 215: The commit def72dc73b03887048e251729075dc168b486df5 is already in #4916. It looks like you accidentally include it in this PR?

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +999,34 @@ one for the result.
...
+
+Filters a hash table by applying a predicate to it's values.
+
+@examples[
+#:eval the-eval
+(hash-filter-values (for/hash ([num '(1 2 3 4 5)]) (values num (add1 num))) (λ (v) (< v 3)))
+(hash-filter-values (make-hash) (λ (v) (< v 3)))
+]
+
+@history[#:added "8.11.1"]
```
sorawee, line 1014: The current version should be `8.12.0.6`.

racket/collects/racket/hash.rkt
```diff
@@ -63,6 +63,12 @@
                     (combine/key k v (hash-ref hm k))))
         res)))
 
+(define (hash-filter-values ht pred)
+  (for/hash ([(k v) (in-hash ht)] #:when (pred v)) (values k v)))
```
sorawee, line 67: This will always construct an immutable hash with `equal?` as the key comparator, which doesn't seem right to me.

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +999,34 @@ one for the result.
 
 @history[#:added "7.9.0.1"]}
 
+
+@defproc[(hash-filter-values [ht hash?] [pred procedure?])
+         hash?]{
+
+Filters a hash table by applying a predicate to it's values.
```
sorawee, line 1006: typo: its values

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +999,34 @@ one for the result.
...
+
+Filters a hash table by applying a predicate to it's keys.
+
+@examples[
+#:eval the-eval
+(hash-filter-keys (for/hash ([num '(1 2 3 4 5)]) (values num 0)) (λ (k) (< k 3)))
+(hash-filter-keys (make-hash) (λ (k) (< k 3)))
+]
+
+@history[#:added "8.11.1"]
```
sorawee, line 1029: ditto

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +999,34 @@ one for the result.
...
+}
+
+
+@defproc[(hash-filter-keys [ht hash?] [pred procedure?])
+         hash?]{
+
+Filters a hash table by applying a predicate to it's keys.
```
sorawee, line 1021: ditto

racket/collects/racket/hash.rkt
```diff
@@ -84,4 +90,6 @@
                         #:combine/key
                         (-> any/c any/c any/c any/c)]
                        #:rest (listof hash?)
-                       (and/c hash? immutable?))])
+                       (and/c hash? immutable?))]
+ [hash-filter-values (-> (hash? procedure?) hash?)]
```
sorawee, line 94: This doesn't look like a valid contract to me. There's an extra pair of parens?

pkgs/racket-doc/scribblings/reference/hashes.scrbl
dr-neptune: yes, that other PR was accidentally done on the master branch and seems to have carried over. Will remove this change from this PR

racket/collects/racket/hash.rkt
dr-neptune: thank you, I'm not used to using contracts in my daily programming so I wasn't sure how to structure these. I've updated the file to remove the extra parens

racket/collects/racket/hash.rkt
dr-neptune: Thanks, I was aware of immutable/mutable but have never delved into different hash comparators. I've updated the function to try to maintain the right configuration and added new tests/examples

sorawee:
What do you think of just providing one function `hash-filter` (or `hash-filter/copy` to be consistent with `hash-map/copy`)? The procedure would consume both key and value. That way, you only need one function instead of two.

dr-neptune:
> What do you think of just providing one function `hash-filter` (or `hash-filter/copy` to be consistent with `hash-map/copy`)? The provided procedure would consume both key and value. That way, you only need one function instead of two.

While I agree it makes sense to have a function that consumes the key and value (`hash-filter`, `hash-filter/copy`), I do think it feels cleaner on the user side to still provide key-specific and value-specific keywords, i.e. providing `hash-filter`, `hash-filter-keys`, `hash-filter-values`. Putting the onus to handle both key and value in the provided lambda has lead me to default to the less ideal `(map fn (hash-keys ht))` pattern in the past.

That said, I am open to whatever makes the most sense given the community and API philosophy in Racket, as I am still very new to the idea of adding functionality here

pkgs/racket-test-core/tests/racket/hash.rktl
```diff
@@ -3,7 +3,7 @@
 
 (Section 'hash)
 
-(require racket/hash)
+(require racket/hash (only-in racket list? number? null? symbol? eq? eqv?))
```
sorawee, line 6: I don't think this is needed. They are all provided in the `base` environment already.

pkgs/racket-test-core/tests/racket/hash.rktl
```diff
@@ -123,6 +123,70 @@
...
+;; hash-filter-key and hash-filter-value:
+
+;; filtering by key with strings
+(test #hash(["one" . 1])
+      hash-filter-keys
+      #hash(["one" . 1] ["two" . 2] ["three" . 3])
+      (λ (k) (string=? k "one")))
+
+;; filtering by value with strings
+(test #hash(['carrot . "vegetable"])
```
sorawee, line 136: I think this test is wrong. Shouldn't it have `['apple . "fruit"]` and `['banana . "fruit"]` in the resulting hash?

Also, the key here is `'(quote carrot)`, not `'carrot`. I would recommend using `(hash key1 val1 key2 val2)` instead of `#hash(...)`.

sorawee: Sorry for misleading! I somehow thought it's filtering _out_.

racket/collects/racket/hash.rkt
```diff
@@ -63,6 +63,24 @@
                     (combine/key k v (hash-ref hm k))))
         res)))
 
+(define (hash-filter-values ht pred)
+  (let* ([mutable (not (immutable? ht))]
```
sorawee, line 67: Nit: use `define` instead of `let*`.

sorawee: Also, it's not use mutable vs immutable. There're also weak hash table and ephemeron hash table.

dr-neptune:
> > What do you think of just providing one function `hash-filter` (or `hash-filter/copy` to be consistent with `hash-map/copy`)? The provided procedure would consume both key and value. That way, you only need one function instead of two.
> 
> While I agree it makes sense to have a function that consumes the key and value (`hash-filter`, `hash-filter/copy`), I do think it feels cleaner on the user side to still provide key-specific and value-specific keywords, i.e. providing `hash-filter`, `hash-filter-keys`, `hash-filter-values`. Putting the onus to handle both key and value in the provided lambda has lead me to default to the less ideal `(map fn (hash-keys ht))` pattern in the past.
> 
> That said, I am open to whatever makes the most sense given the community and API philosophy in Racket, as I am still very new to the idea of adding functionality here

I've updated the PR to include `hash-filter` and rewritten `hash-filter-values/keys` to use `hash-filter`

sorawee:
@AlexKnauth That's my first thought too, but there's `hash-map/copy` that works in a similar way, so I don't think it's weird.

dr-neptune:
@sorawee @AlexKnauth should I try splitting this up into `hash-filter!` for mutable hash tables and `hash-filter` for immutable hash tables, or is this fine as is? Alternatively, should I rename the current `hash-filter` to `hash-filter/copy` since it makes a copy, and then build out the previously mentioned `hash-filter`, `hash-filter!`?

AlexKnauth:
`hash-filter` should not mutate the hash table it's given.

racket/collects/racket/hash.rkt
```diff
@@ -63,6 +63,23 @@
                     (combine/key k v (hash-ref hm k))))
         res)))
 
+(define (hash-filter ht pred)
+  (cond [(immutable? ht)
+         (for ([(k v) (in-hash ht)])
```
sorawee, line 68: Use `in-immutable-hash` for speed boost.

racket/collects/racket/hash.rkt
```diff
@@ -63,6 +63,23 @@
...
+           (when (not (pred k v))
+             (set! ht (hash-remove ht k))))]
+        [else
+         (for ([(k v) (in-hash ht)])
```
sorawee, line 72: As @AlexKnauth said, let's make a new hash table via `hash-copy-clear`, and then add stuff to it, so that we don't mutate the original hash table.

racket/collects/racket/hash.rkt
```diff
@@ -1,5 +1,5 @@
 #lang racket/base
-(require racket/contract/base)
+(require racket/contract/base racket/match)
```
sorawee, line 2: If there's no need for `match` in this file anymore, I would revert this line.

racket/collects/racket/hash.rkt
```diff
@@ -64,15 +66,15 @@
...
-  (cond [(immutable? ht)
-         (for ([(k v) (in-hash ht)])
-           (when (not (pred k v))
-             (set! ht (hash-remove ht k))))]
-        [else
-         (for ([(k v) (in-hash ht)])
-           (when (not (pred k v))
-             (hash-remove! ht k)))])
-  ht)
+  (define fresh-ht (hash-copy-clear ht))
```
sorawee, line 69: I think the change that @AlexKnauth had in mind is:

(*Caution: code not tested*)

```
(define (hash-filter ht pred)
  (cond
    [(immutable? ht)
     (for/fold ([ht ht]) ([(k v) (in-immutable-hash ht)])
       (cond
         [(pred k v) ht]
         [else (hash-remove ht k)]))]
    [else
     (define new-ht (hash-copy-clear ht))
     (for ([(k v) (in-hash ht)] #:when (pred k v))
       (hash-set! new-ht k v))
     new-ht]))
```

In particular:
- Avoid `set!`. Use `for/fold` to communicate to the next iteration, rather than via mutation.
- `in-immutable-hash` must appear under `for` syntactically for the optimization to kick in, so you can't pass it as an argument like that.
- Definitely avoid stuff from `racket/function`, which somewhat increases the start up time (IIRC).

dr-neptune: Thanks, will update later tonight or tomorrow morning. Lots to learn! I appreciate you two giving all this feedback

AlexKnauth: If `#:when` works for one, it should work for both. Or maybe `#:unless`, I'm not sure.

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +986,115 @@ one for the result.
...
+  ;; Filtering key-value pairs where the key is a list and the value is a symbol
+  (hash-filter (hash (list 1 2) 'pair (vector 3 4) 'vector)
+               (λ (k v) (and (list? k) (symbol? v))))
+
+  ;; Filtering key-value pairs of mixed types based on custom logic
+  (hash-filter (hash "one" 1 2 "two" "three" 3)
+               (λ (k v) (and (not (number? k)) (number? v) (> v 1))))
+]
+
+@history[#:added "8.12.0.6"]
```
sorawee, line 1024: Sorry, just noticed that we are now at 8.12.0.9. Could you make an update?

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +986,115 @@ one for the result.
...
+
+  ;; Filtering keys of mixed types: numbers and strings
+  (hash-filter-keys (hash "one" 1 2 "two" "three" 3) (lambda (k) (number? k)))
+
+  ;; Filtering keys that are symbols
+  (hash-filter-keys (hash 'apple "fruit" 'carrot "vegetable" "banana" "fruit")
+                    (lambda (k) (symbol? k)))
+]
+
+@history[#:added "8.12.0.6"]
```
sorawee, line 1059: Ditto.

racket/collects/racket/hash.rkt
```diff
@@ -63,6 +63,24 @@
                     (combine/key k v (hash-ref hm k))))
         res)))
 
+(define (hash-filter ht pred)
+  (cond
+    [(immutable? ht)
+     (for/fold ([ht ht]) ([(k v) (in-immutable-hash ht)]
+                          #:when (not (pred k v)))
```
sorawee, line 70: There's `#:unless`, FWIW. But maybe you want to keep `#:when` to be consistent with `#:when` below, which is fine too.

sorawee (approved):

racket/collects/racket/hash.rkt
dr-neptune: sure thing, changed to unless

dr-neptune:
What is needed to move this forward?

sorawee:
I cannot merge it due to the merge conflicts. Can you pull the latest `master`, rebase your changes on top of it, squash them to one commit, and then force push?

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +986,115 @@ one for the result.
...
+applied simultaneously to the keys and values of @racket[ht]. The output hash table
+retains the mutability and the key comparison predicate (e.g., @racket[eqv?],
+@racket[equal-always?], @racket[equal?]) of the input hash table @racket[ht], ensuring that
+the structural and operational properties of the original hash are preserved
+in the output. This allows for nuanced filtering where the inclusion of a key-value pair
+depends on both the key and value meeting the criteria defined by @racket[pred].
+
+@examples[
+  #:eval the-eval
+  ;; Filtering key-value pairs where the key is less than 3 and value is even
```
sorawee, line 1004: Oh, these comments won't be rendered, btw. You need to use `code:comment`. Not sure how much you care about that.

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +986,115 @@ one for the result.
...
+         hash?]{
+
+Filters the @racket[hash?] @racket[ht] based on a predicate @racket[pred] applied to both its keys and values.
+This function constructs a new hash table that includes only those key-value pairs
+from the input @racket[ht] for which the predicate @racket[pred] returns true when
+applied simultaneously to the keys and values of @racket[ht]. The output hash table
+retains the mutability and the key comparison predicate (e.g., @racket[eqv?],
+@racket[equal-always?], @racket[equal?]) of the input hash table @racket[ht], ensuring that
+the structural and operational properties of the original hash are preserved
+in the output. This allows for nuanced filtering where the inclusion of a key-value pair
```
sorawee, line 999: Nit: I would remove this "This allows for nuanced ...". The reference manual usually just describes the functionality. It doesn't need to motivate why things should be used. That belongs to the guide.

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -986,4 +986,115 @@ one for the result.
...
+
+@defproc[(hash-filter-values [ht hash?] [pred procedure?])
+         hash?]{
+
+Filters the @racket[hash?] @racket[ht] based on a predicate @racket[pred] applied to its values.
+This function returns a new hash table containing only the key-value pairs for which
+the predicate @racket[pred] returns true when applied to the values of @racket[ht].
+The resulting hash table retains the mutability and the key comparison predicate
+(e.g., @racket[eq?], @racket[eqv?], @racket[equal-always?], @racket[equal?]) of the input
+hash table @racket[ht]. This ensures that the characteristics of the input hash are preserved
```
sorawee, line 1071: Nit (same as above): I would remove "This ensures that the characteristics of the input hash are preserved in the output, allowing for a seamless transition between different types of hash tables."

pkgs/racket-doc/scribblings/reference/hashes.scrbl
```diff
@@ -214,16 +214,6 @@ mappings can hide earlier mappings.
...
-@examples[
-#:eval the-eval
-(make-hash)
-(make-hash '([0 . 1] [42 . "meaning of life"] [2 . 3]))
-(make-hash '([0 . 1] [1 . 2] [0 . 3]))
-(make-hash (list (cons 0 1) (cons 'apple 'orange) (cons #t #f)))
-(make-hash '((0 1) (1 2) (2 3)))
-(make-hash (list (cons + -)))
-]
-
```
AlexKnauth, line 226: Why delete these examples for `make-hash`, and also `hash-map` below?

sorawee: I f-ed it up after rebasing to resolve conflicts. Sorry. Redoing it.

dr-neptune:
Hi @sorawee @AlexKnauth , is there anything that still needs to be done here?

sorawee:
Sorry, looks good to me. Merged.

dr-neptune:
Thanks so much!

mflatt:
@sorawee @dr-neptune — ping

There's always the possibility of backward incompatibility when an export is added to a module. Our general rule is to add an export only when it does not create a conflict with any package registered at pkgs.racket-lang.org. (This is a compromise between a completely consistent approach to backward compatibility and practical evolution of APIs, since code not registered as a package can still break.) Usually, when there's a conflict, we just don't make the addition to the original library — which may mean noyt having the binding at all, or it may mean an extra library or submodule. Sometimes, the remedy is to get a helpful package author to change code. In any case, this needs to be resolved, since we're trying to get the v8.13 release out.

dr-neptune:
My apologies, I didn't see the previous ping by @samth . I'm not super familiar with this process. My first guess based on the error message is that the `Odysseus` package has some version of `hash-filter`. Would it make sense to try to supersede it with `hash-filter` in base `racket/hash`? `Odysseus` looks to be a general purpose helper library:

https://pkgs.racket-lang.org/package/odysseus

My thought is that filtering a hash makes sense in a std lib as opposed to a loaded external lib.

Does this make sense to you @mflatt @sorawee @prozion ?

raco setup:   /home/root/user/.local/share/racket/snapshot/pkgs/odysseus/odysseus/optimize.rkt:7:9: module: identifier already required
raco setup:     at: hash-filter
raco setup:     in: "hash.rkt"
raco setup:     also provided by: racket/hash

samth:
The best solution is for https://github.com/prozion/odysseus/pull/2 to get merged (thanks @AlexKnauth). If @prozion doesn't respond about that then maybe we need to change the name or revert the addition.

sorawee:
By April 16, if the odysseus PR is not merged, I think we should just revert this PR.

sorawee:
Let's revert the PR for now.

PR #4918, opened 2024-01-30
Test m1 runners

samth:


pkgs/racket-benchmarks/tests/racket/benchmarks/treelist/bm.rkt
```diff
@@ -0,0 +1,298 @@
+#lang racket/base
```
sorawee, line 1: Accidentally included?

samth: Yeah as you can see the branch includes #4913

samth:
Just as a note, this PR took about 13 minutes for the Mac test, whereas #4913 took 28 minutes. So it looks like the M1 runners are quite fast.

samth:
Note that this generates a bunch of (new?) warnings about old C code in zlib:

```

cd racket/src/build/cs/c/ChezScheme/pb/zlib && /Applications/Xcode_15.0.1.app/Contents/Developer/usr/bin/make static
gcc -g -O2 -Wall -DOS_X -DOS_X -DHAVE_HIDDEN -I. -c -o example.o test/example.c
test/example.c:88:6: warning: a function definition without a prototype is deprecated in all versions of C and is not supported in C2x [-Wdeprecated-non-prototype]
void test_compress(compr, comprLen, uncompr, uncomprLen)
     ^
```

sorawee:
Should this be reverted? It generates various errors:

https://github.com/racket/racket/actions/runs/7746169161/job/21123775895#step:6:278

samth:
Maybe CGC builds on aarch64 need `--disable-future` explicitly? Probably @mflatt knows.

sorawee:
Also, this PR _switched_ from macos-latest to macos-14 (and not simply adding macos-14). Is this intentional? That is, do we want to drop testing Mac with Intel?

mflatt:
An explicit `--disbale-futures` isn't needed, but the CI rule uses an explicit `--enable-futures`.

Issue #4922, opened 2024-02-04
Treelist chaperone/contract problems with functional update


LiberalArtist:
(I found this while thinking further about https://github.com/racket/racket/pull/4913#discussion_r1472969940.)

Consider the following program:
```racket
#lang racket
(require racket/treelist)
(define/contract lazy-natural-treelist
  (treelist/c natural? #:lazy? #t #:flat? #f)
  (treelist 1 2 3))
(treelist-add lazy-natural-treelist "abc")
```

With the 8.12.0.7-2024-02-03-901a31301a snapshot, the implicit printing produces the following error:

```
2024-02-03-901a31301a/collects/racket/contract/private/blame.rkt:346:0: lazy-natural-treelist: broke its own contract
  promised: natural?
  produced: "abc"
  in: the elements of
      (treelist/c natural?)
  contract from: 
      (definition lazy-natural-treelist)
  blaming: (definition lazy-natural-treelist)
   (assuming the contract is correct)
  at: 3-unsaved-editor:3:17
```

First of all, even if this program ought to produce an error, it seems like the blame ought to lie with the party that added the prohibited value.

More broadly, this example illustrates some of what I found surprising about the design of `chaperone-treelist`.

I typically think about chaperones, especially in the context of contracts, as enabling interposition on operations that provide a "communication channel" across a "boundary". I don't think of immutable values with functional update as providing such a communication channel. While a library may have specific requirements for the immutable values I supply to its functions, I expect to be able to use the values it returns to me freely within my own code. For example, there were a variant of the `xml` library that used a treelist-based `xexpr/c`, I would find it very surprising if I couldn't take parts of treelists read from XML and `treelist-add`/`treelist-cons` non-`xexpr/c` values to them.

[...]

PR #4928, opened 2024-02-06
change `chaperone-treelist` to support state-passing and more operations

mflatt:
Thread a "state" value through interposition procedures for `chaperone-treelist` and family. On each operation that produces an updated treelist, the updated state is associated with the chaperoned result, so that interposition procedures can track the evolution of the wrapped treelist as needed (with no dependence on mutable state).

Previously, operations that remove elements for a treelist had no interposition procedures, but new ones have been added so that they get the opportunity to update the state.

This is a backward-incomptible change within the development sequence, but on the grounds that treelists are new and experimental, anyway.

### DIFF
--- a/pkgs/base/info.rkt
+++ b/pkgs/base/info.rkt
@@ -14,7 +14,7 @@
 
 ;; In the Racket source repo, this version should change exactly when
 ;; "racket_version.h" changes:
-(define version "8.12.0.10")
+(define version "8.12.0.11")
 
 (define deps `("racket-lib"
                ["racket" #:version ,version]))
--- a/pkgs/racket-doc/scribblings/reference/treelists.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/treelists.scrbl
@@ -62,6 +62,20 @@ This operation takes @math{O(N log N)} time to construct a treelist of
 (treelist 1 "a" 'apple)
 ]}
 
+@defproc[(make-treelist [size exact-nonnegative-integer?] [v any/c]) treelist?]{
+
+ Returns a @tech{treelist} with size @racket[size], where
+ every element is @racket[v].
+ This operation takes @math{O(log N)} time to construct a
+ treelist of @math{N} elements.
+
+ @examples[
+ #:eval the-eval
+ (make-treelist 0 'pear)
+ (make-treelist 3 'pear)
+ ]
+
+@history[#:added "8.12.0.11"]}
 
 @deftogether[(
 @defproc[(treelist-empty? [tl treelist?]) boolean?]
--- a/pkgs/racket-doc/scribblings/reference/treelists.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/treelists.scrbl
@@ -182,14 +182,9 @@ Produces a treelist like @racket[tl], except that the element at
 @defproc[(treelist-append [tl treelist?] ...) treelist?]{
 
 Appends the elements of the given @racket[tl]s into a single
-@tech{treelist}. If @math{M} non-chaperoned treelists are given and
-the resulting treelist's length is @math{N}, then appending takes
-@math{O(M log N)} time.
-
-If any other than the first @racket[tl] is chaperoned via
-@racket[chaperone-treelist] with a procedure for its @racket[_ref-proc],
-the time to produce the result is @math{O(N)} for @math{N} elements
-added to the first @racket[tl].
+@tech{treelist}. If @math{M} treelists are given and the resulting
+treelist's length is @math{N}, then appending takes @math{O(M log N)}
+time.
 
 @examples[
 #:eval the-eval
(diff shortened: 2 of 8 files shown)


mflatt:
This in consultation with Robby and meant to address  #4922, but `treelist/c` here is changed only to use the new API.

LiberalArtist:
I see how this change enables a lazy `treelist/c`.

Unless I'm missing something, though, it seems like this still doesn't support cooperating code unwrapping a chaperoned treelist via a property, nor accessing the new "state" value. (Well, maybe you could do something tricky like abort to a particular prompt tag if present.)

This change also seems to go even further down the path of providing arbitrary interposition procedures for functional update operations. Maybe that's ok, if it works in practice, and `chaperone-treelist` ends up being a low-level function that must be used with care to avoid surprising behavior as in https://github.com/racket/racket/issues/4922. But it still seems different to me than the kinds of interposition points existing `chaperone-*` procedures expose.

rfindler:
Hi @LiberalArtist : is your goal to have a the chaperone layer fully implement the communication-channel/boundary aspects of contracts (and then contracts take it from there to give a fuller checking system)? If so, then I think that's probably close to what we had in mind back when we designed the original contract/chaperone divide, but we had something slightly more nuanced. We wanted to have a construct that would allow interposition and would serve as a building block for chaperones but would not give away more power over the primitive values than was already there. So there was this extra goal of not allowing new behaviors (e.g., disallowing impersonators on immutable values is part of that) plus the goal wasn't that chaperones would **be** the boundary part, but instead that it would be possible to implement the boundary parts using the chaperone layer (without having to do crazily inefficient stuff). So in that sense, this PR represents an experiment in that direction. As you pointed out, the lazy immutable treelist contracts do not do the boundary parts correctly and hopefully the little extra state piece will allow the contract to implement the boundary parts. (I'm sorry I haven't gotten to actually trying this out yet but I hope to get some other, pressing things done and then spend some time doing that.) [The paper](https://users.cs.northwestern.edu/~robby/pubs/papers/oopsla2012-stff.pdf) tries to explain this and may do a better job than I'm doing here.

LiberalArtist:
I think what I'm trying to pick at is where to draw the line for "reasonable" interposition on treelists. Or, from a slightly different point of view, what kinds of use cases are treelist chaperones intended to support?

I take one of the lessons of chaperones/impersonators to be that there is a trade-off. On the one hand, we want to allow enough interposition to do interesting things. (In particular, I definitely agree that the chaperone layer should enable implementing contracts as a library, rather than building in too many details of contracts at the chaperone level.) On the other hand, we want to constrain what interposition is allowed to avoid inordinate costs to reasoning, performance, and/or reasoning about performance.

I'm concerned that both the status quo and the current state of this PR allow interposition in some ways that I, at least, find surprising. Secondarily, this PR doesn't seem to enable some useful patterns like allowing cooperating code to unwrap chaperones via properties.

Maybe it will help to try to be a bit more concrete. Currently, chaperones change the algorithmic complexity of `treelist-append`:

> If `M` non-chaperoned treelists are given and the resulting treelist’s length is `N`, then appending takes `O(M log N)` time. If any other than the first `tl` is chaperoned via `chaperone-treelist with a procedure for its `ref-proc`, the time to produce the result of `O(N)` for `N` elements added to the first `tl`.

This is because [the implementation](https://github.com/racket/racket/blob/a0424377ca71775fbd23761fdb36b86ed9423821/racket/collects/racket/treelist.rkt#L1329-L1352) allows a non-`#f` `ref-proc` for a right-hand treelist to interpose on the each element being extracted and added to the left-hand treelist.

I think it might be possible to avoid the slowdown by not allowing `ref-proc` to interpose at that point. Instead, both the left and the right treelist could contribute a layer of re-chaperoning (rather than only the left treelist), with updates to properties and/or "state". Here's a sketch of how that might work for the lazy contract use case:

```racket
#lang racket
(require racket/treelist)
(define/contract left
  (treelist/c number? #:flat? #f #:lazy? #t)
  (treelist 1 'wrong 3))
(define/contract right
  (treelist/c symbol? #:flat? #f #:lazy? #t)
  (treelist 'a 'b "oops"))
(define both
  (treelist-append left right))
(treelist-ref both 0) ; -> 1
(treelist-ref both 3) ; -> 'a
(treelist-ref both 1) ; error: 'wrong
(treelist-ref both 5) ; error: oops
```

Even aside from performance, `treelist-ref` seems to me more like a boundary than `treelist-append` does, and thus a better place to get the error. I hope such a change might also avoid the current API's asymetric treatment of chaperoned treelists: I understand (more or less, at least) how it makes sense for the implementation of `treelist-append` to treat the leftmost RRB tree as the base and extend it with the others, but that seems like an implementation detail that ideally users wouldn't have to think about (if possible with reasonable efficiency etc.).

Providing a mechanism to update property values and/or "state" in the re-chaperoning of a treelist result also seems like it would address my secondary point about uses of properties for cooperating code that don't seem to be supported currently.

I don't have a great idea for what an API to interpose on the re-chaperoning of result treelists should look like. This is one of the things that made me start looking for precedents. I was surprised at how few examples I found of chaperones for interposing on functional update operations. Since Racketeers are generally enthusiastic about immutable datastructures with functional update, I think it's particularly worth thinking through `chaperone-treelist` to provide a good example for future libraries to imitate.

Of the functions documented in [the "Impersonators and Chaperones" reference section](https://docs.racket-lang.org/reference/chaperones.html), only `chaperone-hash` provides a direct way to interpose on functional update, specifically on `hash-ref`, `hash-remove`, and `hash-clear`. However, as I started to write in https://github.com/racket/racket/issues/4922, `chaperone-hash` seems provides the same API for immutable hashes as mutable ones, which doesn't seem to address these issues. In particular, consider this issue from https://github.com/racket/racket/issues/4922: 

> Finally (and somewhat tangentially), [re](https://github.com/racket/racket/pull/4913#discussion_r1472969940):
>
> > Yes, using an extra impersonator property to allow unwrapping is the usual way to go. The mutable-treelist implementation uses that trick itself, but it piggy-backs on the table of chaperone procedures that already uses a property.
>
> I don't think this technique quite works with the current `chaperone-treelist`. As the following program illustrates, currently, functionally updating a chaperoned treelist drops impersonator properties from the result. For unwrapping via a property to work, the result would need not only to have the property, but to map it to the non-chaperoned version of the resulting treelist.

This PR changes `chaperone-treelist` to match the behavior of `chaperone-hash`&emdash;i.e., in the language of this PR, the result "is chaperoned with the same procedures and properties as" the original, rather than dropping all properties&mdash;but neither this PR nor `chaperone-hash` seem to enable the unwrapping pattern @mflatt describes, because functional update operations don't provide an opportunity to update property values on the result. Here's an example of that limitation using `chaperone-hash`:

```racket
#lang racket
(define-values [i-p:orig has-orig? get-orig]
  (make-impersonator-property 'orig))
(define h0
  #hash([1 . a]
        [2 . b]
        [3 . c]))
(define h0+chap
  (chaperone-hash
   h0
   (λ (h k) ; ref-proc
     (values k (λ (h k v)
                 v)))
   (λ (h k v) ; set-proc
     (values k v))
   (λ (h k) ; remove-proc
     k)
   (λ (h k) ; key-proc
     k)
   i-p:orig h0))
(define h1
  (hash-set h0+chap 4 'd))
(impersonator? h1) ;-> #t
(has-orig? h1) ;-> #t
(equal? h1 (get-orig h1)) ;-> #f !!
(hash-has-key? (get-orig h1) 4) ;-> #f !!
```

Some more observations from [the "Impersonators and Chaperones" reference section](https://docs.racket-lang.org/reference/chaperones.html):

  - `chaperone-struct-type`'s ability to interpose on constructors and guards, as well as `chaperone-struct`'s ability to interpose on `struct-info`, provide some very indirect support for interposing on functional update
  - `chaperone-procedure*` and `chaperone-vector*` suggest a need to think about overriding of properties by further impersonators, which I definitely haven't thought about
  - `chaperone-procedure`'s optional additional return values of a `result-rapper-proc` and `'mark key value` for continuation marks could be potential models for interposing on results

The other precedents I found for functional data structures in the standard library are those like `gen:stream`, `gen:dict`, and `gen:set` that support chaperones via [`chaperone-generics`](https://docs.racket-lang.org/reference/struct-generics.html#%28form._%28%28lib._racket%2Fgeneric..rkt%29._chaperone-generics%29%29). The `chaperone-generics` API accepts properties to attach as "a list with an even number of elements, where the first element of the list is an impersonator property, the second element is its value, and so on" (i.e. a [property list](https://www.cs.cmu.edu/Groups/AI/html/cltl/clm/node108.html) in the Lisp sense, like the tail of arguments `chaperone-struct` and similar would be `apply`ed to). Actually interposing on generic methods is not handled by `chaperone-generics` directly: users are given the a procedure and can use `chaperone-procedure` or some derived operation explicitly. Here's an sketch of how `chaperone-generics` allows re-chaperoning of the result of a functional update operation:

```racket
#lang racket
(require racket/generic)
(struct my-dict (d) ; https://github.com/racket/racket/issues/4934
  #:transparent
  #:methods gen:dict
  [(define/generic ref dict-ref)
   (define dict-ref
     (case-lambda
       [(this k)
        (ref (my-dict-d this) k)]
       [(this k fail)
        (ref (my-dict-d this) k fail)]))
   (define/generic set dict-set)
   (define (dict-set this k v)
     (my-dict (set (my-dict-d this) k v)))])
(define-values [i-p:sym-dict known-sym-dict? unwrap-sym-dict]
  (make-impersonator-property 'sym-dict))
(define (chaperone-sym-dict orig)
  (chaperone-generics gen:dict orig
    [dict-set (λ (super)
                (chaperone-procedure
                 super
                 (λ (this k v)
                   (if (symbol? k)
                       (values chaperone-sym-dict this k v)
                       (values this k v)))))]
    #:properties (list i-p:sym-dict orig)))
(define d0
  (my-dict #hash([a . 1])))
(define d0+chap
  (chaperone-sym-dict d0))
(define d1
  (dict-set d0+chap 'b 2))
(known-sym-dict? d1) ; -> #t
(equal? d1 (unwrap-sym-dict d1)) ; -> #t
(dict-ref (unwrap-sym-dict d1) 'b) ; -> 2
(define d2
  (dict-set d1 "not a symbol" 0))
(known-sym-dict? d2) ; -> #f
```

Hopefully some of this is useful!

mflatt:
@LiberalArtist — Thanks for your analysis!

I've pushed several changes in the latest attempt:

 * The current state can be extracted from a chaperoned treelist using a key that is supplied when the treelist is chaperoned as `#:state-key`.
 * In addition to `#:append` for chaperoning, there's `#:prepend` and `#:append2`, which together form the API for merging chaperones. The `#:append2` case is optional, but it can implement a binary operation somewhat along the same lines as `prop:equal+hash`. There's no `treelist-ref` involved with a chaperoned `treelist-append`.
 * Chaperones and impersonators for mutable treelists have been simplified, because a chaperone "state" separate from the mutable state doesn't turn out to make any sense.

mflatt:
I'm planning to merge this. It's not clear that this set of changes is right, yet, so I wouldn't say that these changes are the last word on treelist chaperones. But it seems enough of a good direction to replace the current API.

LiberalArtist:
I haven't had time to experiment with the changes yet, hence my not commenting earlier, but I agree they seem like a good direction!

PR #4929, opened 2024-02-07
expander: add `syntax-local-make-definition-context-introducer`

usaoc:
##### Checklist
- [x] Feature
- [x] tests included
- [x] documentation

### Description of change
This function makes an introducer that encapsulates a scope that should be pruned from `quote-syntax` forms, much like an {inside,outside}-edge scope.  This can be useful for implementing forms like Rhombus `let` (which is indeed the motivation; see racket/rhombus-prototype#471).


usaoc:
To clean up potential confusion seeing the mentioning Rhombus issue: this PR does *not* intend to introduce low-level mechanism for the user to replicate what definition contexts already do.  Rather, it’s intended for situation where you want to create scopes that are should be pruned for `quote-syntax`, so it’s limited in scope.

mflatt:
Since the `as-inside-edge?` argument affects only the debugging name of the scope, would it make sense to generalize that argument to be any symbol that isn't `'macro`?

Specifically, I have in mind that this function is useful to add use-site scopes to Rhombus's enforestation-level expansion. The name doesn't matter, but it can help with debugging if things are distinct. Also, seeing a debugging category for Rhombus `let` by be helpful.

Changing to a symbol would also help clarify that debugging is the only effect of the choice.

mflatt:
As It may also be that `syntax-local-apply-transformer` is the better way to handle use-site scopes in Rhombus, as @michaelballantyne points out in racket/rhombus-prototype#474. But if we're going to expose the more primitive building block directly, then the generalization away from `as-inside-edge?`.

usaoc:
I incorporated the suggestion and made some clarification to the purpose of this function.  I’m not sure how much we still want this change, though.

mflatt:
Thanks for the updates!

Making sure I'm not confused: This is still a useful shortcut for implementing something like `let` in Rhombus, right? Even with Rhombus using `syntax-local-apply-transformer`, the scope for each `let` is separate from the ones managed directly by `syntax-local-apply-transformer`.

usaoc:
Yes, if we keep the current implementation strategy for `let`, that is, managing a single scope for each `let`.

PR #4939, opened 2024-02-25
CS: improve hash code for immutable hash tables

### LINKED ISSUE #4902: Very high chance of collision for hash datatype
8.12 [cs]

**What program did you run?**

```
(set-count
 (set-union
  (for/set ([i (in-range 100000)])
    (equal-hash-code (hash i #t)))
  (for/set ([i (in-range 100000)])
    (equal-hash-code (hash (number->string i) #t)))
  (for/set ([i (in-range 1000)])
    (equal-hash-code (hash (integer->char i) #t)))))
```

The set size should be relatively high (that is, there should be few collisions).

Instead, the result is 16. That is a hash table with 1 association will almost always collide.

N/A

N/A

mflatt:
The old hash-code function for immutable hash tables tried using the mask in HAMT nodes as a summary and substitute for key hashes. That's not good enough, though, as illustrated by the case of making different hash tables all with 1 key (so there are only 16 possible masks in the sole node).

Closes #4902

### DIFF
--- a/pkgs/racket-test-core/tests/racket/hash.rktl
+++ b/pkgs/racket-test-core/tests/racket/hash.rktl
@@ -869,4 +869,25 @@
 
 ;; ----------------------------------------
 
+(let* ([ht
+        (for/fold ([ht #hash()]) ([i (in-range 100000)])
+          (hash-set ht (equal-hash-code (hash i #t)) #t))]
+       [ht
+        (for/fold ([ht ht]) ([i (in-range 100000)])
+          (hash-set ht (equal-hash-code (hash (number->string i) #t)) #t))]
+       [ht
+        (for/fold ([ht ht]) ([i (in-range 100000)])
+          (hash-set ht (equal-hash-code (list i)) #t))]
+       [ht
+        (for/fold ([ht ht]) ([i (in-range 100000)])
+          (hash-set ht (equal-hash-code (vector i)) #t))]
+       [ht
+        (for/fold ([ht ht]) ([i (in-range 1000)])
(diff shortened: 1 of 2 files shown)


racket/src/cs/rumble/hamt-stencil.ss
```diff
@@ -1148,19 +1150,28 @@
                      [(bnode? c)
                       (bnode-hash-code c hash hc)]
                      [else
-                      ;; Hash code needs to be order-independent, so
-                      ;; collision nodes are a problem; simplify by just
-                      ;; using the hash code and hope that collisions are
-                      ;; rare.
+                      ;; Hash code needs to be order-independent with
+                      ;; repsect to collisions; simplify by just using
```
sorawee, line 1154: nit: respect

pkgs/racket-test-core/tests/racket/hash.rktl
```diff
@@ -869,4 +869,25 @@
...
+        (for/fold ([ht ht]) ([i (in-range 100000)])
+          (hash-set ht (equal-hash-code (list i)) #t))]
+       [ht
+        (for/fold ([ht ht]) ([i (in-range 100000)])
+          (hash-set ht (equal-hash-code (vector i)) #t))]
+       [ht
+        (for/fold ([ht ht]) ([i (in-range 1000)])
+          (hash-set ht (equal-hash-code (hash (integer->char i) #t)) #t))])
+  ;; allow some collisions, at worst 2 collisions per key; at the time of
+  ;; writing, we expect collisions with chars, but not other collsisions
```
sorawee, line 888: nit: collisions

PR #4943, opened 2024-02-26
Add `vector-extend`.

samth:
This provides access to optimized Chez implementations of vector extension.


racket/src/cs/rumble/vector.ss
```diff
@@ -5,7 +5,7 @@
     (unless (and (fixnum? n)
                  (fx< n 1000))
       (guard-large-allocation who 'vector n (foreign-sizeof 'void*)))
-    (#2%make-vector n v)]))
+    (#3%make-vector n v)]))
```
mflatt, line 8: A leftover change that you didn't mean to commit?

samth: I think I was unsure if this change was safe to make (because I was unable to determine what the difference between these was).

mflatt:
I wonder how much difference this makes. If the check to dispatch to `guard-large-allocation` is inlined, I'd expect the safe `make-vector` to be about as fast as an unsafe version. But maybe that check doesn't get inlined along with `#2%make-vector` or maybe it costs more than I think.

As far as I can tell, this would be the first unsafe primitive outside of `ffi/unsafe` to involve an arbitrary amount of allocation, so documenting an acceptable/safe argument might need some care.

racket/src/cs/rumble/vector.ss
mflatt: `#3%make-vector` or `make-vector` in this context is an unsafe `make-vector`. That would be bad, because a should-be-safe `(make-vector #f)` at the Racket level (as implemented by the enclosing function) would crash instead of raising an exception.

mflatt:
Ah, I see. At the Racket level, I think adding a new vector function as exported by `racket/vector` may be the right choice, so a suitable length check can be in that function. Maybe `vector-extend` that takes a vector, number of items to add, and (optionally) a value for the new slots?

samth:
Would `vector-extend` be implemented in rumble directly, or atop `unsafe-make-vector`?

sorawee:
This should bump a new version of `base`, so that `racket/data` can specify a minimal version needed.

samth:
I added a commit implementing it on top of `unsafe-make-vector`.

mflatt:
> Would vector-extend be implemented in rumble directly, or atop unsafe-make-vector?

> I added a commit implementing it on top of unsafe-make-vector.

My intent was to avoid having `unsafe-make-vector`, and in particular avoid having to specify a safe constraint on the vector length. If `extend-vector` is implemented at the rumble layer, then the implementation can use `guard-large-allocation` as appropriate.

The `vector-extend` and `vector*-extend` implemention here do not check the `extra-space` argument before passing it along to `unsafe-make-vector`, so that's a problem. Even to the degree that they use safe operations, they should check arguments to report errors as being from `vector-extend` or `vector*-extend`. But I'm suggesting that this implementation and checking would be in rumble (and BC), instead.

samth:
I've re-implemented this along the lines you suggest. But I implemented it to take a new size, rather than a size to add, since the new size is what my use of the operation has available.

pkgs/racket-benchmarks/tests/racket/benchmarks/treelist/bm.rkt
```diff
@@ -286,18 +286,25 @@
 
 (module+ main
   (measure 1000000
-           10)
+           10
+	   '(mut-treelist gvector))
```
mflatt, line 290: These changes were not intended to be included?

racket/src/bc/src/vector.c
```diff
@@ -1291,6 +1303,49 @@ static Scheme_Object *vector_star_append(int argc, Scheme_Object *argv[])
...
+    scheme_raise_out_of_memory(who, "creating vector of length %s",
+			       scheme_make_provided_string(argv[1], 1, NULL));
+  }
+
+  if (argc == 3)
+    fill = argv[2];
+  else
+    fill = scheme_make_integer(0);
+
+  new_vec = scheme_make_vector(new_len - SCHEME_VEC_SIZE(v), fill);
```
mflatt, line 1330: Is a check that `new_len > = SCHEME_VEC_SIZE(v)` needed, or is it somehow covered above?

racket/src/cs/schemified/schemify.scm
mflatt, line 1: Changes to `schemify.scm` not intended, right?

mflatt:
This looks like a good direction overall. Just tests and docs to go?

racket/src/cs/rumble/vector.ss
```diff
@@ -568,3 +568,45 @@
...
+    (vector-extend v new-size 0)]
+   [(v new-size fill)
+    (check who vector? v)
+    (check who exact-nonnegative-integer? new-size)
+    (let ([old-size (vector-length v)])
+      (unless (<= old-size new-size)
+	(raise-arguments-error who
+                               "new length is shorter than existing size"
+                               "new length" new-size 
+                               "existing size" old-size))
```
mflatt, line 584: I wonder about "length" versus "size" here. Maybe both should be described as "length"?

samth: Ok.

samth:
Yes, that seems right (and fixing the bugs the tests find).

samth:
Tests and docs in place (and passing on CS and BC).

mflatt:
@samth — I added a small commit, which you can squash after vetting.

racket/src/bc/src/vector.c
```diff
@@ -1312,7 +1312,7 @@ static Scheme_Object *do_vector_extend(const char *who, int chaperone_ok, int ar
   if (chaperone_ok && SCHEME_NP_CHAPERONEP(v))
     v = SCHEME_CHAPERONE_VAL(v);
   if (!SCHEME_VECTORP(v))
-    scheme_wrong_contract(who, "vector?", 0, argc, argv);
+    scheme_wrong_contract(who, chaperone_ok ? "vector?" : "(and/c vector? (not/c immutable?))", 0, argc, argv);
```
samth, line 1315: Did you mean impersonator here?

samth: Yeah I remember looking at that and being confused. I'll fix all of them.

samth:
I fixed all those error messages and squashed in your changes.

mflatt:
LGTM

PR #4944, opened 2024-02-28
docs: correct typos in “Treelists” section

usaoc:
##### Checklist
- [x] Bugfix
- [x] tests included
- [x] documentation

### Description of change
This corrects some typos in “Treelists” section, and fixes a bug in `mutable-treelist-sort!` that I noticed during the process.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/treelists.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/treelists.scrbl
@@ -14,7 +14,9 @@ way that supports many operations in @math{O(log N)} time: accessing
 an element of the list by index, adding to the front of the list,
 adding to the end of the list, removing an element by index, replacing
 an element by index, appending lists, dropping elements from the start
-or end of the list, and extracting a sublist. The base for the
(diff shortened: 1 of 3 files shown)


pkgs/racket-doc/scribblings/reference/treelists.scrbl
```diff
@@ -54,7 +54,7 @@ Returns a @tech{treelist} with @racket[v]s as its elements in order.
 
 This operation takes @math{O(N log N)} time to construct a treelist of
 @math{N} elements. Unless otherwise specified, other operations on a
-treelist of length @math{N} also take @math{O(log N)} time.
+treelist of length @math{N} take @math{O(log N)} time.
```
sorawee, line 57: It seems weird to document the running time of "other operations" in this `treelist` function. I think the introductory paragraph would work better.

mflatt: I moved it, since I had put this here in the first place.

mflatt:
Thanks for these corrections!

PR #4946, opened 2024-03-05
Skip let-values arity check in unsafe mode.

samth:
This simplifies a lot of generated code, so much that I wonder if @mflatt already thought of this and decided against it.


mflatt:
No objection. In most cases, I would expect the error hanlding to be removed by cp0. For example, even though the generated `.scm` files get smaller, there's very little change in the generated `racket.boot` file. This change might have a bigger effect on other cases, though, and maybe it's useful to save work along the way.

PR #4956, opened 2024-03-22
add `unsafe-impersonate-hash`

mflatt:
Unlike `impersonate-hash`, `unsafe-impersonate-hash` works on immutable hash tables. As a result, interposition procedures on hash table operations can implement a custom hashing strategy (e.g., by wrapping keys in another struct as they're added, and then unwrapping keys whenever they are extracted).

Racket already has `make-custom-hash`, but the result of `make-custom-hash` is not a hash table in the sense of `hash?`. So, the idea here is to enable custom hashing for a hash table that works with `hash-ref`, etc.

The new operation is unsafe, because impersonating an immutable data structure can be misleading: two immutable hash tables could be `equal?` but behave very differently. The new operation is intended for use only when the impersonated hash table is not itself visible.

The intent is *not* that Racket programs would use `unsafe-inpersonator-hash` directly, but that a custom-hashing library can be built on it. Adding `unsafe-impersonate-hash` is an especially small change to the Racket core to enable that.

Issues:

* As of the initial implementation, there's no way to make `hash-copy` create a table with the same hash function.
* Along similar lines, hash tables with custom hashing would get serialized as having one of the four primitive equality functions.
* Even more generally along those lines, the change to have more kinds of hash tables could break code that assumes there are only four possible equalities for hashing.

The first issue could be addressed by having `hash-copy` propagate wrapper functions to a new mutable hash table — in the same way that `hash-set` propagates to an immutable table. The other two issues seem like inherent limitations.


samth:
It seems like with a slightly larger change (for example, providing the constructor function rather than an existing table) this could be made safe. If that's right, it seems worth it to me.

mflatt:
Maybe the primitive could be unsafe, but a safe addition to `racket/hash` could use it in the intended way?

I was starting to think that this change is probably not worthwhile, since uses of `make-custom-hash` seem rare. But I see that `make-custom-hash` is used in the `datalog`, `python`, `redex`, and `plai` packages. Also, custom hashing is the essence of`free-identifier=?` and `bound-identifier=?` tables, which are useful in macros. So, maybe it's useful enough.

rmculpepper:
I'm unclear on the benefits, and I dislike the idea of changing `hash?`, which currently recognizes a collection of concrete datatypes, into something that occupies an ad hoc halfway point between concrete datatype and interface.

Instead of abusing the definition of a concrete datatype, I think we should improve the support for the corresponding interface (`dict?`), or if there is some obstacle to that (backwards compatibility, perhaps?), maybe we should introduce a new interface.

Why is the `dict` interface not a suitable solution to the underlying problem?

If the goal is just to let `hash-ref` etc work with other things, because the name is more attractive than `dict-ref`, or maybe the dispatch speed, then the core could add a `prop:hashlike` property and a `hashlike?` predicate, and the contract of `hash-ref`'s first argument could be expanded from `hash?` to `hashlike?`. Functions could gradually opt in to supporting hashlike arguments, which IMO would be better than gradually discovering the list of functions that are broken on nonstandard hashes.

mflatt:
Yes, the goal is to be able to use `hash-ref` with its current performance on the current representations of hash tables, but also supporting an extended representation. I don't know how to so easily get the same behavior and performance out of `dict-ref` with our current compilation framework.

I can see how `hashlike?` might be a good direction for expanding hash-table functionality in Racket, but I'm not yet convinced that a Racket extension in that direction is worthwhile. Meanwhile, I am convinced that a long-term experiment is worthwhile at the Rhombus level — and that experiment may inform an eventual choice for Racket. So, I'm not in favor of adding something like `hashlike?` now, since that would be a bigger change than is needed for the experiment.

mflatt:
Even just pushing the Rhombus experiment a little further, I find that I need a little more built-in support to work with `equal?` to make things fit together, so I've (reluctantly) added that.

LiberalArtist:
> Yes, the goal is to be able to use `hash-ref` with its current performance on the current representations of hash tables, but also supporting an extended representation. I don't know how to so easily get the same behavior and performance out of `dict-ref` with our current compilation framework.

Rather than the general `dict-ref`, maybe it would be viable to use the specialized procedures like `custom-hash-ref` from `racket/private/custom-hash` (or a tuned version: `custom-key` for example seems like a candidate for `#:authentic` and `#:sealed`). It strikes me that the core of `hash-ref` at the Rumble level is already generic over a few representations:
```racket
(define (hash-ref/none ht k)
  (cond
   [(mutable-hash? ht)
    (cond
     [(eq-mutable-hash? ht)
      •••]
     [else
      •••])]
   [(intmap? ht)
    •••]
   [(and (impersonator? ht)
         (authentic-hash? (impersonator-val ht)))
    (impersonate-hash-ref ht k)]
   [else
    (raise-argument-error 'hash-ref "hash?" 0 ht k)]))
```
I would hope adding one more branch to get `hash-or-custom-ref` wouldn't have a prohibitive impact on performance.

mflatt:
A rough measurement from running the program below on my machine is that an extra dispatch layer costs about 10% of the time of lookup for a small, `eq`-based table (i.e., worst-case scenario). That would be acceptable in trade for some benefits, but generalized hashing does not seem to me worth that cost (or the work of implementing the wrappers). The idea behind this PR is to add functionality in a easy way with minimal risk of increasing costs.

```
#lang racket/base

(define hteq (for/hasheq ([i (in-range 5)]) (values i i)))
(define ht (for/hash ([i (in-range 5)]) (values (format "~a" i) i)))
(define mhteq (hash-copy hteq))

(struct other ())

(define (x-hash-ref ht k)
  (if (other? ht)
      (list ht)
      (hash-ref ht k)))

(define N 10000000)

(time
 (for/fold ([v #f]) ([i (in-range N)])
   (hash-ref mhteq 1)))

(time
 (for/fold ([v #f]) ([i (in-range N)])
   (x-hash-ref mhteq 1)))

(time
 (for/fold ([v #f]) ([i (in-range N)])
   (hash-ref hteq 1)))

(time
 (for/fold ([v #f]) ([i (in-range N)])
   (x-hash-ref hteq 1)))

(time
 (for/fold ([v #f]) ([i (in-range N)])
   (hash-ref ht "1")))

(time
 (for/fold ([v #f]) ([i (in-range N)])
   (x-hash-ref ht "1")))
```

mflatt:
One more thought: a difference between `unsafe-impersonate-hash` in Racket for Rhombus and `$`-prefixed names in Chez Scheme for Racket is that the latter are not documented and have to be accessed using Chez Scheme's `$primitive` form (which is usually written with the shorthand `#%`). So, it's easier to pretend than the `$` names just don't exist. We could easily go that route with `unsafe-impersonate-hash` by not exporting it from `racket/unsafe/ops`, and Rhombus would access the `#%unsafe` module directly.

An undocumented, unexported binding seems like a worse option to me, but it has the advantage of committing even less to `unsafe-impersonate-hash` while an experiment proceeds in Rhombus.

LiberalArtist:
> We could easily go that route with `unsafe-impersonate-hash` by not exporting it from `racket/unsafe/ops`, and Rhombus would access the `#%unsafe` module directly.
> 
> An undocumented, unexported binding seems like a worse option to me, but it has the advantage of committing even less to `unsafe-impersonate-hash` while an experiment proceeds in Rhombus.

I agree that `#%unsafe` seems worse that `racket/unsafe/ops` as a *solution*, but it does seem especially low-impact as temporary way to enable an experiment. If the experiment works out on the Rhombus side, presumably it will need some story for Racket interop, and maybe more insight into the implications will have emerged by that point. (Including perhaps evidence that adding `unsafe-impersonate-hash` to `racket/unsafe/ops` would be low-impact in practice.)

One detail that struck me was that unsafely-impersonated hashes would satisfy not only `hash?` but also e.g. `hash-eq?`, which seems like it could be quite surprising.

> I was starting to think that this change is probably not worthwhile, since uses of `make-custom-hash` seem rare. But I see that `make-custom-hash` is used in the `datalog`, `python`, `redex`, and `plai` packages. Also, custom hashing is the essence of`free-identifier=?` and `bound-identifier=?` tables, which are useful in macros. So, maybe it's useful enough.

A scenario when I've used `define-custom-hash-types` (and `define-custom-set-types`, and similar: I think I ultimately opted to implement `gen:dict` on a wrapper struct myself) is for values that have some kind of unique identifier, like a primary key in a database: I've wanted the container to behave as a set of values and to be able to look up the value for a given key.

mflatt:
Ok, switched to export from `#%unsafe`, only.

PR #4957, opened 2024-03-23
ChezScheme: replace call to iconv procedure with ICONV macro

Bogdanp:
The availability of the iconv procedure varies with different build settings, so I believe the ICONV macro should be used here instead.


sorawee:
Is this better submitted to https://github.com/cisco/ChezScheme/?

Bogdanp:
Moved to [cisco/ChezScheme](https://github.com/cisco/ChezScheme/pull/818).

PR #4962, opened 2024-03-31
cs,build: add pbchunk object files to libracketcs.a

Bogdanp:
I noticed that linking against a pbchunk build of `libracketcs.a` fails due to missing symbols because the pbchunk object files are not included in the archive. This change addresses that problem.


mflatt:
Thanks!

PR #4966, opened 2024-04-05
feat: move rackunit/log to raco/testing

### LINKED ISSUE #4965: [Feature Request] rackunit/log should be moved to racket
`raco test` requires `rackunit` installation (it particularly needs `rackunit/log`). Non `rackunit` testing framework also needs to require `rackunit/log` to register testing results.

I think it would be better to move `rackunit/log`, which is actually very small and self-contained, to Racket itself, so that `rackunit` doesn't need to be installed for `raco test` / other testing framework to work.

sorawee:
Previously, `rackunit/log` is used by both `raco test` and various testing framework to log testing results. However, this organization doesn't make much sense, as it gives privileges to rackunit over other testing frameworks (technically `rackunit/log` lives in the `testing-util-lib` package, so it's not a rackunit package proper, but it's still very close)

This commit moves `rackunit/log` to `raco/testing`, which makes `raco test` "neutral". This also eliminates the dependency on `testing-util-lib`/`rackunit-lib`, for both `raco test` itself, and also other testing frameworks.

Must be merged along with racket/rackunit#174

Fixes #4965

### DIFF
--- a/pkgs/racket-doc/scribblings/raco/test.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/test.scrbl
@@ -4,8 +4,9 @@
           "common.rkt" 
           (for-label racket/runtime-path
                      racket/base
+                     racket/contract
                      launcher/launcher
-                     rackunit/log
+                     raco/testing
                      compiler/module-suffix
                      compiler/cm))
 
@@ -192,7 +193,7 @@ The @exec{raco test} command accepts several flags:
  @item{@DFlag{table} or @Flag{t}
        --- Print a summary table after all tests. If a test uses
        @racketmodname[rackunit], or if a test at least uses
-       @racket[test-log!] from @racketmodname[rackunit/log] to log
+       @racket[test-log!] from @racketmodname[raco/testing] to log
        successes and failures, the table reports test and failure
        counts based on the log.}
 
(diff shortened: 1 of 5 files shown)


samth:
I'm not sure exactly what the problem is here. Is it that the module name has `rackunit` in it? Or that the code is in the `racket/rackunit` repository? Or that it's not part of the `"base"` package?

sorawee:
All of them, yes.

It's just weird that if you want to implement a testing framework (say, https://docs.racket-lang.org/chk/), you need to require `rackunit/log` to integrate with `raco test`.

samth:
So would just renaming the library be enough? Why is it a problem for `chk` to depend on `testing-util-lib`?

I am reluctant to add new things to `base` unless there's a real benefit.

sorawee:
What about this? I'm not gonna add it to `base`, but instead to `compiler-lib`. Would that address your concern?

samth:
That's where raco test itself is, right? That seems like a good approach.

sorawee:
Correct. I will do that then.

sorawee:
@samth done, but let's hold off merging this until the release branch is created.

pkgs/racket-doc/scribblings/raco/test.scrbl
```diff
@@ -378,4 +379,32 @@ the test's output is prefixed with a
...
+ @racket[(current-error-port)]. If @racket[exit?] is true, then if there were
+ failures, calls @racket[(exit 1)].}
+
+@defboolparam[test-log-enabled? enabled? #:value #t]{
+ When set to @racket[#f], @racket[test-log!] is a no-op. This is useful to
+ dynamically disable certain tests whose failures are expected and shouldn't be
+ counted in the test log, such as when testing a custom check's failure
+ behavior.}
+
+@history[#:added "1.13"]
```
usaoc, line 410: I’ve just noticed this, but this history note should go under the `defmodule`, right?  It looks extremely out of place since `current-test-invocation-directory` is added, which comes with another history note.

benknoble: That seems reasonable, though I think other modules tend to put them at the end? (Maybe `current-test-invocation-directory`’s note should move?)

sorawee: I agree with @usaoc that this should be placed above, probably immediately before `test-log!`.

PR #4969, opened 2024-04-11
generalize contract-in to have the same features as contract-out

rfindler:
This is ready to be merged, IMO, but shouldn't be until the release branches are created.


pkgs/racket-doc/scribblings/reference/contracts.scrbl
```diff
@@ -2044,7 +2059,9 @@ the export.
           (eval:error (recip 1+2i))]
 
 @history[#:changed "7.3.0.3" @list{Added @racket[#:unprotected-submodule].}
-         #:changed "7.7.0.9" @list{Started ignoring @racket[ignored-id].}]
+         #:changed "7.7.0.9" @list{Started ignoring @racket[ignored-id].}
+         #:changed "8.12.0.13" @list{Added @racket[contract-in]}
```
sorawee, line 2063: I think the change to add the `8.12.0.13` entry should be added to `master` right now as a separate commit.

rfindler:
> Just to make sure I understand correctly:
> 
>     1. The addition of `contract-in` will be a part of 8.13.
> 
>     2. The further improvement will be a part of 8.14
> 
>     3. The further improvement is backward compatible relative to the addition of `contract-in`.
> 
> 
> Are these correct?

Yes, these are all correct.

> (If these are not the case, I think it would be more reasonable to defer the addition of `contract-in` to 8.14, or simply merge them all in 8.13.)

I considered doing them all in 8.13, but since the change is to big and affects `contract-out`, I feel a bit nervous doing it now.

The change that adds the simpler form of `contract-in` (that's already on master) doesn't require changes to `contract-out` and has a lot fewer moving pieces so I'm pretty confident in that part.

mfelleisen:
+1

I support Robby’s decision (and accept his rationale).

PR #4971, opened 2024-04-14
zuo: fix some compiler warnings

benknoble:
I figured a patch gave a better starting point for discussion, but I admit I'm not confident in some of these changes and I'm opening this mostly to generate conversation.

For the unused parameters, I considered introducing
```
#define UNUSED(x) (void)x
```
but didn't know if that fit the project style. I certainly didn't think we want to deal with compiler-specific attributes like `__unused`.

---

I often compile with a CFLAGS that looks like

-Wall -Wextra -Wstrict-prototypes -Wold-style-definition -Wshadow -Wpointer-arith -Wcast-qual -pedantic -O2 -std=c11

(I actually typically include -Wmissing-prototypes but Zuo produces many warnings from that which are probably harmless.) These changes cover the generated warnings (mostly shadowed variables, unused function arguments, and missing void parameter lists). Interesting cases are in the commit message.

./zuo build.zuo check passes after all these changes are compiled.

Here's the remaining warning:
```
zuo.c:979:21: warning: comparison of integers of different signs: 'zuo_int32_t' (aka 'int') and 'unsigned long long' [-Wsign-compare]
      for (i = 0; i < len / sizeof(zuo_int32_t); i++)
                  ~ ^ ~~~~~~~~~~~~~~~~~~~~~~~~~
```

My understanding of C and integer arithmetic is woefully incomplete, so I'm not sure if casting (on which side of <?) is appropriate or if there is a better solution.

### DIFF
--- a/.github/workflows/zuo-cflags.yml
+++ b/.github/workflows/zuo-cflags.yml
@@ -1,0 +1,31 @@
+---
+name: Zuo with Strict Compiler Flags
+
+on:  # yamllint disable-line rule:truthy
+  push:
+    branches:
+      - master
+    paths:
+      - "racket/src/zuo/**"
+  pull_request:
+    paths:
+      - "racket/src/zuo/**"
+
+jobs:
+  build-linux:
+    runs-on: ubuntu-20.04
+    defaults:
+      run:
+        working-direcotry: "racket/src/zuo"
+
+    env:
+      CFLAGS: "-Werror -Wall -Wextra -Wstrict-prototypes -Wold-style-definition -Wshadow -Wpointer-arith -Wcast-qual -pedantic -O2 -std=c11"
+
+    steps:
+      - uses: actions/checkout@v4
+        with:
+          fetch-depth: 100
+      - name: Compile
+        run: cc $CFLAGS zuo.c -o zuo
+      - name: Check
+        run: ./zuo build.zuo check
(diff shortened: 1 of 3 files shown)


mflatt:
Thanks! These changes look like a good idea to me. I agree about `UNUSED`, but the style would be to use `ZUO_UNUSED`.

When I tried compiling with the flags you list, I saw many more warnings on various platforms, so I've pushed another commit to address those. Warnings from `-Wmissing-prototypes` repersent places where I meant to use `static`, so those changes are in the extra commit, too. And there was one place where `zuo.c` didn't compile with old MSVC, so I fixed that. (As far as I can tell, everything still compiles with C89, too.)

benknoble:
Thanks!

- I saw the cast on the arithmetic warning: I suppose there we're truncating the sizeof's `unsigned long long`? I guess I would need to stare at that code and the standard a little better to understand it.
- RE: my commit's comment on primitive dispatching and stack magic, I think I finally realized that those procedures implement differently primitive calling conventions (where 0 means "no arguments). That explains the unused warning better.

LiberalArtist:
Maybe it would be worth adding a CI action to test these flags, either here or in the Zuo repository.

benknoble:
I'll push a first draft of a CI workflow for that when I get an internet connection for my laptop. It could use some improvements (such as running on a matrix of build platforms).

mflatt:
I think CI is better (only) at the https://github.com/racket/zuo repo. I'd like to merge this change there, first, anyway.

@benknoble, I could create a new PR at the racket/zuo repo, but I'll leave it to you if you'd prefer to create it. Let me know.

benknoble:
Ah, I misunderstood the preferred direction. I will get this opened over there instead.

PR #4974, opened 2024-04-16
net/smtp and net/imap: add XOAUTH2 authentication support

mflatt:
Also, change debugging logging `net/smtp` to use a logger.


LiberalArtist:
IIUC, the support here is specifically for supplying the token to the IMAP or SMTP server: that is, the library user is responsible for putting the user through the OAuth 2.0 flow, or getting a token somehow, right? I'm not sure if it's worth mentioning that in the docs.

mflatt:
Yes, that's right. I've adjusted the docs to try to clarify.

PR #4976, opened 2024-04-18
Fix kill-thread description in threads.scrbl

gregr:
##### Checklist

- [x] documentation

### Description of change

This is an update to the reference documentation for kill-thread.

Like thread-suspend, kill-thread fails if it is managed by a custodian that is not current-custodian or one of its subordinates.  This new kill-thread description is based on the phrasing already used for the thread-suspend description.


mflatt:
Thanks!

gregr:
No problem.

PR #4982, opened 2024-04-25
`chaperone-treelist` fixes and doc edits

LiberalArtist:
This PR has two bug fixes and several documentation edits related to `chaperone-treelist`.

The bugs:

1. ~https://github.com/racket/racket/commit/4530cd9559fb63110da7351ad5ff6b034f16f754 documented that `#:ref` could be `#f`, but `check-chaperone-arguments` didn't actually allow it. This is fixed in ed70e7ffcb8cc91505759675eeaf3d50637c2636.~
3. The documented contract for `#:append2` had the wrong number of arguments (i.e. one of the state arguments was missing).

I would be happy to split this PR up further if either of those fixes seem important enough to put into 8.13 at this late stage: I only wish I had discovered them sooner!

The other non-documentation change in this PR is that I made `#:ref` optional.

Several things that I have *not* done (yet?):

1. ~At one point I considered whether `#:state` should also be optional, but I decided against it: if there were really a use-case for chaperoning for which state was irrelevant, it would be better to provide an alternate function that wouldn't need the corresponding arguments and return values, either.~
2. I'd like to do something about places that say procedures `must accept @racket[tl]` and produce results `chaperoned with the same procedures and properties as @racket[tl]`. The argument `tl` is without the layer of chaperoning added by this call to `chaperone-treelist`, so `treelist-chaperone-state` and impersonator-property accesses on that `tl` won't work. Perhaps pedantically, the argument `tl` might also be a derived treelist, rather than the `tl` from the formal arguments to `chaperone-treelist`. That also means the result treelist more precisely gets "the same procedures and properties" added by this call to `chaperone-treelist`, or something like that. But I haven't yet figured out a concise way of writing (the important parts of) that.
4. In the example I added, `tl` and `state` are italicized as though by `@var{}`, whereas `pos`, `v`, `other`, and `other-state` are not. They could easily all be italicized via `_`, but it might be more correct to italicize none of them.
5. After the paragraph beginning `When two chaperoned treelists are given to @racket[treelist-append] and @racket[append2-proc] is not used`, I'd like to say more about what happens when `append2-proc` *is* used, but I'm still confirming that I understand the details I think I want to write.

More generally I have some observations now that I've (finally) experimented with the improved treelist chaperoning a bit, but I probably won't finish writing those up tonight and, having found the bugs here, I didn't want to delay reporting them. (In brief, this still seems like a fine direction.)

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/treelists.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/treelists.scrbl
@@ -389,24 +389,24 @@ Like @racket[for/list] and @racket[for*/list], but generating
 @defproc[(chaperone-treelist [tl treelist?]
                              [#:state state any/c]
                              [#:state-key state-key any/c (list 'fresh)]
-                             [#:ref ref-proc (or/c #f (treelist? exact-nonnegative-integer? any/c any/c
-                                                       . -> . any/c))]
+                             [#:ref ref-proc (treelist? exact-nonnegative-integer? any/c any/c
+                                              . -> . any/c)]
                              [#:set set-proc (treelist? exact-nonnegative-integer? any/c any/c
                                               . -> . (values any/c any/c))]
                              [#:insert insert-proc (treelist? exact-nonnegative-integer? any/c any/c
                                                     . -> . (values any/c any/c))]
-                             [#:append append-proc (treelist? treelist? any/c
-                                                    . -> . (values treelist? any/c))]
-                             [#:prepend prepend-proc (treelist? treelist? any/c
-                                                      . -> . (values treelist? any/c))]
-                             [#:append2 append2-proc (or/c #f (treelist? treelist? any/c
-                                                               . -> . (values treelist? any/c any/c))) #f]
                              [#:delete delete-proc (treelist? exact-nonnegative-integer? any/c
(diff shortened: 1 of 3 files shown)


mflatt:
I think the better fix for `#:ref` as `#f` is probably to remove that option from the contract in the docs, and then also remove the unneeded branch in the implementation. A `#f` was allowed at one point to enable efficient `treelist-append`, but the addition of more `append` hooks addressed that problem more completely.

LiberalArtist:
> I think the better fix for `#:ref` as `#f` is probably to remove that option from the contract in the docs, and then also remove the unneeded branch in the implementation.

That makes sense. I've pushed a revised series with a docs-only patch, then the implementation change. I changed `chaperone-mutable-treelist` and `impersonate-mutable-treelist` to match, and in the process noticed that `chaperone-mutable-treelist` should be documented as returning a non-chaperone impersonator.

> In the example I added, `tl` and `state` are italicized as though by `@var{}`, whereas `pos`, `v`, `other`, and `other-state` are not. They could easily all be italicized via `_`, but it might be more correct to italicize none of them.

I used `_`, since it was easy, for the sake of at least local consistency.

PR #4984, opened 2024-04-26
Fix equal-always? on fl/fxvector, stencil-vector

### LINKED ISSUE #4983: `flvector?` and `fxvector?` are missing `equal-always?` support
8.12 [cs]

**What program did you run?**

```scheme
Welcome to Racket v8.12 [cs].
> (require racket/flonum)
> (define foo (flvector 1.0 2.0 3.0))
> (define bar (flvector 1.0 2.0 3.0))
> foo
(flvector 1.0 2.0 3.0)
> bar
(flvector 1.0 2.0 3.0)
> (equal-always? foo bar)
#t
> (flvector-set! foo 1 20.0)
> foo
(flvector 1.0 20.0 3.0)
> bar
(flvector 1.0 2.0 3.0)
> (equal-always? foo bar)
#f
```

And the same thing for fxvectors:

```scheme
> (require racket/fixnum)
> (define bar (fxvector 1 2 3))
> (define foo (fxvector 1 2 3))
> foo
(fxvector 1 2 3)
> bar
(fxvector 1 2 3)
> (equal-always? foo bar)
#t
> (fxvector-set! foo 1 20)
> foo
(fxvector 1 20 3)
> bar
(fxvector 1 2 3)
> (equal-always? foo bar)
#f
```

Generally, for types that are defined with attention to `equal-always?` support, the result of `equal-always?` doesn't vary when a value is mutated. Since flvectors and fxvectors are mutable, they should only be `equal-always?` if they're `eq?`, similarly to mutable strings or vectors.

Alternatively, some mutation operations could be considered to be part of an unsafe API, like `unsafe-set-immutable-car!`. If that's what `flvector-set!` and `fxvector-set!` are considered to be, the present behavior may be correct. But that's not the impression I get from the documentation; I think these were simply types whose equality behavior was overlooked when `equal-always?` was introduced.

AlexKnauth:
##### Checklist

- [x] Bugfix
- [x] tests included

### Description of change

Fixes `equal-always?` and `chaperone-of?` on `flvector`, `fxvector`, and `stencil-vector`, to return false on different values with the same contents at the time, because they are mutable.

Fixes #4983.

- [x] equality tests
- [x] hash-code tests
- [x] Racket CS
- [x] Racket BC

### DIFF
--- a/pkgs/racket-test-core/tests/racket/basic.rktl
+++ b/pkgs/racket-test-core/tests/racket/basic.rktl
@@ -118,6 +118,10 @@
 (test #t equal? 2 2)
 (test #t equal? (make-vector 5 'a) (make-vector 5 'a))
 (test #t equal? (box "a") (box "a"))
+(test #t equal? (make-flvector 5 0.0) (make-flvector 5 0.0))
+(test #t equal? (make-fxvector 5 0) (make-fxvector 5 0))
+(test #t equal? (stencil-vector #b10010 'a 'b) (stencil-vector #b10010 'a 'b))
+
 (test #f equal? "" (string #\null))
 
 (test #f equal? 'a "a")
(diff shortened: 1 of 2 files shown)


mflatt:
Looks ok to me. I worry a little that the fallthrough to `#%equal?` has changed, or maybe I'm misreading; even if I'm reading right, I can't see any reason that the change would create a problem.

AlexKnauth:
Yes, the fallthrough to `#%equal?` has changed in that it only falls through to that when the mode is `equal?` or `impersonator-of?`, not `equal-always?` or `chaperone-of?`. Are there any more immutable data structures that would need that fall through that I haven't accounted for?

mflatt:
There are no more that I can think of.

In trying some things, I discovered that paths are currently not handled right for `equal-always?`. The repairs (which I have ready to push later) are independent of your changes. If there's something else wrong like that on CS, I expect the problem will be again with the struct and not with `equal?`.

PR #4988, opened 2024-05-07
raco test: set current-test-invocation-directory

benknoble:
Running "raco test foo/bar/test.rkt" will cause Racket to change directory to "foo/bar" before invoking "test.rkt", and error messages are printed relative to (current-directory) like

    raco test: "foo/bar/test.rkt"
    --------------------
    fail
    FAILURE
    name:       check-equal?
    location:   test.rkt:5:0
    actual:     1
    expected:   2
    --------------------
    1/1 test failures

For an editor to consume this, it has to post-process the output to correlate "raco test: path" with "location: partial path" (which is difficult when running tests in parallel). This is not ideal; output with

    location:   foo/bar/test.rkt:5:0

is immediately useable by editors for jumping to the error.

Expose the original directory in which "raco test" was launched so that rackunit (or other test frameworks) can appropriately adjust filepaths when reporting errors.

##### Checklist

- [x] Bugfix
- [ ] Feature
- [ ] tests included
- [x] documentation

### DIFF
--- a/pkgs/compiler-lib/raco/testing.rkt
+++ b/pkgs/compiler-lib/raco/testing.rkt
@@ -4,10 +4,25 @@
 
 (provide test-log-enabled?
          test-log!
-         test-report)
+         test-report
+         current-test-invocation-directory)
+
+;; records the original "raco test" directory while raco test changes directory
+;; to invoke tests
+(define current-test-invocation-directory
+  (make-parameter
+   #f
+   (λ (path)
+     (cond
+       [(path-string? path) (path->directory-path (simplify-path path))]
+       [(not path) path]
+       [else (raise-argument-error 'current-test-invocation-directory
+                                   "(or/c #f path-string?)"
+                                   path)]))
+   'current-test-invocation-directory))
 
 (define test-log-enabled?
-  (make-parameter #t (lambda (v) (and v #t))))
+  (make-parameter #t (lambda (v) (and v #t)) 'test-log-enabled?))
 
 (define TOTAL 0)
 (define FAILED 0)
--- a/pkgs/racket-doc/scribblings/raco/test.scrbl
+++ b/pkgs/racket-doc/scribblings/raco/test.scrbl
@@ -407,4 +407,19 @@ with @exec{raco test} should also use this library to log test results.
  counted in the test log, such as when testing a custom check's failure
  behavior.}
 
+@defparam*[current-test-invocation-directory
+            path
+            (or/c #f path-string?)
+            (or/c #f path?)
+            #:value #f]{
+Contains the directory from which tests were invoked by, @emph{e.g.}, @exec{raco
+test}. This may differ from @racket[current-directory] when the test runner
+changes directory before invoking a specific test file and should be set by test
+runners to reflect the directory from which they were originally invoked.
+
+This should be used by test reports to display appropriate path names.
+
+@history[#:added "1.14"]
+}
+
 @history[#:added "1.13"]
--- a/pkgs/compiler-lib/compiler/commands/test.rkt
+++ b/pkgs/compiler-lib/compiler/commands/test.rkt
@@ -1151,6 +1151,7 @@
   "Save stdout and stderr to file, overwrite if it exists."
   (set! default-output-file file)]
  #:args file-or-directory-or-collects-or-pkgs
+ (current-test-invocation-directory (current-directory))
  (define (test)
    (define file-or-directory
      (maybe-expand-package-deps file-or-directory-or-collects-or-pkgs))
--- a/pkgs/compiler-lib/info.rkt
+++ b/pkgs/compiler-lib/info.rkt
@@ -12,7 +12,7 @@
 
 (define pkg-authors '(mflatt))
 
-(define version "1.13")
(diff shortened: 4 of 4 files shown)


pkgs/compiler-lib/raco/testing.rkt
```diff
@@ -4,7 +4,13 @@
...
 (provide test-log-enabled?
          test-log!
-         test-report)
+         test-report
+         current-test-invocation-directory)
+
+;; records the original "raco test" directory while raco test changes directory
+;; to invoke tests
+(define current-test-invocation-directory
+  (make-parameter #f))
```
usaoc, line 13: It’d be ideal if this parameter has a guard that checks the contract and normalizes the path string like `current-directory`.  I see the RackUnit-side change assumes that the parameter returns a *path* as opposed to path string, which makes such change even more important.  Less importantly, a name should be given to `make-parameter`.  (Since you’re changing this file, could you also change `test-log-enabled?` to add a name?)

pkgs/racket-doc/scribblings/raco/test.scrbl
```diff
@@ -407,4 +407,15 @@ with @exec{raco test} should also use this library to log test results.
  counted in the test log, such as when testing a custom check's failure
  behavior.}
 
+@defparam[current-test-invocation-directory test-invocation-directory (or/c #f path-string?) #:value #f]{
```
usaoc, line 410: And this should be documented like `current-directory` (notice how the setter and getter have different contracts), if the above change is made.

benknoble:
I believe the new version addresses @usaoc's comments. Should the out contract use `path-for-some-system?` since that's what the guard actually produces, or is `path?` sufficient documentation here?

range-diff:
```
1:  216c2bd1f5 ! 1:  df01b0d1dc raco test: set current-test-invocation-directory
    @@ Commit message
         that rackunit (or other test frameworks) can appropriately adjust
         filepaths when reporting errors.
     
    +    Also add a name to the test-log-enabled? parameter while we're here.
    +
      ## pkgs/compiler-lib/compiler/commands/test.rkt ##
     @@ pkgs/compiler-lib/compiler/commands/test.rkt: (define-values (sum _ deps)
        "Save stdout and stderr to file, overwrite if it exists."
    @@ pkgs/compiler-lib/raco/testing.rkt
     +;; records the original "raco test" directory while raco test changes directory
     +;; to invoke tests
     +(define current-test-invocation-directory
    -+  (make-parameter #f))
    ++  (make-parameter
    ++   #f
    ++   (λ (path)
    ++     (cond
    ++       [(path-string? path) (path->directory-path (simplify-path path))]
    ++       [(not path) path]
    ++       [else (raise-argument-error 'current-test-invocation-directory
    ++                                   "(or/c #f path-string?)"
    ++                                   path)]))
    ++   'current-test-invocation-directory))
      
      (define test-log-enabled?
    -   (make-parameter #t (lambda (v) (and v #t))))
    +-  (make-parameter #t (lambda (v) (and v #t))))
    ++  (make-parameter #t (lambda (v) (and v #t)) 'test-log-enabled?))
    + 
    + (define TOTAL 0)
    + (define FAILED 0)
     
      ## pkgs/racket-doc/scribblings/raco/test.scrbl ##
     @@ pkgs/racket-doc/scribblings/raco/test.scrbl: with @exec{raco test} should also use this library to log test results.
       counted in the test log, such as when testing a custom check's failure
       behavior.}
      
    -+@defparam[current-test-invocation-directory test-invocation-directory (or/c #f path-string?) #:value #f]{
    ++@defparam*[current-test-invocation-directory
    ++            path
    ++            (or/c #f path-string?)
    ++            (or/c #f path?)
    ++            #:value #f]{
     +Contains the directory from which tests were invoked by, @emph{e.g.}, @exec{raco
     +test}. This may differ from @racket[current-directory] when the test runner
     +changes directory before invoking a specific test file and should be set by test
```

usaoc:
I think a path or string will always end up as a path for the current platform, unless you explicitly produce a path for a different platform, so `path?` is correct here.

rfindler:
Sorry if I am missing something, but on the input side would path-string?
Be acceptable?

benknoble:
> Sorry if I am missing something, but on the input side would path-string? be acceptable?

Yes; indeed, the documentation and code (should) both agree that the parameter accepts `(or/c #f path-string?)`. The output coerces to `(or/c #f path?)`.

benknoble:
@mflatt it looks like I've missed a case somewhere: when I do `raco test -j 8 foo/bar/test.rkt`, I still get the new output. But with `raco test -j 8 foo/bar/test{1,2}.rkt`, I get the old output (meaning the new parameter hasn't been set, I think).

Any pointers?

mflatt:
@benknoble With `-j 8`, parallelism can be implemented with a thread, place, or process. For places, maybe line 121 of `test.rkt` is the right point to set the parameter in the new place, and maybe the value to be used for the parameter needs to be passed to that place. For processes, probably the value of the parameter needs to be passed as a command-line argument to be received by the `process` submodule.

benknoble:
Thanks. I started thinking along those lines; the code references are helpful.

PR #4989, opened 2024-05-08
Add `match-λ[*][*]`.

NoahStoryM:
##### Checklist

- [x] Feature
- [x] Documentation

### Description of change

This PR introduces Unicode variants of common lambda forms in Racket. A previous PR #1301 suggested similar changes but was closed. This PR revives that idea.


racket/collects/racket/match/define-forms.rkt
```diff
@@ -14,21 +14,29 @@
...
+     (provide match match* match/values
+              match-lambda match-lambda* match-lambda**
+              (rename-out [match-lambda   match-λ]
+                          [match-lambda*  match-λ*]
+                          [match-lambda** match-λ**])
+              match-let match-let*
+              match-let-values match-let*-values
               match-define match-define-values
               match-letrec match-letrec-values
-	      match/values match/derived match*/derived match-define-values
```
NoahStoryM, line 30: In the original code, `match-define-values` is provided twice.

racket/collects/racket/private/pre-base.rkt
```diff
@@ -200,6 +200,8 @@
              (rename new-prop:procedure prop:procedure)
              (rename #%app #%plain-app)
              (rename lambda #%plain-lambda)
+             (rename λ #%plain-λ)
```
NoahStoryM, line 203: I'm not sure if it should be `(rename λ #%plain-λ)` or `(rename lambda #%plain-λ)`.

racket/collects/racket/private/class-internal.rkt
```diff
@@ -331,8 +331,9 @@
               (append
                (kernel-form-identifier-list)
                (list
-                (quote-syntax #%app) ; racket/base app, as opposed to #%plain-app
+                (quote-syntax #%app) ; racket/base #%app, as opposed to #%plain-app
```
sorawee, line 334: Please revert all changes to this file. `free-identifier=?` already takes renaming into account. There's no need to test against another renamed form.

sorawee (commented):
Changes to `match-λ[*][*]` is nice. 

I really doubt that `#%plain-λ` is a good idea. No one will use `#%plain-lambda` anyway, since it's inferior than the regular `lambda`. It's only used inside the Racket internals, and also when people want to pattern match against expanded code. But in those two cases, `#%plain-λ` won't have much value.

My personal opinion is that this PR should be restricted to only changes in `match`-related form.

NoahStoryM:
> I really doubt that #%plain-λ is a good idea. No one will use #%plain-lambda anyway, since it's inferior than the regular lambda. It's only used inside the Racket internals, and also when people want to pattern match against expanded code. But in those two cases, #%plain-λ won't have much value.

Initially, I added the λ forms based on search results for `L:racket -lambda`. I have never used `#%plain-lambda`, so removing `#%plain-λ` is fine by me.

> My personal opinion is that this PR should be restricted to only changes in match-related form.

`case-λ` is likely to be widely used, and the related code change is just a single line: `(rename case-lambda case-λ)`. Is it necessary to split this PR? Of course, restricting this PR to only include changes to match-related forms is also acceptable to me.

racket/collects/racket/private/class-internal.rkt
NoahStoryM: This file already handles similar processes for `λ`, as seen in: `(or (free-identifier=? #'lam #'lambda) (free-identifier=? #'lam #'λ))`. I assumed the relationship between `lambda` and `λ` was similar to that between `case-lambda` and `case-λ`, which is why I treated `#%plain-λ` and `case-λ` in the same manner. If supporting `case-λ` is considered, do we still need to revert all changes to this file?

NoahStoryM: I read the code, and if I understand correctly, `lambda` and `λ` in `racket/base` are respectively `new-lambda` and `new-λ` in `kw.rkt`:
```
   (define-syntaxes (new-lambda new-λ)
     (let ([new-lambda ...])
       (values new-lambda new-lambda)))
```
Is it necessary to create a new PR and define `new-λ` through the renaming mechanism?

sorawee: For any big change like this (including this PR actually), I think it makes sense to get preliminary feedback from https://racket.discourse.group/ first. For example, it could be that people are somehow relying on `lambda` and `λ` not being `free-identifier=?` for some reasons, and the change would break their code.

But I agree that _ideally_ `lambda` should be `free-identifier=?` to `λ`.

rfindler: It seems unlikely that changing how `lambda` and `λ` are bound is going to work but before a discussion, just checking to see if racket (and a bunch of packages, maybe starting with main-distribution) builds would be informative.

pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -662,17 +662,20 @@ many values to expect from @racket[expr].
   ]
 }
 
-@defform[(match-lambda clause ...)]{
+@deftogether[(@defform[(match-lambda clause ...)]
+              @defform[(match-λ clause ...)])]{
 
 Equivalent to @racket[(lambda (id) (match id clause ...))].
 }
```
mflatt, line 669: Please add `@history[#:changed "8.13.0.5" @elem{Added @racket[match-λ].}]` here before the closing curly brace, and similar for the other two. I'll bump the Racket version number after merging this and related PRs.

mflatt (commented):
I think this will be ready after adding `history` notes to the documentation. You could also squash the commits, or I can squash when merging. Thanks!

pkgs/racket-doc/scribblings/reference/match.scrbl
```diff
@@ -666,12 +666,16 @@ many values to expect from @racket[expr].
...
+
+@history[#:changed "8.13.0.5" @elem{Added @racket[match-λ].}]
 }
 
 @deftogether[(@defform[(match-lambda* clause ...)]
               @defform[(match-λ* clause ...)])]{
 
 Equivalent to @racket[(lambda lst (match lst clause ...))].
+
+@history[#:changed "8.13.0.5" @elem{Added @racket[match-λ].}]
```
sorawee, line 678: This should be `match-λ*`. Ditto for `match-λ**` below.

sorawee (changes requested):

sorawee (approved):

samth (approved):

PR #4991, opened 2024-05-09
Define `λ` through the renaming mechanism.

NoahStoryM:
##### Checklist

- [x] Feature
- [x] tests included

### Description of change

To put it simply, this PR has the following changes:

Before changes:
```racket
(#%provide new-lambda new-λ)

(define-syntaxes (new-lambda new-λ)
   (let ([new-lambda ...])
     (values new-lambda new-lambda)))
```

After changes:
```racket
(#%provide new-lambda (rename new-lambda new-λ))

(define-syntax (new-lambda stx) ...)

```

I didn't know how to check the impact of this change on all packages, so I just ran `raco setup` and found no exception.

See also https://github.com/racket/racket/pull/4989#discussion_r1594165915.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/procs.rktl
+++ b/pkgs/racket-test-core/tests/racket/procs.rktl
@@ -449,10 +449,6 @@
 (test -4 procedure-arity-mask apply)
 
 ;; ------------------------------------------------------------
-;; Check if `lambda` and `λ` access the same binding.
-(test #t free-identifier=? #'lambda #'λ)
-
-;; ------------------------------------------------------------
 ;; Check arity reporting for methods.
 
 (map
--- a/pkgs/racket-test-core/tests/racket/procs.rktl
+++ b/pkgs/racket-test-core/tests/racket/procs.rktl
@@ -449,6 +449,10 @@
 (test -4 procedure-arity-mask apply)
 
 ;; ------------------------------------------------------------
+;; Check if `lambda` and `λ` access the same binding.
+(test #t free-identifier=? #'lambda #'λ)
+
+;; ------------------------------------------------------------
 ;; Check arity reporting for methods.
 
 (map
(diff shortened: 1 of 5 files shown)


pkgs/racket-test-core/tests/racket/procs.rktl
```diff
@@ -448,6 +448,10 @@
 
 (test -4 procedure-arity-mask apply)
 
+;; ------------------------------------------------------------
+;; Check if `lambda` and `λ` access the same binding.
+(test #t free-identifier=? #'lambda #'λ)
+
```
mflatt, line 454: The `syntax.rktl` file is probably a better place for this check, since that one is more about syntactic forms, binding, and syntax checking. But I don't object if you really prefer the test here.

mflatt (changes requested):
Given the potential to change expansion behavior, I think it would be worth adding a note to the documentation for `lambda` and `λ` to record the change. I suggest `@history[#:changed "8.13.0.5" @elem{Adjusted binding so that @racket[(free-identifier=? #'λ #'lambda)] produces @racket[#t].}]` added before the closing brace of line 2106 of `syntax.scrbl`.

samth:
It looks like this broke a bunch of packages. A representative failure is here: https://plt.cs.northwestern.edu/pkg-build/server/built/fail/phc-toolkit.txt

The failures mostly are in Typed Racket-related packages, but reverting the changes since 8.13 in Typed Racket does not fix the problems.

mflatt:
It's not clear that those failures are related to this PR. In fact, they report version 8.13.0.4, but I bumped the version to 8.13.0.5 just after this merge.

Based on the error message, it seems more likely that the failure is related to a847564e707f81f60a441f49b1f6dd7918cb9d19. I'm able to replicate the error, so I'll investigate more.

mflatt:
Commit a79cd8050c727ba8e10afab6c17f8e57264ad4bf is intended to repair an expander problem that caused trouble for TR-related packages, including phc-toolkit.

PR #4996, opened 2024-05-19
Update `list-prefix?` to avoid duplicate checks

Antigen-1:
### Description of change
Small changes that fix a typo in a comment and avoid duplicate checks in the `list-prefix?` function.


sorawee:
Thank you!

sorawee:
FWIW, it would be fair to consider this a bugfix if you can demonstrate that it was unnecessarily slow before the PR. But you can also consider it a feature (which counts general improvements).

PR #4997, opened 2024-05-19
raco test: set more current-test-invocation-directory

benknoble:
This adds the effects of df01b0d1dc (raco test: set current-test-invocation-directory, 2024-05-07) to --place and --process mode, too; as a result, parallel invocations with -j work, too.

The changes to the subprocess argv processing and place channel received list processing do perhaps beg the question: is the protocol for unpacking the values communicated via side-channel error prone?

##### Checklist

- [x] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation

Specifically, this addresses issues in https://github.com/racket/racket/pull/4988#issuecomment-2119274879

### DIFF
--- a/pkgs/compiler-lib/compiler/commands/test.rkt
+++ b/pkgs/compiler-lib/compiler/commands/test.rkt
@@ -68,8 +68,9 @@
   (require raco/testing
            racket/file
            compiler/private/cm-minimal)
-  ;; Arguments are a temp file to hold test results, the module
-  ;; path to run, and the `dynamic-require` second argument:
+  ;; Arguments include a temp file to hold test results, the module path to run,
+  ;; and the `dynamic-require` second argument. See the 'process case of
+  ;; dynamic-require-elsewhere.
   (define argv (current-command-line-arguments))
   (define result-file (vector-ref argv 0))
   (define test-module (read (open-input-string (vector-ref argv 1))))
@@ -77,7 +78,10 @@
   (define d (read (open-input-string (vector-ref argv 3))))
   (define make? (read (open-input-string (vector-ref argv 4))))
   (define errortrace-path-or-false (read (open-input-string (vector-ref argv 5))))
-  (define args (list-tail (vector->list argv) 6))
+  (define test-invocation-path (bytes->path (read (open-input-string (vector-ref argv 6)))))
+  (define args (list-tail (vector->list argv) 7))
+
+  (current-test-invocation-directory test-invocation-path)
 
   ;; In case PLTUSERHOME is set, make sure relevant
   ;; directories exist:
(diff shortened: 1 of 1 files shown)


benknoble:
The following test driver is useful (given simple test files):
```sh
#! /bin/sh

for mode in --process --place; do
  if ! raco test $mode -j 2 foo/bar/test*.rkt 2>&1 | grep 'location:.*foo/bar'; then
    printf 'failure: %s\n' $mode >&2
    exit 1
  fi
done
```

Roughly,
```console
# tree foo
foo
└── bar
    ├── test.rkt
    └── test2.rkt

2 directories, 2 files
# diff foo/bar/test*
# cat foo/bar/test.rkt
#lang racket
(require rackunit)
(test-equal? "fail" 1 2)
```

pkgs/compiler-lib/compiler/commands/test.rkt
```diff
@@ -274,9 +281,12 @@
                       (format "~s" d)
                       (format "~s" make?)
                       (format "~s" (and load-errortrace? errortrace-module-path))
+                      (format "~s" (cond
+                                     [(current-test-invocation-directory) => path->string]
```
mflatt, line 285: This should be `path->bytes` with `bytes->path` on the other end, `path->string` does not always round-trip with `string->path`.

benknoble: Indeed. I notice that `normalize-module-path` uses `path->string` and is also put in the subprocess arguments. Something to audit/fix another day?

`git range-diff`:
```
1:  b3a32b284c ! 1:  50ed11c902 raco test: set more current-test-invocation-directory
    @@ Commit message
     
         This adds the effects of df01b0d1dc (raco test: set
         current-test-invocation-directory, 2024-05-07) to --place and --process
    -    mode, too; as a result, parallel invocations with -j work, too.
    +    mode, too; as a result, parallel invocations with -j work, too. To pass
    +    the path to a subprocess, we need to convert to bytes (since paths do
    +    not round trip through strings).
     
         The changes to the subprocess argv processing and place channel received
         list processing do perhaps beg the question: is the protocol for
    @@ pkgs/compiler-lib/compiler/commands/test.rkt: (define rt-module (read (open-inpu
        (define make? (read (open-input-string (vector-ref argv 4))))
        (define errortrace-path-or-false (read (open-input-string (vector-ref argv 5))))
     -  (define args (list-tail (vector->list argv) 6))
    -+  (define test-invocation-path (read (open-input-string (vector-ref argv 6))))
    ++  (define test-invocation-path (bytes->path (read (open-input-string (vector-ref argv 6)))))
     +  (define args (list-tail (vector->list argv) 7))
     +
     +  (current-test-invocation-directory test-invocation-path)
    @@ pkgs/compiler-lib/compiler/commands/test.rkt: (define ps
                            (format "~s" make?)
                            (format "~s" (and load-errortrace? errortrace-module-path))
     +                      (format "~s" (cond
    -+                                     [(current-test-invocation-directory) => path->string]
    ++                                     [(current-test-invocation-directory) => path->bytes]
     +                                     [else #f]))
                            args)))
                 (define proc (list-ref ps 4))
```

mflatt: > Indeed. I notice that `normalize-module-path` uses `path->string` and is also put in the subprocess arguments. Something to audit/fix another day?

Yes, agreed. Thanks for the fixing the directory, at least!

PR #4998, opened 2024-05-20
Fix the `one-arg-check` function

Antigen-1:
##### Checklist

- [x] Bugfix
- [x] Tests included

### Description of change
Fix the checker of `vector-filter` and `vector-filter-not`. It should check whether the second argument is a `vector?` according to its comment.

```
Welcome to Racket v8.13 [cs].
> (vector-filter (lambda (x) x) (list 1))
in-vector: contract violation
  expected: vector
  given: '(1)
 [,bt for context]
```

This is definitely a bugfix, but I have problems in including tests. Is it necessary? Where can I include new tests?

### DIFF
--- a/racket/collects/racket/vector.rkt
+++ b/racket/collects/racket/vector.rkt
@@ -96,7 +96,7 @@
   (unless (and (procedure? f) (procedure-arity-includes? f 1))
     (raise-argument-error name "(any/c . -> . any/c)" 0 f))
   (unless (vector? v)
-    (raise-argument-error name "vector?" 1 v)))
+    (raise-argument-error name "vector?" v)))
 
 (define (vector-filter f v)
   (one-arg-check f v 'vector-filter)
--- a/racket/collects/racket/vector.rkt
+++ b/racket/collects/racket/vector.rkt
@@ -94,7 +94,9 @@
 ;; uses name for error reporting
 (define (one-arg-check f v name)
   (unless (and (procedure? f) (procedure-arity-includes? f 1))
-    (raise-argument-error name "(any/c . -> . any/c)" 0 f)))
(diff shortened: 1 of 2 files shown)


racket/collects/racket/vector.rkt
```diff
@@ -94,7 +94,9 @@
 ;; uses name for error reporting
 (define (one-arg-check f v name)
   (unless (and (procedure? f) (procedure-arity-includes? f 1))
-    (raise-argument-error name "(any/c . -> . any/c)" 0 f)))
+    (raise-argument-error name "(any/c . -> . any/c)" 0 f))
```
sorawee, line 97: If I read correctly, `0` is provided here as a `bad-pos`, so we are using the second form of [`raise-argument-error`](https://docs.racket-lang.org/reference/exns.html#%28def._%28%28quote._~23~25kernel%29._raise-argument-error%29%29). As a result, there should be one more argument at the end: `v`.

racket/collects/racket/vector.rkt
```diff
@@ -94,7 +94,9 @@
...
+  (unless (vector? v)
+    (raise-argument-error name "vector?" v)))
```
sorawee, line 99: For consistency with the above `raise-argument-error`, let's also use the second form. That is, include `1` as `bad-pos`, and also add `f` as an argument before the last argument.

sorawee (changes requested):

sorawee (approved):

sorawee:
Thanks!

Re `err/rt-test`: yeah, I don't think there's any documentation. I do not know the history, but it looks like the system predates RackUnit. A lot of these tests should be migrated to RackUnit, IMO.

PR #5004, opened 2024-05-29
Chez+racket: support building with cosmopolitan libc

jjsullivan5196:
##### Checklist

- [X] Feature
- [x] documentation

### Description of change

This patch adds some minimal scaffolding to allow building and running Chez and the Racket main distribution with cosmopolitan libc, affording access to 6 operating systems with a single binary https://github.com/jart/cosmopolitan

The main contents of this patch are a couple of platform checks, and an addition to Racket's bootup procedure such that it can load the start image and packages from an embedded ZIP object store provided by the platform. I know this isn't exactly new in terms of what `raco exe` can do, but it does collapse the amount of binary images needed to ship to different platforms, while taking away ABI variability/build churn.

The shell script included in the commit message demonstrates shipping the main distribution as a single binary file, which runs on amd64 Linux, Windows 10, MacOS, NetBSD, OpenBSD, and FreeBSD. These are all covered by the static minimal racket binary built with the cosmopolitan toolchain. aarch64 should also work, though I still need to try it out.

Of course, this isn't all sunshine and rainbows. Cosmo does ship its own implementation of a POSIX ABI to cover most OS interfacing, but Racket assumes a lot about the underlying system from build time (such as it pertains to the foreign interface for instance, getting the gui package to work might be a challenge). I don't want to make outlandish claims of what's possible or worth doing, but I thought this was interesting and useful enough to share with the maintainers. Also where I might not be expert in the details of the platform, @jart or @ahgamut should be able to provide some clarification. I'd be happy to make any changes the authors would prefer of course. I'll also add an accompanying PR to the Chez repo https://github.com/cisco/ChezScheme/pull/837

### DIFF
--- a/racket/src/ChezScheme/c/expeditor.c
+++ b/racket/src/ChezScheme/c/expeditor.c
@@ -677,7 +677,7 @@ static void s_ee_set_color(int color_id, IBOOL background) {
 # include <unistd.h>
 # include <time.h>
 #endif
-#if !defined(__GLIBC__) && !defined(__OpenBSD__) && !defined(__NetBSD__) && !defined(__linux__) && !defined(__EMSCRIPTEN__) && !defined(NO_USELOCALE)
+#if !defined(__GLIBC__) && !defined(__COSMOPOLITAN__) && !defined(__OpenBSD__) && !defined(__NetBSD__) && !defined(__linux__) && !defined(__EMSCRIPTEN__) && !defined(NO_USELOCALE)
 # include <xlocale.h>
 #endif
 
--- a/racket/src/ChezScheme/c/prim5.c
+++ b/racket/src/ChezScheme/c/prim5.c
@@ -749,8 +749,8 @@ static ptr s_system(const char *s) {
 #ifdef WIN32
   return Sinteger(status);
 #else
-  if WIFEXITED(status) return Sinteger(WEXITSTATUS(status));
-  if WIFSIGNALED(status) return Sinteger(-WTERMSIG(status));
+  if (WIFEXITED(status)) return Sinteger(WEXITSTATUS(status));
+  if (WIFSIGNALED(status)) return Sinteger(-WTERMSIG(status));
   S_error("system", "cannot determine subprocess exit status");
   return 0 /* not reached */;
 #endif /* WIN32 */
--- a/racket/src/ChezScheme/c/version.h
+++ b/racket/src/ChezScheme/c/version.h
@@ -86,7 +86,7 @@ FORCEINLINE void store_unaligned_uptr(uptr *addr, uptr val) {
 /*****************************************/
 /* Operating systems                     */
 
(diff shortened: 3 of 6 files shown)


jjsullivan5196:
@samth 

> Also, significant parts of base Racket depend on the ffi; for example the ability to use openssl to download packages over HTTPS or use sqlite to manage databases when building code and documentation. Do those work with this set of changes?

On linux it does, but I can't say the same for other OSes as it looks like `(system-type)` only uses information available at build time. I don't know if that would be trivial to change, I did see the source in `system.ss` and it didn't look too bad, but I'm not familiar with how the build steps capture that information. (ie the main distribution built on linux will still eval `(system-type 'os*)` to `'linux` on windows) Also note this doesn't touch the behavior for other libc implementations. I am able to get a REPL on Windows 10 and Linux with the build distribution in `racket.com` at least.

> I'm confused about why this particular situation for locating .boot files should be hardcoded. Can using --enable-scheme in configure handle this case? If not, why not -- maybe it should be extended.

I agree that its not particularly ideal, but since the APE image always provides the embedded zip archive, it seems like the most reliable solution for storing the boot images. I could try to work it into the build steps somehow, but I didn't want to disrupt how the existing targets work.

Also

```
john@crunchbox:~/src/racket/racket/src/build$ ../configure --help-cs | grep scheme
  --enable-scheme=<path>  use <path> as Chez Scheme for build
```

Maybe you were referring to a different configure option? As far as I understand, there's no need to use different Chez sources for this to work.

samth:
If you want to build a cross build, you can see the `--build`, `--host`, and `--target` `configure` options. But I guess this points to a bigger issue -- what should this even do when run on a machine different from how it was compiled? Building against cosmopolitan libc doesn't sound very useful if the result can only run on one platform, but it seems like changing that would require much bigger changes than proposed here. 

`--enable-scheme` should point to how to find the `scheme.boot` and `petite.boot` files. `racket.boot` is an intermediate build product that gets combined with the other boot files into the `racket` executable. So I'm confused about why they would be in a global store.

jjsullivan5196:
> but it seems like changing that would require much bigger changes than proposed here.

That's a fair point. I do imagine searching for dynamic libraries/filling the ABI gaps would be enough of a challenge on its own to get right, considering how things are currently done with the `foreign` package. I can try it as a recreational exercise, but I might not be the right person to carry that forward.

I can leave this open as a draft, but if there's anyone who would be more interested in getting this to work correctly, they can start from what I have here. I'll try to tidy up the details where I can at least. Thanks for taking a look @samth !

mflatt:
I merged the Chez Scheme part of this (by syncing with the Chez Scheme commit) and also changed the `u_char`s (but to `unsigned char`, because C89).

For the boot-file part, probably there should be a new `configure` option to disable embedded boot files and specify where to find them. If that sounds like something that could work, I can look into adding that.

PR #5007, opened 2024-06-02
feat(main.zuo): add support for $JOBS environment variable

virchau13:
##### Checklist

(Not sure what to put here - this is just a convenient build system change)

### Description of change

Currently, if you want to change the number of threads used during a top-level Racket build, you must edit the Makefile. This isn't particularly conducive to automated building (it requires a Makefile diff), which bit me trying to build on a particularly slow device. This adds support for a `$JOBS` environment variable to conveniently change the number of threads for the build.

### DIFF
--- a/main.zuo
+++ b/main.zuo
@@ -52,7 +52,11 @@
     (define cpus (let ([s (lookup 'CPUS)])
                    (and (not (equal? s ""))
                         s)))
-    (define simple-jobs (or cpus
+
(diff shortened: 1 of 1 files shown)


benknoble:
I might be missing something, but is the problem that `zuo . JOBS=<N>` doesn't override the `cpus` count? Perhaps those should instead flipped in `(or cpus (lookup "JOBS") …)`?

virchau13:
Er, yes it does. Sorry about that... :P

PR #5018, opened 2024-06-20
Support `#:gc-toplevel?` option in `#lang compiler/demod`.

samth:
Connects to the option of the same name in `demodularize`. Still needs tests/docs.


mflatt:
Thanks! I have a similar change that is part of making definition pruning work better — new PR momentarily.

PR #5019, opened 2024-06-20
Improve definition pruning for the demodularizer

mflatt:
Adds `#:prune-definitions` to `compiler/demo`, similar to #5018. Also adds a new pass to enable more pruning, and improves the interaction of syntax-perserving mode and pruning so that they work together.

With these changes, `#lang racket "hello"` can be demodularized and pruned to a 34k ".zo" (AArch64), instead of 500k with the old pruning implementation or 1.8M without pruning. Pruning the Rhombus implementation saves 10-15%, since most of it is exported, with a resulting size of 5.5M. Pruning `#lang rhombus "hello"` is still 465k, so there's plenty of room for improvement, but that's still usefully smaller than the Rhombus implementation.


samth:
With this PR, running a module that uses `compiler/demod` in racket-mode (or with eg `raco expand`) errors with the message:

```
?: invalid context for a compiled module from module-begin expansion
  in: ((#%module-begin find-collection/run #:exe))
```

mflatt:
> With this PR, running a module that uses `compiler/demod` in racket-mode (or with eg `raco expand`) errors with the message:
> 
> ```
> ?: invalid context for a compiled module from module-begin expansion
>   in: ((#%module-begin find-collection/run #:exe))
> ```

Seems that I lost `(syntax-local-compiling-module?)` somewhere along the way. I'll push a repair.

PR #5020, opened 2024-06-21
Switch some file operations to be statically-bound kw functions.

samth:
This is designed to make these functions work better with demodularization.

Includes support for helping `typed/racket` with `make-temporary-file`, including
and ugly hack only needed to support TR testing, which hopefully someone has a 
better idea for.

### DIFF
--- a/racket/collects/racket/file.rkt
+++ b/racket/collects/racket/file.rkt
@@ -10,10 +10,11 @@
          make-directory*
          make-parent-directory*
 
-         make-temporary-file
-         make-temporary-directory
-         make-temporary-file*
-         make-temporary-directory*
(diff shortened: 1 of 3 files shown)


samth:
Typed Racket support is in racket/typed-racket#1380

mflatt:
Instead of `temp-file/dir-xform`, could the same result be achieved by expanding to the creation of a submodule that exports the procedure?

samth:
You mean a submodule that has an export for the identifier naming the procedure? Should there be one submodule for all of them, or one per definition?

mflatt:
Seems like either way could work, but one submodule that accumulates all via `module+` may be most convenient.

mflatt:
Latest LGTM

PR #5023, opened 2024-06-24
Update concurrency.scrbl

luistung:
some mistake in section 18.7's code. there should be "<=" instead of ">="

##### Checklist

- [x] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation

### Description of change

according to the requirement " wait until at least a certain number of items have been produced.",  the condition should be "(<= (car waiter) total-items-seen)"

### DIFF
--- a/pkgs/racket-doc/scribblings/guide/concurrency.scrbl
+++ b/pkgs/racket-doc/scribblings/guide/concurrency.scrbl
@@ -579,7 +579,7 @@ to wait until at least a certain number of items have been produced.
            (code:comment "we check to see if there has been enough")
            (code:comment "production")
            (cond
-             [(>= (car waiter) total-items-seen)
+             [(<= (car waiter) total-items-seen)
               (code:comment "if so, we send a message back on the channel")
               (code:comment "and continue the loop without that item")
               (handle-evt


sorawee:
Does this look correct? If so, let's merge it. @rfindler @mflatt

benknoble:
This looks right to me: each waiter is waiting until _at least_ `n` items (`(car waiter)`) have been produced, which is tracked by `total-items-seen`. So we should only notify a waiter when `total-items-seen >= n`, or `(>= total-items-seen (car waiter))` (equivalently `(<= (car waiter) total-items-seen))`). Perhaps swapping the comparison order (and naming some intermediate pieces) would make the logic clearer.

benknoble:
I must also be missing something, but I don't see where the waiter's channel (`(cdr waiter)`) is actually written to? It's not clear from [`channel-put-event` docs](https://docs.racket-lang.org/reference/channel.html#%28def._%28%28quote._~23~25kernel%29._channel-put-evt%29%29) that when this event is handled it has _actually_ sent a value (but it must or the example wouldn't terminate).
```scheme
; if so, we send a message back on the channel
; and continue the loop without that item
(handle-evt
 (channel-put-evt
  (cdr waiter)
  (void))
 (λ (_) (loop items total-items-seen (remove waiter waiters))))
```

rfindler:
Maybe this should be tested? We can pull it out of a file, I suppose, to
put it into the docs programmatically.

Robby

benknoble:
Ok, here's something really telling: Using `(wait-at-least 5)` instead of `10` without this change deadlocks; using it with this change completes.

In every run I have tried that doesn't deadlock (see previous sentence), the waiter events don't get handled until all 10 items have been produced (the example code at the end of this comment has a println that should fire when deciding between the handle-evt and the never-evt for each waiter. But it only sees `total-items-seen` as 10, not any of the intermediate results.). When using `(wait-at-least 10)`, then, the choice of `>=` or `<=` is immaterial and in fact hides what I'd consider a bug in the example. Adding `(sleep (random 1 3))` before the `produce` lines demonstrates the system a little better, including the fact that `>=` is wrong.

Here's the buggy code instrumented to show it is wrong:
```racket
#lang racket

(define/contract (produce x)
  (-> any/c void?)
  (channel-put producer-chan x))

(define/contract (consume)
  (-> any/c)
  (channel-get consumer-chan))

(define/contract (wait-at-least n)
  (-> natural? void?)
  (define c (make-channel))
  ; we send a new channel over to the
  ; main loop so that we can wait here
  (channel-put wait-at-least-chan (cons n c))
  (void (channel-get c)))

(define producer-chan (make-channel))
(define consumer-chan (make-channel))
(define wait-at-least-chan (make-channel))

(void
 (thread
  (λ ()
    (let loop ([items '()]
               [total-items-seen 0]
               [waiters '()])
      ; instead of waiting on just production/
      ; consumption now we wait to learn about
      ; threads that want to wait for a certain
      ; number of elements to be reached
      (apply
       sync
       (handle-evt
        producer-chan
        (λ (i) (loop (cons i items)
                     (+ total-items-seen 1)
                     waiters)))
       (handle-evt
        (if (null? items)
            never-evt
            (channel-put-evt consumer-chan (car items)))
        (λ (_) (loop (cdr items) total-items-seen waiters)))

       ; wait for threads that are interested
       ; the number of items produced
       (handle-evt
        wait-at-least-chan
        (λ (waiter) (loop items total-items-seen (cons waiter waiters))))

       ; for each thread that wants to wait,
       (for/list ([waiter (in-list waiters)])
         ; we check to see if there has been enough
         ; production
         (println `([waiter ,waiter] [total-items-seen ,total-items-seen]))
         (cond
           [(>= (car waiter) total-items-seen)
            ; if so, we send a message back on the channel
            ; and continue the loop without that item
            (handle-evt
             (channel-put-evt
              (cdr waiter)
              (void))
             (λ (_) (loop items total-items-seen (remove waiter waiters))))]
           [else
            ; otherwise, we just ignore that one
            never-evt])))))))

; an example (non-deterministic) interaction
(define thds
  (for/list ([i (in-range 10)])
    (thread (λ ()
              (sleep (random 1 3))
              (produce i)
              (wait-at-least 10)
              (display (format "~a -> ~a\n" i (consume)))))))

(for ([thd (in-list thds)])
  (thread-wait thd))
```

Here's one version of the working code with the same instrumentation:
```racket
#lang racket

(define/contract (produce x)
  (-> any/c void?)
  (channel-put producer-chan x))

(define/contract (consume)
  (-> any/c)
  (channel-get consumer-chan))

(define/contract (wait-at-least n)
  (-> natural? void?)
  (define c (make-channel))
  ; we send a new channel over to the
  ; main loop so that we can wait here
  (channel-put wait-at-least-chan (cons n c))
  (void (channel-get c)))

(define producer-chan (make-channel))
(define consumer-chan (make-channel))
(define wait-at-least-chan (make-channel))

(void
 (thread
  (λ ()
    (let loop ([items '()]
               [total-items-seen 0]
               [waiters '()])
      ; instead of waiting on just production/
      ; consumption now we wait to learn about
      ; threads that want to wait for a certain
      ; number of elements to be reached
      (apply
       sync
       (handle-evt
        producer-chan
        (λ (i) (loop (cons i items)
                     (+ total-items-seen 1)
                     waiters)))
       (handle-evt
        (if (null? items)
            never-evt
            (channel-put-evt consumer-chan (car items)))
        (λ (_) (loop (cdr items) total-items-seen waiters)))

       ; wait for threads that are interested
       ; the number of items produced
       (handle-evt
        wait-at-least-chan
        (λ (waiter) (loop items total-items-seen (cons waiter waiters))))

       ; for each thread that wants to wait,
       (for/list ([waiter (in-list waiters)])
         ; we check to see if there has been enough
         ; production
         (println `([waiter ,waiter] [total-items-seen ,total-items-seen]))
         (cond
           [(<= (car waiter) total-items-seen)
            ; if so, we send a message back on the channel
            ; and continue the loop without that item
            (handle-evt
             (channel-put-evt
              (cdr waiter)
              (void))
             (λ (_) (loop items total-items-seen (remove waiter waiters))))]
           [else
            ; otherwise, we just ignore that one
            never-evt])))))))

; an example (non-deterministic) interaction
(define thds
  (for/list ([i (in-range 10)])
    (thread (λ ()
              (sleep (random 1 3))
              (produce i)
              (wait-at-least 5)
              (display (format "~a -> ~a\n" i (consume)))))))

(for ([thd (in-list thds)])
  (thread-wait thd))
```

Here's code that deadlocks:
```racket
#lang racket

(define/contract (produce x)
  (-> any/c void?)
  (channel-put producer-chan x))

(define/contract (consume)
  (-> any/c)
  (channel-get consumer-chan))

(define/contract (wait-at-least n)
  (-> natural? void?)
  (define c (make-channel))
  ; we send a new channel over to the
  ; main loop so that we can wait here
  (channel-put wait-at-least-chan (cons n c))
  (void (channel-get c)))

(define producer-chan (make-channel))
(define consumer-chan (make-channel))
(define wait-at-least-chan (make-channel))

(void
 (thread
  (λ ()
    (let loop ([items '()]
               [total-items-seen 0]
               [waiters '()])
      ; instead of waiting on just production/
      ; consumption now we wait to learn about
      ; threads that want to wait for a certain
      ; number of elements to be reached
      (apply
       sync
       (handle-evt
        producer-chan
        (λ (i) (loop (cons i items)
                     (+ total-items-seen 1)
                     waiters)))
       (handle-evt
        (if (null? items)
            never-evt
            (channel-put-evt consumer-chan (car items)))
        (λ (_) (loop (cdr items) total-items-seen waiters)))

       ; wait for threads that are interested
       ; the number of items produced
       (handle-evt
        wait-at-least-chan
        (λ (waiter) (loop items total-items-seen (cons waiter waiters))))

       ; for each thread that wants to wait,
       (for/list ([waiter (in-list waiters)])
         ; we check to see if there has been enough
         ; production
         (println `([waiter ,waiter] [total-items-seen ,total-items-seen]))
         (cond
           [(>= (car waiter) total-items-seen)
            ; if so, we send a message back on the channel
            ; and continue the loop without that item
            (handle-evt
             (channel-put-evt
              (cdr waiter)
              (void))
             (λ (_) (loop items total-items-seen (remove waiter waiters))))]
           [else
            ; otherwise, we just ignore that one
            never-evt])))))))

; an example (non-deterministic) interaction
(define thds
  (for/list ([i (in-range 10)])
    (thread (λ ()
              (sleep (random 1 3))
              (produce i)
              (wait-at-least 5)
              (display (format "~a -> ~a\n" i (consume)))))))

(for ([thd (in-list thds)])
  (thread-wait thd))
```

mflatt:
Thanks, all! I'll merge this change for now. I agree that the documentation on channels and put events is unclear and that the example should be improved more, but I'll work on those afterward.

PR #5024, opened 2024-06-24
unixstyle-install.rkt: unbreak for powerpc

barracuda156:
##### Checklist

- [x] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation

### Description of change

Fixes the build on Darwin PowerPC, which is currently broken due to misspecified Mach-O magic numbers.

### DIFF
--- a/racket/collects/setup/unixstyle-install.rkt
+++ b/racket/collects/setup/unixstyle-install.rkt
@@ -258,8 +258,11 @@
                                                            #""
                                                            r))))])
     (cond [(or (regexp-match #rx#"^\177ELF" magic)
+               ;; Mach-O magic numbers for LE/BE, 32/64-bit
                (regexp-match #rx#"^\316\372\355\376" magic)
-               (regexp-match #rx#"^\317\372\355\376" magic))
+               (regexp-match #rx#"^\317\372\355\376" magic)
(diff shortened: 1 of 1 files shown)


racket/collects/setup/unixstyle-install.rkt
```diff
@@ -259,7 +259,9 @@
                                                            r))))])
     (cond [(or (regexp-match #rx#"^\177ELF" magic)
                (regexp-match #rx#"^\316\372\355\376" magic)
```
mflatt, line 261: Worth adding a comment (further improving over the original) that these are all Mach-O variants, 32-bit and 64-bit in both little and big endian?

barracuda156: Yeah, sure, will add now.

P. S. Let’s have `ppc64` here as well, it does not hurt even if it is not built, as of now; otherwise this problem is obscure enough that someone else gonna spend a few hours unnecessarily later on trying to debug it.

barracuda156: @mflatt I have added a comment. Could you take a look if this is okay?

mflatt:
Thanks!

PR #5026, opened 2024-06-27
Set up `object%` in a way that cooperates with definition pruning.

samth:
With this change, using the demodularizer on

```
#lang racket "hello"
```

produces a zo file that's only 6.6k.


mflatt:
LGTM

PR #5027, opened 2024-06-28
Fix regexps docs

gus-massa:
Fix minor typos in the `bstr` parameter name

##### Checklist
- [X] documentation

---

Side questions:

One of the typos was in `regexp-quote` and I got confused. Which locale is used in `regexp-quote` when it compare bytes in case insensitive?

```
(regexp-match (regexp-quote #"Á" #f) #"á") ; ==> (#"\341")
(regexp-match (regexp-quote #"á" #f) #"Á") ; ==> (#"\301")
(regexp-match (regexp-quote #"A" #f) #"a") ; ==> (#"a")
(regexp-match (regexp-quote #"a" #f) #"A") ; ==> (#"A")
```


sorawee:
I don't know an answer to your question. But the fix looks good to me, so I will merge it now.

PR #5028, opened 2024-06-30
openssl: delay evaluation to prevent error within sandbox

sorawee:
Currently:

    (require racket/sandbox)
    (define eval (make-evaluator 'racket))
    (eval '(require openssl))

fails on Linux with the error:

    file-exists?: `exists' access denied for /usr/lib/ssl/cert.pem
      context...:
       /usr/share/racket/collects/openssl/mzssl.rkt:244:0: x509-root-sources

This is because the `ssl-default-verify-sources` parameter eagerly calls `file-exists?` on paths without the read permission.

This PR delays the parameter initialization so that `file-exists?` is not called eagerly.

### DIFF
--- a/racket/collects/openssl/mzssl.rkt
+++ b/racket/collects/openssl/mzssl.rkt
@@ -36,6 +36,7 @@ TO DO:
          racket/string
          racket/lazy-require
          racket/include
+         racket/promise
          "libcrypto.rkt"
          "libssl.rkt"
          "private/ffi.rkt"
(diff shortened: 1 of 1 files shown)


sorawee:
This PR is untested, because I don't have a Linux machine. Can anyone with a machine help me testing it?

samth:
The example you showed works with this change, and not without it.

One test fails for me in the `tests/openssl` collection both with and without the patch:

```
raco test: "./test-keylog.rkt"
raco test: @(test-responsible 'mflatt)
--------------------
server context retains keylogger
ERROR
name:       check-equal?
location:   test-keylog.rkt:115:6

read-bytes: SSL read failed (error:0A000126:SSL routines::unexpected eof while reading)
--------------------
```

sorawee:
Thank you! I will consider the failing test outside of the scope of this PR then. Would be good to fix it eventually though.

sorawee:
@samth @rmculpepper @mflatt OK to merge?

samth:
This is ok by me

PR #5030, opened 2024-07-02
Fix issue with `raco test` handling of special path strings

NoahStoryM:
##### Checklist

- [x] Bugfix
### Description of change

Some files (e.g., `𝒞.rkt`) were incorrectly handled by `raco test`, because the `normalize-module-path` function converts any path to a string directly:
```
𝒞.rkt: raco test: non-zero exit: 1
raco test: "𝒞.rkt"
dynamic-require: contract violation
  expected: (or/c module-path? module-path-index? resolved-module-path?)
  given: "𝒞.rkt"
  context...:
   /usr/share/racket/pkgs/compiler-lib/compiler/commands/test.rkt:92:2
   body of (submod "/usr/share/racket/pkgs/compiler-lib/compiler/commands/test.rkt" process)
   body of top-level
```
I have updated the `normalize-module-path` function so that paths are wrapped with `(file ...)` to explicitly mark them as file paths.


benknoble:
Related comments: https://github.com/racket/racket/pull/4997#discussion_r1606110404

pkgs/compiler-lib/compiler/commands/test.rkt
```diff
@@ -517,11 +518,10 @@
           v))]))
 
 (define (normalize-module-path p)
-  (cond
-   [(path? p) (path->string p)]
-   [(and (pair? p) (eq? 'submod (car p)))
-    (list* 'submod (normalize-module-path (cadr p)) (cddr p))]
-   [else p]))
+  (match p
+    [(? path?) `(file ,(path->string p))]
```
samth, line 522: Do we really want to convert this to a string?

NoahStoryM: `normalize-module-path` currently serves two purposes:

1. Printing the path as a string for logging or display purposes.
2. Processing the path to use it as an argument in subprocess calls.

Using `serialize` instead of `normalize-module-path` might be more appropriate for handling `rt-module` and `test-module`. If so, is it necessary to recover `normalize-module-path`?

mflatt:
Thanks! Separating the tasks of serialization and reporting in a human-readable form is a good move, and this looks like a good improvement overall. I'll wait a day or so to give others time to check before merging, though.

pkgs/compiler-lib/compiler/commands/test.rkt
benknoble: The problem with (2) from the comments I linked is that paths don't round-trip through strings in the general case (I forget what change I made for that though: it's probably in the PR comments I linked somewhere).

samth:
This change seems to have broken something about timeouts. See http://drdr.racket-lang.org/67241/racket/share/pkgs/typed-racket-test/external/tr-random-testing.rkt for example, where the test is killed after 90 seconds even though https://github.com/racket/typed-racket/blob/master/typed-racket-test/external/tr-random-testing.rkt#L404-L408 configures it to have a 600 second timeout. Since this is configured using submodules, and the PR changes some submodule-handling code, I'm suspicious that's related.

PR #5034, opened 2024-07-07
fix various bugs in syntax/flatten-begin

sorawee:
- Raise a syntax error (instead of a contract error) for (flatten-begin #'())
- Make (flatten-begin #'(1 2)) a bad syntax error instead of an internal error on syntax-track-origin
- For consistency with #'(1 2), make (flatten-begin #'(1)) a syntax error instead of producing an empty list
- Make (flatten-all-begins #'(begin . 1)) a syntax error, instead of producing (list #'(begin . 1))
- Make (flatten-all-begins #'(begin (1))) produce (list #'(1)), instead a syntax error


mflatt:
LGTM

PR #5046, opened 2024-07-27
Synonyms for raco pkg commands

benknoble:
## Checklist

- [x] Feature
- [x] documentation

## Description

This PR adds to `raco pkg` the synonym `uninstall`
(for `remove`), which is now the preferred command name.

I recommend viewing the changes in this PR with something that can ignore
whitespace changes, given the shifting indentation of a few blocks.

The first few
commits are preparation that implement a notion of aliases for `svn-style-command-line` and macros that expand to it, in addition to entirely shifting top-level definitions to the left. The last commit
second implements the actual aliasing and documents it.

### DIFF
--- a/pkgs/racket-doc/pkg/scribblings/pkg.scrbl
+++ b/pkgs/racket-doc/pkg/scribblings/pkg.scrbl
@@ -10,9 +10,9 @@
 @(define @|Planet1| @|PLaneT|)
 
 @(define package-name-chars
-   @list{@litchar{a} through @litchar{z}, 
-         @litchar{A} through @litchar{Z}, 
-         @litchar{0} through @litchar{9}, 
+   @list{@litchar{a} through @litchar{z},
+         @litchar{A} through @litchar{Z},
+         @litchar{0} through @litchar{9},
          @litchar{_}, and @litchar{-}})
 
 @(define raco-doc '(lib "scribblings/raco/raco.scrbl"))
@@ -132,7 +132,7 @@ The @tech{package source} types are:
 is the basename of the archive file. The @tech{checksum} for archive
 @filepath{f.@nonterm{ext}} is given by the file @filepath{f.@nonterm{ext}.CHECKSUM}.
 The valid archive formats
-are (currently) @filepath{.zip}, @filepath{.tar}, @filepath{.tgz}, 
+are (currently) @filepath{.zip}, @filepath{.tar}, @filepath{.tgz},
 @filepath{.tar.gz}, and
 @filepath{.plt}.
 Other than a @litchar{type} query, which affects inference as described below,
@@ -440,7 +440,7 @@ to @racket['user].
 The @exec{raco pkg} command provides package-management tools via
 sub-commands.
 
-@subcommand{@command/toc{install} @nonterm{option} ... @nonterm{pkg-source} ... 
+@subcommand{@command/toc{install} @nonterm{option} ... @nonterm{pkg-source} ...
  --- Installs the given @tech{package sources} (eliminating exact-duplicate @nonterm{pkg-source}s).
      If a given @nonterm{pkg-source} is @seclink["concept:auto"]{auto-installed} (to satisfy some other package's
      dependency), then it is promoted to explicitly installed.
@@ -454,7 +454,7 @@ sub-commands.
      only @nonterm{pkg-source} argument. See the @DFlag{clone} flag
      below for more details.
 
- The @exec{install} sub-command accepts 
+ The @exec{install} sub-command accepts
  the following @nonterm{option}s:
 
  @itemlist[
@@ -478,7 +478,7 @@ sub-commands.
 
  @item{@DFlag{deps} @nonterm{behavior} --- Selects the behavior for dependencies, where @nonterm{behavior} is one of
   @itemlist[
-   @item{@exec{fail} --- Cancels the installation if dependencies are uninstalled or version requirements are unmet. 
+   @item{@exec{fail} --- Cancels the installation if dependencies are uninstalled or version requirements are unmet.
         This behavior is the default for non-@tech{interactive mode}.}
    @item{@exec{force} --- Installs the package(s) despite missing dependencies or version requirements.
          Forcing an installation may leave package content in an inconsistent state. Implied packages
@@ -519,7 +519,7 @@ sub-commands.
         The package is identified
         as a @tech{single-collection package} or a @tech{multi-collection package} at the
         time that it is installed, and that categorization does not change even if the @schemeidfont{collection}
-        definition in @filepath{info.rkt} is changed (i.e., the package must be removed and re-installed
+        definition in @filepath{info.rkt} is changed (i.e., the package must be uninstalled and re-installed
         for the change to take effect).}
 
   @item{@DFlag{static-link} --- Implies @DFlag{link}, and also indicates that subdirectories
@@ -567,7 +567,7 @@ sub-commands.
  @item{@Flag{i} or @DFlag{installation} --- Shorthand for @exec{--scope installation}.}
  @item{@Flag{u} or @DFlag{user} --- Shorthand for @exec{--scope user}.}
  @item{@DFlag{scope-dir} @nonterm{dir} --- Select @nonterm{dir} as the @tech{package scope}.}
- 
+
  @item{@DFlag{catalog} @nonterm{catalog} --- Uses @nonterm{catalog}s instead of of the currently configured
        @tech{package catalogs}. This flag can be provided multiple times. The catalogs are tried in the order provided.}
 
@@ -640,7 +640,7 @@ sub-commands.
   @item{@DFlag{no-setup} --- Does not run @exec{raco setup} after installation. This behavior is also the case if the
         environment variable @envvar{PLT_PKG_NOSETUP} is set to any non-empty value.}
 
-  @item{@DFlag{no-docs} or @Flag{D} --- Does not render documentation during setup after installation. This flag has no effect 
+  @item{@DFlag{no-docs} or @Flag{D} --- Does not render documentation during setup after installation. This flag has no effect
         with @DFlag{no-setup}.}
 
   @item{@DFlag{recompile-only} ---Constrains @exec{raco setup} to at most recompile a module from
@@ -654,7 +654,7 @@ sub-commands.
   @item{@DFlag{batch} --- Disables @deftech{interactive mode}, suppressing potential prompts for a user
                           (e.g., about package dependencies or clone sharing).}
 
-  @item{@DFlag{no-trash} --- Refrains from moving updated or removed packages to a trash folder.}
+  @item{@DFlag{no-trash} --- Refrains from moving updated or uninstalled packages to a trash folder.}
 
   @item{@DFlag{fail-fast} --- Breaks @exec{raco setup} as soon as any error is encountered.}
  ]
@@ -671,7 +671,7 @@ sub-commands.
          #:changed "8.0.0.13" @elem{Added @litchar{git-url} as a @DFlag{type} option.}]}
 
 
-@subcommand{@command/toc{update} @nonterm{option} ... @nonterm{pkg-source} ... 
+@subcommand{@command/toc{update} @nonterm{option} ... @nonterm{pkg-source} ...
 --- Checks the specified package names for @tech{package updates} or
 replaces existing package installations with the given sources. If an
 update or replacement cannot be installed (e.g. it conflicts with
@@ -709,7 +709,7 @@ argument.
 If a @tech{package scope} is not specified, the scope is inferred from
 the given @nonterm{pkg-source}s.
 
- The @exec{update} sub-command accepts 
+ The @exec{update} sub-command accepts
  the following @nonterm{option}s:
 
  @itemlist[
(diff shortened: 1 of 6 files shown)


racket/collects/pkg/main.rkt
```diff
@@ -154,651 +154,657 @@
...
-                  (unless (or (not name) (package-source->name name))
-                    ((current-pkg-error) (format "~e is an invalid package name" name)))
-                  ;; if no sources were supplied, and `--pkgs` was not
-                  ;; explicitly specified, install the current directory
-                  ;; as a linked directory
-                  (define-values (sources a-type*)
-                    (if (and (not pkgs) (null? pkg-source))
+(define-for-syntax (aliases names command-clause)
+  (for/list ([name (in-list names)])
+    (cons name command-clause)))
```
benknoble, line 159: Should this use `map` instead of `for/list`?

racket/collects/pkg/main.rkt
```diff
@@ -154,651 +154,657 @@
...
-                  (when (and name (> (length pkg-source) 1))
-                    ((current-pkg-error) (format "the --name flag only makes sense with a single package source")))
-                  (unless (or (not name) (package-source->name name))
-                    ((current-pkg-error) (format "~e is an invalid package name" name)))
-                  ;; if no sources were supplied, and `--pkgs` was not
-                  ;; explicitly specified, install the current directory
-                  ;; as a linked directory
-                  (define-values (sources a-type*)
-                    (if (and (not pkgs) (null? pkg-source))
+(define-for-syntax (aliases names command-clause)
```
sorawee, line 157: This means the body will get duplicated, right? Is there a way to modify `commands` so that we don't need to duplicate the code?

racket/collects/pkg/main.rkt
```diff
@@ -154,651 +154,657 @@
...
+                                    #:pull-behavior pull
+                                    #:link-dirs? link-dirs?
+                                    #:infer-clone-from-dir? (not (or link static-link copy))
+                                    #:dry-run? dry-run
+                                    #:use-trash? (not no-trash)))))
+                (setup "updated" no-setup no-docs recompile-only #f setup-collects jobs))))]
+          ;; ----------------------------------------
+          #,@(aliases
+              (list #'remove #'uninstall)
+              #'["Remove packages"
```
sorawee, line 387: Should all texts that have the word "remove" get changed to "uninstall" to be symmetric with "install"?

racket/collects/pkg/main.rkt
benknoble: I think one way would be to have a procedure that is called instead, but the macro would still have to forward a lot of arguments (twice). I'll check, but I didn't see the body here being large.

Modifying commands to allow alias specs instead of make-commands might work; commands did a lot more work to expand into the (relatively undocumented?) svn-style command-line macro, and I didn't want to go too many layers deep.

racket/collects/pkg/main.rkt
benknoble: I wondered about this and one other thing: `raco pkg uninstall` suggests running `raco pkg remove --auto` (because the text is hard-coded). It might be nice to say uninstall there, too. But `remove` should probably continue to use "remove"? Unclear.

The output of `raco pkg -h` also lists both synonyms with the same help description. An alternative would be to have the synonyms list as in the docs ("synonym for X"). 

I now wonder about adding a spec for the synonyms that accepts arbitrary arguments and (somehow) re-invokes `raco pkg <original> <args>…`, but that only lets you tweak the help text unless we add a side channel for the command-name. Then again, maybe that could use the raco binding that gives you the correct command name?

benknoble:
This probably also needs a version note, which I'll get to later when we bang out the final version.

mflatt:
I agree about the mismatch between `install` and `remove`. As noted in the issue, those command names were chosen to match a prominent precedent, but I think `uninstall` is enough of an improvement that we could switch to that terminology. The word `add` (like `remove`) can mean too many different things, and I would not want to encourage it over `install`.

The best solution in my mind is

* Change `remove` to `uninstall` and adjust messages accordingly.

* Create `remove` as an alias, and make it documented and behave as such. Maybe `raco pkg remove` should just print "Redirecting to `raco pkg uninstall`" and then deliver all arguments to `raco pkg uninstall` for any checking that is needed. This kind of alias is probably best implemented by pushing the concept through a few layers, but that seems worthwhile to me if it's the behavior that we want.

benknoble:
I've pushed a new version which I think addresses most of the feedback; I'll appreciate extra eyes on the docs and macros.

range-diff:
```
-:  ---------- > 1:  c259e01d05 prepare: make svn-style-command-line cases call procedures
-:  ---------- > 2:  992a255e50 prepare: support #:alias declarations in svn-style-command-line
-:  ---------- > 3:  e9149630dd prepare: push alias support into pkg/commands
1:  704f445713 ! 4:  ece7456fa7 pkg: unindent top-level forms
    @@ Metadata
     Author: D. Ben Knoble <ben.knoble+github@gmail.com>
     
      ## Commit message ##
    -    pkg: unindent top-level forms
    +    pkg: dedent top-level forms
     
         Best-viewed-with: --ignore-space-change
     
2:  c625521d01 < -:  ---------- raco pkg: add synonyms for install, remove
-:  ---------- > 5:  4dfc5aaca9 raco pkg: rename remove to uninstall
```

racket/collects/pkg/main.rkt
benknoble: This is handled by the new version.

racket/collects/pkg/main.rkt
benknoble: Also handled by the new version.

racket/collects/pkg/main.rkt
benknoble: Removed entirely.

benknoble:
This appears not quite ready, however, because `remove` (and presumably other aliases) don't handle unknown flags correctly.

benknoble:
> This appears not quite ready, however, because `remove` (and presumably other aliases) don't handle unknown flags correctly.

Note to self: a likely way to make this work is to insert `--` before parsing for the alias, which should then remove it when pushing to the next parser.

benknoble:
I've added the proper support for arbitrary flags; for example, `raco pkg install jsond && raco pkg remove --auto jsond` still works at this PR's tip (as does `… && raco pkg uninstall --auto jsond`), where `jsond` was chosen as a small package with few dependencies to make the test fast to run.

<details>
<summary>range-diff</summary>

```
1:  c259e01d05 = 1:  c259e01d05 prepare: make svn-style-command-line cases call procedures
2:  992a255e50 ! 2:  ebddf6625c prepare: support #:alias declarations in svn-style-command-line
    @@ Commit message
           and the impl procedures can refer to each other.
         - The alias description, being computed at run time, must be unquoted to
           appear in `((name description) ...).
    +    - To allow the aliasing command to accept arbitrary arguments, prepend
    +      "--" to the arguments it parses. This marker is stripped by
    +      command-line when handing off to the formals, so no cdr is needed when
    +      invoking the aliased command. To support this change, each clause must
    +      be allowed to determine an argv expression that depends on the name of
    +      the input parameter (for most clauses this is an identity to preserve
    +      backwards compatibility).
     
         Also refactor (lambda (l) (display l) (newline)) to displayln.
     
    @@ racket/collects/planet/private/command.rkt: (define-syntax (svn-style-command-li
     -         #'(let* ([p prog]
     +     (with-syntax* ([impl-table (generate-temporary #'impl-table)]
     +                    [(((name description long-description body ...)
    -+                       accum formals arg-help-strs final-expr)
    ++                       accum formals arg-help-strs final-expr
    ++                       input argv)
     +                      ...)
     +                     (map (lambda (clause)
     +                            (syntax-case clause ()
    @@ racket/collects/planet/private/command.rkt: (define-syntax (svn-style-command-li
     +                                  ignored remainder '("remainder")
     +                                  (begin
     +                                    (eprintf "Redirecting to `~a ~a`\n" p invoke)
    -+                                    ((hash-ref impl-table invoke) remainder)))]
    ++                                    ((hash-ref impl-table invoke) remainder))
    ++                                  remainder (cons "--" remainder))]
     +                              [[name description long-description body ...
     +                                     #:args formals final-expr]
     +                               #'((name description long-description body ...)
    -+                                  ignored formals (pimap symbol->string 'formals) final-expr)]
    ++                                  ignored formals (pimap symbol->string 'formals) final-expr
    ++                                  remainder remainder)]
     +                              [(name description long-description body ...
     +                                     #:handlers (lambda (accum . formals) final-expr) arg-help-strs)
     +                               #'((name description long-description body ...)
    -+                                  accum formals arg-help-strs final-expr)]))
    ++                                  accum formals arg-help-strs final-expr
    ++                                  remainder remainder)]))
     +                          (syntax->list #'(clause ...)))]
     +                    [(n ...) (generate-temporaries #'(name ...))]
     +                    [(impl ...) (generate-temporaries #'(name ...))])
    @@ racket/collects/planet/private/command.rkt: (define-syntax (svn-style-command-li
                        [a args]
                        [n name] ...
                        [impl
    -@@ racket/collects/planet/private/command.rkt: (define-syntax (svn-style-command-line stx)
    +-                   (λ (remainder)
    ++                   (λ (input)
    +                      (parameterize ([current-svn-style-command n])
    +                        (command-line
    +                         #:program (format "~a ~a" p n)
    +-                        #:argv remainder
    ++                        #:argv argv
    +                         body ...
    +                         #:handlers
                              (λ (accum . formals) final-expr)
                              arg-help-strs
                              (λ (help-string)
3:  e9149630dd = 3:  aa1a8c6fee prepare: push alias support into pkg/commands
4:  ece7456fa7 = 4:  abfc30132b pkg: dedent top-level forms
5:  30e7575bf5 = 5:  54c51a4a54 raco pkg: rename remove to uninstall
-:  ---------- > 6:  46965135e1 pkg uninstall: update user messages for new command name
```

</details>

pkgs/racket-doc/pkg/scribblings/pkg.scrbl
```diff
@@ -840,8 +840,12 @@ the given @nonterm{pkg}s.
          #:changed "6.1.1.6" @elem{Added the @DFlag{no-trash} flag.}
          #:changed "6.4.0.14" @elem{Added the @DFlag{dry-run} flag.}
          #:changed "7.2.0.8" @elem{Added the @DFlag{recompile-only} flag.}
-         #:changed "7.4.0.4" @elem{Added the @DFlag{no-docs}, @Flag{D} flags.}]}
+         #:changed "7.4.0.4" @elem{Added the @DFlag{no-docs}, @Flag{D} flags.}
+         #:changed "8.14.0.2" @elem{Renamed from @command-ref{remove} to @command-ref{uninstall}.}]}
+
+@subcommand{@command/toc{remove}---A synonym for @command-ref{uninstall}.
```
mflatt, line 846: Nit: Although it may be typographically an abuse of em dashes, the convention overall here is to have spaces around `---`.

mflatt (changes requested):
Looks good to me, with a very small change request. Assuming it's ok, you can squash the commits, or I can squash when merging.

benknoble:
> Looks good to me, with a very small change request. 

I can make that change.

> Assuming it's ok, you can squash the commits, or I can squash when merging.

Is there a specific set of commits you'd like squashed? I'm happy to meet Racket's standard for commits.

I went to a bit of trouble to keep the commits self-contained and well-described, though I'll admit the last one about messages could probably go with the other changes, and the prepare commits might better describe what they prepare for. 

I'm not a fan of the [information wipeout that squashing incurs](https://benknoble.github.io/blog/2024/08/02/github-squash/), at least when the commits are already full of useful info :)

mflatt:
Ah, nevermind about squashing. I should have looked more at the individuals commits. Squashing is a default, but since you created the commits to stay separate, I'm happy to keep them that way.

pkgs/racket-doc/pkg/scribblings/pkg.scrbl
benknoble: Fixed! The range-diff is trivial but included:
```
1:  c259e01d05 = 1:  c259e01d05 prepare: make svn-style-command-line cases call procedures
2:  ebddf6625c = 2:  ebddf6625c prepare: support #:alias declarations in svn-style-command-line
3:  aa1a8c6fee = 3:  aa1a8c6fee prepare: push alias support into pkg/commands
4:  abfc30132b = 4:  abfc30132b pkg: dedent top-level forms
5:  54c51a4a54 ! 5:  81355921f0 raco pkg: rename remove to uninstall
    @@ pkgs/racket-doc/pkg/scribblings/pkg.scrbl: the given @nonterm{pkg}s.
     +         #:changed "7.4.0.4" @elem{Added the @DFlag{no-docs}, @Flag{D} flags.}
     +         #:changed "8.14.0.2" @elem{Renamed from @command-ref{remove} to @command-ref{uninstall}.}]}
     +
    -+@subcommand{@command/toc{remove}---A synonym for @command-ref{uninstall}.
    ++@subcommand{@command/toc{remove} --- A synonym for @command-ref{uninstall}.
      
     +@history[#:changed "8.14.0.2" @elem{Made @command-ref{remove} an alias.}]}
      
6:  46965135e1 = 6:  853f171b91 pkg uninstall: update user messages for new command name
```

mflatt:
Thanks!

mflatt:
I'll update the test.

benknoble:
Looks like a check for the specific advice needs changed? Or the advice change could be reverted.

benknoble:
mflatt fixed it in 1ecba9f, which also points me to all the test files I missed. Thanks!

PR #5047, opened 2024-07-27
glob: clarify that only path values are returned

benknoble:
## Description of change

Summarizing [1], experimental evidence suggested that `glob` and cousins
only returned values matching `path?` and never `string?`. Digging into
the source, glob depends on `in-directory` from
racket/collects/racket/private/for.rkt, which sorts the results
`directory-list` by `path<?`. So it certainly appears that `glob` only
returns path values. Correct both documentation and contract.

Also update `in-directory` documentation to note the kind of values
returned by its sequence (it has no corresponding source contract).

Also cleanup adjacent whitespace.

[1]: https://racket.discourse.group/t/does-glob-ever-return-strings-if-not-should-we-update-the-docs/3049/2

Best-viewed-with: --ignore-space-change

## Checklist

- [x] Bugfix (?)
- [x] documentation

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/sequences.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/sequences.scrbl
@@ -595,7 +595,7 @@ each element in the sequence.
 @defproc[(in-directory [dir (or/c #f path-string?) #f]
                        [use-dir? ((and/c path? complete-path?) . -> . any/c)
                                  (lambda (dir-path) #t)])
-         sequence?]{
+         (sequence/c path?)]{
   Returns a sequence that produces all of the paths for files,
   directories, and links within @racket[dir], except for the
   contents of any directory for which @racket[use-dir?] returns
--- a/pkgs/racket-doc/file/scribblings/glob.scrbl
+++ b/pkgs/racket-doc/file/scribblings/glob.scrbl
@@ -64,7 +64,7 @@ has the same meaning as @racket['("foo.rkt" "bar.rkt")].
 }
 ]
 
-@defproc[(glob [pattern glob/c] [#:capture-dotfiles? capture-dotfiles? boolean? (glob-capture-dotfiles?)]) (listof path-string?)]{
+@defproc[(glob [pattern glob/c] [#:capture-dotfiles? capture-dotfiles? boolean? (glob-capture-dotfiles?)]) (listof path?)]{
   Builds a list of all paths on the current filesystem that match any glob
   in @racket[pattern]. The order of paths in the result is unspecified.
 
@@ -100,7 +100,7 @@ Examples:
 }
 }
 
-@defproc[(in-glob [pattern glob/c] [#:capture-dotfiles? capture-dotfiles? boolean? (glob-capture-dotfiles?)]) (sequence/c path-string?)]{
+@defproc[(in-glob [pattern glob/c] [#:capture-dotfiles? capture-dotfiles? boolean? (glob-capture-dotfiles?)]) (sequence/c path?)]{
   Returns a stream of all paths matching the glob @racket[pattern],
   instead of eagerly building a list.
 }
--- a/racket/collects/file/glob.rkt
+++ b/racket/collects/file/glob.rkt
@@ -8,8 +8,8 @@
 (provide
   glob/c
   (contract-out
-    [glob (->* [glob/c] [#:capture-dotfiles? boolean?] (listof path-string?))]
-    [in-glob (->* [glob/c] [#:capture-dotfiles? boolean?] (sequence/c path-string?))]
+    [glob (->* [glob/c] [#:capture-dotfiles? boolean?] (listof path?))]
+    [in-glob (->* [glob/c] [#:capture-dotfiles? boolean?] (sequence/c path?))]
     [glob-match? (->* [glob/c path-string?] [#:capture-dotfiles? boolean?] boolean?)]
     [glob-quote (->i ([ps path-string?]) [r (ps) (if (path? ps) path? string?)])]
     [glob-capture-dotfiles? (parameter/c boolean?)]))


pkgs/racket-doc/scribblings/reference/sequences.scrbl
```diff
@@ -595,7 +595,7 @@ each element in the sequence.
 @defproc[(in-directory [dir (or/c #f path-string?) #f]
                        [use-dir? ((and/c path? complete-path?) . -> . any/c)
                                  (lambda (dir-path) #t)])
-         sequence?]{
+         (sequence/c path?)]{
```
benknoble, line 598: Should these all be `path-for-some-system?` instead?

sorawee:
The "co-authored-by" in the commit message should be greghendershott.com, right?

benknoble:
Well, that solves that mystery 😅

shhyou:
Is it okay to omit the `racket/private/for` changes? I only see changes to the white spaces, or am I missing anything?

benknoble:
> Is it okay to omit the `racket/private/for` changes? I only see changes to the white spaces, or am I missing anything?

"Also cleanup adjacent whitespace" was meant to refer to that. It could be dropped if desired.

sorawee:
Cleaning adjacent whitespace would certainly be great. I think the issue that @shhyou raised is that the change in for.rkt is not *adjacent* whitespace clean up. The file has no other change besides whitespace clean up. 

If it's easy for you to do, let's leave that file out, to avoid complicating git blame.

pkgs/racket-doc/scribblings/reference/sequences.scrbl
mflatt: I think `path?` is right everywhere, since the paths come from the current filesystem, so they're definitely for the current platform.

benknoble:
> Cleaning adjacent whitespace would certainly be great. I think the issue that @shhyou raised is that the change in for.rkt is not _adjacent_ whitespace clean up. The file has no other change besides whitespace clean up.

Fair enough. It was adjacent only in that I'd gone poking around there while researching the code. 

> If it's easy for you to do, let's leave that file out, to avoid complicating git blame.

It is easy (`git restore -s @~ racket/collects/racket/private/for.rkt` and amend the commit, which I've done and pushed), but it doesn't hurt `git blame`: `git blame -w racket/collects/racket/private/for.rkt | grep Knoble` doesn't get any hits while 7f7bc52ad5f97c47d1274009195718a3ec9c4285 is checked out.

---

Here's the range-diff:
```
1:  7f7bc52ad5 ! 1:  b12b22118d glob: clarify that only path values are returned
    @@ Commit message
         Also update `in-directory` documentation to note the kind of values
         returned by its sequence (it has no corresponding source contract).
     
    -    Also cleanup adjacent whitespace.
    -
         [1]: https://racket.discourse.group/t/does-glob-ever-return-strings-if-not-should-we-update-the-docs/3049/2
     
    -    Best-viewed-with: --ignore-space-change
    -
      ## pkgs/racket-doc/file/scribblings/glob.scrbl ##
     @@ pkgs/racket-doc/file/scribblings/glob.scrbl: has the same meaning as @racket['("foo.rkt" "bar.rkt")].
      }
    @@ racket/collects/file/glob.rkt
          [glob-match? (->* [glob/c path-string?] [#:capture-dotfiles? boolean?] boolean?)]
          [glob-quote (->i ([ps path-string?]) [r (ps) (if (path? ps) path? string?)])]
          [glob-capture-dotfiles? (parameter/c boolean?)]))
    -
    - ## racket/collects/racket/private/for.rkt ##
    -@@ racket/collects/racket/private/for.rkt: (module for '#%kernel
    -              (rename *in-port in-port)
    -              (rename *in-lines in-lines)
    -              (rename *in-bytes-lines in-bytes-lines)
    --             
    -+
    -              in-hash
    -              in-hash-keys
    -              in-hash-values
    -@@ racket/collects/racket/private/for.rkt: (module* expand #f
    -   ;; redefininition of functions not in #%kernel
    -   (begin-for-syntax
    -     (define (format-id ctx str . args)
    --      (define datum 
    -+      (define datum
    -         (string->symbol (apply format str (map syntax->datum args))))
    -       (datum->syntax ctx datum))
    -     (define (join-ids ids sep) ; joins ids with sep; ids = stx-pair
    -@@ racket/collects/racket/private/for.rkt: (define (check-stream l)
    -   (define (in-stream l)
    -     (check-stream l)
    -     (make-do-sequence (lambda () (:stream-gen l))))
    --  
    -+
    -   (define (:stream-gen l)
    --    (values 
    -+    (values
    -      unsafe-stream-first unsafe-stream-rest values l unsafe-stream-not-empty? #f #f))
    - 
    - 
    -   ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
    -   ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
    -   ;; hash sequences
    --  
    -+
    -   ;; assembles hash iterator functions to give to make-do-sequence
    -   (define :hash-gen
    -     (case-lambda
    -@@ racket/collects/racket/private/for.rkt: (define-syntax (define-in-hash-sequences stx)
    -      [(_ element-types: V ...)
    -       (with-syntax
    -        ([VAL (join-ids #'(V ...) #'+)])
    --       (with-syntax 
    -+       (with-syntax
    -         ([IN-HASH-DEFINER (format-id #'VAL "define-in-hash-~as-seq" #'VAL)])
    -         #'(begin
    -            ;; 1) define sequence syntax definer
    -@@ racket/collects/racket/private/for.rkt: (define-syntax (IN-HASH-DEFINER stx)
    -                 [HASHTYPE? #'(lambda (ht) (and (hash? ht) (p? ht) (... ...)))]
    -                 [ERR-STR
    -                  (datum->syntax #'HASHTYPE
    --                  (if (null? (syntax->list #'(p? (... ...)))) 
    -+                  (if (null? (syntax->list #'(p? (... ...))))
    -                       "hash?"
    --                      (string-append 
    -+                      (string-append
    -                         "(and/c hash? "
    --                        (symbol->string 
    --                          (syntax->datum (join-ids #'(p? (... ...)) #'" "))) 
    -+                        (symbol->string
    -+                          (syntax->datum (join-ids #'(p? (... ...)) #'" ")))
    -                         ")")))])
    -                (with-syntax
    -                 ([-first (format-id #'PREFIX "~a-first" #'PREFIX)]
    -@@ racket/collects/racket/private/for.rkt: (define (check-ranges who type-name vec start stop step len)
    -     (unless (and (exact-integer? step) (not (zero? step)))
    -       (raise-argument-error who "(and/c exact-integer? (not/c zero?))" step))
    -     (when (and (< start stop) (< step 0))
    --      (raise-arguments-error who 
    -+      (raise-arguments-error who
    -                              "starting index less than stopping index, but given a negative step"
    -                              "starting index" start
    -                              "stopping index" stop
    -                              "step" step))
    -     (when (and (< stop start) (> step 0))
    --      (raise-arguments-error who 
    -+      (raise-arguments-error who
    -                              "starting index more than stopping index, but given a positive step"
    -                              "starting index" start
    -                              "stopping index" stop
    -@@ racket/collects/racket/private/for.rkt: (define-for-syntax ((make-inner-recur/foldr/lazy fold-vars delayed-id delayer-id
    -                          ...)]
    -                      [else #'()]))
    -              expr ...))]))
    --  
    -+
    -   (define-syntax (push-under-break stx)
    -     (syntax-case stx ()
    -       [(_ inner-recur fold-bind [expr ...] next-k break-k final?-id)
    -        (let loop ([l (syntax->list #'(expr ...))] [pre-accum null])
    -          (cond
    --          [(null? l) 
    -+          [(null? l)
    -            ;; No #:break form
    -            #'(inner-recur fold-bind [expr ...] (if final?-id break-k next-k))]
    -           [(eq? '#:break (syntax-e (car l)))
    -@@ racket/collects/racket/private/for.rkt: (define-for-syntax (split-for-body stx body-stx)
    -                   (list (reverse pre-kw) (reverse post-kw)))]
    -              [(memq (syntax-e (car exprs)) '(#:break #:final))
    -               (if (pair? (cdr exprs))
    --                  (loop (cddr exprs) 
    -+                  (loop (cddr exprs)
    -                         (append (list* (cadr exprs) (car exprs) post-kw)
    -                                 pre-kw)
    -                         null)
    -@@ racket/collects/racket/private/for.rkt: (define wrap? (and combine*
    -         ;; Let `derived-id-stx' complain about the missing bindings and body expression:
    -         [(_ . rest)
    -          #`(#,derived-id-stx #,stx fold-bind . rest)])))
    --  
    -+
    -   (define-syntax define-syntax-via-derived
    -     (syntax-rules ()
    -       [(_ id derived-id nonnested-id fold-bind wrap rhs-wrap combine combine*)
    -@@ racket/collects/racket/private/for.rkt: (define-for-syntax (for_/vector stx orig-stx for_/vector-stx for_/fold/derived-s
    -                            orig-stx
    -                            ([vec (make-vector 16)]
    -                             [i 0])
    --                           (for-clause ...) 
    -+                           (for-clause ...)
    -                            middle-body ...
    -                            (let ([new-vec (if (eq? i (unsafe-vector*-length vec))
    -                                               (grow-vector vec)
    -@@ racket/collects/racket/private/for.rkt: (define wrapped-fc
    -               (let ([v (make-vector len fill-expr)])
    -                 (unless (zero? len)
    -                   (for_/fold/derived
    --                   orig-stx 
    -+                   orig-stx
    -                    ([i 0])
    -                    (limited-for-clause ...)
    -                    middle-body ...
    -@@ racket/collects/racket/private/for.rkt: (define wrapped-fc
    -                    (unsafe-fx+ 1 i)))
    -                 v))))]
    -       [(_ #:length length-expr (for-clause ...) body ...)
    --       (for_/vector #'(fv #:length length-expr #:fill 0 (for-clause ...) body ...) 
    -+       (for_/vector #'(fv #:length length-expr #:fill 0 (for-clause ...) body ...)
    -                     orig-stx for_/vector-stx for_/fold/derived-stx wrap-all?)]))
    - 
    -   (define-syntax (for/vector stx)
    -@@ racket/collects/racket/private/for.rkt: (define-sequence-syntax *in-input-port-chars
    - 
    -   (define (dir-list full-d d acc)
    -     (for/fold ([acc acc]) ([f (in-list (reverse (sort (directory-list full-d) path<?)))])
    --	      (cons (build-path d f) acc)))
    -+      (cons (build-path d f) acc)))
    - 
    -   (define (next-body l d init-dir use-dir?)
    -     (let ([full-d (path->complete-path d init-dir)])
    -@@ racket/collects/racket/private/for.rkt: (define (initial-state orig-dir init-dir)
    - 	(sort (directory-list init-dir) path<?)))
    - 
    -   (define in-directory
    --    (case-lambda 
    -+    (case-lambda
    -      [() (in-directory #f (lambda (d) #t))]
    -      [(orig-dir) (in-directory orig-dir (lambda (d) #t))]
    -      [(orig-dir use-dir?)
    -@@ racket/collects/racket/private/for.rkt: (define d (car l))
    - 	  pair?
    - 	  #f
    - 	  #f)))]))
    --     
    -+
    -   (define-sequence-syntax *in-directory
    -     (λ () #'in-directory)
    -     (λ (stx)
    -@@ racket/collects/racket/private/for.rkt: (define-sequence-syntax *in-directory
    - 	 [((d) (_)) #'[(d) (in-directory #f)]]
    - 	 [((d) (_ dir)) #'[(d) (in-directory dir (lambda (d) #t))]]
    - 	 [((d) (_ dir use-dir?-expr))
    --	  #'[(d) 
    -+	  #'[(d)
    - 	     (:do-in
    - 	      ([(orig-dir) (or dir #f)]
    -                [(init-dir) (current-directory)]
```

mflatt:
Thanks!

PR #5048, opened 2024-07-28
net/platform-ssl: avoid cross-system-type in stub-replacement macro

Bogdanp:
As suggested by @rmculpepper, this moves the stub-replacement dispatch to runtime to fix an issue with code distributed in "built" form. The difference in memory use between the two approaches is negligible.

See also https://github.com/racket/racket/pull/4884#issuecomment-2254275684.


mflatt:
Thanks!

PR #5052, opened 2024-08-05
doc: add an example for make-parameter

sorawee:
Related to https://racket.discourse.group/t/parameterize-not-working-as-expected-in-8-13/3071

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/parameters.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/parameters.scrbl
@@ -50,6 +50,22 @@ applied to the initial @racket[v].
 The @racket[name] argument is used as the parameter procedure's name
 as reported by @racket[object-name].
 
+@examples[
+(define p (make-parameter 1))
+(p)
+(p 3)
+(p)
+(define q (make-parameter "a"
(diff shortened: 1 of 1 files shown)


mflatt:
This looks helpful, thanks! I think I'd be inclined to use `parameterize` for most examples, instead of emphasizing imperative use of parameters, but you may have already considered the trade-offs there.

sorawee:
Yeah, I intentionally didn't use `parameterize` here because there are already a lot of clear examples of `parameterize`  under `parameterize`.

We could demonstrate how `parameterize` works here as well, but then I think it would be weird to have examples in both `make-parameter` and `parameterize` being similar.

However, I understand your concern that people might see the example, and thought that the imperative use is the way to use parameters. Maybe we can improve the prose above at the beginning of the section to make clear that `parameterize` is preferred?

rfindler:
My take is that people will skim the prose miss that point and think
imperatively updating parameters is the way to go. Updating the examples to
use parameterize first and then have some examples to demonstrate the guard
and imperatively setting the parameter (with comments explaining the two
gotchas maybe) is the way to go.

Robby

PR #5057, opened 2024-08-16
cs,iOS: add iOS-specific machine type and make OpenSSL work

Bogdanp:
With this change, PB iOS builds report `(system-type)` as `unix` and `(system-type 'os*)` as `ios`. The former could be made to report `macosx` instead, but then more code would have to change and I'm not sure pretending to be macOS is a good idea in the long term.

### DIFF
--- a/.github/workflows/ci-pr.yml
+++ b/.github/workflows/ci-pr.yml
@@ -101,3 +101,53 @@ jobs:
           --enable-scheme=${{ github.workspace }}/host-racket/src/build/cs/c
         make
         make install
+
+  build-pb-ios:
+    runs-on: macos-14
+    steps:
+    - uses: actions/checkout@v4
+    - name: Build LibFFI
+      run: |
+        set -euxo pipefail
+        brew install automake libtool
+        git clone https://github.com/libffi/libffi
+        cd libffi
+        git checkout v3.4.6
+        ./autogen.sh
+        python generate-darwin-source-and-headers.py --only-ios
+        xcodebuild \
+          -configuration release \
+          -target libffi-iOS \
+          -scheme libffi-iOS \
+          -sdk "$(xcode-select -p)/Platforms/iPhoneOS.platform/Developer/SDKs/iPhoneOS.sdk" \
+          -derivedDataPath dist \
+          IPHONEOS_DEPLOYMENT_TARGET=12
+    - name: Build PB iOS Racket
+      run: |
+        set -euxo pipefail
+        make fetch-pb
+        mkdir -p racket/src/build
+        cd racket/src/build
+        cat >libffi.pc <<EOF
+          prefix=${GITHUB_WORKSPACE}/libffi/dist/Build/Products/Release-iphoneos
+          exec_prefix=\${prefix}
+          libdir=\${exec_prefix}
+          toolexeclibdir=\${libdir}
+          includedir=\${prefix}/include/ffi
+
+          Name: libffi
+          Description: Library supporting Foreign Function Interfaces
+          Version: 3.4.6
+          Libs: -L\${toolexeclibdir} -lffi
+          Cflags: -I\${includedir}
+        EOF
+        cat libffi.pc
+        env PKG_CONFIG_PATH="$(pwd)" \
+          ../configure \
+            --host=aarch64-apple-darwin \
+            --enable-ios=iPhoneOS \
+            --enable-pb \
+            --enable-racket=auto \
+            --enable-libffi
+        make
+        make install
--- a/racket/src/cs/rumble/system.ss
+++ b/racket/src/cs/rumble/system.ss
@@ -36,6 +36,7 @@
 
 (define os*-symbol
   (case (reflect-machine-type)
+    [(a6ios ta6ios arm64ios tarm64ios) 'ios]
     [(a6osx ta6osx
             i3osx ti3osx
             arm64osx tarm64osx
@@ -74,6 +75,10 @@
 
 (define arch-symbol
   (case (reflect-machine-type)
+    [(a6ios ta6ios)
+     'x86_64]
+    [(arm64ios tarm64ios)
+     'aarch64]
     [(a6osx ta6osx
             a6nt ta6nt
             a6le ta6le
@@ -122,6 +127,8 @@
 
 (define link-symbol
   (case (reflect-machine-type)
+    [(a6ios ta6ios arm64ios tarm64ios)
+     'framework]
     [(a6osx ta6osx i3osx ti3osx arm64osx tarm64osx)
      (if unix-style-macos?
          'static
@@ -133,6 +140,7 @@
 
 (define so-suffix-bytes
   (case (reflect-machine-type)
+    [(a6ios ta6ios arm64ios tarm64ios) (string->utf8 ".dylib")]
     [(a6osx ta6osx i3osx ti3osx arm64osx tarm64osx ppc32osx tppc32osx) (string->utf8 ".dylib")]
     [(a6nt ta6nt i3nt ti3nt arm64nt tarm64nt) (string->utf8 ".dll")]
     [else (string->utf8 ".so")]))
--- a/racket/src/cs/rumble/system.ss
+++ b/racket/src/cs/rumble/system.ss
@@ -29,14 +29,17 @@
 
 (define os-symbol
   (case (reflect-machine-type)
-    [(a6osx ta6osx i3osx ti3osx arm64osx tarm64osx ppc32osx tppc32osx)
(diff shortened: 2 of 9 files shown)


mflatt:
I think the intent is that this mode works with `--enable-pb`, and the "build-ios" CI task failed because it doesn't use that flag. Is that right? 

My next thought that was the `configure` changes should only apply when `--enable-pb` is supplied. But then I worry about the mismatch between `(system-type)` for iOS without `--enable-pb` and with `--enable-pb`. So, maybe even trigger the change only with a combination of `--enable-pb` and `--enable-xox`? Or maybe, given that `--enable-xox` is an increasingly anachronistic name anyway, some new configuration flag? Or would it be better to make `(system-type)` report `'unix` without `--enable-pb`?

Bogdanp:
> I think the intent is that this mode works with --enable-pb, and the "build-ios" CI task failed because it doesn't use that flag. Is that right?

That's right, although I didn't realize this change would break regular iOS builds in this way.

> My next thought that was the configure changes should only apply when --enable-pb is supplied. But then I worry about the mismatch between (system-type) for iOS without --enable-pb and with --enable-pb. So, maybe even trigger the change only with a combination of --enable-pb and --enable-xox? Or maybe, given that --enable-xox is an increasingly anachronistic name anyway, some new configuration flag? Or would it be better to make (system-type) report 'unix without --enable-pb?

My vote would be to make `(system-type)` report `unix` in all configurations on iOS. Looking at some of the places where core libraries dispatch on a `macosx` `(system-type)`, my impression is that some of them would do the wrong thing on iOS anyway (eg. the openssl package in this change -- I don't think it will ever be possible to load a `libcrypto.dylib` on a non-jailbroken iOS like it is on macOS), and others are unnecessary on iOS (eg. creating the `GRacket.app` launcher). I think it would also be strange to have the non-PB iOS builds report an `'os*` of `macosx` while the PB builds report `ios`, so I'll try to make the existing iOS build pass with this change over the weekend -- it looks like all it might need is a machine definition for `tarm64ios`.

Bogdanp:
I think the latest commit should make the build pass, though I guess I'll have to make that change against the chez repo instead. 

Re. `(system-type)`, I'll take some time this weekend and experiment with having the `iOS` builds report `macosx` in case that's better after all, so I'm going to convert this to a draft for now.

rfindler:
FWIW, the gui libraries use system-type a bunch. I didn't try it, but just
based on my memory, it seems not very difficult to convert them for the new
result of system-type.

.github/workflows/ci-pr.yml
```diff
@@ -101,3 +101,44 @@ jobs:
...
+
+  build-pb-ios:
+    runs-on: macos-14
+    steps:
+    - uses: actions/checkout@v4
+    - name: Fetch LibFFI
+      run: |
+        set -euxo pipefail
+        curl -O https://racket.defn.io/libffi-aarch64-ios-3.4.6.tar.gz
+        echo '1a312af2ef169f40bbc9a2e9bdbd52b9e8de35e7  libffi-aarch64-ios-3.4.6.tar.gz' | shasum --check -
```
Bogdanp, line 113: I'm a bit anxious about downloading built artifacts from non-Racket infra like this, but I haven't been able to get LibFFI to cross-compile for iOS inside GH Actions. Other options here would be:

1. Defer adding this build job until I can figure out how to get LibFFI to compile in GHA.
2. Host the built archive somewhere in the Racket infra (maybe even a GH repo?).
3. Keep as-is since the checksum protects us from tampering.

Bogdanp:
After sleeping on it, I think keeping `(system-type)` as `macosx` is probably for the best, or at least neutral compared to `unix`. The cases where `macosx` would not work are just as likely not to work as `unix`, and this way the change itself is smaller (I was wrong in my original PR comment thinking that reporting `macosx` would require more changes). For the cases where `macosx` doesn't work well on iOS, we can just repair them as needed going forward (as @rfindler points out, that shouldn't be difficult in general).

If this all seems OK, I'll go ahead and make a PR to the ChezScheme repo for the chez-specific changes. I haven't done it here, but I think I'm going to have to also bump the `scheme-version`, so PB files will also have to be regenerated.

mflatt:
This looks like a fine direction to me, and I agree that making a Chez Scheme PR would be the next step. Thanks!

LiberalArtist:
This reminded me of https://github.com/racket/racket/pull/4702#discussion_r1267417037 in which I noted that BC already can return `'ios`:

https://github.com/racket/racket/blob/244535d8b41169ee69d077cd327090c12f9dd969/racket/src/bc/sconfig.h#L651-L657

(BC also has `'android` as a possibility, which obviously comes with a different set of considerations, but, to the extent that we can have a consistent rationale thinking about `(system-type 'os*)`, that seems desirable:)

https://github.com/racket/racket/blob/244535d8b41169ee69d077cd327090c12f9dd969/racket/src/bc/sconfig.h#L86-L92

racket/src/cs/rumble/system.ss
```diff
@@ -29,13 +29,17 @@
...
      (if unix-style-macos? 'unix 'macosx)]
     [(a6nt ta6nt i3nt ti3nt arm64nt tarm64nt) 'windows]
     [else 'unix]))
 
 (define os*-symbol
   (case (reflect-machine-type)
+    [(a6ios ta6ios arm64ios tarm64ios)
+     (if unix-style-macos?
+         'unix
+         'ios)]
```
LiberalArtist, line 42: At a minimum, I think this change should be consistent with BC:

https://github.com/racket/racket/blob/244535d8b41169ee69d077cd327090c12f9dd969/racket/src/bc/sconfig.h#L651-L657

IIUC the current behavior would be to return `'darwin` rather than `'unix`. (I think iOS is arguably Darwin, depending on where exactly you draw the line for "Darwin" between XNU and Mac OS.)

It might be more useful, though, to change both to return `'ios` even in `unix-style-macos?` builds: you can distinguish `unix-style-macos?` in any case with `(system-type 'os)` (no `*`), and `'ios` would let you distinguish iOS from Mac OS.

racket/src/cs/rumble/system.ss
```diff
@@ -74,6 +78,10 @@
 
 (define arch-symbol
   (case (reflect-machine-type)
+    [(a6ios ta6ios)
+     'x86_64]
+    [(arm64ios tarm64ios)
+     'aarch64]
```
LiberalArtist, line 84: I would mildly prefer adding these to the existing clauses that return `'x86_64` and `'aarch64`, respectively.

As I wrote in https://github.com/racket/racket/pull/4702 and https://github.com/racket/racket/issues/3948#issuecomment-1481803208, I think a much better solution, though out of scope for this PR, would be to push more of this work to Chez Scheme. I'm still not sure of a design I'd find fully convincing for that, though `(constant architecture)` might be enough for this specifically &hellip;

racket/src/cs/rumble/system.ss
```diff
@@ -133,6 +143,7 @@
 
 (define so-suffix-bytes
   (case (reflect-machine-type)
+    [(a6ios ta6ios arm64ios tarm64ios) (string->utf8 ".dylib")]
     [(a6osx ta6osx i3osx ti3osx arm64osx tarm64osx ppc32osx tppc32osx) (string->utf8 ".dylib")]
```
LiberalArtist, line 147: I'd merge the two `(string->utf8 ".dylib")` clauses.

racket/src/cs/rumble/system.ss
```diff
@@ -122,6 +130,8 @@
 
 (define link-symbol
   (case (reflect-machine-type)
+    [(a6ios ta6ios arm64ios tarm64ios)
+     'framework]
     [(a6osx ta6osx i3osx ti3osx arm64osx tarm64osx)
      (if unix-style-macos?
          'static
```
LiberalArtist, line 137: It would be good to check that `(system-type 'link)` as a whole is consistent with BC and with the docs in:

https://github.com/racket/racket/blob/244535d8b41169ee69d077cd327090c12f9dd969/pkgs/racket-doc/scribblings/reference/runtime.scrbl#L65-L72

The documentation for `system-type` doesn't mention the existence of `unix-style-macos?` builds at all, but the fact that it says `Mac OS` instead of `'macosx` could be interpreted to contradict the current possibility of returning `'static`. I also wonder if it would make sense to check `unix-link-shared?` there.

Bogdanp: I think it would probably be best to handle this in a separate PR.

LiberalArtist: Yes, that's probably right. I'm not even entirely sure where or how `(system-type 'link)` is used.

LiberalArtist: I've opened https://github.com/racket/racket/issues/5066 to keep track of this.

Bogdanp:
I think this is ready to go now.

I don't think the CI Win build failure is related to this change.

PR #5072, opened 2024-09-07
racket/(mutable-)treelist: report `exact-nonnegative-integer?`

usaoc:
##### Checklist
- [x] Bugfix

### Description of change
In `(mutable-)treelist-{ref,sublist(!)}`, when given a fixnum but exact negative integer.

Also unify the reported "valid range" for fixnum and non-fixnum cases.


mflatt:
Thanks!

PR #5078, opened 2024-09-25
search: ranking results based on package status

mflatt:
This is an updated version of #4510. The main difference is that it doesn't try to infer a package from a tag. A package is determined as the package that contains the document, not the defined identifier, which locks things down a little more among documents. It also makes the notion of "base" and "main distribution" configurable via "config.rktd".

Within a group of search results of the same score, search results associated with the base packages will come first, followed by those in main-distribution packages, and then everything else.

Search results also have an additional indicator on the right border, which can be hovered to read the description. This commit changes the indicator colors compared to the original PR, using bright blue and faded blue to reflect an ordering of base -> distro -> other.

Tracking the package that contains a document requires some plumbing in the index-loading machinery, and it triggers a schema change in "docindex.sqlite". If you try this comment and then rewind, discard generated "docindex.sqlite" files.

### DIFF
--- a/pkgs/base/info.rkt
+++ b/pkgs/base/info.rkt
@@ -14,7 +14,7 @@
 
 ;; In the Racket source repo, this version should change exactly when
 ;; "racket_version.h" changes:
-(define version "8.14.0.4")
+(define version "8.14.0.5")
 
 (define deps `("racket-lib"
                ["racket" #:version ,version]))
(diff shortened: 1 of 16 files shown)


mflatt:
Forgot that this needs a Scribble update: https://github.com/mflatt/scribble/tree/docindex
I can bump the version there and increase the dependency here.

Issue #5069, opened 2024-09-07
(string-contains? ...) has inconsistent input order


Heuzi:
This is a repost of "string-contains has inversed parameters in documentation #3356 " but I think it wasn't properly discussed/solved.

If using beginning student language, by default, the system uses:

(string-contains? sub-string full-string)
![image](https://github.com/user-attachments/assets/880f1533-2dcd-4251-a88c-8a41b0858574)

But this is inconsistent with some of the documentation:

![image](https://github.com/user-attachments/assets/3e6df9da-7e57-4ef7-a5d9-24646e0a8ef4)
![image](https://github.com/user-attachments/assets/e11a6c76-622a-4a2f-b89d-e4ed03ec24c7)

The images are screenshots of:
https://docs.racket-lang.org/reference/strings.html#%28def._%28%28lib._racket%2Fstring..rkt%29._string-contains~3f%29%29
https://docs.racket-lang.org/htdp-langs/beginner.html#%28def._htdp-beginner._%28%28lib._lang%2Fhtdp-beginner..rkt%29._string-contains~3f%29%29

I understand the different (string-contains? ...) actually come from different packages. The inconsistency is unnecessarily confusing.

gus-massa:
To add some confussion

    (regexp-match? "l" "apple.scm") ;==> #t

shhyou:
Closing because no fixes are available, unless there is a "Student Languages only" distribution.

1. My current action is to tell the students to use Blue Boxes at the top-right corner and/or _"View documentation ..."_ from the pop-up menu.
2. The doc search page currently shows "HtDP" from #5080 as of v8.15 which is an improvement.

PR #5080, opened 2024-09-29
update the search index page to support language families

mflatt:
Depends on racket/scribble#443.

A "language family" here is a set of languages that share a common module-path syntax and direct interoperability, such as Racket or Rhombus. A program might be written with modules from more than one language family, of course, but the intent to make search results clearer about which interfaces are most intended to be used from a particular starting language and notation.

Another goal is to keep a Racket distribution Racket-focused, and not clutter it with unnecessary and largely redundant "Racket" labels. Naturally, an eventual Rhombus (or any other language family) distribution should be possible, and it should similarly be Rhombus-focused. So, this commit adds a notion of an installation's mahin language family, which can be configured when a distribution is built (i.e., configured through "config.rktd").

Here's what it looks like when a search finds a name in both Racket and Rhombus modules (using not-yet-merged changes to Rhombus to add language-family information to documentation):

Screenshot 2024-09-28 at 8 22 10 PM

Note that Rhombus hits are tagged as "Rhombus", but Racket hits are left alone, because Racket is the main language family for this installation.

By clicking on the gear icon above the search results, the search can be configured to only report matches for the Racket language family:

Screenshot 2024-09-28 at 8 22 47 PM

Here's that configuration panel open with "Rhombus" just selected:

Screenshot 2024-09-28 at 8 23 13 PM

The language-family shortcuts in that panel are populated by the families that show up in installed documentation.

Note that in all cases where results are shown for the Rhombus `pict` library, the module path is reported in Rhobus syntax as `pict`, instead of Rack syntax as `(lib "pict/main.rhm")`.

As the screen shots illustrate, there are some cosmetic changes independent of the language-family support. ASCII art like "[?]" and "[!]" are replaced with icons, the search context is reported at the top instead of the bottom, and there are many smaller changes.


samth:
Are there other languages in the Racket distribution or ecosystem currently that seem like they'd be good candidates for this? I can imagine a "Typed Racket" or "Scribble" or "Beginning Student" label being useful but I'm not sure exactly where you'd draw the boundary or whether there are other considerations?

mfelleisen:
+1 for the teaching languages.  When novice students used the search in the past, they seemed confused. In recent years, they used Google and often found the wrong results.  So should we make an attempt to break these five languages out? (In a way we want each level to be separated … but that’s already indicated and who knows what younglings actually notice.)

mflatt:
The documentation within the commit offers this characterization of "language family":

> A language family is not merely a module-based language, but instead stands for a set of languages that share a module-naming convention; as a rule of thumb, a language family is distinct enough that it might have its own downloadable distribution. 

This isn't a concrete rule, but continuing that perspective, "HtDP" makes sense to me as a language family. I can imagine pointing students to a distribution that is only set up for the teaching languages and teachpacks; a student using those languages doesn't need to know about anything else in the Racket distribution. But "Beginner Student" is too specific for a language family, both because you don't want students download five different distributions during a semester, and because the connections among the languages are important.

"Typed Racket" seems less like a language family to me, as things stand, because I don't think the idea is that TR programmers always stay in TR or can work effectively without realizing that non-TR languages exist. But if there's an effort to make it more distinct in that way, then a language family would make sense.

Languages like Plait or Shplait do not seem like good candidates for a language family, because they're covered by just `plait` or `shplait` as a language module. If there were an ecosubsystem of Plait or Shplait library packages, somehow, then I guess they'd graduate to language-family status, but that evolution isn't going to happen for those specific languages.

Qi seems more in a TR-like space now, but it seems distinct and stand-alone enough in principle to be a language family, depending on how the community around the language evolves and perceives itself.

In principle, there's room for lots of language families in this perspective. The search-page interface here is set up for a relatively small number of language families in an installation, though, and "docs.racket-lang.org" is an installation in that sense.

That's as far as I got, anyway.

shhyou:
> [samth](https://github.com/samth) wrote:
> Are there other languages in the Racket distribution or ecosystem currently that seem like they'd be good candidates for this? I can imagine a "Typed Racket" or "Scribble" or "Beginning Student" label being useful but I'm not sure exactly where you'd draw the boundary or whether there are other considerations?

Perhaps another example could be Scheme and the SRFI libraries.

+1 for a "Teaching Languages distribution" configured to prioritize HtDP and teaching pack bindings. The labels can partially address #5069.

mfelleisen:
@mflatt I totally agree that the separate levels of *SL should all go into one family.  Otherwise it wouldn't be a "family" :-) 

@mikesperber Mike may wish to gather his teaching languages in a separate family (which might be less urgent because his docs are in a distinct natural language).

samth:
Typed Racket certainly isn't intended to be a separate ecosystem, but nonetheless people often want to know, when searching for a library, if it's typed. So that seems in between the *SL case and the case of Scribble, for example.

mflatt:
I ended up further generalizing the plumbing for an index entry's language family so that it can be set a different layers. In the case of the HtDP languages, the "HtDP" language family is specified in the "info.rkt" `scribblings` clause. In the case of documentation rendered through `scribble/rhombus`, the default language family is configured to be "Rhombus" by the `title` function — but that default can be overridden on the outside via `scribblings`, or it can be overridden on the inside at the granularity of a section or an individual index entry.

benknoble:
Should there be a way to specify a language family without providing a sort key? (Perhaps #f should be allowed to mean “default”?) Or would we prefer to require thinking about sort order when overriding default language families?

PR #5082, opened 2024-10-03
reference: correct note on `make-keyword-procedure`

usaoc:
##### Checklist
- [x] documentation

### Description of change
Specifically, specify object name and arity encoding of the resulting procedure when `plain-proc` is not provided, because the documented default value doesn't make that clear.


mflatt:
Thanks!

PR #5083, opened 2024-10-04
Refactor vector-count and add an optional argument to vector-member

Antigen-1:
##### Checklist

- [x] Feature
- [x] tests included
- [x] documentation

### Description of change

- The `member` function accepts an optional `is-equal?` argument but the `vector-member` function doesn't. So I hope to add this argument to the latter.
- Use the existing `varargs-check` function to check the arguments of `vector-count`.


mflatt:
Thanks for these improvements! Although I don't really expect any trouble, I'm currently planning to wait until after the v8.15 branch to merge.

mflatt:
I added a `history` note and adjusted the contract used in the error message to be like `member` (but left the documentation as-is). If it's ok, I can squash and merge.

mflatt:
Thanks for the reminder. I got side-tracked for a couple of days, but merged now.

PR #5084, opened 2024-10-06
xml,reader: improve performance

Bogdanp:
The KMP implementation turns out to be about 2-3x slower on inputs with long CDATA than this implementation. On short CDATA, this implementation is about 10% better.

Testing on the attached file, the following program:

```racket
#lang racket/base

(require racket/port
         xml)

(define data
  (call-with-input-file "test.xml"
    port->bytes))

(time
 (for ([_ (in-range 100)])
   (call-with-input-bytes data
     read-xml/document)))
```

returns:

```
cpu time: 4160 real time: 4317 gc time: 189
```

on HEAD

and 

```
cpu time: 1482 real time: 1549 gc time: 81
```

on this branch.

[test.xml.zip](https://github.com/user-attachments/files/17270588/test.xml.zip)

### DIFF
--- a/racket/collects/xml/private/reader.rkt
+++ b/racket/collects/xml/private/reader.rkt
@@ -1,7 +1,8 @@
 #lang racket/base
 (require racket/contract/base
          racket/list
-         "structures.rkt")
+         (only-in "structures.rkt" make-document make-prolog)
+         (except-in (submod "structures.rkt" unsafe) make-document make-prolog))
 
 (provide/contract
  [read-xml (() (input-port?) . ->* . document?)]
--- a/racket/collects/xml/private/reader.rkt
+++ b/racket/collects/xml/private/reader.rkt
@@ -270,16 +270,16 @@
 
 ;; lex-attributes : Input-port (-> Location) -> (listof Attribute)
 (define (lex-attributes in pos)
-  (let* ([result_list
+  (let* ([result-list
           (let loop ()
             (skip-space in)
             (cond [(name-start? (peek-char-or-special in))
                    (cons (lex-attribute in pos) (loop))]
                   [else null]))]
-         [check_dup (check-duplicates result_list (lambda (a b) (eq? (attribute-name a) (attribute-name b))))])
-    (if check_dup
-        (lex-error in pos "duplicated attribute name ~a" (attribute-name check_dup))
-        result_list)))
+         [check-dup (check-duplicates result-list eq? #:key attribute-name)])
+    (if check-dup
+        (lex-error in pos "duplicated attribute name ~a" (attribute-name check-dup))
+        result-list)))
 
 ;; lex-attribute : Input-port (-> Location) -> Attribute
 (define (lex-attribute in pos)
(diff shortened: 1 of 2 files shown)


Bogdanp:
> And, just to confirm, there are no functions inside those structs, right? (I didn't see any.) So there's no blame that got erased that'd fall on the client of the xml module?

That's right.

> And finally, I do wonder if that's the minimal set to drop to get the speedup, but that's more idle curiosity than anything.

Initially, I only switched to the unsafe versions of `make-{element,attribute,entity,cdata}` and maybe one or two more (the commonality being that they were all used inside loops), but I ended up gaining (IIRC) ~50ms-100ms by switching all of them, so I figured it was worth it. Probably, for `prolog` and `document` we can keep the contracts w/o any perf impact so I'll bring those back.

Bogdanp:
The last set of changes squeezes another 10% improvement on the file in the original post and a 25% improvement on an even larger file ([test2.xml.zip](https://github.com/user-attachments/files/17322656/test2.xml.zip)).

racket/collects/xml/private/reader.rkt
```diff
@@ -332,50 +332,39 @@
...
+  (define c (non-eof read-char-or-special in pos))
+  (unless (name-start? c)
+    (lex-error in pos "expected name, received ~e" c))
+  (let loop ([chars (list c)])
+    (define c (non-eof peek-char-or-special in pos))
+    (cond
+      [(eq? c 'special)
+       (lex-error in pos "names cannot contain non-text values")]
+      [(name-char? c)
+       (loop (cons (read-char in) chars))]
```
benknoble, line 365: [non-blocking: I'm not well-versed in this code] This looks like it throws away c, but the original code threw away the next read; is c from a "peek", which I suppose would make these are equivalent?

Bogdanp: >  is c from a "peek", which I suppose would make these are equivalent?

Yes. The previous code peeked c then threw away the read. This code peeks then uses the read. The behavior is equivalent, but this version is shorter.

mflatt:
@Bogdanp Should these commits be merged as-is, or should they be squashed?

Bogdanp:
@mflatt I've squased the intermediate commits, but would prefer to keep these three final commits separate.

mflatt:
Thanks!

benknoble:
For future readers interested in what changed between the final rebase, here's the range-diff :) Fortunately I was able to recover the old version by fetching the (full-length) SHA from Bogdan's fork (the abbreviated commit hash was not sufficient).

<details><summary>range-diff:</summary>

```
1:  710704b8df ! 1:  ea90e8aea5 xml,reader: improve performance of gen-read-until-string
    @@ Commit message
         with long CDATA than this implementation.
     
      ## racket/collects/xml/private/reader.rkt ##
    +@@
    + #lang racket/base
    +-(require racket/contract
    ++(require racket/contract/base
    +          racket/list
    +-         racket/match
    +          "structures.rkt")
    + 
    + (provide/contract
     @@ racket/collects/xml/private/reader.rkt: (define (read-xml-element-helper pos in start)
        (cond
          [(start-tag? start) (read-element start in pos)]
2:  6bd3d1d7c2 < -:  ---------- xml,reader: drop no-longer-used racket/match import
3:  81b17355e6 ! 2:  8c23f95d55 xml,reader: avoid enforcing contracts during parse
    @@ Commit message
         xml,reader: avoid enforcing contracts during parse
     
         This change further reduces the time it takes to parse large xml
    -    documents by about 1/3 for the file tested in 710704b.
    +    documents by about 1/3 for the file tested in ea90e8.
     
      ## racket/collects/xml/private/reader.rkt ##
     @@
    @@ racket/collects/xml/private/reader.rkt
      (require racket/contract/base
               racket/list
     -         "structures.rkt")
    -+         (submod "structures.rkt" unsafe))
    ++         (only-in "structures.rkt" make-document make-prolog)
    ++         (except-in (submod "structures.rkt" unsafe) make-document make-prolog))
      
      (provide/contract
       [read-xml (() (input-port?) . ->* . document?)]
4:  06d4a66956 < -:  ---------- xml,reader: use contracted versions of make-{document,prolog}
5:  bc80d14190 < -:  ---------- xml,reader: avoid port overhead while lexing pcdata
6:  923857aea5 < -:  ---------- xml,reader: prefer check-duplicates with #:key argument
7:  ea1b61d8c1 ! 3:  e883cd4d79 xml,reader: avoid port overhead in lex-name and read-until
    @@ Metadata
     Author: Bogdan Popa <bogdan@defn.io>
     
      ## Commit message ##
    -    xml,reader: avoid port overhead in lex-name and read-until
    +    xml,reader: avoid port overhead while lexing
     
      ## racket/collects/xml/private/reader.rkt ##
    -@@ racket/collects/xml/private/reader.rkt: (define data
    +@@ racket/collects/xml/private/reader.rkt: (define (lex-tag-cdata-pi-comment in pos)
    + 
    + ;; lex-attributes : Input-port (-> Location) -> (listof Attribute)
    + (define (lex-attributes in pos)
    +-  (let* ([result_list
    ++  (let* ([result-list
    +           (let loop ()
    +             (skip-space in)
    +             (cond [(name-start? (peek-char-or-special in))
    +                    (cons (lex-attribute in pos) (loop))]
    +                   [else null]))]
    +-         [check_dup (check-duplicates result_list (lambda (a b) (eq? (attribute-name a) (attribute-name b))))])
    +-    (if check_dup
    +-        (lex-error in pos "duplicated attribute name ~a" (attribute-name check_dup))
    +-        result_list)))
    ++         [check-dup (check-duplicates result-list eq? #:key attribute-name)])
    ++    (if check-dup
    ++        (lex-error in pos "duplicated attribute name ~a" (attribute-name check-dup))
    ++        result-list)))
    + 
    + ;; lex-attribute : Input-port (-> Location) -> Attribute
    + (define (lex-attribute in pos)
    +@@ racket/collects/xml/private/reader.rkt: (define (skip-space in)
    + ;; lex-pcdata : Input-port (-> Location) -> Pcdata
    + ;; deviation - disallow ]]> "for compatibility" with SGML, sec 2.4 XML spec
    + (define (lex-pcdata in pos)
    +-  (let ([start (pos)]
    +-        [data (let ()
    +-                (define out (open-output-string))
    +-                (let loop ()
    +-                  (let ([next (peek-char-or-special in)])
    +-                    (cond
    +-                      [(or (eof-object? next)
    +-                           (not (char? next))
    +-                           (eq? next #\&)
    +-                           (eq? next #\<))
    +-                       (void)]
    +-                      [(and (char-whitespace? next) (collapse-whitespace))
    +-                       (skip-space in)
    +-                       (write-char #\space out)
    +-                       (loop)]
    +-                      [else
    +-                       (write-char next out)
    +-                       (void (read-char in))
    +-                       (loop)])))
    +-                (get-output-string out))])
    +-    (make-pcdata start (pos) data)))
    ++  (define start (pos))
    ++  (define data
    ++    (let loop ([chars null])
    ++      (let ([next (peek-char-or-special in)])
    ++        (cond
    ++          [(or (eof-object? next)
    ++               (not (char? next))
    ++               (eq? next #\&)
    ++               (eq? next #\<))
    ++           (apply string (reverse chars))]
    ++          [(and (char-whitespace? next) (collapse-whitespace))
    ++           (skip-space in)
    ++           (loop (cons #\space chars))]
    ++          [else
    ++           (loop (cons (read-char in) chars))]))))
    ++  (make-pcdata start (pos) data))
      
      ;; lex-name : Input-port (-> Location) -> Symbol
      (define (lex-name in pos)
```

</details>

PR #5086, opened 2024-10-10
CS: rewrite several `~s is not a valid ....` errors

usaoc:
##### Checklist
- [x] Bugfix
- [ ] Feature
- [x] tests included
- [ ] documentation

### Description of change
For `make-{string,vector,fxvector,flvector,bytevector}`, report `exact-nonnegative-integer?` contract violation or OOM error.

For `integer->char`, report the same contract violation as on BC.


mflatt:
Thanks!

Issue #4730, opened 2023-08-30
[Feature Request] `open-input-nowhere`


LiberalArtist:
We have `open-output-nowhere` to discard unwanted output &aacute; la `/dev/null`. I often want an analogous way to create empty input ports.

One concrete use-case is running a process with `system*`, e.g.:

```
#lang racket

(parameterize ([current-input-port (open-input-string "")])
  (system* (find-executable-path "ping")
           "-c" "1" "racket-lang.org"))
```

Without `parameterize`ing the `current-input-port`, the program you run will inherit `current-input-port`, which it might not understand, might unexpectedly consume, or might not be supposed to have access to. DrRacket will display a widget in the interactions window for typing input.

As workarounds, `(open-input-string "")` and `(open-input-bytes #"")` aren't bad, but `(open-input-nowhere)` would more clearly communicate intent. It might also provide an opportunity for a more efficient implementation.

sorawee:
The documentation for `make-input-port` actually already gave an implementation for `open-input-nowhere`. Someone who wants to work on this should benchmark how it performs against `(open-input-bytes #"")`.

PR #5087, opened 2024-10-17
Add `open-input-nowhere`

Mehbark:
##### Checklist

- [x] Feature
- [x] tests included
- [x] documentation

### Description of change
Adds `open-input-nowhere`, an `input-port?` analogue of `open-output-nowhere` requested in issue #4730. Basic benchmarking informed the use of `(open-input-bytes #"")` for the implementation, but I'd appreciate help getting this up to mainline standards—I'm no expert.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/port-lib.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/port-lib.scrbl
@@ -79,7 +79,7 @@ lines like @racket[read-bytes-line].
 The input port is closed unless @racket[close?] is @racket[#f].
 
 @examples[#:eval port-eval
-(port->bytes-lines 
+(port->bytes-lines
  (open-input-string "line 1\nline 2\n  line 3\nline 4"))
 ]
 
@@ -179,9 +179,9 @@ input ports as it becomes available.
 @history[#:changed "6.90.0.19" @elem{Added the @racket[name] argument.}]}
 
 
-@defproc[(make-input-port/read-to-peek 
+@defproc[(make-input-port/read-to-peek
           [name any/c]
-          [read-in (bytes? 
+          [read-in (bytes?
                     . -> . (or/c exact-nonnegative-integer?
                                  eof-object?
                                  procedure?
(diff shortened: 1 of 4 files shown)


benknoble:
> * There are a lot of whitespace cleanups in this PR that makes it hard to see what the change actually is. Can you remove those?

GitHub has a toggle to ignore whitespace changes when reviewing diffs, and git diff/log/show accept several options to ignore whitespace (such as `--ignore-space-change`/`-b`, which should be sufficient here, and `--ignore-all-space`/`-w`).

rfindler:
> > * There are a lot of whitespace cleanups in this PR that makes it hard to see what the change actually is. Can you remove those?
> 
> GitHub has a toggle to ignore whitespace changes when reviewing diffs, and git diff/log/show accept several options to ignore whitespace (such as `--ignore-space-change`/`-b`, which should be sufficient here, and `--ignore-all-space`/`-w`).

Ah, thanks, I didn't know / had forgotten that (despite seeing this advice go by before in other PR discussions, perhaps by you? sorry!). 

Still, tho, let's not merge substantive changes together with whitespace fixes. It just makes trouble for future us.

samth:
The port tests are here: https://github.com/racket/racket/blob/master/pkgs/racket-test-core/tests/racket/port.rktl

rfindler:
> > Although I am puzzled as I expected to find some port tests in pkgs/racket-test/tests but I didn't so I'm unsure where the test should go.
> 
> Same here, I'll read carefully and add them where appropriate. 

Great!

> Should I squash all the commits into one when I'm done?

I think for this particular change it is best to combine them. I'm happy to do it when you're done, however; no problem. When I'm looking at old commits, I find it very helpful to see both the docs changes and the test changes right there with the change. When changes are bigger it may make sense to split things out, it is just this one seems nice and self-contained.

Mehbark:
Tests added and passing (I added failing ones first to check that they actually run), but there's some more irrelevant whitespace changes; let me know what else is needed from my end. Thanks!

rfindler:
Thanks, I've merged this. (I don't know how to make the changes and send them back to this pull request so I just pushed them to master directly.)

PR #5088, opened 2024-10-19
Improve the error message of bytes-append* and fix bytes-join

Antigen-1:
##### Checklist

- [x] Bugfix
- [x] Feature
- [x] tests included
- [x] documentation

### Description of change

1. Check the arguments of `bytes-append*`.
2. It would be better if `bytes-join` always returns a newly allocated mutable byte string, just like `bytes-append`.

### DIFF
--- a/racket/collects/racket/bytes.rkt
+++ b/racket/collects/racket/bytes.rkt
@@ -2,9 +2,21 @@
 
 (provide bytes-append* bytes-join)
 
-(define bytes-append*
-  (case-lambda [(strs) (apply bytes-append strs)] ; optimize common case
-               [(str . strss) (apply bytes-append (apply list* str strss))]))
+(define (check-bytes-list* l name)
+  (let loop ((l l))
+    (cond ((null? (cdr l))
+           (define tail (car l))
+           (or (and (list? tail) (andmap bytes? tail))
+               (raise-argument-error name "(listof bytes?)" tail))
+           tail)
+          (else
+           (define e (car l))
+           (or (bytes? e)
+               (raise-argument-error name "bytes?" e))
+           (cons e (loop (cdr l)))))))
+
+(define (bytes-append* bstr-list-or-bstr . null-or-bstr-list*)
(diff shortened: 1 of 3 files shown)


shhyou:
Rebasing to racket/racket@3c14248 should include the fix for the foreign test

racket/collects/racket/bytes.rkt
```diff
@@ -2,9 +2,22 @@
...
+           (define tail (car l))
+           (or (and (list? tail) (andmap bytes? tail))
+               (raise-argument-error name "(listof bytes?)" tail))
+           (foldl cons tail p))
+          (else
+           (define e (car l))
+           (or (bytes? e)
+               (raise-argument-error name "bytes?" e))
+           (loop (cdr l) (cons e p))))))
+
```
mflatt, line 17: A direct `cons` on the recursion seems simpler here — and it should be faster, although that's unlikely to matter:

```
(define (check-bytes-list* l name)
  (let loop ([l l])
    (cond
      [(null? (cdr l))
       (define tail (car l))
       (unless (and (list? tail) (andmap bytes? tail))
         (raise-argument-error name "(listof bytes?)" tail))
       tail]
      [else
       (define e (car l))
       (unless (bytes? e)
         (raise-argument-error name "bytes?" e))
       (cons e (loop (cdr l)))])))
```

racket/collects/racket/bytes.rkt
```diff
@@ -2,9 +2,22 @@
...
 (define bytes-append*
-  (case-lambda [(strs) (apply bytes-append strs)] ; optimize common case
-               [(str . strss) (apply bytes-append (apply list* str strss))]))
+  (case-lambda [(strs) (apply bytes-append (check-bytes-list* (list strs) 'bytes-append*))] ; optimize common case
+               [(str . strss) (apply bytes-append (check-bytes-list* (cons str strss) 'bytes-append*))]))
```
mflatt, line 20: The common case isn't much of an optimization anymore, so maybe keep just the second one?

mflatt (approved):

mflatt:
Thanks!

PR #5090, opened 2024-10-20
Change treelist/c to never be lazy by default

lexi-lambda:
For the reasons described in https://github.com/racket/racket/issues/4922#issuecomment-2408449144, I believe the current default behavior of `treelist/c` when given a chaperone contract is a very bad idea. To summarize the points from that comment:

* The overhead of the bookkeeping used by lazy `treelist/c` to track which elements belong to the original treelist is substantial, and the effect compounds if the treelist repeatedly crosses a contract boundary.

* The overhead is surprising and challenging to reason about. There are no other contracts that work this way, and the costs are not documented.

* Making laziness the default exclusively for non-flat chaperone contracts feels extremely arbitrary. It also adds to the surprise, as I doubt most users of the contract system are consciously thinking about which contracts are flat and which ones are not, while even fewer are thinking about which contracts are chaperone contracts and which require impersonators.

This change does technically change behavior of existing programs. However, I doubt that many (if any?) users are intentionally relying on the current behavior. Moreover, if the behavior is to be changed, doing it sooner rather than later seems wise.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -705,11 +705,7 @@ necessarily @racket[eq?] to the input.
 
 @defproc[(treelist/c [ctc contract?]
                      [#:flat? flat? any/c (flat-contract? ctc)]
-                     [#:lazy? lazy? any/c
-                      (cond
-                        [(flat-contract? ctc) #f]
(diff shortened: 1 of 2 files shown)


mflatt:
A simpler and more consistent default on laziness sounds right to me, and I agree that we can probably change the behavior now without too much trouble.

PR #5093, opened 2024-10-25
db: Improve `prepare` to properly handle virtual statements

lexi-lambda:
The `prepare` function is documented and contracted as accepting virtual statements, but the implementation did not handle them, resulting in an internal error. Fortunately, handling them is an easy fix.

Docs and tests for this change are in `db-doc` and `db-test` and are therefore in a separate racket/db#33 pull request.


rmculpepper (approved):
Looks good to me.

PR #5103, opened 2024-11-12
docs: provide example for the `call/cont` / `call/cc` comparison

omentic:
##### Checklist

- [x] documentation

### Description of change

This adds a short example to the page in the Racket Guide on continuations, to showcase the difference between `call/comp` and `call/cc`. I didn't understand what the page meant at first glance, and understood better after trying this out in the REPL, so thought it might be helpful to have immediately on the page itself.

I didn't see anything about a formatter in `build.md` or `contributing.md`, so I formatted things to a best-effort.

(feel free to directly edit as desired)


mflatt:
Thanks!

PR #5104, opened 2024-11-14
allow `make-parameter` as cross-phase persistent

mflatt:
Allow non-system parameters to cross phases by allowing `make-parameter` in a cross-phase persistent mode. Use that capability for the `syntax-parse` configuration parameter.

Racket has various built-in parameters that span phases. Those parameters provide a way to escape from the expander's phase separation, so I've always hoped/wished that they could go away. But their phase-spanning behavior is handy and maybe unavoidable for system-wide configuration, such as adapting the environment to a language's conventions. That kind of capability should not be restricted to built-in parameters. Absent a better idea, I'm finally ready to yield on allowing `make-parameter` in a cross-phase persistent module.


samth:
Should `racket/private/gen-temp` use parameters as a replacement for mutation to be cross-phase persistent?

samth:
The test failure is a crash in the GC I think: 

```
exception raised by error display handler: invalid memory reference.  Some debugging context lost; original exception raised: $collect-rendezvous: cannot return to the collect-request-handler
```

https://github.com/racket/racket/actions/runs/11828069088/job/32957460057?pr=5104#step:23:105

mflatt:
> Should `racket/private/gen-temp` use parameters as a replacement for mutation to be cross-phase persistent?

Unless i'm confused, making `racket/private/gen-temp`  cross-phase persistent would be counterproductive. It caches an introducer while compiling a module, which is good for minimizing the number of different scopes within a module. Using a single introducer across modules and phases, however, would pollute a module's scopes with bindings for other modules. The pollution would typically get pruned as a compiled module is marshaled to ".zo", but it would slow down binding resolution and retain garbage meanwhile.

To put it another way, `racket/private/gen-temp` seems to be taking advantage of phase separation, not in need of circumventing it.

PR #5106, opened 2024-11-21
More treelist utility functions

AlexKnauth:
##### Checklist

- [ ] Bugfix
- [x] Feature
- [x] tests included
- [x] documentation

### Description of change

Adds 7 new treelist functions:
- `treelist-index-of`: like list `index-of`
- `treelist-index-where`: like list `index-where`
- `treelist-splitf`: like list `splitf-at`
- `treelist-keep-members`: like set intersect, but with configurable equality instead of arbitrary arity
- `treelist-skip-members`: list set subtract, but with configurable equality instead of arbitrary arity, also like list `remove*` but with opposite argument order
- `treelist-flatten`: like list `flatten`
- `treelist-flatten-once`: like list `append*` on one argument

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/treelists.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/treelists.scrbl
@@ -347,6 +347,107 @@ found, the result is @racket[#f]. For a constant-time
 (treelist-find items symbol?)
 ]}
 
+@defproc[(treelist-index-of [tl treelist?]
+                            [v any/c]
+                            [eql? (any/c any/c . -> . any/c) equal?])
+         (or/c exact-nonnegative-integer? #f)]{
+
+Returns the index of the first element in @racket[tl] that is
+@racket[eql?] to @racket[v].
+If no such element is found, the result is @racket[#f].
+
+@examples[
+#:eval the-eval
+(define items (treelist 1 "a" 'apple))
+(treelist-index-of items 1)
+(treelist-index-of items "a")
+(treelist-index-of items 'apple)
+(treelist-index-of items 'unicorn)
+]}
+
+@defproc[(treelist-index-where [tl treelist?] [pred (any/c . -> . any/c)])
+         (or/c exact-nonnegative-integer? #f)]{
+
+Returns the index of the first element in @racket[tl] where applying
+@racket[pred] to the element produces a true value.
+If no such element is found, the result is @racket[#f].
+
+@examples[
+#:eval the-eval
+(define items (treelist 1 "a" 'apple))
+(treelist-index-where items number?)
+(treelist-index-where items string?)
+(treelist-index-where items symbol?)
+(treelist-index-where items void?)
+]}
+
+@defproc[(treelist-splitf [tl treelist?] [pred (any/c . -> . any/c)])
+         (values treelist? treelist?)]{
+
+Splits @racket[tl] into 2 treelists returned as values.
+The first treelist contains elements taken successivly from @racket[tl]
+as long as they satisfy @racket[pred].
+The second treelist the rest of the elements of @racket[tl], from the
+first element not satisfying @racket[pred] and onward.
+
+@examples[
+#:eval the-eval
+(treelist-splitf (treelist 2 4 5 8) even?)
+(treelist-splitf (treelist 2 4 5 8) odd?)
+(treelist-splitf (treelist 2 4 6 8) even?)
+(treelist-splitf (treelist 2 4 6 8) odd?)
+]}
+
+@defproc[(treelist-keep-members [tl treelist?]
+                                [other-tl treelist?]
+                                [eql? (any/c any/c . -> . any/c) equal?])
+         treelist?]{
+
+Produces a treelist with only members of @racket[tl] that are also
+members of @racket[other-tl], according to @racket[eql?].
+
+@examples[
+#:eval the-eval
+(treelist-keep-members (treelist 1 2 3 2 4 5 2) (treelist 2 1))
+(treelist-keep-members (treelist 1 2 3 2 4 5 2) (treelist 4 3 5))
+]}
+
+@defproc[(treelist-skip-members [tl treelist?]
+                                [other-tl treelist?]
+                                [eql? (any/c any/c . -> . any/c) equal?])
+         treelist?]{
+
+Produces a treelist with only members of @racket[tl] that are not
+members of @racket[other-tl], according to @racket[eql?].
+
+@examples[
+#:eval the-eval
+(treelist-skip-members (treelist 1 2 3 2 4 5 2) (treelist 4 3 5))
+(treelist-skip-members (treelist 1 2 3 2 4 5 2) (treelist 2 1))
+]}
+
+@defproc[(treelist-flatten [v any/c]) treelist?]{
+
+Flattens a tree of nested treelists into a single treelist.
+
+@examples[
+#:eval the-eval
+(treelist-flatten
+ (treelist (treelist "a") "b" (treelist "c" (treelist "d") "e") (treelist)))
+(treelist-flatten "a")
+]}
+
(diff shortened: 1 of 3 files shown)


jackfirth:
Can you also add `sequence->treelist`?

AlexKnauth:
`sequence->treelist` added

jackfirth:
Two performance notes on `sequence->treelist`:

- If the input is already a treelist, return it unchanged without iterating at all. That way functions can accept any sequence and convert it to a treelist immediately without any cost for users who already have a treelist.
- If the input is a list, call the `list->treelist` function. That way there's no performance difference between `sequence->treelist` and `list->treelist` for users.

AlexKnauth:
Added shortcuts for conversions to treelist from treelist, vector, and list.

mflatt:
What's the rationale for adding these particular functions? Some of these seem arbitrary to me, like`treelist-keep-members`, but maybe there's some precedent or rationale that I'm missing.

AlexKnauth:
These are the list functions that I needed to re-implement for some projects I've been writing in Rhombus. Both `keep-members` and `skip-members` came out of wanting to use set intersect and subtract operations on lists in Rhombus, but not wanting to call the subtract operation `subtract` to avoid potential confusion on duplicate elements.

mflatt:
> These are the list functions that I needed to re-implement for some projects I've been writing in Rhombus.

I'm having trouble seeing this as a rationale for adding to a core library. It seems like we should generally avoid adding things to this library, since everyone who loads `racket/treeelist` will pay for them.

I can see an argument for adding a few more functions that are known from experience to be very broadly useful. The  `treelist-flatten` and `sequence->treelist` functions are probably in that category, and maybe `treelist-index-of`.

As you may guess, I'm not a fan of the way `racket/list` evolved to have such a large set of functions. A lot of them should have been organized by task, and not in `racket/list` just because they work on lists.

jackfirth:
Perhaps `sequence-index-of` and `sequence-index-where` would be more valuable additions?

AlexKnauth:
I don't think including the functionality of `remove*` in `racket/list` was a mistake. It's a very common thing to do with lists, and I think Rhombus should have an operation with the same behavior. I considered other names like `remove_all` and `subtract`, but... I didn't want to let people think it was the same as repeatedly calling `remove` for 1 element at a time.

AlexKnauth:
Perhaps if a `filter` method is implemented with `~keep` and `~skip` keywords, use cases for `a.keep_members(b)` and `a.skip_members(b)` could be replaced with `a.filter(~keep: b.has_element(_))` and `a.filter(~skip: b.has_element(_))`.

jackfirth:
I'm not a fan of making `treelist-filter` work differently from `filter`, `vector-filter`, `hash-filter`, `sequence-filter`, etc. It should take a single predicate argument rather than separate `#:keep` and `#:skip` keyword arguments.

mflatt:
To get a sense of what should be in a Racket library, I tried grepping all packages from a January snapshot, where each grep was of the form `''[ ()]map[ ()]'` (but with `map` replaced). Checking things like `map` and `append` provides scale.

```
map            21930
append          8262
list-ref        4684
filter          3675
member          3075
memq            2464
for-each        2047
take            2197
drop            1120
flatten          816
apply append     735  ; literally, so on the same line
partition        497
sequence->list   476
append*          374
shuffle          316
add-between      280
drop-right       291
memv             272
index-of         227
filter-map       211
filter-not       202
group-by         187
argmax           154
argmin           140
splitf-at        101
check-duplicates  97
take-right        96
index-where       50
remf              15
indexes-of         6
takef-right        6
splitf-at-right    5
indexes-where      3

```

These are just rough counts, of course, given the grep approach. A number like 6 means basically unused, except that there are things that track or copy  "list.rkt".

I read this as confirming that `index-of` is borderline, and `splitf-at` and `index-where` are significantly more rare. It's true that `index-of` can be written with `index-where`, but if the point is to provide useful things directly, `index-of` seems like the one that might be worthwhile.

The `append*` function and `apply append` combination do show up a lot. This is is an operation where treelists are different from lists, though, so I'm not sure it's right to conclude that it will be as useful to treelists. If we just go by these numbers, then I'd concede that `treelist-append*` belongs.

The `partition` function shows up more than I expected (just because I never use it). I'd say it's borderline, but more worthwhile than other things we're considering.

The `shuffle` function is useful, but I'd say it belongs in a different module.

The relative rareness of `filter-not` is a little bit of an argument against keywords for `treelist-filter`. I still think the stronger argument is to stay like `filter` to keep Racket libraries more consistent.

My current conclusions on what to include:
 * `treelist-filter` - yes (not with keywords)
 * `treelist-append*` - yes
 * `treelist-flatten` - yes
 * `treelist-partition` - probably
 * `sequence->treelist` - probably
 * `treelist-index-of` - maybe
 * `treelist-index-where` - no
 * `treelist-splitf` - no

Further Rhombus discussion probably belongs in the other repo, but I would be inclined to omit `treelist-append*` in favor of writing `List.append(& lists)` (and, of course, have keyword arguments for `List.filter` for consistency with other forms).

AlexKnauth:
Would you be happier if functions that split the list into 2 values, such as `partition`, `split`, and `splitf`, were placed in a separate module like `racket/treelist/split`? Just thinking of that split because not everyone likes dealing with multiple values.

The index operations like `index-of` and `index-where` have greater potential for treelists than they did for pairlists because random access by index is more efficient for treelists: should that bump them into the main module? Or should that be separated into a `racket/treelist/index` module?

mflatt:
As much as I like the treelist datatype, I don't see so many Racket programs moving to them that we need `racket/treelist/X` libraries, so far. (It looks like 3 non-Rhombus packages use them, based on a freshened package snapshot.) That may change in the long run, but Rhombus seems like a better place to experiment with list-library organizations. Adding to Racket means forever, but Rhombus is free to change, at least for for a little while.

With that in mind, maybe it's best to add only `treelist-filter`, `treelist-append*`, and `treelist-flatten`, because we agree that they're useful, and because list-function uses support that impression (with the caveat that append might be used differently with treelist). Or maybe it's best to not add anything to `racket/treelist`, since that isn't necessary to add functions to Rhombus — or even particularly useful in the short run, since we wouldn't want to bump the required Racket version for Rhombus.

pkgs/racket-doc/scribblings/reference/treelists.scrbl
```diff
@@ -317,6 +317,20 @@ results. For a constant-time @racket[proc], this operation takes
...
+Produces a treelist with only members of @racket[tl] that satisfy
+@racket[keep].
+
+@examples[
+#:eval the-eval
+(treelist-filter even? (treelist 1 2 3 2 4 5 2))
+(treelist-filter odd? (treelist 1 2 3 2 4 5 2))
+(treelist-filter (λ (x) (not (even? x))) (treelist 1 2 3 2 4 5 2))
+(treelist-filter (λ (x) (not (odd? x))) (treelist 1 2 3 2 4 5 2))
+]}
```
mflatt, line 332: Needs a `history` note (and the same for the other additions)

pkgs/racket-test-core/tests/racket/treelist.rktl
```diff
@@ -56,6 +56,9 @@
                                                                       (regexp-quote (symbol->string 'op))
                                                                       ":"))))
 
+(define-syntax-rule (test-values expected actual)
```
mflatt, line 59: It looks like this is not used, anymore.

pkgs/racket-doc/scribblings/reference/treelists.scrbl
AlexKnauth: I've added history notes for 8.15.0.6, the current version. Is that okay, or should this include a version bump to 8.15.0.7?

mflatt: I think "8.15.0.6" is good.

PR #5107, opened 2024-11-21
[doc]: Correct the form of `contract-in`

NoahStoryM:
##### Checklist

- [ ] Bugfix
- [ ] Feature
- [ ] tests included
- [x] documentation

### Description of change

`(contract-in in-out-item ...)` -> `(contract-in module-path in-out-item ...)`

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -1944,7 +1944,7 @@ earlier fields.}}
 
 @deftogether[(@defform[
  #:literals (struct rename)
- (contract-in require-spec in-out-item ...)]
+ (contract-in module-path in-out-item ...)]
               @defform[
  #:literals (struct rename)
  (contract-out unprotected-submodule in-out-item ...)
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/contracts.scrbl
```diff
@@ -1944,8 +1944,8 @@ earlier fields.}}
 
 @deftogether[(@defform[
  #:literals (struct rename)
- (contract-in in-out-item ...)]
-               @defform[
+ (contract-in require-spec in-out-item ...)]
```
sorawee, line 1947: I don't think `require-spec` is right. It must be a module path. We can't do:

```
(require (contract-in (only-in racket add1)
                      [add1 (-> number? number?)]))
```

which fails with:

```
contract violation
  expected: (and/c syntax? (lambda (s) (module-path? (syntax->datum s))))
  given: #<syntax:3-unsaved-editor:3:22 (only-in racket add1)>
```

sorawee:
Thanks!

sorawee:
Ughh, I should have squashed the commits. My bad :/

rfindler:
Thanks!

PR #5112, opened 2024-11-23
racket/string: add `string-find`

mflatt:
Add a variant of `string-contains?` that reports the location within the first string where the second string was found.

The `string-contains?` implementation already determines a starting position for the found string, so it seems a shame not to report it.

I worry that the name `string-find` is too obvious and could conflict with existing definitions, but I don't find the string `string-find` anywhere in currently registered package sources.


sorawee:
A better name IMO is `string-index`, but that's already taken by srfi-13.

PR #5113, opened 2024-11-24
[doc]: Corrected return type for `mutable-treelist`

NoahStoryM:
##### Checklist

- [ ] Bugfix
- [ ] Feature
- [ ] tests included
- [x] documentation

### Description of change

This PR addresses a minor error in the mutable-treelist documentation:
`(mutable-treelist v ...) → treelist?` -> `(mutable-treelist v ...) → mutable-treelist?`

I noticed that in the `mutable-treelist-append!` documentation, both `treelist`
and `mutable-treelist` are valid for `other-tl`. However, the contract of
`append-proc` in `chaperone-mutable-treelist` and `impersonate-mutable-treelist`
is `(mutable-treelist? treelist? . -> . treelist?)`. I don't know much about these
two procedures, so I'm not sure if there is a problem here.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/treelists.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/treelists.scrbl
@@ -556,7 +556,7 @@ noted.
 Returns @racket[#t] if @racket[v] is a @tech{mutable treelist},
 @racket[#f] otherwise.}
 
-@defproc[(mutable-treelist [v any/c] ...) treelist?]{
+@defproc[(mutable-treelist [v any/c] ...) mutable-treelist?]{
 
 Returns a @tech{mutable treelist} with @racket[v]s as its elements in order.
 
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/treelists.scrbl
```diff
@@ -336,13 +336,14 @@ operation takes @math{O(N)} time.
 @defproc[(treelist-find [tl treelist?] [pred (any/c . -> . any/c)]) any/c]{
 
 Checks each element of @racket[tl] with @racket[pred] until the result
-is a true value, and then returns that element. If no such element is
+is a non-false value, and then returns that element. If no such element is
```
mflatt, line 339: The convention throughout the documentation is to use `a true value` (as opposed to `#true`) to mean a value other than `#false`.  So, I don't think this should change.

pkgs/racket-doc/scribblings/reference/treelists.scrbl
```diff
@@ -336,13 +336,14 @@ operation takes @math{O(N)} time.
...
 found, the result is @racket[#f]. For a constant-time
 @racket[pred], this operation takes @math{O(N)} time.
 
 @examples[
 #:eval the-eval
 (define items (treelist 1 "a" 'apple))
+(treelist-find items number->string)
```
mflatt, line 346: This seems like a fine addition, but since it's not a typical case, I think it would be better as the last example here.

mflatt (changes requested):
Thanks! The main change is clearly good, but I have reservations about the other two.

PR #5114, opened 2024-11-24
[doc]: Clarify `end` behavior in `vector-sort[!]` when set to `#f`

NoahStoryM:
##### Checklist

- [ ] Bugfix
- [ ] Feature
- [ ] tests included
- [x] documentation

### Description of change

This PR updates the `vector-sort` and `vector-sort!` documentation to clarify the behavior of the `end` and `key`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/vectors.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/vectors.scrbl
@@ -492,10 +492,13 @@ Like @racket[vector-member], but finds an element using @racket[eq?].
 (define v1 (vector 4 3 2 1))
 v1
 (vector-sort v1 <)
+v1
 (vector-sort v1 < 2 #f #:key #f)
+v1
 (define v2 (vector '(4) '(3) '(2) '(1)))
 v2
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/vectors.scrbl
```diff
@@ -485,22 +485,27 @@ Like @racket[vector-member], but finds an element using @racket[eq?].
  not modified). This sort is stable (i.e., the order of ``equal''
  elements is preserved).
 
+ If @racket[end] is @racket[#f], it is replaced with
+ @racket[(vector-length vec)].
+
 @mz-examples[#:eval vec-eval
 (define v1 (vector 4 3 2 1))
-(vector-sort v1 <)
 v1
```
sorawee, line 493: I think the point of having `v1` after `(vector-sort v1 <)` originally is to show whether `vector-sort` has side-effect. So we should either move `v1` to the end of the sequences of `(vector-sort v1 ....)`. Alternatively, if you don't think it's worth the effort to do the demonstratation, then we should just delete the lone `v1` / `v2`.

PR #5118, opened 2024-11-27
Fix use of empty-tag-shorthand in write-xexpr

joergen7:
##### Checklist

- [x] Bugfix
- [ ] Feature
- [x] tests included
- [ ] documentation

### Description of Change

See also: Issue #5119 

In `xml`, the function `write-xexpr` uses the parameter `empty-tag-shorthand` to determine which elements to shorthand. However, all tags with uppercase letters are ignored.

This pull request allows tags with uppercase letters to be shorthanded. Generally the test for shorthanding is case-insensitive now.

### DIFF
--- a/pkgs/racket-test/tests/xml/test-xexpr.rkt
+++ b/pkgs/racket-test/tests/xml/test-xexpr.rkt
@@ -1,0 +1,20 @@
+#lang racket/base
+
+(module+ test
+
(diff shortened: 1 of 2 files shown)


samth:
This has a bunch of test failures in the CI. Can you look into that?

joergen7:
It passes CI now.

PR #5123, opened 2024-11-30
Add match-case-lambda

Antigen-1:
##### Checklist

- [x] Feature
- [x] tests included
- [ ] documentation

### Description of change

Add a `match-case-lambda` function to the `racket/match` library.


sorawee:
Isn't this simply `match-lambda*` where `list` is omitted? I wonder if it's worth having a new form, and whether the implementation could much simpler by reusing work already did in `match-lambda*`.

PR #5124, opened 2024-11-30
raco pkg migrate: support --auto for compatibility with install/update

benknoble:
## Checklist

- [x] Feature
- [x] tests included
- [x] documentation

## Description of change

Add `--auto` to `raco pkg migrate` to make the interface similar to `raco pkg` `install`/`update`.

PR #5125, opened 2024-11-30
PR template: use appropriate header hierarchy

benknoble:
Use same-level h2s instead of h5 followed by h3: the latter has less semantic sense. If you take the PR title to be the h1 element (which is how GitHub marks it up), then direct child elements in the body should first be h2, then nested h3, etc.

This should also improve the screen-reader experience, which often have trouble making sense of heading levels that jump through the hierarchy.

## Checklist

- [x] (Contributor) documentation

## Description of change

Change some heading levels in the PR template. (Bet you didn't notice that I did it manually in this PR description and in that of #5124 ?)

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -8,7 +8,7 @@ For more information about how to contribute please see
 Bug fixes and new features should include tests.
 -->
 
-##### Checklist
+## Checklist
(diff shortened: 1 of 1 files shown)


sorawee:
Thanks!

PR #5136, opened 2024-12-14
Add `mutable-treelist-prepend!`

NoahStoryM:
## Checklist

- [ ] Bugfix
- [x] Feature
- [x] tests included
- [x] documentation

## Description of change

This PR adds `#:at-end?` to `mutable-treelist-append!`. `at-end` is a boolean that determines whether elements from `other-tl` are appended to the end (default behavior) or prepended to the beginning of `tl`.


NoahStoryM:
After reviewing the previous changes, I realized that modifying `mutable-treelist-append!` to include a `#:at-end?` flag could conflict with the existing behavior of `chaperone-mutable-treelist` and `impersonate-mutable-treelist`.

Would defining a separate `mutable-treelist-prepend!` function be a better way to achieve the same functionality?

mflatt:
I wonder about the order of arguments to `mutable-treelist-prepend!`. It would make more sense to me to keep the treelist to be mutated as first, following the convention that all `mutable-treelist-...` functions take the treelist to modify first. Also, even though reversing the arguments keeps them in the result order, both changing "append" to "prepend" and reversing the order feels like a kind of double negative.

NoahStoryM:
The reason for the current argument order in `mutable-treelist-prepend!` is to maintain consistency with the `prepend-proc` order in `chaperone-treelist`:

```
The prepend-proc procedure must accept a treelist being appended with tl, tl, and the current chaperone state; it must produce a chaperone replacement for the first treelist, which is prepended for the result of `treelist-append` on the chaperone, and an updated state.
```

If we change the order in `mutable-treelist-prepend!` and move `mtl` to the first position, should we also update the order in `chaperone-mutable-treelist` accordingly?

Personally, I don't have a strong preference for either order. Initially, I planned to put `mtl` first, but I realized that would create an inconsistency between `mutable-treelist-prepend!`, `chapterone-treelist`, and `chapterone-mutable-treelist`.

mflatt:
> Personally, I don't have a strong preference for either order. Initially, I planned to put mtl first, but I realized that would create an inconsistency between `mutable-treelist-prepend!`, `chapterone-treelist`, and `chapterone-mutable-treelist`.

I see what you mean, and I vaguely remember going back and forth on the argument order for `prepend-proc`. Somehow, though, it feels more ok for me to let the argument order vary for the chaperone callbacks, while it seems important for `mytable-treelist-prepend!` to keep the target mutable treelist first. So, I would be ok letting those have different orders.

PR #5139, opened 2024-12-16
Add case* forms with quasiquote support for dynamic clause selection

NoahStoryM:
## Checklist

- [ ] Bugfix
- [x] Feature
- [x] tests included
- [x] documentation

## Description of change

This PR adds `case*` and its variants (`case*/equal`, `case*/equal-always`, `case*/eq`, `case*/eqv`). These new forms extend the behavior of `case` by using `quasiquote` instead of `quote` for `datum` in each clause.


jackfirth:
Why not use `match` with `==` instead?

NoahStoryM:
> Why not use `match` with `==` instead?

I encountered a scenario where I wanted to branch based on which function `f` was. While it’s true that `match` with `==` and `or` can achieve this, I think that using `case*` made the code more readable. After all, `case` itself could also be replaced with `match` and `==`.

sorawee:
This is not correct.

```
#lang racket

(require racket/case)

(case* add1
  [(,add1) -1]
  [(,0) 0]
  [(,1) 1]
  [(,2) 2]
  [(,3) 3]
  [(,4) 4]
  [(,5) 5]
  [(,6) 6]
  [(,7) 7]
  [(,8) 8]
  [(,9) 9]
  [(,10) 10])
```

evaluates to `(void)`, while:

```
(case* add1
  [(,add1) -1]
  [(,0) 0]
  [(,1) 1]
  [(,2) 2]
  [(,3) 3]
  [(,4) 4]
  [(,5) 5]
  [(,6) 6]
  [(,7) 7]
  [(,8) 8]
  [(,9) 9])
```

evaluates to `-1`.

This is because when the number of cases are low, `case` simply searches for a clause to evaluate sequentially, but when the number of cases is past a certain threshold, `case` switches the algorithm to binary search or hash-table lookup. 

The main value of `case`-dispatching is that we can compute statically how to effectively match a value to a clause to evaluate, and that's the reason why the form is very constraining. If we allow dynamic clause selection, then that benefit is gone.

I agree with @jackfirth that `match` seems like a better mechanism for what you are trying to achieve. If the usage of `or` and `==` is too verbose for you, you can create a new match expander, or even an entirely new macro that expands to `match`.

> After all, case itself could also be replaced with match and ==.

FWIW, it's the other way around: [`match` expands to `case`](https://github.com/racket/racket/pull/4522).

NoahStoryM:
Thank you for your detailed feedback and insights! I'll explore alternative approaches using `match`.

PR #5144, opened 2024-12-21
docs: add a note that filtered-in &co. require (for-syntax racket/base)

benknoble:
## Checklist

- [x] documentation

## Description of change

Otherwise the fixnum sample doesn't work in #lang racket/base, for example.

It seems to me this should happen automatically, but I'm probably missing the key ingredient that prevents that from making sense.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/fixnums.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/fixnums.scrbl
@@ -24,6 +24,8 @@ The expected use of the @racketmodname[racket/fixnum] library is for
 code where the @racket[require] of @racketmodname[racket/fixnum] is
 replaced with
 
+@margin-note{See the documentation of @racket[filtered-in] for use with @racket[@#,(hash-lang) @#,racketmodname[racket/base]].}
+
 @racketblock[(require (filtered-in
                        (λ (name)
                          (and (regexp-match #rx"^unsafe-fx" name)
--- a/pkgs/racket-doc/scribblings/reference/syntax.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/syntax.scrbl
@@ -1580,6 +1580,16 @@ aliens
   procedure must return either a string for the import's new name or
   @racket[#f] to exclude the import.
 
+  @margin-note{
+    The second part of @racket[filtered-in] is expand-time code evaluated in the
+    scope of the enclosing module. Accordingly, most uses need
+    @racket[(require (for-syntax racket/base))] if @racketmodname[racket/base]
+    is not already imported @racket[for-syntax]. For example,
+    @racket[@#,(hash-lang) @#,racketmodname[racket]] establishes this import
+    automatically, while @racket[@#,(hash-lang) @#,racketmodname[racket/base]]
+    does not.
+  }
+
   For example,
   @racketblock[
     (require (filtered-in
@@ -1687,6 +1697,8 @@ Examples:
 
  Analogous to @racket[filtered-in], but for filtering and renaming
  exports.
+
+  @margin-note{See the documentation of @racket[filtered-in] for use with @racket[@#,(hash-lang) @#,racketmodname[racket/base]].}
 
   For example,
   @racketblock[


mflatt:
The implementation of this change doesn't seem right to me. The `hash-lang-note` form from the Guide is meant to identify libraries and bindings that are in `racket` but not `racket/base`. The generated words happen to line up with the intent here, but that's a kind of pun. Also, I doubt that it's a good idea to import Guide utilities into the Reference, but I'm also not certain we've kept things nicely separated to date.

A margin note to clarify makes sense, but probably it's worth a few words to note that second part of `filtered-in` is expand-time code, and so `(require (for-syntax racket/base))` is needed if `racket/base` is not already imported `for-syntax` (as it is by `racket`).

benknoble:
I've pushed a version 2 with significant changes to wording (and located it in what appears to be the reference's equivalent of the guide's utils module).

I can't decide on location for `filtered-in`/`out`: should it come before or after the `defform`, and inside or out of its body?

<details><summary>range-diff:</summary>

```
1:  d39cdd04dc < -:  ---------- docs: add a note that filtered-in &co. require (for-syntax racket/base)
-:  ---------- > 1:  713c098c3c docs: add a note that filtered-in &co. require (for-syntax racket/base)
```

</details>

mflatt:
I would say that the `margin-note` belongs inside the `defform`, but I know the placement doesn't always optimal then.

I usually look for ways to avoid copying (programatically) largish chunks of text. What if you put the main text in the documentation of `filtered-in`, and then have short margin notes in the "Fixnums" section and the `filtered-out` form? That would both avoid copied text and also shrink the margin notes so that the alignment next to the main text might look ok. But if that doesn't work out well from your perspective, then the current approach here is ok with me.

benknoble:
Perhaps something like this?

<details><summary>range-diff:</summary>

```
1:  713c098c3c ! 1:  3bba377ae9 docs: add a note that filtered-in &co. require (for-syntax racket/base)
    @@ pkgs/racket-doc/scribblings/reference/fixnums.scrbl: The expected use of the @ra
      code where the @racket[require] of @racketmodname[racket/fixnum] is
      replaced with
      
    -+@compile-time-note{The second part of @racket[filtered-in]}
    ++@margin-note{See the documentation of @racket[filtered-in] for use with @racket[@#,(hash-lang) @#,racketmodname[racket/base]].}
     +
      @racketblock[(require (filtered-in
                             (λ (name)
                               (and (regexp-match #rx"^unsafe-fx" name)
     
    - ## pkgs/racket-doc/scribblings/reference/mz.rkt ##
    -@@ pkgs/racket-doc/scribblings/reference/mz.rkt: (define (envvar-indexed s)
    -   @elemref['(caveat "concurrency")]{caveats concerning concurrent modification})
    - @(define (mutable-key-caveat)
    -   @elemref['(caveat "mutable-keys")]{caveat concerning mutable keys})
    -+
    -+(provide compile-time-note)
    -+(define (compile-time-note . what)
    -+  @margin-note{
    -+    @what is expand-time code evaluated in the scope of the enclosing module.
    -+    Accordingly, most uses need @racket[(require (for-syntax racket/base))] if
    -+    @racketmodname[racket/base] is not already imported @racket[for-syntax]. For
    -+    example, @racket[@#,(hash-lang) @#,racketmodname[racket]] establishes this
    -+    import automatically, while @racket[@#,(hash-lang) @#,racketmodname[racket/base]]
    -+    does not.
    -+  })
    -
      ## pkgs/racket-doc/scribblings/reference/syntax.scrbl ##
    -@@ pkgs/racket-doc/scribblings/reference/syntax.scrbl: Like @racket[require-spec], but omitting those imports that would be
    - aliens
    - ]}
    +@@ pkgs/racket-doc/scribblings/reference/syntax.scrbl: aliens
    +   procedure must return either a string for the import's new name or
    +   @racket[#f] to exclude the import.
      
    -+@compile-time-note{The second part of @racket[filtered-in]}
    ++  @margin-note{
    ++    The second part of @racket[filtered-in] is expand-time code evaluated in the
    ++    scope of the enclosing module. Accordingly, most uses need
    ++    @racket[(require (for-syntax racket/base))] if @racketmodname[racket/base]
    ++    is not already imported @racket[for-syntax]. For example,
    ++    @racket[@#,(hash-lang) @#,racketmodname[racket]] establishes this import
    ++    automatically, while @racket[@#,(hash-lang) @#,racketmodname[racket/base]]
    ++    does not.
    ++  }
     +
    - @defform[(filtered-in proc-expr require-spec)]{ 
    - 
    -   Applies an arbitrary transformation on the import names (as strings)
    +   For example,
    +   @racketblock[
    +     (require (filtered-in
     @@ pkgs/racket-doc/scribblings/reference/syntax.scrbl: Examples:
    -   an external name that matches @racket[regexp]. The @racket[regexp]
    -   must be a literal regular expression (see @secref["regexp"]).}
    +  Analogous to @racket[filtered-in], but for filtering and renaming
    +  exports.
      
    -+@compile-time-note{The second part of @racket[filtered-out]}
    ++  @margin-note{See the documentation of @racket[filtered-in] for use with @racket[@#,(hash-lang) @#,racketmodname[racket/base]].}
     +
    - @defform[(filtered-out proc-expr provide-spec)]{
    - 
    -  Analogous to @racket[filtered-in], but for filtering and renaming
    +   For example,
    +   @racketblock[
    +     (provide (filtered-out
```

</details>

Screens:

Capture d’écran 2024-12-28 à 14 36 48
Capture d’écran 2024-12-28 à 14 36 58
Capture d’écran 2024-12-28 à 14 37 09
Capture d’écran 2024-12-28 à 14 37 20
Capture d’écran 2024-12-28 à 14 37 33
Capture d’écran 2024-12-28 à 14 37 36

mflatt:
This looks great to me - thanks!

PR #5151, opened 2024-12-31
Json Representation Customization

samdphillips:
## Checklist

- [x] Feature
- [ ] tests included

## Description of change

This exposes the internal procedures `write-json*` and `read-json*` via a submodule.  Additional arguments have been added to both functions to customize the representation of the JSON objects, lists, and strings.

### Why
Users may want to use the Racket JSON reader and writer but not want to use the `jsexpr?` representation.  For example, in Rhombus it is more natural to use `String` (which are immutable) for object keys, `List` (aka `treelist?`) for JSON lists, and `String` as JSON string values.

### DIFF
--- a/racket/collects/json/main.rkt
+++ b/racket/collects/json/main.rkt
@@ -65,11 +65,15 @@
         any)] ;; jsexpr?
   ))
 
+(module* for-extension #f
+  (provide read-json*))
+
 ;; -----------------------------------------------------------------------------
 ;; CUSTOMIZATION
 
 ;; The default translation for a JSON `null' value
 (define json-null (make-parameter 'null))
+
 
 ;; -----------------------------------------------------------------------------
 ;; PREDICATE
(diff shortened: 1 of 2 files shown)


racket/collects/json/main.rkt
```diff
@@ -95,9 +100,25 @@
...
+               values
+               list?
+               values
+               symbol?
+               symbol->immutable-string
+               string?
+               values))
 
-(define (write-json* who x o jsnull enc indent)
+(define (write-json* who x o jsnull enc indent
```
sorawee, line 113: With these many arguments, would keyword arguments be a good idea? IIUC, calls are also used in a way that can be inlined, so there should be no performance issues.

racket/collects/json/main.rkt
```diff
@@ -166,49 +187,56 @@
...
+             (write-bytes #"}" o))]
           [else (raise-type-error who "legal JSON value" x)]))
   (void))
 
 ;; -----------------------------------------------------------------------------
 ;; PARSING (from JSON to Racket)
 
 (define (read-json [i (current-input-port)] #:null [jsnull (json-null)])
-  (read-json* 'read-json i jsnull))
+  (read-json* 'read-json i jsnull make-immutable-hasheq values string->symbol values))
```
NoahStoryM, line 232: Would it be possible to add a `#:mhash?` keyword to `read-json` to allow users to create mutable hashtables for JSON objects? This would be helpful for scenarios where JSON objects need to be dynamically modified (e.g., updating keys in nested objects) while maintaining backward compatibility with the default immutable hashtables. Since `jsexpr?` already considers mutable hashtables as valid JSON objects, this addition seems like a natural extension.

samdphillips: Hi @NoahStoryM,

That is out of the scope of this PR.  But (if this PR is accepted) it may be easier to propose as a new PR.

Cheers,
Sam

shhyou:
Given their 'internal function' nature, could `read-json*` and `write-json*` in some `json/private/XYZ` module and directly export them rather than adding a submodule?

mflatt:
This looks like a good addition to me. Instead of moving these to a private module, though, it seems better to document the submodule and its exports (along the lines of "Why" above).

samdphillips:
Added documentation, and implemented more uniform keyword usage.

mflatt:
Thanks!

PR #5157, opened 2025-01-03
Fix typo in help messages

halfminami:
## Checklist

- [x] documentation

## Description of change

In what `racket --help` prints, I noticed an unbalanced parenthesis in the label of `-p`. This fixes it.
I looked through command line reference to make sure there isn't the same mistake.


sorawee:
Thank you!

PR #5158, opened 2025-01-03
Fixes syntax in init-decl example

acarrico:
It seems like syntax ```#'internal``` is intended rather than keyword ```#:internal```.


sorawee:
Thank you!

PR #5163, opened 2025-01-07
[json] Add `#:mhash?` to `read-json`

NoahStoryM:
## Checklist

- [ ] Bugfix
- [x] Feature
- [x] tests included
- [x] documentation

## Description of change

This PR adds an optional `#:mhash?` keyword to the `read-json` function.

The current `read-json` implementation creates immutable hashtables for JSON
objects, making it very difficult to modify the content of a `.json` file
(e.g., adding, deleting, or updating keys).

Furthermore, `jsexpr?` already treats mutable hashtables as valid JSON objects,
making this change a natural and consistent extension.

### DIFF
--- a/pkgs/racket-doc/json/json.scrbl
+++ b/pkgs/racket-doc/json/json.scrbl
@@ -129,7 +129,8 @@ the @rfc for more information about JSON.
 @section{Parsing JSON Text into JS-Expressions}
 
 @defproc[(read-json [in input-port? (current-input-port)]
-                    [#:null jsnull any/c (json-null)])
+                    [#:null jsnull any/c (json-null)]
+                    [#:mhash? jsmhash? boolean? #f])
          (or/c jsexpr? eof-object?)]{
   Reads a @tech{jsexpr} from a single JSON-encoded input port @racket[in] as a
   Racket (immutable) value, or produces @racket[eof] if only whitespace
@@ -137,14 +138,28 @@ the @rfc for more information about JSON.
   characters in the port so that a second call can retrieve the
   remaining JSON input(s). If the JSON inputs aren't delimited per se
   (true, false, null), they must be separated by whitespace from the
-  following JSON input. Raises @racket[exn:fail:read] if @racket[in] is not
-  at EOF and starts with malformed JSON (that is, no initial sequence of bytes
-  in @racket[in] can be parsed as JSON); see below for examples.
+  following JSON input.
+
+  @racket[jsmhash?] determines whether the generated hash table is mutable.
+
+  Raises @racket[exn:fail:read] if @racket[in] is not at EOF and starts with
+  malformed JSON (that is, no initial sequence of bytes in @racket[in] can be
+  parsed as JSON); see below for examples.
 
 @examples[#:eval ev
   (with-input-from-string
     "{\"arr\" : [1, 2, 3, 4]}"
     (λ () (read-json)))
+
+  (immutable?
+   (with-input-from-string
+     "{\"arr\" : [1, 2, 3, 4]}"
+     (λ () (read-json))))
+
+  (immutable?
+   (with-input-from-string
+     "{\"arr\" : [1, 2, 3, 4]}"
+     (λ () (read-json #:mhash? #t))))
 
   (with-input-from-string
     "\"sandwich\""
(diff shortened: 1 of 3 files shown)


jackfirth:
I don't think this feature is the best idea. For one thing, it's strange for `read-json` to produce mutable hashes and _immutable_ lists. Perhaps a better approach would be to make the kinds of functional updates of immutable nested hashes you're talking about easier to do?

NoahStoryM:
For example, suppose I have an `a.json` file with the following content:

```
{"a":[{"b":2},{"c":3}]}
```

I want to update the value of `"c"` to `-3` within this JSON file.
But I can't find a good approach for achieving this kind of update.
I think the `#:mhash?` feature would provide a straightforward way to handle such scenarios.

jackfirth:
You could use [lenses](https://docs.racket-lang.org/lens/).

```scheme
(lens-set (lens-compose (hash-ref-lens 'c) second-lens (hash-ref-lens 'a))
          jsexpr
          -3)
```

Alternatively, I could see room for `jsexpr-ref` and `jsexpr-set` functions that take a sequence of symbols / list indices and strings them together into a nested get or set:

```scheme
(jsexpr-ref jsexpr (list 'a 2 'c)) ; returns 3 in your example
(jsexpr-set jsexpr (list 'a 2 'c) -3)
```

shhyou:
You can use the recently added `read-json*` function which I think is going to be included in v8.16:

[Documentation of `read-json*` from the snapshot build](https://users.cs.utah.edu/plt/snapshots/current/doc/json/index.html#%28def._%28%28submod._%28lib._json%2Fmain..rkt%29._for-extension%29._read-json%2A%29%29)

Added in #5151 by @samdphillips.

NoahStoryM:
> You can use the recently added read-json* function which I think is going to be included in v8.16:
>
> Documentation of read-json* from the snapshot build
>
> Added in #5151 by @samdphillips.

Yes, `read-json*` can indeed address this scenario. My understanding is that
`read-json*` seems to lean towards defining a data structure that is different
from what `jsexpr?` recognizes as JSON.

Currently, `jsexpr?` already accepts mutable hashtables as valid JSON objects.
Therefore, I think it's reasonable to add support for mutable hashtables in the
original `read-json` function. Modifying the value of a JSON file feels like a
common use case, and the existing functions doesn't handle it well
without relying on additional libraries or `read-json*`.


> Alternatively, I could see room for jsexpr-ref and jsexpr-set functions that take a sequence of symbols / list indices and strings them together into a nested get or set:

I'm not sure if these functions can handle more complex scenarios effectively.
I tried implementing similar functions to generate new jsexpr in the past.
However, I found that as the logic for modifying the file became more complex,
my code also grew increasingly complicated. So my solution ended up being to read
the file with `read-json` first, then convert it to use mutable hashtables, and
then process it.

This makes sense to me because the use case inherently involves modifying JSON.
Using immutable data structures in this context often feels less straightforward
than working directly with mutable ones, which can more naturally express the
intent of modifying nested JSON objects.

jackfirth:
> Currently, `jsexpr?` already accepts mutable hashtables as valid JSON objects.
Therefore, I think it's reasonable to add support for mutable hashtables in the
original `read-json` function.

I disagree with this line of reasoning. It's very reasonable for functions to be liberal in accepting both immutable and (immediately-copied) mutable data, but conservative in emitting only immutable data. Just because a function accepts mutable data doesn't mean it should be able to produce it.

> I'm not sure if these functions can handle more complex scenarios effectively.
I tried implementing similar functions to generate new jsexpr in the past.
However, I found that as the logic for modifying the file became more complex,
my code also grew increasingly complicated. So my solution ended up being to read
the file with `read-json` first, then convert it to use mutable hashtables, and
then process it.

 Would you be willing to share these scenarios? There might be other options that address the issues more clearly. Extending core APIs to produce mutable data isn't generally a thing we like to do without good reason.

NoahStoryM:
> I disagree with this line of reasoning. It's very reasonable for functions to be liberal in accepting both immutable and (immediately-copied) mutable data, but conservative in emitting only immutable data. Just because a function accepts mutable data doesn't mean it should be able to produce it.

Using `read-json*` already solves my use case for modifying JSON files. I
personally feel that unless I specifically want to use a new data structure to
represent JSON, I'd prefer to solve the JSON processing problems with the
existing jsexpr functions. That said, if this reasoning isn't sufficient to
justify modifying the existing `read-json` function, feel free to close this PR.

> Would you be willing to share these scenarios? There might be other options that address the issues more clearly. Extending core APIs to produce mutable data isn't generally a thing we like to do without good reason.

I'd be happy to. I'll need some time to dig up and simplify the old code to better
illustrate these scenarios. I think it might make sense to share these examples
on Racket Discourse for further discussion.

PR #5166, opened 2025-01-08
json: add `#:replace-malformed-surrogate?` argument

samdphillips:
## Checklist

- [x] Feature
- [x] tests included
- [x] documentation

## Description of change

As specified in the the JSON grammar 16 bit unicode escapes are allowed with `\uHHHH`.  To write a 32 bit code point the specification requires the escapes to be written with surrogate pairs ala UTF-16.  When these surrogate pairs are malformed the previous implementation would raise an exception.  This patch changes the behavior to raise an error or replace the malformed bytes with the unicode replacement character. The replacement behavior implemented here is consistent with other JSON processing tools.

### DIFF
--- a/racket/collects/json/main.rkt
+++ b/racket/collects/json/main.rkt
@@ -370,7 +370,7 @@
          (define-values (e* new-result new-pos)
            (let resync ([e e] [result result] [pos pos])
              (cond
-               [(<= #xD800 e #xDFFF)
+               [(<= #xD800 e #xDBFF)
                 (cond
                   [(equal? (peek-bytes 2 0 i) #"\\u")
                    (read-bytes 2 i)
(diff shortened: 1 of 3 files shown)


racket/collects/json/main.rkt
```diff
@@ -356,21 +367,29 @@
...
-                [else
-                 (err "bad string \\u escape, bad second half of a UTF-16 pair")])]
-             [else e]))
-         (keep-char (integer->char e*) result pos converter)]
+         (define-values (e* new-result new-pos)
+           (let resync ([e e] [result result] [pos pos])
+             (cond
+               [(<= #xD800 e #xDFFF)
+                (cond
+                  [(equal? (peek-bytes 2 0 i) #"\\u")
```
mflatt, line 375: Is this right when only the second half of a surrogate pair is present? It seems like it might try to read (another) second half and even generate a bad character if there are two second halves in a row, but maybe I'm missing something.

samdphillips: You're right.  I've added more tests and adjusted the implementation to handle this.

PR #5169, opened 2025-01-10
Update scribble.scrbl Use ``DrRacket'', not Dr. Racket.

spdegabrielle:
specify  'DrRacket' is correct 'Dr. Racket' style guide


spdegabrielle:
I am mystified why adding a line of text causes buildtest-win to fail with 
`exception raised by error display handler: invalid memory reference.  Some debugging context lost; original exception raised: $collect-rendezvous: cannot return to the collect-request-handler`

sorawee:
That's an intermittent failure. You can submit a new CI run if you really want the CI to be green, but reviewers would be aware that the failure isn't caused by the change.

sorawee:
And anyway, it's now a duplicate of #5171. Closing.

PR #5171, opened 2025-01-10
Update scribble.scrbl

spdegabrielle:
Use ``DrRacket'', not Dr. Racket.


spdegabrielle:
I know its a duplicate, but I'm going to let it do the checks because I don't know why the other failed

PR #5172, opened 2025-01-10
Optimize `for/mutable-treelist` and `treelist-reverse` performance

NoahStoryM:
## Checklist

- [ ] Bugfix
- [ ] Feature
- [x] tests included
- [x] documentation

## Description of change

This PR optimizes the performance of `for/treelist` and `treelist-reverse` by introducing new implementations that leverage `for/vector` and `vector->treelist`. See also https://racket.discourse.group/t/for-treelist-vs-vector-treelist/3441/5 .

Example:
```racket
#lang racket

(require racket/treelist)

(define (my/treelist-reverse tl)
  (define len (treelist-length tl))
  (define vec (make-vector len))
  (for ([el (in-treelist tl)]
        [i (in-range (sub1 len) -1 -1)])
    (vector-set! vec i el))
  (vector->treelist vec))

(define tl (make-treelist 10000000 'pear))
(void
 (time (my/treelist-reverse tl))
 (time (treelist-reverse tl))
 (newline))


(define-for-syntax (make-for/treelist for/vector-id vector->treelist-id)
  (lambda (stx)
    (syntax-case stx ()
      [(_ binds body0 body ...)
       (quasisyntax/loc stx
         (#,vector->treelist-id
          (#,for/vector-id binds body0 body ...)))]
      [(_ #:length length-expr #:fill fill-expr binds body0 body ...)
       (quasisyntax/loc stx
         (#,vector->treelist-id
          (#,for/vector-id #:length length-expr #:fill fill-expr binds body0 body ...)))]
      [(_ #:length length-expr binds body0 body ...)
       (quasisyntax/loc stx
         (#,vector->treelist-id
          (#,for/vector-id #:length length-expr binds body0 body ...)))])))

(define-syntax my/for/treelist
  (make-for/treelist #'for/vector #'vector->treelist))
(define-syntax my/for*/treelist
  (make-for/treelist #'for*/vector #'vector->treelist))

(void
 (time (my/for/treelist #:length 10000000 ([i 10000000]) i))
 (time (my/for/treelist ([i 10000000]) i))
 (time (for/treelist ([i 10000000]) i))
 (newline))
```

Output:
```
cpu time: 427 real time: 428 gc time: 256
cpu time: 1842 real time: 1849 gc time: 288

cpu time: 220 real time: 221 gc time: 146
cpu time: 558 real time: 560 gc time: 353
cpu time: 1396 real time: 1400 gc time: 179

```

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/treelists.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/treelists.scrbl
@@ -464,6 +464,8 @@ Like @racket[for/vector] and @racket[for*/vector], but generating
 
 @examples[
 #:eval the-eval
+(for/treelist ([i (in-range 10)]) i)
+(for/treelist #:length 15 ([i (in-range 10)]) i)
 (for/treelist #:length 15 #:fill 'a ([i (in-range 10)]) i)
 ]}
 
@@ -959,6 +961,8 @@ Like @racket[for/vector] and @racket[for*/vector], but generating
 
 @examples[
 #:eval the-eval
+(for/mutable-treelist ([i (in-range 10)]) i)
+(for/mutable-treelist #:length 15 ([i (in-range 10)]) i)
 (for/mutable-treelist #:length 15 #:fill 'a ([i (in-range 10)]) i)
 ]}
 
--- a/pkgs/racket-doc/scribblings/reference/treelists.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/treelists.scrbl
@@ -455,17 +455,16 @@ If @racket[s] is infinite, this function does not terminate.
 @history[#:added "8.15.0.6"]}
 
 @deftogether[(
-@defform[(for/treelist (for-clause ...) body-or-break ... body)]
-@defform[(for*/treelist (for-clause ...) body-or-break ... body)]
+@defform[(for/treelist maybe-length (for-clause ...) body-or-break ... body)]
+@defform[(for*/treelist maybe-length (for-clause ...) body-or-break ... body)]
 )]{
 
-Like @racket[for/list] and @racket[for*/list], but generating
+Like @racket[for/vector] and @racket[for*/vector], but generating
 @tech{treelists}.
 
 @examples[
 #:eval the-eval
-(for/treelist ([i (in-range 10)])
-  i)
+(for/treelist #:length 15 #:fill 'a ([i (in-range 10)]) i)
 ]}
 
 @defproc[(chaperone-treelist [tl treelist?]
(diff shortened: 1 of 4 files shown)


pkgs/racket-doc/scribblings/reference/treelists.scrbl
```diff
@@ -455,17 +455,16 @@ If @racket[s] is infinite, this function does not terminate.
...
 
-Like @racket[for/list] and @racket[for*/list], but generating
+Like @racket[for/vector] and @racket[for*/vector], but generating
 @tech{treelists}.
 
 @examples[
 #:eval the-eval
-(for/treelist ([i (in-range 10)])
-  i)
+(for/treelist #:length 15 #:fill 'a ([i (in-range 10)]) i)
```
sorawee, line 467: I think we should add new examples, rather than modifying the existing example. The original form is probably going to be the common use, still, and we should include that.

NoahStoryM:
I also considered optimizing `sequence->treelist`:

```racket
(define (sequence->treelist s)
  (cond
    [(treelist? s) s]
    [(vector? s) (vector->treelist s)]
    [(list? s) (list->treelist s)]
    [(sequence? s) (for/treelist #:length (sequence-length s) ([el s]) el)]
    [else
     (raise-argument-error* 'sequence->treelist 'racket/primitive "sequence?" s)]))
```

However, to use `sequence-length`, the racket/sequence module must be required. Since racket/sequence already depends on racket/treelist, this creates a circular dependency issue.

Notably, `sequence->treelist` was introduced after Racket 8.15, and Racket 8.16 has not been released yet. Would it make sense to submit a separate PR to move the `sequence->treelist` function to racket/sequence instead of racket/treelist? This would align with the current design of related functions like `sequence->list`, which is in racket/sequence rather than racket/list.

jackfirth:
That `sequence->length` optimization wouldn't be correct anyway since sequences can have side effects during iteration. The `sequence->treelist` function should only iterate a generic sequence once.

mflatt:
The `for/treelist` form should not be implemented in terms of `for/vector`, because `for/vector` doesn't play well with continuations. I think that internal state is ok for `for/vector`, which is usually for a mutation-oriented content, anyway — but not `for/treelist`. Along the same lines, I don't think there should be a `#:length` option for `for/treelist`. I would not have the same objections for making these changes to `for/mutable-treelist`.

Implementing `for/treelist` in terms of `for/list` would provide performance in for large lists while still working right with continuations. But measuring just the time for length 10000000 is not good enough for a performance argument. Few programs create a list of that length, while programs that create lists of length 1, 2, or 3 are very common. On my machine, `for/treelist` appears to be faster than `for/list` plus `list->treelist` until somewhere between 5 and 10 iterations. The gap between `for/treelist` and using `for/vector` also becomes very small. So, that's why `for/treelist` is implemented the way it is right now. Below is the program I used to check again.

```
#lang racket
(require racket/treelist)

(define-syntax-rule (measure N M)
  (begin
    (printf "~a times ~a length\n" N M)
    (println "for/treelist")
    (time
     (treelist-length
      (for/fold ([v #false]) ([i (in-range N)])
        (for/treelist ([j (in-range M)])
          j))))

    (println "for/list -> treelist")
    (time
     (treelist-length
      (for/fold ([v #false]) ([i (in-range N)])
        (list->treelist
         (for/list ([j (in-range M)])
           j)))))

    (println "for/vector -> treelist")
    (time
     (treelist-length
      (for/fold ([v #false]) ([i (in-range N)])
        (vector->treelist
         (for/vector ([j (in-range M)])
           j)))))

    (println "for/vector #:length -> treelist")
    (time
     (treelist-length
      (for/fold ([v #false]) ([i (in-range N)])
        (vector->treelist
         (for/vector #:length M ([j (in-range M)])
           j)))))))

(measure 1
         1000000)
(measure 100000
         10)
(measure 200000
         5)
(measure 1000000
         1)
```

NoahStoryM:
Thank you for the detailed explanation! I noticed that `list->treelist` first converts the list to a vector and then to a treelist. This was surprising to me, as I initially thought it might be implemented using `for/treelist`. Observing this difference in implementation is what brought my attention to the performance gap.

Given this, I wonder if `list->vector` might also have issues working well with continuations, similar to the concerns with `for/vector`. Would it be worth considering re-implementing `list->treelist` using `for/treelist` instead to ensure better compatibility with continuations?

jackfirth:
There's no continuation problem in `list->treelist` because there's no arbitrary user code executed during the iteration, like there is with `for/treelist`. So there's no way to capture a continuation midway through the iteration.

PR #5177, opened 2025-01-12
doc(openssl): fix the link to openssl-ciphers(1)

benknoble:
The current link redirects to
https://docs.openssl.org/master/man1/openssl-cmds/, from which you need to scroll to "SEE ALSO" to get to
https://docs.openssl.org/master/man1/openssl-ciphers/, where we wanted to be in the first place.

## Checklist

- [x] documentation


sorawee:
Thanks!

PR #5184, opened 2025-01-19
ci: extract run-racket-tests.sh and run more tests on PR and push

Bogdanp:
Reduces some, but not all of the duplication in the workflows and runs more tests on PR and push. In particular, I didn't touch the asan and windows workflows since the former relies on skipping failing tests to produce a report and the latter can't run bash scripts.

### DIFF
--- a/.github/scripts/run-racket-tests.sh
+++ b/.github/scripts/run-racket-tests.sh
@@ -1,0 +1,75 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+HERE="$(dirname "$0")"
+BIN="$HERE/../../racket/bin"
+RACKET="$BIN/racket"
+RACO="$BIN/raco"
+CPUS="$("$RACKET" -e '(processor-count)')"
+
+"$RACO" pkg install --auto --skip-installed db-test racket-test
+
+do_test() {
+    "$RACO" test --timeout 300 -j "$CPUS" "$@"
+}
+
+
+# What Gets Tested
+# ~~~~~~~~~~~~~~~~
+# These tests run in GitHub Actions, where the environment isn't suited
+# to run all the tests. This script tries to run most of the tests that
+# can easily run in that environment. The rest of the tests are run by
+# DrDr[1] whenever changes are made to the master branch.
+#
+# [1]: http://drdr.racket-lang.org/
+
+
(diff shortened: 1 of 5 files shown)


.github/workflows/ci-pr.yml
```diff
@@ -17,28 +17,9 @@ jobs:
...
-        raco test -l tests/openssl/basic
-        raco test -l tests/openssl/https
-        raco test -l tests/match/main
-        raco test -l tests/zo-path
-        raco test -c tests/xml
-        raco test --timeout 300 -c tests/future
-        raco test -l tests/db/all-tests
-        raco test -c tests/stxparse
-        raco test -c tests/syntax
+      run: bash .github/scripts/run-racket-tests.sh
```
sorawee, line 22: It seems unfortunate to me that the testing script needs to have a precondition that some packages are already installed. Is it feasible to do just `make CPUS="$(nproc)"`, and move the package installation inside `run-racket-tests.sh`?

If not, I think the testing script should be commented with what packages are required for it to be functional.

pkgs/net-test/tests/net/git-checkout.rkt
```diff
@@ -59,13 +59,13 @@
    [else
     (error 'compare "no such file: ~s" a)]))
 
-(when git-exe
+(when (and git-exe (not (getenv "GITHUB_ACTIONS")))
```
samth, line 62: Why change this here?

Bogdanp: The version of `git` that comes with the GH actions Linux environment is too old and doesn't support the `-b` flag to `git init`. We could try to update it in the action, but this seemed easier (and this test is already not being run).

.github/scripts/run-racket-tests.sh
```diff
@@ -0,0 +1,75 @@
...
+tests/setup
+tests/stxparse
+tests/syntax
+tests/units
+tests/utils
+tests/xml
+EOF
+                   )
+
+for collection in $COLLECTIONS_TO_TEST; do
```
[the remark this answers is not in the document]
Bogdanp: Feel free! I'm not a bash expert so this pattern was just what came to mind at the time.

Issue #5189, opened 2025-01-26
string-append is slower than necessary/makes more garbage


endobson:
**What version of Racket are you using?**
Welcome to Racket v8.16.0.1 [cs].

**What program did you run?**
Test harness: https://gist.github.com/endobson/2cef417bf0d57b27abc5db6e0f7e51d8

**What should have happened?**
I would have expected the string-append* and built in routines to have similar performance to my hand rolled ones.

**If you got an error message, please include it here.**
```
Results

MyAppend* Strings:
Size: 4000000
cpu time: 136 real time: 139 gc time: 24
cpu time: 126 real time: 128 gc time: 25
cpu time: 191 real time: 193 gc time: 99
MyAppend2 Strings:
Size: 4000000
cpu time: 122 real time: 124 gc time: 25
cpu time: 116 real time: 117 gc time: 23
cpu time: 214 real time: 216 gc time: 118
MyAppend3 Strings:
Size: 4000000
cpu time: 386 real time: 398 gc time: 23
cpu time: 217 real time: 220 gc time: 121
cpu time: 112 real time: 115 gc time: 23
Append* Strings:
Size: 4000000
cpu time: 717 real time: 719 gc time: 557
cpu time: 794 real time: 796 gc time: 632
cpu time: 493 real time: 495 gc time: 321
```

**Please include any other relevant details**
Discussion on discord: https://discord.com/channels/571040468092321801/618895179343986688/1333131083004379146
String-append implementation in chez: https://github.com/cisco/ChezScheme/blob/main/s/5_4.ss#L40

It seems like it uses a singular non-tail recursive loop that computes the size on the way forward through the list and sets the subparts on the way back through the list. This can result in storing a large amount of data in the stack/continuation as opposed to two tail-recursive loops, and the hypothesis is that this is the cause of the difference in behavior.

mflatt:
Small clarification: There are two relevant sets of allocations in `string-append*`. The non-tail loop as described in the report is one, and that's inside `string-append`. But there's also the use of `apply` in `string-append*`, which (together with the rest arg in the `string-append` implementation) means that the given list is copied.

The list copy via `append` creates 24 bytes of allocation per element: 16 bytes per cons cell and 8 bytes per element of an intermediate argument array that the arg list is created from. (That's the difference between `my-string-append*` and `my-string-append`.) Each frame in the non-tail loop within `string-append` appears to be about 80 bytes, so that allocation does dominate, but the 24 from `apply` is also significant.

[...]

PR #5190, opened 2025-01-27
Avoid using apply in string-append*.

endobson:
Second change suggested in #5189.

TODO:
Needs improvement of error messages

## Checklist

- [X ] Bugfix
- [ ] tests included
- [ ] documentation

## Description of change
Change string-append* to avoid apply and do manual copying. Still needs some work though.

* I'm not sure how to format the error message for this (string-append* used to just pass through to string-append).
* I don't think we need to care about mutated strings during the loop, but it came up as a thought. (The same race can happen with any implementation that precomputes the size).

### DIFF
--- a/racket/collects/racket/string.rkt
+++ b/racket/collects/racket/string.rkt
@@ -16,13 +16,27 @@
   (provide build-kmp-table))
 
 
+;; TODO: What are the semantics if the strings themselves are changed during the execution?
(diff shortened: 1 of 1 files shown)


shhyou:
Superceded by https://github.com/racket/racket/issues/5189#issuecomment-2629055983 (1ccfe8989e15b7044cea5b1ec27c0398bd035404)

PR #5200, opened 2025-02-04
net,http-client: avoid sending Content-Length: 0 when there is no body

Bogdanp:
This fixes an issue where some hosts can fail when sent a GET request with a Content-Length header. One such host right now is memberfulcontent.net:

    $ curl -H'Content-Length: 0' -v \
          'https://relayfm.memberfulcontent.com/rss/1234?auth=XXX' >/dev/null
    ...
    < HTTP/2 500
    < date: Tue, 04 Feb 2025 08:56:04 GMT
    < content-type: text/html; charset=UTF-8
    < content-length: 4335
    ...


mflatt:
Thanks!

PR #5214, opened 2025-02-18
Zuo: remove unused private `modulo` binding

usaoc:
## Description of change
This is a remnant of the change that removed `modulo` from `zuo/kernel`, but actually `modulo` isn't used anywhere in private modules.  Moreover, the implementation actually has a bug: `(+ r n)` should be `(+ r m)`.  I think it's best to just remove it since it's unused.


mflatt:
Thanks!

PR #5216, opened 2025-02-24
raco pkg archive --relative bugfix

CutieDeng:
I encountered an error while trying to create an exportable code repository using the raco pkg archive --relative command. I tested this with the following versions of Racket: 8.13, 8.14, and 8.15, and the issue was reproducible in all of them (I did not test version 8.12, so any additional testing is welcome).

## Checklist

- [x] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation

## Description of change

I made the following modification to the catalog-copy code, adding a check for the existence of include-table to avoid illegal hash-ref access.


mflatt:
Thanks!

Issue #5210, opened 2025-02-14
[Feature Request] More list algorithms


jocyber:
Coming from other functional programming languages, I find that the "Pairs and Lists" API is missing a few handy algorithms that I find myself having to implement consistently. These are the ones I had in mind:

* scanl
Like foldl but outputs a list of all the intermediate results. [Reference](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:scanl)

* scanr
Like foldr but outputs a list of all the intermediate results. [Reference](https://hackage.haskell.org/package/base-4.21.0.0/docs/Prelude.html#v:scanr)

* windowing / sliding
Outputs a sliding window of the list, same as Haskell's [divvy](https://hackage.haskell.org/package/split-0.2.3.4/docs/Data-List-Split.html#v:divvy)

Using `for` forms is fine when the window is small, but quickly becomes repetitive when the window becomes larger.

* slice-by 
Takes in a binary predicate and groups the values together that satisfy the predicate. This is slightly different from the existing `group-by` function. For example:
```
> (slice-by eq? '(1 1 1 2 2 1 1))
'((1 1 1) (2 2) (1 1))
```

This is especially one I find myself re implementing a lot using takef and drop.

I would be interested in implementing these myself for lists.

halfminami:
I just wanted to mention that [algorithms](https://docs.racket-lang.org/algorithms/index.html) implements these nice functions!
Namely [scanl](https://docs.racket-lang.org/algorithms/index.html#%28def._%28%28lib._algorithms%2Fmain..rkt%29._scanl%29%29), [scanr](https://docs.racket-lang.org/algorithms/index.html#%28def._%28%28lib._algorithms%2Fmain..rkt%29._scanr%29%29), [sliding](https://docs.racket-lang.org/algorithms/index.html#%28def._%28%28lib._algorithms%2Fmain..rkt%29._sliding%29%29), and [chunk-by](https://docs.racket-lang.org/algorithms/index.html#%28def._%28%28lib._algorithms%2Fmain..rkt%29._chunk-by%29%29). `scan`s are not identical to the Haskell ones, but I hope this helps!

jocyber:
Oh, cool! Any reason these shouldn't be within the standard library, though? I'd like to be able to use them in environments where I can't import them, like Leetcode and such (I tried importing this). I know this is a minor thing but just want your thoughts.

halfminami:
I solve LeetCode problems in Racket, too!
[...]

PR #5218, opened 2025-02-25
pairs/lists: Introduce 4 additional algorithms

jocyber:
## Description
Adds some more algorithms that are commonly found in other functional programming languages.
* slice-by
* windows
* running-foldl
* running-foldr

## Checklist
- [x] Feature
- [x] tests included
- [x] documentation

## Testing details
* ran unit tests

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/pairs.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/pairs.scrbl
@@ -449,6 +449,42 @@ each call to @racket[proc]).
   (foldr (lambda (v l) (cons (add1 v) l)) '() '(1 2 3 4))]}
 
 
+@defproc[(running-foldl [proc procedure?] [init any/c] [lst list?] ...+)
+         list?]{
+
+Like @racket[foldl], but produces a list containing all the results of applying 
+@racket[proc] as well as the initial accumulator.
+
+@mz-examples[
+  (running-foldl + 0 '(1 2 3))
+  (running-foldl + 0 '())
+  (running-foldl (lambda (a b acc) 
+                   (* acc (+ a b))) 
+                 1 
+                 '(1 2)
+                 '(3 4))]
+@history[#:added "8.16"]{}
+}
+
+
+@defproc[(running-foldr [proc procedure?] [init any/c] [lst list?] ...+)
+         list?]{
+
+Like @racket[running-foldl], but produces the intermediate results from the right
+like @racket[foldr].
+
+@mz-examples[
+  (running-foldr + 0 '(1 2 3))
+  (running-foldr + 0 '())
+  (running-foldr (lambda (a b acc) 
+                   (* acc (+ a b))) 
+                 1 
+                 '(1 2)
+                 '(3 4))]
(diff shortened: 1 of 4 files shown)


shhyou:
Nit: the `#:added` version is "8.16.0.2".

https://github.com/racket/racket/blob/1f49402c4f54bd25826b9a7c80096cd7d0da7843/pkgs/base/info.rkt#L17

jocyber:
@shhyou Thanks for the feedback. Is this merge blocking?

mflatt:
I have reservations about adding even more to `racket/list`. Anything added to that library imposes a cost on (most) Racket programs to load code that they aren't using. That's not a concern in a language with a more static form of compilation and linking, but it's an issue for Racket.

Also, we don't want to add new exports that would create conflicts for existing modules. The name `windows` seems a common enough name that it might cause a conflict somewhere.

If Racket users think these are important enough to be always available from `racket/list`, then we can figure it out, but I'm not immediately convinced that these are common enough operations to have in `racket/list`.

jocyber:
@mflatt So I'm confused about the decision making here. These are definitely very useful utilities that functional programmers use all the time, and they complement the language well. Yet, additional list accessors like `eleventh`, `twelfth`, etc were approved? I think the ones being presented here are significantly more useful than something that could just be an alias for `list-ref ls x`, for example. I got a lot of consensus around this in the linked issue, so be sure to check that out. Furthermore, I think that after this change Racket is definitely reaching it's limit on what should be added, I would agree with you on that. So curiously how much of a cost on loading would adding 4 more algorithms really bring? I'm asking this out of ignorance. 

I get your point about `windows`, though. Maybe `sliding-window` would be a better name?

Finally, are there any guidelines on what should and shouldn't be added? It would be helpful to avoid adding even more beyond this if it were in some guideline.

mflatt:
I see now that there was some discussion on the linked issue #5210, and it already covered much of this ground, but I'll at least answer the direct questions here.

> Yet, additional list accessors like `eleventh`, `twelfth`, etc were approved?

I can only report that I was dismayed by those additions. They were merged before I could say anything.

> So curiously how much of a cost on loading would adding 4 more algorithms really bring? I'm asking this out of ignorance.

It will be very small and difficult to measure. I have not done the work to measure this, but I'm pretty sure that no one would actually notice the difference for this set of additions, as long as there are no conflicts. But it's a non-zero cost, and enough additions like this create a measurable difference.

> Finally, are there any guidelines on what should and shouldn't be added?

No, there's no simple rule to apply here. There's a simple rule of not adding anything that creates a conflict for registered packages (again, discussed in #5210), but that's only a small facet here. It's always a judgement call.

jocyber:
I'm going to just raise a revision here. Under `racket/collects/racket` we can add another `list` subdirectory and put some additional modules under there. This will prevent these additional functions from being loaded by the `racket` and `racket/list` collections. I tested it out and the functions are not loaded unless I explicitly add the following: `(require racket/list/iteration)`. This doesn't seem to clash with `racket/list` on its own.

PR #5220, opened 2025-03-02
Add sequence-tools package

jocyber:
## Description
Adds some more algorithms in a new package that are commonly found in other functional programming languages.
* slice-by
* windows
* running-foldl
* running-foldr

## Checklist
- [x] Feature
- [x] tests included
- [x] documentation

## Testing details
* ran unit tests
*  Ensured that the bindings are not loaded unless the package is 
   * Installed with `raco pkg install sequence-tools`
   * The modules are imported `(require racket/list/iteration)` and `(require racket/list/grouping`


mflatt:
Thanks for this updated PR!

About the name: How likely is this package is evolve to have sequence-generic functions, instead of sticking with list functions? I'm wondering whether `more-list-lib` might be a better package name, but maybe I don't have the right long-term expectation here.

Since the documentation is in `racket-doc`, then I don't think we need the package without `-lib`.

The documentation should use `examples` instead of `codeblock`, and I think `for-label` imports of `racket/list/....` are needed. The `racket-doc` package should declare a dependency on the new package. We wouldn't normally use `@history[#:added "1.0"]` for things that were always in a package, so I recommend dropping that.

jocyber:
@mflatt 
```
About the name: How likely is this package is evolve to have sequence-generic functions, instead of sticking with list functions? I'm wondering whether more-list-lib might be a better package name, but maybe I don't have the right long-term expectation here.
```

During our conversation in the linked issue, someone expressed a concern that this change would create a further divide between the different sequence libraries like streams, generic sequences, etc. I'm inclined to agree with that. I would like this packages name to be generic enough so that we can add extensions to those as well, if desired (without needing a brand new one). I don't think it's a huge deal and I'm willing to change the name if you really want, but as a user, I do find the inconsistency between the different libraries somewhat annoying.  

```
The documentation should use examples instead of codeblock, and I think for-label imports of racket/list/.... are needed. The racket-doc package should declare a dependency on the new package.
```
I think I tried this and had issues, but I'll give another go at it. 

```
We wouldn't normally use @history[#:added "1.0"] for things that were always in a package, so I recommend dropping that.
```
Yeah, I noticed this during development but I wasn't 100% sure if you would want this. I will update in revision.

sorawee:
@jocyber did you intentionally close this?

jocyber:
I did not.

PR #5225, opened 2025-03-08
Add sequence-tools package

jocyber:
## Description
Adds some more algorithms in a new package that are commonly found in other functional programming languages.
* slice-by
* windows
* running-foldl
* running-foldr

## Checklist
- [x] Feature
- [x] tests included
- [x] documentation

## Testing details
* ran unit tests
*  Ensured that the bindings are not loaded unless the package is 
   * Installed with `raco pkg install sequence-tools-lib`
   * The modules are imported `(require racket/list/iteration)` and `(require racket/list/grouping`


jocyber:
Apologies for the confusion. I messed up my commits. Addressed the comment from Matthew here: https://github.com/racket/racket/pull/5220

pkgs/racket-doc/scribblings/reference/pairs.scrbl
```diff
@@ -449,6 +449,53 @@ each call to @racket[proc]).
...
+          (require racket/list/iteration)]
+
+
+@; ----------------------------------------
+@section{More List Iteration}
+
+@defmodule[racket/list/iteration #:packages ("sequence-tools-lib")]
+
+The bindings in this section are provided by the @racket[sequence-tools-lib] packages 
+which acts as an extension to the base sequence libraries.
```
mflatt, line 465: It would be more consistent with other places to use `@note-lib-only[racket/list/iteration]` instead of `defmodule` and the two lines text here.

It's true that the module is provided by a different package, but it's even more relevant that the bindings aren't provided by `racket`. Also, the package name can be inferred. (I can imagine that `#:packages` might have been needed along the way, before everything was set up right.)

pkgs/racket-doc/scribblings/reference/pairs.scrbl
```diff
@@ -1658,6 +1702,40 @@ produces a true value.
 }
 
 
+@; ----------------------------------------
+@section{More List Grouping}
+
+@defmodule[racket/list/grouping #:packages ("sequence-tools-lib")]
+
+The bindings in this section are provided by the @racket[sequence-tools-lib] packages 
+which acts as an extension to the base sequence libraries.
```
mflatt, line 1711: And `@note-lib-only[racket/list/grouping]` here.

mflatt (changes requested):
This looks about ready to merge — just a small suggestion (in two places) to use a non-obvious declaration form that's specific to the Reference.

PR #5230, opened 2025-03-19
Fix analysis categories for the scanbuild workflow

bogdanap:
## Checklist

- [ ] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation

## Description of change

Hello from GitHub Code Scanning! 😄👋

We noticed your repo had an unusually large number of active categories, which I think was unintended.

By including the commit SHA in the category name, each scan will create a new category.
This means you won't be able to track alert changes between two different runs of the same tool.
Similarly, if an alert is fixed in recent runs, it will continue to appear as open because it's still present in all the other scan categories.

You can see this manifested on the alert page, where there are multiple versions of the same alert created for each new scan:
image

This change should ensure Code Scanning works as expected and allows the correct management of alerts.

Unfortunately, you'll still be stuck with all the old categories unless you manually delete them via the API.
If it makes things easier, we can clean them up for you.

### DIFF
--- a/.github/workflows/scanbuild_static-analysis.yml
+++ b/.github/workflows/scanbuild_static-analysis.yml
@@ -219,4 +219,4 @@ jobs:
       if: ${{ steps.chunk_presence.outputs.presence == '1' }}
       with:
         sarif_file: workspace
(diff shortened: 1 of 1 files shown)


sorawee:
Thank you! Will these automatically expire? If so, I think we can just wait for it to expire.

PR #5231, opened 2025-03-21
docs: SHA1 is mandatory for package checksums

otherjoel:
For background: https://joeldueck.com/code/pluto

The old wording seems to indicate a package’s checksum can be *any* unique string — i.e. it doesn’t actually _have_ to be computed using SHA1.

However, when using [manual deployment][md], `raco pkg` will actually compute the SHA1 hash of `pkg.zip` and bail if the contents of `pkg.zip.CHECKSUM` don’t match (including if `pkg.zip.CHECKSUM` is not provided).

I think there is a case to be made that the old wording should be kept, and that `raco pkg` should stop verifying the contents of `pkg.zip`.[^1] But for now the docs should be updated to match current behavior.

[^1]: Verifying the checksum doesn’t provide real protection against anything, since anyone who can fiddle with or intercept `pkg.zip` can also do so with the provided checksum.

[md]: https://docs.racket-lang.org/pkg/getting-started.html#%28part._manual-deploy%29


otherjoel:
> I think there is a case to be made that the old wording should be kept, and that `raco pkg` should stop verifying the contents of `pkg.zip`.

I've opened #5232 to discuss this alternative.

PR #5234, opened 2025-03-24
[json] Repair `#:make-key` function argument

samdphillips:
## Checklist
- [x] Bugfix
- [ ] tests included
- [ ] documentation

## Description of change
Before the `#:make-key` argument function would receive whatever the
`#:make-string` argument function returned.  If it wasn't a `string?`
this would raise an exception.  Now the `#:make-key` argument function
consistenetly receives `string?` value.

### DIFF
--- a/racket/collects/json/main.rkt
+++ b/racket/collects/json/main.rkt
@@ -416,13 +416,18 @@
   ;;
   (define (read-hash)
     (define (read-pair)
-      (define k (read-json))
-      (unless (string? k) (err "non-string value used for json object key"))
-      (define ch (skip-whitespace))
-      (when (eof-object? ch)
+      (define ch0 (skip-whitespace))
+      (cond
+        [(eqv? ch0 #\") (read-byte i)]
+        [(eof-object? ch0) (read-byte i)
+                           (err "unexpected end-of-file while parsing a json object pair")]
+        [else (err "non-string value used for json object key")])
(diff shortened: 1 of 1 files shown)


racket/collects/json/main.rkt
```diff
@@ -416,13 +416,18 @@
...
-      (define k (read-json))
-      (unless (string? k) (err "non-string value used for json object key"))
-      (define ch (skip-whitespace))
-      (when (eof-object? ch)
+      (define ch0 (skip-whitespace))
+      (cond
+        [(eqv? ch0 #\") (read-byte i)]
+        [(eof-object? ch0) (read-byte i)
+                           (err "unexpected end-of-file while parsing a json object pair")]
+        [else (err "non-string values used for json object key")])
```
mflatt, line 424: "values" -> "value"?

samdphillips: Yep, not sure how that slipped in.

mflatt:
Since the old behavior was in a release, I wonder whether this merits a `history` note in the documentation. The old behavior was less good, but it was something we could write down — and, in the short term, a history note may help with managing compatibility across versions. Then again, the relevant funcitonality is a corner that seems unlikely to be much used before the next release (except for a Rhombus binding), so maybe the change is too small to matter. I currently lean towards "too small to matter".

PR #5239, opened 2025-03-28
Link to `raco demod` in `raco exe`

jyn514:
## Checklist

- [x] documentation

## Description of change

`demod` has better runtime performance than `make`, and is closer to the semantics that people coming from compiled languages will expect.

image

PR #5240, opened 2025-03-28
Document how to document raco itself

jyn514:
## Checklist

- [x] documentation

## Description of change

this would have helped with https://github.com/racket/racket/pull/5239

### DIFF
--- a/build.md
+++ b/build.md
@@ -321,6 +321,8 @@ make
 ./racket/bin/raco docs raco
 ```
 
+Note that the raco built by `make base` does not support the `docs` subcommand. You have to do a full build.
+
 #### 1.7.3. Installing Packages
 
 After you’ve built and installed minimal Racket, you could install
--- a/build.md
+++ b/build.md
@@ -18,8 +18,9 @@ and how to contribute to Racket development.
 >> [1.6 More Instructions: Building Racket CS and Racket BC](#16-more-instructions-building-racket-cs-and-racket-bc)  
 >> [1.7 Even More Instructions: Building Racket Pieces](#17-even-more-instructions-building-racket-pieces)  
 >>> [1.7.1 Building Minimal Racket](#171-building-minimal-racket)  
->>> [1.7.2 Installing Packages](#172-installing-packages)  
->>> [1.7.3 Linking Packages for In-Place Development Mode](#173-linking-packages-for-in-place-development-mode)  
+>>> [1.7.2 Documenting Packages](#172-documenting-packages)  
+>>> [1.7.3 Installing Packages](#173-installing-packages)  
+>>> [1.7.4 Linking Packages for In-Place Development Mode](#173-linking-packages-for-in-place-development-mode)  
   
 > [2 Distributing Racket Variants](#2-distributing-racket-variants)  
 >> [2.1 Running Build Farms](#21-running-build-farms)  
@@ -308,7 +309,19 @@ bytecode to be machine-independent, which is mainly useful when the
 generated installation will be used as a template for other platforms or
 for cross-compilation.
 
-#### 1.7.2. Installing Packages
+#### 1.7.2. Documenting Packages
+
+Packages shipped with the racket base installation, such as `raco`, are documented in this repository.
+For these packages, `raco docs` links to the installation `raco` came from.
+As a result, `raco docs` show you the docs for your *install*, not the docs for this repository.
+
+To build docs for raco itself, first build raco from source, then run it:
+```
+make
+./racket/bin/raco docs raco
+```
+
+#### 1.7.3. Installing Packages
 
 After you’ve built and installed minimal Racket, you could install
 packages via the package-catalog server, completely ignoring the content
@@ -327,7 +340,7 @@ before the default package catalogs, specify the catalog’s URL as the
 
   `make .... SRC_CATALOG=<url>`
 
-#### 1.7.3. Linking Packages for In-Place Development Mode
+#### 1.7.4. Linking Packages for In-Place Development Mode
 
 With an in-place build, you can edit packages within `"pkgs"` directly
 or update those packages with `git pull` plus `raco setup`, since the


benknoble:
Rather ironically, I think you actually want to modify `pkgs/racket-build-guide/build.scrbl` and then `make -C pkgs/racket-build-guide`, or something. (Which is sort of documented in `pkgs/racket-build-guide/Makefile`, but could use a note elsewhere if it doesn't exist.)

shhyou:
FWIW, the `docs` command is installed by [the `racket-doc` package](https://github.com/racket/racket/tree/master/pkgs/racket-doc).

samth:
I think the problem that led to this is addressed in some sense already in the build guide, here: https://docs.racket-lang.org/racket-build-guide/contribute.html#%28part._main-contribute%29 That is, that the correct way to contribute to this repository is to first build Racket from a checkout of this repository. Is there something that should be added to the build guide to make that clearer, or to clarify what is in this repository beyond just the Racket runtime and compiler?

benknoble:
Would a cross-link from the Build guide to these explanations in the Contributing guide help? That seemed like a good TL;DR of the ways things are separate-but-together.

It might also help to clarify how to get from a minimal build to something more complete?

samth:
Based on what @jyn514 wrote, I think the issue was trying to use a downloaded build of Racket to contribute to this repository, which gets you into lots of problems. Unfortunately because (a) downloading Racket is how almost everyone gets Racket to start with and (b) building Racket takes a long time, this happens repeatedly to people who want to contribute. I'm not sure what the best way to address that is though.

benknoble:
> That document is the build guide. You can read it here too. https://github.com/racket/racket/blob/master/build.md#3-contributing-to-racket-development

Doh!

benknoble:
> Based on what @jyn514 wrote, I think the issue was trying to use a downloaded build of Racket to contribute to this repository, which gets you into lots of problems. Unfortunately because (a) downloading Racket is how almost everyone gets Racket to start with and (b) building Racket takes a long time, this happens repeatedly to people who want to contribute. I'm not sure what the best way to address that is though.

My understanding, but maybe I'm conflating 2 separate issues, was that the OP used `make base` _and_  needed to use built raco. So really 2 problems solved by a single (attempted) doc change.

samth:
Right the described change is about how to use the result of what you just built (as opposed to the other version of Racket). Maybe a description of how to use that when you have already installed a different version of Racket is needed (not specific to documenting or `raco`)?

jyn514:
i have several thoughts here.

it would be good to mention that build.md is generated by `make -C pkgs/racket-build-guide` and mirrored as HTML at https://docs.racket-lang.org/racket-build-guide/; it wasn't clear to me and it sounds like it wasn't clear to Ben either, even though Ben works on racket quite a lot. that also helps explain to me some of the design decisions (e.g. using scribble instead of plain markdown).

> I think the problem that led to this is addressed in some sense already in the build guide, here: https://docs.racket-lang.org/racket-build-guide/contribute.html#%28part._main-contribute%29 That is, that the correct way to contribute to this repository is to first build Racket from a checkout of this repository. Is there something that should be added to the build guide to make that clearer, or to clarify what is in this repository beyond just the Racket runtime and compiler?

it is possible that section implies it. it was not clear to me from skimming that it also applies to things besides the `racket/bin/*` executables; i don't think the docs are generated by plain `make`. in particular someone on discord (much more familiar with the racket ecosystem than i!) suggested running `raco docs raco-exe`, which *looks* like it works, but actually does the wrong thing.

i don't think there is any harm in having duplicate documentation, especially when the documentation is specialized to the exact thing someone might want to do. if we can also extend the general documentation to more clearly imply this, then great. but i think being very explicit and direct is a good thing in documentation.

i would also like to mention that the docs are split up between section 1 (which seems to be targeted at installing from source, but also has docs on building for devs) and section 3 (which is explicitly targeted at devs). it would be nice to condense that somehow.

> Based on what @jyn514 wrote, I think the issue was trying to use a downloaded build of Racket to contribute to this repository, which gets you into lots of problems. Unfortunately because (a) downloading Racket is how almost everyone gets Racket to start with and (b) building Racket takes a long time, this happens repeatedly to people who want to contribute. I'm not sure what the best way to address that is though.

kind of. i assumed to start that `raco build` and `raco docs` would not work. i skimmed through the readme and could not find documentation for how to build the docs for raco itself. at that point i started skimming the makefile, assumed that `make server` would generate a webserver with the docs, and ran that. that did not build the docs, and put my build tree in a strange state (the makefile called it "server mode") that i had to run `git clean -dfx` to get out of. then i tried `make base`, which build a `raco` that did not have a `docs` subcommand. and only then did i run a full `make`, which took quite a while, but did build a raco that supported what i wanted.

my takeaways from that:
- is there any way we can get away from "modes"? it makes sense to me that people would want to do partial builds of the repo, but why does that require changing the state in such a way that i can't get back to a full build?
- can we have a partial build that has a `raco docs` subcommand? it doesn't have to necessarily be `make base`, but it would be nice not to have to bootstrap a full distribution to generate HTML 😓
- it would really be nice to document how to build the docs. it was not at all clear to me a-priori how to do so, *even assuming* that my existing install of racket wouldn't work.

> Right the described change is about how to use the result of what you just built (as opposed to the other version of Racket). Maybe a description of how to use that when you have already installed a different version of Racket is needed (not specific to documenting or `raco`)?

yes, it would be nice to document where all the output paths are. i ran `make` and didn't know what had been built, and it was worse because i'm not familiar with racket package structures and didn't think to look in `./racket/`.

it sounds to me like you don't want this to be specific to `docs` and `raco`, and i am not entirely sure why. if you like i am happy to add documentation for other subcommands, and to note that "docs raco" is just an example and will work for other packages and subcommands.

samth:
I'm just going to describe how things work first, and then we can discuss how to better document them, since I think the problem is that we already think we documented them correctly.

1. The build guide says:
```
The main Racket repository also has the source to the Racket Reference, Racket Guide, and other core-ish documentation, including the source to the document that you are reading. Those document sources are in the repository’s "pkgs" directory.

Finally, the main repository includes a few other packages that are especially tightly bound to the runtime-system implementation, such as the "compiler-lib" package or the "racket-test" package. Those package sources are also in the repository’s "pkgs" directory.

To develop improvements to any of those parts of Racket, following the usual GitHub-based workflow:

* Fork the Racket repository.

* Create an in-place build as described in [Building Racket from Source](https://github.com/racket/racket/blob/master/build.md#1-building-racket-from-source).

* Make your changes and rebuild with make or make as-is or raco setup, where raco setup is the best choice when modifying Racket libraries that are in "collects" or a package. If your changes involve modifying things that are part of the racket executable, then a simple make may not suffice; see “Modifying Racket” in "racket/src/README.txt" for more information.

* Commit changes to your fork and [submit a pull request](https://help.github.com/en/articles/creating-a-pull-request).
```

The first paragraph is supposed to indicate that contributing to the documentation for most `raco` commands is included in that list, and that therefore you should follow the listed steps.

2. The documentation, including the documentation you were editing, is in fact built when you run `make`, because `make` in its last step runs `raco setup` which includes building the documentation.

3. One problem is that you ran `make server`, which you effectively never want to do; that's part of the steps for building your own Racket distribution. Which step of the documentation led you to try that? This is the only "mode", as you say, and the fact that it breaks your repository is indeed unfortunate, but that's somewhat alleviated by the fact that you never need to do it.

4. Running `make base` indeed generates a minimal version of Racket, which isn't what you want either (unless you're just changing the runtime or compiler, and usually not then either). Maybe the documentation should more explicitly indicate that shortcuts are not a good idea. 

5. If you want to build just the reference and not other things, you could run `make PKGS=racket-doc` but note that this will not speed up the build as much as you would hope. 

6. What I mean by not specific to this issue is that so far I don't see what is specific to documenting `raco` as opposed to documenting something else that's part of this repository, like `append`, or making non-documentation changes (like changing how `append` or `raco` are implemented). 

7. Where `make` puts the results is documented in the build guide; see 1.2 under "In-place build", the first bullet.
 
One thing that this discussion demonstrates is that the build guide contains all the information but because it contains a lot of information the right piece can be hard to find.

The other important point is that for contributing to _this repository_, the guide is trying to say the following:

> Run `make`. Ignore any existing Racket installations you have, and use the resulting build, which is in `./racket/`. After your change, run `make` again to build everything. This will potentially take a long time. Do not try to take shortcuts. 

I'm not sure if that's a helpful summary, though, or where in the build guide it should go.

jyn514:
that's very helpful, thanks. i didn't realize `make` also generated docs. i see now that that's documented under https://github.com/racket/racket/blob/master/build.md#31-main-repository-contributions. 

(as an aside, is there a way to search on all pages for scribble's html output? that would have taken me a while to find with the html interface. i know about https://docs.racket-lang.org/search/index.html?q=documentation but not how to limit it to the build guide.)

could we extend either https://github.com/racket/racket/blob/master/build.md#13-quick-instructions-in-place-build or https://github.com/racket/racket/blob/master/build.md#15-more-instructions-building-racket to say *what* is built? it's documented in the contribution section, and in https://github.com/racket/racket/blob/master/build.md#14-quick-instructions-unix-style-install, but it was not clear to me that `make` alone would generate docs; and i assumed 1.4 didn't apply to me because i wasn't doing a unix-style install.

could we extend https://github.com/racket/racket/blob/master/build.md#171-building-minimal-racket to clarify what "minimal racket" means? i thought it would be the same as https://download.racket-lang.org/ and it was not. i am still not exactly sure the difference between `make` and `make base`.

---

> One problem is that you ran make server, which you effectively never want to do; that's part of the steps for building your own Racket distribution. Which step of the documentation led you to try that? This is the only "mode", as you say, and the fact that it breaks your repository is indeed unfortunate, but that's somewhat alleviated by the fact that you never need to do it.

i had a hard time navigating the docs so i gave up and read through the makefile instead. the docs did not tell me to run `make server`.

> One thing that this discussion demonstrates is that the build guide contains all the information but because it contains a lot of information the right piece can be hard to find.

it also documents a lot of things "by implication". building the docs is documented as:
> The main Racket repository also has the source to the Racket Reference, Racket Guide, and other core-ish documentation, including the source to the document that you are reading. Those document sources are in the repository’s "pkgs" directory. [...]
> To develop improvements to any of those parts of Racket, following the usual GitHub-based workflow: [...]
> Make your changes and rebuild with make

this requires reading around 6 paragraphs in and around those docs; which is ok, except that it's not clear while skimming that you *need* to read those paragraphs. it also doesn't mention the raco docs specifically, which makes it hard to tell that they're included in "other core-ish documentation".

there are other things like this; for instance `make PKGS=racket-doc` is documented as "To install a subset of the packages that would otherwise be installed, supply a PKGS value to make". `racket-doc` is in the `pkgs/` folder, and it does say "The "pkgs" directory contains packages", so the info is there, but i did not assume that "installing a package" includes building its documentation and that's not stated explicitly.

---

in general, i think a lot of the confusion is because i come from a background where "building" is always a separate step from "installing", and i just immediately dismissed all the "installing" docs as not relevant to me. i'm not sure how to clarify that.

jyn514:
> This seems like a reasonable thing to add, although I'm not sure exactly how to phrase it. Maybe "everything in the racket distribution including documentation and tests"? This is sort of what it says under "in-place build" in section 1.2

1.2 says this:
> It creates a build in the "racket" subdirectory and installs packages that you specify (or the "main-distribution" plus "main-distribution-test" package by default).

again, i really feel like this is very implicit: it's using "build" and "install" as synonyms, and it does not mention that it includes anything other than the executables themselves. i think "builds everything in the racket distribution including documentation and tests" would be much more clear.

> Minimal Racket as described in the build guide should be basically the same as what you get by downloading Minimal Racket from https://download.racket-lang.org. What difference did you see?

it generated a `raco` exe that did not know how to run the `raco docs` subcommand. see https://discord.com/channels/571040468092321801/618895179343986688/1355207189114916995. i expected `make base` not to contain docs but i did not expect `raco` to have different abilities based on the make command i ran.

> Do you have suggestions for making it clearer what to read in your situation?

maybe we can add more links *between* the sections? for instance, section 1 is titled "Building Racket from Source", so i assumed it was what i wanted; if it mentioned it was mostly for installing to the system, and had linked to the "Main-Repository Contributions" section, it would have saved me a lot of time.

> The other thing is that the build process is actually very simple unless you are trying to do less work because you're impatient (I certainly fall into this category regularly). Maybe that should be emphasized?

i think that would be great, yeah. maybe we can add an 'introduction' section that mentions that, as well as that "building" and "installing" are used as synonyms?

hmm actually i see that it already says "you can try just running make but we recommend that you at least consider the information [...]" - maybe we can make 🥁 that suggestion stronger?

> In general this seems like it lines up with my general sense that people get confused by the build guide by bringing ideas from elsewhere and thus not reading the relevant parts or trying to get around things; maybe the docs should more explicitly advise against this.

well, i feel like the guide is the one making assumptions, not me 😅 it's true i missed all of section 3, but i think the docs as a whole would be much more clear if they consistently stuck to saying "build" instead of switching between "build" and "install", and explicitly called out that using a global `raco` already installed on the system won't work. the second thing especially seems like a mistake that people would often make.

samth:
> again, i really feel like this is very implicit: it's using "build" and "install" as synonyms, and it does not mention that it includes anything other than the executables themselves. i think "builds everything in the racket distribution including documentation and tests" would be much more clear.

I think it actually isn't using "build" and "install" as synonyms and that's part of the issue. It's creating a _build_ of Racket (that is, compiling Racket and creating an installation with the relevant config files) and then _install_ing a bunch of packages. 

What about the following:

> It creates a build in the "racket" subdirectory including, by default, the entire default Racket distribution, with documentation and tests (specifically, it installs the "main-distribution" plus "main-distribution-test" package). You can also specify a different set of packages to build; see the discussion of `PKGS` below.

------

> it generated a raco exe that did not know how to run the raco docs subcommand. see https://discord.com/channels/571040468092321801/618895179343986688/1355207189114916995. i expected make base not to contain docs but i did not expect raco to have different abilities based on the make command i ran.

This is exactly the behavior of Minimal Racket if you download it. `raco` is extensible and installing more packages (for `docs`, it's implemented in the `scribble-lib` package).

----------

> maybe we can add more links between the sections? for instance, section 1 is titled "Building Racket from Source", so i assumed it was what i wanted; if it mentioned it was mostly for installing to the system, and had linked to the "Main-Repository Contributions" section, it would have saved me a lot of time.

I think more links between sections would be good although I think the "Building Racket from Source" section is exactly what you wanted (eg, just run `make`). 

-------

> i think that would be great, yeah. maybe we can add an 'introduction' section that mentions that, as well as that "building" and "installing" are used as synonyms?

I don't think "building" and "installing" are used as synonyms, but (a) installing packages does compile things (that seems normal though) and (b) there's not usually an "install" step for Racket itself. 

------

> hmm actually i see that it already says "you can try just running make but we recommend that you at least consider the information [...]" - maybe we can make 🥁 that suggestion stronger?

Yes I definitely think making that suggestion stronger would be good.


--------

> explicitly called out that using a global raco already installed on the system won't work

Yes calling this out explicitly would be a good idea.

benknoble:
> > i think that would be great, yeah. maybe we can add an 'introduction' section that mentions that, as well as that "building" and "installing" are used as synonyms?
> 
> I don't think "building" and "installing" are used as synonyms, but (a) installing packages does compile things (that seems normal though) and (b) there's not usually an "install" step for Racket itself.

with particular emphasis on

> there's not usually an "install" step for Racket itself.

It's possible I've misunderstood you, but I seem to have the impression that most Racket developers use `/path/to/clone/of/racket/bin`. On the contrary, I use whatever I get from my packager, which presumably is taking an install step; further, I would be shocked if there was no way to go from `racket/bin/<stuff>` + packages to a (Unix-like?) install into `PREFIX`, given `make unix-style PREFIX=…` is supported.

On a closer read, though, you can't just go from `make` to `install` with the right arguments; it seems you have to `make unix-style PREFIX=… DESTDIR=…` instead.

Speaking of which, it's starting to feel like `in-place` and `unix-style` are also different "modes" 😅 There's 

**TL;DR** The way that (platform) packagers and package-consumers see Racket looks a little more like a traditional `make && make install`, right? Which suggests there is (or can be) an install step for Racket. But the way that Racket hackers see Racket is the "in-place" mode, which intentionally lacks an install step?

jyn514:
> One challenge is that the current build guide is trying to serve multiple constituencies -- people who want to download, compile, and install Racket; people who are packaging Racket for others; people who want to contribute to Racket.
> 
> Maybe an organization of the build guide based more around what people want to do, or a guide to it, would help. But I worry about adding more; it seems pretty clear that the existing length is already a barrier to people getting what they need from it.

fwiw, rust-lang/rust handles this by just fully putting the guides for different constituencies into different documents (which it calls INSTALL.md and CONTRIBUTING.md): 
image

benknoble:
Just a note that some reorganization around goals fits the spirit of [diátaxis](https://diataxis.fr), though I'm not suggesting that we adopt that approach specifically. (Still, the Guide/Reference split prominent in Racket docs is wonderful; perhaps the Build Guide needs a companion Reference and a simpler edition?)

mflatt:
I took a fresh look at the Build Guide after reading through this thread. Here are the main things I understood and the corresponding changes in commit 7722ce74308e2c101dfeb60e310994a36ac823ae:

 * The way that "build" is a kind of "install" for in-place builds was not clear. The new version makes the intended equivalence explicit.
 * More goal-oriented information is needed. The various build options in section 1.2 now start with italicized text to explain what that option is for.
 * The "Contributing to Racket Development" section is easily overlooked. Along the same lines as goal signposting, a few places in section 1 now tell a reader when to skip to section 3.
 * HTML is more readable than Markdown, and it's not clear why both forms of the document exist. The Markdown rendering now starts out with a pointer to the HTML version at https://docs.racket-lang.org (with a suitable caution that it may not match the sources you're reading).

benknoble:
I can't speak for @jyn514 but I'm happy with @mflatt 's recent changes.

There may or may not still be some missing info on "how to document raco" (or more generally, packages developed here), but perhaps now we can find a good place for that? I'd want to reread to see if that's already present though.

jyn514:
i agree that @mflatt's changes help a lot, in particular "Building and installing packages implies that documentation provided by those packages is built and locally installed, too" is very helpful.

happy to close this.

PR #5243, opened 2025-04-02
VT Input Mouse Support for Windows Console

dzoep:
## Checklist

- [x] Feature

## Description of change

This PR adds support for translating MOUSE_EVENT_RECORD[1] structures in windows console input event buffer to VT console sequences as generated by most modern terminal emulators in the SGR mouse reporting mode[2] (see the SET_SGR_EXT_MODE_MOUSE). The mechanism is the same as employed by the code already present on windows terminals which do not support VT sequences for cursor keys (see the code just above the main section added in `expeditor.c`. The static buffer had to be extended to allow storing strings which may be as long as "\e[<99;999;999m".

By default this does not inject any character sequences into the input because the windows console must be programmatically configured to send them. For an example of how this can be achieved after setting the terminal into "raw mode", see [3] which contains an example of appropriate GetConsoleMode/SetConsoleMode based on Microsoft's documentation.

[1] https://learn.microsoft.com/en-us/windows/console/mouse-event-record-str
[2] https://www.invisible-island.net/xterm/ctlseqs/ctlseqs.html#h2-Mouse-Tracking
[3] https://gitlab.com/racketeer/tui-term/-/blob/master/term/tty-input-port.rkt?ref_type=heads#L54


shhyou:
Closing because I think mflatt has already merged dzoep's PR (cisco/ChezScheme#930) into Racket's Chez copy through 55d986dcd3602457994a44f2fafedffad4f5229d.

PR #5244, opened 2025-04-03
reference: clarify `datum->syntax` recursive conversion

usaoc:
## Checklist
- [x] documentation

## Description of change
Say only immutable prefab structures are recursively converted.


mflatt:
Thanks!

PR #5246, opened 2025-04-06
XML element tag should not be allowed to be empty

joergen7:
| attribute        | value                                |
|------------------|--------------------------------------|
| Racket version   | 8.16.0.4-2025-04-03-14fc734705 [cs]  |
| #lang            | racket/base                          |
| collection       | xml/private/xexpr-core               |
| tags             | bug                                  |

### Observation

In the `xml` collection, an XML element's name is given as a symbol. E.g., the following code verifies and formats a simple element:

```racket
#lang racket/base

(require xml)

(define e '(myelement))

(values
 (xexpr? e)
 (xexpr->string e))

```

In Racket, symbols can be empty `'||`, just like strings `""`. However, XML element tags are not allowed to be empty. Nevertheless neither `xexpr?` produces `#f` nor does `xexpr->string` decline its input:

```racket
#lang racket/base

(require xml)

(define e '(||))

(values
 (xexpr? e)
 (xexpr->string e))

```

In my opinion, this case is relevant, because oftentimes users generate element tag names dynamically.

### Expectation

`(xexpr '(||))` should be `#f`. `(xexpr->string '(||))` should raise an error.

### Approach

In the definition of the function `incorrect-xexpr?` replace the call to `symbol?` with a call to `non-empty-symbol?`. We introduce `non-empty-symbol?` analogous to the function `non-empty-string?` in `racket/string`.

### DIFF
--- a/racket/collects/xml/private/xexpr-core.rkt
+++ b/racket/collects/xml/private/xexpr-core.rkt
@@ -3,6 +3,8 @@
          racket/contract/base
          racket/contract/combinator
          racket/pretty
+         (only-in racket/string
+                  non-empty-string?)
          "core.rkt")
 
 
@@ -75,6 +77,11 @@
       (true-k)))
 
 
+(define (non-empty-symbol? x)
+  (and
+   (symbol? x)
+   (non-empty-string? (symbol->string x))))
+
 ;; incorrect-xexpr?: any -> (or/c #f exn:invalid-xexpr)
 ;; Returns an exn:invalid-xexpr if the xexpr has incorrect structure.
 ;; Otherwise, returns #f.
@@ -94,7 +101,7 @@
              (current-continuation-marks)
              x)]
            [else
-            (if (symbol? (car x))
+            (if (non-empty-symbol? (car x))
                 (cond [(has-attribute-pairs? x)
                        (define maybe-exn (erroneous-attribute-pairs? (cadr x)))
                        (cond [maybe-exn maybe-exn]


racket/collects/xml/private/xexpr-core.rkt
```diff
@@ -75,6 +77,11 @@
       (true-k)))
 
 
+(define (non-empty-symbol? x)
+  (and
+   (symbol? x)
+   (non-empty-string? (symbol->string x))))
```
Bogdanp, line 83: Nit: I would replace `symbol->string` with `symbol->immutable-string` to avoid an unnecessary allocation and copy. 

The `symbol->string` implementation is [about 10x slower](https://gist.github.com/Bogdanp/9ada3e8dec7e274d47c38e971b5bb934) compared to an `eq`-based implementation (which I assume you didn't go with due to `(string->uninterned-symbol "")`), and a `symbol->immutable-string` version is only 5x slower.

Also, it's probably worth adding a couple of tests for this change.

rfindler: Maybe a short-circuit with and `eq?` then? Something like this?
```
#lang racket/base
(require racket/symbol)

(define (non-empty-symbol? x)
  (cond
    [(symbol? x)
     (cond
       [(symbol-interned? x)
        (not (eq? x '||))]
       [else
        (not (= 0 (string-length (symbol->immutable-string x))))])]
    [else #f]))

(require rackunit)
(module+ test
  (check-true (non-empty-symbol? 'x))
  (check-false (non-empty-symbol? '||))
  (check-false (non-empty-symbol? 0))
  (check-false (non-empty-symbol? (string->uninterned-symbol "")))
  (check-true (non-empty-symbol? (string->uninterned-symbol "x"))))
```

mflatt:
My impression is that the XML doesn't constrain symbols at all, so it allows characters that would not be allowed as well as empty symbols. For example, it allows `=` as an element name. Given the absence of other checking, disallowing the empty symbol seems a little arbitrary.

If I'm not mistaken about that, then adding a mode that constrains element names to allowed forms could be useful, and then checking for non-empty symbols should be part of that.

racket/collects/xml/private/xexpr-core.rkt
joergen7: I have gone with @rfindler's suggestion. I'm unsure if I have put the unit tests in the right spot though. If someone has a better place than a `(module+ test ...)` in the same file, I'm happy to go with that.

LiberalArtist:
> ```racket
> (string->xexpr (xexpr->string x))
> ```
> should give me the same expression back.

Even setting aside well-formedness issues, this (reasonable!) wish quickly encounters issues defining what “the same” is. There are many different  non-`equal?` x-expression representations that represent the same data model. Consider the following:
```racket
'(demo () "Hello" "," " " "World")
'(demo "Hello, World")
```

(That's without even getting into the way that the `xml` library distinguishes different concrete syntaxes that are “the same” in terms of the XML Infoset.)

I can definitely see the value in preventing ill-formed XML, but I suspect there are too many existing uses of x-expressions that accept or even rely on the ability to use them as an abstract data model, without reference to the constraints of serialization  into XML concrete syntax. I think such functionality would need to be opt-in, either with new functions or a new library (even just a wrapper).

joergen7:
I see your point. The argument that not all x-expressions, in fact, /need/ to be encodable and decodable because they are still valid data representations in their own right, would make this pr, in fact, obsolete. If no pro-argument for narrowing the definition for `xexpr?` surface, I would close the pr.

joergen7:
@rfindler I'm trying to apply property-based testing to find bugs in Racket. The assertion that `xexpr->string` and `string->xexpr` are each other's inverse looked like an obvious candidate to me.

I am sceptical myself whether it was even feasible (or even desired) to come to a predicate `xexpr?` that is congruent with valid XML.

From this, I plan to move on to find a valid XML document for which `xexpr?` is false. Because I'm sure if I can find such an example this would be more interesting.

But I'm open to suggestions.

PR #5250, opened 2025-04-17
docs: note the default imports provided by single-{unit,signature} modules

benknoble:
For many readers, this was a reasonable implicit assumption. It's good to be explicit, though.

This is partly in response to a Discord question on the same topic.

## Checklist

- [x] documentation

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/units.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/units.scrbl
@@ -839,9 +839,11 @@ contract.  The unit name is used for the positive blame of the contract.
 
 @section[#:tag "single-unit"]{Single-Unit Modules}
 
-When @racketmodname[racket/unit] is used as a language name with
-@hash-lang[], the module body is treated as a unit body.  The
-body must match the following @racket[_module-body] grammar:
+As a language name with @hash-lang[], @racketmodname[racket/unit] provides all
+bindings of @racketmodname[racket/unit] and @racketmodname[racket/base] except
+for @racket[%#module-begin], and the @racketmodname[racket/unit] module body is
+treated as a unit body. The body must match the following @racket[_module-body]
+grammar:
 
 @racketgrammar*[
 #:literals (import export require begin)
@@ -857,7 +859,7 @@ body must match the following @racket[_module-body] grammar:
               derived-require-form]]
 
 After any number of @racket[_require-decl]s, the content of the module
-is the same as a @racket[unit] body.
+is the same as a @racket[unit] body with access to @racketmodname[racket/base].
 
 The resulting unit is exported as @racket[_base]@racketidfont["@"],
 where @racket[_base] is derived from the enclosing module's name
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/units.scrbl
```diff
@@ -857,7 +857,7 @@ body must match the following @racket[_module-body] grammar:
               derived-require-form]]
 
 After any number of @racket[_require-decl]s, the content of the module
-is the same as a @racket[unit] body.
+is the same as a @racket[unit] body with access to @racketmodname[racket/base].
```
benknoble, line 860: I'm not in love with the specific wording, so hopefully this kickstarts a conversation about something reasonable.

mflatt: How about this as a way to start the section, instead?

> As a language name with `#lang`, `racket/unit` provides all bindings of `racket/unit` and `racket/base` except for `%#module-begin`, and the `racket/unit` module body is treated as a unit body.

benknoble: <details><summary>range-diff: Adjusted, including incorporating something similar for `racket/signature` with cross-link back to the "single-unit" section.</summary>

```
1:  b79bb62279 ! 1:  c8390d2684 docs: note the default imports provided by single-{unit,signature} modules
    @@ Commit message
         to be explicit, though.
     
      ## pkgs/racket-doc/scribblings/reference/units.scrbl ##
    +@@ pkgs/racket-doc/scribblings/reference/units.scrbl: contract.  The unit name is used for the positive blame of the contract.
    + 
    + @section[#:tag "single-unit"]{Single-Unit Modules}
    + 
    +-When @racketmodname[racket/unit] is used as a language name with
    +-@hash-lang[], the module body is treated as a unit body.  The
    +-body must match the following @racket[_module-body] grammar:
    ++As a language name with @hash-lang[], @racketmodname[racket/unit] provides all
    ++bindings of @racketmodname[racket/unit] and @racketmodname[racket/base] except
    ++for @racket[%#module-begin], and the @racketmodname[racket/unit] module body is
    ++treated as a unit body. The body must match the following @racket[_module-body]
    ++grammar:
    + 
    + @racketgrammar*[
    + #:literals (import export require begin)
     @@ pkgs/racket-doc/scribblings/reference/units.scrbl: body must match the following @racket[_module-body] grammar:
                    derived-require-form]]
      
    @@ pkgs/racket-doc/scribblings/reference/units.scrbl: body must match the following
      The resulting unit is exported as @racket[_base]@racketidfont["@"],
      where @racket[_base] is derived from the enclosing module's name
     @@ pkgs/racket-doc/scribblings/reference/units.scrbl: suffix). If the module name ends in @racketidfont{-unit}, then
    + 
      @section{Single-Signature Modules}
      
    - @defmodulelang[racket/signature]{The @racketmodname[racket/signature]
    +-@defmodulelang[racket/signature]{The @racketmodname[racket/signature]
     -language treats a module body as a unit signature.}
    -+language treats a module body as a unit signature with access to
    -+@racketmodname[racket/base].}
    ++@defmodulelang[racket/signature]{The @racketmodname[racket/signature] language
    ++treats a module body as a unit signature in the same way that
    ++@racketmodname[racket/unit] treats @seclink["single-unit"]{a module body as unit
    ++body}: it provides all bindings of @racketmodname[racket/signature] and
    ++@racketmodname[racket/base] except for @racket[%#module-begin].}
      
      The body must match the following @racket[_module-body] grammar:
      
```

</details>

benknoble:
I think the test just needs retried; it looks like a failure to access something over the net? I don't see any commits between here and latest master that touch `racket/pkgs/net-test`. cc @mflatt

samth:
I re-ran the test.

PR #5251, opened 2025-04-19
Add deprecated aliases

jackfirth:
This pull request adds a `deprecated-alias` rename transformer and a `(define-deprecated-alias alias-id target-id)` macro, which exports `alias-id` as an alias of `target-id`.

## Description of change

The `define-deprecated-alias` macro provides `alias-id` as a rename transformer that expands to `target-id`. It allows external tools to use `syntax-local-value/immediate` on uses of `alias-id` to determine that it's a deprecated alias and what identifier to use instead.

This feature is intended for use in Typed Racket and Resyntax, as Typed Racket has several macros like `struct:`, `for:`, etc. that are legacy aliases for non-colon-suffixed equivalents like `struct` and `for`. Providing a standardized mechanism for defining deprecated aliases allows library authors to coerce Resyntax into automatically migrating client code when a library renames something. A full audit of main-distribution use cases remains to be performed, as there are likely other packages besides TR that could make use of this feature.


sorawee:
Is it reasonable to use this macro in a couple of places as examples?

jackfirth:
@sorawee Yes, but all the examples I've found so far are in other repositories such as Typed Racket and Scribble.

mflatt:
This looks good to me, but the new reference section is too small/narrow to be at the top level of that manual. I think it would fit inside the "Reflection and Security" chapter (which is "security.scrbl").

jackfirth:
@mflatt Fixed. I also tweaked the docs and tests a bit, and I adjusted the implementation to make `(deprecated-alias-target (deprecated-alias id-stx))` preserve `eq?`-ness.

mflatt:
Thanks!

benknoble:
Exciting! Now to figure out tool and editor integration.

PR #5254, opened 2025-04-24
disallow the empty string to be a key in plist-dict

joergen7:
## Checklist

- [x] Bugfix
- [ ] Feature
- [x] tests included
- [ ] documentation

### Observation

It is possible to define and encode plist-dicts with the empty string as a key:

    #lang racket/base

    (require
     xml/plist
     racket/port)

    (define v0 '(dict (assoc-pair "" (array))))

    (define s
      (with-output-to-string
        (lambda ()
          (write-plist v0 (current-output-port)))))

    (displayln s)

The XML document output looks like this:

    <?xml version="1.0" encoding="UTF-8"?>
    <!DOCTYPE plist SYSTEM "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
    <plist version="0.9"><dict><key></key><array></array></dict></plist>

Hovever, when trying to read such a dict back using `read-plist`, Racket raises an obscure error.

    (call-with-input-string
     s
     (lambda (port)
       (read-plist port)))

The error message being

    cadr: contract violation
      expected: (cons/c any/c pair?)
      given: '(key)

### Approach

This pr makes `plist-value?` require dicts to have non-empty keys.

### DIFF
--- a/pkgs/racket-test/tests/xml/plist.rkt
+++ b/pkgs/racket-test/tests/xml/plist.rkt
@@ -1,0 +1,40 @@
+#lang racket/base
+
+(module+ test
+  (require
+   (only-in rackunit
+            test-case
+            check-pred
+            check-false
+            check-equal?)
+   (only-in racket/port
+            with-output-to-string
+            call-with-input-string)
+   (only-in xml/plist
+            plist-value?
+            write-plist
+            read-plist))
+
+  (define (plist->string pl)
+    (with-output-to-string
+      (lambda ()
+        (write-plist pl (current-output-port)))))
+
+  (define (string->plist s)
+    (call-with-input-string
+     s
+     (lambda (port)
+       (read-plist port))))
+
+  (define (check-inverse v0)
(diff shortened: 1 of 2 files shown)


joergen7:
@samth Here I suggest a narrowing of a contract again. This triple (`plist-value?` `write-plist` `read-plist`) is not so much different from the cases we had before, e.g., the triple (`xexpr?` `xexpr->string` string->xexpr`).

Is there a way to make the following propsition true:

For all `x` that satisfy `(plist-value? x)`: applying `read-plist` on the output of `(write-plist x)` produces `x`.

There are at least two ways I can think of:
- narrow the definition of `plist-value?` to catch all cases where `read-plist` fails
- relax `read-plist` to entail all values that could possibly satisfy `plist-value?`

This pr reflects the first option. Is the second option more viable? Is there a third way? Is the goal uninteresting to begin with?

samth:
I am definitely not the plist expert here. Maybe @mflatt has thoughts?

PR #5262, opened 2025-05-08
Make the deftech for "X-expression" point somewhere useful

otherjoel:
I frequently use the term “X-expression” in Scribble docs. With this change it's possible to use `@tech{X-expression}` and have it link somewhere that actually explains what an X-expression is and what qualifies as an X-expression.


mflatt:
Thanks!

Issue #5257, opened 2025-04-28
Package regressions for 8.17


samth:
### Build Failures
- [x] 450lang https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/450lang.txt
      Package removed

- [x] s-lib https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/s-lib.txt
      Package removed

- [x] vscode-dark-theme https://plt.cs.northwestern.edu/release-pkg-build/server/built/fail/vscode-dark-theme.txt
Package removed

### Test Failures
- [ ] g-code-tools https://plt.cs.northwestern.edu/release-pkg-build/server/built/test-fail/g-code-tools.txt
Caused by https://github.com/racket/draw/commit/a801d886f63c1382796c68c7013bd23c129b4057 which added new methods to `dc<%>` PR to fix at https://github.com/GThad/g-code-tools/pull/1
### Other Failures
- [x] typed-map (missing a dependency on `typed-map-doc`). Not a release issue.

samth:
@SuzanneSoy I would open a PR for `typed-map` but the gitea instance seems not to be responding.

samth:
I created a PR to fix the issue in `g-code-tools`: https://github.com/GThad/g-code-tools/pull/1

rfindler:
It looks like @GThad (the owner of the g-code-tools repository) has not been active on github since 2018. Hopefully they are still available to merge the pull request!

jbclements:
Many thanks for creating this issue, @samth . My inclination would not be to let the g-code-tools problem halt the release, I'm going to close this issue, and regard blockers as complete. Feel free to un-close this issue if you (any of you!) think the close is premature.

samth:
I feel somewhat conflicted here. On the one hand, I understand entirely @mflatt's feeling that it's late in the release cycle to make significant changes to `racket/draw`. On the other hand, I am sad if more packages gradually end up in a non-working state over time, as seems likely here. So I'm unsure about what to do.

jbclements:
Isn't it possible to apply the proposed fix for the next release? Is the issue the public nature of the interface change?

mflatt:
> Isn't it possible to apply the proposed fix for the next release? Is the issue the public nature of the interface change?

Right. Because the methods are added to the public interface `dc<%>`, there's no taking them back later.

[...]

PR #5264, opened 2025-05-12
racket/class: add support interface method default implementations

mflatt:
The change extends `interface` to allow a declared method to have a default implementation.

 * The syntax is `[<id> #:default <expr>]` or `[<id> <contract-expr> #:default <expr>]` in place of a plain `<id>` or `[<id> <contract-expr>]` within `interface`, where `<expr>` must have a method shape (i.e., `lambda` or similar)
 * A default method implementaiton can use `this` or call another declared method directly.
 * An interface can add or replace the default-method implementation of a superinterface. If multiple superinterfaces provide defaults for a method and no replacement is provided, then it's an error — unless the default from both superinterfaces comes from a shared ancestor interface.

Related to #5257

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/class.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/class.scrbl
@@ -125,13 +125,14 @@ The class system allows a program to define a new class (a
 ]
 
 An @deftech{interface} is a collection of method names to be
-implemented by a class, combined with a derivation requirement. A
+implemented by a class, potentially with default implementations
+some methods, combined with a derivation requirement. A
 class @deftech{implements} an interface when it
 
 @itemize[
 
- @item{declares (or inherits) a public method for each variable in the
- interface;}
+ @item{declares (or inherits) a public method for each method in the
+ interface (that does not have a default implementation);}
 
  @item{is derived from the class required by the interface, if any; and}
 
@@ -145,6 +146,9 @@ implements. Each class also implements an implicitly-defined interface
 that is associated with the class. The implicitly-defined interface
 contains all of the class's public method names, and it requires that
 all other implementations of the interface are derived from the class.
+When a class implements an interface but does not explicitly
+declare an implementation of a method that has a default implementation
+in the interface, then the default implementation is used for the class.
 
 A new interface can @deftech{extend} one or more interfaces with
 additional method names; each class that implements the extended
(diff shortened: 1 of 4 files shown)


rfindler:
I tried out a few examples and I think it makes sense. 

There is perhaps one example that I think behaves correctly but it might be worth just looking at:

This program
```
(define i<%> (interface () (m #:default (λ () "i%"))))

(define c%
  (class* (class* object% (i<%>) (super-new)) ()
    (define/override (m) "c%")
    (super-new)))

(send (new c%) m)
```

returns `"c%"` which seems okay and this program

```
(define i<%> (interface () (m #:default (λ () "i%"))))

(define c%
  (class* object% (i<%>)
    (define/override (m) "c%")
    (super-new)))

(send (new c%) m)
```

signals an error saying that the `define/override` isn't okay in `c%` because there is no `m` in the superclass to override. 

I think this makes sense but it does seem a little counterintuitive that there is going to be an `m` in `c%` if the `define/override` isn't there but we cannot override it. And the error message wording is technically correct because it says "superclass" does not provide and, indeed, it is an interface that's providing the method.

(Also, the other test cases that I tried out seem to be subsumed by the test cases in the pull request.)

mfelleisen:
Since @rfindler 's second program looks like Java, I figured I'd rewrite it: 

```
class Main {
  public static void main(String argv[]) { 
    I c = new C();
    c.m();
  }
}


interface I {
  default void m() {
    System.out.println("I");
  }
}


class C implements I {
  @Override
  public void m() {
    System.out.println("C");
  }
}    
	   
```

I wonder whether we should give credit where credit is due.

mflatt:
This is a good point, and one that I should have drawn out in the first place. The intent of `#:default` was to have something minimal to avoid compatibility problems and not have to worry about `super` calls. But the simplification may create trouble in the future, so I'll explore filling in more with `#:public` instead of `#:default`, `#:override` for overriding a superinterface implementation, and requiring `override` in implementing classes.

mflatt:
I've pushed a revision that uses `#:public` and `#:override` (for overriding a superinterface implementation) instead of `#:default`, that requires `override` instead of `public` to replace an interface-implemented method, and that allows `super` to get to a unique overridden method. Using `super` when multiple interfaces implement a method triggers an error when the `super` call is evaluated. Also, `super` is not currently supported in interface method implementations.

PR #5271, opened 2025-05-30
extracter: status output: make parameterizable

jesboat:
Previously, the `(log-status ...)` in the expander's extractor unconditionally printed to stdout. Now, it uses a parameter `current-status-output-port` which defaults to stdout. This enables tests to capture and inspect the output.

Test plan: used in next PR


benknoble:
I can’t speak for the rest of the project, but I wouldn’t personally mind if these 2 commits were together, since there’s a dependency relationship. (If you’re concerned about reviewers, it is possible—and encouraged—to review topic branches one commit at a time. It’s not something GitHub’s design particularly encourages, but that’s largely a design failure ;) )

mflatt:
Different approach in #5272

PR #5272, opened 2025-05-30
extracter: rewrite, bugfix, and test mutated-vars

jesboat:
mutated-vars traverses an expression in a linklet to discover which
variables might be mutated, to enable some optimizations in the
extraction process. Previously, it had one *known* case of silent
unsoundness; however, there were a couple bugs which caused other cases
of unsoundness.

This PR rewrites the let-values and letrec-values. First, we addrss all
the bugs (unintentional unsoundness); explicit mutation is now detected
in all cases.

For the implicit mutation, we take a conservative approach. We recognize
some specific obviously-safe instances of letrec-values, specifically,
when the rhs of a letrec is: a plain lambda, a case-lambda, the pattern
generated by a desugared keyword procedure, and non-recursive data. If
every rhs is safe, then no implicit mutations are observable. Otherwise,
we (a) conservatively output that all bindings are potentially mutated,
which is sound and seems inefficient but is fine in practice because the
effectively never happens in what the extractor is actually used for;
and (b) print a warning.

Concretely, a not-known-as-safe letrec occurrs in the actual code the
extracter is used for falls into three categories:

1. Two cases where the reference to the letrec variable is inside a
   lambda, but the lambda is not directly bound in the same letrec,
   and the (very basic) analysis can't see that the lambda won't be
   invoked until all bindings are defined;
2. One case where cross-thread synchronization prevents the variable
   from being referenced before its definition;
3. One case which is probably a bug.

(I can make a PR to fix those, at which point the warning could be made
into an error, if desired.)

There's an extensive suite of unit tests. Currently, they have to
be run manually with `(require (submod "simplify-defn.rkt" test))`,
but if the build system is modified later to run test submodules,
they'll get run, at least.

Depends on racket/racket#5271

### DIFF
--- a/racket/src/expander/extract/simplify-defn.rkt
+++ b/racket/src/expander/extract/simplify-defn.rkt
@@ -1,6 +1,7 @@
 #lang racket/base
 (require racket/list
          racket/match
+         racket/string
          "../host/correlate.rkt"
          "../common/set.rkt"
          "../compile/side-effect.rkt"
@@ -20,20 +21,47 @@
       (seteq)
       (set-union (car args) (apply union-all (cdr args)))))
 
-;; compute the variables that are the target of a set! in e
+;; Compute the variables that are the target of a set! in e. Note
+;; that, at this point, the linklet has been modified so all bindings
+;; have unique identifiers, so we can equate the binding with the
+;; identifier.
+;;
+;; In some sense, this is only used for performance optimizations;
+;; if `mutated-vars` said "everything is potentially mutated!", that
+;; would not change the meaning of programs. However, some of those
+;; optimizations are actually quite important in practice. Specifically,
+;; knowing that a variable is not mutated enables (soundly) folding
+;; `(variable-reference-constant? (#%variable-reference v))` into #t,
+;; which, after some further optimizations, allows keyword procedures
+;; to use the optimized calling convention. In fact, the analysis is
+;; sufficient for *all* keyword procedures to get optimized this way
+;; for the code that we need to schemify, which then enables dropping
+;; all of the keyword procedure handling as unused.
 (define (mutated-vars e)
   (match e
     [`(set! ,i ,e) (set-add (mutated-vars e) i)]
-    [`(let-values ,cl ,e) 
-     (define cl* (map (lambda (c)  (mutated-vars (cadr c))) cl))
-     (define binds (apply seteq (apply append (map car cl))))
-     (set-union (apply union-all (map mutated-vars cl*)) (set-remove (mutated-vars e) binds))]
-    [`(letrec-values ,cl ,e)
-     ;; UNSOUND --- assume that variables are defined before use
-     ;; (i.e., no visible implicit assignment)
-     (define cl* (map (lambda (c) (mutated-vars (cadr c))) cl))
-     (define binds (apply seteq (apply append (map car cl))))
-     (set-remove (set-union (mutated-vars e) (apply union-all (map mutated-vars cl*))) binds)]
+    [`(let-values ([(,name ...) ,es] ...) ,e)
+     (apply union-all (mutated-vars e) (map mutated-vars es))]
+    [`(letrec-values ([(,name* ...) ,bound-exprs] ...) ,body-expr)
+     (define names (apply union-all (map list->seteq name*)))
+     (define names-because-implicit-assignments/or-false
+       (for/or ([bound-expr (in-list bound-exprs)])
+         (match bound-expr
+           [`(lambda ,_ ,_) #f]
+           [`(case-lambda ,_ ...) #f]
+           [`(let-values ([(,_ ...) (lambda ,_ ,_)] ...) (case-lambda ,_ ...))
+            ; ^-- desugared form of a lambda with optional args
+            #f]
+           [els #:when (set-empty? (set-intersect (frees els) names)) #f]
+           [els
+            (reject-suspect-letrec e)
+            ;; Using all of `names` here is sound, and a more precise analysis
+            ;; would not be useful. See comments on GH PR 5272
+            names])))
+     (apply union-all
+            (mutated-vars body-expr)
+            (or names-because-implicit-assignments/or-false (seteq))
+            (map mutated-vars bound-exprs))]
     [`(lambda ,args ,e) (mutated-vars e)]
     [`(case-lambda [,args ,e] ...) (apply union-all (map mutated-vars e))]
     [`(,sym ,e ...)
@@ -43,18 +71,17 @@
     [`(quote ,_) (seteq)]
     [e #:when (or (boolean? e) (number? e) (string? e) (bytes? e))
        (seteq)]
-    [(list app ...) (apply union-all (map mutated-vars app))]
-    [(? hash?) (seteq)]))
+    [(list app ...) (apply union-all (map mutated-vars app))]))
 
 ;; compute the free variables of e
 (define (frees e)
   (match e
     [(? symbol?) (set e)]
-    [`(let-values ,cl ,e) 
+    [`(let-values ,cl ,e)
      (define cl* (map (lambda (c)  (frees (cadr c))) cl))
      (define binds (apply seteq (apply append (map car cl))))
      (set-union (apply union-all cl*) (set-remove (frees e) binds))]
-    [`(letrec-values ,cl ,e) 
+    [`(letrec-values ,cl ,e)
      (define cl* (map (lambda (c)  (frees (cadr c))) cl))
      (define binds (apply seteq (apply append (map car cl))))
      (set-remove (set-union (frees e) (apply union-all cl*)) binds)]
@@ -81,11 +108,11 @@
        [(#t) (simp e1)]
        [(#f) (simp e2)]
        [else `(if ,e0* ,(simp e1) ,(simp e2))])]
-    [`(let-values ,cl ,e) 
-     (define names (apply append (map car cl)))     
+    [`(let-values ,cl ,e)
+     (define names (apply append (map car cl)))
      (define simp-body (simplify-expr e vars (lambda (e) (or (memq e names) (safe-ref? e))) seen-defns))
      (define body-frees (frees simp-body))
-     (define cl* (filter-map 
+     (define cl* (filter-map
                   (lambda (c)
                     (define vars (car c))
                     (define rhs (simp (cadr c)))
@@ -100,7 +127,7 @@
                       [else (list vars rhs)]))
                   cl))
      `(let-values ,cl* ,simp-body)]
-    [`(letrec-values ,cl ,e) 
+    [`(letrec-values ,cl ,e)
      (define names (apply append (map car cl)))
      (define cl* (map (lambda (c) (list (car c) (simp (cadr c)))) cl))
      `(letrec-values ,cl* ,(simplify-expr e vars (lambda (e) (or (memq e names) (safe-ref? e))) seen-defns))]
@@ -131,7 +158,7 @@
 (define (simplify-definitions linklet-expr)
   (log-status "Simplifying definitions...")
   (define body (bootstrap:s-expr-linklet-body linklet-expr))
-  
+
   (define all-mutated-vars
     (for/fold ([s (seteq)]) ([e (in-list body)])
       (cond [(defn? e)
@@ -147,7 +174,7 @@
         (not (any-side-effects? e #f #:known-defns seen-defns))))
 
   (define (safe-ref? s) (hash-ref seen-defns s #f))
-  
+
   (define new-body
     (let loop ([body body])
       (cond [(null? body) null]
@@ -159,7 +186,7 @@
                     #:when (defn? d))
                (add-defn-known! seen-defns all-mutated-vars (defn-syms d) (defn-rhs d)))
              (define e (car body))
-             (define new-defn 
+             (define new-defn
                (list 'define-values (defn-syms e) (simplify-expr (defn-rhs e) all-mutated-vars safe-ref? seen-defns)))
              (add-defn-known! seen-defns all-mutated-vars (defn-syms e) (defn-rhs e))
              (cons new-defn (loop (cdr body)))]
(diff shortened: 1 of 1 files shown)


mflatt:
This looks generally good, and I agree that the early reference in`special-filename?` is a bug.

What did you have in mind as a change for the other three? I would be inclined to log the warning with `log-warning` instead of `log-status`, and then leave it at that.

Using `log-warning` should also allow the output to be captured for a test. Using `log-status` for warnings doesn't seem quite right to me.

[Edit: wrote `log-status` once when `log-warning` was intended]

jesboat:
Ah, wasn't aware of log-warning; that's probably better. I'll update that.


Two of the other cases are when the proc passed to `wrap-event` wants to return the wrapped event.

`port-closed-evt` in `io/port/close.rkt`
https://github.com/racket/racket/blob/6b4649ed489ac575c8e841f285594e62dad5bd06/racket/src/io/port/close.rkt#L61-L62
`replace-evt` in `thread/sync.rkt`
https://github.com/racket/racket/blob/6b4649ed489ac575c8e841f285594e62dad5bd06/racket/src/thread/sync.rkt#L760-L7691

I don't remember how I was thinking of addressing them before. The least objectionable ways I can think of now would be to use an explicit `set!` (which is gross, although it might be what ultimately ends up in the `.scm` anyway) or refactoring things so that the *users* of the procedure in the wrap-event pass in the wrap-event when they call it. Something like

```racket
(struct real-wrap-evt (...) ...)
(define (wrap-evt evt wrap)
  (real-wrap-event evt (lambda (self . args) (apply wrap args))))
(define (wrap-evt-wrap we)
  (lambda args (apply real-wrap-event-wrap we args)))
```



The final case is this:

`call-in-nested-thread` in `thread/nested-thread.rkt`
https://github.com/racket/racket/blob/6b4649ed489ac575c8e841f285594e62dad5bd06/racket/src/thread/nested-thread.rkt#L22-L58

The reference is `(thread-dead! t)`; I think changing it to `(thread-dead! (current-thread/in-atomic))` would suffice.

mflatt:
Thanks for explaining more how the `wrap-evt` and `nested-thread` might change. I'm still inclined to leave those alone, though, and just have the sound analysis log a potentially helpful warning.

jesboat:
&lt;philosophizing&gt;
The thing I hate about warnings (in the codebases I've spent time on, at least) is that they tend to build up, unless you have people who are *very* scrupulous. Some warning like this, hiding in the middle of a bunch of build output... not many people would notice if they introduce one (particularly because they'd have to know *which* ones were expected.) Absent  a possible plan to eliminate them, I tend not to propose new warnings (versus making them suppressible errors, but we don't have a general way to do that, here.)
&lt;/philosophizing&gt;

How about (1) modify `nested-thread`, as it's simple to so so; (2) have the analysis act soundly; (3) when the analysis notices the rhs of a binding is the wrap-evt pattern ("detected" syntactically), generate no message; (4) in other cases, error with a message saying something like "Hi! We noticed this unexpected case. Please look at what's triggering it, and either modify the source to not trigger or verify that it's not a bug and modify this function to shush about it". (Now that I think about it, it actually seems like this check might be a relatively decent static analysis for potential bugs anyway.)

samth:
On the general warning point, all stderr output in any part of the build process or running any of the code in tree is an error in CI, so it basically always gets fixed.

mflatt:
I thought the analysis is acting soundly. In the `nested-thread` and `wrap-evt` cases, it is correctly approximating those as "maybe used too early", unlike before. Is that not right?

Meanwhile, the code in the three places is as intended. Each case is using `letrec` to tie a knot that would otherwise have to be tied in a less declarative way. Rewriting to be explicit with `set!` in this case doesn't feel like an improvement to the code. But it's true that using `letrec` that way can go wrong, and we could change the language (that the expander is implemented in) to disallow using `letrec` without function right-hand sides. The three uses might be forced to use a macro that expands to `let` plus `set!`, for example, which would keep it declarative enough.

It sounds like neither of us is a fan of warnings. It's also true that I suggested `log-warning` so that someone would have to opt into looking at these an other warnings. (The warnings would not show up by default in CI.) I see the warning as an extra analysis tool that someone might run and look at — and using the warning system helps put those tools in one place, in a sense — but that doesn't cause ok programs to be rejected. I guess, if the analysis had caught reachable bugs, then it would be a stronger indicator that a flexible `letrec` is not a good language feature, and I would be more interested in changing the language. As it is, I think the warning is potentially useful and the code is ok in its current form. Anyway, that hopefully better explains my opinion here.

samth:
Does this cause the keyword function infrastructure to be included in the generated code when it wasn't previously? That's the major potential issue when making this analysis more conservative here.

jesboat:
> Does this cause the keyword function infrastructure to be included in the generated code when it wasn't previously? That's the major potential issue when making this analysis more conservative here.

It does not. I think almost all of the keyword function processing happens during expansion, anyway. There might be some vestiges of runtime support left behind (maybe the macros leave some calls to runtime support functions or something, I don't know off the top of my head). Either way, this code doesn't care; it just recognizes some obviously-to-humans safe patterns. The patterns it recognizes are ones that we humans decide are worth writing, and a pattern which handles what fancy procedures get expanded into was worth writing :)

(Planning on replying to mflatt's comment tomorrow, but the tldr is "yeah, makes sense")

jesboat:
@mflatt, I *almost* entirely agree.

> I thought the analysis is acting soundly. In the nested-thread and wrap-evt cases, it is correctly approximating those as "maybe used too early", unlike before. Is that not right?

Yes, that is correct.

> Meanwhile, the code in the three places is as intended. Each case is using letrec to tie a knot that would otherwise have to be tied in a less declarative way. Rewriting to be explicit with set! in this case doesn't feel like an improvement to the code.

The code would be uglier for the wrap-evt cases, and I'm fine with saying "the tool should adapt to the code, and not force the code to adapt to the tool" for this. :) The thing that makes me sad is that it currently introduces spurious warnings, so personally I'd be a fan of modifying the thread code (PR #5278; I think the change is pretty neutral from a readability perspective), and having mutated-vars have some variety of diagnostic which is explicitly suppressed for the event ones. Does that sound reasonable? (PR Tentatively updated for this.)

In terms of whether that diagnostic should be a warning or an error and where it gets printed, let me know what you want and i'll update the PR accordingly.

jesboat:
> > Does this cause the keyword function infrastructure to be included in the generated code when it wasn't previously? That's the major potential issue when making this analysis more conservative here.
> 
> It does not. I think almost all of the keyword function processing happens during expansion, anyway. There might be some vestiges of runtime support left behind (maybe the macros leave some calls to runtime support functions or something, I don't know off the top of my head). Either way, this code doesn't care; it just recognizes some obviously-to-humans safe patterns. The patterns it recognizes are ones that we humans decide are worth writing, and a pattern which handles what fancy procedures get expanded into was worth writing :)

(Talked with @samth out-of-band, as I had misunderstood the question. The *actual* question was "if the analysis is now sound and think that more variables are mutated, does that cause any `variable-reference-constant?` which would previously get folded to true now get folded to false, thereby causing the optimization that keyword procedures will only do when `variable-reference-constant?` is true?". I checked the generated file and confirmed that this does not happen. Additionally, I added a comment to the source about how `mutated-vars` affects keyword procedure invocations in an important way.)

racket/src/expander/run/status.rkt
```diff
@@ -26,3 +32,4 @@
          (list* "\n" (make-string (string-length prefix) #\space) s
                 (loop (+ (string-length prefix) slen)
                       (cdr vals)))])]))))
+
```
benknoble, line 35: Spurious?

racket/src/expander/extract/simplify-defn.rkt
```diff
@@ -20,20 +21,90 @@
       (seteq)
       (set-union (car args) (apply union-all (cdr args)))))
 
-;; compute the variables that are the target of a set! in e
+; This function gets called when encountering a 'suspect letrec': some
+; letrec expression where the bound variable is used somewhere inside
+; the bound expression, and we can't proove (given the information
```
benknoble, line 26: s/proove/prove

racket/src/expander/extract/simplify-defn.rkt
```diff
@@ -20,20 +21,90 @@
...
+; available here) that uses is safe: that the implicit mutation that
```
benknoble, line 27: s/uses/use

racket/src/expander/extract/simplify-defn.rkt
```diff
@@ -20,20 +21,90 @@
...
+; happens as the letrec defines the variables.
```
benknoble, line 28: This reads oddly to me, so I wonder if I missed something or if a clause is missing.

I expected to read `that use is safe: [clarify what "is safe means"]`, and I think I got `that use is safe: that the implicit mutation […].`, where I think I would have expected `is not observable` or something similar before the full stop.

racket/src/expander/extract/simplify-defn.rkt
```diff
@@ -20,20 +21,90 @@
...
+;
+; Generally, we want to emit a warning in this case, as it's possible
+; that we stumbled on an actual bug (e.g. the bug fixed in git commit
+; b9b017261bf35cc42b341ca5d3b8f17e955416c2.) However, there are also
```
benknoble, line 32: Since this commit is stable (in master), it might be nicer to use "reference formatting" (what `git show -s --format=reference` does): `b9b017261b (io: repair wrong-variable use in checking for special Windows paths, 2025-06-03)`.

racket/src/expander/extract/simplify-defn.rkt
```diff
@@ -20,20 +21,90 @@
...
+; a handful of cases where the code is (obviously to humans) safe,
+; and our analysis is too primitive to reccognize it. What we do in
```
benknoble, line 34: s/cc/c

racket/src/expander/extract/simplify-defn.rkt
```diff
@@ -20,20 +21,90 @@
...
+; this case (see GH #5272) is to act soundly and indicate a warning,
```
benknoble, line 35: I prefer links or prose to issue numbers, just in case we ever lose this record, but other people don't care so much and I'm not a stickler for it :)

benknoble (commented):
Just some minor comments I noticed since I've been following this out of curiosity.

<details><summary>range-diff with differences between recent versions (helps me review)</summary>

```
-:  ---------- > 1:  b086766edf nested-thread.rkt: change a mostly irrelevant thing
1:  135b94f8c3 = 2:  3d82d977ca extracter: status output: make parameterizable
2:  8e12905a5b ! 3:  31c9e929cc extracter: rewrite, bugfix, and test mutated-vars
    @@ Commit message
         we (a) conservatively output that all bindings are potentially mutated,
         which is sound and seems inefficient but is fine in practice because the
         effectively never happens in what the extractor is actually used for;
    -    and (b) print a warning.
    +    and (b) print a warning (but see below.)
     
    -    Concretely, a not-known-as-safe letrec occurrs in the actual code the
    -    extracter is used for falls into three categories:
    +    Concretely, not-known-as-safe letrecs occurr very rarely in the actual
    +    code that the extracter is used for. We found 4 instances across three
    +    categories:
     
         1. Two cases where the reference to the letrec variable is inside a
            lambda, but the lambda is not directly bound in the same letrec,
            and the (very basic) analysis can't see that the lambda won't be
    -       invoked until all bindings are defined;
    -    2. One case where cross-thread synchronization prevents the variable
    -       from being referenced before its definition;
    -    3. One case which is probably a bug.
    +       invoked until all bindings are defined. We detect those specific
    +       cases using questionable heuristics and suppress the warning. (Note
    +       that we don't change the result of the analysis, only suppress
    +       the warning; this is why we're okay with the hacky detection.)
     
    -    (I can make a PR to fix those, at which point the warning could be made
    -    into an error, if desired.)
    +    2. One case where cross-thread synchronization prevents the variable
    +       from being referenced before its definition. We were able to slightly
    +       modify that code.
    +
    +    3. One case which was an actual bug; fixed in b9b0172.
     
         There's an extensive suite of unit tests. Currently, they have to
         be run manually with `(require (submod "simplify-defn.rkt" test))`,
    @@ Commit message
         they'll get run, at least.
     
      ## racket/src/expander/extract/simplify-defn.rkt ##
    +@@
    + #lang racket/base
    + (require racket/list
    +          racket/match
    ++         racket/string
    +          "../host/correlate.rkt"
    +          "../common/set.rkt"
    +          "../compile/side-effect.rkt"
     @@ racket/src/expander/extract/simplify-defn.rkt: (define (union-all . args)
            (seteq)
            (set-union (car args) (apply union-all (cdr args)))))
      
     -;; compute the variables that are the target of a set! in e
    ++; This function gets called when encountering a 'suspect letrec': some
    ++; letrec expression where the bound variable is used somewhere inside
    ++; the bound expression, and we can't proove (given the information
    ++; available here) that uses is safe: that the implicit mutation that
    ++; happens as the letrec defines the variables.
    ++;
    ++; Generally, we want to emit a warning in this case, as it's possible
    ++; that we stumbled on an actual bug (e.g. the bug fixed in git commit
    ++; b9b017261bf35cc42b341ca5d3b8f17e955416c2.) However, there are also
    ++; a handful of cases where the code is (obviously to humans) safe,
    ++; and our analysis is too primitive to reccognize it. What we do in
    ++; this case (see GH #5272) is to act soundly and indicate a warning,
    ++; *except* that we suppress the warning when the code looks like one
    ++; of the cases that a human reviewed.
    ++;
    ++; If you've found your way to this comment because a warning showed up
    ++; and you're investigating :) then you can (1) set `debug` to #t below;
    ++; (2) regenerate the schemified files, which will generate the `.rktl`
    ++; linklet files in the process; (3) use the debugging info to identify
    ++; the relevant code in the linklet, and then trace that back to the
    ++; `.rkt` source; and (4) either fix the bug, if there ids one, or
    ++; otherwise add a case to this match statement to find your code and
    ++; suppress the warning. If possible, keep the match limited so that
    ++; it doesn't suppress warnings possibly-unsafe patterns unlike the
    ++; one you found, and definitely don't do anything beyond supressing
    ++; warnings because we want the analysis to remain sound.
    ++(define (maybe-warn-about-suspect-letrec letrec-expr)
    ++  (match letrec-expr
    ++    [`(letrec-values ([(,name) (,some-event-ctor ,e1s ... (lambda ,_ ,e2) ,e3s ...)]) ,body)
    ++     #:when (or (string-contains? (symbol->string some-event-ctor) "wrap-evt")
    ++                (string-contains? (symbol->string some-event-ctor) "replacing-evt"))
    ++     #:when (andmap (lambda (e)
    ++                      (set-empty? (set-intersect (frees e) (seteq name))))
    ++                    (append e1s e3s))
    ++     (void)]
    ++
    ++    [els
    ++     (log-warning "WARN: unable to show that implicit assignment in letrec is unobservable")
    ++     (log-warning "Details about this message and next steps are in the source code.")
    ++     (let ([debug #f])
    ++       (when debug
    ++         (log-status "full letrec expression: ~a" letrec-expr)))]))
    ++
     +;; Compute the variables that are the target of a set! in e. Note
     +;; that, at this point, the linklet has been modified so all bindings
     +;; have unique identifiers, so we can equate the binding with the
     +;; identifier.
    ++;;
    ++;; In some sense, this is only used for performance optimizations;
    ++;; if `mutated-vars` said "everything is potentially mutated!", that
    ++;; would not change the meaning of programs. However, some of those
    ++;; optimizations are actually quite important in practice. Specifically,
    ++;; knowing that a variable is not mutated enables (soundly) folding
    ++;; `(variable-reference-constant? (#%variable-reference v))` into #t,
    ++;; which, after some further optimizations, allows keyword procedures
    ++;; to use the optimized calling convention. In fact, the analysis is
    ++;; sufficient for *all* keyword procedures to get optimized this way
    ++;; for the code that we need to schemify, which then enables dropping
    ++;; all of the keyword procedure handling as unused.z
      (define (mutated-vars e)
        (match e
          [`(set! ,i ,e) (set-add (mutated-vars e) i)]
    @@ racket/src/expander/extract/simplify-defn.rkt: (define (union-all . args)
     -     (set-remove (set-union (mutated-vars e) (apply union-all (map mutated-vars cl*))) binds)]
     +    [`(let-values ([(,name ...) ,es] ...) ,e)
     +     (apply union-all (mutated-vars e) (map mutated-vars es))]
    -+    [`(letrec-values ([(,name* ...) ,es] ...) ,e)
    ++    [`(letrec-values ([(,name* ...) ,bound-exprs] ...) ,body-expr)
     +     (define names (apply union-all (map list->seteq name*)))
     +     (define names-because-implicit-assignments/or-false
    -+       (for/or ([bound-expr (in-list es)])
    ++       (for/or ([bound-expr (in-list bound-exprs)])
     +         (match bound-expr
     +           [`(lambda ,_ ,_) #f]
     +           [`(case-lambda ,_ ...) #f]
    @@ racket/src/expander/extract/simplify-defn.rkt: (define (union-all . args)
     +            #f]
     +           [els #:when (set-empty? (set-intersect (frees els) names)) #f]
     +           [els
    -+            (log-status "WARN: unable to show that implicit assignment in letrec is unobservable")
    -+            (let ([debug #t])
    -+              (when debug
    -+                  (log-status "(names ~a) (bound-exprs ~a) (body ~a)" name* es e)))
    ++            (maybe-warn-about-suspect-letrec e)
     +            ; Using all of `names` here is sound, and a more precise analysis
     +            ; would not be useful. See comments on GH PR 5272
     +            names])))
     +     (apply union-all
    -+            (mutated-vars e)
    ++            (mutated-vars body-expr)
     +            (or names-because-implicit-assignments/or-false (seteq))
    -+            (map mutated-vars es))]
    ++            (map mutated-vars bound-exprs))]
          [`(lambda ,args ,e) (mutated-vars e)]
          [`(case-lambda [,args ,e] ...) (apply union-all (map mutated-vars e))]
          [`(,sym ,e ...)
```

</details>
 <details><summary>range-diff with differences between versions</summary>

```
-:  ---------- > 1:  b086766edf nested-thread.rkt: change a mostly irrelevant thing
1:  135b94f8c3 = 2:  3d82d977ca extracter: status output: make parameterizable
2:  8e12905a5b ! 3:  31c9e929cc extracter: rewrite, bugfix, and test mutated-vars
    @@ Commit message
         we (a) conservatively output that all bindings are potentially mutated,
         which is sound and seems inefficient but is fine in practice because the
         effectively never happens in what the extractor is actually used for;
    -    and (b) print a warning.
    +    and (b) print a warning (but see below.)
     
    -    Concretely, a not-known-as-safe letrec occurrs in the actual code the
    -    extracter is used for falls into three categories:
    +    Concretely, not-known-as-safe letrecs occurr very rarely in the actual
    +    code that the extracter is used for. We found 4 instances across three
    +    categories:
     
         1. Two cases where the reference to the letrec variable is inside a
            lambda, but the lambda is not directly bound in the same letrec,
            and the (very basic) analysis can't see that the lambda won't be
    -       invoked until all bindings are defined;
    -    2. One case where cross-thread synchronization prevents the variable
    -       from being referenced before its definition;
    -    3. One case which is probably a bug.
    +       invoked until all bindings are defined. We detect those specific
    +       cases using questionable heuristics and suppress the warning. (Note
    +       that we don't change the result of the analysis, only suppress
    +       the warning; this is why we're okay with the hacky detection.)
     
    -    (I can make a PR to fix those, at which point the warning could be made
    -    into an error, if desired.)
    +    2. One case where cross-thread synchronization prevents the variable
    +       from being referenced before its definition. We were able to slightly
    +       modify that code.
    +
    +    3. One case which was an actual bug; fixed in b9b0172.
     
         There's an extensive suite of unit tests. Currently, they have to
         be run manually with `(require (submod "simplify-defn.rkt" test))`,
    @@ Commit message
         they'll get run, at least.
     
      ## racket/src/expander/extract/simplify-defn.rkt ##
    +@@
    + #lang racket/base
    + (require racket/list
    +          racket/match
    ++         racket/string
    +          "../host/correlate.rkt"
    +          "../common/set.rkt"
    +          "../compile/side-effect.rkt"
     @@ racket/src/expander/extract/simplify-defn.rkt: (define (union-all . args)
            (seteq)
            (set-union (car args) (apply union-all (cdr args)))))
      
     -;; compute the variables that are the target of a set! in e
    ++; This function gets called when encountering a 'suspect letrec': some
    ++; letrec expression where the bound variable is used somewhere inside
    ++; the bound expression, and we can't proove (given the information
    ++; available here) that uses is safe: that the implicit mutation that
    ++; happens as the letrec defines the variables.
    ++;
    ++; Generally, we want to emit a warning in this case, as it's possible
    ++; that we stumbled on an actual bug (e.g. the bug fixed in git commit
    ++; b9b017261bf35cc42b341ca5d3b8f17e955416c2.) However, there are also
    ++; a handful of cases where the code is (obviously to humans) safe,
    ++; and our analysis is too primitive to reccognize it. What we do in
    ++; this case (see GH #5272) is to act soundly and indicate a warning,
    ++; *except* that we suppress the warning when the code looks like one
    ++; of the cases that a human reviewed.
    ++;
    ++; If you've found your way to this comment because a warning showed up
    ++; and you're investigating :) then you can (1) set `debug` to #t below;
    ++; (2) regenerate the schemified files, which will generate the `.rktl`
    ++; linklet files in the process; (3) use the debugging info to identify
    ++; the relevant code in the linklet, and then trace that back to the
    ++; `.rkt` source; and (4) either fix the bug, if there ids one, or
    ++; otherwise add a case to this match statement to find your code and
    ++; suppress the warning. If possible, keep the match limited so that
    ++; it doesn't suppress warnings possibly-unsafe patterns unlike the
    ++; one you found, and definitely don't do anything beyond supressing
    ++; warnings because we want the analysis to remain sound.
    ++(define (maybe-warn-about-suspect-letrec letrec-expr)
    ++  (match letrec-expr
    ++    [`(letrec-values ([(,name) (,some-event-ctor ,e1s ... (lambda ,_ ,e2) ,e3s ...)]) ,body)
    ++     #:when (or (string-contains? (symbol->string some-event-ctor) "wrap-evt")
    ++                (string-contains? (symbol->string some-event-ctor) "replacing-evt"))
    ++     #:when (andmap (lambda (e)
    ++                      (set-empty? (set-intersect (frees e) (seteq name))))
    ++                    (append e1s e3s))
    ++     (void)]
    ++
    ++    [els
    ++     (log-warning "WARN: unable to show that implicit assignment in letrec is unobservable")
    ++     (log-warning "Details about this message and next steps are in the source code.")
    ++     (let ([debug #f])
    ++       (when debug
    ++         (log-status "full letrec expression: ~a" letrec-expr)))]))
    ++
     +;; Compute the variables that are the target of a set! in e. Note
     +;; that, at this point, the linklet has been modified so all bindings
     +;; have unique identifiers, so we can equate the binding with the
     +;; identifier.
    ++;;
    ++;; In some sense, this is only used for performance optimizations;
    ++;; if `mutated-vars` said "everything is potentially mutated!", that
    ++;; would not change the meaning of programs. However, some of those
    ++;; optimizations are actually quite important in practice. Specifically,
    ++;; knowing that a variable is not mutated enables (soundly) folding
    ++;; `(variable-reference-constant? (#%variable-reference v))` into #t,
    ++;; which, after some further optimizations, allows keyword procedures
    ++;; to use the optimized calling convention. In fact, the analysis is
    ++;; sufficient for *all* keyword procedures to get optimized this way
    ++;; for the code that we need to schemify, which then enables dropping
    ++;; all of the keyword procedure handling as unused.z
      (define (mutated-vars e)
        (match e
          [`(set! ,i ,e) (set-add (mutated-vars e) i)]
    @@ racket/src/expander/extract/simplify-defn.rkt: (define (union-all . args)
     -     (set-remove (set-union (mutated-vars e) (apply union-all (map mutated-vars cl*))) binds)]
     +    [`(let-values ([(,name ...) ,es] ...) ,e)
     +     (apply union-all (mutated-vars e) (map mutated-vars es))]
    -+    [`(letrec-values ([(,name* ...) ,es] ...) ,e)
    ++    [`(letrec-values ([(,name* ...) ,bound-exprs] ...) ,body-expr)
     +     (define names (apply union-all (map list->seteq name*)))
     +     (define names-because-implicit-assignments/or-false
    -+       (for/or ([bound-expr (in-list es)])
    ++       (for/or ([bound-expr (in-list bound-exprs)])
     +         (match bound-expr
     +           [`(lambda ,_ ,_) #f]
     +           [`(case-lambda ,_ ...) #f]
    @@ racket/src/expander/extract/simplify-defn.rkt: (define (union-all . args)
     +            #f]
     +           [els #:when (set-empty? (set-intersect (frees els) names)) #f]
     +           [els
    -+            (log-status "WARN: unable to show that implicit assignment in letrec is unobservable")
    -+            (let ([debug #t])
    -+              (when debug
    -+                  (log-status "(names ~a) (bound-exprs ~a) (body ~a)" name* es e)))
    ++            (maybe-warn-about-suspect-letrec e)
     +            ; Using all of `names` here is sound, and a more precise analysis
     +            ; would not be useful. See comments on GH PR 5272
     +            names])))
     +     (apply union-all
    -+            (mutated-vars e)
    ++            (mutated-vars body-expr)
     +            (or names-because-implicit-assignments/or-false (seteq))
    -+            (map mutated-vars es))]
    ++            (map mutated-vars bound-exprs))]
          [`(lambda ,args ,e) (mutated-vars e)]
          [`(case-lambda [,args ,e] ...) (apply union-all (map mutated-vars e))]
          [`(,sym ,e ...)
```

</details>

jesboat:
@mflatt friendly ping :)


me:
> mflatt:
> > Meanwhile, the code in the three places is as intended. Each case is using letrec to tie a knot that would otherwise have to be tied in a less declarative way. Rewriting to be explicit with set! in this case doesn't feel like an improvement to the code.
> 
> The code would be uglier for the wrap-evt cases, and I'm fine with saying "the tool should adapt to the code, and not force the code to adapt to the tool" for this. :) The thing that makes me sad is that it currently introduces spurious warnings, so personally I'd be a fan of modifying the thread code (PR #5278; I think the change is pretty neutral from a readability perspective), and having mutated-vars have some variety of diagnostic which is explicitly suppressed for the event ones. Does that sound reasonable? (PR Tentatively updated for this.)
> 
> In terms of whether that diagnostic should be a warning or an error and where it gets printed, let me know what you want and i'll update the PR accordingly.

mflatt:
Sorry for the delay. I have been traveling the past few weeks, and I figured I might be able to think more clearly about this after getting back home.

After getting back and thinking for another day, here are options that I could get behind, in order or my preference:

 * Adjust this PR to use `log-warning` instead of `log-status` and leave it at that.

   There would be a potentially helpful diagnostic in place, but someone interested would need to turn on warning output to see it. They'd also need to look past three false positives, at least.

 * Change the language for extracted code to allow only the currently recognized `letrec` patterns, make the warning an error, and change code in three places to use `set!` explicitly.

 * Like the second option, but using `(current-thread/in-atomic)` as in #5272 instead of `set!`. (I prefer just always using `set!` when `define` is disallowed, but I don't feel strongly on this point.)
 
The option that I've left out is having an analysis that is expected to fail sometimes and where there's a suggestion that the analysis should be adjusted to avoid an error. We should just have a well-defined language. Meanwhile, potentially helpful analyses (related to the language's shortcomings) should be warnings in the sense of `log-warning`.

In any case, I'm eager to get this PR's correction to soundness in place!

mflatt:
@jesboat — I pushed a concrete verson of the middle bullet in my suggestion: https://github.com/mflatt/racket/tree/extracter-mutated-vars-bugfix
Would this be ok with you to merge?

jesboat:
> @jesboat — I pushed a concrete verson of the middle bullet in my suggestion: https://github.com/mflatt/racket/tree/extracter-mutated-vars-bugfix Would this be ok with you to merge?

Hi, sorry for the delay... I typically treat things like this as highly asynchronous, didn't realize you were actively waiting.

I split your commit into two (one for the change to io and thread (with you marked as author) and the expander (with me marked as author)) and updated the commit messages. That's pushed as the PR's content now. The tree is identical to https://github.com/mflatt/racket/tree/extracter-mutated-vars-bugfix.

mflatt:
> I typically treat things like this as highly asynchronous, didn't realize you were actively waiting.

No problem! I found things to do meanwhile. :)

Thanks for the updates, and I expect to merge soon.

PR #5275, opened 2025-06-01
io: fix fallback for (find-system-path 'host-addon-dir)

Bogdanp:
Without this change, a call to `(find-system-path 'host-addon-dir)` would end up in a call to `(find-system-path #f)`, because the `addon-dir` variable is initialized to `#f`.


sorawee:
Thank you!

PR #5276, opened 2025-06-03
doc: fix incorrect references in `racket/signature`

NoahStoryM:
## Checklist

- [ ] Bugfix
- [ ] Feature
- [ ] tests included
- [x] documentation

## Description of change

- Replaced incorrect reference to `sig-spec` with the correct `sig-elem`.
- Updated the link from `@secref["creatingunits"]` to `@racket[define-signature]` to reflect the correct grammar reference.


mflatt:
Thanks!

PR #5278, opened 2025-06-05
nested-thread.rkt: change a mostly irrelevant thing

jesboat:
Right now, there's a bit of code which does:

    (define t
      (make-thread
        (lambda ()
          (wait-for-event)
          ...
          (do-stuff t))))
    (trigger-the-event)

This results in a knot getting tied with a letrec, analogously to:

    (letrec ([t undefined])
      (set! t (make-thread ...))
      ...)

This is fine from a correctness perspective, but the algorithms in the extractor aren't nearly powerful enough to reason about threading primitives (see PR #5272) and we'd like to be able to add a warning for cases where the expander can't prove safety. In this case, we can tweak the code *very* slightly: change a place where the thread refers to itself by using `(current-thread/in-atomic)` instead of `t`.

### DIFF
--- a/racket/src/cs/schemified/thread.scm
+++ b/racket/src/cs/schemified/thread.scm
@@ -12413,84 +12413,85 @@
                     (let ((result_0 #f))
                       (let ((result-kind_0 #f))
                         (let ((ready-sema_0 (1/make-semaphore)))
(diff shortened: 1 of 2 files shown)


mflatt:
In #5272

PR #5280, opened 2025-06-12
add #:properties option for struct

rmculpepper:
## Checklist

- [x] Feature
- [x] tests included
- [x] documentation

## Description of change

Add a `#:properties` option to `struct` that takes an association list (same type as `make-struct-type`).

This makes `struct` a better target for macros, since the calculation of properties can be separated from the rest of the struct declaration. Multiple property values can be calculated in a single expression position, making it easier for property values to share intermediate calculations and syntax bindings.

This PR does not change the expansion of existing `struct` definitions, if any tools rely on that.

Unlike `#:property`, `#:properties` does not signal an error if used with `#:prefab`. The property list might be empty, in which case there is no conflict, and `make-struct-type` will catch the error otherwise.


mflatt:
LGTM, but it would be good to also have a test with `#:prefab` and `#:properties null`.

PR #5282, opened 2025-06-20
Make xexprs serializable via read and write

stevebyan:
To enable this Scribble pull request: https://github.com/racket/scribble/pull/498
make xexprs serializable via read and write by converting some structs to pre-fab structs.

The intent is to enable blogging systems like Greg Hendershott's "Tadpole", but using Scribble markup rather than Markdown. Greg gets away with serializing `xexpr`s because he doesn't insert comments, CDATA, or PCDATA in his Markdown source.

## Checklist

- [ ] Bugfix
- [x] Feature
- [ ] tests included
- [ ] documentation

## Description of change

Convert the comment, CDATA, and PCDATA structs to pre-fab structs.

I should add a test for serializing those structs, and amend the xexpr docs to mention that they are serializable.

### DIFF
--- a/pkgs/racket-test/tests/xml/test.rkt
+++ b/pkgs/racket-test/tests/xml/test.rkt
@@ -392,7 +392,7 @@ END
      
      (test-read-xml/element/exn
       "<!-- comment --><br />"
-      "read-xml: parse-error: expected root element - received (comment ")
+      "read-xml: parse-error: expected root element - received '#s(comment \" comment \"")
      
      (test-read-xml/element
       "<title><![CDATA[hello world[mp3]]]></title>"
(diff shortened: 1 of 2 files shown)


LiberalArtist:
> Convert the comment, CDATA, and PCDATA structs to pre-fab structs.

Using prefab structs is problematic. Currently, the constructors for these structs are provided with contracts, so (unless there are bugs in this library) a value that satisfies a predicate like `cdata?` is guaranteed to have valid contents. With prefab structs, however, anyone can construct invalid instances, e.g. `#s(cdata #"not" #"valid" #"cdata")`.

Instead, these structs should be serializable in the sense of `racket/serialize`.

It is a bit fiddly to set up contracts to be checked during deserialization, especially because we don't want the performance overhead of `struct-guard/c`. I'm drafting a PR with the hard parts.

LiberalArtist:
Also, `location` needs to be serializable because it can occur in `source-start` or `source-stop`.

LiberalArtist:
My serializable version is in https://github.com/racket/racket/pull/5283

stevebyan:
Thank you, LiberalArtist! Closing this pull request in favor of yours.

PR #5283, opened 2025-06-21
xml: make structs serializable, but enforce contracts

LiberalArtist:
This addresses the issue with https://github.com/racket/racket/pull/5282 that could allow evading contracts.

Since an X-expression can contain a `location` in a `source-start` or `source-stop`, I needed to add serialization support to `xml/private/structures` in addition to `xml/private/core`. At that point, it seemed to make sense to make the structs that aren't used in X-expressions serializable, too.

One caveat is that contract violations on deserialization will blame the `deserialize-info` submodules. That's not theoretically ideal, but seems practically fine.

Still needs docs and tests.

### DIFF
--- a/racket/collects/xml/private/core.rkt
+++ b/racket/collects/xml/private/core.rkt
@@ -1,28 +1,90 @@
 #lang racket/base
-(require racket/contract/base)
+(require racket/runtime-path
+         (for-syntax racket/base))
 
 ;; Core structures needed for `xml/xexpr'
 
-(provide (all-defined-out))
+(provide permissive-xexprs
+         (struct-out source)
+         (struct-out comment)
+         (struct-out p-i)
+         (struct-out pcdata)
+         (struct-out cdata)
+         valid-char?)
 
 ; permissive-xexprs : parameter bool
 (define permissive-xexprs (make-parameter #f))
 
+; support for serialization with contracts enforced on deserialization
+(module serialization-support racket/base
+  (require racket/serialize-structs
+           (for-syntax racket/syntax
+                       racket/struct-info
+                       racket/base))
+  (define ((cycles-not-allowed name))
+    (error name "invalid serialization;\n cycles not allowed"))
+  (define-for-syntax ((make-define-serializable-struct deserialize-info-mpi
+                                                       get-checked-constructor)
+                      stx)
+    (syntax-case stx ()
+      [(_ name-maybe-super (fld ...))
+       (with-syntax* ([(name super-fld-ref ...)
+                       (syntax-case #'name-maybe-super ()
+                         [name
+                          (identifier? #'name)
+                          #'(name)]
+                         [(name super)
+                          #`(name
+                             #,@(reverse (list-ref (extract-struct-info (syntax-local-value #'super))
+                                                   3)))])]
+                      [(name-fld-ref ...)
+                       (for/list ([id (in-list (syntax->list #'(fld ...)))])
+                         (format-id id "~a-~a" #'name id))]
+                      [deserialize-info:name-v0
+                       (format-id #'name "deserialize-info:~a-v0" #'name)])
+         #`(begin
+             (define-struct name-maybe-super (fld ...)
+               #:property prop:serializable
+               (make-serialize-info (λ (this)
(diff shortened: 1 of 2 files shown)


stevebyan:
Thank you!

LiberalArtist:
I've added docs and tests.

While working on the tests, I found that a bug in `xexpr/c`, `xexpr?`, `corrext-xexpr?`, and `validate-xexpr?` incorrectly accepted instances of the `pcdata` struct. I've changed them to match the documentation and other parts of the implementation. To be clear, actually trying to use a `pcdata` instance as an X-expression was already very broken, e.g.:
```
$ racket
Welcome to Racket v8.17 [cs].
> (require xml)
> (xexpr->string (pcdata #f #f "pcdata"))
""
```
Since this is so broken, I'm hoping it will work to make this change, but I've put it in a separate commit in case someone is somehow relying on the current behavior and we have to revert. I've put that commit *before* the one adding serialization, because, when the structs become serializable, the module path providing their deserialize info becomes part of the API. Having tried it both ways, it's much nicer to move from `(submod xml/private/structures deserialize-info)` to `(submod xml/private/core deserailize-info)` than the other way around. It can be just:
```racket
(require (only-in (submod "core.rkt" deserialize-info) deserialize-info:p-i-v0))
(provide deserialize-info:p-i-v0)
```
rather than:
```racket
(provide deserialize-info:p-i-v0)
(define deserialize-info:p-i-v0
  (dynamic-require (module-path-index-join '(submod "." deserialize-info)
                                           mpi-for-contracts)
                   'deserialize-info:p-i-v0))
```

I also included some other doc tweaks with the serialization commit. The addition of `no-external-dtd` was the result of one of those side-quests.

racket/collects/xml/private/core.rkt
```diff
@@ -1,28 +1,86 @@
...
+             (module+ deserialize-info
+               (provide deserialize-info:name-v0)
+               (define deserialize-info:name-v0
+                 (make-deserialize-info (#,get-checked-constructor name)
+                                        (cycles-not-allowed 'name))))))]))
+  (provide (for-syntax make-define-serializable-struct)))
+(require 'serialization-support)
+(define-runtime-module-path-index deserialize-info-mpi '(submod "." deserialize-info))
+(module+ deserialize-info
+  (define-runtime-module-path-index mpi-for-contracts "structures.rkt")
```
LiberalArtist, line 63: I wanted to double-check that this doesn't need to be:
```racket
(define-runtime-module-path mpi-for-contracts xml/private/structures)
```

pkgs/racket-doc/xml/xml.scrbl
```diff
@@ -115,20 +190,65 @@ Represents an attribute within an element.}
...
+]}
 
 @defstruct[(pcdata source) ([string string?])]{
 
-Represents PCDATA content.}
+Represents textual content, i.e@._ what the
+@link["https://www.w3.org/TR/REC-xml/#dt-chardata"]{XML specification} calls
+@deftech{character data}.
+
+More specifically, this library library has several representations for
```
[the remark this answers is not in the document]
LiberalArtist: Thanks!

racket/collects/xml/private/core.rkt
LiberalArtist: Oh, I was thinking about the one that is currently `(define-runtime-module-path-index mpi-for-contracts "structures.rkt")`. I vaguely remembered that some ways of mixing styles of references could cause trouble for `raco exe` in some cases, but I've never actually encountered those problems myself, so I'm not very clear about it.

mflatt: I guess if someone referred to "core.rkt" through a file path instead of a module path, then there would be a difference for both of those relative module references, and there could be a mismatch. For that case, using relative paths as here might be better, or it might be worse, depending on the context. In any case, referencing "core.rkt" both through a file path and a module path would create other (related) problems with `raco exe`, and I don't think we need to worry about it.

mflatt:
This looks fine to me, so I'll merge soon if there are no objections.

stevebyan:
Wow! That's one heck of a commit. I've so much to learn. Thanks again.

PR #5295, opened 2025-07-08
Edit worksp/README.txt to integrate extra flag

rfindler:
Trying to put the information directly in the place where the person following the instructions will need it, but not sure if I've done this edit correctly.


mflatt:
Instead of this direction, I'm removing the need for `/addmissingpkgs`. When updating READMEs for that change, I'll keep the idea of getting rid of the "Building from a Source Distribution" section.

PR #5296, opened 2025-07-10
net,http-client: handle RST during body read

Bogdanp:
An RST packet received while reading a response with a Content-Length header would cause a thread reading from the body output port to block indefinitely as the piping thread would crash without closing the output side of the pipe. The same situation for a chunked transfer would lead to a slightly better outcome, namely that the output side of the pipe would get closed, so the reader wouldn't block indefinitely, but it would end up receiving partial data with no notice that something went wrong in the process.

This change wraps the input side of these pipes into a custom input port that propagates any errors from the piping thread to the thread reading the input port.

Unfortunately, it's not possible to trigger an RST using portable Racket code, so the added test only works on POSIX-compatible platforms and relies on the test machine having at least a C compiler.

EDIT: Looks like the test server doesn't cause an RST packet to be sent on Linux, so I'll have to look into that later.

### DIFF
--- a/pkgs/net-lib/info.rkt
+++ b/pkgs/net-lib/info.rkt
@@ -11,4 +11,4 @@
 (define license
   '(Apache-2.0 OR MIT))
 
-(define version "1.3")
+(define version "1.3.1")
--- a/pkgs/net-test/tests/net/http-client.rkt
+++ b/pkgs/net-test/tests/net/http-client.rkt
@@ -1,12 +1,18 @@
 #lang racket/base
 (module+ test
-  (require rackunit
-           racket/tcp
-           racket/port
-           (for-syntax racket/base)
-           racket/list
+  (require (for-syntax racket/base)
            (prefix-in hc: net/http-client)
-           (prefix-in u: net/url))
+           (prefix-in u: net/url)
+           racket/list
+           racket/match
+           racket/port
+           racket/runtime-path
+           racket/system
+           racket/tcp
+           rackunit)
+
+  (define-runtime-path rst-server.rkt
+    "http-client/rst-server.rkt")
 
   (define-syntax regexp-replace**
     (syntax-rules ()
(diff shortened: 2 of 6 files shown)


mflatt:
Minor: Could `rst-server-constants` be written to a "compiled" subdirectory instead of adding a ".gitignore"? I think it generally works best to put all ephemeral data in "compiled" (even data that's not ".zo"-related).

Does adding a custom-port layer create a measurable performance difference? I'm guessing that there's enough going on otherwise that it's in the noise, and I don't insist if measuring doesn't seem worthwhile. But that's something I always worry about with custom ports, and maybe you've checked already.

Bogdanp:
> Minor: Could `rst-server-constants` be written to a "compiled" subdirectory instead of adding a ".gitignore"? I think it generally works best to put all ephemeral data in "compiled" (even data that's not ".zo"-related).

Done! I used `(use-compiled-file-paths)`, but let me know if you wanted me to just use `"compiled"` literally.

> Does adding a custom-port layer create a measurable performance difference? I'm guessing that there's enough going on otherwise that it's in the noise, and I don't insist if measuring doesn't seem worthwhile. But that's something I always worry about with custom ports, and maybe you've checked already.

I hadn't measured it because I couldn't see another way to signal the error while preserving backwards-compatibility, but I went ahead and wrote [this](https://gist.githubusercontent.com/Bogdanp/962c558d515e4ef5ecb6a1be5d98f7df/raw/d02b0d26b88947317fbf326df7280be328180d50/port.rkt) after your comment. On my machine, I get the following measurements:

```
=== test/same-process #<procedure:make-pipe> 4096
cpu time: 9 real time: 9 gc time: 1
=== test/same-process #<procedure:make-pipe/err> 4096
cpu time: 14 real time: 14 gc time: 1
=== test/yes #<procedure:make-pipe> 4096
cpu time: 226 real time: 229 gc time: 1
=== test/yes #<procedure:make-pipe/err> 4096
cpu time: 358 real time: 363 gc time: 2
=== test/network #<procedure:make-pipe> 4096
cpu time: 35 real time: 212 gc time: 0
=== test/network #<procedure:make-pipe/err> 4096
cpu time: 21 real time: 359 gc time: 0
```

In raw terms, reading from the custom port is about 1.5x slower, but when the network is involved it doesn't seem to matter (in the example above, the custom port ended up being slower, but other times it's faster or the same and changing the path of the request to a 30MB file yields the same results).

I'm happy to take a different approach if there is one, though. I also considered just handling the error and closing the output side of the pipe like the chunked reader was doing, but that doesn't feel right because the reading side won't notice the error.

EDIT: Another idea would be to make the custom port opt-in and have the default behavior be to close the output side of the pipe without signalling an error.

mflatt:
Thanks for providing a performance check! A 50%-at-worst slowdown seems acceptable in a part that is unlikely to be the bottleneck.

PR #5299, opened 2025-07-12
Fix sign of lcm of negative non-integers

gus-massa:
## Checklist

- [X] Bugfix
- [X] tests included
- [X] documentation

## Description of change

Fix the sign of `lcm` in CS for non-integers, both `ratnums?` and `flonums?` 

    (lcm -9/2 15/2) ; ==> currently -45/2 should be 45/2
    (lcm -4.5 7.5) ; ==> currently -1.5 should be 1.5

I added tests for these cases for `lcm` and just in case also for `gcd`.
I think BC is correct, but let's wait until the test finish.

I also tried to fix the docs, but I don't like my fix too much, The current online version probably has a missing line and it's wrong/confusing. Perhaps I can merge only the fix of the code and someone else can fix the docs later.


gus-massa:
Commit 1) Added more tests

Commit 2) Changed the signature of `lcm` in the docs. The result may be `+inf.0` for example `(lcm 3E307 5E307)`
  and `(lcm #e3E999 1.0)`. Even if someone writes a better text, this change is necessary.

Commit 3) Still failing. No idea how to fix BC. Perhaps drop this commit completely.

mflatt:
The BC problem is opaque macro names. The commit needs `SCHEME_FLOATP` (recognizes flonums and single flonums) instead of `SCHEME_FLTP` (recognizes only single flonums).

I'll push that change and merge.

PR #5300, opened 2025-07-13
io,port: fix "who" for port-closed? contract error

Bogdanp:


mflatt:
Merged with the generated file updated as 2f92b4f8aa8d6d3b1dbaf72725be1be54b739183.

PR #5301, opened 2025-07-14
net,url: remove URL fragment on http-conn-send

### LINKED ISSUE #5221: The `get-impure-port` function should not deliver the fragment part of the URL on the wire
8.16

**What program did you run?**

On the command line, I started a TCP listener using `nc -l 9303`

Then, in DrRacket, I ran this program:

```
#lang racket

(require net/url)

(get-impure-port (string->url "http://localhost:9303/index.html#zzzz"))
```

On the terminal side, I saw this:

```
GET /index.html#zzzz HTTP/1.1
Host: localhost:9303
User-Agent: Racket/8.15.900 (net/http-client)
Content-Length: 0
```

... indicating that the fragment part, `#zzzz`, was appearing as part of the request. 

@LiberalArtist reports:

"To my surprise, [RFC 3986 § 3.5](https://datatracker.ietf.org/doc/html/rfc3986#section-3.5) indeed says:

    Fragment identifiers have a special role in information retrieval systems as the primary form of client-side indirect referencing, allowing an author to specifically identify aspects of an existing resource that are only indirectly provided by the resource owner. As such, the fragment identifier is not used in the scheme-specific processing of a URI; instead, the fragment identifier is separated from the rest of the URI prior to a dereference, and thus the identifying information within the fragment itself is dereferenced solely by the user agent, regardless of the URI scheme. Although this separate handling is often perceived to be a loss of information, particularly for accurate redirection of references as resources move over time, it also serves to prevent information providers from denying reference authors the right to refer to information within a resource selectively. Indirect referencing also provides additional flexibility and extensibility to systems that use URIs, as new media types are easier to define and deploy than new schemes of identification."

The `GET` line (I believe) should instead have been

```
GET /index.html HTTP/1.1
```

taimoorzaeem:
## Checklist

- [x] Bugfix
- [ ] Tests included
- [ ] Documentation

There were **no** regressions with this change, so I am not sure how exactly should I test this. Not sure if the documentation needs to be updated. Any pointers or directions for these would be a great help in closing this issue.

## Description of change

The `get-impure-port` function sends the URL fragment in the http connection request. The fragment part should not be sent to the server and should be handled at client side.

This commit removes the fragment part from the `get-impure-port` function call.

Closes #5221.

### DIFF
--- a/racket/collects/net/url.rkt
+++ b/racket/collects/net/url.rkt
@@ -214,7 +214,9 @@
                         (if (null? (url-path url))
                             (values #t (list (make-path/param "" '())))
                             (values (url-path-absolute? url) (url-path url)))])
-            (make-url #f #f #f #f abs? path (url-query url) (url-fragment url)))))))
+            ;; We don't send the URL fragment to the server, hence
+            ;; the last #f
+            (make-url #f #f #f #f abs? path (url-query url) #f))))))
 
   (hc:http-conn-send! hc access-string
                       #:method (if get? #"GET" #"POST")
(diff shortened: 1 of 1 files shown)


benknoble:
It looks like this got swallowed in the noise; I see @jbclements and @jackfirth on the original issue, and this sounds like something @mflatt or @Bogdanp might be interested in?

PR #5304, opened 2025-07-17
Back to the Futures Part II: parallel threads

mflatt:
This is a first cut at adding *parallel threads* to Racket. Racket's original kind of threads are called *coroutine threads* in updated documentation.

The `thread` function changes to

```
 (thread thunk
        [#:pool pool
         #:keep-results? keep-results?]) -> thread?
   pool: (or/c #f 'own parallel-thread-pool?) = #f
   keep-results?: any/c = #f
```

The default `#f` for `#:pool` creates a coroutine thread, like before. Supplying`'own` for `#:pool` creates a parallel thread that uses its own OS thread. Supplying a thread pool creates a parallel thread that is backed with the pool's OS threads (i.e., OS threads that are separate from the coroutine OS thread, but shared by threads in the pool—still parallel to each other, too). [edit:] If `keep-results?` is true, then `thread-wait` will report the result of `thunk`; that's an extension for coroutine threads as well as parallel threads.

The implementation is based on the existing *future*s layer, and so performance improvements can be expected when large regions a thread's work would run without blocking in a future. The difference is that when a blocking operation is encountered in a parallel thread, it is kicked over to a coroutine shadow thread; once the blocking operation completes[*], the continuation can be moved afterward back to the parallel thread's future.

[*] The notion of a blocking operation "completing" falls out naturally in the way most blocking operations are implemented. Consistent with the existing terminology for futures, we could say that all blocking operations for futures turn into synchronizing operations for parallel threads.

From the perspective of a library that aims to be thread-safe, this change doesn't introduce much new machine-level concurrency compared to what was possible before via futures. It does make machine-level concurrency more accessible, so we'll likely find places that weren't synchronized correctly before. Also, it's only mostly the same, because a library could conceivably depend on a blocking operation to ensure sequential consistency afterward—but that seems unlikely compared to the probably of plain old lock bugs.

Note that if a Racket program uses semaphores, then it's only the semaphore wait or post that is atomic, and the part in between can run in parallel. That's almost certainly what you want, but it illustrates how, from a library perspective, parallel threads create more possible machine-level concurrency (as parallelism) than just futures.

Blocking regions (that become merely synchronizing) are primarily identified by `start-atomic` and `end-atomic` pairs. Those atomicity commands are used within the Racket implementation, and they are also exposed via `ffi/unsafe/atomic`. That generally the key to this adaptation. When a parallel thread is running, between calls to those two functions, the continuation is moved to a coroutine thread with a (fairly expensive) synchronization. As far as Racket code is concerned, though, it all runs inside `(current-thread)`. The difficult part of the implementation is making the coroutine thread scheduler and future thread scheduler play together, and making things like `kill-thread`, `break-thread`, and `thread-suspend` work like they should.

All filesystem and network I/O remains blocking/synchronizing, so moving a program with a lot of I/O into a parallel thread makes it much slower. Improving that is probably the next step.


mflatt:
Some things I forgot to mention:

* To try a full Racket build, you'll need a patch like this for TR: https://github.com/mflatt/typed-racket/commit/473404512356120f5ec7dcb0c271b009ddf5a8c1
* BC isn't going to have parallel threads, obviously. It will ignore `#:pool`, except to complain about a closed pool. I haven't yet updated it fully.
* The futures visualizer should be useful for understanding parallel-thread performance. I'm not sure it is yet, but it should be a matter of getting logging right on the Racket side.

mflatt:
I'm not so happy with `#:keep-results?` and a boolean, so I plan to try out `#:keep 'results` and ~~`#:keep 'nothing`~~ `#:keep #f` instead.

racket/src/thread/README.txt
```diff
@@ -7,6 +7,14 @@ more primitive layer must also provide `break-enabled-key` and special
 handling for looking up a mark with that key so that an egine-specific
 default thread cell is produced.
 
+Thread
+------
+
+"Thread" or "Racket thread" here means a Racket thread in the sense of
+`thread`. Thse are "green threads" within a place; they do not run
+concurrent with each other. Synchronization at this level is often
+implemented by using atomic mode via `start=atomic` and `end-atomic`.
```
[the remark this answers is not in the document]
eutro: Another typo with "Thse"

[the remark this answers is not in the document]
mflatt: Thanks — I've fixed these, but not doubt introduced more while adding to this file.

racket/src/thread/atomic.rkt
```diff
@@ -7,27 +7,28 @@
          "debug.rkt"
          (for-syntax racket/base))
 
-(provide atomically
-         current-atomic
+(provide current-atomic ; tecnicall
```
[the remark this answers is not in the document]
mflatt: I decided not to put a comment there.

pkgs/racket-doc/scribblings/reference/eval-model.scrbl
```diff
@@ -982,20 +970,43 @@ for the cell in the created thread. For a non-preserved thread cell, a
...
+
+The possibility of shared state imposes a cost on some operations,
+particularly in the case of sharing among parallel threads, and
+using parallel threads can easily make a computation slower than
+using coroutine threads when the underlying primitives resort to a
+more pessimistic mode.
+@tech{Futures} and @tech{places} are alternatives to parallel threads
+that provide different trade-offs in sharing constraints and performance.
+@tech{Futures} sometimes achieve better performance by limiting
+operations that run in parallel; as a result, they can created and
```
sorawee, line 998: Typo: _be_ created and completed

pkgs/racket-test/tests/racket/stress/parallel-custodian.rkt
```diff
@@ -0,0 +1,29 @@
...
+
+(for ([i (in-range 1000)])
+  (define ts
+    (for/list ([j (in-range 8)])
+      (parameterize ([current-custodian (make-custodian)])
+        (thread
+         (lambda ()
+           (custodian-shutdown-all (current-custodian)))
+         #:pool 'own))))
+  (map thread-wait ts))
```
sorawee, line 19: Curious: why is this a `map` rather than `for-each`?

pkgs/racket-doc/scribblings/reference/threads.scrbl
```diff
@@ -195,10 +212,19 @@ otherwise.}
...
+         any]{
 
 Blocks execution of the current thread until @racket[thd] has
-terminated. Note that @racket[(thread-wait (current-thread))]
+terminated. If the thread's procedure raised an exception or otherwise
+aborted to the thread's initial @tech{prompt}, @racket[fail-k] is
+called to produce the result of @racket[thread-wait]. Otherwise, if
+the thread records its results (see @racket[#:keep] in
+@racket[thread]), those results are returned, while @|void-const| is
+return if the thread does not keep its results.
```
sorawee, line 225: This program prints `#<void>`. Does this agree with the above description?

```
#lang racket

(define t
  (thread
   #:keep 'results
   (λ ()
     (sleep 5)
     1)))

(break-thread t)
(println (thread-wait t (λ () 123)))
```

mflatt: That example was misbehaving due to s bug, now fixed.

pkgs/racket-test-core/tests/racket/parallel.rktl
```diff
@@ -273,6 +273,15 @@
   (test 'ok thread-wait (thread (lambda () 'ok) #:pool 'own #:keep 'results))
   (test-values '(ok more) (lambda () (thread-wait (thread (lambda () (values 'ok 'more)) #:pool 'own #:keep 'results)))))
 
+(let ()
+  (define t (parameterize ([current-error-port (open-output-bytes)])
+              (thread #:keep 'results
+                      (λ ()
+                        (sleep (system-idle-evt))
```
sorawee, line 280: Do you mean `sync` here rather than `sleep`? Also, I'm curious why `system-idle-evt` is chosen over `never-evt`.

mflatt: Oops, yes. The thread wasn't getting there before getting broken, which is the intent of the test, but not guaranteed. Using `(sync (never-evt))` would be fine — could be optimized very differently than intended internally, although it won't be.

pkgs/racket-test/tests/racket/stress/parallel-custodian.rkt
mflatt: A `for-each` would certainly make sense, but my fingers just happen to type `map` more readily.

mflatt:
Support for parallel threads affects performance programs that do not use parallel threads. The main effects, currently: `hash-ref` on a mutable `eq?`- or `eqv?`-based table takes 2 to 3 times as long, reading and writing small numbers of bytes through the fastest port is up to 1.75x times as long (on x86_64), and getting a random number is slower by 15% or so. Those operations are too common to force a parallel thread to synchronize with the scheduler thread, so programs running in the scheduler thread now need to take a lock. The `hash-ref` optionation, didn't need to do that before, port operations have relied on atomic mode instead of a port-specific lock, and the random-number-generator state was previous broken with respect to futures. A specialized lock implementation is in each case to get as close as it does.

These are fairly fast operations—on the order of a microsecond—and a realistic program is going to mix these with much slower operations. So, I think the tradeoff will be ok, but it's possible that there's room for improvement.

For concreteness, here are some `hash-ref`, `hash-set!`, and port numbers, with numbers for immutable hash tables provide for comparison and as a sanity check (and it turns out that the new lock is faster for `hash-set!` on Apple Silicon):

```
Apple M2
--------

msecs for 100M operations                  old       new

mutable hasheq size 1, hash-ref            439       733  <-
immutable hasheq size 1, hash-ref          763       765
mutable hasheq size 129, hash-ref          469       732  <-
immutable hasheq size 129, hash-ref       1589      1598

mutable hasheq size 129, hash-set!        1159       910  <-
immutable hasheq size 129, hash-set!      1211      1211
mutable hasheq size 129, hash-set!        1200       899  <-
immutable hasheq size 129, hash-set       6260      6264

msecs for 10k operations, 1k times         old       new
write/read-byte on bytes port              124       145  <-
write-/read-bytes! #"abc" on bytes port    358       380  <-



x86_64 i7-7700 (Kaby Lake)
--------------------------

msecs for 100M operations                  old       new

mutable hasheq size 1, hash-ref            648      1831  <-
immutable hasheq size 1, hash-ref         1257      1251
mutable hasheq size 129, hash-ref          640      1822  <-
immutable hasheq size 129, hash-ref       2585      2705

mutable hasheq size 129, hash-set!        1776      2003  <-
immutable hasheq size 129, hash-set!      2233      1974
mutable hasheq size 129, hash-set!        1852      2019  <-
immutable hasheq size 129, hash-set       8123      7694

msecs for 10k operations, 1k times         old       new
write/read-byte on bytes port              217       346  <-
write-/read-bytes! #"abc" on bytes port    581       686  <-
```

[ht.txt](https://github.com/user-attachments/files/21371452/ht.txt)
[io.txt](https://github.com/user-attachments/files/21371453/io.txt)

mflatt:
For an I/O-intensive program, probably the biggest remaining blocker is anything that interactions with custodians, which includes opening and closing files. I have ideas about how to improve that, but not for this PR.

I'll test more and plan to merge tomorrow. After that, I can confidently promise that further testing with the `master` branch and snapshots will turn up problems.

PR #5306, opened 2025-07-23
docs(zuo): unify `target`'s 2nd parameter name to `get-rule`

shenlebantongying:
- [x] documentation

The doc of [`target`](https://docs.racket-lang.org/zuo/zuo-build.html#%28def._%28%28lib._zuo-doc%2Ffake-zuo..rkt%29._target%29%29) says the 2nd parameter return a `rule`.

However, this parameter has 2 names `get-rule` and `get-deps` across the doc. I find this is confusing.

So when talking about `target`, which is all occurrences of `get-deps`, let's consistently use `get-rule`.

close https://github.com/racket/zuo/issues/28


mflatt:
Thanks for this improvement! I merged it first to racket/zuo, and then here as d2a972cfcf262a9b957be9f47524285b83c410ff.

PR #5307, opened 2025-07-25
vectors.scrbl: warn users that (make-vector n (vector)) creates alias to the same vector

jestarray:
## Checklist
- [x] documentation

## Description of change
warn users that (make-vector n (vector)) creates alias to the same vector and provide a solution. I and someone else on discord have been hit by this and wasted an hour or so finding what the issue was

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/vectors.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/vectors.scrbl
@@ -50,14 +50,14 @@ all elements, so for mutable data, mutating an element will affect other element
 This function takes time proportional to @racket[size].
 
 A common mistake is using @racket[make-vector] to create nested vectors.
-The following creates a vector that contain 3 vectors to the same vector.
+The following creates a vector that contain 3 references to the same vector.
 @examples[
-  (make-vector 3 (vector)) ; wrong
+  (make-vector 3 (vector)) ; probably not what you'd want
 ]
 
 To create nested vectors, do this instead:
 @examples[
-(for/vector ([_ (in-range 0 n)])
+(for/vector ([_ 3)])
   (vector))
 ]
 }
--- a/pkgs/racket-doc/scribblings/reference/vectors.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/vectors.scrbl
@@ -50,15 +50,14 @@ all elements, so for mutable data, mutating an element will affect other element
 This function takes time proportional to @racket[size].
 
 A common mistake is using @racket[make-vector] to create nested vectors.
-The following creates a vector that contain 3 references to the same vector.
-@examples[
-  (make-vector 3 (vector)) ; probably not what you'd want
-]
+The fact that @racket[v] is shared for all elements means that @racket[(make-vector 3 (make-vector 4))]
+would not be a good way of representing a mutable 3x4 matrix, for example, since just one mutable vector
+would be shared three times. Using `for/vector` as follows more likely produces the intended result, since
+it evaluates `(make-vector 4)` separately for each iteration:
 
-To create nested vectors, do this instead:
 @examples[
-(for/vector ([_ 3)])
-  (vector))
+(for/vector ([i (in-range 3))])
+  (make-vector 4))
 ]
 }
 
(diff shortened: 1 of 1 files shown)


pkgs/racket-doc/scribblings/reference/vectors.scrbl
```diff
@@ -47,7 +47,20 @@ all elements, so for mutable data, mutating an element will affect other element
   v
 ]
 
-This function takes time proportional to @racket[size].}
+This function takes time proportional to @racket[size].
+
+A common mistake is using @racket[make-vector] to create nested vectors.
+The following creates a vector that contain 3 vectors to the same vector.
+@examples[
+  (make-vector 3 (vector)) ; wrong
```
jbclements, line 55: Strictly speaking, this is not "wrong", and you might sometimes want it... it's just that you often don't. Would you consider changing "wrong" to "probably not what you want" ?

pkgs/racket-doc/scribblings/reference/vectors.scrbl
```diff
@@ -47,7 +47,20 @@ all elements, so for mutable data, mutating an element will affect other element
...
+]
+
+To create nested vectors, do this instead:
+@examples[
+(for/vector ([_ (in-range 0 n)])
+  (vector))
```
jbclements, line 61: Two things: 

* You used "3" in the first example, I would use it in the second one as well.
* I think it's probably slightly more idiomatic to omit the in-range, and just write
```
(for/vector ([_ 3])
  (vector))
```

... but I could be persuaded that I'm wrong.

pkgs/racket-doc/scribblings/reference/vectors.scrbl
```diff
@@ -47,7 +47,20 @@ all elements, so for mutable data, mutating an element will affect other element
   v
 ]
 
-This function takes time proportional to @racket[size].}
+This function takes time proportional to @racket[size].
+
+A common mistake is using @racket[make-vector] to create nested vectors.
+The following creates a vector that contain 3 vectors to the same vector.
```
jbclements, line 53: I think you meant "3 references to the same vector", not "3 vectors to the same vector".

jbclements:
I should also have said: thanks for working on this!

pkgs/racket-doc/scribblings/reference/vectors.scrbl
jestarray: Ah, I didn't know that, I just copied what jack said on discord. It's better not to use `in-range` since it's shorter

jestarray:
np! first time contributing to this repo so bear with me :sweat_smile:

soegaard:
Since the length is known, it is worth using `#:length`.

```
(for/vector #:length n ([_ (in-range 0 n)])
  (vector)
```

pkgs/racket-doc/scribblings/reference/vectors.scrbl
jackfirth: I actually prefer the explicit `in-range`, personally. I think the way `for` loops treat bare numbers is unintuitive. But Resyntax has no official stance on the issue.

jestarray:
> Since the length is known, it is worth using `#:length`.
> 
> ```
> (for/vector #:length n ([_ (in-range 0 n)])
>   (vector))
> ```

is the `in-range` needed? It seems like it works without it? Seems a bit annoying to need to keep the two `n`'s in sync

mflatt:
How about rephrasing to something like this, which calls back more directly to the first paragraph and positions it as an example of a more general problem (and also avoids the empty vector, which would be perfectly fine to share):

```
The fact that @racket[v] is shared for all elements means that @racket[(make-vector 3 (make-vector 4))]
would not be a good way of representing a mutable 3x4 matrix, for example, since just one mutable vector
would be shared three times. Using `for/vector` as follows more likely produces the intended result, since
it evaluates `(make-vector 4)` separately for each iteration:

@examples[
(for/vector ([i (in-range 3))])
  (make-vector 4))
]
```
I've opted here to use `i` to avoid shadowing `_`, and I also used `in-range` but not `#:length`.

jestarray:
alright, copied and pasted the revised explanation in

mflatt:
Thanks!

PR #5310, opened 2025-07-26
test,http-client: adjust test to avoid failing on BC

Bogdanp:
I noticed that, after #5296 was merged, the [BC tests](https://github.com/racket/racket/actions/runs/16532995890/job/46762657013#step:8:1680) started failing. I think it's not actually necessary to interrupt the process, since it shuts down on its own after serving the request, so this change should fix the failure.

### DIFF
--- a/pkgs/net-test/tests/net/http-client.rkt
+++ b/pkgs/net-test/tests/net/http-client.rkt
@@ -462,37 +462,29 @@
   ;; Test that a RST from a client during body transfer doesn't deadlock.
   (when (memq (system-type 'os) '(unix macosx))
     (for ([chunked? (in-list '(#f #t))])
-      (match-define (list stdout stdin _pid stderr control)
+      (match-define (list stdout stdin _pid stderr _control)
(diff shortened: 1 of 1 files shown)


mflatt:
I addressed the failure by making adjustments in rktio, working around bad behavior by Mac OS (although it took me a couple of tries: 3bbd436d2351415c0b90144ce95f331f0fee955e, 43a94c3e4201a97dc6dd6ba4288e5178c63b8941). So, even though interrupting is not necessary, I'd prefer to keep it as a regression check.

Bogdanp:
> I addressed the failure by making adjustments in rktio, working around bad behavior by Mac OS (although it took me a couple of tries: [3bbd436](https://github.com/racket/racket/commit/3bbd436d2351415c0b90144ce95f331f0fee955e), [43a94c3](https://github.com/racket/racket/commit/43a94c3e4201a97dc6dd6ba4288e5178c63b8941)). So, even though interrupting is not necessary, I'd prefer to keep it as a regression check.

Sounds good to me! Thanks!

PR #5312, opened 2025-07-30
Fix undeclared __clear_cache.

rmolives:
## Checklist

- [x] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation

## Description of change

When I was compiling Rackert for Riscv and Loongarch, I encountered an issue

```
[  361s] ChezScheme/c/clearcache.c:48:3: error: call to undeclared function '__clear_cache';  ISO C99 and later do not support implicit function declarations [-Wimplicit-function-declaration]
[  361s]    48 |   __clear_cache((char *)start, (char *)end);
[  361s]       |   ^
[  361s] 1 error generated.
This issue is due to the fact that implicit function declarations are not allowed after ISO C99, so I fixed it.
```

### DIFF
--- a/racket/src/ChezScheme/c/clearcache.c
+++ b/racket/src/ChezScheme/c/clearcache.c
@@ -45,6 +45,7 @@ void S_doflush(uptr start, uptr end) {
 #ifdef S_TARGET_OS_IPHONE
   sys_icache_invalidate((void *)start, (char *)end-(char *)start);
 #else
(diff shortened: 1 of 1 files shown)


gus-massa:
Thanks for the report. You posted this also in https://github.com/cisco/ChezScheme/pull/962 and @mflatt already replied to it. 99% sure this will be fixed as usual first in Chez Scheme and ported here afterwards.

PR #5314, opened 2025-08-07
Update strncpy in regexp.c (bc)

bdeket:
#### Checklist
- [X] Bugfix

#### Description of change
In the last few months something changed on my machine that broke my build script for snapshot racket-bc on linux.
I think it has to do with how strict gcc is but anyway...
If `strncpy` is not defined a conflicting definition to what is used is given, this PR changes the signature to bring it in line.
But maybe this can be deleted all together since i don't see any use of `strncpy` in `regexp.c`.
(with this PR local build works again)

### DIFF
--- a/racket/src/bc/src/regexp.c
+++ b/racket/src/bc/src/regexp.c
@@ -4252,7 +4252,7 @@ regstrcspn(char *s1, char *e1, char *s2)
 }
 
 #ifndef strncpy
-  extern char *strncpy();
+  extern char *strncpy(char *, const char *, long unsigned int);
 #endif
 
 /*


mflatt:
That must be an odd leftover from way back. I agree that it can be just deleted, so I'll push that change.

benknoble:
Indeed, `git log --follow -S strncpy -- racket/src/bc/src/regexp.c` finds e41b2fb359 (Added src, set properties for all DOS files., 2005-05-27) immediately. That looks like an initial import of some code. If that code is at least as old as 1987, perhaps it was leftover from supporting an old C compiler.

mflatt:
Removed in 0c09999db9c2521411b6d855926d977c3769c9ed.

samth:
@benknoble If you look at https://github.com/racket/old-plt/blob/3a0f0320d6b793ce1e5bf05aff04f77f67a0141c/src/mzscheme/src/regexp.c you can see that it was there in the very first version of regexp.c that Matthew added. If you look here https://github.com/garyhouston/regexp.old/blob/master/regexp.c at what I think was the source Matthew copied the `extern` is not there.

benknoble:
Thanks @samth! I need to get my clone set up with a graft to that old repository for history spelunking.

PR #5332, opened 2025-09-12
Fix #3167

### LINKED ISSUE #3167: with-intercepted-logging intercepts too much with #:logger
The documentation for `with-intercepted-logging` states:

> If #:logger is specified, intercepts events sent to that logger, otherwise uses a new child logger of the current logger.

But it seems that it captures not only events sent to that logger, but anything sent to it and `(current-logger)`.

```rkt
(define-logger bar)
(with-intercepted-logging
  displayln
  (λ () (log-warning "hello"))
  #:logger bar-logger
  'warning)
```

I thought this wouldn't print anything (since we never invoke `log-bar-warning`), but it does. This is because `with-intercepted-logging` [sets the `current-logger`](https://github.com/racket/racket/blob/master/racket/collects/racket/logging.rkt#L55) to the passed logger. This seems, at least to me, to be the wrong behavior.

camoy:
## Checklist

- [x] Bugfix
- [x] tests included

## Description of change

This PR fixes the behavior described in #3167 to match both intuitive expectations and the documentation. I did not end up going with the solution that I originally proposed in the issue because it turns out that the strategy the code currently uses is fundamentally broken. If `proc` captures a continuation and restores it later, it can cause `channel-put` to execute after the thread has terminated, resulting in hanging. The solution I ended up going with, which resolves both this issue and my original one, is to use `dynamic-wind` to launch and teardown the receiver thread when execution enters and exits the dynamic extent of `proc`. 

I have two tests. The first confirms that #3167 is fixed. The second confirms that logs are intercepted when a continuation is invoked that re-enters `proc`.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/logger.rktl
+++ b/pkgs/racket-test-core/tests/racket/logger.rktl
@@ -280,6 +280,35 @@
     warning-counter))
 (test 2 test-intercepted-logging2)
 
+;; From issue #3167
+(define (test-intercepted-logging3)
+  (define-logger bar)
+  (define ok? #t)
+  (with-intercepted-logging
+    (λ (l) (set! ok? #f))
+    (λ () (log-warning "hello"))
+    #:logger bar-logger
+    'warning)
+  ok?)
+(test #t test-intercepted-logging3)
+
+(define (test-intercepted-logging4)
+  (define msg #f)
+  (define kont #f)
+  (call-with-continuation-prompt
+   (λ ()
+     (with-intercepted-logging
+       (λ (l) (set! msg l))
+       (λ ()
+         (when (call-with-composable-continuation
+                (λ (k)
+                  (set! kont k) #f))
+           (log-warning "hello")))
+       'warning)))
+  (kont #t)
+  (and msg #t))
+(test #t test-intercepted-logging4)
+
 ; --------------------
 ;; Check that a blocked log receiver is not GCed if
 ;; if might receiver something
--- a/racket/collects/racket/logging.rkt
+++ b/racket/collects/racket/logging.rkt
@@ -40,22 +40,21 @@
                 (intercept l)
                 (loop)]))))))
 
-(define (with-intercepted-logging interceptor proc #:logger [logger #f]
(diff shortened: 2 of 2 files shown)


mflatt:
What effect does this change have on existing packages that use `with-intercepted-logging` with `#:logger`?

Some packages that  seem to use it (not a complete list): `video`, `pollen`, `resyntax` (in tests), `make-log-interceptor`, and `trivial`.

camoy:
That's a good question, I just took a quick look at these packages.

## `video`

Used in two places. The [first](https://github.com/videolang/video/blob/3c69669063c56ff8d269768589cb9506a33315e5/video/tests/ffmpeg.rkt#L42-L50) doesn't use the logs at all and seems to just be a workaround. The [second](https://github.com/videolang/video/blob/3c69669063c56ff8d269768589cb9506a33315e5/video/private/devices.rkt#L89-L116) is interesting in that it seems to me that without this PR the code would actually be buggy. There is an assumption that the log message matches one of three regular expressions, but that is only true if the log interceptor only intercepts logs from `ffmpeg-logger`, which it may not in the original code. After this PR, I think the code is correct.

## `pollen`

Looks like `pollen` just vendors the old version of `with-intercepted-logging` for compatibility with old versions of Racket (and consequently doesn't use it).

## `make-log-interceptor`

Wraps `with-intercepted-logging` with a slightly different interface and functionality. Should not be negatively affected.

## `trivial`

The use [here](https://github.com/bennn/trivial/blob/6da6a2edea1912c7b28d0bcaca5ffe09bf247ca5/trivial/private/test-common.rkt#L61-L74) is only in tests and manually filters out logs not related to `ttt`. So the check on line 67 is redundant with this PR.

rfindler:
One (very) concrete thing to check is if the tests continue to pass with the change. If they don't, we should take some care because it can be difficult to get some maintainers to accept pull requests.

camoy:
I checked and tests pass on these projects with the patch, other than `video`. I wasn't able to get `video` to work at all because it requires an older `ffmpeg` than I have on my machine.

mfelleisen:
Ask Leif. But I think this is the abandoned version. (Perhaps Leif can confirm and then remove the package.)

PR #5333, opened 2025-09-13
Add a simple check for method defns to struct

### LINKED ISSUE #1711: struct exposes generic-method-table in error message for bad #:methods
This highly incorrect structure declaration
```
(struct buggy ()
  #:methods gen:dict "not a methods section")
```
gives the error message `generic-method-table: bad syntax in: (generic-method-table gen:dict . "not a methods section")`, rather than reporting the error in terms of `struct`.

samdphillips:
## Checklist
- [x] Bugfix
- [x] tests included

## Description of change
Adds an outer level check to `struct` for `#:methods` sub-forms to ensure the methods definitions are a nested list form.

resolves #1711

### DIFF
--- a/pkgs/racket-test/tests/generic/errors.rkt
+++ b/pkgs/racket-test/tests/generic/errors.rkt
@@ -136,4 +136,13 @@
                            (define-generics AA [meth AA . a])
                            (meth 1 2))))
              "rest args present")
+
+  ;; https://github.com/racket/racket/issues/1711
+
+  (check-exn #rx"struct: .* not a parenthesized sequence of method definitions"
+             (lambda () (convert-compile-time-error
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks for this improvement! But for the error message, I would normally reserve the word "list" for a value. For a syntactic form, I would use "sequence" or "parenthesized sequence". (Granted, "sequence" also has a run-time meaning.)

samdphillips:
> Thanks for this improvement! But for the error message, I would normally reserve the word "list" for a value. For a syntactic form, I would use "sequence" or "parenthesized sequence". (Granted, "sequence" also has a run-time meaning.)

Thanks.  "List" didn't sound correct to me either.

PR #5334, opened 2025-09-14
add derived `prop:sequence` to `gen:set`

### LINKED ISSUE #2526: Generic gen:set sets should satisfy sequence?
The [sequence documentation](https://docs.racket-lang.org/reference/sequences.html) seems to say that all sets are sequences, but instances of custom set types implemented using `gen:set` currently do not satisfy `sequence?` and cannot be used directly as sequences. This is surprising, especially as they do work with e.g. `in-set`.

Here's a small example:
```racket
#lang racket

(require racket/generic
         rackunit)

(struct my-set (set)
  #:transparent
  #:methods gen:set
  [(define/generic super:set-member? set-member?)
  (define (set-member? this v)
    (super:set-member? (my-set-set this) v))
  (define/generic super:set-add set-add)
  (define (set-add this v)
    (super:set-add (my-set-set this) v))
  (define/generic super:set-remove set-remove)
  (define (set-remove this v)
    (super:set-remove (my-set-set this) v))
  (define/generic super:set-first set-first)
  (define (set-first this)
    (super:set-first (my-set-set this)))
  (define/generic super:set-empty? set-empty?)
  (define (set-empty? this)
    (super:set-empty? (my-set-set this)))
  (define/generic super:set-copy-clear set-copy-clear)
  (define (set-copy-clear this)
    (super:set-copy-clear (my-set-set this)))])

(check-pred null? (set->list (my-set (set))))

(check-pred sequence? (in-set (my-set (set))))

(check-pred sequence? (my-set (set))
            "this one fails")
```

samdphillips:
## Checklist
- [x] Bugfix
- [x] tests included

## Description of change
resolves #2526

### DIFF
--- a/pkgs/racket-test-extra/tests/racket/gh_2526.rkt
+++ b/pkgs/racket-test-extra/tests/racket/gh_2526.rkt
@@ -1,0 +1,31 @@
+#lang racket/base
+
+(require racket/generic
+         racket/set
+         rackunit)
+
+(struct my-set (set)
+  #:transparent
+  #:methods gen:set
+  [(define/generic super:set-member? set-member?)
+  (define (set-member? this v)
+    (super:set-member? (my-set-set this) v))
+  (define/generic super:set-add set-add)
+  (define (set-add this v)
+    (super:set-add (my-set-set this) v))
+  (define/generic super:set-remove set-remove)
+  (define (set-remove this v)
+    (super:set-remove (my-set-set this) v))
+  (define/generic super:set-first set-first)
+  (define (set-first this)
+    (super:set-first (my-set-set this)))
+  (define/generic super:set-empty? set-empty?)
+  (define (set-empty? this)
+    (super:set-empty? (my-set-set this)))
+  (define/generic super:set-copy-clear set-copy-clear)
+  (define (set-copy-clear this)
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks!

mflatt:
I think we're going to have to revert this one. It causes problems with packages like "ricoeur-tei-utils" and "relation-lib", which add their own `gen:sequence` or `prop:sequence`.

samdphillips:
That's unfortunate.  I'll reach out to the maintainers so we can see about including this in the next release.

mflatt:
While fixing all packages is enough for our release rules, the worry that this may be a more widespread problem. Multiple packages having trouble with the change may be a symptom of breaking a lot of code, which we generally don't want to do. I'm not sure, but it might be better in this case to document that `gen:set` should be paired with `gen:sequence`.

samdphillips:
Yes, that sounds like it may be the best path.  I found couple more similar cases searching github just now.

LiberalArtist:
As the author of both https://github.com/racket/racket/issues/2526 and one or the offending packages, I've been trying to think if there's some way `gen:set` could add only a default implementation of the sequence interface, so an explicit `prop:sequence` could still be allowed to override it.

PR #5339, opened 2025-09-21
Adds missing require statement in unit.scrbl example

acarrico:
## Checklist
- [x ] documentation

## Description of change
Adds missing require statement in unit.scrbl example.


mflatt:
Although this part of the guide is following a convention where the REPL is meant to be cumulative over subsections — so the `require` in the previous section is meant to be in effect for this section — it looks like this `require` is the only thing really carried over beyond the module-defining files. So, an arguably redundant `require` seems ok and helpfully clearer.

Thanks!

acarrico:
I see what you mean. My impression had been that 14.3 would be inferring the same definitions as 14.2, so it was meant to be fresh, but I guess you could just clobber them.

Thank you!

PR #5341, opened 2025-09-23
Fix `sequence-ref` to avoid unnecessary lookahead consumption

NoahStoryM:
## Checklist

- [x] Bugfix
- [ ] Feature
- [x] tests included
- [ ] documentation

## Description of change

Fix `sequence-ref` to stop consuming extra elements from stateful sequences.

### Problem

The current implementation of `sequence-ref` uses `#:unless (j . < . i)` which causes the loop to continue after reaching the target index, leading to unnecessary consumption of an additional element from the sequence. See also #5338.

### Example

```racket
> (define v 0)
> (define s
    (make-do-sequence
     (λ ()
       (define (pos->element _) v)
       (define (continue-with-pos? _) (set! v (add1 v)) (< v 100))
       (values pos->element void (void) continue-with-pos? #f #f))))
> (define-values (more? get) (sequence-generate s))
> (displayln (get))
1
> (displayln (sequence-ref s 0))
2
> (displayln (sequence-ref s 0))
4
> (displayln (sequence-ref s 1))
7
> (displayln (sequence-ref s 1))
10
> (displayln (get))
12
> (displayln (sequence-length s))
87
```

### DIFF
--- a/pkgs/racket-test-core/tests/racket/sequence.rktl
+++ b/pkgs/racket-test-core/tests/racket/sequence.rktl
@@ -174,8 +174,8 @@
   (define s
     (make-do-sequence
      (λ ()
-       (define (pos->element _) v)
-       (define (continue-with-pos? _) (set! v (add1 v)) (< v 100))
+       (define (pos->element _) (set! v (add1 v)) v)
+       (define (continue-with-pos? _) (< v 100))
        (values pos->element void (void) continue-with-pos? #f #f))))
   (define-values (more? get) (sequence-generate s))
   (test 1 'sequence-ref (get))
--- a/pkgs/racket-test-core/tests/racket/sequence.rktl
+++ b/pkgs/racket-test-core/tests/racket/sequence.rktl
@@ -184,7 +184,7 @@
   (test 5 'sequence-ref (sequence-ref s 1))
   (test 7 'sequence-ref (sequence-ref s 1))
   (test 8 'sequence-ref (get))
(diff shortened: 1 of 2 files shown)


sorawee:
I could be missing something, but does this accomplish what you want? 

```
(define (ref s i)
  (let ([v (for/first ([v (in-values*-sequence s)]
                       [j (in-range (add1 i))]
                       #:unless (j . < . i))
             (or v '(#f)))])
    (cond
      [(not v)
       (raise-arguments-error
        'sequence-ref
        "sequence ended before index"
        "index" i
        "sequence" s)]
      [(list? v) (apply values v)]
      [else v])))
```

NoahStoryM:
I think this version looks correct, and it should behave as intended.
I’m just not sure whether this approach is better compared to the sequence-generate version.

Since this one matches the behavior of the original implementation, I’ll go ahead and use it for now.

sorawee:
An issue with the `(for ([_ (in-range i)] ...)` + `sequence-generate` revisions above is that it's O(i) rather than O(min(i, |s|)). Maybe there's a way to make them work, but I haven't tried.

sorawee:
Is this good to merge for the upcoming release? @mflatt

jbclements:
@mflatt bump.... ready to merge this one? I've stared at it for about 10 minutes, and if I stare at it for another 20 or so, I'll probably understand it pretty well.

PR #5342, opened 2025-09-23
add `black-box`

mflatt:
The `black-box` function is useful for testing and benchmarking. The name is based on Rust's https://doc.rust-lang.org/beta/std/hint/fn.black_box.html.

Since it's only for tasks like benchmarking, putting `black-box` in a library instead of `racket/base` would make sense. But the convenience of having `black-box` in `racket/base` (so always available) is especially appealing to me, so I can write quick examples and tests without thinking too hard about it, so that's what this PR does. I have checked that the name `black-box` isn't defined by any package at pkgs.racket-lang.org, although it is used as a struct field name in `racket-chipmunk`.

Examples from the docs:

```
> (let ([to-power 100])
    (let loop ([i 1000])
      (unless (zero? i)
        ; call to `expt` is optimized away entirely, since
        ; there's no effect and the result is unused:
        (expt 2 to-power)
        (loop (sub1 i)))))
> (let ([to-power 100])
    (let loop ([i 1000])
      (unless (zero? i)
        ; call to `expt` is optimized to just returning a folded
        ; constant, instead of calling `expt` each iteration:
        (black-box (expt 2 to-power))
        (loop (sub1 i)))))
> (let ([to-power (black-box 100)])
    (let loop ([i 1000])
      (unless (zero? i)
        ; in safe mode, calls `expt`, because `to-power` is not
        ; known to be a number; optimized away in unsafe mode:
        (expt 2 to-power)
        (loop (sub1 i)))))
> (let ([to-power (black-box 100)])
    (let loop ([i 1000])
      (unless (zero? i)
        ; arithmetic really performed every iteration, since the
        ; `to-power` value is assumed unknown, and the `expt`
        ; result is assumed to be used, even in unsafe mode:
        (black-box (expt 2 to-power))
        (loop (sub1 i)))))
```

Another example is where I currently sometimes write
```
(define (f-that-i-do-not-want-to-inline x)
  ....)
(set! f-that-i-do-not-want-to-inline f-that-i-do-not-want-to-inline) ; rely on `set!` pessimism
(for ([i (in-range N)])
  (f-that-i-do-not-want-to-inline i))
```
I can write

```
(define f-that-i-do-not-want-to-inline
  (black-box
    (lambda (x)
      ....)))
(for ([i (in-range N)])
  (f-that-i-do-not-want-to-inline i))
```


mflatt:
I don't think there's been enough discussion about the `racket/base` part of this PR to merge it.

Generally, I'm opposed to adding things to `racket/base`. Creating no conflicts with registered packages is a necessary condition, but not sufficient. One reason to put a binding in `racket/base` or even `racket` is that it's useful enough. There's also precedent for putting primitive compiler directives in `racket/base`.

I expect to use `black-box` almost as often as I use `time` going forward, so I'd be happy with convenience here. But for `black-box` to be in `racket/base`, it should probably be something that others also expect to use, and maybe that's not the case. Unless others jump to say that the convenience of `racket/base` really helps, then one of these is probably a better option:
* put it in `racket/black-box` module; or
* name it `#%black-box` following the convention of `#%` meaning "I'm sorry you had to see this, but I need a low-level thing here", but still in `racket/base` on the grounds that it is fundamental enough (along the same lines as `assert-unreachable`).

Bogdanp:
I'm also generally of the opinion that things should be added to `racket/base` sparingly, but in this case my vote would be to include it as is.

jackfirth:
I lean towards `racket/black-box`. I think end users are much more likely than compiler devs to use `time` without `black-box`, as compiler devs are more likely to be interested in the optimizer's behavior specifically rather than the performance of some large end-to-end application or library codepath. I can think of many instances where I wanted to use `time` but had no particular need to combine it with something equivalent to `black-box`.

That said, I don't feel too strongly about it.

rfindler:
My thoughts are all about precedent here. If we put this in, what criteria are we going to use to say "no" in the future and will that criteria feel capricious to future contributors?

Alternatively, will we just end up adding a lot of stuff to `racket/base`? I guess that adding lots of stuff to `racket/base` is a bad outcome but I don't feel particularly strongly about that being the eventual outcome. (I have not struggled with startup time the way @mflatt has .... 😅)

gus-massa:
It's somewhat like `void/reference-sink`?  Does `black-box` guaranty that the reference is not collected under the hood? Does  `void/reference-sink` guaranty that a mutable argument is not modified or stored somewhere?

Is it possible to optimize `black-box` like this? [I'm not planing to implement them for now.]

    (number? (black-box 7))   ==>   (begin (black-box 7) #t)

    (let ([b (box 7)]) (black-box b) (number? (unbox b)) )   ;==> no optimization

I'm using in some test `(zero? (random 1))` as a non-optimizable `#t`. A sufficient smart compiler can optimize it. I'm just wondering how it's related... 

    (define (hidden-true)  (let ([b (box #t|)]) (black-box b) (unbox b)))     ;==> no optimization, but it's #t

mflatt:
Yes, the `black-box` function implies everything that `void/reference-sink` does about its argument, but by returning its argument instead of `(void)`, `black-box` is a little more general. In this PR, the implementation of `(void/reference-sink x)` is changed to `(begin (black-box x) (void))`.

No, the following optimization shouldn't happen, because optimizer passes are supposed to specifically treat `black-box` as potentially returning any value:

```
(number? (black-box 7))   ==>   (begin (black-box 7) #t)
```

Yes, `(black-box #t)` is intended to be a replacement for `(zero? (random 1))`.

mflatt:
> If we put this in, what criteria are we going to use to say "no" in the future and will that criteria feel capricious to future contributors?

FWIW, I did offer a couple of criteria above:

> One reason to put a binding in racket/base or even racket is that it's useful enough. There's also precedent for putting primitive compiler directives in racket/base.

We might add to that "about as expensive to exclude from `racket/base` out as keep it in, since the implementation has to be built in". But it's not clear to me that these are sufficient reasons, still.

rfindler:
> FWIW, I did offer a couple of criteria above:

Right, sorry. I wrote too quickly. 

> One reason to put a binding in racket/base or even racket is that it's useful enough. 

I fear this will feel capricious when used to say "no", at least not without making it a little bit more concrete. 

> There's also precedent for putting primitive compiler directives in racket/base.

That one's clear and makes sense to me as a rule. 

> We might add to that "about as expensive to exclude from racket/base out as keep it in, since the implementation has to be built in". 

I am not understanding this one. What expense is involved with not putting something into racket/base?

mflatt:
> I am not understanding this one. What expense is involved with not putting something into racket/base?

This seems to detailed to matter, but: A primitive in `#%kernel` that isn't in `racket/base` goes in a list of things to exclude from a bulk import of `#%kernel`, which involves a small amount of symbol resolution and tracking. That's too small to measure, but it's about the same as keeping it included, where the symbolic name will instead stay in a summary table of bulk-imported names. This issue is specific to things that have to be built in enough, anyway, to cooperate with the compiler.

rfindler:
> > I am not understanding this one. What expense is involved with not putting something into racket/base?
> 
> This seems to detailed to matter, but: A primitive in `#%kernel` that isn't in `racket/base` goes in a list of things to exclude from a bulk import of `#%kernel`, which involves a small amount of symbol resolution and tracking. That's too small to measure, but it's about the same as keeping it included, where the symbolic name will instead stay in a summary table of bulk-imported names. This issue is specific to things that have to be built in enough, anyway, to cooperate with the compiler.

Oh, I think I get it now: when something has to go into `#%kernel` to be implemented then the cost of keeping it out of `racket/base` is basically the same as the cost of including it. 

Thus, for any new name that conflicts with nothing on the package server and has to be implemented at the `#%kernel` level, there is no strong argument for keeping it out of `racket/base`, so we might as well put them in and this might as well be an acceptable reason, globally, for putting new things into `racket/base`. Put another way, any argument against including such things would have to be about putting it into the language at all; once it meets that bar, there's no reason to focus on the `racket/base` inclusion anymore. (right?)

usaoc:
> In this PR, the implementation of `(void/reference-sink x)` is changed to `(begin (black-box x) (void))`.

Not very important, but I suppose `reference-sink` in `src/thread/sink.rkt` can also be replaced by `black-box`?

samth:
I think the fact that it's being used in other core parts of the system (like the `reference-sink` functions) is a reason to add it to `racket/base`.

mflatt:
Thanks everyone. I think we've arrived at a clear rationale for including `black-box` in `racket/base`.

> I suppose `reference-sink` in `src/thread/sink.rkt` can also be replaced by `black-box`?

Good point. I'll update that implementation.

PR #5347, opened 2025-10-06
doc: update url for “The WTF-8 Encoding” (Sapin 2018)

LiberalArtist:
Moved to <https://wtf-8.codeberg.page>: see <https://github.com/SimonSapin/wtf-8/commit/e27dc7c6e0d25e9276274ad42eb25dc30aba8c1f>, specifically <https://github.com/SimonSapin/wtf-8/blob/e27dc7c6e0d25e9276274ad42eb25dc30aba8c1f/index.html#L4-L6>.


sorawee:
I cannot access https://wtf-8.codeberg.page/ either.

Would it be a good idea to just link to the original page but with archive.org?

LiberalArtist:
> I cannot access https://wtf-8.codeberg.page/ either.

What is happening when you tried? Clicking on the link in your message worked for me.

sorawee:
Works for me now... Not sure what's going on yesterday.

Issue #5351, opened 2025-10-17
Cross-phase persistent submodules of non-xpp modules break with errortrace


jesboat:
**What version of Racket are you using?**

8.18.0.17-2025-10-17-3686704bf1
(from commit 3686704bf1)

on MacOS

**What program did you run?**

The following two programs:

```racket
(module outer '#%kernel
  (module u '#%kernel
    (#%declare #:cross-phase-persistent)))
```

**What should have happened?**

I'd expect it to run after clicking "Run", and then produce the "Interactions disabled: (quote #%kernel) does not support a REPL (no #%top-interaction)" message in the interactions pane.

**If you got an error message, please include it here.**

Instead, I get this as a syntax error:

module: cannot be cross-phase persistent due to required modules
  example: required module #<module-path-index:drracket/private/drracket-errortrace-key> is not cross-phase-persistent in: #:cross-phase-persistent

Presumably, this is DrRacket-specific, given that the problem module is in "drracket".

The program works fine if outer module is also cross-phase-persistent.

samth:
Do you have "Debugging" on in the language menu? What happens if you turn it off?

jesboat:
> Do you have "Debugging" on in the language menu? What happens if you turn it off?

Yes. Changing from "Debugging" to "No debugging or profiling" makes the error go away.

samth:
I think the fix here is that the errortrace annotator should just skip modules that are cross-phase persistent.

LiberalArtist:
That would make sense, but I think https://github.com/racket/racket/pull/5352 would still be useful more generally for debugging.

jesboat:
> I think the fix here is that the errortrace annotator should just skip modules that are cross-phase persistent.

@samth it's confusing that there's no issue when the outer module is cross-phase persistent, though.

> That would make sense, but I think [#5352](https://github.com/racket/racket/pull/5352) would still be useful more generally for debugging.

@LiberalArtist this and #5352 are orthogonal. That one is "make it easier to see why a module is ineligible to be cross-phase persistent"; this is "DrRacket is making some modules ineligible when it shouldn't".

rfindler:
FWIW, here's how to reproduce this outside of DrRacket (where `tmp3.rkt` contains the above code):

```
$  racket -l errortrace -t ~/tmp3.rkt
/Users/robby/tmp3.rkt:3:15: module: cannot be cross-phase persistent due to required modules
  at: #:cross-phase-persistent
  in: module
  errortrace...:
[...]

PR #5352, opened 2025-10-17
[expander] show which required module breaks cross-phase-perstence

jesboat:
Modules which are declared cross-phase persistent are only permitted to require other modules if those modules are also cross-phase persistent. Previously, the expander's error when this check fails was just:

    module: cannot be cross-phase persistent due to required modules

This is not particularly helpful. This modifies the expander to track which required module wasn't cross-phase-persistent (or an arbitrarily chosen such module if multiple required modules weren't cross-phase persistent) and include it in the error message. For example, #5351 triggers the message

    module: cannot be cross-phase persistent due to required modules
      example: required module #<module-path-index:drracket/private/drracket-errortrace-key> is not cross-phase-persistent in: #:cross-phase-persistent

(Note that #5351 shouldn't result in an error. I wrote this change to help figure out why it is, but the change is useful anyway.)

**Pending before commit:**

- The error message prints the module as a `~a`-formatted module-path-index, which is not ideal. On the other hand, cross-phase-persistent modules are a rare and advanced feature, and some error messages regarding them already include some implementation gunk (even references to expander-internal structs sometimes), so I don't think this is a big deal. (And TBH I was too lazy to figure out the right way to pleasantly render the mpi.)
- The `in:` field is rendered on the same line as the `example:` field in DrRacket. I think this is a DrRacket issue.


mflatt (commented):
I haven't tried building with this, yet, but it looks like a good idea and a fine implementation (even though I have little suggestions). Let us know when it's in a shape that you think should be merged, which might be the PR as-is.

jesboat:
(1) Added a prequel commit to strengthen the invariants tested for cross-phase-persistent modules in the test suite, and to enable a test for the content of the error message. (2) Rebased. (3) Eliminated "xpp". (4) Revised the error message to

    test.rkt:3:13: module: cannot be cross-phase persistent because some required module isn't
      at: #:cross-phase-persistent
      required module: #<module-path-index:racket/base>

(I kept the `/#f` in the `required-non-cross-phase-module/#f` name tho.)

Assuming all the CI passes, I have no objections to merging.

PR #5354, opened 2025-10-19
Record `literal-set->predicate` disappeared uses

### LINKED ISSUE #5353: Uses of `literal-set->predicate` don't add binding arrows
8.18.0.11-2025-08-08-74d4f401d8 [cs]

**What program did you run?**
```scheme
#lang racket/base
(require syntax/parse)

(define-literal-set arithmetic-functions
  (+ - * / expt sqrt))

(literal-set->predicate arithmetic-functions)
```

There should be an arrow in DrRacket from the definition of `arithmetic-functions` to its use inside `literal-set->predicate`.

jackfirth:
## Checklist

- [x] Bugfix
- [ ] tests included

## Description of change

Fixes #5353. Not sure how to test this though.

### DIFF
--- a/racket/collects/syntax/parse/private/litconv.rkt
+++ b/racket/collects/syntax/parse/private/litconv.rkt
@@ -246,20 +246,22 @@ cause an error, so don't worry about that case.)
 (define-syntax (literal-set->predicate stx)
   (syntax-case stx ()
     [(literal-set->predicate litset-id)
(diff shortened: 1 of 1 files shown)


jackfirth:
Alright, I've confirmed locally that this works at least.

mflatt:
@rmculpepper — Does this look ok to you?

rmculpepper:
Thanks, merged.

PR #5357, opened 2025-10-22
Make syntax->string work on boxes.

dvanhorn:
## Checklist

- [x] Bugfix
- [x] tests included

## Description of change

The `syntax->string` function from `syntax/to-string` does not currently work for literal boxes and instead prints the string representation of the inner syntax object (e.g. `(syntax->string #'(#&1))` produces `"#&#<syntax:17-unsaved-editor:3:21 1>"`).  This commit adds functionality to traverse the inner syntax object appropriately.  A couple simple tests are included.

The same problem exists for literal hash tables, but is more complicated to solve since (a) there are several different kinds of literal hash tables, and (b) the keys of the hash table are not syntax objects and so it's unclear how `syntax->string` should  print the result.


mflatt:
Thanks!

PR #5358, opened 2025-10-23
add `#:on-length-mismatch` to `for` and variants

mflatt:
When `for` traverses multiple sequences at once, iteration stops when any of the sequences terminates. That's often useful, and practically any use of `in-naturals` depends on that behavior. But sometimes it would be better to have an error, especially in a new abstraction that expands to a `for` variant. Checking the lengths of multiple sequences is possible in some cases, but that's not possible for all kinds of sequences, so building a hook into `for` is helpful.

This PR adds an `#:on-length-mismatch` clause to `for`. Its associate expression is evaluated just before deciding that iteration terminates (because one of the sequence ends) in the case that some sequence still has elements. The expectation is that the expression raises an exception, but if it produces a value, iteration terminates as if #:on-length-mismatch` wasn't present.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/for.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/for.scrbl
@@ -24,7 +24,8 @@ The @racket[for] iteration forms are based on SRFI-42
                            (code:line #:unless guard-expr)
                            (code:line #:do [do-body ...])
                            break-clause
-                           (code:line #:splice (splicing-id . form))]
+                           (code:line #:splice (splicing-id . form))
+                           (code:line #:on-length-mismatch mismatch-expr)]
                [break-clause (code:line #:break guard-expr)
                              (code:line #:final guard-expr)]
                [body-or-break body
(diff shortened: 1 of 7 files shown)


pkgs/racket-doc/scribblings/reference/for.scrbl
```diff
@@ -111,6 +113,15 @@ binding and expansion. Support for @racket[#:splice] clauses is
 intended less for direct use in source @racket[for] forms than for
 building new forms that expand to @racket[for].
 
+At @racket[#:on-length-mismatch mismatch-expr] clause is similar to
```
samdphillips: Change to `A @racket[#:on-length-mismatch mismatch-expr] clause`

mflatt: Fixed, but I went with `An`.

mflatt:
The most recent version of the commit here fixes an interaction with `in-parallel` and `in-indexed`, where the latter is built with `in-parallel` and `in-naturals`. Previously, `in-parallel` was always spliced into the enclosing `for` as an optimization, but with `#:on-length-mismatch`, that triggers an error when the starting form was `in-indexed`. Now, `in-parallel` is spliced only if there's no `#:on-length-mismatch` afterward.

At the same time, `in-parallel` is limited, because it requires each parallel sequence to produce a single value. So, the latest commit also adds the generalization `in-parallel-values`.

gus-massa:
Too late, but probably not important.

Is it possible to add `in-naturals/do-not-complain-if-the-other-sequence-is-shorter`?

mflatt:
I considered that direction, but it doesn't work out nicely, especially with the way that `in-naturals` and similar can be used as functions to create a sequence. Having that intent expressed by combining via `in-parallel` worked out much more nicely.

Issue #5364, opened 2025-11-02
internal error on thread-receive in parallel thread


rmculpepper:
On Racket 9.0.0.3 [cs], the following program causes Racket to crash with an internal error:

    > (thread #:pool 'own (lambda () (thread-receive)))
    #<thread>
    internal-error: tried to deschedule a descheduled thread
     at "internal-error"
     at "do-thread-deschedule!"
     at "thread-deschedule!"
     at "touch-blocked"
     at "call-in-empty-metacontinuation-frame"

I expected `(thread-receive)` to act like other operations that temporarily move the computation to the backing coroutine thread.

PR #5365, opened 2025-11-02
fix thread-receive in parallel thread

rmculpepper:
See #5364. I'm not completely sure this is the right fix, but it seems to follow the comments in src/thread/atomic.rkt and the implementation pattern of unsafe-semaphore-wait.

## Checklist

- [x] Bugfix
- [ ] Feature
- [x] tests included
- [ ] documentation


mflatt:
I'm trying to double-check, but so far, this looks right to me.

mflatt:
Thanks!

PR #5367, opened 2025-11-03
make make-struct-type accept `'current` for an inspector

jesboat:
make make-struct-type accept `'current` for an inspector

Summary:

Currently, `make-struct-type` defaults to `(current-inspector)` if no inspector
is specified. However, specifying an inspector is mandatory to specify anything
involving optional args which come after, e.g. that fields should be immutable.
At first glance, this is fine.

However, `make-struct-type` is allowed at the module level of
cross-phase-persistent modules, which have certain restrictions, and
`(current-inspector)` can't be specified explicitly. This leaves authors of
cross-phase-persistent in an unpleasant situation: either have structs be
opaque but always mutable, with no guard, etc.; or have structs which can be
immutable but are transparent.

So, change `make-struct-type` to allow `'current` for an inspector to mean
the current inspector.

There are a couple ways to alternate ways to resolve this tension. Allowing
`(current-inspector)` as a cross-expr in general sounds riskier than modifying
make-struct-type in a backwards-compatible manner. Another would be to ban
non-transparent structs, which seems very undesirable. Finally, we could leave
things inconsistent, which seems sad.

Thanks to @mflatt for helping with the documentation update, and the
corresponding changes to BC's optimizer in next commit.

Test Plan:

Tests added.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/struct.rktl
+++ b/pkgs/racket-test-core/tests/racket/struct.rktl
@@ -1957,19 +1957,27 @@
 ;; ----------------------------------------
 
 (let-values ([(insp) (make-inspector)])
-  (let-values ([(type1 make1 pred1 sel1 set1) (make-struct-type 'foo1 #f 0 0 #f null insp)]
+  (let-values ([(type0 make0 pred0 sel0 set0) (make-struct-type 'foo1 #f 0 0 #f null)]
+               [(type0x make0x pred0x sel0x set0x) (make-struct-type 'foo1 #f 0 0 #f null (current-inspector))]
+               [(type1 make1 pred1 sel1 set1) (make-struct-type 'foo1 #f 0 0 #f null insp)]
                [(type2 make2 pred2 sel2 set2) (make-struct-type 'foo2 #f 0 0 #f null 'current)]
                [(type3 make3 pred3 sel3 set3) (parameterize ([current-inspector insp])
                                                 (make-struct-type 'foo3 #f 0 0 #f null))]
+               [(type3x make3x pred3x sel3x set3x) (parameterize ([current-inspector insp])
+                                                     (make-struct-type 'foo3 #f 0 0 #f null (current-inspector)))]
                [(type4 make4 pred4 sel4 set4) (parameterize ([current-inspector insp])
                                                 (make-struct-type 'foo4 #f 0 0 #f null 'current))])
+    (err/rt-test (begin (struct-type-info type0) #t) exn? "inspector")
+    (err/rt-test (begin (struct-type-info type0x) #t) exn? "inspector")
     (test '#t (begin (struct-type-info type1) #t))
     (err/rt-test (begin (struct-type-info type2) #t) exn? "inspector")
     (test '#t (begin (struct-type-info type3) #t))
+    (test '#t (begin (struct-type-info type3x) #t))
     (test '#t (begin (struct-type-info type4) #t))
     (parameterize ([current-inspector insp])
       (err/rt-test (begin (struct-type-info type1) #t) exn:fail? "inspector")
       (err/rt-test (begin (struct-type-info type3) #t) exn:fail? "inspector")
+      (err/rt-test (begin (struct-type-info type3x) #t) exn:fail? "inspector")
       (err/rt-test (begin (struct-type-info type4) #t) exn:fail? "inspector"))))
 
 ; We also try with a module which we hope gets the optimization which does
--- a/racket/src/bc/src/optimize.c
+++ b/racket/src/bc/src/optimize.c
@@ -470,7 +470,9 @@ static Scheme_Object *extract_specialized_proc(Scheme_Object *le, Scheme_Object 
 int scheme_omittable_expr(Scheme_Object *o, int vals, int fuel, int flags,
                           Optimize_Info *opt_info, Optimize_Info *warn_info)
      /* Checks whether the bytecode `o` returns `vals` values with no
-        side-effects and without pushing and using continuation marks.
+        side-effects and without pushing and using continuation marks,
+        except that `make-struct-type` is allowed to use continuation marks
+        unless `OMITTABLE_REALLY_NO_MARKS` is in `flags`.
         A -1 for `vals` means that any return count is ok.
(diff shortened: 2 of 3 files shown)


mflatt:
This idea seems fine, but I think the new argument value should be `'current` instead of `#t`.

As noted in your Discord discussion, this PR raises an underlying issue, which is that a cross-phase-persistent module picks up some state when it is instantiated by getting the current inspector. That doesn't seem good, but I also don't see an alternative. (It's not just structure types, but also the instantiation-time code inspector associated with the module.)

jesboat:
Revising to 'current instead of #t. Once CI finishes, I'm going to convert this back to a draft in the hopes of adding a couple tests (and actually confirming first-hand it works on BC.)

pkgs/racket-test-core/tests/racket/struct.rktl
```diff
@@ -1956,4 +1956,22 @@
...
+(let-values ([(insp) (make-inspector)])
+  (let-values ([(type1 make1 pred1 sel1 set1) (make-struct-type 'foo1 #f 0 0 #f null insp)]
+               [(type2 make2 pred2 sel2 set2) (make-struct-type 'foo2 #f 0 0 #f null 'current)]
+               [(type3 make3 pred3 sel3 set3) (parameterize ([current-inspector insp])
+                                                (make-struct-type 'foo3 #f 0 0 #f null))]
+               [(type4 make4 pred4 sel4 set4) (parameterize ([current-inspector insp])
+                                                (make-struct-type 'foo4 #f 0 0 #f null 'current))])
+    (test '#t (begin (struct-type-info type1) #t))
+    (err/rt-test (begin (struct-type-info type2) #t) exn? "inspector")
+    (test '#t (begin (struct-type-info type3) #t))
```
jesboat: > `(test '#t (begin (struct-type-info type3) #t))`

This test, and only this test, fails on BC (macos aarch64, base rev f4eb1b9a7a) and I do not have the foggiest idea why. If anybody watching sees something obviously wrong, or could suggest a possible next step for me to debug, that would be great.

mflatt:
The test failure was due to a bug in BC that predates your changes: the optimize could drop a `with-continuation-mark` around a `make-struct-type` where the latter consults the parameterization.

I've added a commit to the branch here to fix that bug. I considered pushing a separate fix, but the optimizer needs to be updated for `make-struct-type` with `'current`, and it seems better to just have everything together. If you'd prefer to rearrange the patches at that level, it's ok with me.

I also added another commit as a documentation suggestion: use `(current-inspector)` instead  or `current-inspector` and (more importantly) adding a `history` note. I'm imagining that you'd merge that to your commit.

Overall, I think these changes are ready to merge when it's in a state that you like.

mflatt:
On further thought, this should have a version bump, since it can change the way that a `make-struct-type` call is compiled (where the default `'current` gets inlined into the original call site).

Also, I notice now that `make-struct-type-install-properties` needs to be updated from `#t` to `'current`, right?

jesboat:
@mflatt,

Thanks for the review and debugging!

> The test failure was due to a bug in BC that predates your changes: the optimize could drop a with-continuation-mark around a make-struct-type where the latter consults the parameterization.
>
> I've added a commit to the branch here to fix that bug. I considered pushing a separate fix, but the optimizer needs to be updated for make-struct-type with 'current, and it seems better to just have everything together. If you'd prefer to rearrange the patches at that level, it's ok with me.

Thank you! I don't think it's worth it to split the commit, although I did update the commit messages to refer to each other.

> I also added another commit as a documentation suggestion: use (current-inspector) instead or current-inspector and (more importantly) adding a history note. I'm imagining that you'd merge that to your commit.

Thank you, done.

> Overall, I think these changes are ready to merge when it's in a state that you like.

Ready, modulo any CI failures, and the version bump and `make-struct-type-install-properties` comment below.

> On further thought, this should have a version bump, since it can change the way that a make-struct-type call is compiled (where the default 'current gets inlined into the original call site).

Added, I think.

> Also, I notice now that make-struct-type-install-properties needs to be updated from #t to 'current, right?

I think I did that, but maybe I missed something?

Jade

jesboat:
The `#t`-versus-`'current` in the default args with `make-struct-type-install-properties` is a bug. This is sufficient to trigger it:

```
(module u racket/base
  (define-values (struct:test7 test71.1.unreadable:test71.1
                  test7? test7-f7)
    (let-values ([(struct:_4 make-_5 ?_7 -ref_7 -set!_8)
                      (make-struct-type 'test7 #f 1 0 #f
                        (list (cons prop:liberal-define-context #t)))])
      (values
        struct:_4
        make-_5
        ?_7
        (make-struct-field-accessor -ref_7 0 'f7)))))
```

I was surprised to not hit this before; I've been running DrRacket from this PR for a while now. Empirically, it looks like the criteria are:

- the `make-struct-type` call (in the source) to be in a context that the optimizer expects for when it lowers to a `make-struct-type-install-properties`
- but not generated by `struct` or `define-struct` (they fill out the full argument list)
- no supertype, no auto fields, `#f` for the auto values is specified (otherwise the optimization isn't attempted, and a `make-struct-type` call is in the schemified)
- properties are specified (otherwise there's no need to call `make-struct-type-install-properties`)
- no `prop:procedure`, no guard, no explicit inspector, no immutables, no explicit constructor name (otherwise the inspector needs to be explicit in the source

It seems ...difficult to write a reliable test for this and I am kinda inclined to not. Are you okay committing with the bug fixed, but no explicit test for it?

mflatt:
> It seems ...difficult to write a reliable test for this and I am kinda inclined to not. Are you okay committing with the bug fixed, but no explicit test for it?

I'm not sure I follow here. Doesn't the `module` form you wrote work as a test? It could be dropped into a ".rktl" file with a `(require 'u)` afterward. For these kinds of tests, I usually use a long (presumably distinct) name for the module instead of `u`.

jesboat:
I suspect that would work. The thing I feel unease at is making sure the test continues to actually test the right thing, in the sense that if the behavior of the optimization changes even slightly, then the test would start compiling down to a lower-level `make-struct-type` call and not the `make-struct-type-install-properties` and would then be silently worthless. I'll add it, though.

PR #5377, opened 2025-11-18
complex tan/tanh

### LINKED ISSUE #3324: improving tan of complex arguments
To implement other math functions I need a better tan for complex arguments. I can define one locally but that seems a pity.

a start for this better tan function would be:
```
(define (mytan z)
  (cond
    [(< (imag-part z) -354) 0.0-1.0i]
    [(< 354 (imag-part z)) 0.0+1.0i]
    [else
     (define a (* 2 (real-part z)))
     (define b (* 2 (imag-part z)))
     ;(define denom (+ (cos a) (cosh b)))
     (define denom (real-part (* 2 (cos z) (cos (conjugate z)))))
     (println (list (+ (cos a) (cosh b)) (real-part (* 2 (cos z)(cos (conjugate z))))))
     (make-rectangular
      (/ (sin a) denom)
      (/ (sinh b) denom))]))
```
unfortunately tan is defined in [https://github.com/racket/racket/blob/42cb80bc70e1fb97dabd1e0a2dff5d93257b1f40/racket/src/racket/src/number.c#L2877](numbers.c) and sinh is not available at this point (I'm guessing), so I don't know how to proceed.

Areas where this function outperforms the previous one:
- large imag-part (>800) => correctly go to 0±1
- small imag-part and real-part near k*π/2 (provided that sinh uses an improved implementation)

I need to check further to see if there are no points where this performs worse than the current implementation, but so far it looks ok.

bdeket:
Prevent underflow in the calculation of complex-`tan` and complex-`tanh`

Although this improves the results in the affected regions, both have still inaccuracies up to 8ulp (real+imag)

closes #3324 

continuing discussion there

## Checklist
- [X] Feature
- [X] tests included


bdeket:
With this change maximum observed `flulp-error` is now 6 in my random testing.

Due to the mismatch of primitive functions exported by CS and BC, for `tanh` (in racket/math on CS) the code now transforms this in terms of `tan` (in CS) that transforms it back to `tanh`.
The alternative would be to bring the exports of BC in line with CS.

soegaard:
@bdeket @pavpanchekha @mflatt 

I just fell over this one, but confused that there were two PRs.

Is this the new one - and is it ready to commit?

bdeket:
I think there is only 1 PR (this one), and 1 Issue #3324.
This PR is ready - apart from the test included I also did random testing against a `bigfloat` implementation in both CS and BC.
But I was hoping someone would review the BC implementation. And maybe address the comment from gambiteer in the issue (regarding too large code change for too little gains)

PR #5382, opened 2025-11-27
add `exn-classify-errno`

mflatt:
Filesystem and network exceptions usually have an `errno` field that provides more specific information, but it's a platform-specific encoding. The `exn-classify-errno` function can convert the most common errors to a symbol that represents an error in a portable way, so result symbols include `'ENOENT` as "file not found," `'EEXIST` as "file exists already," and `'EACCESS` as "permission denied."

It's implemented at the rktio level based on `lookup-errno` in BC's FFI and an LLM-generated Windows-to-Posix mapping.


mflatt:
Merged as 01bd403766b46e52ed646a82d09441f4c83c70fc

PR #5384, opened 2025-11-30
Add MSVC query to system-type 

ndykman:
## Checklist

- [x] Feature

## Description of change
This change allows developers to query if Racket was built locally via Microsoft Visual Studio (and not cross compiled). 
The main purpose of this is to allow for different dynamic libraries to be used if that is the case. This is needed as vcpkg dependencies for Microsoft Visual Studio use a different convention for naming libraries (many don't have lib as the start of the filename). It does so by checking for a specific environment variable that is set on a command prompt or Powershell prompt for development, but not by the vcvarsall.bat script. 

As part of this, the libcrypto and libssl libraries have been updated to use version 3 of openssl. There are no new tests as the existing tests cover functionality. A working version of racket and raco have been built from source and they run without any issues.  

This work also supports updating dynamic libraries in packages like draw-lib, gui-lib and math-lib to use MSVC versions as part of testing and development in Windows and to insure MSVC builds work in general.  

Note: There is no intent here to have MSVC builds be officially distributed or supported.

### DIFF
--- a/racket/src/bc/src/string.c
+++ b/racket/src/bc/src/string.c
@@ -189,7 +189,7 @@ static void cache_locale_or_close(int to_bytes, rktio_converter_t *cd, char *le)
 
 ROSYM static Scheme_Object *sys_symbol, *sys_os_symbol, *sys_arch_symbol;
 ROSYM static Scheme_Object *link_symbol, *machine_symbol, *vm_symbol, *gc_symbol;
-ROSYM static Scheme_Object *so_suffix_symbol, *so_mode_symbol, *word_symbol;
+ROSYM static Scheme_Object *so_suffix_symbol, *so_mode_symbol, *word_symbol, *msvc_symbol;
 ROSYM static Scheme_Object *os_symbol, *os_star_symbol, *arch_symbol;
 ROSYM static Scheme_Object *fs_change_symbol, *target_machine_symbol, *cross_symbol;
 ROSYM static Scheme_Object *racket_symbol, *cgc_symbol, *_3m_symbol, *cs_symbol;
@@ -251,6 +251,7 @@ scheme_init_string (Scheme_Startup_Env *env)
   REGISTER_SO(so_suffix_symbol);
   REGISTER_SO(so_mode_symbol);
   REGISTER_SO(word_symbol);
+  REGISTER_SO(msvc_symbol);
   REGISTER_SO(os_symbol);
   REGISTER_SO(os_star_symbol);
   REGISTER_SO(arch_symbol);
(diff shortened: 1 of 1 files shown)


mflatt:
A way to trigger a different library-linking mode on Windows makes sense, but I'm skeptical that it should be based on whether Racket was compiled with MSVC. After all, compiling with MSVC is mainly the way I work on Windows currently. I think we want something based on a configure-time choice, instead.

The way to reflect it into the Racket level is through `(system-library-subpath)`.  Although `(system-library-subpath)` was originally related to picking up native-compiled forms of Racket code, it evolved into the mechanism for selecting library conventions. (In the old, old days, compiling with MSVC vs. MinGW produced different `(system-lbrary-subpath)` results. But that was before I spent a lot of effort getting Racket to behave the same with either compiler.)

A different `(system-library-subpath)` result will cause different native-library selections at level of Racket's package system, which is part of what you want. Library names at run time also can also be conditioned on `(system-library-subpath #f)`, which is something that we've done in the past.

Racket CS is almost set up for this. Try changing `racket/src/cs/c/winfig.bat` to add `echo PLT_CS_SLSP_SUFFIX=-alt >> Makefile` after other similar lines that initialize `Makefile`, and also set the environment variable `PLT_CS_SLSP_SUFFIX` to `-alt` before running `nmake`. That should cause `(system-library-subpath #f)` to have `-alt` added to the end.

If that looks like a promising direction, we'd want to add a way to configure without changing `winfig.bat` and without remembering to set an environment variable — and some way that doesn't involve having to remember to pass a configuration to `nmake` (or `make` or even `configure`) every time. I'd be happy to work on that, if initial experiments adjusting `winfig.bat` work out.

ndykman:
> A way to trigger a different library-linking mode on Windows makes sense, but I'm skeptical that it should be based on whether Racket was compiled with MSVC. After all, compiling with MSVC is mainly the way I work on Windows currently. I think we want something based on a configure-time choice, instead.

I completely agree here and I apologize that I didn't catch this earlier. It is really about the target binary: `x64-windows-dynamic` or `x64-mingw-dynamic` (using vcpkg conventions). Of course, the dynamic part is a given here. 

It seems that `(system-library-subpath)` would be the direction to go here as there are hooks into the package manager already. My plan will be to set `PLT_CS_SLSP_SUFFIX` to something like `msvc_native` and go from there.  

I have some questions on how to best integrate this with the `(define-runtime-lib)` macro. My current approach was to add something like `(windows-native)` to support a completely separate list of file names. Again, this wasn't just about names being different, but to allow for additional libraries to be loaded or to exclude libraries that don't need to be loaded. 

Leaving open for now until further information is gathered.

mflatt:
Your plan for `define-runtime-lib` sounds ok to me.

The more I think out it, `system-library-subpath` needs to be part of the story, but matching the result of `system-library-subpath` is not great. In part, it's not great because an architecture is part of the returned subpath, but it's not relevant to the library choice. It's nicer to match a simple result from `system-type`, as in your patch.

If I understand, the distinction you want to make is the same as the "natipkg" (use native libraries supplied by Racket packages) versus "nonatipkg" (use packages ad managed by the system) distinction that has been evolving already. That distinction is currently reflected consistently only in `system-library-subpath`, and it just happens that `(system-type 'os)` and version search lists for "Unix" libraries bridged the gap, but making the distinction explicit would make sense in general.

So, what you do think about adding `(system-type 'native-package)` to report `'racket` or `'system`, or something like along those lines, where is would be configured at the same time as an adjustment to `system-library-subpath`?

ndykman:
Closing as this is superseded by #5387.

PR #5387, opened 2025-12-02
add `(system-type 'platform)`, `(system-type 'os-find)`, and `--enable-sofind=<conv>`

mflatt:
The new `system-type` modes refine the way that platforms are identified for the purposes of (1) locating native libraries at run time, and (2) picking platform-specific Racket packages like "racket-win32-arm64-3" or "racket-x86_64-linux-natipkg-3". The new `--enable-sofind=<conv>`argument for `configure` (or `/sofind <conv>` for `winfig.bat`) configures the result of `(system-type 'os-find)`, which is in turn included as a suffix on `(system-type 'platform)` and `(system-library-subpath #f)`.

* `(system-type 'platform)` returns the same string as `(path->string (system-library-subpath #f))`, but it in a simpler and clearer way. This addition will enable a migration away from `system-library-subpath` , which was not really intended for the role it has ended up serving.
    
* `(system-type 'so-find)` reports `'system`, `'natipkg`, or another symbol that is configured with `--enable-sofind`. This should make it easier to condition native-library choices on an environment (selected at build time) without having to parse a suffix out of `(system-library-subpath #f)`. For example, when using libraries managed by MacPorts or vcpkg, then the names or versions of the libraries can be different than in Racket packages like "racket-win32-x86_64-3", and we might update Racket libraries to try different names depending on a combination of `(system-type)` and `(system-type 'so-find)`.

As a bonus, arguments that would otherwise be specified with `make CONFIGURE_ARGS=....` or `make PKGS=....` can be stashed in `build/CONFIGURE_ARGS` or `build/PKGS`. That way, a Git checkout can be configured without having to remember the arguments with every `make`.

Related to #5384

### DIFF
--- a/main.zuo
+++ b/main.zuo
@@ -195,6 +195,7 @@
   (define (build-base subdir token [options (hash)])
     (define also? (hash-ref options 'also? #f))
     (define prefix (hash-ref options 'prefix #f))
+    (define msvc? (eq? 'windows (hash-ref (runtime-env) 'toolchain-type)))
     (define targets
       (configured-targets-at (hash
                               'configure (if (eq? 'windows (hash-ref (runtime-env) 'toolchain-type))
@@ -203,8 +204,10 @@
                               'inputs (list (at-source "racket/src" subdir "Makefile.in"))
                               'outputs (list (at-source "racket/src/build" subdir "Makefile"))
                               'configure-args (append
-                                               (shell->strings (shell-subst-pwd (lookup 'CONFIGURE_ARGS_qq)
-                                                                                (at-source "racket/src/build" subdir)))
+                                               (if msvc?
+                                                   '()
+                                                   (shell->strings (shell-subst-pwd (lookup 'CONFIGURE_ARGS_qq)
+                                                                                    (at-source "racket/src/build" subdir))))
                                                (shell->strings (lookup 'CONFIGURE_ARGS))
                                                (let ([f (at-source "build/CONFIGURE_ARGS")])
                                                  (if (file-exists? f)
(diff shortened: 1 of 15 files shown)


ndykman:
Commenting as to motivation. There are number of build choices for native libraries (and for the racket platform itself). This pertains mostly for Windows, but there are also clang-based builds on Linux as well. As such, matching up what dynamic library is to be loaded into racket (and what packages those libraries come from) needed to be a bit more flexible.   

The landscape in terms of Windows has notably changed. For a while, `mingw64` (and `msys2`) were needed to build a lot of libraries used by cross-platform software. Today, many libraries can be built natively by the Microsoft Visual Compilers (MSVC) toolchain. This is enabled by `vcpkg`, which is a cross platform tool for managing and building third-party libraries. 

However, `mingw64` and `msvc` will use different naming conventions by default. Trying to force one to use the other overly complicates things. So, the best thing to do is use these features to change what libraries are loaded in the racket code itself. This makes sense as new versions of the libraries might require additional libraries to be loaded. 

The `mingw` toolchain is also changing. For a long time, `mingw` builds used `msvcrt` as the core C runtime library. This library is ancient. `mingw`, as version 12 (2024-05-29),  uses the `ucrt` library by default; this is available on Windows by default and  is completely modern (full Unicode, C99 support). The current dynamic libraries for racket still use the older mingw builds.

Change doesn't happen overnight. By adding this feature, it should make it much easier to test new tool chains and slowly move what the `natipkg` target is for future versions. 

I have validated using vcpkg to manage all the third party dynamic libraries that racket (and associated libraries) use and created a build of racket (and DrRacket) with MSVC using those libraries. This version loads and is working in day to day use.

PR #5389, opened 2025-12-05
use the 'sub-range-binding property in struct-copy

rfindler:
This changes the `struct-copy` macro so that it uses the [newly added](https://github.com/racket/drracket/commit/7b87b4b9de098911fb6c6433f399c00cd262139e) support for `'sub-range-binding`.

For example:

Screenshot 2025-12-04 at 7 23 57 PM

Without this commit that would have only a single arrow (to the `posn-x`).


mflatt:
LGTM

rfindler:
Thanks!

I just realized I can lift the "string length" computation outside the loop. I'll do that before pushing.

PR #5395, opened 2025-12-10
Include 'system in keyword list

soegaard:
Leaving `system` out of the keyword list happens to work, since `system` is handled in the last clause. My guess is that `system` wasn't left out on purpose though.

## Checklist

- [ ] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation


mflatt:
Thanks!

PR #5398, opened 2025-12-11
BC: Make interned-char? equivalent to char?

gus-massa:
After making chars play nice with be `eq?`, it's better to sync the implementation of `kernel:interned-char?` to return `#t` for all chars, so it's the same than `char?` like in CS.

This is an internal undocumented function that is used only in `optimize.c` because the types need a pointer to a function. It's not strictly necessary to fix this, the difference only cause a few minor missed optimizations.

I'm not sure about backward comparability in the `#%kernel` module, because another possibility is to just remove this function completely.


mflatt:
Thanks! I'm inclined to just leave `interned-char?`, too.

PR #5400, opened 2025-12-12
Replace `gen-temp.rkt` with new `make-stx-id-counter` in stx.rkt

jesboat:
Currently, "gen-temp.rkt" exists to provide just `gen-temp-id`, which some low-level macros use to (surprise) generate IDs for temporaries. This used a global counter (modulo multiple instantiations of gen-temp.rkt), which is entirely unnecessary. The only uses of `gen-temp-id` are:

- cond, which can just use a simple single hygenic identifier. (Why was it using gen-temp-id? I've no idea, but it's been that way as far back as the svn history goes, as far as I can tell.)

- with-syntax, which uses `gen-temp-id` to bind the evaluation results of the RHS's of the syntax bindings. These identifiers need to be unique within any invocation of with-syntax, but standard hygene will take care of everything beyond that.

- syntax-case, which uses it similarly to `with-syntax`.

We replace `gen-temp-id` with `make-stx-id-counter`, a new function in stx.rkt which maintains a local counter; `with-syntax` and `syntax-case` use that, and `cond` just uses a boring identifier.

This allows us to eliminate gen-temp.rkt, which lives at the bottom of the module hierarchy, and reduces module instantiations by `racket/base` by ~3%.

Test plan: (1) code review the `cond` changes; (2) manually skim the `make derived` changes, and the only ones are alpha-renaming from the `cond` change which gives me confidence that with-syntax and syntax-case are still working; (3) DrRacket runs; (4) CI.


mflatt:
This looks like a good idea to me. It looks like an "expobs.rktl" test is failing, but that's likely a regression test that is looking for a trace matching a recorded one, and the solution is almost certainly to update the recording.

samth:
Just for historical interest, here's the previous implementation of `cond` before @mflatt added `gen-temp-id` in v3.99.02: https://github.com/racket/racket/blob/11d145b7392df482ab4e0a8dcbe5b8ab88b9a5d0/src/mzscheme/src/startup.ss#L595 It doesn't have any mutable state.

jesboat:
I'll look into the test.

@samth, it's got a gensym at https://github.com/racket/racket/blob/11d145b7392df482ab4e0a8dcbe5b8ab88b9a5d0/src/mzscheme/src/startup.ss#L638 and https://github.com/racket/racket/blob/11d145b7392df482ab4e0a8dcbe5b8ab88b9a5d0/src/mzscheme/src/startup.ss#L653 which is (I'd call it) global state. That traces back to the first `src/mzscheme/src/startup.ss` in src/mzscheme/src/startup.ss (svn: r4).

samth:
Ah ok, counting gensym that goes back to the beginning of startup.ss here: https://github.com/racket/old-plt/blame/47ffa80baf8782badf90a2545453de9f06bbab8e/src/mzscheme/src/startup.ss and further back to https://github.com/racket/old-plt/blob/3a0f0320d6b793ce1e5bf05aff04f77f67a0141c/src/mzscheme/src/macro.ss#L192. Before that I think everything used the zodiac macro system and isn't really comparable.

jesboat:
> It looks like an "expobs.rktl" test is failing, but that's likely a regression test that is looking for a trace matching a recorded one, and the solution is almost certainly to update the recording.

The test failing is https://github.com/racket/racket/blob/1a74e93caa/pkgs/racket-test-core/tests/racket/expobs.rktl#L119-L121, the trace of a simple `define-syntax-rule` expansion.

After modifying expobjs.rktl to regenerate the trace, the only difference was the removal of one `track-syntax` event: https://github.com/racket/racket/blob/1a74e93caa/pkgs/racket-test-core/tests/racket/expobs-regression.rktd#L560

I don't have a good sense of where to go from here... try to figure out what that line is coming from, and how this change would have affected it? Update the trace without understanding why? Infer there's a problem with the change? Advice welcome.. :/

mflatt:
Yes, just update the trace. It's not reporting a problem with the change, just that something about scope sets encountered during the expansion has changed — and that should expected.

mflatt:
Thanks!

PR #5401, opened 2025-12-14
RFC: [expander] cross-phase persistent: allow more quote, quote-syntax

jesboat:
# Description of commit

Previously, `quote` at the module level (i.e. on the rhs of a `define-values` and not wrapped in a lambda, i.e. would run at module instantiation time) was limited to a small subset of atomic data, and `quote-syntax` was forbidden to appear anywhere in the module.

Loosen these restrictions, significantly. Specifically:

Define *simple data* as:

- numbers (including extflonums), booleans, symbols, chars, keywords, null
- immutable strings and immutable bytes
- immutable vectors with simple data for values
- immutable hasheqs with simple data for keys and values
- immutable boxes with simple data for content
- immutable prefab structs with simple data for field values

(These are the things the standard reader can read, except that it excludes non-hasheq hashes, because cross-phase-persistent modules previously allowed hasheqs but not other hashes.)

Then, the restrictions for `quote` are then relaxed so that any simple data can be quoted at the module level. (As before, arbitrary data can be quoted when guarded by a lambda.)

For `quote-syntax`, the restriction to simple data is enforced throughout the module. This is necessary because the stx's quoted by a quote-syntax are, in essence, lifted to the module level.

# RFC

Thoughts? In particular:

- Should "loosen restrictions on quote" and "allow quote-syntax" be separate commits, or separate PRs?
- Allowing quote-syntax seems okay, at least insofar as it works for some simple cases. Does this work all the time, or are there sharp edges I'm not aware of? Does it break expander invariants?
- If palatable, I'd want to add tests, update the docs, and add appropriate changelog entries before merging.

One potential application of allowing quote-syntax is to allow transformers to be written in a cross-phase-persistent module. However, that couldn't (by itself, at least) reduce the number of instances or bindings needed, since the cross-phase-persistent module can't create the syntax bindings.


mflatt:
I have been unclear on whether `quote-syntax` is going to work out, but now I'm pretty sure that it will need extra changes to make sense. I think the changes are maybe not all that difficult, but we should figure out whether it's going to work before allowing `quote-syntax`.

Here's a module of the sort that I think you're aiming for:

```
;; "m1.rkt"
(module m1 '#%kernel
  (#%require (for-syntax '#%kernel))
  (#%provide m)
  (#%declare #:cross-phase-persistent)
  (define-syntaxes (m)
    (lambda (stx)
       (quote-syntax ok)))
  (define-values (ok) "yes"))
```

If I hack the expander to (1) skip the check whether `#:cross-phase-persistent` is ok, and (2) make `namespace-module-instantiate!` run the phase-1 component of a cross-phase persistent module (using `(instantiate! 0 1 (or (namespace-root-namespace ns) ns))` and `(instantiate! 0 0 (or (namespace-root-namespace ns) ns))`), then this works:
```
;; "m1.rkt"
(module m2 '#%kernel
  (#%require "m1.rkt")
  (display m)
  (newline))
```
However, this does not:
```
(module m3 '#%kernel
  (#%require (for-syntax '#%kernel
                         "m1.rkt"))
  (define-syntaxes (m3)
    (begin
      (display m)
      (newline)
      void)))
```
The problem is that `m3` is using `m` at phase 1, and the way "m1.rkt" is imported allows the `m` macro to be found, but expansion injects an `ok` that has a binding only at phase 0.

To make this work, `module` expansion needs to create a different kind of scope of its body. In `expand-module`, where `inside-scope` is created by `new-multi-scope`, it needs to be `new-scope`. That will make module-level binding information cross-phase. (Also, in "root-expand-context.rkt", `apply-post-expansion` and `post-expansion-scope` need to accommodate a `scope?` instead of just `shifted-multi-scope?`.) There may be more problems that larger experiments will uncover, but it's plausible to me that this can work out.

The remaining problem is that `inside-scope` needs to be created — and so the choice between `new-scope` and `new-multi-scope` needs to be made — well before `#%declare` forms are discovered in the current module-expansion process. Maybe the `module` expander will need to look for `(#%declare #:cross-phase-persistent)` in a special way (looking for it literally?).

As an aside, cross-phase bindings in a `(#%declare #:cross-phase-persistent)` module would mean that phase 0 and phase 1 cannot have different bindings for the same identifier. I image that's as expected, though.

jesboat:
My interest in quote-syntax in cross-phase-persistent modules was mainly for core-ish macro, many of which (and, or, when, unless, the non-core `let`s, non-keyword non-optional `define`s, a couple others) can be implemented pretty-straightforwardly in `racket/kernel`, where quote-syntax is limited to capturing bindings exported by `#%kernel` and the source location. (Bindings to core forms can also be procured during the dynamic extent of a transformer using `namespace-module-identifier`.) If `quote-syntax` would be simple to allow for that, then the bodies of transformers could be in cross-phase-persistent modules. If there was also a way for cross-phase-persistent modules to provide some of their bindings as syntax bindings instead of value bindings, then many (most?) of the simpler macros could live in a cross-phase-persistent module. (Notable exceptions would be `cond` and `quasiquote`, which need to `quote-syntax` e.g. `else` and `unquote`.)

Sadly, it seems that `quote-syntax` does have sharp edges, and @mflatt and I have intuition ([discord](https://discord.com/channels/571040468092321801/893314076346826852/1449806943756615851)) that flattening is likely to be a bigger bang-for-the-buck.

Given that, I'm tempted to:
- Split the `quote` changes out into a PR (with tests and docs) which could be merged with less concern
- Shelve consideration of `quote-syntax` in cross-phase-persistent modules (or of implementing macros in them more generally)
- Focus effort on flattening

Sound ok?


> The remaining problem is that inside-scope needs to be created — and so the choice between new-scope and new-multi-scope needs to be made — well before #%declare forms are discovered in the current module-expansion process. Maybe the module expander will need to look for (#%declare #:cross-phase-persistent) in a special way (looking for it literally?).

In the current model, looking for it literally almost works: the only syntax bindings cross-phase-persistent modules can end up with are core ones, none of which (except begin splicing, I suppose) can expand to `#%declare`, but it would get uglier because (AIUI) the `#%declare` should be detected by `free-identifier=?` which would mean processing `#%require`'s....  If I imagine writing `module` as a macro, I feel like I'd end up with something which, when it hits the `#:cross-phase-persistent` request, restarts the entire expansion process with cross-phase-persistent restrictions in place. (I do not mean to suggest that the expander should work that way, or that it would be a reasonably-scoped change.)

jesboat:
> Given that, I'm tempted to:
> 
> * Split the `quote` changes out into a PR (with tests and docs) which could be merged with less concern
> * Shelve consideration of `quote-syntax` in cross-phase-persistent modules (or of implementing macros in them more generally)
> * Focus effort on flattening
> 
> Sound ok?

Going forward with this plan. The `quote` changes are #5407. Closing this one now.

PR #5402, opened 2025-12-14
Make racket kernel entirely kernel

jesboat:
# Description

make `#lang racket/kernel` entirely kernel

The catch before was that the reader submod used `module-reader`, which is quite complex and decidedly not written in `#%kernel`. However, perhaps unsurprising given `#lang racket/kernel`'s simplicity, only the most basic features of module-reader were used, and it is quite straightforward to inline the necessary components and rewrite them in `#%kernel`.

This change means that the `#lang` syntax can be used to write modules which are entirely independent from everything other than "kernel.rkt" itself.

# Notes

~~Pending before merge:~~

- ~~Is this desirable? Arguably the additional complexity in `kernel.rkt` is still too much "just" to be able to use the `#lang` syntax in more places in the core implementation.~~ Discussed in comments.
- ~~Probably should have a note in the history file.~~ History file is updated at release time.
- ~~Is it possible to easily test this?~~ I think "no".

### DIFF
--- a/racket/collects/racket/kernel.rkt
+++ b/racket/collects/racket/kernel.rkt
@@ -1,7 +1,233 @@
 (module kernel '#%kernel
+
   (#%provide (all-from '#%kernel))
 
   (#%declare #:cross-phase-persistent)
 
-  (module reader syntax/module-reader
-    #:language 'racket/kernel))
+
+  (module reader '#%kernel
+    (#%module-begin
+
+     (#%require '#%paramz)
+
+     (#%provide (rename lang:read read)
+                (rename lang:read-syntax read-syntax)
+
+                get-info
+                get-interaction-info)
+
+     (#%declare #:cross-phase-persistent)
+
+     (define-values (the-lang) (quote '#%kernel))
+
+
+     ;
+     ;
+     ;                                ;     ;
+     ;   ;;;;;;                       ;     ;
+     ;   ;    ;;                      ;
+     ;   ;     ;  ;;;;     ;;;    ;;; ;   ;;;    ; ;;;    ;;; ;
+     ;   ;     ;  ;  ;;   ;   ;   ;  ;;     ;    ;;   ;   ;  ;;
+     ;   ;    ;; ;    ;       ;  ;    ;     ;    ;    ;  ;    ;
+     ;   ;;;;;   ;;;;;;   ;;;;;  ;    ;     ;    ;    ;  ;    ;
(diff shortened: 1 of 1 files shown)


samth:
This seems fine as an approach but wouldn't moving the submodule to racket/kernel/lang/reader.rkt and not doing anything otherwise be simpler?

mflatt:
There's also already the option of writing `(module m '#%kernel ....)`. Can you say more about why the combination of `#lang` and minimal build dependencies would be useful compared to writing `(module ....)`?

(FWIW, I don't see how moving the submodule to a "reader.rkt" would help.)

jesboat:
The things this was meant to address is that, in the current state of affairs, (a) `#lang racket/kernel` can't be used to implement anything transitively used by module-reader and (b) it can't be used to implement cross-phase-persistent modules IIRC, since they pick up implicit dependencies on their reader. (I might be misremembering the second one.)

Personally, I prefer `#lang` to `(module X '#%kernel ...)`, and generally was thinking of it as a desideratum to enable use of that in lower-layers of the stack. (I like avoiding the extra layer of indentation, and needing the parentheses-pair which span the entire module.) Happy to abandon.

mflatt:
I do think you're mistaken about whether `#lang racket/kernel` can be used for cross-phase persistent modules. This works:
```
#lang racket/kernel
(#%declare #:cross-phase-persistent)
(define-values (x) 10)
```

But you're certainly right that `#lang racket/kernel` cannot be used to implement anything that `syntax/module-reader` depends on. My sense is that this is too much to avoid the extra parentheses of `(module ....)`, but maybe I have a high tolerance for using different (and less convenient) languages for lower layers.

jesboat:
> I do think you're mistaken about whether #lang racket/kernel can be used for cross-phase persistent modules.

Yeah, I don't know why I thought that.

> But you're certainly right that #lang racket/kernel cannot be used to implement anything that syntax/module-reader depends on. My sense is that this is too much to avoid the extra parentheses of (module ....), but maybe I have a high tolerance for using different (and less convenient) languages for lower layers.

Want me to close, or wait a bit to see if others have thoughts?

jesboat:
Any chance of a merge, or should I close?

mflatt:
My vote is still not to do this, because the amount of needed code is seems a lot more than the parentheses it will save.

jesboat:
Ack. Closing.

PR #5406, opened 2025-12-20
discussion/proposal: merging some lower-level racket/base stuff together for perf reasons

jesboat:
Making this PR mostly as a discussion point for @mflatt, @samth, and others who are interested.

This code is not intended for review as-is.

# Background

For the past $time, I've been hacking around with the Racket implementation, specifically the parts below `#lang racket/base` and above Chez, with an idea towards finding potential performance improvements in things that matter to people. In discord, @samth suggested a couple possible places where performance improvements could be helpful

> - reduce Racket startup time for running a single expression
> - make it more palatable to write cli utilities in Racket by reducing their overhead
> - make documentation builds faster by making namespaces faster to create
> - reducing the footprint of big libraries like rhombus, which is what motivated compiler/demod in its current form

As a first pass, I was poking through the lower layers of racket/base's in-Racket implementation. It turns out some of the low-level modules can be merged together, with relatively little loss of expressiveness. Specifically:

- cond.rkt
- define-et-al.rkt
- private/define.rkt
- member.rkt
- qq-and-or.rkt
- reverse.rkt
- one or two more IIRC

can be combined into one `core-macros.rkt` and cross-phase-persistent `pico.rkt`.

This branch is approximately what I benchmarked using https://gist.github.com/jesboat/0ef53a4c93ea0f1a154c1460d5b5ecdb to get https://docs.google.com/spreadsheets/d/1Lb5_tRu825SazTYaPPONXDSCHIDY79rcNLNcgYc5Its/edit?gid=778923661#gid=778923661 (see also [discord](https://discord.com/channels/571040468092321801/893314076346826852/1449071065425903786)), although it has been rebased since then. It also appears to show significant improvements in startup time (e.g. `time racket -I racket/base </dev/null`).

# This PR

This commit is a snapshot of my working tree, which includes:

- above changes
- #5402 "make `#lang racket/kernel` entirely kernel" which is maybe not going to be merged at all
- peppering a bunch of `(#%declare #:require=define)` in racket/base dependencies for unrelated reasons
- maybe a couple other things which was accidentally included

The development history is squashed out, and I would not consider this PR as an actual request to pull anything. Rather, it's (a) a place for me to attach the working tree and (b) a place to discuss proposed changes in the comments.

# What next?

My suggestion would be for reviewers (/me glances awkwardly at @mflatt?) to:
- skim the PR as a sanity check.
- look at https://github.com/jesboat/racket/blob/merge-lower-layers-squashed/racket/collects/racket/private/core-macros.rkt and https://github.com/jesboat/racket/blob/merge-lower-layers-squashed/racket/collects/racket/private/pico.rkt (see below)
- inspect the branch locally to assess perf gains as desired

If the perf gains are worth the loss of expressiveness from combining the various lower-layer modules, then I would start breaking the changes into smaller logical commits which could be reviewed and (hopefully) merged.

As a final note, the main distribution does not currently compile with this because compatibility-lib's mzscheme imports some private stuff from racket/base. Minimal changes to make it work with this branch are https://gist.github.com/jesboat/9c987896b94deb1d554597fbfe384cd5, which is also something I'd need to tidy up before potential upstreaming.

#  Overview of new files

core-macros.rkt
- begin-for-phases: new macro
- and/or: previous implementation, no changes (I have an alternate implementation lying around which unifies both of them and I think reduces O(n^2) to O(n) behavior, but it breaks some optimizations)
- when/unless: fresh implementation
- cond: previous implementation, but with `let`, `let*`, let-loop, and/or, qq unfolded
- let_: previous implementation
- define-values-for-syntax: new implementation
- define, define-syntax, define-for-syntax: new implementation, which is intended to support everything except keywords and optional args. It can replace the pre-keyword `define` forms in private/define.rkt and also the `-define` forms in define-et-al
- top-interaction: essentially the same implementation (there's not many ways to do it)
- let/ec: new implementation
- qq-and-or.rkt: old implementation, no changes

pico.rkt:
- member: new implementation
- memw: previous implementation
- reverse: previous implementation
- car+cdr: new function

### DIFF
--- a/pkgs/racket-test-core/tests/racket/stx.rktl
+++ b/pkgs/racket-test-core/tests/racket/stx.rktl
@@ -556,7 +556,7 @@
 ;; Check tracking of (formerly) primitive expanders
 
 (test '(let) (tree-map syntax-e) (syntax-property (expand #'(let ([x 10]) x)) 'origin))
-(test '(let*) (tree-map syntax-e) (syntax-property (expand #'(let* ([x 10]) x)) 'origin))
+(test '((let*) let*-values let*) (tree-map syntax-e) (syntax-property (expand #'(let* ([x 10]) x)) 'origin))
 (test '(let) (tree-map syntax-e) (syntax-property (expand #'(let loop ([x 10]) x)) 'origin))
 (test '(letrec) (tree-map syntax-e) (syntax-property (expand #'(letrec ([x 10]) x)) 'origin))
 (test '(let*-values) (tree-map syntax-e) (syntax-property (expand #'(let*-values ([(x) 10]) x)) 'origin))
--- a/pkgs/racket-test-core/tests/racket/syntax.rktl
+++ b/pkgs/racket-test-core/tests/racket/syntax.rktl
@@ -2254,9 +2254,9 @@
 (syntax-test #'(let*) #rx"missing binding")
 (syntax-test #'(let* ([x 10])) #rx"missing body")
 (syntax-test #'(let) #rx"missing name or")
-(syntax-test #'(let x) #rx"missing binding pairs")
-(syntax-test #'(let ([10 10])) #rx"not an identifier")
-(syntax-test #'(let x ([10 10])) #rx"not an identifier")
+(syntax-test #'(let x) #rx"missing binding pairs or")
+(syntax-test #'(let ([10 10])) #rx"missing binding pairs or")
+(syntax-test #'(let x ([10 10])) #rx"missing body")
 (syntax-test #'(letrec) #rx"missing binding")
 (syntax-test #'(letrec ([x 3])) #rx"missing body")
 
--- a/racket/collects/racket/private/core-syntax.rkt
+++ b/racket/collects/racket/private/core-syntax.rkt
@@ -17,8 +17,6 @@
              call/ec let/ec
              let let* letrec
              let*-values  ; {let,letrec}-values are kernel
-             let-syntax letrec-syntax
-             let-syntaxes letrec-syntaxes
              quasiquote
              and or)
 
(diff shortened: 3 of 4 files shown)


mflatt:
This seems like a good idea. I have been worried about (1) the compatibility issues (as you mention); (2) whether `core-macros.rkt` and/or `pico.rkt` might be more difficult to maintain; and (3) whether performance improvements will be measurable. But for (1), I think you're clear on the compatibility question, and there can be compatibility stubs as needed. For (2), the more time I spend with `core-macros.rkt` and `pico.rkt`, the more they look just better. For (3), I don't know, but it seems like this is on a path to improving the implementation independent of performance improvements, anyway.

jesboat:
@mflatt awesome. I'll start on breaking it up into reviewable chunks, adding the stubs, etc.

jesboat:
Status update:

- The groundwork for this (e.g. #5425, #5431, racket/compatibility#15) has been fully merged
- The first couple PRs from these changes proper (#5441, #5447, #5450) have been merged. More PRs to follow :)
- I just pushed an update to this PR with a snapshot of my current working copy, including an okay-ish commit history. There is again a bit of unrelated stuff included, but it's a bit better now.
- Startup times ("start REPL, evaluate `(exit)`") no longer differ to the extent that I could trivially benchmark them, even on the commits I benchmarked originally. I do not know why.
- More reliable benchmarks on various combinations of namespace creation and evaluating expressions ([code](https://gist.github.com/jesboat/0ef53a4c93ea0f1a154c1460d5b5ecdb), [results](https://docs.google.com/spreadsheets/d/1fmY3u3jKHWybAzjfTeIdTnGY-JoKcizd13VwEGPoJyk/edit?gid=1486117898#gid=1486117898)) do still show improvements, ranging from 2.9% on real time for "create namespace, require racket/base, evaluate an expression" to 11% on CPU time for "just `make-base-empty-namespace`"

jesboat:
Status update:
- The duplicate identifier changes have been extended and split into an orthogonal PR, https://github.com/racket/racket/pull/5464 (not yet merged)
- https://github.com/racket/racket/pull/5453 was created and merged
- https://github.com/racket/racket/pull/5473 was created and should be merge-ready within a couple days

After those, all of qq-and-or.rkt, define-et-al.rkt, cond.rkt, define.rkt, top-int.rkt, and performance-hint.rkt will use only kernel syntax, and can be merged into an omnibus file, tentatively named core-syntax.rkt, completing this series.

PR #5407, opened 2025-12-20
[expander] cross-phase persistent: allow more quote

jesboat:
# commit description

[expander] cross-phase persistent: allow more quote

Previously, `quote` at the module level (i.e. on the rhs of a `define-values`
and not wrapped in a lambda, i.e. would run at module instantiation time) was
limited to a small subset of atomic data.  Loosen the restrictions,
significantly. Specifically:

Define *cross-data* as:

- numbers (including extflonums), booleans, symbols, chars, keywords, null
- immutable strings and immutable bytes
- immutable vectors with cross-data for values
- immutable hasheqs with cross-data for keys and values
- immutable boxes with cross-data for content
- immutable prefab structs with cross-data for field values

(These are the things the standard reader can read, except that it excludes
non-hasheq hashes, because cross-phase-persistent modules previously allowed
hasheqs but not other hashes.)

Then, the restrictions for `quote` are then relaxed so that any cross-data can
be quoted at the module level. (As before, arbitrary data can be quoted when
guarded by a lambda.)

The code is written slightly oddly to allow for easy addition of a check
for `quote-syntax` if that becomes eventually allowed; for more on that, see
https://github.com/racket/racket/pull/5401

# context

https://github.com/racket/racket/pull/5401 proposed this and also allowing quote-syntax, but quote-syntax doesn't work in all cases, and is more speculative. this is the smaller change which is just quote.


mflatt:
Sorry for the long delay here! Paging back in, this generally makes sense to me. So, it does just need updated docs and tests.

jesboat:
@mflatt no worries! just as a heads up, i'm about to head mostly AFK for a trip, so i might not be able to circle back to this for a ~3 weeks, but i'll do the docs/tests then. hope you had a good holiday season!

jesboat:
Pushing an update here which adds tests and updates the reference manual. I don't know what version to list in the `@history[#:changed ...]` section, so I wrote 9.2 as a placeholder, but that should probably be something else?


I also took the opportunity to
- Fix the docs to note that `begin` splicing can happen as an expansion step
- Reword one tidbit in `syntax.scrbl` to have a more helpful link
- Fix two prior tests for cross-phase-persistent modules which were malformed
- Update the test cross-phase-persistent test helper function to make it harder to have malformed tests

Those changes didn't seem worth the cognitive overhead of dedicated PRs, but they are separate commits.


When writing the tests, I realized there are actually a handful of inconsistencies between the current docs and implementation (e.g. the implementation checks the result arity of cross-exprs, which is undocumented; the docs prohibit various primitive forms as cross-expr which the implementation allows.) Resolving those inconsistencies seems like it would require changes outside the scope of this PR, so I haven't.

mflatt:
I bumped the version number and adjusted `history` to match the new number, and I plan to merge (unsquashed) soon.

PR #5416, opened 2026-01-13
fix intermittent zip-round-trip test failure

samth:


mflatt:
Thanks!

samth:
I've been annoyed about this bug for a long time.

PR #5419, opened 2026-01-19
Fix documentation of `unsafe-flbit-field`

pavpanchekha:
This PR fixes the documentation of `unsafe-flbit-field`, which was added in 8.15.0.3 and seems to have been documented in a bit of a hurry.


mflatt:
Thanks!

PR #5422, opened 2026-01-30
Fix flags for flbit-field.

samth:
This occurred in DrDr here: https://drdr.racket-lang.org/72051/racket/share/pkgs/math-test/math/tests/flonum-tests.rkt after @pavpanchekha's changes to `math/flonum`.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/flonum.rktl
+++ b/pkgs/racket-test-core/tests/racket/flonum.rktl
@@ -444,6 +444,14 @@
           flbit-field 3.141579e132 i j)))
 
 ;; ----------------------------------------
+;; Regression test to make sure that flbit-field does not internally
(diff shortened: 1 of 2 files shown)


samth:
All the actual debugging here was done by Claude Code; I just wrote the patch.

mflatt:
Thanks! I updated the test to one that reliably fails without the repair.

samth:
The unsafe-flbit-field flags are wrong too, I think.

mflatt:
Good point! I pushed an update here.

pavpanchekha:
Thanks, didn't know this was a thing

samth:
These notifications should go to you instead of Neil these days.

pavpanchekha:
Oh sure yeah

PR #5425, opened 2026-02-08
Tidying and clarifying racket/private modules used by other packages

jesboat:
This PR (two commits) is part of an effort to clarify what in `racket/private` is intended to be used (or is used) by other first-party packages. The overall plan is:

(1) Change first-party packages to use public APIs when possible.

(2) For any packages which have racket/private dependencies which are part of racket/base's implementation, make a new `racket/private/for-thatpackage` module which provides only the appropriate chunks.

(3) For packages which have dependencies on racket/private which are not part of racket/base's implementation, add a comment to the appropriate file noting that it's used by the other package.

This plan is less ambitious than it sounds because parts (1) and (2) consist only of `compatability-lib` and one line in `drracket` and part (3) is just comments.

The overall list was found by
`grep --exclude '*.zo' --exclude '*.dep' --exclude compiled --exclude .trash -r racket/private racket/share/pkgs | sort  | perl -lne 'print if m{(?<![a-z0-9-])racket/private}'`
in a built working copy of the main repo.

This effort excludes Typed Racket, which is (at my current level of familiarity with it, at least) un-untangleable.

### DIFF
--- a/racket/collects/racket/private/class-internal.rkt
+++ b/racket/collects/racket/private/class-internal.rkt
@@ -1,4 +1,18 @@
 #lang racket/base
+
+;; Note that this module only semi-private, in the sense that other packages
+;; which are in the main distribution but in separate repositories require
+;; it. This means any backwards-incompatible API changes need coordination.
+;; The following packages are known to use this module:
+;;
+;; - compatability-lib
+;;
+;; Please try not to add more packages which require this file. Instead,
+;; consider whether a public API is appropriate, or if a currently-private
+;; API should be made public. If not, then make a new file
+;; "racket/private/for-yourpackage.rkt" which exports the necessary
+;; definitions. See "racket/private/for-compatability-lib.rkt" for an example.
+
 (require (for-syntax racket/base)
          (only-in racket/list remove-duplicates)
          racket/stxparam
--- a/racket/collects/racket/private/generic.rkt
+++ b/racket/collects/racket/private/generic.rkt
@@ -1,4 +1,18 @@
 #lang racket/base
+
+;; Note that this module only semi-private, in the sense that other packages
+;; which are in the main distribution but in separate repositories require
+;; it. This means any backwards-incompatible API changes need coordination.
+;; The following packages are known to use this module:
+;;
+;; - data-lib
+;;
+;; Please try not to add more packages which require this file. Instead,
+;; consider whether a public API is appropriate, or if a currently-private
+;; API should be made public. If not, then make a new file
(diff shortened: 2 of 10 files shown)


jesboat:
friendlyping

rfindler:
Hi-- I'm working on a change to `object-c` which, along the way, renames `class-c-old.rkt` to `class-c.rkt` (as the change to `object-c` obviates the work that I started when making `class-c-new.rkt`, a looong time ago). The changes are nearing completion and are on these branches:
* [racket](https://github.com/rfindler/racket/tree/new-object-c)
* [compatbility](https://github.com/rfindler/compatibility/tree/new-object-c)
* [typed racket](https://github.com/rfindler/typed-racket/tree/new-object-c)

samth:
This looks good to me. I agree that currently Typed Racket is not worth trying to disentangle; it should just be changed as needed to support your changes to Racket.

jesboat:
> Hi-- I'm working on a change to `object-c` which, along the way, renames `class-c-old.rkt` to `class-c.rkt` (as the change to `object-c` obviates the work that I started when making `class-c-new.rkt`, a looong time ago). The changes are nearing completion and are on these branches:

@rfindler, I can skip adding the comment to `class-c-old.rkt` if you'd prefer?

rfindler:
Whatever you feel is best; hopefully we finish up the new object/c implementation soon and merge that, so another option is to merge this after that. That said, I'm not expecting the performance results to slow us down, but they might.

jesboat:
It was easy enough to update the PR to skip edits to class-c-old.rkt, so I went ahead and did that. :)

rfindler:
It may make sense to make compatibility-lib not depend on the private parts of the contracts. It can probably expand into a use of `object-contract` instead of the private constructor function.

jesboat:
> It may make sense to make compatibility-lib not depend on the private parts of the contracts. It can probably expand into a use of `object-contract` instead of the private constructor function.

@rfindler I have no objection :) but it seems orthogonal to this PR?

jesboat:
I think this is ready to merge?

otherjoel:
I'm not sure if my code review comment came through but “compatible” and “compatibility” are misspelled throughout this PR.

jesboat:
@otherjoel they did not :(

https://github.com/racket/racket/pull/5434 and https://github.com/racket/compatibility/pull/17 fix the spelling, but it renames the module so it requires simultaneous commits in both repos, and would be the second cross-repo breaking change in 3 hours... up to the reviewers whether to take the hit or leave it spelled incorrectly

Sorry everybody...

PR #5428, opened 2026-02-13
A new implementation of object/c

rfindler:
This implementation uses impersonator properties to interpose on `object-ref` but, instead of returning a new class, it returns a special value that indicates that this object has a contract on it. Method invocation and field access all check for the new value and faciliate contract checking when they see it.

Along the way, replace instanceof/c, object-contract, and dynamic-object/c so they all use this same implementation internally.

This commit also abandons the approach in class-c-new (which was an attempt in this same general direction, but not as good), so there is now just one class-c file (which is based on class-c-old).

----

@bennn is running the TR benchmarks to confirm the change isn't problematic (he already ran an earlier version of this change and found some good speedups; this latest run is to make sure the finalized version of the code is still good).

This also requires coordinated changes in the compatibility-lib and typed-racket-lib.


samth:
Have you looked at whether there are other packages on pkgs.racket-lang.org that require some of the private modules you've changed?

rfindler:
> Have you looked at whether there are other packages on pkgs.racket-lang.org that require some of the private modules you've changed?

The big one is, I think, that class-c-old.rkt that got renamed. These are all of the places that grep turns up when looking for "class-c-old" in the code registered with the pkg server as of January 14th:

```
./fmt/tests/test-cases/large.rkt:         ;; for class-c-old.rkt:
./fmt/tests/test-cases/large.rkt:         ;; end class-c-old.rkt requirements
./fmt/tests/benchmarks/class-internal.rkt:         ;; for class-c-old.rkt:
./fmt/tests/benchmarks/class-internal.rkt:         ;; end class-c-old.rkt requirements
./racket-test/tests/racket/contract/object.rkt:  ;; `new-cls` (in class-c-old.rkt, currently line 1368))
./typed-racket-lib/typed-racket/utils/opaque-object.rkt:         (only-in racket/private/class-c-old
./compatibility-lib/mzlib/class.rkt:         racket/private/class-c-old)
./compatibility-lib/mzlib/private/contract-object.rkt:         racket/private/class-c-old
```

Maybe that's good enough before merging and we can use the pkg build to sort out any other changes after merging? Or are there other checks to do at this stage?

samth:
Yeah that looks like it's just the things you have PRs for plus comments. I'm pleasantly surprised.

rfindler:
I was worried about this before and did some checks (that's why the pkg build code I got is from a month ago) and I am happy to put stuff back based on what happens when the pkg build runs. The rename is just to be less embarrasing, not anything substantial!

PR #5430, opened 2026-02-15
add `#%foreign-inline`

mflatt:
The `#%foreign-inline` syntactic form is a new case in the grammar of fully expanded expressions. The "foreign" part of `#%foreign-inline` is the underlying compiler and runtime system that handles linklets --- so, Chez Scheme in the case of Racket CS.

The `#%foreign-inline` form is unsafe, and correct use is up to the programer who writes a `#%foreign-inline` form. Specifically, the form wrapped by `#%foreign-value` must not refer to the enclosing environment (the macro expander will prevent that, anyway), throw an exception, or otherwise rely on the current continuation.

A `#%foreign-value` form can have a `#:pure` annotation to tell the Racket compiler (at the schemify level) that the form has no side effects, and it can have `#:copy` to enable propagation of the `#%foreign-value` form like a constant.

Adding a new form to the fully expanded grammar means that any tool that processes expanded Racket programs will need to handle it, eventually. The new form will not show up at first --- nothing in the main Racket libraries will use it, for now --- and tools not yet updated will only potentially break when it starts being used in some relevant program. Small updates to `errotrace` and `web-server` are ready to go.

This new form is motivated by work in progress on a revised FFI that is better turned to Chez Scheme than `ffi/unsafe`. The new FFI needs to use of syntactic forms in Chez Scheme, instead of just functions, and `#%foreign-inline` provides a general mechanism for those kinds of tasks that must be closely tied to the underling implementation.


samth:
The windows CI failure seems to be an actual bug.

samth:
Based on offline conversation with @mflatt, I found it helpful to think of this as "like `vm-eval`, but at compile-time".

mflatt:
Merged as fc04bbf154226cec11893ae4f07f16fda4e37bcd

gus-massa:
IIUC a long time ago I tried something like `(if ((vm-eval 'fxzero?) x) "O" "X")` but it was super slow, in particular I think `cp0/cptypes` were not able to see `fxzero?`.  (I don't remember the details.) I guess this solves the speed problem.

PR #5431, opened 2026-02-15
remove unused "break-parameterization" struct (easy)

jesboat:
Previously, the module defined two structs:

1. Struct `struct:break-paramz` (defined with `make-struct-type` and, confusingly, with reflexive name `break-parameterization`) used for the break parameterization

2. Struct `break-parameterization` (defined with the only remaining use of `-define-struct`) which is unused locally and not `provide`d anywhere. This goes at least as far back as v3.99.0.2, git 39cedb62edf, svn r7706.

Remove the unused one, which will enable removing `-define-struct` once https://github.com/racket/compatibility/pull/15 is also merged


mflatt:
Thanks!

PR #5435, opened 2026-02-18
for/foldr: Add missing syntax-local-introduce

lexi-lambda:
When I initially added `for/foldr` in #2656, I omitted a use of `syntax-local-introduce` on the identifier provided to `#:delay-with`. I have no explanation for why I chose to do this, as it is simply wrong: the identifier is smuggled into the expansion via the closure produced by a helper macro, but it is morally an input to the macro, so it needs `syntax-local-introduce`.

Omitting this usage doesn’t cause any immediate problems, but it does cause the identifier to gain a macro-introduction scope, which makes it no longer `original?`, breaking Check Syntax. This PR corrects the issue.


mflatt:
Thanks!

PR #5437, opened 2026-02-19
Make racket/{list,string,promise} more extraction-friendly.

samth:
Change `apply apply` to avoid a dependency on `keyword-apply`.

Avoid fancy uses of `letrec` that are not provably safe.


mflatt:
For the `apply apply` problem, would it make sense to change `new-apply` in "pre-base.rkt" to recognize the pattern and produce plain `apply apply`? It looks like it would be a relatively easy change, but I might be missing something.

samth:
I thought about that but this seemed like a simpler change. I'd be happy to implement that idea and we'll see what it looks like.

samth:
I've implemented that idea. It's always helpful to be reminded why writing macros with pattern matching is good.

PR #5438, opened 2026-02-20
ffi/serialize-cstruct test: add _gcable to nonatomic pointer fields

samth:
Two inner test modules defined `_aq` with `[d _aq-pointer/null]` (missing `_gcable`) while using `#:malloc-mode 'nonatomic`. Without `_gcable`, reading the d field returns a non-gcable wrapper whose raw C address is not updated when GC moves the struct. This produces stale pointers that read garbage data or crash with "invalid memory reference".

Add `(_gcable ...)` to the d field in both inner modules so the wrapper is GC-traced and updated when the struct moves. Add a regression test that verifies data read through a wrapper remains correct after GC and that serialize/deserialize preserves cycles.

### DIFF
--- a/pkgs/racket-test-extra/tests/ffi/serialize-cstruct.rkt
+++ b/pkgs/racket-test-extra/tests/ffi/serialize-cstruct.rkt
@@ -25,6 +25,7 @@
 
 (define aq1 (make-aq/mode (make-fish 6) #f))
 (set-aq-d! aq1 aq1)
+(collect-garbage)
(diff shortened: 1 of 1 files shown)


mflatt (approved):
This looks like the right test repair to me.

The three `(collect-garbage)`s in a row in the regression test seem like overkill — the test fails with one `(collect-garbage)` if the cruicial `__gcpointer` is removed — but ok.

samth:
I removed the extra collect-garbage calls.

PR #5444, opened 2026-02-24
net/http-client: propagate decoding errors to http-conn-receive!ing thread

Bogdanp:
In #5296, I added make-pipe/err to handle RSTs received by the pumping thread, but missed that decoding also happens through a background thread. This change makes the decoding threads also use make-pipe/err.

Related to Bogdanp/racket-http-easy#35.


mflatt:
Thanks!

PR #5448, opened 2026-02-26
xml,reader: slightly improve read-xml performance 

Bogdanp:
I noticed the repeated lookups on `collapse-whitespace` were taking a long time while profiling `read-xml` on a 14MB file, so this seemed like an easy win. With this change in place, reading that file takes about 339ms, versus about 382ms before.

The change to use `for/and` instead of `andmap` doesn't make a difference in this case, but it can't hurt, either. Happy to drop it, though.


mflatt:
Thanks!

PR #5449, opened 2026-02-26
Fix flaky test with promises and concurrency

jesboat:
For the code

    (let* ([p (delay/sync (sync never-evt))]
           [th (thread (lambda () (force p)))])
      (test #f promise-forced? p)
      (test #f promise-running? p)
      (sync (system-idle-evt))
      ...)

it's unlikely with cooperative threads that `th` starts to force the promise before `promise-running?` has a chance to look at it, but not outright impossible. Add a bit of synchronization to make sure things happen in the order we want them to.


mflatt:
Thanks!

PR #5450, opened 2026-02-27
define-et-al.rkt: remove dead `-define-struct` + add section headers

jesboat:
# define-et-al.rkt: remove unused -define-struct macro

This is currently unused. The most recent three uses were: a prior version of `syntax-parameterize` (circa 2018), an ad-hoc struct which was itself dead for even longer (finally removed in https://github.com/racket/racket/commit/bf48460cf4498e47c975103a250e6449dfefbc6f), and an unnecessary reference in compatability-lib (see racket/compatability#15).

Now that those are all gone, we can remove it.

# define-et-al.rkt: add section headers

[No detailed commit message]


mflatt:
Thanks!

PR #5457, opened 2026-03-02
Rename `qq-and-or.rkt` to `core-syntax.rkt`; add section headers

jesboat:
This rename makes twofold sense. First, `qq-and-or.rkt` is already more than just what it says on the tin-- it also defines `let`, `let*`, and `letrec`. Second, as part of the effort to reduce startup time, we plan on merging more core macros into the same file; for more info about the overall effort, see https://github.com/racket/racket/pull/5406

Testing: manually reviewed `git grep qq-and-or` in this repo; the only hits are #%requires which this commit adjusts. Searched all `*.rkt` and `*.ss` files in my copy of the package archive; the hits are `stxparse-info` which appears to include a copy of a decent chunk of racket/base's implementation and is broken anyway, and one hit in TR which will need a coordinated change.

# NOTE coordinated change in TR

This requires a backwards-compatible change to Typed Racket before this can be committed, so TR is aware of the rename. The changes do not need to be simultaneous. The PR on the Typed Racket side is https://github.com/racket/typed-racket/pull/1499 . The proper order for things is probably (1) get approval for both PRs, (2) merge the PR to Typed Racket and wait a couple hours for the package server to update, (3) merge this.

### DIFF
--- a/racket/collects/racket/phase+space.rkt
+++ b/racket/collects/racket/phase+space.rkt
@@ -1,5 +1,5 @@
 (module phase+space '#%kernel
-  (#%require "private/qq-and-or.rkt")
+  (#%require "private/core-syntax.rkt")
   
   (#%provide phase?
              space?
--- a/racket/collects/racket/private/case.rkt
+++ b/racket/collects/racket/private/case.rkt
@@ -3,8 +3,8 @@
 ;;       [http://scheme2006.cs.uchicago.edu/07-clinger.pdf]
 
 (module case '#%kernel
-  (#%require '#%paramz '#%unsafe "qq-and-or.rkt" "cond.rkt" "define.rkt" "fixnum.rkt"
-             (for-syntax '#%kernel "define-et-al.rkt" "qq-and-or.rkt" "cond.rkt"
+  (#%require '#%paramz '#%unsafe "core-syntax.rkt" "cond.rkt" "define.rkt" "fixnum.rkt"
+             (for-syntax '#%kernel "define-et-al.rkt" "core-syntax.rkt" "cond.rkt"
                          "stxcase-scheme.rkt"
                          "qqstx.rkt" "define.rkt" "sort.rkt" "fixnum.rkt"
                          "stx.rkt"))
--- a/racket/collects/racket/private/collect.rkt
+++ b/racket/collects/racket/private/collect.rkt
@@ -1,5 +1,5 @@
 (module collect '#%kernel
-  (#%require "qq-and-or.rkt"
+  (#%require "core-syntax.rkt"
              "path.rkt"
              "kw.rkt"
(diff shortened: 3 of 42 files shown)


jesboat:
> This requires a backwards-compatible change to Typed Racket before this can be committed, so TR is aware of the rename. The changes do not need to be simultaneous. The PR on the Typed Racket side is https://github.com/racket/typed-racket/pull/1499 . The proper order for things is probably (1) get approval for both PRs, (2) merge the PR to Typed Racket and wait a couple hours for the package server to update, (3) merge this.

Friendly ping on step 1 for this PR: confirming that we still want to move forward with merging the low-level modules, and confirming that `core-syntax.rkt` is an acceptable filename?

jesboat:
> > This requires a backwards-compatible change to Typed Racket before this can be committed, so TR is aware of the rename. The changes do not need to be simultaneous. The PR on the Typed Racket side is [racket/typed-racket#1499](https://github.com/racket/typed-racket/pull/1499) . The proper order for things is probably (1) get approval for both PRs, (2) merge the PR to Typed Racket and wait a couple hours for the package server to update, (3) merge this.
> 
> Friendly ping on step 1 for this PR: confirming that we still want to move forward with merging the low-level modules, and confirming that `core-syntax.rkt` is an acceptable filename?

@mflatt ?

jesboat:
> > This requires a backwards-compatible change to Typed Racket before this can be committed, so TR is aware of the rename. The changes do not need to be simultaneous. The PR on the Typed Racket side is [racket/typed-racket#1499](https://github.com/racket/typed-racket/pull/1499) . The proper order for things is probably (1) get approval for both PRs, (2) merge the PR to Typed Racket and wait a couple hours for the package server to update, (3) merge this.
> 
> Friendly ping on step 1 for this PR: confirming that we still want to move forward with merging the low-level modules, and confirming that `core-syntax.rkt` is an acceptable filename?

@mflatt ping?

mflatt:
Sorry for being the bottleneck here — I just kept not getting around to understanding the coordinated change that would be needed.

I'm fine with this change, and we should go ahead with it.

samth:
Given that, I'm going to merge racket/typed-racket#1499.

jesboat:
> I'm fine with this change, and we should go ahead with it.

@mflatt thanks! This should be ready to commit, now.

For actually folding other files into core-syntax.rkt, do you want one PR with separate commits for each source file, or separate PRs for each source file? Normally I go for small PRs, but I'm not sure that would actually offer value here.

> Given that, I'm going to merge [racket/typed-racket#1499](https://github.com/racket/typed-racket/pull/1499).

@samth thanks!

mflatt:
Will merge after my build completes (just in case).

I think one PR to merge the files makes the most sense.

PR #5464, opened 2026-03-07
duplicate ID handling

jesboat:
# PR Notes

This PR introduces two new functions in racket/private/stx,
`stx-find-duplicate-ids` and `raise-if-duplicate-identifiers`. Then,
throughout the codebase, places where duplicate identifier syntax errors
are raised are changed to use one of the two new functions. This causes
those syntax errors to highlight all copies of the duplicate identifier,
instead of just one.

It also centralizes two previous chunks of code which did duplicate ID
checking: the implementation of `check-duplicate-identifier` and an
implementation in the contract code. The implementation is written in
pure `#%kernel` in the intent that it will also soon replace a third
duplicate-identifier check in the implementation of `let`.

The new implementation uses straightforward loops instead of a `let/ec`
previously used in `check-duplicate-identifier`. I have not benchmarked
this, but would be quite surprised if the `let/ec` version wins.

Unlike the implementation in `let` which uses a list-based loop for
small input sizes, we use a hashtable for all input sizes (except for 0
and 1, for which there can be no duplicates). This is based on some
benchmarking done on lists of various sizes which have no duplicates:
https://gist.github.com/jesboat/a47f090dca2094b5cd36598ebb5e6406

There is an ad-hoc test suite which tests almost all the changes:
https://gist.github.com/jesboat/83e728cd39a818d9a977b7bb1bf3c9f9
It checks that duplicate identifiers are detected and that the
exn:fail:syntax contains the expected number of identifiers all of
which have distinct source locations. It is missing tests for three
changes:

- "duplicate identifier" in racket/collects/racket/private/unit/unit-infer.rkt
- 2nd "duplicate declared identifier" in racket/collects/racket/private/class-internal.rkt
- "duplicate identifier" in racket/collects/racket/private/class-internal.rkt

There is no fundamental reason those aren't tested; I'm just not
familiar enough with units and classes that writing a test which tickles
those particular places would have been straightforward.

# Commit log

## duplicate id handling: new helper functions

Introduce new helper functions:

`stx-find-duplicate-ids` is like check-duplicate-identifier, except it
returns all copies of the duplicated identifier, in a form which makes
it easy to pass them to `raise-syntax-error` and get a syntax error which
highlights all uses. (This commit does not introduce a public variant,
although it might be desirable to add one in the future.)

`raise-if-duplicate-identifiers` does exactly what it says on the tin.

These functions are tested in the rest of the PR series.

## duplicate id handling: use new helpers in various places

Use the new helpers in various places across the codebase, where
the changes are unobtrusive. There are ~29 uses total, of which this
PR changes 28.

An uncommitted ad-hoc test suite tests all but 3 (one in racket/unit, two in
racket/class). I don't know how to integrate it into the Racket test infra,
but it's at:
https://gist.github.com/jesboat/83e728cd39a818d9a977b7bb1bf3c9f9

## duplicate id handling: use new helpers in contract-in, contract-out

The change to use the new duplicate ID helper functions in the code for
`(require (contract-in ...))` and `(provide (contract-out ...))` is
not trivial, and is presented here as its own commit.

It is tested by the uncommitted test suite (along with most of the
changes in the previous commit) at:
https://gist.github.com/jesboat/83e728cd39a818d9a977b7bb1bf3c9f9

## duplicate id handling: replace check-duplicate-identifier implementation

Replace the old implementation (which was written using an, um, odd style
using an escape continuation) with a simple delegate to the new function.

Add some basic test cases.

### DIFF
--- a/pkgs/racket-test-core/tests/racket/stx.rktl
+++ b/pkgs/racket-test-core/tests/racket/stx.rktl
@@ -1,7 +1,7 @@
 
 (load-relative "loadtest.rktl")
 
-(require racket/syntax-srcloc)
+(require racket/syntax-srcloc racket/private/stx)
 
 (Section 'stx)
 
@@ -2919,6 +2919,17 @@
   (check '((require (for-label racket/base))) #f 0)
   (check '() 1 1)
   (check '() #f  #f))
+
+;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
+;; test some helper functions
+
+(test #f check-duplicate-identifier empty)
+(test #f check-duplicate-identifier (list #'a #'b))
+(test/compare bound-identifier=? #'a check-duplicate-identifier (list #'a #'a))
+(test/compare bound-identifier=? #'a check-duplicate-identifier (list #'a #'b #'a))
+(test/compare bound-identifier=? #'a check-duplicate-identifier (list #'b #'a #'a))
+(test #f check-duplicate-identifier (list #'a ((make-syntax-introducer) #'a)))
+
 
 ;; ;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;;
 
--- a/racket/collects/racket/private/stxcase-scheme.rkt
+++ b/racket/collects/racket/private/stxcase-scheme.rkt
@@ -11,19 +11,8 @@
   (-define (check-duplicate-identifier names)
     (unless (and (list? names) (andmap identifier? names))
       (raise-argument-error 'check-duplicate-identifier "(listof identifier?)" names))
-    (let/ec escape
-      (let ([ht (make-hasheq)])
-	(for-each
-	 (lambda (defined-name)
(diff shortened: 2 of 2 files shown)


rfindler:
I'm loving the better error messages that're going to come out of this! DrRacket's <menukey>-. keyboard shortcut is going to get a lot more useful!

jesboat:
Opportunity for discussion: which IDs (first occurrence of the duplicated identifier, second occurrence i.e. first duplicate, other occurrences) should end up where (`sub-expr` argument or `extra-sources` argument) in the raise-syntax-error calls.

jesboat:
friendly ping

rfindler:
No strong opinions from me! I think DrRacket sorts them in the buffer and goes through them in order anyway (not that that should factor into the thinking necc.)

jesboat:
If nobody has strong opinions, I'd propose to merge the PR as it is currently; I believe that results in the `exprs` field of the syntax error containing the duplicate IDs in sorted order, which seems fine.

PR #5467, opened 2026-03-12
Reject non-linear `match` patterns with mismatched ellipsis depth

### LINKED ISSUE #5442: Repeated identifier in nested list pattern: "unbound identifier;  also, no #%top syntax transformer is bound"
I was writing this function:
```racket
(define (remove-common-prefix xss)
  (match xss
    ((cons (cons x _)
           (list (cons ys _) ...))
     #:when (andmap (lambda (y) (equal? x y)) ys)
     (remove-common-prefix (map cdr xss)))
    (_ xss)))
(require rackunit)
(check-equal? (remove-common-prefix '((1 2 3 4) (1 2 5) (1 2 3 6)))
              '((3 4) (5) (3 6)))
```
And I wanted to write it like this:
```racket
(define (remove-common-prefix xss)
  (match xss
    ((cons (cons x _)
           (list (cons x _) ...))
     (remove-common-prefix (map cdr xss)))
    (_ xss)))
```
But I got this unfriendly error from Racket 9.0:
```
x28: unbound identifier;
 also, no #%top syntax transformer is bound in: x28
```
Whereas the code below succeeds at least:
```racket
(match '((4 5) 4 5)
  ((cons x
         (list x ...))
   x))
```

samth:
Non-linear pattern variables (same variable used multiple times) in match are now required to be at the same ellipsis depth. Patterns like `(cons x (list x ...))` or `(list a ... a)` where a variable appears both inside and outside of ... are syntax errors.

This requires re-implementing an optimization for patterns like `(list x ...)`, since it still have to track that `x` was used non-linearly.

This also repairs crashes/incorrect behavior for non-linear patterns across ... boundaries (fixes #5442, #2152, #1386).

Backwards compatibility: this may be incompatible in two ways:

1. It creates syntax errors for previously working programs. These programs have had a warning when compiled for the past decade.

2. It adds a check that the non-linear patterns actually match the same value. This may cause some matches to fail where previously they (incorrectly) succeeded.

I have run a full pkg-build which confirms that problem (1) does not seem to appear. I hope for testing between now and the next release to confirm whether problem (2) is significant.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/match.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/match.scrbl
@@ -126,6 +126,12 @@ In more detail, patterns match as follows:
        @racketidfont{not} sub-patterns are independent. The binding for @racket[_id] is
        not available in other parts of the same pattern.
 
+       If @racket[_id] is used multiple times at different ellipsis
+       depths---that is, some uses are under @racketidfont{...} and
+       others are not, or they are under different @racketidfont{...}
+       patterns---a syntax error is raised.
+       See @secref["match-nonlinear-ellipsis"] for details and examples.
+
        @examples[
        #:eval match-eval
        (match '(1 2 3)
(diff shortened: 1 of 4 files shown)


mflatt:
LGTM

PR #5485, opened 2026-04-09
CS: fix `struct-constructor-procedure?` on builtin struct types

jesboat:
Somebody familiar with this code should review this PR; the fix *seems* straightforward given surrounding code and other uses of `|#%struct-constructor|`, but I am unfamiliar with this layer of the implementation and don't know if there are might be unintended side effects.

# Commit message

Struct types declared in Rumble using the Rumble `struct` macro were constructing the constructors without passing them through `|#%struct-constructor|`, so `struct-constructor-procedure?` couldn't recognize them. Fix it.

Testing:

    Welcome to Racket v9.2.0.1-2026-04-08-cbc702d25f* [cs].
    > (struct-constructor-procedure? exn)
    #t
    >

### DIFF
--- a/pkgs/racket-test-core/tests/racket/struct.rktl
+++ b/pkgs/racket-test-core/tests/racket/struct.rktl
@@ -38,6 +38,8 @@
     (arity-test makex 0 0)
     (arity-test sel 2 2)
     (arity-test set 3 3)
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks!

PR #5493, opened 2026-04-18
correct minor error in some contract strings

jesboat:
`procedure-arity-includes?/c` -> `procedure-arity-includes/c`


mflatt:
Thanks!

PR #5500, opened 2026-04-28
Add configurable compression levels for zip and deflate

### LINKED ISSUE #4931: Support configurable (and matching) compression level in `zip` and `deflate`
The EPUB standard [requires](https://www.w3.org/TR/epub-33/#sec-zip-container-mime) that the `mimetype` file in an OCF zip container not be compressed. I've been using system commands to build these containers, but due to cross-platform issues (https://github.com/otherjoel/perfect-edition-pollen/issues/1) it would be nice to be able to build a valid zip container with Racket code.

Currently neither the [`zip`](https://docs.racket-lang.org/file/zip.html#%28def._%28%28lib._file%2Fzip..rkt%29._zip%29%29) function nor the [`deflate`](https://docs.racket-lang.org/file/gzip.html#%28def._%28%28lib._file%2Fgzip..rkt%29._deflate%29%29) function provide a way for callers to adjust the compression level from a hardcoded default, but the underlying code seems like it would be able to use any valid value (0–9) if one was provided.

Additionally, this hardcoded default appears to be mismatched between `deflate` (which [always uses `6`](https://github.com/racket/racket/blob/fa06ed5f44e97e29847a1d5f0ed2e51c45634cf5/racket/collects/file/gzip.rkt#L128)) and `zip` (which [always uses `8`](https://github.com/racket/racket/blob/fa06ed5f44e97e29847a1d5f0ed2e51c45634cf5/racket/collects/file/zip.rkt#L28) for the metadata). This might be harmless, or maybe I’m reading it wrong, but if it is incorrect it could be fixed as part of this issue.

If you prefer, I could try my hand at adding a PR with tests/docs to add an optional argument to both functions.

soegaard:
This PR was written by Codex.
It fixes issue #4931.

### Add configurable compression levels for `zip` and `deflate`

This PR addresses #4931 by adding configurable compression levels to `file/gzip` and `file/zip`.

### What changed

- Added `#:level` to `deflate`, accepting levels `0..9`
- Added `#:level` to `zip` and `zip->output`, also accepting `0..9`
- Preserved existing default behavior:
  - `deflate` still defaults to level `6`
  - `zip` still defaults to deflated ZIP entries

### ZIP behavior

This change also fixes a semantic mismatch in the current ZIP implementation.

Previously:
- `zip` recorded ZIP compression method `8` in entry metadata
- but the underlying compression behavior was hard-wired separately through `deflate`

Now:
- ZIP method is handled explicitly as ZIP method
- `#:level 0` produces stored ZIP entries (method `0`)
- `#:level 1..9` produces deflated ZIP entries (method `8`)

This makes the API usable for cases like EPUB containers, where some files must be uncompressed in the ZIP archive.

### `deflate` behavior

For `deflate`, `#:level 0` produces an uncompressed deflate stream, while `#:level 1..9` use the existing deflate compressor with the requested tuning.

### Tests

Added focused coverage for the new option:

- `file/gzip`: round-trip test for `deflate #:level 0`
- `file/zip`: verifies that `zip->output #:level 0` writes ZIP method `0`, while the default still writes method `8`

Also re-ran broader file/archive tests and rebuilt the documentation.

### Docs

Updated the `file/gzip` and `file/zip` documentation to describe the new `#:level` argument.

Fixes #4931

### DIFF
--- a/pkgs/racket-doc/file/scribblings/gzip.scrbl
+++ b/pkgs/racket-doc/file/scribblings/gzip.scrbl
@@ -38,7 +38,8 @@ report on Unix).}
 
 
 @defproc[(deflate [in input-port?]
-                  [out output-port?])
+                  [out output-port?]
+                  [#:level level exact-integer? 6])
          (values exact-nonnegative-integer?
                  exact-nonnegative-integer?
                  exact-nonnegative-integer?)]{
@@ -47,6 +48,10 @@ Writes @exec{pkzip}-format ``deflated'' data to the port @racket[out],
 compressing data from the port @racket[in].  The data in a file
 created by @exec{gzip} uses this format (preceded with header
 information).
+
+The @racket[level] argument controls compression from @racket[0] to
+@racket[9]. A level of @racket[0] produces an uncompressed
+@tt{deflate} stream, while higher levels trade speed for compression.
 
 The result is three values: the number of bytes read from @racket[in],
 the number of bytes written to @racket[out], and a cyclic redundancy
--- a/pkgs/racket-doc/file/scribblings/zip.scrbl
+++ b/pkgs/racket-doc/file/scribblings/zip.scrbl
@@ -8,7 +8,8 @@ utilities to create @exec{zip} archive files, which are compatible
 with both Windows and Unix (including Mac OS) unpacking. The actual
 compression is implemented by @racket[deflate].}
 
-@defproc[(zip [zip-file path-string?] [path path-string?] ...
+@defproc[(zip [zip-file path-string?] [path-or-entry (or/c path-string? zip-entry?)] ...
+              [#:level level exact-integer? 6]
               [#:timestamp timestamp (or/c #f exact-integer?) #f]
               [#:get-timestamp get-timestamp
                                (path? . -> . exact-integer?)
@@ -22,14 +23,22 @@ compression is implemented by @racket[deflate].}
          void?]{
 
 Creates @racket[zip-file], which holds the complete content of all
-@racket[path]s.
+@racket[path-or-entry]s. Each @racket[path-or-entry] is either a path
+that refers to an existing file or directory on the filesystem, or it
+is a @racket[zip-entry] that can override the compression level for
+that entry or describe an entry without needing corresponding content
+on the filesystem.
 
-The given @racket[path]s are all expected to be
+The given paths among @racket[path-or-entry]s are all expected to be
 relative path names of existing directories and files (i.e., relative
-to the current directory).  If a nested path is provided as a
-@racket[path], its ancestor directories are also added to the
-resulting @exec{zip} file, up to the current directory (using
-@racket[pathlist-closure]).
+to @racket[(current-directory)]). If a nested path is provided as a
+path or string, its ancestor directories are also added to the resulting
+@exec{zip} archive, up to the current directory (using
+@racket[pathlist-closure]). The @racket[path-or-entry]s will appear
+within the @exec{zip} archive in the same order in which they are provided.
+Settings contained in a @racket[zip-entry] will be used for that entry
+in the archive in preference to values provided as keyword arguments to
+this function.
 
 Files are packaged as usual for
 @exec{zip} files, including permission bits for both Windows and Unix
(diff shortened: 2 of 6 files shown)


otherjoel:
@soegaard thanks! not sure if you noted this aspect in a [later comment](https://github.com/racket/racket/issues/4931#issuecomment-1937798786):

> As I’m digging into it, I realize that being able to specify the deflate compression level doesn’t fully cover my immediate need. EPUB zip containers have to include the mimetype file at zero compression; that file also has to be the first one in the archive; and it is desirable for the rest of the files to be compressed normally. So what I really need, in addition to being able to set the compression level, is to be able to specify the order of the files in the zip container as well as separate compression levels for certain files.

So what's really needed is to be able to guarantee the order of files and to specify separate compression levels for each. I proposed this approach:

> add a zip-entry struct that consists of a pathname and a compression level, and an alternative function that works like zip but, instead of paths, accepts a list of zip-entry structs and provides guarantees about preserving the order of the entries in the zip archive.

soegaard:
Support for `zip-entry` (in style with `tar-entry`) was needed to support EPUB files.

The interface implemented is in the pdf:

[zip-with-zip-entry.pdf](https://github.com/user-attachments/files/27184219/zip-with-zip-entry.pdf)

soegaard:
Updated description of the PR after squashing additional commits.

### Add configurable ZIP and deflate compression support

Fixes #4931.

This change expands `file/zip` and `file/gzip` to support configurable compression in a way that also covers the EPUB-style use case that motivated the issue.

### What changed

- Added `#:level` to `deflate`, accepting values `0..9`
- Added `#:level` to `zip` and `zip->output`
- Preserved existing defaults:
  - `deflate` still defaults to level `6`
  - ZIP output still defaults to deflated entries

### ZIP behavior

This change separates ZIP entry method from deflate tuning.

- `#:level 0` creates stored ZIP entries (method `0`)
- `#:level 1..9` creates deflated ZIP entries (method `8`)

That makes it possible to create archives with uncompressed entries when required, instead of always writing deflated entries.

### `zip-entry` support

This PR also extends `zip-entry` so it is closer in spirit to `tar-entry`.

`zip` and `zip->output` now accept `zip-entry` values that can describe:

- an existing filesystem path
- a synthetic file entry backed by bytes, an input port, or a thunk producing an input port
- a synthetic directory entry

This means ZIP archives can now be constructed without first creating a matching directory tree on disk.

To keep archive paths well-formed, `zip-entry` paths are required to be relative, matching the `tar-entry` convention.

### Why this helps EPUB

For EPUB-like ZIP containers, callers can now:

- ensure specific entries appear first
- store entries without compression when required
- provide generated content directly, without temporary files

### Tests and docs

Updated docs for `file/gzip` and `file/zip`, and added regression coverage for:

- `deflate #:level 0`
- ZIP method `0` vs `8`
- ordered `zip-entry` usage
- synthetic ZIP entries without backing files
- thunk- and port-backed synthetic content
- metadata overrides
- size mismatch and malformed-entry error cases

Validation run:

- `racket/bin/raco setup file`
- `racket/bin/raco make racket/collects/file/zip.rkt pkgs/racket-test/tests/file/zip-round-trip.rkt`
- `racket/bin/racket pkgs/racket-test/tests/file/gzip.rkt`
- `racket/bin/racket pkgs/racket-test/tests/file/zip-round-trip.rkt`
- `racket/bin/racket pkgs/racket-test/tests/file/packers.rkt`
- `racket/bin/racket pkgs/racket-test/tests/file/test-docs-complete.rkt`

soegaard:
@otherjoel 

Did you get to a chance to skim [zip-with-zip-entry.pdf](https://github.com/user-attachments/files/27184219/zip-with-zip-entry.pdf) ?

I am interested in knowing whether this is enough for EPUBs?

otherjoel:
@soegaard It looks like the interface described in that PDF would be sufficient for generating EPUB-compliant ZIP archives. I'm not able to test it at this time, though.

For EPUB purposes, the only thing that seems to be missing from the document is explicit guarantees about preserving the order of the files inside the resulting archive. There is some language in there that gestures towards this, however. In particular, this paragraph under `zip` could use some edits to make some things clearer:

> The given paths among `path-or-entry`s are all expected to be relative path names of existing directories and files (i.e., relative to ~~the current directory~~ <mark>`current-directory`</mark>). If a nested path is provided as a ~~plain path~~ <mark>path-string</mark>, its ancestor directories are also added to the resulting zip file, up to the current directory (using `pathlist-closure`).  ~~A `zip-entry` is included as given, so it can be used to control the order of entries and their compression within the archive, including entries whose contents are provided as byte strings or input ports.~~ <mark>The `path-or-entry`s will appear within the zip archive in the same order in which they are provided. Settings contained in a `zip-entry` will be used for that entry in the archive in preference to values provided as keyword arguments to this function.</mark>

* Clarifies what is meant by "plain path" (I'm guessing what you meant here)
* Makes the guarantee about (and mechanism for) controlling order of entries more explicit
* Separate the usefulness of zip-entry from language about controlling order of entries, which applies to both zip-entry and path-string arguments

mflatt:
Thanks! I adjusted the docs as suggested, and also adjusted the `zip-entry` implementation to perform field contract checks early instead of late.

PR #5503, opened 2026-04-29
Normalize tar-entry string paths

### LINKED ISSUE #5502: `file/tar`: `tar->output` crashes when `tar-entry` uses a string path
Found by Codex

### `file/tar`: `tar->output` crashes when `tar-entry` uses a string path

`tar-entry` is documented to accept a `path-string?` path, but `tar->output` currently crashes if the `path` field is a string.

Reproducer:

```racket
#lang racket/base
(require file/tar)

(define b (open-output-bytes))
(parameterize ([current-output-port b])
  (tar->output
   (list (tar-entry 'directory "d" #f 0 #hash()))))
```

Current result:

```text
path->bytes: contract violation
  expected: path-for-some-system?
  given: "d"
```

The failure appears to come from `tar-one-entry` / `split-tar-name` here:

- [racket/collects/file/tar.rkt](/Users/soegaard/Dropbox/GitHub/racket/racket/collects/file/tar.rkt:52)

The documented contract is here:

- [pkgs/racket-doc/file/scribblings/tar.scrbl](/Users/soegaard/Dropbox/GitHub/racket/pkgs/racket-doc/file/scribblings/tar.scrbl:130)

So the docs say string paths are supported, but the implementation assumes a path object before calling `path->bytes`.

Expected behavior:

- either normalize string `tar-entry` paths to path values before processing
- or tighten the documented contract if only path values are intended

Given the docs, normalizing string paths seems like the right fix.

soegaard:
Written by Codex.

### Normalize string paths in `tar-entry`

Fixes #5502.

`file/tar` documents `tar-entry` paths as `path-string?`, but `tar->output` could crash when a `tar-entry` used a string path, because the implementation eventually passed that string to `path->bytes` without normalizing it first.

This PR fixes that by normalizing `tar-entry` paths before they are processed in `file/tar`.

### What changed

- normalize a `tar-entry` path when it is a string or byte string before using it as a path value
- do the same for `tar-entry` link targets, which had the same assumption

### Result

This now works as documented:

```racket
#lang racket/base
(require file/tar)

(define b (open-output-bytes))
(parameterize ([current-output-port b])
  (tar->output
   (list (tar-entry 'directory "d" #f 0 #hash()))))
```

### Tests

Added a focused regression in `pkgs/racket-test/tests/file/unpackers.rkt` that exercises `tar->output` with a `tar-entry` whose path is a string and checks that it round-trips correctly through `untar`.


soegaard:
Note: The failed check is unrelated to `tar`. It's a futures tests failing on BC.

[CI Win / buildtest-win (3m) (pull_request)](https://github.com/racket/racket/actions/runs/25104782690/job/73563089568?pr=5503)

mflatt:
Thanks! I updated slightly to make the normalization more directly about strings.

PR #5504, opened 2026-04-29
Allow tar-entry arguments in high-level tar APIs

### LINKED ISSUE #5501: `file/tar`: high-level `tar` does not actually accept `tar-entry` arguments
Found by Codex

### `file/tar`: high-level `tar` does not actually accept `tar-entry` arguments

The `file/tar` docs say that `tar` accepts `(or/c path-string? tar-entry?)` arguments, but the current implementation still passes all arguments through `pathlist-closure`, which only accepts real paths.

As a result, this documented usage fails:

```racket
#lang racket/base
(require file/tar racket/file)

(define p (make-temporary-file "~a.tar"))
(delete-file p)

(tar p
     (tar-entry 'directory (string->path "d") #f 0 #hash()))
```

Current result:

```text
simplify-path: contract violation
  expected: (or/c path-string? path-for-some-system?)
  given: #<tar-entry>
```

The problem appears to be here:

- [racket/collects/file/tar.rkt](/Users/soegaard/Dropbox/GitHub/racket/racket/collects/file/tar.rkt:264)

`tar` calls `pathlist-closure` on all arguments, including `tar-entry` values.

The docs that currently promise `tar-entry` support are here:

- [pkgs/racket-doc/file/scribblings/tar.scrbl](/Users/soegaard/Dropbox/GitHub/racket/pkgs/racket-doc/file/scribblings/tar.scrbl:17)

Expected behavior:

- `tar` should accept `tar-entry` arguments as documented, similar to `tar->output`
- plain paths can still go through `pathlist-closure`
- `tar-entry` values should be preserved and handled by the tar writer instead of being passed to `pathlist-closure`

soegaard:
Written by Codex.

### Allow `tar-entry` arguments in high-level tar APIs

Fixes #5501.

`file/tar` documents `tar` as accepting `(or/c path-string? tar-entry?)` arguments, but the high-level implementation was still passing all arguments through `pathlist-closure`, which only accepts real paths.

As a result, documented uses like passing a `tar-entry` directly to `tar` could fail before the tar writer ever saw the entry.

### What changed

- added a helper that closes over only plain path arguments
- preserves `tar-entry` values in place instead of sending them through `pathlist-closure`
- uses that helper in both:
  - `tar`
  - `tar-gzip`

### Result

High-level tar creation now accepts `tar-entry` arguments as documented, while still preserving the existing closure behavior for ordinary filesystem paths.

### Tests

Added focused regressions in `pkgs/racket-test/tests/file/unpackers.rkt` for:

- `tar` with a `tar-entry`
- `tar-gzip` with a `tar-entry`

These checks verify that the high-level APIs preserve and emit the entry correctly.


mflatt:
Thanks!

PR #5505, opened 2026-04-29
Avoid hanging in net/url-port test startup

soegaard:
Written by Codex.

**Summary**

This updates `tests/net/url-port` to avoid hanging indefinitely during server startup.

The test previously chose a random fixed port range and then waited for each server thread to reach `tcp-accept` before continuing. If `tcp-listen` failed during startup, the server thread could die before signaling readiness, leaving the main test thread blocked forever. That matches the CI timeout we saw.

**What changed**

- Use `tcp-listen 0` so the OS assigns an available port instead of guessing a random fixed port.
- Capture the actual listener ports during startup and use those ports for the client requests.
- Replace the implicit “ready once `accept` runs” coordination with an explicit startup channel.
- Add a timeout while waiting for startup, so a broken server setup fails the test instead of hanging.
- Update the redirect case to use the dynamically assigned second port.

**Why**

This makes the test more robust on CI, especially on systems where binding a chosen port can fail due to permissions, policy, or port conflicts.

**Validation**

- `raco make pkgs/net-test/tests/net/url-port.rkt`

I couldn’t fully run the socket-based test in the current sandbox because `tcp-listen` is blocked here, so the main runtime validation should come from CI.

### DIFF
--- a/pkgs/net-test/tests/net/url-port.rkt
+++ b/pkgs/net-test/tests/net/url-port.rkt
@@ -5,47 +5,64 @@
          tests/eli-tester)
 
 (define (run-tests scheme wrap-ports skip-actual-redirect?)
-  (define first-port-no (+ 10000 (random 50000)))
   (define ((make-tester url->port) response . responses)
     (define server-cust
       (make-custodian))
-    (define wait-sema
-      (make-semaphore))
-    (define my-accept
-      (λ (l)
-        (semaphore-post wait-sema)
-        (tcp-accept l)))
+    (define startup-ch
+      (make-channel))
+    (define listener-ports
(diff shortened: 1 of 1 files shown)


mfelleisen:
Isn't it possible to write a test with mock connections?

soegaard:
As I understand it, the library does not currently support mocking.

soegaard:
FWIW  This PR was motivated by this failing test:

https://github.com/racket/racket/actions/runs/25104782690/job/73563089539?pr=5503

It is still running.

image

soegaard:
https://github.com/racket/racket/actions/runs/25124550748/job/73633922740?pr=5506

This PR is stuck the same place.

PR #5499, opened 2026-04-28
Finish `core-syntax.rkt` merge, and 2 small related commits

jesboat:
This PR finishes (!) the giant series of PRs to merge files into core-syntax.rkt.  All the origin files by now use only kernel syntax (and helpers from the cross-phase-persistent `stx.rkt`), so we can just cut-and-paste the definitions from the origin files into core-syntax.rkt, and adjust the requires in other files. (And then adjust the requires in other files again. And again.) The full commit log follows. Some notes for review

1. Two non-rkt files need adjusting, too:

    pkgs/racket-doc/scribblings/reference/reference.scrbl
    racket/src/bc/src/makeexn

These seem straightforward enough. The doc search index still points `define-values` at the right place, which I think is sufficient to test the change to reference.scrbl, and the GH build of 3m should be sufficient to test the makeexn change.

2. In letstx-scheme.rkt, I removed the "rename `define` to `-define`" and "rename `define-syntax` to `-define-syntax`" imports in favor of just writing `define-syntax` in the body of the file. This avoids annoyances with `(#%provide (all-from _))`.

3. In each of the "absorb X into core-syntax.rkt" commits, adjusting requires was done with essentially `s/file.rkt/core-syntax.rkt` across the repo, manual review of the diff, and then confirmation with a search for `file.rkt` and `racket/private/file`. ("Essentially `s/file.rkt/core-syntax.rkt`" is glossing over care to avoid hitting unrelated files, e.g. `norm-define.rkt`)

4. I manually checked `all-from.*core-syntax.rkt`.

5. I also snuck in a commit to remove `small-scheme.rkt`, which appears unused in both 1p and 3p packages. It didn't seem worth a dedicated PR.

# Commit log

## eliminate some uses of the old `-define-syntax` macro

This is just `s/-define-syntax/define-syntax/` in a couple places in letstx-scheme.rkt. The motivation is to remove the import of `-define-syntax` which causes unpleasant behavior with `all-from` (or `all-from-except`) provides.

This is part of a larger effort to centralize and modernize some core syntax constructs; see #5406 for more on that.

## racket/private/core-syntax: absorb cond.rkt

Move cond.rkt's definitions (with no changes) into core-syntax.rkt, adding a section header. Remove cond.rkt. Adjust the requires of other files appropriately.

Note that this leads to some files requiring `core-syntax.rkt` more than once; this is tidied in a future commit in this PR.

This is part of a larger effort to centralize and modernize some core syntax constructs; see #5406 for more on that.

[...]

PR #5506, opened 2026-04-29
Fix unintentional provide of old-cond from racket/base

jesboat:
The recent merge of files into `core-syntax.rkt` (PR #5499) inadvertently started providing `old-cond` from `racket/base`. Specifically, f00c7f2c74c76df3626bfd86834b5ed17b15dcca changed pre-base.rkt from

    (#%require (all-except "define.rkt" define define-syntax define-for-syntax))
    (#%provide (all-from "define.rkt"))

to "core-syntax.rkt" without taking into consideration that core-syntax provided old-cond.

Fix by having core-syntax simply not provide old-cond (and provide `old-cond` to its one user through a submodule instead.)

### DIFF
--- a/racket/collects/racket/private/core-syntax.rkt
+++ b/racket/collects/racket/private/core-syntax.rkt
@@ -14,11 +14,14 @@
              define-values-for-syntax
              when unless
              call/ec let/ec
-             cond old-cond else =>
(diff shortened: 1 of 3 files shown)


mflatt:
Thanks!

PR #5507, opened 2026-04-30
Stabilize ephemeral listener startup in url-port test

soegaard:
Written by Codex.

**Summary**

This updates `tests/net/url-port` to make ephemeral listener startup more robust.

The current test already uses `tcp-listen 0` and reports startup failures instead of hanging indefinitely, but we still saw intermittent `Address already in use` failures during listener creation on macOS CI.

**Change**

- Keep the existing startup-hang fix structure.
- Continue to avoid socket reuse for ephemeral listeners.
- Add a small retry loop around the `tcp-listen 0` call used by the test’s custom listener startup path.

The retry only applies to transient `exn:fail:network` failures during listener creation, using a short backoff before retrying.

**Why**

Even when asking the OS to choose an ephemeral port, listener setup could still fail intermittently on macOS runners. A few retries make the test startup path more stable without changing the `net/url` behavior being tested.

**Validation**

- Local run:
  - `raco test pkgs/net-test/tests/net/url-port.rkt`
- GitHub Actions passed on:
  - `soegaard/racket`
  - branch `codex/fix-url-port-ephemeral-reuse`

### DIFF
--- a/pkgs/net-test/tests/net/url-port.rkt
+++ b/pkgs/net-test/tests/net/url-port.rkt
@@ -4,6 +4,18 @@
          openssl
          tests/eli-tester)
 
+(define (listen-on-ephemeral-port max-allow-wait hostname)
+  (let loop ([tries 8] [pause 0.01])
+    (with-handlers ([exn:fail:network?
+                     (λ (e)
+                       (cond
+                         [(zero? tries)
+                          (raise e)]
+                         [else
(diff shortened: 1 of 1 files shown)


mflatt:
After discussing this more, we went with moving retires into `tcp-listen` in the case of port 0. That seems to have addressed the problem in the url-port test and others.

PR #5511, opened 2026-05-11
Minor .gitignore changes

ndykman:
This is a small change to the ```.gitignore``` files. This includes ignoring any ```.vscode``` folder or Windows build artifacts anywhere in the project. Also, it ignores the ```pkg-src``` directory which one could use to place package cloned sources to insure they are not picked up by accident.  It also ignores ```Makefile``` where those files are to be created by running ```configure``` or a similar script.  

This should better prevent artifacts from VS Code and Windows development from being committed.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -112,7 +112,7 @@ CFLAGS_FOR_BUILD =
 
 # This branch name must be changed each time the pb boot files are
 # updated:
-PB_BRANCH = v10.4.0-pre-release.4-1
+PB_BRANCH = v10.5.0-pre-release.1-1
 PB_REPO = https://github.com/racket/pb
 
 # Set to empty for Git before v1.7.10:
--- a/pkgs/base/info.rkt
+++ b/pkgs/base/info.rkt
@@ -14,7 +14,7 @@
 
 ;; In the Racket source repo, this version should change exactly when
 ;; "racket_version.h" changes:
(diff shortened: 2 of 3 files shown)


mflatt:
Thanks for this patch!

The `.vscode` addition certainly makes sense.

I can also see excluding Windows build artifacts more generally. I'm unclear on how they would show up, but as a possible example, I vaguely remember ending up with some files in the output `lib` or executable directory when I turned on some compiler flag (probably related to debugging).

Excluding `Makefile` doesn't seem like it should be necessary, and there are enough `Makefile`s in the repo that it seems like it might be confusing. A `configure` script (or similar) should always be run in a `build` subdirectory or outside a checkout, and `Makefile` is just one of many files generated when configuring and building. Do you have a situation where using a `build` subdirectory doesn't work?

I'm not inherently opposed to adding `pkg-src`, but there's already `extra-pkgs`, which is what we have been using for that purpose, so maybe it would make sense to stick with that?

ndykman:
I can't remember how ```Makefile``` snuck in, but it was being picked up. Could have been a weird error. Also, for Windows artifacts, I usually do that to just avoid weird corner cases that can occur.  

I am fine with using ```extra-pkgs``` of course. If you want, free free to modify this however you'd like or if you need to update the pull request, let me know.

PR #5512, opened 2026-05-14
Add structured template strings

CutieDeng:
## Summary

- Add a new `racket-tstring` package that implements structured template strings, including parsing, expansion, rendering, and reader support.
- Add the `tstring` package as the public wrapper/alias and register both packages in the build configuration.
- Add `racket/interactive/tstring` support for REPL use.

## Tests

- `raco make pkgs/racket-tstring/tests/*.rkt pkgs/tstring/tests/*.rkt`
- `raco test pkgs/racket-tstring/tests pkgs/tstring/tests` (95 tests passed)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -129,7 +129,7 @@ EXTRA_REPOS_BASE =
 # to include in a distribution; if "{}" appears at the start,
 # then the content of "build/PKGS" can override when that file
 # exists:
-PKGS = {} main-distribution main-distribution-test
+PKGS = {} main-distribution main-distribution-test racket-tstring tstring
 
 # Needed for any distribution (not meant to be configured):
 REQUIRED_PKGS = racket-lib
--- a/pkgs/racket-tstring/info.rkt
+++ b/pkgs/racket-tstring/info.rkt
@@ -1,0 +1,23 @@
+#lang info
+
+(define collection "racket-tstring")
+
+(define deps
+  '("base")
+) ; end define deps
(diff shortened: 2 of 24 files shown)


CutieDeng (commented):
This feature branch introduces a non-trivial feature and still needs more testing and validation. Please treat this PR as ready for review and discussion, but do not merge it immediately.

usaoc:
I think there’s no reason these packages need to be in this repo.  (We already have `@`-notation for string interpolation, anyway.)  Maybe consider submitting them to the [Racket Package Index](https://pkgs.racket-lang.org/)?

CutieDeng:
> I think there’s no reason these packages need to be in this repo. (We already have `@`-notation for string interpolation, anyway.) Maybe consider submitting them to the [Racket Package Index](https://pkgs.racket-lang.org/)?

From the implementation side, bare `f"..."` / `t"..."` syntax and REPL support require reader-level integration, so this cannot be meaningfully delivered as an ordinary library after the reader has already parsed the source.

From the feature side, this follows Python’s established f-string model and its newer t-string/template-string direction (PEP 498, PEP 750), and explores whether first-tier Racket support can improve everyday REPL ergonomics and structured-string workflows.

samth:
It's true that it needs reader-level integration, but Racket's reader is extensible and #lang notation with meta-languages like `at-exp` makes it easy to add this to any Racket program.

PR #5515, opened 2026-05-18
Make `variable-reference->module-foo` functions work on pure linklet VRs

### LINKED ISSUE #5514: variable-reference->module-foo functions crash on pure-linklet variable-references
v9.2.0.4-2026-05-18-7498b0ab9b* [cs].

**What program did you run?**

```
#lang racket
(require racket/linklet)

(define l
  (compile-linklet
   '(linklet () (vr)
      (define-values (vr) (#%variable-reference)))))

(define inst (instantiate-linklet l '()))
(define vr (instance-variable-value inst 'vr))

(variable-reference->module-path-index vr)
(variable-reference->resolved-module-path vr)
(variable-reference->module-source vr)
```

Thanks to @soegaard for the repro

Not a crash

jesboat:
Previously, the implementation of

    variable-reference->namespace
    variable-reference->module-path-index
    variable-reference->resolved-module-path
    variable-reference->phase
    variable-reference->module-base-phase
    variable-reference->module-declaration-inspector

all assumed that, if a variable-reference was associated with a linklet instance, that that instance's data (in the sense of `racket/linklet`'s `instance-data`) was a namespace. When the expander is run as a normal program, this merely produces an exception when one of the namespace accessors tries to get a field of a non-namespace, but when the expander is built and used as the expander for a CS executable, it produces an invalid memory reference.

Instead, make all of the above functions raise an error when used on a variable-reference which hails from a linklet instance whose data is not a namespace. An alternative would be to specify placeholder results for some or all of the functions; this was discarded on the grounds that code using these functions is probably unprepared to deal with weird variable-references and surfacing the error early is probably better.

I was too lazy to write proper regression tests, but https://gist.github.com/jesboat/5beee827a6ebdaa3713f69c60bbccd2c exercise the functionality of a couple of the functions decently.

Closes #5514.


mflatt:
Thanks! I updated the patch to include the test.

PR #5524, opened 2026-06-01
Add missing time.h include

hannesbraun:
## Checklist

- [x] Bugfix

## Description of change

When packaging Racket 9.2 for Alpine Linux (which is using musl libc), the build failed on ppc64le, riscv64 and s390x due to `nanosleep` not being available. So I added the missing include of `time.h` to resolve this.


mflatt:
Thanks!

Since `nanosleep` is used only when `HAVE_POLL_SYSCALL`, I moved the `#include` under an existing `#ifdef HAVE_POLL_SYSCALL`, just to be conservative (although it seems unlikely to create any trouble as an unconditional `#include`).

PR #5530, opened 2026-06-09
Documentation Changes for rktio.h

ndykman:
## Checklist
- [ ] Bugfix
- [ ] Feature
- [ ] tests included
- [x] documentation

## Description of change

Changes the `rktio.h` file to use Doxygen style comments. Previous versions of this file had comments after the function or struct declarations which made reading the file very difficult. 

Also, there is a minor change to rktio_process.c that changes field names to be consistent with the rest of the library in terms of using `fd` versus `rfd` variables. 

There are zero changes to the public API. This is merely for those that are working with the library in the future.


mflatt:
@ndykman — I appreciate the time you've put into this!

I've tried to approach it with an open mind. I must say, though, that this change does not appeal to me. I'll offer my reasons, but please note that my conclusion below is not an outright "no".

To start, I wonder whether I'm missing some context. Are you running Doxygen to render documentation, or does your programming environment format/index information in a way similar to rendering? I'm unlikely to look at this information in any way other than as text, so that might explain our differing perceptions.

You write

> Previous versions of this file had comments after the function or struct declarations which made reading the file very difficult.

I don't agree with that conclusion. (Opinions will vary.) Pretty much all documentation I read — whether Racket documentation, man pages, or Doxygen-rendered pages — shows a definition or prototype followed by an explanation. I am familiar with a C convention of putting explanation comments before a declaration, and I started with that order for this file out of habit, but I was much happier after switching the order to match other (rendered) documentation.

Many things feel noisy or redundant to me. The `///` boilerplate and especially having to read/write `/// @param rktio The context....` every time is uncomfortable. Similarly, something like `/// @param rfd The rktio file descriptor` doesn't tell me anything that wasn't already in the prototype. Functions that were described all together like `rktio_read_in`, `rktio_write_in`, and `rktio_read_converted_in` get split out into multiple copies that repeat the same thing.

Where I used to see some clear sections that always start `/*************************************************/`, those hints are mostly (bit not always) gone. Sometimes there are sections, but in a less visually distinct format (to me, in my environment). Probably fixable.

Not really about the layout: You've rewritten a lot of descriptions in ways that I think are not correct or that seem less clear to me. (For example, `rktio_read_in` is not about a position in a file.) I imagine that the original text wasn't clear to you, so that's a place for improvement in any case. Another nit that could be fixed easily enough: the rewrites seem to use backquote less consistently around code names.

Left to my own devices, I would take your diff and try to improve comments based on the explanations that you changed, but leave the format basically intact. That would make sense if I continued to be the main producer and consumer of these comments — because I've been working on this code for a long time, I'm set in my ways, and I associate Doxygen with low-quality documentation.

I recognize, though, that you're inspired to make these changes because you're working on the rktio implementation. If you insist that this is needed, and especially if others find this format useful, I can go along. I think a lot of work will be necessary to preserve some details (things that I knew were important to tell myself, at least), but it can be done.

ndykman:
I understand this is a big change. I was tempted to just keep this in my fork and merge as needed (the file doesn't change much). I decided to give this a go in the spirit of making things easier for other contributors and I do think this is a critical change in that regard.

This was not a case of me rewriting things just because "DOxygen better". The old header was so frustrating to me that I that the time needed to rewrite the header was worth it.  It would also help me understand the library. This was me being stubborn as a C/C++/C# programmer.

I note that there are some duplicate comments and there were some editorial changes that need to be reviewed and I apologize that I made too many changes in that regard. I do think it is worth while to revisit the overall documentation. 

> To start, I wonder whether I'm missing some context. Are you running Doxygen to render documentation, or does your programming environment format/index information in a way similar to rendering? I'm unlikely to look at this information in any way other than as text, so that might explain our differing perceptions.

VSCode and other IDEs understand DOxygen style comments. If I hover over a function, it will show me the documentation:

VSCode

This is incredibly useful. Yes, it does mean you have much more boilerplate as you need to document each function. But, tools like VSCode make that easy. Believe me, I did not write all that boilerplate by hand. And yes, even I got exhausted and left out some parts in the documentation. 

I am not insisting that DOxygen be used. However, DOxygen hasn't been a bad sign in quite sometime. The Boost standard library (one of the largest and best regarded C++ library) uses it. 

For open source projects, I always prefer established conventions unless there is a really good reason to do otherwise. I don't think a personal style preference (albeit an understandable one) is good enough in this case. I also note that the C code in ChezScheme uses the standard C convention for comments as well.

mflatt:
Thanks for the explanation and screen shot!

With that in mind, and after learning more about how IDEs treat Doxygen-format comments, I wonder how you would feel about an intermediate variant along these lines:

https://gist.github.com/mflatt/29e21f9c5554a3861d4317aecc0d9cf5

Like your version, this moves descriptions to before prototypes, and it enables the description to show up in VSCode. It doesn't show a breakout of parameters and results (leaving that to the description text, where not obvious), but it's easier for me to read and maintain without IDE support.

ndykman:
> Thanks for the explanation and screen shot!
> 
> With that in mind, and after learning more about how IDEs treat Doxygen-format comments, I wonder how you would feel about an intermediate variant along these lines:
> 
> https://gist.github.com/mflatt/29e21f9c5554a3861d4317aecc0d9cf5

I have zero issues with that. Again, I apologize that DOxygen was a bit of a red herring. Using VS Code, it was second nature. 

> 
> Like your version, this moves descriptions to before prototypes, and it enables the description to show up in VSCode. It doesn't show a breakout of parameters and results (leaving that to the description text, where not obvious), but it's easier for me to read and maintain without IDE support.

Agreed. 

Now, do you want me to de-oxygenize the comments? Or you would want to work on it to make sure it meets your needs?

As commentary for anybody reading, it's not my intent to change `rktio` just because. This is all motivated by native thread support in Racket and what that enables in terms of modern I/O. This isn't just about performance. It is also about matching how many production languages approach I/O and being able to more effectively educate and show students the pitfalls and advantages of these newer approaches.

ndykman:
I decided to remove the DOxygen markup myself as I did start all of this after all. Also, `port.c` should work now with the small name changes I made to a struct.

mflatt:
Thanks! This is generally good, but we'll need to either convert remaining `//` to `/* ... */` or update `parse.rkt` to handle `//` comments. Ok if I make those repairs and then force-push a simplified commit sequence here? Then, you can double-check before we megre.

ndykman:
> Thanks! This is generally good, but we'll need to either convert remaining `//` to `/* ... */` or update `parse.rkt` to handle `//` comments. Ok if I make those repairs and then force-push a simplified commit sequence here? Then, you can double-check before we megre.

I just added a change that removed all the `//` style comments. Go ahead and double check everything and if you are happy, just merge and squash if needed. If I overlapped your work on this, I apologize, but I thought the workflow would be easier overall.

PR #5538, opened 2026-06-26
add `ffi/unsafe/runtime-lib`

mflatt:
Like `define-runtime-path` (which it builds on), `define-runtime-lib` both binds a name to a foreign library and declares relevant libraries at compile time to be picked up by `raco exe` and `raco dist`.

The `define-runtime-lib` form is based on the same-named form that has long been exported by `racket/draw/private/utils/libs`, but modernized, generalized, and made public. The intent is that `racket/draw/private/utils/libs` will change to use this implementation (while also supporting its legacy syntax).

### DIFF
--- a/pkgs/racket-doc/scribblings/foreign/derived.scrbl
+++ b/pkgs/racket-doc/scribblings/foreign/derived.scrbl
@@ -26,4 +26,4 @@
 @include-section["file.scrbl"]
 @include-section["winapi.scrbl"]
 @include-section["vm.scrbl"]
-
+@include-section["runtime-lib.scrbl"]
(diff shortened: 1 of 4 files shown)


samth:
How is this intended to interact with `ffi2`?

mflatt:
`ffi-lib` and `ffi2-lib` are compatible, so `define-runtime-lib` can be used with `ffi2`.

But that does suggest that relying on `ffi-lib` syntactically is a bad idea (i.e., not a good one to carry over from `racket/draw/private/utils/libs`), so I'll revisit that.

samth:
I also wonder if having the path in `require` be `ffi/unsafe` will look weird when everyone is using `ffi2`.

mflatt:
There are several `ffi/unsafe` libraries that remain relevant for `ffi2` (as noted in https://docs.racket-lang.org/ffi2/interop.html), so I think `ffi2` + `ffi/unsafe` mixtures will be common.

Issue #4804, opened 2023-10-20
[Feature Request] better temporary file names for `call-with-atomic-output-file`


LiberalArtist:
**Is your feature request related to a problem? Please describe.**
It would be nice for `call-with-atomic-output-file` to use temporary files that would be more helpful for debugging, perhaps by incorporating the name of the destination file.

As a concrete example, when chasing down the bug discussed in https://github.com/racket/racket/pull/4802, this is the error message I was starting from:
```
raco setup: --- summary of errors ---                              [4:39:37]
raco setup: error: during building docs for <pkgs>/racket-index/scribblings/main/release.scrbl
raco setup:   open-output-file: error opening file
raco setup:     path: /gnu/store/ksbg66yvwfkyf999kdlglbjkv7iijadq-racket-8.10/share/doc/racket/release/COLLECTS-racket-file.rkt-586-20_16977767711697776771794
raco setup:     system error: Permission denied; errno=13
raco setup:     context...:
raco setup:      /gnu/store/s3xkylhbk8h2cmmk7s1d88ihixjdkdm2-racket-vm-cs-8.10/opt/racket-vm/collects/racket/file.rkt:203:4
raco setup:      /gnu/store/s3xkylhbk8h2cmmk7s1d88ihixjdkdm2-racket-vm-cs-8.10/opt/racket-vm/collects/racket/file.rkt:564:0: call-with-atomic-output-file
raco setup:      /gnu/store/ksbg66yvwfkyf999kdlglbjkv7iijadq-racket-8.10/lib/racket/pkgs/racket-index/setup/scribble.rkt:1577:0: write-
raco setup:      /gnu/store/ksbg66yvwfkyf999kdlglbjkv7iijadq-racket-8.10/lib/racket/pkgs/racket-index/setup/scribble.rkt:1415:0: do-render-time
raco setup:      /gnu/store/ksbg66yvwfkyf999kdlglbjkv7iijadq-racket-8.10/lib/racket/pkgs/racket-index/setup/scribble.rkt:1472:3
raco setup:      .../setup/parallel-do.rkt:389:17
```

Fortunately, `racket-index` had logging I could enable: otherwise, it would have been far from obvious that, when opening `release/COLLECTS-racket-file.rkt-586-20_16977767711697776771794`, it was really trying to write to `release/in.sxref`.

**Describe the solution you'd like**
I think incorporating the name of the destination file in the name of the temporary file would be helpful for debugging. In my example above, maybe the name could have been something like:
```
[...]

PR #5543, opened 2026-07-07
Add better temporary filename for call-with-atomic-output-file

arshakir:
## Checklist

- [ ] Bugfix
- [x] Feature
- [ ] tests included
- [ ] documentation

## Description of change

Solves #4804 
This version switches to `make-temporary-file*` to include a prefix and suffix to help with debugging atomic files. The atomic file name is now generated in the format `<outfile>-<random>-atomic-tmp`

No tests added since I didn't see any for `call-with-atomic-output-file` but I would be open to adding some.

This is my first contribution to Racket, so I'd appreciate any pointers on style or process I might have missed.


LiberalArtist:
I've only taken a quick look, but this seems like just what I had in mind in #4804, and I'm especially glad you tried to tackle it as a first issue. Thanks for contributing!

LiberalArtist:
P.S. The tests for `call-with-atomic-output-file` live here: https://github.com/racket/racket/blob/d32f1545a199f522573c71da3cde0aa89ef8fb6e/pkgs/racket-test-core/tests/racket/filelib.rktl#L217-L255

Adding a test would certainly be ideal!

arshakir:
@LiberalArtist thanks for taking a look. I just added a test to confirm that the temporary filenames format matches `<outfile>-<random>-atomic-tmp`. Let me know what you think

mflatt:
Thanks!

PR #5544, opened 2026-07-07
WIP: Add `keyword-prefixed-field`

jackfirth:
## Checklist

- [x] Feature
- [ ] tests included
- [x] documentation

## Description of change

When using `make-constructor-style-printer`, the current guidance for handling fields that are passed to the struct's constructor by name instead of by position is to use `unquoted-printing-string` to insert quoteless keywords into the printed representation. This has downsides when the struct is too long to print on a single line with the pretty printer. For example, printing `(foo #:a 1 #:b 2)` over multiple lines would print out thusly:

```scheme
(foo
 #:a
 1
 #:b
 2)
```

This is unreadable. Far better is to keep each keyword and argument on the same line unless *that specific field* is the one that's long enough to force the line break. And in that instance, it would be better to indent the field value slightly to better communicate via indentation that it's associated with the keyword on the line above it. Supposing that the `#:a` field in the example above was the one that was excessively long, the above would be more readable like this:

```scheme
(foo
 #:a
   1
 #:b 2)
```

This commit adds a `keyword-prefixed-field` struct wrapping a keyword and an arbitrary value which, when printed, implements the above behavior for a single keyword field. By combining this with `make-constructor-style-printer`, we can achieve the desired pretty printing behavior.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/struct.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/struct.scrbl
@@ -772,7 +772,7 @@ struct instance from being considered quotable. For example:
 (print (list (point2 1 2) (point2 3 4)))
 ]
 
-Keyword arguments can be simulated with @racket[unquoted-printing-string]:
+Keyword arguments can be printed with @racket[keyword-prefixed-field]:
 
 @examples[#:eval struct-eval #:label #f
 (code:comment "Private implementation")
@@ -782,10 +782,8 @@ Keyword arguments can be simulated with @racket[unquoted-printing-string]:
      (make-constructor-style-printer
       (lambda (obj) 'kwpoint)
       (lambda (obj)
-        (list (unquoted-printing-string "#:x")
-              (kwpoint-impl-x obj)
-              (unquoted-printing-string "#:y")
-              (kwpoint-impl-y obj)))))])
+        (list (keyword-prefixed-field '#:x (kwpoint-impl-x obj))
+              (keyword-prefixed-field '#:y (kwpoint-impl-y obj))))))])
 (code:comment "Public ``constructor''")
 (define (kwpoint #:x x #:y y)
   (kwpoint-impl x y))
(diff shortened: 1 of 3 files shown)


LiberalArtist:
I like this idea!

I wonder if there's a slightly more general version&mdash;but I wouldn't want to worsen the ergonomics for this use-case.

For example, I thought of dictionary types, which similarly suffer from their key&ndash;value relationships not being communicated to the pretty printer, but where the keys need not be keywords. A complication is that, for `keyword-prefixed-field`, it makes sense to ignore `write`/`print`/`display` mode for the key part, but that would not be true for a dictionary type.

I also wondered about a generalization to more than two components to try to keep together, but that seemed like a point of diminishing usefulness.

Maybe the answer is that the differences would be big enough that they shouldn't conflated with `keyword-prefixed-field`, especially since this use case has been prominent enough to document (for nearly 9 years! I hadn't noticed https://github.com/racket/racket/commit/5519a8ab458521177a4e3398125280718adb86b7 until seeing this PR).

LiberalArtist:
One possible answer (don&rsquo;t know if it&rsquo;s a good one or not) could be to keep `keyword-prefixed-field` as is, and separately add a `(struct pretty-print-together (lst))` that prints like a list, but with no parentheses.

sorawee:
Should `fmt` to be adjusted accordingly? I think it will cause a lot of changes though. Do you want to post to Discourse first to see whether Racket users have any feedback on this "indentation" style?

jackfirth:
I don't think `fmt` needs to change currently. I don't actually personally think that whitespace formatting for values printed to REPLs needs to exactly match whitespace styles for the expressions that *construct* those values. I went with the indentation style I did because I find keyword struct REPL output far more difficult to read with less indentation.

jackfirth:
@LiberalArtist Regarding generalization, I think there's some options we could explore. But I'd like to handle them separately, in future PRs. Right now all I personally need is this specific API, and I think it's useful and ergonomic enough on its own that we'd still want it as-is even if we had a more general solution.

PR #5547, opened 2026-07-25
Bugfix for `filesystem-root-list` on Windows

ndykman:
## Checklist
- [X] Bugfix

## Description of change

Fixes a bug in which `filesystem-root-list` failed (invalid memory reference) on windows systems with a certain amount of drive letters in use. Looking at the code, it seems this was triggered if more than eight drives were mapped. 

The older version used an API to get drive stats to check if the drive could be accessed by the current user. This doesn't work as the old code expects as a user may have access to folders within the drive without access to the root. Also, one can make a file within the root and technically not stat the drive; this is very rare of course.

As the stat call is expensive and doesn't actually work as expected, it was removed in this code. Also, this behavior is **not** part of the public documentation. Given that, I don't consider this a breaking change.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/filesystem.scrbl
@@ -802,8 +802,11 @@ The resulting paths are always sorted using @racket[path<?].}
 
 @defproc[(filesystem-root-list) (listof path?)]{
 
-Returns a list of all current root directories. Obtaining this list
-can be particularly slow on Windows.}
+Returns a list of all current root directories.
+
+@history[#:changed "9.3.0.2" @elem{Removed filtering of Windows root directories that are
+                                   currently inaccessible, which was intended to
+                                   avoid removable-storage devices that are empty.}]}
 
 @;------------------------------------------------------------------------
 @section[#:tag "filesystem-change"]{Detecting Filesystem Changes}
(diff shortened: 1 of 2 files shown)


mflatt:
Thanks for the repair!

On the filtering by `GetDiskFreeSpaceA`: good point about permissions on a drive's root. In case it's not obvious, the reason this filter existed is due to removable drives. In the old days, there floppy drives and CD drives that would throw up a modal dialog on any attempt to access the drive. Checking them with `SetErrorMode(SEM_FAILCRITICALERRORS)` enabled filtering without creating a modal dialog, so consumers of `(filesystem-root-list)` could more easily avoid dialogs. That aspect of `filesystem-root-list` should have been documented. But devices that provoke that dialog have become rare, as far I know, so it seems likely that we can do without it.

We'll want a `history` note in the docs to report the filtering change, but I can add that.

ndykman:
> Thanks for the repair!
> 
> On the filtering by `GetDiskFreeSpaceA`: good point about permissions on a drive's root. In case it's not obvious, the reason this filter existed is due to removable drives. In the old days, there floppy drives and CD drives that would throw up a modal dialog on any attempt to access the drive. Checking them with `SetErrorMode(SEM_FAILCRITICALERRORS)` enabled filtering without creating a modal dialog, so consumers of `(filesystem-root-list)` could more easily avoid dialogs. That aspect of `filesystem-root-list` should have been documented. But devices that provoke that dialog have become rare, as far I know, so it seems likely that we can do without it.
> 
> We'll want a `history` note in the docs to report the filtering change, but I can add that.

I figured it had to do with removable devices, but I don't think the modal prompt behavior even exists anymore even with devices that used to prompt it.  So, I think this is okay. If needs be, there is a fix that uses `GetDriveTypeA` to check for removable drives (optical and floppy) that would be better.

PR #5555, opened 2026-08-07
Correct barcode module filename in Guide example

slbls:
## Checklist

- [x] documentation

## Description of change

The [Organizing Modules example](https://docs.racket-lang.org/guide/module-basics.html#%28part._module-org%29) names the helper module `db/barcodes.rkt` in its prose and diagram, but the adjacent `db/lookup.rkt` example requires `barcode.rkt`.

Change the example to require `barcodes.rkt` so the references agree.


mflatt:
Thanks!

PR #5558, opened 2026-08-13
Chez Scheme: fuse call-with-values through non-atomic values args

### LINKED ISSUE #4634: [Feature Request] call-with-values reduction in CP0
It would be nice to be able to optimize `call-with-values` with non-trivial producer.

Consider:

```
(define (test x)
  (define-values (a b) (values (add1 x) (sub1 x)))
  (+ a b))
```

This is CP0'ed to:

```
  (letrec ([test.8 (lambda (x_1.9)
                     (#2%call-with-values
                       (lambda ()
                         (let ([app_4.10 (#2%add1 x_1.9)])
                           (#2%values app_4.10 (#3%sub1 x_1.9))))
                       (case-lambda
                         [(a_2.11 b_3.12) (#2%+ a_2.11 b_3.12)]
                         [args.13
                          ($app/no-return
                            2
                            args.13)])))])
    ...)
```

Ideally, CP0 would know that the producer and consumer of `call-with-values` can be fused together.

Interestingly, it can do the fusion in some cases, and I can't really tell what exactly inhibit the optimization. Perhaps when arguments to `values` are not both atomic?

```
(define (test2 x)
  (define-values (a b) (values (add1 x) x))
  (+ a b))
```

CP0'ed to:

```
  (letrec ([test2.8 (lambda (x_1.9)
                     (#2%+ (#2%add1 x_1.9) x_1.9))])
    ...)
```

But this leads to some unexpectedly (un)successful optimization.

```
(define (test x y)
  (define-values (a b) (values x (sub1 y)))
  (+ a b))

(define (test* x)
  (test (add1 x) x))
```

is successfully reduced to:

```
  (letrec ([test*.9 (lambda (x_5.10)
                      (let ([x_1.11 (#2%add1 x_5.10)])
                        (#3%+ x_1.11 (#3%sub1 x_5.10))))]
           [test.12 (lambda (x_1.13 y_2.14)
                      (#2%+ x_1.13 (#2%sub1 y_2.14)))])
    ...)
```

V1C70RYG0D:
## Checklist

- [x] Bugfix
- [ ] Feature
- [x] tests included
- [ ] documentation

## Description of change

Fixes #4634.

cp0's `call-with-values` inlining only recognized a producer whose body,
after stripping a `seq` wrapper, was directly a `(values e ...)` call. When
a `values` argument is non-atomic (e.g. `(add1 x)`), cp0 must bind it to a
temp first to fix evaluation order before constructing the tuple. That
binding is represented not as `letrec`, but as an immediate application of
a single-clause `case-lambda` (`let`'s actual Lsrc shape), which the
existing `result-exp`-based check didn't see through, so the fusion
silently fell back to a residual `call-with-values` call:

```racket
(define (test x)
  (define-values (a b) (values (add1 x) (sub1 x)))
  (+ a b))
```

## Fix

Added `find-values-in-tail`/`rewrap-tail` to `s/cp0.ss`, generalizing
`result-exp`/`non-result-exp` to also look through `seq`,
`letrec`/`letrec*`, and that case-lambda-application shape, recursively
(so multiple non-atomic args, needing nested bindings, are handled too).
These are new, narrowly-scoped helpers rather than a change to the
existing global `result-exp`/`non-result-exp`, which are used in 30+
other places in this file — this keeps the change confined to the one
optimization being extended.

The fix never strips or reorders the sequencing bindings, only relocates
what they feed into (from the `values` call to the fused consumer call),
so the evaluation-order guarantee they exist for is preserved.

## Testing

- Added 3 regression tests to `mats/cp0.ms`'s `call-with-values` section:
  the issue's reported 2-value case, its already-working sibling case
  (confirming no regression), and a 3-value case with two non-atomic args
  requiring nested bindings.
- Ran the full `mats/cp0.ms` suite plus `6.ms`, `5_3.ms`, and `cptypes.ms`
  via a standalone `pb` Chez Scheme build: zero Bugs/Errors.
- Verified via `PLT_LINKLET_SHOW_CP0` on compiled modules that
  `call-with-values` is fully eliminated and runtime results are correct
  for the 2-value, already-working, and 3-value cases.

### DIFF
--- a/racket/src/ChezScheme/mats/cp0.ms
+++ b/racket/src/ChezScheme/mats/cp0.ms
@@ -382,6 +382,31 @@
           (lambda () (div-and-mod 7 3))
           (lambda (x y) (#2%cons (* x 10) (/ y 10)))))
      '(#2%cons 20 1/10))
+  ; verify call-with-values still fuses producer and consumer when a
+  ; non-atomic `values` argument forces cp0 to bind it to a temp before
+  ; constructing the tuple, instead of leaving a residual call-with-values
+  ; (github issue #4634)
+   (equivalent-expansion?
+     (expand/optimize
+       '(lambda (x)
+          (call-with-values
+            (lambda () (values (add1 x) (sub1 x)))
+            (lambda (a b) (+ a b)))))
+     '(lambda (x) (#3%+ (#2%add1 x) (#2%sub1 x))))
+   (equivalent-expansion?
+     (expand/optimize
+       '(lambda (x)
+          (call-with-values
+            (lambda () (values (add1 x) x))
+            (lambda (a b) (+ a b)))))
+     '(lambda (x) (#2%+ (#2%add1 x) x)))
+   (equivalent-expansion?
+     (expand/optimize
+       '(lambda (x)
+          (call-with-values
+            (lambda () (values (add1 x) (sub1 x) (* x 2)))
+            (lambda (a b c) (+ a b c)))))
+     '(lambda (x) (#3%+ (#2%add1 x) (#2%sub1 x) (#2%* 2 x))))
 
   (equivalent-expansion?
    (expand/optimize '(lambda (t) (#3%$value (if t 1 (values 3 3 3))) #t))
--- a/racket/src/ChezScheme/s/cp0.ss
+++ b/racket/src/ChezScheme/s/cp0.ss
@@ -961,6 +961,37 @@
           [(seq ,e1 ,e2) `(seq ,e1 ,body)]
           [else body])))
 
+    ; Like `result-exp`, but also looks through the wrappers cp0 introduces
+    ; to sequence evaluation of non-atomic subexpressions before combining
+    ; them, e.g. the `(add1 x)` in `(values (add1 x) (sub1 x))` gets bound
+    ; to a temp first. cp0 represents that binding not as `letrec`, but as
+    ; an immediate call to a single-clause `case-lambda` (i.e., `let`'s
+    ; actual internal Lsrc shape). Used to recognize a producer that
+    ; tail-calls `values` even after that sequencing.
+    (define (find-values-in-tail e)
+      (nanopass-case (Lsrc Expr) e
+        [(seq ,e1 ,e2) (find-values-in-tail e2)]
+        [(letrec ([,x* ,e*] ...) ,body) (find-values-in-tail body)]
+        [(letrec* ([,x* ,e*] ...) ,body) (find-values-in-tail body)]
+        [(call ,preinfo (case-lambda ,preinfo2 (clause (,x* ...) ,interface ,body)) ,e* ...)
+         (find-values-in-tail body)]
+        [(call ,preinfo ,pr ,e* ...)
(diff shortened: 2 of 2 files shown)


samdphillips:
This probably should be opened against the [cisco/ChezScheme](https://github.com/cisco/ChezScheme) repo.

mflatt:
I agree that a Chez Scheme change would need to go through `cisco/ChezScheme`, but things to consider with this PR:

* In looking at the cp0 output, you might think that a closure will be allocated when the example runs, since `call-with-values` is still in the cp0 output. But backend optimizations will compile this a multiple-value "let", after all.

* Then again, even though the backend turns out to avoid `call-with-values`, it's possible that other optimizations (at either the cp0 or earlier in the backend passes) will be missed because `call-with-values` sticks around through the cp0 phase. So, this kind of change is potentially useful.

* Then again again, within Racket, there's an even earlier set of optimizations at the schemify level. Inlining and conversion from `call-with-values` to `let-values` are particularly useful at that level, and schemify is also the level that is introduces a `let` for `(add1 x)`. Adding a conversion in schemify will then turn out to make a cp0 optimization less useful for Racket programs. So, a schemify change (to this repo) is probably more useful overall.

* In the `cp0.ss` change, `find-values-in-tail` can traverse a body arbitrarily deeply. Given the way `find-values-in-tail` is triggered by a pattern that it also traverses through, an N-deep nesting of encoded `let`s can take quadratic time, right? Whether in cp0 or schemify, an optimization pass will need to avoid going quadratic. (@gus-massa seems to have this detail in mind in #4634).

* The tests added here to `cp0.ms` pass without the change to `cp0.ss`. That is, the new tests do not really check the change to `cp0.ss`. Tests could need to use `let` more to show the effect.

I recommend working toward a schemify change instead of a cp0 change, although both are potentially useful.

V1C70RYG0D:
You were right on every point, and I want to follow up honestly on what I found digging in.

**The test gap**: confirmed. I rebuilt with the `cp0.ss` change reverted but the `mats/cp0.ms` additions kept, and the tests passed either way. Root cause: the raw `expand/optimize` harness that file uses compiles a plain, single-clause `lambda` consumer, not the arity-checking `case-lambda` that `define-values` actually produces via schemify. In that simpler shape, cp0's existing inlining already fully resolves the producer/consumer on its own, with or without my change. The bug only shows up once schemify's `case-lambda` wrapper is in the picture, which that harness doesn't reproduce.

**The quadratic-time concern**: also fair — `find-values-in-tail` walks the same structure it's triggered by, so nested `call-with-values` chains could compound.

**The layer**: you were right that this belongs in schemify, not cp0. I traced it to the actual introduction point — `make-let-values` in `racket/src/schemify/left-to-right.rkt` always wraps a multi-value binding in `call-with-values` + an arity-checking `case-lambda`, even when the right-hand side is already statically known to be a matching `values` call. Fixed it there instead: #5563. Verified via `schemify-demo.rkt`, a full from-source rebuild with `PLT_LINKLET_SHOW_CP0`, and the real `optimize.rktl` suite (1389 tests, including two pre-existing tests whose expectations needed updating because of the fix).

Closing this one in favor of #5563.

PR #5559, opened 2026-08-13
syntax/parse: track `'disappeared-use` in syntax-class definitions

usaoc:
## Checklist
- [x] Bugfix

## Description of change
Currently, `define-syntax-class`, `define-splicing-syntax-class`, and `define-eh-alternative-set` are missing a `with-disappeared-uses`. Add this, so that the recorded keywords (including `pattern` and syntax-pattern keywords) are attached as `'disappeared-use` to the expansion.

Related to the `define-eh-alternative-set` change, record `pattern` in each eh-alt, and record the referenced eh-alt-set id in `~eh-var`.

An additional commit in this PR also cleans up some `syntax-case` fenders in `"rep.rkt"` that I noticed along the way.

### DIFF
--- a/racket/collects/syntax/parse/private/parse.rkt
+++ b/racket/collects/syntax/parse/private/parse.rkt
@@ -61,30 +61,31 @@
  (define (tx:define-*-syntax-class stx splicing?)
    (syntax-case stx ()
      [(_ header . rhss)
-      (parameterize ((current-syntax-context stx))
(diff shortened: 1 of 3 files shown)


mflatt:
This looks good as far as I can tell. @rmculpepper?

rmculpepper:
Merged, thanks!

PR #5560, opened 2026-08-14
Enable Racket to build and run on Haiku OS x86_64.

tealeg:
## Checklist

- [ ] Bugfix
- [x] Feature
- [ ] tests included
- [ ] documentation

## Description of change

As my changes to Chez, to support Haiku OS on x86_64, have now been merged into the Racket codebase, I was able to make these very simple changes to support building Racket on that platform too.  I've been running this successfully since the 26th of July and I'm able to work happily in DrRacket on the platform, use raco, etc. etc.   

Building on Haiku should be as simple as `make in-place`, so long as you have installed the correct dependencies (I'll go about making a proper package for Haiku when and if support is "official").  

However, I'd appreciate some pointers on how to run exhaustive tests for Haiku.  I've poked around a bit but didn't find my way in.

### DIFF
--- a/racket/src/cs/c/configure
+++ b/racket/src/cs/c/configure
@@ -4801,6 +4801,14 @@ case "$host_os" in
     MACH_OS=qnx
     use_flag_pthread=no
     ;;
+  haiku*)
+    MACH_OS=hk
+    CFLAGS="${CFLAGS} -I/boot/system/develop/headers -I/boot/system/develop/headers/posix"
+    LDFLAGS="${LDFLAGS} -L/boot/system/develop/lib"
+    LIBS="${LIBS} -lm -lpthread -lroot -lnetwork -lposix_error_mapper"
+    LINK_DYNAMIC="-rdynamic"
+    add_iconv_lib="-liconv"
+    ;;
   *)
     ;;
 esac
--- a/racket/src/cs/c/configure.ac
+++ b/racket/src/cs/c/configure.ac
@@ -376,6 +376,14 @@ case "$host_os" in
     MACH_OS=qnx
     use_flag_pthread=no
     ;;
+  haiku*)
+    MACH_OS=hk
+    CFLAGS="${CFLAGS} -I/boot/system/develop/headers -I/boot/system/develop/headers/posix"
+    LDFLAGS="${LDFLAGS} -L/boot/system/develop/lib"
+    LIBS="${LIBS} -lm -lpthread -lroot -lnetwork -lposix_error_mapper"
+    LINK_DYNAMIC="-rdynamic"
+    add_iconv_lib="-liconv"
+    ;;
   *)
     ;;
 esac


mflatt:
Thanks!

The merge pattern in the original PR left GitHub thinking there were conflicts, so I've pushed a collapsed commit to the branch here. Can you check that it's still the right set of changes. (I went several rounds with `git diff` to try to make sure, but it's easy to go wrong.)

For testing, go to `pkgs/racket-test-core/tests/racket/` and run `racket test.rkt`.

tealeg:
@mflatt those are the right changes still.  Thanks for going through that pain!  

I'll start running the tests and see what's still broken.  Thanks.

tealeg:
Hmm.  It seems to Time out during the threading section.  I will investigate.  

```sh
Section(basic)
Section(unicode)
Section(rx)
Section(reading)
Section(readtable)
Section(printing)
Section(macro)
Section(syntax)
Section(letrec)
Section(procs)
Section(stx)
Section(linklet)
Section(module)
Section(submodule)
Section(stxparam)
Section(numbers)
Section(gambit-numeric)
Section(unsafe)
Section(foreign-inline)
Section(object)
Section(struct)
Section(struct-derived)
Section(threads)


TIMEOUT -- ABORTING!
~/src/3rdparty/racket/pkgs/racket-test-core/tests/racket>
```

tealeg:
I've narrowed the problem down to [this line](https://github.com/racket/racket/blob/50f1f60628c5b50f1aeeca27e50a5af42381731e/pkgs/racket-test-core/tests/racket/thread.rktl#L852). 

Working to understand why now.

tealeg:
> I've narrowed the problem down to [this line](https://github.com/racket/racket/blob/50f1f60628c5b50f1aeeca27e50a5af42381731e/pkgs/racket-test-core/tests/racket/thread.rktl#L852).
> 
> Working to understand why now.

Specifically, if I swap the order of this line and the next one (causing the port to be closed before I attempt to sync) then everything completes.

mflatt:
I may get a chance to investigate the test failure. But whichever of us figures that out, it seems like this PR is a step forward, so I'm inclined to merge it if there are no objections.

tealeg:
> I may get a chance to investigate the test failure. But whichever of us figures that out, it seems like this PR is a step forward, so I'm inclined to merge it if there are no objections.

Yes, with this PR we have Racket working on HaikuOS for x86_64, but we know that there's at least one, subtle bug around threading.  FWIW, if one swaps the lines as I describe above (presumably forcing the `sync` to return the right thing, because the port got closed?), then the rest of the test suite completes.  I'd propose that I just pull this out to a separate issue and continue to work on it there.  

I am still looking at this FWIW, but it will take me a while to be really certain I understand what's going on with the tests and why specifically that's an issue for Haiku.  This is very much a weekend-only activity for me at the moment.

PR #5561, opened 2026-08-14
track literals as `'disappeared-use` in `syntax-case` and alike

usaoc:
## Checklist
- [x] Bugfix

## Description of change
This change makes it so that binding arrows are drawn for literals provided in `syntax-case`.

An additional, related change in this PR makes it so that the binding arrow for `free-identifier=?` is always drawn in `syntax-case*` (previously, the original `free-identifier=?` sometimes won’t appear in the expansion).


mflatt:
The implementation change looks good to me. I expect that the test failure is due to expansion details changing where the regression trace just needs to be updated/regenerated.

usaoc:
Should be passing now.

PR #5563, opened 2026-08-19
schemify: flatten let-values/define-values whose RHS is a direct values call

### LINKED ISSUE #4634: [Feature Request] call-with-values reduction in CP0

V1C70RYG0D:
## Checklist

- [x] Bugfix
- [ ] Feature
- [x] tests included
- [ ] documentation

## Description of change

Follow-up to #5558, moving the fix to the layer @mflatt identified as the
right one there.

Fixes #4634.

`make-let-values` (in `racket/src/schemify/left-to-right.rkt`) always wraps
a multi-value `let-values`/`define-values` binding in `call-with-values`
plus an arity-checking `case-lambda`, even when the right-hand side, after
schemification, is already directly a `(values e ...)` call with exactly
as many results as the binding expects. That arity is known statically in
that case, by construction: the `values` call and the binding come from
the same source-level form, so there's nothing to check at run time, and
the `call-with-values`/`case-lambda` machinery just adds a closure
allocation and a dispatch that a plain `let` can skip entirely:

```racket
(define (test x)
  (define-values (a b) (values (add1 x) (sub1 x)))
  (+ a b))
```

## Fix

Added `find-values-tail`/`rewrap-values-tail`, which look through the
single-binding `let`s that `left-to-right/app` introduces above the
`values` call (to sequence evaluation of non-atomic arguments like
`(add1 x)`) to find a tail `values` call with a matching arity, then bind
each result directly via nested `let`s instead of going through
`call-with-values`. When the arity doesn't match, the check fails and the
existing checked path is used unchanged, so a genuine mismatch still
raises at run time exactly as before.

This was scoped to schemify rather than cp0 (as in #5558) because that's
where the `call-with-values` wrapper and the `(add1 x)` sequencing `let`
actually originate for real Racket programs going through `define-values`.

## Testing

- Verified structurally via `schemify-demo.rkt` on the issue's exact case
  (add1/sub1), its already-working sibling case, a 3-value case with two
  non-atomic arguments needing nested bindings, and a deliberately
  mismatched-arity case (confirmed it still uses the checked
  `call-with-values` path and raises at run time).
- Verified end-to-end via `PLT_LINKLET_SHOW_CP0` on a compiled module:
  `call-with-values` is fully eliminated and runtime results are correct.
- Added tests to `pkgs/racket-test-core/tests/racket/optimize.rktl`'s
  "call-with-values optimization" section, using `define-values`/
  `let-values` (the code path this change actually touches — a
  hand-written `call-with-values` goes through a different schemify
  path and would not exercise this fix). Ran the full file: 1389 tests,
  all passing.
- Two pre-existing tests in that file needed their expectations updated
  to reflect this change:
  - a `let*-values` + `set!` case whose comment noted uncertainty about
    continuation-capture equivalence with a hand-written `let*`; this
    change removes the intermediate producer closure that the
    uncertainty was about, so the two forms are now provably equivalent.
  - a `let-values` case with a module-level `(random)` reference, where
    full inlining down to a single expression no longer happens the same
    way; the value is still bound via a single `let` instead of the
    previous full inlining, with no closure/arity-check overhead either
    way.
- `schemified/*.scm` regenerated via `make derived` per
  `racket/src/cs/README.txt`. This file is machine-generated and the
  diff is mostly gensym renumbering unrelated to the functional change;
  a maintainer regenerating it with their own toolchain may see a
  different (but functionally equivalent) diff.


mflatt:
Thanks, but this patch is not right. Probably AI (I'm assuming) as of now isn't the right tool for this compiler update.

The first problem is that the transformation is wrong, as suggested by one of the tests that was changed. In fairness, `+` cannot actually capture a relevant continuation, but change the example to
```
(module m racket/base
  (provide f)
  (define (f x g)
    (let*-values ([(a b) (values x (g x x))])
      (set! a (add1 a))
      (/ a b))))
```
and use it as 
```
(require 'm)
(call-with-continuation-prompt
 (lambda ()
   (let ([save-k #f])
     (define r
       (f 1
          (lambda (a b)
            (call-with-composable-continuation
             (lambda (k)
               (set! save-k k)
               2)))))
     (cond
       [save-k
        ((begin0 save-k (set! save-k #f)) 2)]
       [else r]))))
```
The answer should be 1, not 3/2.

The other problem is that the new `make-let-values` has the quadratic behavior warned against in #5558.

The change is missing the bigger picture about adjusting the schemify layer. A potential benefit of a (sound) conversion at the schemify layer is to enable more constant folding, copy propagation, and maybe even inlining at that level. But the change here is happening outside the flow of those optimizations, so it helps less than I think a change at this layer should.

mflatt:
Closing with 8817cff25b74d703f6368f9dc1a7c0f3a0fdb8f6 addressing the original issue.

PR #5566, opened 2026-08-21
`rktio` Win32 Concurrency and Other Improvements  

ndykman:
## Checklist

- [ ] Bugfix
- [X] Feature
- [ ] tests included
- [ ] documentation

## Description of change

The use of semaphores has been changed to using critical sections. Critical sections are lighter weight, not shared between processes, and support a variant that does a spin lock before yielding the thread. This is a valuable optimization in some cases that applies to the racket code base. 

In the ChezScheme code, `thread.h` is modified to use a spin count of 4000. This is the value that the Windows heap manager uses and it is a good balance to balance contention versus yielding threads prematurely for operations that are quite short. 

In the `rktio` library, A critical section replaces a semaphore as the global lock for certain initialization tasks. The spin count here is shorter as most paths are longer, but still allows for some benefit in certain cases. 

The previous console code waiting on a console window to be explicitly closed before a program could exit. While this behavior was needed in older versions of Windows, it is no longer needed and can be a real annoyance. The new code removes the semaphore completely and uses an updated API to allocate/free the console window and also insures that a console window does not appear in GUI programs. 

Finally, the code for initializing the `iconv` library in `rktio_convert.c` was cleaned up to greatly reduce the number of branches to speed up the code.


ndykman:
It seems that the CS test has failed (memory access error); I did check run the tests locally before making the pull request. Not sure were the issue lies. Also, the build failed on Linux with a DNS query error. Not sure if this is related or not. 

I have further isolated the error. It seems that calling the `place.rktl` test will cause the `subprocess.rktl` test to fail. Not sure why this is the case with the changes, but this is a start in terms of debugging. 

My apologies in advance, I did think all tests were running without an error, but the raco.exe process was just silently failing on my machine.

ndykman:
There are still too many issues with tests. Closing until proper debugging can be done.

PR #5570, opened 2026-08-24
Update lcm documentation

soegaard:
Racket evaluates `(lcm 0 0.0)` to 0.0.
This matches Chez Scheme and Gambit.


gambiteer:
Gambit GitHub head now has `(lcm 0 1.) => 0`.

soegaard:
@gambiteer Did you change gcd as well?

gambiteer:
`(lcm x y)` is `(* x (quotient y (gcd x y)))`.  If `x` is exact 0, that product should be exact 0.  `gcd` being inexact if one of the arguments is inexact is correct:
```
> (gcd 0 1.)
1.
```

soegaard:
> (lcm x y) is (* x (quotient y (gcd x y))).

That seems an implementation detail for me.

Wikipedia:
> ... lcm(a, b), is the smallest positive integer that is [divisible](https://en.wikipedia.org/wiki/Divisible) by both a and b.
> Since division of integers by zero is undefined, this definition has meaning only if a and b are both different from zero.
> However, some authors define lcm(a, 0) as 0 for all a, since 0 is the only common multiple of a and 0.

LiberalArtist:
From another perspective, in `(gcd 0 0.0)`, the inexact `0.0` may represent a very small positive number, which would indeed mathematically be greater than 0.

In R6RS terms, `(gcd 0 x)` is clearly *not* `0` for every possible exact `x`, so the exception doesn't apply, and we fall back to the general rule to “return an inexact result when any argument is inexact.”

gambiteer:
> Wikipedia:
> 
> > ... lcm(a, b), is the smallest positive integer that is [divisible](https://en.wikipedia.org/wiki/Divisible) by both a and b.
> > Since division of integers by zero is undefined, this definition has meaning only if a and b are both different from zero.
> > However, some authors define lcm(a, 0) as 0 for all a, since 0 is the only common multiple of a and 0.

I don't see that this argues against my position.  Perhaps I don't understand.

soegaard:
When I mentioned `gcd` originally I was thinking of:

   (gcd 1 2.)
   
It currently returns 1.0 but using the same argument, 
one could argue the result should be 1 ?

gambiteer:
I'm sorry, I don't know what "same argument" you're referring to.

soegaard:
The one from Discord, that caused the policy change:

> But this seems like the “One general exception” at the bottom of what you quoted: just as (* 1.0 0) may return exact 0, because (* x 0) would be 0 for any possible exact x, (lcm 0 x) may also return exact 0.

By the same argument `gcd` with an exact 1 argument should return exact 1?

LiberalArtist:
[Edit: I didn't see your reply before writing this.]

> When I mentioned `gcd` originally I was thinking of:
> 
>    (gcd 1 2.)
>    
> It currently returns 1.0 but using the same argument, 
> one could argue the result should be 1 ?
> 

Like @gambiteer, I don't see the argument.

The R6RS formulation we've been talking about for `lcm` would ask if the result of `(gcd 1 x)` for every exact `x` would always be `1`. But the answer to that question is clearly no, per the [docs](https://docs.racket-lang.org/reference/generic-numbers.html#%28def._%28%28quote._~23~25kernel%29._gcd%29%29):

> … for non-integer `n`s, the result is the `gcd` of the numerators divided by the `lcm` of the denominators. …
> Examples: …
> ```racket
> > (gcd 12 81.0)
> 3.0
> > (gcd 1/2 1/3)
> 1/6
> ```

In terms of the R6RS formulation, the fact that `(gcd 1 1/2)` would not be `1` is sufficient to say that the exception to the general rule does not apply.

But returning to `(gcd 1 2.)`, it also makes sense if you take the perspective that inexact `2.` *might* represent mathematical 2, but it equally might represent any of the rational numbers that round to `2.0`, rather than the neighboring 64-bit floats. The fact that `(gcd 1 x)` would be `1` for any exact **integer** `x` does not help us, because we cannot assume that an inexact integer represents an exact integer. So the only way I see to get a well-defined result is to convert  `1` to `1.0` and follow floating-point semantics.

gambiteer:
> The one from Discord, that caused the policy change:
> 
> > But this seems like the “One general exception” at the bottom of what you quoted: just as (* 1.0 0) may return exact 0, because (* x 0) would be 0 for any possible exact x, (lcm 0 x) may also return exact 0.
> 
> By the same argument `gcd` with an exact 1 argument should return exact 1?

Ah, I see, thanks.

My answer: well, maybe.

The fact is that having `(* 0 anything)` and `(/ 0 anything-other-than-exact-0)` evaluate to exact 0, and `(sin 0)` and `(tan 0)` and `(sinh 0)` etc evaluate to exact zero and `(cos 0)` and `(cosh 0)` evaluate to exact 1 is useful in that propagating these rules to, e.g., computing complex products or complex sines/cosines etc.

Maybe I'd take the position that if one finds that `(gcd 1 anything)` being exact 1 is useful when propagating that result through other computations, then it might be a good idea.

soegaard:
Thanks @LiberalArtist 

You convinced me that lcm and gcd isn't as analogous as I thought.

mflatt:
To make sure I'm reading correctly: The conclusion of the discussion was that we really do want to just update the docs and not change any functions, right?

LiberalArtist:
No, I think we want `(lcm 0 0.0)` to produce `0`, as the docs claim. (But no change to `gcd`.)

mflatt:
Ah, I misread

>  Gambit GitHub head now has `(lcm 0 1.) => 0`.

as

>  Gambit GitHub head now has `(lcm 0 1.) => 0.`

mflatt:
Changed the implementation instead of the documentation, after all.

For BC, it turns out that `lcm` with exact `0` and an inexact number did produce exact `0`, except in the case of combination with `0.0`. So, that explains why `lcm` has been documented as treating `0` specially, but the implementation was inconsistent in BC and this was not covered in the test suite.

PR #5571, opened 2026-08-25
Expose contract-random-generate's source of randomness

ndh4:
## Checklist

- [x] Feature
- [x] Tests included
- [x] Documentation

## Description of change

Creates a parameter `contract-random-generate-pseudo-random-generator`, which contract-based generation often uses as a source of randomness in addition to `current-pseudo-random-generator`.

### DIFF
--- a/pkgs/racket-doc/scribblings/reference/contracts.scrbl
+++ b/pkgs/racket-doc/scribblings/reference/contracts.scrbl
@@ -4201,12 +4201,3 @@ ended up returning @racket[contract-random-generate-fail].
 
  @history[#:added "9.3.0.2"]
 }
-
-@defproc[(contract-random-generate-seed [k (integer-in 0 (sub1 (expt 2 31)))])
-         void?]{
-
- Like @racket[random-seed], except that it seeds
- @racket[(current-contract-pseudo-random-generator)]
- rather than the current pseudo-random number generator.
-
- @history[#:added "9.3.0.2"]}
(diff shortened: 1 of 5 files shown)


rfindler:
I think you don't need (want?) to bump the version number for this change. The core racket version number changes often enough that I usually just use the current one for changes like this (so at least the current release will not be new enough).

rfindler:
The "Files Changed" tab looks strange; can you edit the history so that everything is in a single commit and the commit comes after master (that's what we're going to want to merge, anyway).

I think the core of the change we want is to expose a parameter that the random generator uses. That is, to do something like this:

```
(define current-contract-pseudo-random-generator
  (make-parameter (make-pseudo-random-generator)))

(define (rand [x #f]) 
  (if x
      (random x (current-contract-pseudo-random-generator))
      (random (current-contract-pseudo-random-generator))))

(define (rand-seed x)
  (parameterize ([current-pseudo-random-generator current-contract-pseudo-random-generator])
    (random-seed x)))
```

in rand.rkt and then export and document the parameter `current-contract-pseudo-random-generator` (using `@defparam`).

ndh4:
> I think you don't need (want?) to bump the version number for this change. The core racket version number changes often enough that I usually just use the current one for changes like this (so at least the current release will not be new enough).

Got it. I'll lower the version number back down to the current version in the docs.

ndh4:
> The "Files Changed" tab looks strange; can you edit the history so that everything is in a single commit and the commit comes after master (that's what we're going to want to merge, anyway).

Will do.

> I think the core of the change we want is to expose a parameter that the random generator uses.

I agree that it's better to provide a parameter than a raw value. I'll make this change.

rfindler:
Given that `contract-random-generate-seed` is a two line function probably it is a good idea to not export it. I'm actually not even sure where it gets used; maybe nowhere? Anyway, probably good to not add that as an export.

rfindler:
Thanks! I've squashed and pushed.

PR #5575, opened 2026-09-03
Rktio Semaphore and Related Changes

ndykman:
## Checklist

- [ ] Bugfix
- [x] Feature
- [ ] tests included
- [ ] documentation

## Description of change

This is a set of changes around replacing Win32 Semaphore with Win32 CriticalSections and related small changes. This is primary performance based as well as avoiding some bad patterns around loading DLLs.

### DIFF
--- a/racket/src/ChezScheme/c/schsig.c
+++ b/racket/src/ChezScheme/c/schsig.c
@@ -479,40 +479,39 @@ ptr S_save_errno(void) {
     int errno_val;
 
 #ifdef WIN32
-    {
-      ptr tc = get_thread_context();
-      if (CURRENTERRNOSOURCE(tc) == Sfalse) {
-	errno_val = errno;
-      } else if (Schar_value(STRIT(SYMNAME(CURRENTERRNOSOURCE(tc)), 0)) == 'm' /* msvcrt */) {
-	if (!msvcrt_get_errno_ptr) {
-	  HMODULE hm;
-	  get_errno_ptr_t new_get_errno_ptr = NULL;
-	  hm = LoadLibrary("msvcrt.dll");
-	  if (hm)
-	    new_get_errno_ptr = (get_errno_ptr_t)GetProcAddress(hm, "_errno");
-	  if (!new_get_errno_ptr)
-	    new_get_errno_ptr = get_errno_ptr;
-	  while (msvcrt_get_errno_ptr == NULL) {
-	    COMPARE_AND_SWAP_PTR(&msvcrt_get_errno_ptr, NULL, new_get_errno_ptr);
-	  }
-	}
-	errno_val = *(msvcrt_get_errno_ptr());
-      } else {
-	if (!ucrt_get_errno_ptr) {
-	  HMODULE hm;
-	  get_errno_ptr_t new_get_errno_ptr = NULL;
-	  hm = LoadLibrary("ucrtbase.dll");
-	  if (hm)
-	    new_get_errno_ptr = (get_errno_ptr_t)GetProcAddress(hm, "_errno");
-	  if (!new_get_errno_ptr)
-	    new_get_errno_ptr = get_errno_ptr;
-	  while (ucrt_get_errno_ptr == NULL) {
-	    COMPARE_AND_SWAP_PTR(&ucrt_get_errno_ptr, NULL, new_get_errno_ptr);
-	  }
-	}
-	errno_val = *(ucrt_get_errno_ptr());
+  ptr tc = get_thread_context();
+  if (CURRENTERRNOSOURCE(tc) == Sfalse) {
+    errno_val = errno;
+  } else if (Schar_value(STRIT(SYMNAME(CURRENTERRNOSOURCE(tc)), 0)) == 'm' /* msvcrt */) {
+    if (!msvcrt_get_errno_ptr) {
+        HMODULE hm;
+        get_errno_ptr_t new_get_errno_ptr = NULL;
+      hm = GetModuleHandle("msvcrt.dll");
+      if (hm)
+        new_get_errno_ptr = (get_errno_ptr_t)GetProcAddress(hm, "_errno");
+      if (!new_get_errno_ptr)
+      new_get_errno_ptr = get_errno_ptr;
+
+      while (msvcrt_get_errno_ptr == NULL) {
+        COMPARE_AND_SWAP_PTR(&msvcrt_get_errno_ptr, NULL, new_get_errno_ptr);
       }
     }
+  	errno_val = *(msvcrt_get_errno_ptr());
+  } else {
+	  if (!ucrt_get_errno_ptr) {
+	    HMODULE hm;
+	    get_errno_ptr_t new_get_errno_ptr = NULL;
+	    hm = GetModuleHandle("ucrtbase.dll");
+	    if (hm)
+	      new_get_errno_ptr = (get_errno_ptr_t)GetProcAddress(hm, "_errno");
+	    if (!new_get_errno_ptr)
+	      new_get_errno_ptr = get_errno_ptr;
+	    while (ucrt_get_errno_ptr == NULL) {
+	      COMPARE_AND_SWAP_PTR(&ucrt_get_errno_ptr, NULL, new_get_errno_ptr);
+	    }
+	  }
+    errno_val = *(ucrt_get_errno_ptr());
+  }
 #else
     errno_val = errno;
 #endif
--- a/racket/src/rktio/rktio_convert.c
+++ b/racket/src/rktio/rktio_convert.c
@@ -209,7 +209,11 @@ static void init_iconv()
   }
   
   if (!iconv_errno) {
-      iconv_errno = (errno_proc_t) _errno;
+      HANDLE m = LoadLibrary("ucrtbase.dll");
+      if (!m)
+        m = LoadLibrary("msvcrt.dll");
+
+      iconv_errno = (errno_proc_t) GetProcAddress(m, "_errno");
       if (!iconv_errno) {
         iconv = NULL;
         iconv_open = NULL;
--- a/racket/src/rktio/rktio_convert.c
+++ b/racket/src/rktio/rktio_convert.c
@@ -209,11 +209,21 @@ static void init_iconv()
   }
   
   if (!iconv_errno) {
+<<<<<<< HEAD
       HANDLE m = LoadLibrary("ucrtbase.dll");
       if (!m)
         m = LoadLibrary("msvcrt.dll");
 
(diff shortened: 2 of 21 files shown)


ndykman:
Not sure why the tests are failing here when they run on my local machine. I don't want to give the impression that I am making pull requests without doing the basic tests. 

Well, it seems that both my local environment and the build environment are giving the same error. This is hopeful.

ndykman:
This all seems to stem from how `iconv_errno` is setup and used. I still can't tell why this passes in my environment, but fails in this one when the `unicode` tests are run.

ndykman:
So, this seems to be an error that is thrown from the `iconv2.dll` when creating this kind of convertor. I am getting this error locally, so I can debug further.

ndykman:
It seems that touching anything in `rktio_convert.c` isn't a good idea. All tests pass on my local machine. The only change is using a CriticalSection versus a Semaphore. If this breaks on in the github checks, I am really at a loss.

racket/src/rktio/rktio_main.c
```diff
@@ -19,8 +22,9 @@ rktio_t *rktio_init(void)
   memset(rktio, 0, sizeof(rktio_t));
 
 #ifdef RKTIO_SYSTEM_WINDOWS
-  if (!rktio_global_lock)
-    rktio_global_lock = CreateSemaphore(NULL, 1, 1, NULL);
+  BOOL was_created = InterlockedCompareExchangeAcquire(&rktio_global_cs_created, TRUE, FALSE);
+  if (!was_created)
+    InitializeCriticalSectionEx(&rktio_global_cs, CS_SPINCOUNT, CRITICAL_SECTION_NO_DEBUG_INFO);
```
mflatt, line 27: This doesn't seem right. If `rktio_init` could be called in multiple threads, then `was_created` could be set to true before `rktio_global_cs` is initialized. But a constraint on `rktio_init` is that the first call must be non-concurrent with any other call. So, just use `gloabl_cs_created` directly (made `static` and without the `rktio_` prefix, since it shouldn't be sued outside this `.c` file)?

racket/src/rktio/rktio_process.c
```diff
@@ -1177,20 +1177,20 @@ typedef void (*rktio_DeleteProcThreadAttributeList_t)(void *lpAttributeList);
...
-    rktio_UpdateProcThreadAttribute = (rktio_UpdateProcThreadAttribute_t)GetProcAddress(hm, "UpdateProcThreadAttribute");
-    rktio_DeleteProcThreadAttributeList = (rktio_DeleteProcThreadAttributeList_t)GetProcAddress(hm, "DeleteProcThreadAttributeList");
+  // This used to be locked, but there is no need. 
 
-    FreeLibrary(hm);
-  }
-  ReleaseSemaphore(rktio_global_lock, 1, NULL);
+  rktio_InitializeProcThreadAttributeList = (rktio_InitializeProcThreadAttributeList_t)GetProcAddress(hm, "InitializeProcThreadAttributeList");
+  rktio_UpdateProcThreadAttribute = (rktio_UpdateProcThreadAttribute_t)GetProcAddress(hm, "UpdateProcThreadAttribute");
+  rktio_DeleteProcThreadAttributeList = (rktio_DeleteProcThreadAttributeList_t)GetProcAddress(hm, "DeleteProcThreadAttributeList");
```
mflatt, line 1193: We can get rid of `rktio_InitializeProcThreadAttributeList`, etc., and just use `InitializeProcThreadAttributeList`, etc., instead, right?

racket/src/ChezScheme/c/schsig.c
```diff
@@ -479,40 +479,39 @@ ptr S_save_errno(void) {
     int errno_val;
 
 #ifdef WIN32
-    {
```
mflatt, line 482: This looks like probably a good idea, but Chez Scheme changes will need to go through https://github.com/cisco/ChezScheme, and then I keep merged changes in sync. (In very rare cases, we merge Chez Scheme changes here first to try them out, but this doesn't seem like a case where that is needed.)

racket/src/rktio/rktio_convert.c
```diff
@@ -107,10 +107,13 @@ static void init_iconv()
   wchar_t *p;
   int hook_handle = 0;
 
-  WaitForSingleObject(rktio_global_lock, INFINITE);
+  if (iconv_is_ready)
+    return;
+    
+  EnterCriticalSection(&rktio_global_cs);
```
mflatt, line 113: Doesn't the `iconv_is_ready` check need to be within the critical section?

racket/src/rktio/rktio_dll.c
```diff
@@ -206,64 +206,36 @@ rktio_ok_t rktio_dll_close(rktio_t *rktio, rktio_dll_t *dll)
...
-    *lpcbNeeded = i * sizeof(HMODULE);
-    return GetLastError() == ERROR_NO_MORE_FILES;
   }
+
+  HANDLE m = GetModuleHandle("psapi.dll");
+  if (m == NULL)
+    m = LoadLibrary("psapi.dll");
+
+  _EnumProcessModules = (EnumProcessModules_t) GetProcAddress(m, "EnumProcessModules");
+  return _EnumProcessModules(hProcess, lphModule, cb, lpcbNeeded);
```
mflatt, line 238: Is there a reason not to use `EnumProcessModules` directly? Seems like we're way past Windows versions (and headers) where it wasn't available. I guess we may need to add a library to the `windows-libs` list in `build.zuo`.

racket/src/rktio/rktio_fs.c
```diff
@@ -92,14 +92,12 @@ static void init_procs()
...
-    CreateSymbolicLinkProc = (CreateSymbolicLinkProc_t)GetProcAddress(hm, "CreateSymbolicLinkW");
-    DeviceIoControlProc = (DeviceIoControlProc_t)GetProcAddress(hm, "DeviceIoControl");
-    GetFinalPathNameByHandleProc = (GetFinalPathNameByHandle_t)GetProcAddress(hm, "GetFinalPathNameByHandleW");
 
-    FreeLibrary(hm);
+    // All these APIs are available since Windows Vista, so just use 
+    // the direct pointers. 
+    CreateSymbolicLinkProc = (CreateSymbolicLinkProc_t) &CreateSymbolicLinkW;
+    DeviceIoControlProc = (DeviceIoControlProc_t) &DeviceIoControl;
+    GetFinalPathNameByHandleProc = (GetFinalPathNameByHandle_t) &GetFinalPathNameByHandleW;
```
mflatt, line 100: We can get rid of `CreateSymbolicLinkProc`, etc., and just use `CreateSymbolicLinkW`, etc., instead, right?

mflatt (changes requested):

racket/src/rktio/rktio_dll.c
ndykman: I actually agree that we are past versions in which it wasn't available, so calling directly is fine here.

racket/src/rktio/rktio_fs.c
ndykman: Agreed. I didn't know if these needed to be hooked or not.

racket/src/ChezScheme/c/schsig.c
ndykman: I will yank this and submit there.

racket/src/rktio/rktio_convert.c
ndykman: It is there as well on line 115. The idea is not to enter the critical section just to leave right away, so check before entering. Of course, the first thing to do when entering the critical section is to check it as well.

racket/src/rktio/rktio_main.c
ndykman: Oh, I see the convention now for the`rktio_` prefix. The use of `InterlockedCompareExchangeAcquire` insures that the global flag is set atomically. Creating a critical section more than once is really problematic, so enforcing this is important. 

Now, the modern thing to do (but this is a much bigger change) is to use the Win32 `InitOnceExecuteOnce` function for initialization of all resources as this consistently ensures those semantics.

racket/src/rktio/rktio_process.c
ndykman: Agreed.

racket/src/bc/winfig.bat
```diff
@@ -9,8 +9,8 @@ REM Racket, or by "../../build.zuo" after "../../winfig.bat" is used
 
 set SLSP_SUFFIX=
 set ENABLE_CIFY=auto
-set MORE_CFLAGS=
-set MORE_LDFLAGS=
+set MORE_CFLAGS=-DPSAPI_VERSION=2
+set MORE_LDFLAGS=-DPSAPI_VERSION=2
```
mflatt, line 13: These should probably be in "build.zuo", instead. The intent is that anything always needed goes there, while `winfig.bat` has things that make sense to change.

racket/src/worksp/cs/Makefile
```diff
@@ -3,7 +3,7 @@ INCS = /I.. /I..\..\rktio /I..\librktio /I..\..\worksp\cs /I$(SCHEME_DIR)\$(MACH
 
 RKTIO_LIB = ..\..\build\librktio.lib
 BASE_WIN32_LIBS = WS2_32.lib Shell32.lib User32.lib Winmm.lib
-WIN32_LIBS = $(BASE_WIN32_LIBS) RpCrt4.lib Ole32.lib Advapi32.lib
+WIN32_LIBS = $(BASE_WIN32_LIBS) PsApi.lib RpCrt4.lib Ole32.lib Advapi32.lib
```
mflatt, line 6: I think this Makefile isn't used, and it's a leftover from the old system. So, ok to include this change, but also ok to drop it, and I can remove the file separately either way.

mflatt (commented):
Thanks for the revision! Looks like things are moving in a good direction.

racket/src/rktio/rktio_main.c
mflatt: This new version is ok, although using `InterlockedCompareExchangeAcquire` here still seems more than is needed. The `global_cs_created` could just guard the call to `InitializeCriticalSectionEx` and then get set afterward.

racket/src/worksp/cs/Makefile
ndykman: Makes sense to me. This was just for the sake of completeness in case it was used.

racket/src/bc/winfig.bat
ndykman: Totally agree. I still am wrapping my head around how the `build.zuo` works, so if you can make the changes or show me what they need to be, I'd appreciate it.

PR #5581, opened 2026-09-06
Fix `sqlite3-create-function` failing to properly handle NULLs

lexi-lambda:
This fixes #5580 in the obvious way.

PR #5583, opened 2026-09-07
Fix CI failure

sorawee:
Co-authored by GPT-5.6 Sol

## Checklist

- [x] Bugfix
- [ ] Feature
- [ ] tests included
- [ ] documentation

## Description of change

https://github.com/racket/racket/pull/5581 added a test file that [fails the CI](https://github.com/racket/racket/actions/runs/34072960453/job/101593607736). The reason is that it imports `db/unsafe/sqlite3` for `sqlite3-create-function`, but `db/unsafe/sqlite3` is defined in a [different package](https://github.com/racket/db/blob/master/db-lib/db/unsafe/sqlite3.rkt), which is not installed during the test running.

This PR instead directly uses the `unsafe-create-function` method, which is what `sqlite3-create-function` calls under the hood, and can readily be used. This fixes the issue.

### DIFF
--- a/pkgs/racket-test/tests/db/sqlite3.rkt
+++ b/pkgs/racket-test/tests/db/sqlite3.rkt
@@ -1,11 +1,13 @@
 #lang racket/base
 
 (require db/private/pre
-         db/unsafe/sqlite3
+         (only-in db/private/sqlite3/connection
+                  unsafe-create-function)
+         racket/class
          rackunit)
 
 (test-case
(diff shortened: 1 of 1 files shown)


sorawee:
Sigh. That's what happened when I don't test locally.

mflatt:
It might be better to move the test to "db-test", even though it's testing a change in this repo. Otherwise, it seems likely that a new problem test will be added to "racket-tests/tests/db" in the future. @rmculpepper

PR #5586, opened 2026-09-14
Raise the timeout for "fsema-lock.rkt"

samth:


mflatt:
Raising the timeout makes sense, but in this case, I think making the test shorter is probably better. I'll adjust it to perform 1/3 of the work by trying fewer numbers of futures (4, 7, 10, etc., instead of 4, 5, 6, etc.).

